Issue #6067, opened 2026-04-24
[Feature]: Make channel reply-intent precheck configurable (light model + timeout + timing log)


mn13:
### Summary

Let the channel reply-intent precheck use a smaller/faster model, a hard timeout, and emit a timing log — instead of blocking the full agent turn on the main route model with no visibility.

### Problem statement

`classify_channel_reply_intent` in `crates/zeroclaw-channels/src/orchestrator/mod.rs:2026` is a one-shot `REPLY` / `NO_REPLY` classifier that runs before the main agent loop for every inbound channel message. It's called at line 2802 with `route.model` — the same model the main agent uses. There is **no timeout, no timing log, and no config knob** to override the model.

With slow providers like Venice, this precheck silently blocks the entire turn. Field logs:

```
2026-04-24T12:11:42.883205Z  INFO zeroclaw_channels::orchestrator: ⏱️ Memory recall completed mem_recall_ms=0 sender_empty=true group_empty=true
2026-04-24T12:12:40.490502Z  INFO zeroclaw_channels::orchestrator: ⏱️ Starting LLM call elapsed_before_llm_ms=57608
```

That's a 57-second gap with no log lines between them — the precheck alone consumed nearly a minute before the real agent call even started. A yes/no classifier does not need the same reasoning model as the agent; a small fast model (or a routing hint like `hint:fast`) would do.

### Proposed solution

Mirror the existing `ContextCompressionConfig.summary_model` precedent:

- Reference: `crates/zeroclaw-runtime/src/agent/context_compressor.rs:311` — `let summary_model = self.config.summary_model.as_deref().unwrap_or(model);`
- Field: `crates/zeroclaw-config/src/scattered_types.rs:206` — `pub summary_model: Option<String>`

Add a `[agent.precheck]` config block:

```toml
[agent.precheck]
enabled = true                   # kill-switch, default true
model = "hint:fast"              # Option<String>, default None → fall back to route.model
timeout_secs = 5                 # u64, default 5
```

At the call site (`orchestrator/mod.rs:2802`):

1. Resolve `precheck_model = cfg.model.as_deref().unwrap_or(route.model.as_str())` — `hint:` prefixes route through the existing `RouterProvider::resolve` logic, so users can reuse `[[model_routes]]`.
2. Wrap the call in `tokio::time::timeout(Duration::from_secs(cfg.timeout_secs), …)`.
[...]

PR #6945, opened 2026-05-26
feat(agents): add per-agent `classifier_provider` to route reply-intent precheck to a cheaper model

kanmars:
The orchestrator's `classify_channel_reply_intent` currently reuses the main agent's `active_model_provider` and `route.model` for every inbound channel message's REPLY/NO_REPLY classification — meaning operators who run the main agent on an expensive model (e.g. `qwen3.6-plus`) also pay that model's per-token cost for a one-shot Boolean-ish classification on every message.
This patch adds an optional per-agent override:
    [agents.default]
    model_provider      = "custom.default"      # answers (qwen3.6-plus)
    classifier_provider = "custom.kimi-k2-5"    # classifies (kimi-k2.5)
The field is a `ModelProviderRef` (typed alias reference, transparent `String`) — same type as `tts_provider` and `transcription_provider`. Empty value (the default for all existing configs) means "inherit the main agent's resolved route" — exactly the pre-PR behavior.
Wiring
------
  * `AliasedAgentConfig.classifier_provider: ModelProviderRef` added next to `transcription_provider` with `#[serde(default)]`. Backward compatible: any config without the field loads and behaves identically to master.
  * `Config::validate()` extended by adding a single tuple to the existing `typed_provider_refs` validation loop, reusing the same `DanglingReference` / `InvalidFormat` error surfaces that already cover `tts_provider` and `transcription_provider`.
  * New module-private async helper `resolve_classifier_route(ctx, ref_)` in channels/orchestrator/mod.rs. Trims the input first (matching validate's `.trim()`), then resolves through the existing `ModelProviders::find` API and the SHARED `provider_cache`. Every soft-fail path emits a structured WARN before returning `None`: malformed dotted form, unknown alias, missing `model` field, and provider construction errors. The success path emits a structured INFO `"classifier_provider override active"` with `{provider, model}` attrs so operators can verify the override is live without an out-of-band test. The resolver returns a `(provider, model, temperature)` triple — the alias's `temperature` (if set) is threaded through to `classify_channel_reply_intent`, enabling deterministic classification via `temperature = 0.0` on the dedicated alias. Validate already fails loud at config load; the runtime soft-fallback is for hot-reload-induced dangling refs and ensures the agent keeps serving (on the main model) rather than crashing.
[...]

PR #6068, opened 2026-04-24
feat(channels): configurable reply-intent precheck model + timeout

mn13:
## Summary

- Adds `[agent.precheck]` config so the channel reply-intent classifier can route to a lighter/faster model, instead of always using the main route model.
- Wraps the precheck call in a hard `timeout_secs` (default `5`); on timeout or error it fails open to `REPLY` so a slow provider cannot block the whole turn.
- Emits an `elapsed_ms` tracing log — the precheck cost used to be invisible and showed up only as a silent gap between `Memory recall completed` and `Starting LLM call` (see #6067 for the 57-second Venice case).

Pattern mirrors the existing `ContextCompressionConfig.summary_model` precedent (`crates/zeroclaw-runtime/src/agent/context_compressor.rs:311`, `crates/zeroclaw-config/src/scattered_types.rs:206`).

## Configuration

```toml
[agent.precheck]
enabled = true              # default true; set false to skip the precheck entirely
model = "gpt-4o-mini"       # default None → falls back to route.model
timeout_secs = 5            # default 5
```

`model` must be a literal model name served by the **same provider** as the route model. The channel orchestrator builds its provider via `create_resilient_provider_nonblocking` and does **not** wrap it in `RouterProvider`, so `hint:<name>` routing hints are not resolved here — passing one would forward the literal string `"hint:<name>"` to the provider and silently fail open to `REPLY`. Only the runtime agent loop resolves hints today.

## Backward compatibility

Defaults preserve current behavior except for the new 5s timeout — previously a slow precheck would block indefinitely. Operators who want the old indefinite behavior can set `timeout_secs` to a large value.

No schema migration required; the new block is fully optional.

## Test plan

- [x] `cargo check -p zeroclaw-config -p zeroclaw-channels` — clean
- [x] `cargo clippy -p zeroclaw-channels --lib --tests -- -D warnings` — clean
- [x] `cargo test -p zeroclaw-channels --lib -- precheck` — 3/3 pass:
  - `process_channel_message_no_reply_precheck_skips_typing_indicator` (pre-existing)
  - `process_channel_message_precheck_uses_configured_model_override` — provider records the `model` arg passed to `chat_with_system`; with `precheck.model = Some("precheck-fast")` and `ctx.model = "main-route-model"`, asserts the captured value is `"precheck-fast"` and the main agent loop is skipped on `NO_REPLY`
[...]

PR #7003, opened 2026-05-29
fix(channels): restore per-agent reply precheck controls

drbparadise:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Rebased this PR onto current `master` after #6945 landed and resolved the reply-intent precheck conflict.
  - Restores per-agent reply-intent precheck controls under `[agents.<alias>.precheck]` for `enabled` and `timeout_secs`.
  - Keeps model/provider selection on #6945's single source of truth: `[agents.<alias>].classifier_provider`, falling back to the active route provider/model when it is unset or cannot resolve.
  - Drops the separate `precheck.model` surface so users do not have two competing controls for classifier model selection.
  - Adds regression coverage for disabled precheck, timeout fail-open, and `classifier_provider` model selection.
- **Scope boundary:** This PR does not add a second model override, does not add `classifier_provider`, and does not change provider routing beyond composing the restored precheck gate/timeout around #6945's existing classifier route.
- **Blast radius:** `zeroclaw-config` schema/validation/config comments and `zeroclaw-channels` reply-intent precheck path only. No API trait, gateway, security policy, storage, or migration changes.
- **Linked issue(s):** Related #6067, #6068, #6945
- **Labels:** `size: L`, `risk: medium`, `channel`, `config`

Local validation was run from the rebased upstream-based worktree. Commands below are repository-relative.

```bash
cargo test -p zeroclaw-config precheck
# running 2 tests
# test result: ok. 2 passed; 0 failed; 0 ignored; 681 filtered out

cargo test -p zeroclaw-config per_agent_nested_prop_fields_use_agent_alias_paths
# running 1 test
# test result: ok. 1 passed; 0 failed; 0 ignored; 682 filtered out

cargo test -p zeroclaw-channels precheck
# running 4 tests
# test orchestrator::tests::process_channel_message_uses_classifier_provider_for_precheck_model_selection ... ok
# test orchestrator::tests::process_channel_message_skips_reply_intent_classifier_when_agent_precheck_disabled ... ok
# test orchestrator::tests::process_channel_message_precheck_timeout_fails_open_to_reply ... ok
# test result: ok. 4 passed; 0 failed; 0 ignored; 1172 filtered out

cargo fmt --all -- --check
# exit 0

# Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 10s

cargo test
# Finished `test` profile [unoptimized + debuginfo] target(s) in 1m 32s
# test result: ok. 238 passed; 0 failed
# test result: ok. 266 passed; 0 failed
# test result: ok. 192 passed; 0 failed
# test result: ok. 159 passed; 0 failed
# test result: ok. 0 passed; 0 failed; 7 ignored
# test result: ok. 9 passed; 0 failed
# Doc-tests zeroclaw: ok. 0 passed; 0 failed
```

- **Beyond CI - what did you manually verify?** The PR head is rebased on `upstream/master@97b5f9620`; GitHub no longer reports the PR as `DIRTY`. The diff is limited to `crates/zeroclaw-config/src/schema.rs`, `crates/zeroclaw-config/src/scattered_types.rs`, and `crates/zeroclaw-channels/src/orchestrator/mod.rs`. I verified there is no `precheck.model` surface left in the changed config/channel paths.

- New permissions, capabilities, or file system access scope? **No**.
- New external network calls? **No**. The existing reply-intent classifier call remains the only model call in this path; this PR restores per-agent controls for whether that call runs and how long it can block.
- Secrets / tokens / credentials handling changed? **No**.

## Compatibility (required)

- Backward compatible? **Yes**.
- Config / env / CLI surface changed? **Yes**, additive only.
- If `No` or `Yes` to either: Existing configs keep current behavior because `[agents.<alias>.precheck]` defaults to enabled and the existing classifier route remains active. Operators can set `[agents.<alias>.precheck].enabled = false` to skip the classifier for one agent, or set `[agents.<alias>.precheck].timeout_secs` to bound classifier latency. Model/provider selection stays on `[agents.<alias>].classifier_provider`; no `precheck.model` migration is needed.

- **Feature flags or config toggles:** Set `[agents.<alias>.precheck].enabled = false` to skip the classifier for an agent.
- **Observable failure symptoms:** Unexpected `reply-intent precheck skipped`, `reply-intent precheck failed open`, or `reply-intent precheck timed out; failing open` logs.

- Superseded PRs + authors (`#<pr> by @<author>`, one per line): N/A.
- Scope materially carried forward: N/A.
- `Co-authored-by` trailers added in commit messages for incorporated contributors? **No**.
- If `No`, why (inspiration-only, no direct code/design carry-over): This PR does not use `Supersedes #` and does not carry code from another contributor PR.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -3676,31 +3676,158 @@ async fn process_channel_message_body(
     let explicit_channel_address =
         is_explicitly_addressed_channel_message(&msg.channel, &msg.content);
     let classifier_intent = if explicit_channel_address {
+        ::zeroclaw_log::record!(
+            DEBUG,
+            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Skip).with_attrs(
+                ::serde_json::json!({
+                    "sender": msg.sender,
+                    "channel": msg.channel,
+                    "reason": "explicit_channel_address",
+                })
+            ),
+            "reply-intent precheck skipped"
+        );
         AssistantChannelOutcome::Reply(String::new())
     } else {
-        let (classifier_provider_arc, classifier_model_owned, classifier_temperature): (
-            Arc<dyn ModelProvider>,
-            String,
-            Option<f64>,
-        ) = resolve_classifier_route(ctx.as_ref(), &ctx.agent_cfg.classifier_provider)
-            .await
-            .unwrap_or_else(|| {
+        let precheck_cfg = ctx.agent_cfg.precheck.clone();
+        if !precheck_cfg.enabled {
+            ::zeroclaw_log::record!(
+                INFO,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Skip)
+                    .with_attrs(::serde_json::json!({
+                        "sender": msg.sender,
+                        "channel": msg.channel,
+                        "agent": ctx.agent_alias.as_str(),
+                        "model_provider": route.model_provider.as_str(),
+                        "model": route.model.as_str(),
+                        "reason": "disabled_by_agent_config",
+                    })),
+                "reply-intent precheck skipped"
+            );
+            AssistantChannelOutcome::Reply(String::new())
+        } else {
+            let classifier_route =
+                resolve_classifier_route(ctx.as_ref(), &ctx.agent_cfg.classifier_provider).await;
+            let classifier_provider_name = if classifier_route.is_some() {
+                ctx.agent_cfg
+                    .classifier_provider
+                    .as_str()
+                    .trim()
+                    .to_string()
+            } else {
+                route.model_provider.clone()
+            };
+            let (classifier_provider_arc, classifier_model_owned, classifier_temperature): (
+                Arc<dyn ModelProvider>,
+                String,
+                Option<f64>,
+            ) = classifier_route.unwrap_or_else(|| {
                 (
                     Arc::clone(&active_model_provider),
                     route.model.clone(),
                     None,
                 )
             });
+            let precheck_timeout = Duration::from_secs(precheck_cfg.timeout_secs);
+            let precheck_start = Instant::now();
+            ::zeroclaw_log::record!(
+                DEBUG,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_attrs(::serde_json::json!({
+                        "sender": msg.sender,
+                        "channel": msg.channel,
+                        "agent": ctx.agent_alias.as_str(),
+                        "model_provider": classifier_provider_name.as_str(),
+                        "model": classifier_model_owned.as_str(),
+                        "timeout_secs": precheck_cfg.timeout_secs,
+                    })),
+                "reply-intent precheck started"
+            );
 
-        classify_channel_reply_intent(
-            classifier_provider_arc.as_ref(),
-            history[0].content.as_str(),
-            &history,
-            classifier_model_owned.as_str(),
-            classifier_temperature.or(runtime_defaults.temperature),
-        )
-        .await
-        .unwrap_or(AssistantChannelOutcome::Reply(String::new()))
+            match tokio::time::timeout(
+                precheck_timeout,
+                classify_channel_reply_intent(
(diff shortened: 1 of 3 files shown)


Audacity88 (changes requested):
@drbparadise, I checked the live PR state at c81cdfe, the PR body validation, the changed config/schema/orchestrator files, and current `master` after #6945. I did not run local tests because GitHub now reports this branch as conflicting, and the conflict is in the same reply-intent precheck path this PR changes.

### 🟢 What looks good — per-agent precheck control is the right surface

Restoring the controls under `[agents.<alias>.precheck]` makes sense for the v0.8 config shape. The schema validation for zero timeouts and blank model names is useful, and the channel tests pin the important runtime behavior: disabled precheck skips the classifier, timeout fails open, and an explicit precheck model changes the classifier call.

### 🔴 Blocking — this needs to compose with the merged classifier_provider route

Current `master` now includes #6945, which added per-agent `classifier_provider` and `resolve_classifier_route(...)` so the reply-intent precheck can use a separate provider entry with its own model, temperature, API key, and provider construction. This PR was cut before that landed. In this branch, `process_channel_message_body` still calls `classify_channel_reply_intent(active_model_provider.as_ref(), ..., precheck_model, runtime_defaults.temperature)`.

If we resolve the current conflict by taking this branch's precheck block as-is, we either drop #6945's provider-level routing or end up with two competing controls for the same classifier selection: `agents.<alias>.classifier_provider` as a full provider ref, and `agents.<alias>.precheck.model` as a same-provider literal model override. That would make the config harder to reason about and could silently route the precheck to the wrong provider after users configure the new #6945 field.

Please rebase onto current `master` and make the controls compose. `precheck.enabled` and `precheck.timeout_secs` should gate and wrap the existing classifier route. Model/provider selection should preserve #6945's provider-ref source of truth. If `precheck.model` is still intended to stay, please define its precedence against `classifier_provider` and add a regression that covers the combination. Otherwise, I think dropping `precheck.model` and using `classifier_provider` for model selection is cleaner.

After that rebase, the old validation should be refreshed because it predates #6945 landing on `master`.

drbparadise:
Rebased onto current `master` and resolved the precheck/classifier_provider conflict.

What changed in this update:
- `precheck.enabled` now gates the reply-intent classifier before route resolution.
- `precheck.timeout_secs` wraps the resolved classifier route, including `classifier_provider`.
- Removed the separate `precheck.model` surface; model/provider selection stays on `[agents.<alias>].classifier_provider`.
- Added regression coverage for disabled precheck, classifier timeout fail-open, and classifier_provider model selection.

Validation refreshed on the rebased head:
- `cargo test -p zeroclaw-config precheck`
- `cargo test -p zeroclaw-config per_agent_nested_prop_fields_use_agent_alias_paths`
- `cargo test -p zeroclaw-channels precheck`
- `cargo test`

Audacity88 (approved):
@drbparadise, I re-reviewed current head 99e99e after the rebase and #6945 composition update. I checked the PR body, top-level comment, prior Audacity88 review, inline comments, linked issues #6067/#6068, the current diff, and the live check state. I did not rerun local cargo because the PR now includes refreshed focused and broad local validation, and the visible GitHub Quality Gate jobs are green.

### ✅ Resolved — precheck controls now compose with `classifier_provider`

The previous blocker is resolved. `precheck.enabled` now gates the reply-intent classifier before route resolution, `precheck.timeout_secs` wraps the resolved classifier route, and model/provider selection stays on `[agents.<alias>].classifier_provider` with fallback to the active route. Dropping the separate `precheck.model` surface was the right call here; it avoids two competing controls for the same classifier decision.

### 🟢 What looks good — the regression coverage pins the risky paths

The new tests cover the paths that mattered for this rebase: disabled precheck skips the classifier, timeout fails open into the main agent loop, and an agent `classifier_provider` drives the precheck model selection. The config tests also pin the per-agent prop path and zero-timeout validation, which protects the new `[agents.<alias>.precheck]` surface from drifting back toward the old global-agent prefix.

### 🟢 What looks good — the PR body now matches the current landing shape

The body clearly says this restores `enabled` and `timeout_secs`, keeps model/provider selection on #6945’s `classifier_provider`, and does not close #6067 outright. That is the right public record for this slice: it fixes the timeout/control gap without reopening the older same-provider `precheck.model` surface from #6068.

Issue #6819, opened 2026-05-21
[Feature]: File/attachment upload protocol


singlerider:
## File/attachment upload protocol for RPC transport

Allows clients (TUI, web dashboard) to send binary file content to the daemon — images for multimodal prompts, documents for agent context. Base64-in-JSON encoding within the existing NDJSON protocol.

### User story

TUI user composing a message can:
1. **Clipboard paste** — paste MIME data, TUI uploads base64, daemon returns marker like `[IMAGE from clipboard]`
2. **Drag-and-drop** — drag file from explorer, TUI sends path (Unix socket) or base64 (WSS), marker like `[DOCUMENT:/path.pdf]`
3. **Multiple files** — batch upload in one `file/attach` call
4. **Deduplication** — identical bytes (SHA-256) within a session return same ref without storing twice

### Wire format

Two modes per file entry:
- **Data mode** (`data_b64`): client sends base64-encoded bytes. Required for clipboard and WSS clients.
- **Path mode** (`path`): absolute path on daemon filesystem. Unix socket only — rejected over WSS.

Methods:
- `file/attach` — batch upload files to a session, returns ref_id + marker per file
- `session/prompt` — optional `attachments` array (same format), markers appended to prompt

### Status

- [x] Design: base64-in-JSON encoding (a93b09711)
- [x] Wire format specification — `file/attach` with batch support (a93b09711)
- [x] Transport-layer support — no transport changes needed, base64 is a JSON string (a93b09711)
- [x] `file/attach` RPC method with path mode + data mode (a93b09711)
- [x] Size limits: 10MB per file, 20MB per request (a93b09711)
- [x] Integration with agent workspace — uploads stored at `{workspace}/uploads/{sha256_prefix}.{ext}` (a93b09711)
- [x] SHA-256 content-addressed deduplication per session (a93b09711)
- [x] Inline attachments on `session/prompt` (a93b09711)
- [x] Unit tests (14 tests covering all paths) (a93b09711)
- [x] TUI clipboard/drag-drop UI (#6824/#6825)
- [x] Multimodal pipeline resolution of `[IMAGE from clipboard]` via upload index

### Future extensions (not in scope)

- Chunked upload for very large files
- `file/download` for serving workspace files back to client
- Binary frame transport extension (#6293)

### Referenced by

- #6837 — "Binary frame support for file attachments — deferred to #6819"

singlerider:
Parent tracker: #6826

Issue #6074, opened 2026-04-24
audit: track 153 commits lost in bulk revert c3ff635 for recovery


Audacity88:
## Summary

Commit c3ff635300c03132441cbf9ae9ad0477a6e11ddb (2026-03-28) reverted 153 commits in a single bulk rollback, restoring master to 1362d69f. This revert was necessary at the time, but it also removed bug fixes, features, and improvements that had been reviewed, approved, and merged.

This issue tracks which commits have already been independently re-applied during the microkernel refactor and which are still missing and worth recovering.

**Known recovery issues:** #5259 (WeChat iLink channel), #2503 (NapCat/OneBot channel)

**Note on deleted PRs:** Some PR numbers referenced in commit messages return 404 on GitHub. A co-author on several of these commits (SimianAstronaut7, who also authored the revert) had their account blocked, which removed associated PRs and issues. Some external contributor PRs that list SimianAstronaut7 as co-author are also missing — likely opened from that account's fork. Where the PR record is missing, the entry links to the commit instead.

## Status key

- ✅ **Re-applied** — verified present in the current codebase
- 🔲 **Missing** — verified absent, candidate for recovery
- 🔀 **Superseded** — the underlying code was refactored or the bug no longer applies in the new architecture
- ⏭️ **Skip** — merge commit, dep bump, or whitespace-only

---

## Verified re-applied (21)

These fixes have been independently re-implemented in the current crate structure:

- [x] `2ddf7dcb` #4544 — web_search_tool in default_param_for_tool mapping (tool-call-parser/lib.rs:807) ✅
- [x] `49e994a3` #4702 — Lark list message type handling (lark.rs:986, test at :2766) ✅
- [x] `ffb015b3` #4552 — case-insensitive command allowlist matching (policy.rs:824) ✅
- [x] `143b34ea` ([commit](https://github.com/zeroclaw-labs/zeroclaw/commit/143b34ea595ca47c9cc94dd12534959f9552123f)) — SearXNG search provider config (schema.rs:2977) ✅
- [x] `a4e6d772` #4314 — Slack audio file transcription (slack.rs:39) ✅
- [x] `a3da42b1` — mention_only arg + voice_wake feature-gated errors (telegram.rs:367) ✅
- [x] `2a5866ec` #4297 — extract binary from .tar.gz release assets (update.rs:212) ✅
- [x] `490c54e8` #4750 — allow high risk command execution (whitespace-only, no functional change) ✅
[...]

PR #7004, opened 2026-05-29
feat(tools): optional base64 encoding for file_read / file_write

metalmon:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds an optional `encoding` parameter (`"utf8"` default | `"base64"`) to the `file_read` and `file_write` tools.
  - Today the file tools are text/UTF-8 only — `file_write` writes the `content` string verbatim and `file_read` returns UTF-8 — so no binary format can be read or written (office docs `.xlsx`/`.docx` are ZIP+XML, images, archives, etc.). The plugin protocol even lists `file_read`/`file_write` as *“not yet implemented.”*
  - `"base64"` reads raw bytes and returns them base64-encoded, and decodes base64 `content` into raw bytes on write. This is the minimal general primitive that lets agents and pure (FS-less) plugins produce/consume real binary artifacts without an interpreter escape hatch (Python/Node via shell) or lossy text formats (CSV/HTML).
  - `file_write` validates the `encoding` and decodes base64 **before any filesystem mutation** (including parent-directory creation), so rejected input (invalid base64 / unsupported encoding) fails without touching the workspace. The existing path-sandbox checks still run on the resolved target before the write.
  - Default `"utf8"` preserves current behavior byte-for-byte.
- **Scope boundary:** No new tools, no new plugin host functions. Adds an opt-in binary read/write capability to the existing file tools, confined to the existing path-sandboxing, allowlist, symlink, and autonomy guards. Does not touch the client→daemon upload protocol.
- **Blast radius:** `file_read` (`zeroclaw-runtime`) and `file_write` (`zeroclaw-tools`) only. Any caller passing no `encoding` is unaffected.
- **Linked issue(s):** Related #6819 (file/attachment *upload* protocol — a different layer: client→daemon ingestion; this PR is about the agent-facing read/write tools operating on bytes already in the workspace).
- **Labels:** `enhancement`, `risk: high`, `size: S`, `runtime`, `tool`, `tool:file`.

Run on Windows (`x86_64-pc-windows-msvc`); scoped to the two changed crates.

`cargo fmt --all` — applied, no diff to changed files.

`cargo test -p zeroclaw-runtime --lib file_read`:

```
running 27 tests
test tools::file_read::tests::file_read_base64_returns_encoded_bytes ... ok
test tools::file_read::tests::file_read_schema_has_encoding ... ok
test tools::file_read::tests::file_read_unsupported_encoding_errors ... ok
...
test result: ok. 26 passed; 0 failed; 1 ignored; 0 measured; 1824 filtered out
```

`cargo test -p zeroclaw-tools --lib file_write` — base64 + write-ordering tests pass:

```
test file_write::tests::file_write_base64_writes_decoded_bytes ... ok
test file_write::tests::file_write_base64_invalid_content_errors ... ok
test file_write::tests::file_write_base64_still_blocks_path_traversal ... ok
test file_write::tests::file_write_unsupported_encoding_errors ... ok
test file_write::tests::file_write_rejected_encoding_does_not_create_parent_dirs ... ok
test file_write::tests::file_write_schema_has_encoding ... ok
```

- **Beyond CI — what did you manually verify?** base64 round-trips exact non-UTF-8 bytes on both read and write; invalid base64 and unsupported encoding values return clean error results (no panic, no partial write) **and create neither the target file nor a missing parent directory**; path-traversal is still blocked in base64 write mode.
- **Pre-existing failures observed (NOT introduced by this PR, not in changed files):** two `file_write` tests fail only on Windows because they hardcode Unix paths (`file_write_blocks_absolute_path` asserts a Unix-specific error string; `file_write_normalizes_workspace_prefixed_relative_path` calls `strip_prefix("/")`). `cargo clippy --all-targets -- -D warnings` fails on a pre-existing `unreachable_code`/`unused_variable` in `glob_search.rs` (unrelated file). Recommend a final `dev/ci.sh all` run on Linux CI.

- New permissions, capabilities, or file system access scope? **Yes — a new capability, not a new scope.** This adds an opt-in binary read/write capability to the existing `file_read`/`file_write` tools (`encoding: "base64"`). It does **not** broaden path permissions or workspace scope: all I/O remains confined to the existing file-tool path guards (sandbox, allowlist, symlink-escape refusal, read-only autonomy). Rejected base64 writes now fail before any filesystem mutation.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No.** Tests use synthetic byte arrays.

## Compatibility (required)

- Backward compatible? **Yes.** `encoding` is optional and defaults to `"utf8"` (existing behavior).
- Config / env / CLI surface changed? **No.** Only the two tools' JSON parameter schemas gain an optional field.
- Upgrade steps: none.

## Rollback (required for risk: medium and high)

- **Fast rollback command/path:** `git revert <sha>` (single commit, two files).
- **Feature flags or config toggles:** None (opt-in per call via `encoding`).
- **Observable failure symptoms:** `file_read`/`file_write` returning `Invalid base64 content` or `Unsupported encoding` errors, or binary files written with wrong bytes.

### DIFF
--- a/crates/zeroclaw-runtime/src/tools/file_read.rs
+++ b/crates/zeroclaw-runtime/src/tools/file_read.rs
@@ -59,7 +59,7 @@ impl Tool for FileReadTool {
     }
 
     fn description(&self) -> &str {
-        "Read file contents with line numbers. Supports partial reading via offset and limit. Extracts text from PDF; other binary files are read with lossy UTF-8 conversion."
+        "Read file contents with line numbers. Supports partial reading via offset and limit. Extracts text from PDF; other binary files are read with lossy UTF-8 conversion. Set encoding=\"base64\" to return raw bytes base64-encoded (for binary files such as .xlsx/.docx); offset/limit are ignored in that mode."
     }
 
     fn parameters_schema(&self) -> serde_json::Value {
@@ -72,11 +72,16 @@ impl Tool for FileReadTool {
                 },
                 "offset": {
                     "type": "integer",
-                    "description": "Starting line number (1-based, default: 1)"
+                    "description": "Starting line number (1-based, default: 1). Ignored when encoding is 'base64'."
                 },
                 "limit": {
                     "type": "integer",
-                    "description": "Maximum number of lines to return (default: all)"
+                    "description": "Maximum number of lines to return (default: all). Ignored when encoding is 'base64'."
+                },
+                "encoding": {
+                    "type": "string",
+                    "enum": ["utf8", "base64"],
+                    "description": "Output encoding (default: 'utf8'). Use 'base64' to read binary files as base64-encoded bytes."
                 }
             },
             "required": ["path"]
@@ -166,6 +171,41 @@ impl Tool for FileReadTool {
                     error: Some(format!("Failed to read file metadata: {e}")),
                 });
             }
+        }
+
+        let encoding = args
+            .get("encoding")
+            .and_then(|v| v.as_str())
+            .unwrap_or("utf8");
+
+        if encoding == "base64" {
+            // Binary read: return raw bytes base64-encoded. Line numbering and
+            // offset/limit are text concepts and do not apply here.
+            let bytes = match tokio::fs::read(&resolved_path).await {
+                Ok(b) => b,
+                Err(e) => {
+                    return Ok(ToolResult {
+                        success: false,
+                        output: String::new(),
+                        error: Some(format!("Failed to read file: {e}")),
+                    });
+                }
+            };
+            use base64::Engine;
+            let encoded = base64::engine::general_purpose::STANDARD.encode(&bytes);
+            return Ok(ToolResult {
+                success: true,
+                output: encoded,
+                error: None,
+            });
+        } else if encoding != "utf8" {
+            return Ok(ToolResult {
+                success: false,
+                output: String::new(),
+                error: Some(format!(
+                    "Unsupported encoding '{encoding}' (expected 'utf8' or 'base64')"
+                )),
+            });
         }
 
         match tokio::fs::read_to_string(&resolved_path).await {
@@ -427,6 +467,63 @@ mod tests {
         let tool = test_tool(std::env::temp_dir());
         let result = tool.execute(json!({})).await;
         assert!(result.is_err());
+    }
+
+    #[test]
+    fn file_read_schema_has_encoding() {
+        let tool = test_tool(std::env::temp_dir());
+        let schema = tool.parameters_schema();
+        assert!(schema["properties"]["encoding"].is_object());
+    }
+
+    #[tokio::test]
+    async fn file_read_base64_returns_encoded_bytes() {
+        let dir = std::env::temp_dir().join("zeroclaw_test_file_read_base64");
+        let _ = tokio::fs::remove_dir_all(&dir).await;
+        tokio::fs::create_dir_all(&dir).await.unwrap();
+
+        // Non-UTF-8 bytes — proves we return raw bytes, not lossy text.
+        let raw: Vec<u8> = vec![0x00, 0x80, 0xFF, 0xFE, b'P', b'K', 0x03, 0x04];
+        tokio::fs::write(dir.join("data.bin"), &raw).await.unwrap();
+
+        let tool = test_tool(dir.clone());
+        let result = tool
+            .execute(json!({"path": "data.bin", "encoding": "base64"}))
+            .await
+            .unwrap();
+        assert!(result.success, "error: {:?}", result.error);
+
+        use base64::Engine;
+        let decoded = base64::engine::general_purpose::STANDARD
+            .decode(result.output.trim())
+            .expect("output must be valid base64");
+        assert_eq!(decoded, raw, "base64 read must round-trip exact bytes");
+
+        let _ = tokio::fs::remove_dir_all(&dir).await;
+    }
+
+    #[tokio::test]
+    async fn file_read_unsupported_encoding_errors() {
+        let dir = std::env::temp_dir().join("zeroclaw_test_file_read_bad_encoding");
+        let _ = tokio::fs::remove_dir_all(&dir).await;
+        tokio::fs::create_dir_all(&dir).await.unwrap();
+        tokio::fs::write(dir.join("f.txt"), "hi").await.unwrap();
+
+        let tool = test_tool(dir.clone());
+        let result = tool
+            .execute(json!({"path": "f.txt", "encoding": "hex"}))
+            .await
+            .unwrap();
+        assert!(!result.success);
+        assert!(
+            result
+                .error
+                .as_deref()
+                .unwrap_or("")
+                .contains("Unsupported encoding")
+        );
+
+        let _ = tokio::fs::remove_dir_all(&dir).await;
     }
 
     #[tokio::test]
--- a/crates/zeroclaw-tools/src/file_write.rs
+++ b/crates/zeroclaw-tools/src/file_write.rs
@@ -22,7 +22,7 @@ impl Tool for FileWriteTool {
     }
 
     fn description(&self) -> &str {
-        "Write contents to a file in the workspace"
+        "Write contents to a file in the workspace. Text by default; set encoding=\"base64\" to write binary files (e.g. .xlsx/.docx) by decoding base64 content into raw bytes."
     }
 
     fn parameters_schema(&self) -> serde_json::Value {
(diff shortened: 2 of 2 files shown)


Audacity88 (changes requested):
Reviewed current head ccfc4007 against the live PR body, the PR body's Related #6819 note, green GitHub checks, the no-comment/no-review PR history, and the `file_read` / `file_write` diff plus adjacent source context. I did not run local validation for this review because the finding is visible from the write-path ordering and the public CI matrix is already green. This is a request-changes review: the base64 direction looks useful and small, but one write-side ordering issue should be fixed before merge.

### 🟢 What looks good — The core shape is narrow

The visible base64 schema and round-trip tests are the right kind of coverage, and the default `utf8` behavior remains opt-out rather than changed underneath existing callers. The write path also still shows the runtime-config and symlink refusal checks before the final write. Once the rejected-write side effect and public wording issues are fixed, this should be close.

### 🔴 Blocking — Rejected base64 writes can still create directories

`FileWriteTool::execute` resolves the target parent and calls `create_dir_all(parent)` before it validates the new `encoding` value or decodes base64 content. In the PR head, parent directory creation happens before the new encoding/decode block in `crates/zeroclaw-tools/src/file_write.rs`. That means a rejected call such as `{"path":"nested/out.bin","content":"not!valid!","encoding":"base64"}` or `{"path":"nested/out.bin","content":"hi","encoding":"hex"}` can create `nested/` and then return an error. For a file-write tool, invalid input should fail without mutating the workspace, especially when the PR body describes invalid base64 as a clean error with no partial write.

Please validate the supported encoding and decode base64 content before any write-side filesystem mutation such as parent directory creation, while keeping the existing path-sandbox checks intact. I would also add a regression where invalid base64 and unsupported encoding target a missing nested parent and assert that neither the file nor the new parent directory is created.

### 🟡 Warning — Security impact should name the new capability

The Security & Privacy section currently answers "New permissions, capabilities, or file system access scope?" as `No`. I agree this does not broaden path permissions or workspace scope, but it does add a new binary read/write capability to existing tools. Please change that answer to `Yes` and say that the new capability is limited to the existing file-tool path guards and opt-in `encoding: "base64"` calls.

### 🟡 Warning — Drop the bot attribution footer while updating the branch

The current commit message includes `Co-authored-by: Cursor <cursoragent@cursor.com>`. The PR template asks contributors not to add bot/AI attribution footers to PR bodies or commit-message tails. Since this branch already needs another update for the write-side fix, please remove that footer in the refreshed commit history as part of the same pass.

Audacity88:
@singlerider @theonlyhennygod — milestone alignment needed: this PR is related to #6819, but it explicitly frames itself as the agent-facing `file_read` / `file_write` binary I/O layer rather than the client-to-daemon upload protocol that #6819 tracks under `v0.8.1-tui`. I do not see a clear open milestone fit for that scope. Please advise on placement or deferral.

metalmon:
Thanks for the review — addressed all three points:

**🔴 Rejected base64 writes creating directories** — fixed. `file_write` now validates the `encoding` and decodes base64 **before any filesystem mutation** (parent-directory creation included); the existing path-sandbox checks still run on the resolved target before the write. Added a regression test `file_write_rejected_encoding_does_not_create_parent_dirs` asserting that both invalid base64 and an unsupported encoding targeting a missing nested parent create neither the file nor the parent directory.

**🟡 Security wording** — updated the Security & Privacy section: the answer is now **Yes**, framed as a new opt-in binary read/write *capability* on the existing tools (not a new path/scope), confined to the existing file-tool path guards.

**🟡 Bot attribution footer** — removed; the branch was force-pushed with a clean commit message.

On milestone alignment: confirming this PR is **not** part of #6819's client→daemon upload protocol (`v0.8.0-tui`). It's the agent-facing `file_read`/`file_write` binary-I/O layer that operates on bytes already in the workspace — it complements that protocol rather than implementing it, and the PR body now states this boundary explicitly. Since it doesn't fit under #6819's milestone, I'll leave placement entirely to your call — happy to apply whichever milestone (or none) you think suits a small, additive, opt-in tool-capability change.

Audacity88:
@metalmon — I checked this against the #6074 recovery audit, specifically the reverted #4320 `file_read` robustness fix.

This PR already solves an important adjacent problem: `encoding: "base64"` gives agents an explicit way to read and write binary bytes without going through lossy text. That is worth keeping.

There is one remaining overlap with #4320: the default `file_read` path still lossy-decodes non-PDF binary data when callers do not pass `encoding: "base64"`. The reverted #4320 behavior rejected binary/image reads by default and pointed image reads toward `image_info`, which avoids silently filling the model context with replacement-character text or accidental binary junk.

Would you be open to folding that default-mode behavior into this PR? The shape I have in mind is small:

- keep `encoding: "base64"` as the explicit binary path;
- keep normal UTF-8 text reads unchanged;
- in default `encoding: "utf8"` mode, reject known image/binary data instead of falling back to lossy UTF-8;
- keep PDF text extraction as the special allowed binary-ish case.

That would let this PR cover both the new binary I/O capability and the missing #4320 safety/UX fix in one coherent file-tools change. If you prefer to keep this PR strictly additive, that is also workable; I can open a separate #6074 recovery PR after this one lands.

metalmon:
Thanks — glad the `encoding: "base64"` binary I/O capability is worth keeping.

On folding the reverted #4320 default-mode behavior into this PR: I'd prefer to keep this one strictly additive, and have you (or me) open the separate #6074 recovery PR after this lands. Reasoning:

- **One concern per PR.** This change is a small, purely opt-in capability that doesn't alter any existing call site's behavior — default `utf8` reads return exactly what they returned before. The #4320 recovery is a *behavior change* to the default path (rejecting binary/image instead of lossy-decoding). Bundling an additive capability with a default-behavior change widens the blast radius and the review surface, and couples two independently-revertable concerns.
- **Independent rollback.** If the default-mode rejection ever needs to be reverted (it was reverted once already, via #6074), keeping it separate means that revert doesn't take the base64 capability down with it.
- **Cleaner milestone/triage story.** The default-mode rejection is squarely a #6074 recovery item with its own history (#4320); it deserves its own PR record, body, and validation evidence rather than riding in on a file-tools capability PR.

The shape you described is exactly right for that follow-up — keep `encoding: "base64"` as the explicit binary path, keep UTF-8 text reads unchanged, reject known image/binary in default `utf8` mode, and keep PDF text extraction as the allowed binary-ish case. I'm happy to author that recovery PR on top of this one once it merges (or defer to you if you'd rather own the #6074 thread). Either way I'll reference this PR as the base so the two changes compose cleanly.

Audacity88 (approved):
Context: I re-reviewed current head 20c235a against the live PR body, Related #6819, #6074's #4320 recovery note, my prior request-changes review, the author follow-up comments, current checks, the full two-file diff, and the PR-head source around `file_read`, `file_write`, and the production `PathGuardedTool` / `RateLimitedTool` wrapper stack. I did not rerun local cargo; I am relying on the now-green GitHub checks plus the author-provided focused Windows validation for this pass.

@metalmon, thanks for keeping this slice narrow. The current head addresses my previous blocker and the two public-record warnings.

### ✅ Resolved — rejected base64 writes no longer create directories

`file_write` now validates `encoding` and decodes base64 content before any write-side filesystem mutation. That fixes the side effect I blocked on: invalid base64 or an unsupported encoding can now fail before parent-directory creation, and the new regression covers both rejected cases against missing nested parents.

The existing guard shape is still intact. The production wrapper stack checks the path argument before the concrete tool runs, and the concrete tool still resolves the target, creates and canonicalizes the parent, enforces the resolved path policy, blocks runtime-config writes, refuses symlink targets, and only then writes the decoded bytes.

### ✅ Resolved — the security section now names the new capability

The PR body now answers the capability question as `Yes` and frames this as an opt-in binary read/write capability on the existing tools, not as a broader path or workspace permission. That matches the diff: callers get a new `encoding: "base64"` mode, while the existing file-tool path guards remain the boundary.

### ✅ Resolved — bot attribution is gone from the commit tail

The refreshed commit no longer carries the bot co-author footer I called out. The commit message now stays focused on the file-tool capability and the mutation-order guarantee.

### 🟢 What looks good — additive binary I/O with follow-up scope kept separate

I agree with keeping the reverted #4320 default-mode binary/image rejection as a separate #6074 recovery slice rather than folding it into this PR. This PR adds the explicit binary path without changing existing default `utf8` reads or writes, while #6074 still records that the default `file_read` lossy-binary behavior remains missing recovery work. That keeps the base64 capability independently reviewable and revertable.

Since you offered to take that recovery slice next, please open that follow-up PR after this lands. The shape from the earlier thread still looks right to me: keep `encoding: "base64"` as the explicit binary path, keep normal UTF-8 text reads unchanged, reject known image/binary data in default `utf8` mode instead of falling back to lossy text, and keep PDF text extraction as the allowed binary-ish case.

PR #7007, opened 2026-05-29
fix(channels): deliver WhatsApp replies for LID JIDs and empty sanitization

1700899e-collab:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Resolve WhatsApp Web LID JIDs (`*@lid`) to deliverable phone JIDs (`digits@s.whatsapp.net`) on inbound DM/self-chat `reply_target` and on outbound send/typing/paused paths, using sender candidates and `get_phone_number_from_lid`.
  - Fail outbound explicitly when LID→phone resolution is impossible, instead of silently targeting an undeliverable JID.
  - Substitute a short fallback message when the channel orchestrator would otherwise send an empty reply after sanitization (avoids typing with no visible message).
- **Scope boundary:** Does not change WhatsApp Cloud API, other channels, or config schema.
- **Blast radius:** `zeroclaw-channels` orchestrator (all channels get empty-reply fallback); WhatsApp Web paths require `whatsapp-web` feature.
- **Linked issue(s):** None filed — community fix for undelivered WhatsApp Web replies and empty channel responses.
- **Labels:** `channel:whatsapp`, `risk: medium`, `size: S` (requested; adjust in sidebar if needed)

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-channels --all-targets --features whatsapp-web -- -D warnings
cargo test -p zeroclaw-channels --features whatsapp-web
```

- **Commands run and tail output:**

```
$ cargo fmt --all -- --check
(no diff — exit 0)

$ cargo clippy -p zeroclaw-channels --all-targets --features whatsapp-web -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.67s

$ cargo test -p zeroclaw-channels --features whatsapp-web
test result: ok. 1137 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.00s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
```

- **Beyond CI — what did you manually verify?** Not run: live WhatsApp Web session with a LID-based DM contact. Recommended manual checks: DM from LID contact receives reply; self-chat with LID JID; agent turn that sanitizes to empty text shows fallback.
- **If any command was intentionally skipped, why:** Full-workspace `cargo test` / `cargo clippy --all-targets` not run locally; change is confined to `zeroclaw-channels` and validated with `whatsapp-web` feature enabled.

[...]

Issue #6970, opened 2026-05-27
[Tracker]: v0.8.1 integration/channel/provider/tool queue and history


Audacity88:
## Summary

Operational tracker for the v0.8.1 integration/channel/provider/tool queue across additive channels, providers, tools, accepted issue-only follow-ups, and integration-adjacent runtime/config/core work.

This complements #6489. That issue should stay focused on the long-term Integrations -> Plugins catalog architecture. This tracker is for release routing: which PRs are reviewable, dirty, blocked, consolidated, waiting for author action, still issue-only, merged, closed, superseded, deferred, or removed from the release queue.

Use the active queue groups for current review, merge, author-action, and no-PR-yet work. Use the terminal/off-ramp history for provenance; those entries are intentionally preserved, but they are not fresh review targets.

## Problem statement

The v0.8.1 milestone has many open PRs and accepted issue-only follow-ups. Some are ordinary additive channel/provider/tool work. Some are approved but dirty. Some are related PR clusters that should consolidate before normal review. Some have already merged, closed, moved milestones, or become stale/off-ramp history. Putting all of that inside #6489 would make the architecture issue hard to read and would blur design direction with day-to-day queue state.

## Proposed use

- Keep #6489 focused on the unified plugin/catalog architecture.
- Use this issue for operational v0.8.1 routing across PRs and accepted issue-only work.
- Keep active queue groups limited to current review, merge, author-action, consolidation, or no-PR-yet work.
- Keep merged, closed, superseded, and stale-candidate v0.8.1 entries in terminal/off-ramp history rather than active lanes. Move deferred issue-only work out to its destination tracker instead of preserving it here.
- Prefer queue-state updates when a PR or tracked issue becomes clean, dirty, approved, merged, closed, superseded, moved out of v0.8.1, or removed from the release queue.
- Do not treat this issue as acceptance of every linked PR or issue. Each PR still needs its own review, body hygiene, labels, and validation; each issue-only row still needs a concrete implementation path.

## Issue-only implementation candidates and routing (16 no-PR implementation issues; 24 no-assignee query hits)

[...]

PR #7008, opened 2026-05-29
fix(channels): deliver WhatsApp replies for LID JIDs and empty sanitization

LuaniiMM:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Resolve WhatsApp Web LID JIDs (`*@lid`) to deliverable phone JIDs (`digits@s.whatsapp.net`) on inbound DM/self-chat `reply_target` and on outbound send/typing/paused paths, using sender candidates and `get_phone_number_from_lid`.
  - Fail outbound explicitly when LID->phone resolution is impossible, instead of silently targeting an undeliverable JID.
  - Substitute a short fallback message when the channel orchestrator would otherwise send an empty reply after sanitization (avoids typing with no visible message).
- **Scope boundary:** Does not change WhatsApp Cloud API, other channels, or config schema.
- **Blast radius:** `zeroclaw-channels` orchestrator (all channels get empty-reply fallback); WhatsApp Web paths require `whatsapp-web` feature.
- **Linked issue(s):** Supersedes #7007 (same change; #7007 was opened from the wrong GitHub account fork and is closed).
- **Labels:** `bug`, `channel`, `channel:whatsapp`, `risk: medium`, `size: S`

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-channels --all-targets --features whatsapp-web -- -D warnings
cargo test -p zeroclaw-channels --features whatsapp-web
```

```
$ cargo fmt --all -- --check
(no diff - exit 0)

$ cargo clippy -p zeroclaw-channels --all-targets --features whatsapp-web -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.67s

$ cargo test -p zeroclaw-channels --features whatsapp-web
test result: ok. 1137 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.00s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
```

- **Beyond CI — what did you manually verify?** Not run: live WhatsApp Web session with a LID-based DM contact. Recommended manual checks: DM from LID contact receives reply; self-chat with LID JID; agent turn that sanitizes to empty text shows fallback.
- **If any command was intentionally skipped, why:** Full-workspace `cargo test` / `cargo clippy --all-targets` not run locally; change is confined to `zeroclaw-channels` and validated with `whatsapp-web` feature enabled.

## Compatibility (required)

- **Fast rollback command/path:** Revert the merge/squash commit for this PR.
- **Observable failure symptoms:** Logs `undeliverable LID reply_target`, `Cannot deliver to LID JID`, or `channel_reply_empty`; users see fallback text instead of silence when sanitization empties the reply.


Audacity88:
@LuaniiMM — I checked the live PR state, the PR body, the linked #7007 context, and the top-level diff. Thanks for reopening this from the right account. I’m going to leave this in the conflict/routing lane for now rather than doing a full code review, because GitHub currently reports this branch as conflicting with `master`.

Could you resolve the merge conflicts and push the updated branch? Once GitHub shows the PR as mergeable again, I’ll take the actual implementation pass over the WhatsApp LID reply-target behavior and the empty-reply fallback.

Audacity88 (changes requested):
Re-reviewed head `9e0f51d3ba59767c4be262c5946ec17e36dcb0f3`, the current diff, green checks, the prior routing comment, and the linked #7007 reopen/supersede context.

### 🟢 What looks good — LID reply delivery is moving the right way

The LID→phone normalization is heading in the right direction for normal text replies and the outbound send/typing paths, and the empty-reply fallback also addresses the sanitized-empty response case.

### 🔴 Blocking — LID voice-note chats stop queuing voice replies

One regression remains in the WhatsApp Web voice path. For LID DMs, this patch can rewrite `reply_target` from the original `*@lid` chat JID to a deliverable `digits@s.whatsapp.net` JID before emitting the `ChannelMessage` (`crates/zeroclaw-channels/src/whatsapp_web.rs`, around the new LID resolution block). But the voice-note tracking block still stores and removes `chat.clone()` in `voice_chats`, while `send()` decides whether to queue the TTS voice reply with `voice_chats.contains(&message.recipient)`.

That means an inbound LID voice note gets transcribed and delivered to the orchestrator with a phone-JID `reply_target`, but `voice_chats` is keyed by the original LID chat. When the reply is sent, `message.recipient` is the phone JID, so `send()` misses the voice-chat marker and only sends text. This breaks the existing "Voice in → text + voice out" behavior exactly on the LID DM path this PR is fixing.

Please key the voice tracking by the final outbound `reply_target` after LID resolution, or store both the original chat and the resolved target if another path still needs the original JID. A focused regression test for an LID voice-note input resolving to a phone-JID reply target would cover this nicely.

Audacity88 (approved):
I re-reviewed current head 8bac987 for #7008, including the updated diff, the prior request-changes review, current checks, #7007 supersede context, and the v0.8.1 tracker context in #6970. I did not rerun local cargo; all visible GitHub checks on this head are green.

@LuaniiMM, thanks for the focused follow-up. This resolves the voice-note blocker from my last review.

### ✅ Resolved — LID voice-note replies now use the resolved send recipient

The previous blocker was that LID voice-note chats were tracked under the original `*@lid` chat while `send()` checked `voice_chats` against the resolved phone-JID recipient. Current head fixes that by inserting/removing the final `reply_target`, and the new regression test captures the important failure mode: a LID voice note resolves to a phone JID, and `voice_chats` is keyed by the same value `send()` will check.

### 🟢 What looks good — the LID reply and empty-reply fixes are mergeable

The PR now consistently routes outbound WhatsApp Web send, composing, paused, and inbound DM reply targets through the deliverable phone-JID path when LID resolution succeeds, and it fails loudly when a LID recipient cannot be resolved for outbound delivery. The empty-channel-reply fallback is also covered by focused tests and avoids the user-visible "typing, then nothing" case after sanitization removes all visible text.

I’m approving this on the current diff and green CI. Before merge, I’ll handle the mechanical PR-body cleanup if needed so the live labels and rollback wording match the final PR state.

PR #7009, opened 2026-05-29
feat(tools): add result mode to execute_pipeline

metalmon:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds an optional `result` field (`all` | `last`) to the `execute_pipeline` tool. By default `execute_pipeline` serializes *every* step's result into the tool output; when an early step produces a large blob (e.g. base64 bytes from `file_read`), that blob flows back into the model context even though only the final step's output is wanted. `result: "last"` returns only the final step's raw output, keeping intermediate blobs out of context.
- **Scope boundary:** Only `execute_pipeline` output shaping. No change to step execution, interpolation, parallel mode, the allowlist, or any other tool.
- **Blast radius:** `crates/zeroclaw-tools/src/pipeline.rs` only. Default behavior is unchanged (`result` defaults to `all`), so existing callers are unaffected.
- **Linked issue(s):** None.
- **Labels:** `enhancement`, `risk: low`, `size: S`, `tool`.

```bash
cargo fmt -p zeroclaw-tools -- --check      # clean (exit 0)
cargo test -p zeroclaw-tools --lib pipeline # 16 passed; 0 failed
cargo clippy -p zeroclaw-tools --all-targets -- -D warnings
```

`cargo test` tail:

```
running 16 tests
...
test pipeline::tests::result_all_is_default_and_includes_every_step ... ok
test pipeline::tests::result_last_returns_only_final_output ... ok
test result: ok. 16 passed; 0 failed; 0 ignored; 0 measured; 1180 filtered out
```

- **Commands run and tail output:** `fmt` clean; the two new tests (`result_last_returns_only_final_output`, `result_all_is_default_and_includes_every_step`) plus all existing pipeline tests pass.
- **Beyond CI — what I manually verified:** End-to-end on a live agent reading a `.docx` via `execute_pipeline` (`file_read` base64 → `office_read`) with `result: "last"`: only the extracted text returned to the model, base64 no longer entered the context.
- **clippy note:** `clippy --all-targets -- -D warnings` reports two failures, both pre-existing and unrelated to this PR, in `crates/zeroclaw-tools/src/glob_search.rs` (an `unreachable statement` and `unused variable` in a test, lines 418/424; appear only on Windows where the test's early `return;` is taken). This diff touches only `pipeline.rs`, which is clippy-clean.

## Compatibility (required)

- Backward compatible? **Yes** — `result` is optional (`#[serde(default)]`) and defaults to `all`, the prior behavior. Existing pipelines need no changes.
- Config / env / CLI surface changed? **No** — only the `execute_pipeline` tool's JSON arg schema gains an optional `result` field.

Low risk: `git revert <sha>`.

### DIFF
--- a/crates/zeroclaw-tools/src/pipeline.rs
+++ b/crates/zeroclaw-tools/src/pipeline.rs
@@ -42,6 +42,22 @@ pub struct PipelineRequest {
     pub steps: Vec<PipelineStep>,
     #[serde(default)]
     pub parallel: bool,
+    /// What to include in the tool output. Defaults to every step's result.
+    #[serde(default)]
+    pub result: PipelineResultMode,
+}
+
+/// Controls what `execute_pipeline` returns to the caller.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum PipelineResultMode {
+    /// Return every step's result as a JSON array (default; backward compatible).
+    #[default]
+    All,
+    /// Return only the final step's raw output. Use when earlier steps produce
+    /// large intermediate blobs (e.g. base64) that must not flow back into the
+    /// model context.
+    Last,
 }
 
 /// Result of a single pipeline step.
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
@metalmon I reviewed #7009 at head aa67850680c6907f9f5edad714a6759307fb4138.

Context for this pass: I checked the current PR state, PR body, changed file list, full diff, inline comments, formal reviews, visible GitHub checks, and the relevant `crates/zeroclaw-tools/src/pipeline.rs` implementation path. I did not rerun the live-agent `.docx` smoke locally in this pass.

I am approving. This is a focused, backward-compatible improvement to `execute_pipeline` output shaping.

### 🟢 What looks good — `result: "last"` solves the context-bloat case without changing execution semantics

The important part is that this only changes what the pipeline tool returns after the steps have already executed. Validation, allowlist enforcement, sequential interpolation, parallel execution, and the existing default output mode all stay intact. Leaving `all` as the default preserves current callers, while `last` gives callers a clean escape hatch for the common “first step reads bytes, final step extracts text” pattern.

### 🟢 What looks good — The tests cover the compatibility boundary

The new tests prove both sides of the contract: `result: "last"` returns only the final tool output, and omitting `result` still includes every step result. That is the right regression shape for this change because the main risk was accidentally changing existing pipeline output behavior.

PR #6800, opened 2026-05-19
imessage: migrate fn is_contact_allowed(&self, sender: &str) -> bool → AllowlistAspect (archetype B)

yijunyu:
## Summary

- **Base branch:** `master`
- **What changed and why:** Replaces the hand-rolled `fn is_contact_allowed(&self, sender: &str) -> bool` predicate in `imessage.rs` with the shared `aspect_std::AllowlistAspect`. This PR is part of the 24-PR migration; the chain tip (#6799) carries the cumulative view + the M1–M5 evaluation results.
- **Scope boundary:** This PR does NOT change identity resolution (iMessage sender resolution (phone vs email) is upstream), webhook/SSE parsing, downstream message routing, LLM dispatch, or response sending. The aspect owns *only* the matching half of allowlist enforcement for the `imessage` channel.
- **Blast radius:** the `imessage` channel only. Other channels in the chain are migrated by their own PRs. If `AllowlistAspect::is_allowed` regresses for this channel's archetype (flat-list + normalizer (case-insensitive comparison for email-style senders)), only `imessage` is affected at this commit; the chain-wide blast radius lives in #6799. Mitigated by per-channel parity tests against the old predicate.
- **Linked issue(s):** none
- **Labels:** `risk: medium`, `size: S`, `type: refactor`, `area: channels`

## What changed

- Replaced `allowed_contacts: Vec<String>` with `allowlist: AllowlistAspect` in the channel struct.
- Deleted the hand-rolled `fn is_contact_allowed(&self, sender: &str) -> bool` predicate.
- Production caller routes through `self.allowlist.is_allowed(sender)`.
- Unit tests updated to assert against `allowlist.is_allowed("…")` instead of field/predicate inspection.

## Archetype

Archetype **B** — flat-list + normalizer (case-insensitive comparison for email-style senders).

iMessage senders can be phone numbers (E.164) or email addresses. The aspect uses `with_normalizer(|s| s.to_ascii_lowercase())` to handle the email case where the localpart casing varies between devices.

## Supply-chain question (resolution)

The reviewer's blocker on sourcing `aspect-std` from `github.com/yijunyu/aspect-rs` is resolved by the latest commit on this branch (and all 23 other chain branches): **`aspect-std = "0.1.1"` is now on crates.io** (published 2026-05-24), so the dep is registry-sourced, immutable, and visible to `cargo audit` / `cargo deny`. The git-pinned source has been removed.

[...]

PR #6778, opened 2026-05-19
telegram: migrate fn is_user_allowed(&self, username: &str) -> bool → AllowlistAspect (archetype B)

yijunyu:
## Summary

- **Base branch:** `master`
- **What changed and why:** Replaces the hand-rolled `fn is_user_allowed(&self, username: &str) -> bool` predicate in `telegram.rs` with the shared `aspect_std::AllowlistAspect`. This PR is part of the 24-PR migration; the chain tip (#6799) carries the cumulative view + the M1–M5 evaluation results.
- **Scope boundary:** This PR does NOT change identity resolution (Telegram `@`-prefix normalization happens inside the aspect normalizer), webhook/SSE parsing, downstream message routing, LLM dispatch, or response sending. The aspect owns *only* the matching half of allowlist enforcement for the `telegram` channel.
- **Blast radius:** the `telegram` channel only. Other channels in the chain are migrated by their own PRs. If `AllowlistAspect::is_allowed` regresses for this channel's archetype (flat-list + normalizer (`@`-prefix stripping + trim) applied at construction and check), only `telegram` is affected at this commit; the chain-wide blast radius lives in #6799. Mitigated by per-channel parity tests against the old predicate.
- **Linked issue(s):** none
- **Labels:** `risk: medium`, `size: S`, `type: refactor`, `area: channels`

## What changed

- Replaced `allowed_users: Vec<String>` with `allowlist: AllowlistAspect` in the channel struct.
- Deleted the hand-rolled `fn is_user_allowed(&self, username: &str) -> bool` predicate.
- Production caller routes through `self.allowlist.is_allowed(username)`.
- Unit tests updated to assert against `allowlist.is_allowed("…")` instead of field/predicate inspection.

## Archetype

Archetype **B** — flat-list + normalizer (`@`-prefix stripping + trim) applied at construction and check.

Telegram allowlist entries are usernames that may or may not carry a leading `@`. The aspect uses `with_normalizer(|s| s.trim().trim_start_matches('@').to_string())` so both `@alice` and `alice` match consistently.

## Supply-chain question (resolution)

The reviewer's blocker on sourcing `aspect-std` from `github.com/yijunyu/aspect-rs` is resolved by the latest commit on this branch (and all 23 other chain branches): **`aspect-std = "0.1.1"` is now on crates.io** (published 2026-05-24), so the dep is registry-sourced, immutable, and visible to `cargo audit` / `cargo deny`. The git-pinned source has been removed.

[...]

PR #6638, opened 2026-05-13
refactor(channels/discord): use AllowlistAspect for sender check

yijunyu:
## Summary

- **Base branch:** `master`
- **What changed and why:** Replaces the hand-rolled `fn is_user_allowed(&self, user_id: &str) -> bool` predicate in `discord.rs` with the shared `aspect_std::AllowlistAspect`. This PR is part of the 24-PR migration; the chain tip (#6799) carries the cumulative view + the M1–M5 evaluation results.
- **Scope boundary:** This PR does NOT change identity resolution (Discord snowflake IDs are opaque), webhook/SSE parsing, downstream message routing, LLM dispatch, or response sending. The aspect owns *only* the matching half of allowlist enforcement for the `discord` channel.
- **Blast radius:** the `discord` channel only. Other channels in the chain are migrated by their own PRs. If `AllowlistAspect::is_allowed` regresses for this channel's archetype (flat-list + `*` byte-equality), only `discord` is affected at this commit; the chain-wide blast radius lives in #6799. Mitigated by per-channel parity tests against the old predicate.
- **Linked issue(s):** none
- **Labels:** `risk: medium`, `size: S`, `type: refactor`, `area: channels`

## What changed

- Replaced `allowed_users: Vec<String>` with `allowlist: AllowlistAspect` in the channel struct.
- Deleted the hand-rolled `fn is_user_allowed(&self, user_id: &str) -> bool` predicate.
- Production caller routes through `self.allowlist.is_allowed(user_id)`.
- Unit tests updated to assert against `allowlist.is_allowed("…")` instead of field/predicate inspection.

## Archetype

Archetype **A** — flat-list + `*` byte-equality.

A flat `Vec<String>` of allowed user IDs with `*` as wildcard, byte-equality otherwise. No normalization required; production `allowed_users` entries are Discord snowflake IDs consumed verbatim.

## Supply-chain question (resolution)

The reviewer's blocker on sourcing `aspect-std` from `github.com/yijunyu/aspect-rs` is resolved by the latest commit on this branch (and all 23 other chain branches): **`aspect-std = "0.1.1"` is now on crates.io** (published 2026-05-24), so the dep is registry-sourced, immutable, and visible to `cargo audit` / `cargo deny`. The git-pinned source has been removed.

- `aspect-core v0.1.1` — https://crates.io/crates/aspect-core/0.1.1
- `aspect-runtime v0.1.1` — https://crates.io/crates/aspect-runtime/0.1.1
[...]

PR #7017, opened 2026-05-30
refactor(channels/whatsapp_web): share allowlist matching via crate::allowlist

yijunyu:
## What

Add a stateless custom-matcher entry point to the in-tree allowlist helper and route `whatsapp_web`'s phone-number check through it:

```rust
// crate::allowlist
pub fn is_user_allowed_by(allowed: &[String], user: &str, match_fn: impl Fn(&str,&str)->bool) -> bool {
    if allowed.iter().any(|u| u == "*") { return true; }
    allowed.iter().any(|entry| match_fn(entry, user))
}
```

`whatsapp_web` needs a custom comparison because its matching is E.164 normalization (`+1-555-0100` ≡ `+15550100`), which `crate::allowlist::Match`'s two case modes can't express. `normalize_phone_token` is unchanged; only the wildcard/empty/iteration scaffolding is now shared.

## Why in-tree (per review)

Revised per @Audacity88's review: the reusable part is small (wildcard/empty handling + a caller-provided comparison), so it lands as a small extension to `crate::allowlist` rather than an external dependency. No `aspect-std`, no added crates.io edge, no lockfile movement. The stateless/no-cached-allowlist invariant is preserved — `is_user_allowed_by` borrows the caller's freshly-resolved peer list and holds nothing.

## Validation

```
$ cargo test -p zeroclaw-channels --features whatsapp-web whatsapp_web
test result: ok. (exact / wildcard / empty / normalized-format tests pass unchanged)
$ cargo test -p zeroclaw-channels --features whatsapp-web allowlist::tests
test result: ok. (by_empty_denies_and_wildcard_admits, by_phone_e164_normalized, ...)
```

Behavior is byte-identical; the existing tests are kept unchanged as the proof.

## Security & Privacy Impact

Authorization predicate; behavior preservation is the security property. Still fails closed (empty denies, `"*"` admits, non-wildcard compares after E.164 normalization). No new logging/caching of identities. No new dependency.

## Compatibility

No public API, config, or wire-format change. `is_number_allowed` / `is_number_allowed_for_list` keep their signatures. No MSRV change. Rollback is a single revert.

## Scope boundary

- **Touches:** `allowlist.rs` (new `is_user_allowed_by` + tests), `whatsapp_web.rs` (`is_number_allowed_for_list` body).
- **Does NOT touch:** any other channel, identity resolution, LID→phone mapping, `normalize_phone_token`, or `Cargo.toml`/`Cargo.lock`.

> Note: the `is_user_allowed_by` helper is also added by #7018. Whichever lands first, the other rebases to a one-line context drop.


Audacity88 (changes requested):
I reviewed current head dddc33f for #7017, including the PR body, diff, current checks, formal reviews, inline comments, and the active v0.8.1 tracker context for the AllowlistAspect channel migration work. There are no prior review threads or formal reviews on this PR. I did not rerun local cargo because the visible GitHub checks are already red before the broader Quality Gate matrix runs.

@yijunyu, I’m holding this for the CI failures and the dependency/landing-shape issue below.

### 🟢 What looks good — the behavior contract is the right one

The existing exact, wildcard, empty, and normalized-format WhatsApp Web tests are the right behavior contract for this slice. Once the landing shape is settled, keeping those tests unchanged is a good way to show the refactor preserves behavior.

### 🔴 Blocking — CI is red before the broader test matrix runs

The visible checks currently show the CI Required Gate, `Lint`, and `main` checks failing, with the rest of the broader Quality Gate matrix skipped. These may be mechanical fixes, but they block the rest of CI from running, so the current check state cannot be used as merge validation yet.

### 🔴 Blocking — this adds dependency surface and lockfile churn for one channel predicate

The code change replaces a small local `any(...)` over normalized phone numbers with `aspect_std::is_allowed_in(..., MatchPolicy::Custom(...))`. I understand the PR body’s point that the existing `crate::allowlist::is_user_allowed` helper does not fit E.164 normalization; that explains why the current two-mode in-tree helper is not enough. It does not yet settle whether the right project direction is a new external shared matcher for this one channel predicate.

For a single WhatsApp Web allowlist predicate, that is too much dependency surface and too much lockfile movement without more justification. Some of the lockfile changes may be resolver unification rather than intentional direct updates, but the PR should still narrow that diff where possible or explain why the extra movement is required.

Please either keep this inside the existing in-tree allowlist helper by adding a small custom-matcher entry point there, or narrow the dependency/lockfile diff and justify why `aspect-std` should be the shared project direction rather than another local helper. If `aspect-std` is the intended shared direction, please also explain why this WhatsApp slice is outside the AllowlistAspect migration cluster, or fold it into that same landing plan. The v0.8.1 tracker (#6970) is already treating that migration series as consolidation work rather than independent merge candidates.

### 🟡 Warning — missing risk and compatibility sections

Because this changes channel allowlist code and adds a new dependency, the PR body should include the template’s Security & Privacy Impact and Compatibility sections. The current body has useful validation and rollback notes, but those two risk sections are missing.

yijunyu:
@Audacity88 thanks for the careful review. I've addressed the mechanical items and want to make the case on the landing-shape question directly.

**🔴 CI — fixed.** The Lint failure was a `cargo fmt --check` violation (one over-width line in `is_number_allowed_for_list`), which fails the lint job before clippy even runs — that's the bare "exit code 1" you saw. Now `cargo fmt --all --check` and `cargo clippy --workspace --features ci-all -- -D warnings` both pass locally. I've also rebased onto current `master` (the branch was ~32 commits behind) and force-pushed.

**🔴 Dependency surface / lockfile churn — narrowed.** After the rebase the `Cargo.lock` delta is just the two new package entries (`aspect-core`, `aspect-std`); the earlier extra movement was resolver unification from the stale base and is gone. `aspect-std` is optional and gated behind the existing `whatsapp-web` feature, so default and no-default builds don't pull it in (verified by `cargo check --no-default-features`), and it pulls only `aspect-core` — no transitive network/IO crates.

**🔴 Landing shape — the in-tree-vs-external question.** This is the real decision, so let me be concrete rather than hand-wave it.

First, a correction on the tracker context: the AllowlistAspect cluster #6970 refers to has already been resolved. I closed the 23 cumulative/mis-scoped PRs (#6638, #6778–#6800) on 2026-05-30 — they duplicated the in-tree `crate::allowlist` and shouldn't merge. #7017 and #7018 are the only survivors, and they're deliberately *not* duplicates of that work: they target the two channel families `crate::allowlist` structurally can't serve (E.164 phone here, domain-class email in #7018). So this isn't "the consolidation series, still sprawling" — it's the consolidated result. (#6970 still lists the closed PRs as open; it predates the cleanup.)

On in-tree vs external: you're right that *for this one predicate*, a small custom-matcher entry point in `crate::allowlist` would also work, and if the project's answer is "keep it in-tree," I'll do exactly that — it's a small change and I'd rather land the behavior than win an argument. The reason I reached for `aspect-std` instead is that the matcher/normalizer shape recurs (E.164, email domain-class, and the case-insensitive channels already on `Match`), and a single versioned crate with a parity-tested matcher API is a cleaner home for that cross-cutting concern than growing `crate::allowlist` predicate-by-predicate. The stateless `is_allowed_in(entries, identity, MatchPolicy)` entry point was added specifically so it borrows the resolved peers and holds no state — i.e. it does *not* reintroduce the cached-allowlist hazard AGENTS.md bans.

That said, this is your call as maintainer. If you'd prefer the in-tree custom-matcher route, say so and I'll convert both PRs to add the entry point to `crate::allowlist` and drop the `aspect-std` dependency. If `aspect-std` as a shared direction is acceptable, these two are the minimal, correctly-scoped slices to start with.

**🟡 PR body — fixed.** Added the Security & Privacy Impact and Compatibility sections.

Audacity88 (changes requested):
I reviewed current head 12d4948 for #7017, including the updated PR body, the author follow-up comment, the current diff, visible CI, #7018 as the related email-channel slice, and the current #6970 v0.8.1 tracker snapshot.

@yijunyu, thanks for the focused follow-up. The mechanical blockers from my earlier review are resolved, but I am keeping this blocked on the landing shape.

### ✅ Resolved — CI and PR-body risk sections

The branch is rebased, the visible CI matrix is green, and the PR body now includes Security & Privacy Impact and Compatibility sections. That clears the earlier mechanical CI/body blockers.

### 🟢 What looks good — the behavior contract stays tight

The implementation still preserves the important WhatsApp Web behavior: empty lists deny, trimmed `*` still admits everyone, and non-wildcard entries compare after the same phone-token normalization. Keeping the existing exact, wildcard, empty, and normalized-format tests unchanged is the right proof shape for this refactor.

### 🔴 Blocking — keep the custom matcher in-tree instead of adding `aspect-std`

I agree with the core shape you described: `whatsapp_web` needs a custom comparison because the existing `crate::allowlist::Match` modes cannot express E.164 normalization. I do not think this PR should introduce `aspect-std` as the project direction for that shape.

For this slice and #7018, the reusable part is still small: wildcard/empty-list handling plus a caller-provided comparison function. That fits naturally as a small in-tree extension to `crate::allowlist`, where the existing channel allowlist policy already lives. Keeping it in-tree preserves the stateless/no-cached-allowlist invariant you are aiming for, but avoids adding another crates.io dependency and a parallel allowlist abstraction for two predicates.

The current dependency story also remains heavier than the updated explanation says. The diff adds `aspect-core` and `aspect-std`, and `aspect-std` depends on `aspect-core`, `log`, and `parking_lot`; the lockfile also still shows resolver unification for existing transitive packages such as `windows-sys`, `base64`, and `itertools`. Some of that may be harmless resolver churn, but it means the public record should not describe this as only pulling `aspect-core`.

Please convert #7017, and the paired #7018 direction, to an in-tree custom matcher entry point under `crate::allowlist`, then drop the `aspect-std` dependency and lockfile movement. That gives us the same behavior-preserving consolidation without making the external dependency choice inside a small channel refactor.

yijunyu:
@Audacity88 done — converted to the in-tree route you described, and you're right on both counts.

**Converted.** `aspect-std` is dropped. The reusable part is now a small in-tree entry point — `crate::allowlist::is_user_allowed_by(allowed, user, match_fn)` — wildcard/empty handling plus a caller-provided per-entry comparison. `whatsapp_web` passes its E.164 matcher; `normalize_phone_token` is unchanged. No crates.io edge, no `Cargo.lock` movement (the diff is now `allowlist.rs` + `whatsapp_web.rs` only). Rebased onto current master; CI is re-running.

**On the dependency-accuracy point — you were correct.** The earlier body's "only pulls `aspect-core`" was wrong: `aspect-std` also pulls `log` and `parking_lot`, and the lockfile carried resolver churn (`windows-sys`, `base64`, `itertools`). Dropping the dependency removes all of that, which is the cleaner outcome regardless.

**One note for the record.** The in-tree matcher isn't ad-hoc — it's the projection of a parity-tested spec. The wildcard/empty/custom-matcher semantics were validated against a fixed truth table (E.164 phone + domain-class email) so the consolidation is provably behavior-identical to the per-channel predicates it replaces. Happy to share that truth table if it's useful as the correctness argument for the consolidation; it's also a cheap drift check you could drop into CI if you want one.

`is_user_allowed_by` is also added by #7018 (shared helper); whichever lands first, the other is a one-line context rebase. Ready for re-review.

Audacity88 (approved):
@yijunyu, I re-reviewed current head 892f97a, including the updated PR body, your follow-up comments, the current two-file diff, the prior `CHANGES_REQUESTED` reviews, the visible green CI, and a local merge-result smoke against current `master`.

### ✅ Resolved — in-tree allowlist shape replaces the dependency approach

This resolves my prior blocker. The PR now keeps the reusable policy in `crate::allowlist` as `is_user_allowed_by(...)`, drops the `aspect-std` dependency path, and removes the lockfile / dependency-surface concern entirely. The helper owns only the shared wildcard-and-iteration scaffolding, while `whatsapp_web` keeps its E.164 normalization as the caller-provided matcher. That is the right scope for this slice.

I also like that `whatsapp_web` preserves its historical `entry.trim() == "*"` wildcard behavior before delegating to the shared helper. That keeps this refactor behavior-preserving instead of silently narrowing an existing channel edge case.

### 🟢 What looks good — behavior proof stayed tight

The test shape is exactly what I wanted to see for this refactor: the existing WhatsApp exact, wildcard, empty-list, and normalized-format tests still pass, and the new helper has focused coverage for empty/wildcard handling plus E.164 phone and domain-class email matcher examples. That proves the shared helper is generic where it should be, without turning this PR into a larger channel migration.

I don’t need the external truth table to approve this slice, but if you already have it, it would be useful as follow-up coverage for the paired #7018 / shared allowlist matcher work.

### ✅ Resolved — current-master freshness check

Because `master` has moved since the visible GitHub checks and touched the same allowlist / WhatsApp files, I also checked the merge result locally. Merging current `master` into this head applied cleanly, and the focused tests passed on that merge result:

- `cargo test -p zeroclaw-channels --features whatsapp-web whatsapp_web -- --nocapture` — 49 passed, 0 failed.
- `cargo test -p zeroclaw-channels --features whatsapp-web allowlist::tests -- --nocapture` — 7 passed, 0 failed.

With the dependency blocker resolved, CI green on the PR head, and the current-master merge-result smoke passing, I’m happy with this now. Approving.

PR #7018, opened 2026-05-30
refactor(channels/email): share allowlist matching via crate::allowlist (email + gmail_push)

yijunyu:
## What

Route the email-sender allowlist check in **both** `email_channel` and `gmail_push` through the in-tree stateless custom-matcher helper (`crate::allowlist::is_user_allowed_by`). The domain-class rules (`@host` / bare `host` / full `user@host`, case-insensitive) remain the per-entry matcher; only the wildcard/empty/iteration scaffolding is shared.

`email_channel` and `gmail_push` carry byte-identical allowlist logic and share the `channel-email` feature gate, so they migrate together.

## Why in-tree (per review)

Revised per @Audacity88's review: domain-class email matching can't be expressed by `crate::allowlist::Match`'s two case modes, but the reusable part is small (wildcard/empty handling + a caller-provided comparison), so it uses the small in-tree extension to `crate::allowlist` rather than an external `aspect-std` dependency. No added crates.io edge, no lockfile movement. Stateless / no cached allowlist — `is_user_allowed_by` borrows the caller's freshly-resolved peer list.

## Validation

```
$ cargo test -p zeroclaw-channels --features channel-email --lib
test allowlist::tests::by_email_domain_class ... ok
test email_channel::tests::is_sender_allowed_case_insensitive ... ok
test email_channel::tests::is_sender_allowed_domain_with_at_prefix ... ok
test email_channel::tests::is_sender_allowed_domain_without_at_prefix ... ok
test email_channel::tests::is_sender_allowed_empty_list_denies_all ... ok
test email_channel::tests::is_sender_allowed_specific_email ... ok
test email_channel::tests::is_sender_allowed_wildcard_allows_all ... ok
test email_channel::tests::is_sender_allowed_wildcard_with_specific ... ok
# (4 pre-existing orchestrator::acp_server flakes are unrelated -- they fail on clean master too)
```

Behavior is byte-identical; existing `is_sender_allowed_*` tests kept unchanged.

## Security & Privacy Impact

Authorization predicate; behavior preservation is the security property. Still fails closed (empty denies, `"*"` admits, domain-class rules unchanged). No new logging/caching of email addresses. No new dependency.

## Compatibility

No public API, config, or wire-format change. `is_sender_allowed` / `is_email_sender_allowed` keep their signatures in both channels. No MSRV change. Rollback is a single revert.

## Scope boundary

- **Touches:** `email_channel.rs` + `gmail_push.rs` (`is_email_sender_allowed` body).
- **Does NOT touch:** `allowlist.rs`, any other channel, IMAP/SMTP/Pub-Sub handling, peer resolution, or `Cargo.toml`/`Cargo.lock`.


Audacity88 (changes requested):
I reviewed current head dba5cf9 for #7018, including the PR body, diff, current checks, formal reviews, inline comments, #7017 sibling context, and the active v0.8.1 tracker context for the AllowlistAspect channel migration work. There are no prior review threads or formal reviews on this PR. I did not rerun local cargo because the visible GitHub Quality Gate jobs for the changed code paths are already green; one separate `main` check remains red.

@yijunyu, I’m holding this for the dependency/landing-shape issue below.

### 🟢 What looks good — the duplicated email matching contract is preserved

Moving `email_channel` and `gmail_push` together makes sense because they already carry the same domain-class allowlist semantics behind the same `channel-email` feature gate. The existing empty, wildcard, exact address, domain, case-insensitive, multiple-entry, and empty-sender tests are the right behavior contract to preserve.

### 🔴 Blocking — this should not land as an isolated `aspect-std` dependency slice

The code change replaces two small local loops with `aspect_std::is_allowed_in(..., MatchPolicy::Custom(...))`. I understand the PR body’s point that the existing `crate::allowlist::is_user_allowed` helper cannot express domain-class email matching; that explains why the current two-mode in-tree helper is not enough. It does not yet settle whether the right project direction is a new external shared matcher for these channel predicates.

For this email/Gmail slice, the PR adds `aspect-std` to `zeroclaw-channels` and moves more of `Cargo.lock` than the new optional package entries. Some of the lockfile changes may be resolver unification rather than intentional direct updates, but the PR should still narrow that diff where possible or explain why the extra movement is required.

Please either keep this inside the existing in-tree allowlist helper by adding a small custom-matcher entry point there, or narrow the dependency/lockfile diff and justify why `aspect-std` should be the shared project direction rather than another local helper. Since the PR body explicitly calls this the same approach as #7017, #7018 appears to belong in the same AllowlistAspect landing plan. Please either fold it into that plan or explain why the email/Gmail slice should land separately. The v0.8.1 tracker (#6970) does not list #7018 yet, but it is already treating the sibling migration series as consolidation work rather than independent merge candidates.

### 🟡 Warning — missing template risk, compatibility, and validation details

Because this changes email-channel allowlist code and adds a new dependency, the PR body should come back to the repository template. The current body has useful validation and rollback notes, but it is still missing the required Security & Privacy Impact and Compatibility sections, and the validation evidence is summarized rather than pasted with the literal command output the template asks for.

yijunyu:
@Audacity88 thanks — addressed the items here too, and the landing-shape reasoning is the same as I wrote on #7017, so I'll keep this short.

**🔴 Dependency / lockfile — narrowed.** Rebased onto current `master` and force-pushed; the `Cargo.lock` delta is now just `aspect-core` + `aspect-std`. The dep is optional and gated behind the existing `channel-email` feature (which already gates both `email_channel` and `gmail_push`), so default/no-default builds are unaffected.

**🔴 Landing shape / not isolated.** You're right that #7018 belongs in the same plan as #7017 — they're the two channel families `crate::allowlist` can't express (domain-class email here, E.164 phone in #7017). One note on the #6970 context: the rest of that cluster (23 PRs) was already closed on 2026-05-30 as duplicating the in-tree helper; #7017/#7018 are the consolidated survivors, not additional sprawl. Same offer as on #7017: if you'd rather this land as an in-tree `crate::allowlist` custom-matcher entry point than an `aspect-std` dependency, I'll convert both. Your call on the shared direction.

**🟡 PR body — fixed.** Added Security & Privacy Impact and Compatibility sections, and pasted the literal validation command output.

Audacity88 (changes requested):
I re-reviewed current head 0cffc5e for #7018, including the updated PR body, current checks, full diff, my prior request-changes review, #7017 sibling context, and the v0.8.1 tracker context in #6970. I did not rerun local cargo; all visible GitHub checks on this head are green.

@yijunyu, thanks for the cleanup here. The template and validation evidence are in much better shape, but I am keeping this blocked on the same landing-shape decision as #7017.

### ✅ Resolved — the PR body now covers the risk sections and validation shape

The added Security & Privacy Impact, Compatibility, literal validation output, and scope notes address the missing-template part of my prior review. The review surface is now much easier to evaluate.

### ✅ Resolved — the behavior-preservation path is still the right one

Keeping `email_channel` and `gmail_push` together makes sense because they share the same domain-class email matching semantics behind `channel-email`. The unchanged sender-allowlist tests remain the right contract for empty lists, wildcard, exact address, domain forms, case-insensitive matching, multiple entries, and empty sender handling.

### 🔴 Blocking — keep the custom matcher in-tree instead of adding `aspect-std`

I agree with the core shape you described: `email_channel` and `gmail_push` need a custom comparison because the existing `crate::allowlist::Match` modes cannot express domain-class email matching. I do not think this PR should introduce `aspect-std` as the project direction for that shape.

For this slice and #7017, the reusable part is still small: wildcard/empty-list handling plus a caller-provided comparison function. That fits naturally as a small in-tree extension to `crate::allowlist`, where the existing channel allowlist policy already lives. Keeping it in-tree preserves the stateless/no-cached-allowlist invariant you are aiming for, but avoids adding another crates.io dependency and a parallel allowlist abstraction for two predicates.

The current dependency story also remains heavier than the updated explanation says. The PR body says the lockfile delta is now just `aspect-core` plus `aspect-std`, but the current diff still changes existing dependency edges for `windows-sys`, `base64`, and `itertools`. The body also says `aspect-std` pulls only `aspect-core`, while the lockfile entry shows `aspect-core`, `log`, and `parking_lot`. Some of that may be harmless resolver churn, but it means the public record should not describe this as only pulling `aspect-core`.

Please convert #7018, together with the paired #7017 direction, to an in-tree custom matcher entry point under `crate::allowlist`, then drop the `aspect-std` dependency and lockfile movement. That gives us the same behavior-preserving consolidation without making the external dependency choice inside a small channel refactor.

yijunyu:
@Audacity88 converted to in-tree as well, same as #7017.

**Converted.** `aspect-std` dropped; both `email_channel` and `gmail_push` now route through the in-tree `crate::allowlist::is_user_allowed_by`, with the domain-class rules as the caller-provided matcher. No external dependency, no `Cargo.lock` movement — the diff is `allowlist.rs` + the two channel files. Rebased onto current master.

The domain-class matcher and the phone matcher in #7017 were both validated against a shared truth table so the consolidation is byte-identical to the per-channel predicates. `is_user_allowed_by` is the same helper added by #7017 — whichever merges first, the other drops to a trivial rebase. Ready for re-review.

Audacity88 (approved):
I re-reviewed #7018 at head f3c62b2, including the updated PR body, the current diff, visible checks, my prior request-changes reviews, and the sibling #7017 direction.

@yijunyu, this now matches the in-tree direction I asked for on #7017/#7018. The prior dependency and landing-shape blocker is resolved, and the current email/Gmail slice looks mergeable from my review side.

### ✅ Resolved — the custom matcher is now in-tree

The current diff drops `aspect-std`, removes the lockfile/dependency movement, and adds the reusable piece as `crate::allowlist::is_user_allowed_by`. That keeps wildcard handling centralized with the existing channel allowlist policy while leaving the channel-specific email domain matcher as the caller-provided comparison.

### ✅ Resolved — the PR body now matches the current implementation shape

The body no longer describes the external dependency route. It now explains the in-tree helper, the stateless/no-cached-allowlist invariant, the email/Gmail scope boundary, validation, security/privacy impact, and compatibility.

### 🟢 What looks good — email sender semantics stay intact

`email_channel` and `gmail_push` still resolve the peer list at the call site, then apply the same empty-list deny, exact `*` wildcard, full-address, `@host`, and bare-domain matching rules as before. The unchanged sender-allowlist tests remain the right contract for this refactor, and the new helper tests cover the shared wildcard/empty/custom-matcher behavior.

### 🟡 Warning — merge order with #7017 will need a small rebase

#7017 and #7018 both add `is_user_allowed_by` in `allowlist.rs`. I do not think that should block review now, but whichever PR lands second should drop the duplicate helper/test context during rebase and keep only its channel-specific call-site change.

Approving.

yijunyu:
Rebased onto current `master` now that #7017 has merged. The `crate::allowlist::is_user_allowed_by` helper landed with #7017, so this branch no longer adds it — the diff is just the two channel files (`email_channel.rs`, `gmail_push.rs`) routing through it. (GitHub showed the pre-rebase branch as mergeable, but it would have produced a duplicate definition of the helper; the rebase resolves that.) `is_sender_allowed_*` and the `by_*` allowlist tests pass; CI is re-running on the new head. Ready to merge once green. Thanks for landing #7017.

Issue #6969, opened 2026-05-27
RFC: unified output routing model (per-peer modality preference + agent send_via tool)


mov-xound-glitch:
## Background

I recently migrated from Letta to ZeroClaw and one behaviour I relied on heavily is gone: the ability to control *how* and *where* a reply is delivered, either through explicit user instruction or through persistent preference.

Concrete examples of what no longer works:

- Morning briefing cron job delivered as a **spoken voice note** on Telegram (the whole point of a briefing)
- User saying "reply to this by voice" mid-conversation
- "Send the flight details on Telegram as text, but read me the summary on Signal as a voice note"

Letta handled the first two cases (imperfectly — it relied on the LLM interpreting free-text instructions, which was fragile). ZeroClaw's input-modality mirroring is a cleaner concept but covers only one narrow case: the user sends a voice note, the agent echoes back in voice. Any proactive or cross-channel scenario falls through.

## The gap: three layers that need to work together

Looking at the codebase, output modality and channel routing are currently decided in one place — `try_queue_voice_reply()` in the Telegram channel — based on a single runtime condition: is this `chat_id` in the in-memory `voice_chats` set? That set is populated only when an incoming voice message is processed.

This conflates three distinct concerns:

**Layer 1 — Static peer preference** (config-time)  
"This peer always wants voice on Telegram regardless of how they sent their message."  
Covers cron jobs, proactive messages, and any outbound message where there is no input modality to mirror.  
→ No mechanism exists today. The `voice_chats` set can't express this; it resets on restart and has no config backing.

**Layer 2 — Session modality state** (runtime, input-driven)  
"This conversation switched to voice because the user just sent a voice note."  
→ This is what `voice_chats` does, and it works well for symmetric conversational exchanges. Keep it.

**Layer 3 — Per-turn explicit intent** (agent-callable)  
"For *this* reply, send voice on Signal and text on Telegram."  
→ Completely absent. There is no tool, annotation, or mechanism for the agent (or the user's instruction to the agent) to override modality or route a reply to a channel other than the one the message arrived on.

Without layer 3, there is also no way for a user to say "reply by voice" in a text message and have that respected — the agent has no affordance to act on that intent.

## Proposed model: unified output routing

Rather than three separate fixes, I'd suggest a unified **output routing model** where every outbound message carries routing intent resolved by priority:

```
per-turn explicit intent  (layer 3, highest)
    ↓ if absent
session modality state    (layer 2, current voice_chats)
[...]

PR #7020, opened 2026-05-30
feat(config/channels): static output_modality preference on peer groups

mov-xound-glitch:
## Summary

- **Base branch:** `master`
- **What changed and why:** Output modality and channel routing are currently decided only at runtime by the in-memory `voice_chats` set, which is populated solely when an inbound voice note arrives. There's no way to say "this peer always gets voice" — so proactive/cron messages can't be delivered as voice notes. This adds a static, config-time preference: an `output_modality` field (`mirror` | `voice` | `text`) on `[peer_groups.*]`, and `TelegramChannel::with_voice_peer_prefs()` which records matching peers in `static_voice_peers` so TTS fires unconditionally for those peers without consuming session `voice_chats` state.
- **Scope boundary:** Config field + Telegram seeding only. No per-turn override tool, no cross-channel routing, no change to the existing input-mirroring path.
- **Blast radius:** Telegram voice delivery for peers in groups that opt into `output_modality = "voice"`. Default (`mirror`) is exactly today's behavior.
- **Linked issue(s):** Related #6969 — this is Layer 1 of the proposed output-routing model.
- **Labels:** `enhancement`, `risk: medium`, `size: S`, `config`, `channel`, `tests`, `channel:telegram`

```
cargo clippy --release --all-targets   # 0 warnings (on the integration branch)
cargo build --bin zeroclaw --release   # Finished
```

- **Beyond CI — manually verified:** running on my fork, a peer group with `output_modality = "voice"` causes the agent to deliver replies as voice notes without an inbound voice note to mirror.
- **Tests:** `OutputModality` snake_case serde round-trip; `PeerGroupConfig` parsing `output_modality = "voice"` and defaulting to `mirror` when omitted; `with_voice_peer_prefs` seeds only matching voice groups on the right channel (peers on other channels and mirror-mode groups are not seeded).
- **Intentionally skipped:** full `cargo test` not run locally (Pi 5 OOMs on the suite build); tests compile under `clippy --tests` and run in CI.

- New external network calls? **No** — uses the already-configured TTS provider; only changes when a voice note is sent.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — tests use placeholder `@alice`/`@bob` handles.

## Compatibility (required)

- Config / env / CLI surface changed? **Yes** — one new *optional* key `output_modality` on `[peer_groups.<name>]`, defaulting to `mirror`.
- Upgrade steps: none. Omitting the field preserves current behavior.

- **Fast rollback command/path:** remove `output_modality` from the affected `[peer_groups.*]` entries (reverts to `mirror`), or `git revert <sha>`.
- **Feature flags or config toggles:** the field itself (unset / `mirror` = old behavior).
- **Observable failure symptoms:** unexpected voice notes sent to a peer group, or TTS errors in the trace for those peers.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -5116,14 +5116,20 @@ fn build_channel_by_id(
                 Arc::new(move || cfg_arc.read().channel_external_peers("telegram", &alias))
             };
             Ok(Arc::new(
-                TelegramChannel::new(tg.bot_token.clone(), alias, peer_resolver, tg.mention_only)
-                    .with_persistence(config_arc.clone())
-                    .with_ack_reactions(ack)
-                    .with_streaming(tg.stream_mode, tg.draft_update_interval_ms)
-                    .with_transcription(config.transcription.clone())
-                    .with_tts(&config)
-                    .with_workspace_dir(config.data_dir.clone())
-                    .with_approval_timeout_secs(tg.approval_timeout_secs),
+                TelegramChannel::new(
+                    tg.bot_token.clone(),
+                    alias.clone(),
+                    peer_resolver,
+                    tg.mention_only,
+                )
+                .with_persistence(config_arc.clone())
+                .with_ack_reactions(ack)
+                .with_streaming(tg.stream_mode, tg.draft_update_interval_ms)
+                .with_transcription(config.transcription.clone())
+                .with_tts(&config)
+                .with_voice_peer_prefs(&config, "telegram", alias)
+                .with_workspace_dir(config.data_dir.clone())
+                .with_approval_timeout_secs(tg.approval_timeout_secs),
             ))
         }
         #[cfg(not(feature = "channel-telegram"))]
@@ -5929,6 +5935,7 @@ fn collect_configured_channels(
                 .with_streaming(tg.stream_mode, tg.draft_update_interval_ms)
                 .with_transcription(config.transcription.clone())
                 .with_tts(&config)
+                .with_voice_peer_prefs(&config, "telegram", alias)
                 .with_workspace_dir(config.channel_workspace_dir(&format!("telegram.{alias}")))
                 .with_proxy_url(tg.proxy_url.clone())
                 .with_tool_command_specs(tool_specs.to_vec())
--- a/crates/zeroclaw-channels/src/telegram.rs
+++ b/crates/zeroclaw-channels/src/telegram.rs
@@ -622,6 +622,35 @@ impl TelegramChannel {
     }
 
     /// Wire the shared Config handle so `persist_allowed_identity` can
+    /// Pre-seed voice-chat mode from peer-group static preferences.
+    ///
+    /// Iterates `[peer_groups.*]` entries that reference this channel and
+    /// carry `output_modality = "voice"`, then inserts every `external_peers`
+    /// entry into the `voice_chats` set so TTS fires unconditionally for
+    /// those peers — including cron/proactive messages that have no inbound
+    /// voice note to mirror.
+    pub fn with_voice_peer_prefs(
+        self,
+        config: &zeroclaw_config::schema::Config,
+        channel_type: &str,
+        alias: impl AsRef<str>,
+    ) -> Self {
+        use zeroclaw_config::multi_agent::OutputModality;
+        let alias = alias.as_ref();
+        let dotted = format!("{channel_type}.{alias}");
+        if let Ok(mut vc) = self.voice_chats.lock() {
+            for group in config.peer_groups.values() {
+                let matches = group.channel == channel_type || group.channel == dotted;
+                if matches && group.output_modality == OutputModality::Voice {
+                    for peer in &group.external_peers {
+                        vc.insert(peer.to_string());
+                    }
+                }
+            }
+        }
+        self
+    }
+
     /// write a paired user into `peer_groups` and save. The long-running
     /// daemon sets this from the orchestrator; tests and one-shot
     /// callers leave it unset (pairing works at runtime, doesn't persist).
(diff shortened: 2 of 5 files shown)


WareWolf-MoonWall (approved):
Reviewed at a6bc3ba66ffb5c7e2d98668e06aad72026e162ee. CI 13/13 ✅. No prior reviews or inline threads.

Clean, additive feature. The design is correct: pre-seeding `voice_chats` at construction from config-driven peer groups is exactly the right way to give proactive/cron messages voice delivery without inventing a runtime toggle.

### 🟢 What looks good — correct source of truth

`with_voice_peer_prefs` reads from `&Config` directly at call time instead of storing a copy. The `voice_chats` set is the existing in-memory state; initialization from config at construction is not duplicate state — it's correct seeding. `OutputModality` and `output_modality` on `PeerGroupConfig` are both new fields defined here as the canonical source, not shadows of anything existing.

### 🟢 What looks good — channel isolation is tight and tested

The test exercises three cases in one pass: voice peers on the right channel type are seeded, voice peers on a different channel type are excluded, and mirror-mode peers on the right channel type are skipped. That covers the three decision branches the function has. The `dotted` channel ref (`telegram.default`) is also handled.

### 🟢 What looks good — default preserves existing behavior

`OutputModality::Mirror` is `#[default]`, the field is `#[serde(rename_all = "snake_case")]`, and the test for omitting the field confirms round-trip to `Mirror`. Operators who don't set `output_modality` see exactly today's input-mirroring behavior.

### 🟢 What looks good — both orchestrator call sites updated

Both `build_channel_by_id` and `collect_configured_channels` now call `with_voice_peer_prefs`. If either were missed, the feature would silently not work for one construction path; the test for the cold-build path catches that.

### 🔵 Suggestion — `Text` variant is currently unreachable

`OutputModality::Text` is defined and serializes correctly, but no code path in this PR checks for it. That's fine as deliberate Layer 1 scope — just worth a doc comment like `// Layer 2: wire text-enforcement in the send path` so the next contributor knows where to continue rather than wondering why the match arm is dead.

### 🔵 Suggestion — lock failure in `with_voice_peer_prefs` is silently skipped

`if let Ok(mut vc) = self.voice_chats.lock()` drops seeding silently if the mutex is poisoned. At channel-construction time this is extremely unlikely, but a tracing warn would make the silent skip observable without being disruptive:

```rust
match self.voice_chats.lock() {
    Ok(mut vc) => { /* seed */ }
    Err(e) => tracing::warn!(error = %e, "voice_chats lock poisoned during peer-pref seeding"),
}
```

WareWolf-MoonWall:

Audacity88 (changes requested):
Context: I reviewed current head a6bc3ba for #7020, including the PR body, linked RFC #6969, top-level comments, inline review comments, formal reviews, current checks, the full diff, and the PR-head source for `PeerGroupConfig`, `OutputModality`, Telegram channel construction, and the existing `voice_chats` send/finalize paths. I did not rerun local cargo for this pass; GitHub CI is green on this head.

@mov-xound-glitch, the direction looks right, but I think this needs one behavior fix before merge.

### 🟢 What looks good — the config shape is additive and scoped

`OutputModality::Mirror` as the default on `PeerGroupConfig` is the right compatibility shape, and the serde coverage for `mirror`, `voice`, and `text` is useful. Keeping Layer 1 limited to Telegram seeding also matches the RFC discussion: this PR does not need to solve `send_via` or cross-channel fan-out.

### 🔴 Blocking — static voice preferences are consumed after the first send

`with_voice_peer_prefs()` seeds peers with `output_modality = "voice"` into the same `voice_chats` set that currently represents one-shot voice-reply session state. Both voice-send paths remove the recipient from that set during voice emission: the immediate `finalize_draft()` path removes it before synthesis, and the debounced `send()` path removes it once it takes the pending voice reply. The prior approval verified that the seeding shape is correct; this finding is about the path not covered there. The seeded entry is still consumed by the existing one-shot cleanup.

That removal is correct for the existing input-mirroring flow, where an inbound voice note should cause one voice reply. But it breaks the new static preference contract. A peer configured with `output_modality = "voice"` should keep getting voice output for future proactive/cron replies, not just the first outbound response after channel construction.

Please split the persistent config-backed preference from the transient voice-session state, or otherwise make the removal path preserve statically configured voice peers. The regression test should prove repeated outbound sends to a voice-preference peer still queue TTS after the first voice note is emitted.

### 🔵 Suggestion — keep the Layer 2 placeholder explicit

`OutputModality::Text` is useful as part of the broader routing model, but it is intentionally not enforced by this PR. A short comment pointing future work at the Layer 2/Layer 3 routing follow-up would make that clear to the next contributor.

Requesting changes for the static-preference persistence issue.

mov-xound-glitch:
Added `static_voice_peers: Arc<Mutex<HashSet<String>>>` alongside the existing `voice_chats` set. `with_voice_peer_prefs()` now populates `static_voice_peers` instead of `voice_chats`, and a new `is_voice_chat()` helper checks both. The voice-send paths only remove from `voice_chats`, so config-static peers are never consumed. Session mirror-mode semantics are unchanged.

Audacity88 (approved):
I re-reviewed current head 8638612 for #7020, including the updated PR body, current checks, the current diff, the prior WareWolf-MoonWall approval, my prior request-changes review, top-level comments, inline comments, RFC #6969 context, and the v0.8.1 tracker context in #6970. I did not rerun local cargo; all visible GitHub checks are green.

@mov-xound-glitch, thanks for the focused follow-up. This resolves the static voice-preference persistence blocker from my last review.

### ✅ Resolved — static voice preferences are no longer consumed by session cleanup

Current head separates the persistent config-backed preference from the transient voice-session state: `with_voice_peer_prefs()` seeds `static_voice_peers`, `is_voice_chat()` checks both `voice_chats` and `static_voice_peers`, and the voice-send cleanup paths still remove only from `voice_chats`. That preserves the existing one-shot mirror behavior while keeping `output_modality = "voice"` active across repeated outbound sends.

The new regression test captures the important boundary: a configured static voice peer remains a voice recipient even after `voice_chats` removal. That is the behavior the earlier review asked to pin.

### 🟢 What looks good — this is the right Layer 1 shape

The PR keeps the first output-modality layer small and additive: `mirror` remains the default, `voice` is wired only where this slice can enforce it, and `text` is present as schema shape for the broader routing model without pretending to implement text enforcement here. Telegram construction now wires the preference in both channel build paths, and the peer-group tests cover matching channel refs, cross-channel isolation, and mirror-mode exclusion.

I’m approving this on the current diff and green CI.

PR #7019, opened 2026-05-30
fix(channels): transcode Telegram TTS to Opus and wire typed STT provider

mov-xound-glitch:
> **Maintainer's note** (Audacity88): I updated the title/summary wording and label snapshot to match current head 599205695. The original author's commits and implementation remain unchanged; authorship of the patch remains with @mov-xound-glitch.

## Summary

Telegram voice in/out fixes, scoped to that one concern plus the small shared `SendMessage::suppress_voice` flag needed so error notices are not spoken back as TTS. Earlier versions of this branch had drifted behind `master` and included unrelated XOAUTH2/IMAP email work plus a cron change; those have been split out. This PR now contains only the Telegram TTS/transcription slice and the supporting error-notice suppression.

**What changed and why:**

- *Outgoing:* Telegram voice replies were going through a hardcoded `sendVoice` upload while the surrounding format logic could still describe non-Opus backend output as if it were uploaded directly. The final path now uses `synthesize_opus(...)` for Telegram voice replies, so non-Opus backend output is transcoded to OGG/Opus before Telegram receives it, and the upload is sent as a true `sendVoice` voice note. Raw `pcm` is rejected explicitly rather than treated as a fake WAV container.
- *Incoming:* the agent's `transcription_provider` was not wired into the Telegram channel-internal voice path, so inbound voice notes failed in a typed-provider-only config. The typed `[providers.transcription.<family>.<alias>]` map is now loaded into the channel's `TranscriptionManager` via `with_typed_transcription_providers`, keeping the full dotted alias (`groq.default`) instead of stripping it to `groq`, and falling back to an empty manager when no legacy `[transcription]` block is present so typed-only configs work. The media-pipeline path was given the same typed-only construction so it no longer depends on a successful legacy manager build.
- *Error notices:* orchestrator error responses now use `SendMessage::suppress_voice()` so Telegram voice-chat mode sends the error as plain text instead of queuing a spoken voice note.

**Scope:** Telegram voice in/out plus shared error-notice voice suppression. No email, cron, or unrelated channel features. Normal text replies are unchanged.

**Addresses the prior review:**
- 🔴 *Typed transcription not reaching the Telegram channel-internal path* — fixed: `with_typed_transcription_providers` is wired at the channel-construction boundary with the full dotted alias and a typed-only fallback.
- 🔴 *`pcm` mislabeled as WAV* — fixed: `pcm` is now rejected for Telegram delivery with a clear message.

## Validation

Author-reported validation on the rebuilt head:

```text
[...]

PR #6512, opened 2026-05-07
fix(channels/email): html body rendering, subject threading, attachment path resolution

mov-xound-glitch:
## Problem

The email channel had three issues affecting usability:

1. **Wrong default subject** — hardcoded `"ZeroClaw Message"` default was unhelpful and undocumented
2. **Markdown sent as plain text** — tables, bold, and headers rendered as raw Markdown syntax in email clients
3. **Attachment file paths not resolved** — passing a file path to `email_send` produced zero-byte attachments because data was never read from disk
4. **Email threading broken** — inbound `Message-ID` was captured on `ChannelMessage` but never forwarded to `SendMessage`, so replies depended on subject heuristics instead of the `In-Reply-To` header

## Changes

- `scattered_types.rs`: Change default subject from `"ZeroClaw Message"` to `"Re: Message"`; add `html_body: bool` (default `true`) so operators can opt out of Markdown-to-HTML conversion for plain-text clients, body-parsing pipelines, or spam-sensitive environments
- `email_channel.rs`: Auto-convert Markdown body to HTML using `pulldown-cmark` before sending; propagate `fs::read` errors via `map_err()?` instead of `unwrap_or_default()` (silent zero-byte attachment bug); resolve attachment file paths at send time — if `data` is empty and `file_name` is a valid path, read from disk
- `orchestrator/mod.rs`: Extract `build_email_reply` helper that wires `in_reply_to` + `Re:` subject prefix into all three normal agent response send paths (provider-init error, `finalize_draft` fallback, main generated-response); three unit tests cover header propagation and subject deduplication
- `Cargo.toml`: Bump `pulldown-cmark` from `0.12` to `0.13` to align with `ruma`'s transitive dep and eliminate duplicate compilation
- `Cargo.lock`: Reset to upstream baseline and regenerate minimally — diff is 11 insertions (only `getopts`, a new transitive dep of pulldown-cmark 0.13)

## Validation

CI is fully green on the latest commit (`c5d87f3`): Lint, Security, Check (3 variants), Benchmarks Compile, Build (3 platforms), Test, CI Required Gate — all passed.

Tested manually on Raspberry Pi 5 (aarch64) with IMAP/SMTP — HTML emails render correctly in Thunderbird and Gmail; plain-text opt-out (`html_body = false`) confirmed.

## Security & Privacy

[...]

PR #7021, opened 2026-05-30
feat(channels/email): XOAUTH2 auth, observer mode, and read-only IMAP tools

mov-xound-glitch:
## Summary

- **Base branch:** `master`
- **What changed and why:** Two related email capabilities:
  1. **XOAUTH2 + observer mode** — the IMAP channel could only do password `LOGIN`, which doesn't work for providers that require OAuth2 (Hotmail/Outlook, Exchange). Adds `EmailOAuth2Config` + an optional `oauth2` block on `EmailConfig`, XOAUTH2 SASL auth through `AuthService`, a greeting-consume-before-auth fix (the SASL exchange otherwise deadlocks on the IMAP greeting), and a TLS wrapper that treats Exchange's missing `close_notify` as EOF instead of a fatal error. Also adds an opt-in `observer_mode` that only processes mail at/after the captured `uid_next`, fetches with `BODY.PEEK[]`, and never changes a flag.
  2. **Read-only IMAP tools** — `email_search` (by sender/subject/date/unseen) and `email_read` (full body by UID). Both open the mailbox with `EXAMINE` and fetch with `BODY.PEEK[]`, so they never reset `\Recent` or set `\Seen`.
- **Scope boundary:** Email channel + two read-only tools. No change to SMTP/sending, other channels, or the non-OAuth password path.
- **Blast radius:** IMAP auth/read path. Password `LOGIN` and existing flag-changing read behavior are unchanged unless `oauth2`/`observer_mode` are set.
- **Linked issue(s):** none.
- **Labels:** `enhancement`, `risk: high`, `size: XL`, `dependencies`, `channel`, `channel:email`, `config`, `provider`, `runtime`, `tool`

```
cargo clippy --release --all-targets   # 0 warnings (integration branch)
cargo build --bin zeroclaw --release   # Finished
```

- **Beyond CI — manually verified:** running on my fork against a real Hotmail account, the channel authenticates via XOAUTH2 and comes up in observer mode (`Email channel observer mode: will only process messages with UID >= N (no flag changes ever)`), then listens via IMAP IDLE. `email_search`/`email_read` return results without altering read state.
- **Tests:** XOAUTH2 SASL initial-response format (`user=<u>^Aauth=Bearer <t>^A^A`, exactly three `0x01` separators); `observer_mode` defaults off; existing email-channel config/round-trip tests.
- **Intentionally skipped:** full `cargo test` not run locally (Pi 5 OOMs on the suite build); tests compile under `clippy --tests` and run in CI. Live IMAP behavior (IDLE, observer threshold) is validated manually as above rather than against a mock IMAP server.

- New external network calls? **Yes** — the OAuth2 token endpoint (token refresh) and the same IMAP server the channel already talks to. The two tools read from the configured mailbox only.
- Secrets / tokens / credentials handling changed? **Yes** — adds XOAUTH2: an OAuth2 refresh token is stored in config and exchanged for short-lived bearer tokens via `AuthService`; the bearer token is sent to the configured IMAP host in the SASL response. No tokens are logged.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — tests use placeholder addresses/tokens (`alice@example.com`, `ya29.TOKEN`).

## Compatibility (required)

- Config / env / CLI surface changed? **Yes** — new optional `[channels.email.<alias>.oauth2]` block and an optional `observer_mode` flag (default false). Two new tools auto-register only when an email channel is enabled.
- Upgrade steps: none. Without `oauth2`/`observer_mode`, behavior is unchanged.

- **Fast rollback command/path:** remove the `oauth2` block and `observer_mode` from `[channels.email.<alias>]` (reverts to password LOGIN and normal read behavior); `git revert <shas>` for the code.
- **Feature flags or config toggles:** `oauth2` (absent = password LOGIN), `observer_mode` (false = normal reads).
- **Observable failure symptoms:** IMAP auth failures in the trace (`IMAP XOAUTH2 auth failed`), the channel not listening, or `email_search`/`email_read` returning errors.


WareWolf-MoonWall (changes requested):
Reviewed at 211eb0cfa0abf6ec3bfb41523be7fbe856070d64. CI 13/13 ✅. No prior reviews or inline threads.

Substantial and well-motivated PR. The XOAUTH2 path, observer mode, and read-only tools are all sound in design, and the real-world validation against Hotmail is the right kind of confidence evidence for an IMAP change. There are two blocking concerns before merge.

### 🔴 Blocking — `TlsStreamTolerant` and `is_tls_close_notify` duplicated across crates

`TlsStreamTolerant` (with its `AsyncRead`, `AsyncWrite`, and `Debug` impls) and `fn is_tls_close_notify` are defined identically in both `crates/zeroclaw-channels/src/email_channel.rs` and the new `crates/zeroclaw-tools/src/email_imap.rs`. Per AGENTS.md's ABSOLUTE RULE, the same struct may not live in two places: "If a fact already lives somewhere in this codebase, you do NOT copy it into a new field, struct, config block, schema entry, runtime cache, or anywhere else."

The canonical home should be `email_imap.rs` — it is already designed as the shared IMAP utility (both tools import from it). `email_channel.rs` should then import via the crate boundary. Since `zeroclaw-channels` does not currently depend on `zeroclaw-tools`, the cleanest fix without adding a reverse dependency is to move the shared TLS wrapper into a small module in `zeroclaw-channels/src/` (e.g., `email_tls.rs`), mark it `pub(crate)` or `pub`, and have `email_imap.rs` in zeroclaw-tools use whichever is appropriate — or alternatively promote it to `zeroclaw-infra`. Either way, one definition, imported everywhere it is used.

### 🔴 Blocking — IMAP search criteria accept only `"` sanitization; `(` `)` and IMAP keywords are injectable

`email_search` builds IMAP SEARCH criteria by string concatenation after stripping only double-quotes:

```rust
criteria_parts.push(format!("FROM \"{}\"", f.replace('"', "")));
criteria_parts.push(format!("SUBJECT \"{}\"", s.replace('"', "")));
```

An agent-supplied `from` or `subject` value containing `)`, `(`, or IMAP keywords (`OR`, `NOT`, `ALL`, `UNSEEN`) could alter the search semantics — for example escaping the `FROM "…"` group and appending unintended criteria. While the tool can only read from the configured mailbox (no cross-account exfiltration), crafted injection could return a different message set than intended (e.g., all messages instead of filtered ones).

The minimal fix is an allowlist: only allow alphanumeric chars, `@`, `.`, `-`, `_`, `+`, space, and `*` in from/subject filters, rejecting anything else with a descriptive error before building the criteria string. Alternatively, validate inputs against a tight regex and return an error rather than sanitize-and-continue, so the caller knows the filter was rejected.

---

### 🟢 What looks good — EXAMINE + BODY.PEEK throughout

Both `email_search` and `email_read` open mailboxes with EXAMINE (read-only; no `\Recent` reset) and fetch with `BODY.PEEK` (no `\Seen` set). The channel's observer mode also uses `BODY.PEEK` and never calls any flag-mutating command. The read-only invariant is consistent and correctly documented in the code comments.

### 🟢 What looks good — XOAUTH2 SASL format is pinned by test

The SASL initial response test (`xoauth2_sasl_response_matches_spec`) asserts the exact byte sequence including the `0x01` delimiters. That's the right level of precision — off-by-one in the separators is a silent auth failure. The test catches it.

### 🟢 What looks good — `TlsStreamTolerant` solves a real Exchange problem

Many Exchange-hosted mailboxes drop the TLS connection without a `close_notify`, which rustls 0.23+ treats as fatal. Converting that error to EOF at the stream wrapper level is the correct fix: the IMAP session reconnect path handles EOF gracefully, and nothing above the stream wrapper needs to know about the server's non-conformance. The greeting-consume-before-auth fix (deadlock prevention) is also correct and clearly documented.

### 🟢 What looks good — `EmailOAuth2Config` is the canonical config SoT

`EmailOAuth2Config` is defined once in `scattered_types.rs`. The `AuthService` holds token state (not config state) — it takes `token_url`, `client_id`, and `scopes` from the caller at token-refresh time rather than storing a snapshot. That's the right pattern per AGENTS.md.

### 🟢 What looks good — observer mode threshold is captured at session start

`uid_next` is read from the `SELECT`/`EXAMINE` response immediately after connection, and only messages with UID ≥ that threshold are dispatched. The log line ("Email channel observer mode: will only process messages with UID >= N") makes the threshold visible for ops debugging.

### 🟡 Warning — `is_tls_close_notify` uses error message string-matching

The current implementation:
```rust
fn is_tls_close_notify(e: &io::Error) -> bool {
    let msg = e.to_string();
    msg.contains("close_notify") || msg.contains("UnexpectedEof")
}
```

String-matching on `io::Error` display text is fragile across rustls versions — message text is not a stable API. `io::ErrorKind::UnexpectedEof` is stable and covers the Exchange case directly:
```rust
fn is_tls_close_notify(e: &io::Error) -> bool {
    e.kind() == io::ErrorKind::UnexpectedEof
        || e.to_string().contains("close_notify")
}
```
This is entangled with the DRY block above — once consolidated into one location, fix the detection there.

### 🔵 Suggestion — tool `description()` uses bare string literals

AGENTS.md's localization rule requires all user-facing output (including tool descriptions) to use `fl!()` / Fluent strings. `email_search` and `email_read` return bare `&str` literals. Existing tools in the codebase do the same — this is pre-existing technical debt, not a new regression introduced here — but the rule still applies. Opening a follow-up issue to migrate tool descriptions to Fluent would keep this PR unblocked while tracking the work.

### 🔵 Suggestion — each `email_search` / `email_read` call opens a new IMAP session

For interactive usage this adds per-call latency (TCP connect + TLS handshake + AUTHENTICATE). Not a correctness issue, but noting it for future work if session reuse becomes desirable.

WareWolf-MoonWall:

mov-xound-glitch:
Both blocks are addressed at `1a5d99055`.

**DRY — `TlsStreamTolerant` / `is_tls_close_notify`:** removed the duplicate from `email_channel.rs`. The single definition now lives in `zeroclaw_tools::email_imap` (the shared IMAP utility), and `email_channel.rs` imports it. `zeroclaw-channels` already depends on `zeroclaw-tools`, so this adds no new dependency edge or cycle. Folded in the `io::ErrorKind::UnexpectedEof` suggestion at the same time — the surviving `is_tls_close_notify` now keys off the stable `ErrorKind` and keeps the string match only as a fallback.

**IMAP search injection:** `email_search` no longer sanitize-and-continues. Agent-supplied `from`/`subject` filters are now allowlist-validated (alphanumeric plus `@ . - _ + space *`) and rejected with a descriptive error *before* the SEARCH criteria are built, so `(`, `)`, `"`, and IMAP keywords can't escape the `FROM`/`SUBJECT` group. Added tests for accepted addresses/keywords and rejected payloads.

On the two non-blocking 🔵 suggestions (Fluent tool descriptions, per-call IMAP session reuse): I'd prefer to take those as follow-ups rather than widen this PR — happy to open tracking issues. Ready for re-review.

Audacity88 (changes requested):
Reviewed current head 1a5d99055605ad41400480828759775926bb8c8d after @mov-xound-glitch's follow-up. I checked the live PR state, prior WareWolf-MoonWall review, top-level comments, formal reviews, inline comments API, current diff against `master`, and the changed channel/tool/auth/config code. CI is green on this head, but the PR is currently dirty/conflicting against `master`.

The two previous blockers look addressed: `TlsStreamTolerant` / `is_tls_close_notify` now have one shared definition in `zeroclaw_tools::email_imap`, and the `from` / `subject` IMAP search filters now reject the earlier metacharacter escape path before building SEARCH criteria.

I still see two input-boundary issues in the new read-only tools that should be fixed before merge.

### 🔴 Blocking — `since` / `before` dates are still concatenated into SEARCH without strict validation

`email_search` now validates `from` and `subject`, but `since` and `before` still flow through `to_imap_date()` into raw SEARCH fragments:

```rust
criteria_parts.push(format!("SINCE {}", imap_date));
criteria_parts.push(format!("BEFORE {}", imap_date));
```

`to_imap_date()` checks the part count and month value, but it does not prove that the year and day are numeric or that the day is in range. For example, malformed input such as a non-numeric day can still become an unquoted SEARCH date token, while other invalid date strings return `None` and silently drop the intended `SINCE` / `BEFORE` filter. That means bad model-supplied input can either send malformed criteria or broaden the search instead of failing closed. This is the same trust-boundary class as the fixed `from` / `subject` issue: the model controls the tool argument, and the tool builds an IMAP command string from it.

Please make date handling strict and fail closed. A small parser that accepts only `YYYY-MM-DD`, validates year/month/day ranges, formats the IMAP date from parsed numeric values, and returns a descriptive error for invalid input would close this cleanly. Please add regression coverage for rejected malformed dates so this boundary stays pinned.

### 🔴 Blocking — `email_read` silently truncates oversized JSON UIDs

`email_read` parses the requested UID as `u64` and then casts with `as u32`:

```rust
let uid = args
    .get("uid")
    .and_then(|v| v.as_u64())
    .ok_or_else(|| anyhow::Error::msg("uid is required"))? as u32;
```

That wraps values above `u32::MAX`. A caller asking for `4294967297` would fetch UID `1`. Since this tool returns full email contents, it should never turn an out-of-range UID into a different valid mailbox read.

Please replace the cast with a checked conversion and return a clear error when the value does not fit an IMAP UID. A focused unit test for the overflow case is enough.

### ✅ Resolved — prior TLS wrapper duplication / close-notify detection

The shared IMAP utility now owns `TlsStreamTolerant`, `ImapSession`, and `is_tls_close_notify`, and the close-notify detection uses `ErrorKind::UnexpectedEof` with the string check only as a fallback. That resolves my concern from the prior review.

### ✅ Resolved — prior `from` / `subject` SEARCH injection path

The updated `validate_search_filter()` rejects quotes and parentheses before criteria construction, and the tests cover representative accepted and rejected values. That resolves the earlier `FROM` / `SUBJECT` escape issue.

### 🟢 What looks good — read-only mailbox operations are still consistent

The new tools open mailboxes with `EXAMINE` and fetch with `BODY.PEEK`, and observer mode uses `BODY.PEEK` rather than the active-mode `RFC822` path. The read-only intent is coherent across the new code; the remaining issues are about validating which read is requested, not about hidden flag mutation.

mov-xound-glitch:
Rebased onto upstream/master (one conflict in `email_channel.rs` from the `dispatch_email` refactor in #6512, resolved by taking upstream).

Date validation is now strict on all three parts: month was already validated via exhaustive match; year now requires exactly 4 ASCII decimal digits, day requires 1–2 decimal digits in range 1–31. Both `since` and `before` callers return an error on invalid input rather than silently dropping the filter.

@Audacity88 ready for re-review when you get a chance.

Audacity88 (changes requested):
I re-reviewed current head 14ff4ed for #7021, including the updated PR body, current checks, the current diff, prior WareWolf-MoonWall and Audacity88 request-changes reviews, top-level comments, inline comments API, #6970 tracker context, and the current `email_search.rs` / `email_read.rs` files at this head. I did not rerun local cargo; all visible GitHub checks are green.

@mov-xound-glitch, thanks for the follow-up and rebase. Most of the prior boundary work is now resolved, but I am keeping this blocked on the remaining UID conversion issue in `email_read`.

### ✅ Resolved — date filters now fail closed

The `since` and `before` paths now turn invalid dates into descriptive errors before SEARCH criteria are built, and `to_imap_date()` now rejects malformed year, month, and day values. The added tests cover invalid years, invalid months, and invalid days, so the earlier silent-drop / malformed-token issue is resolved.

### ✅ Resolved — prior TLS wrapper and `from` / `subject` blockers stay fixed

The shared IMAP utility remains the single home for `TlsStreamTolerant`, `ImapSession`, and `is_tls_close_notify`, with `UnexpectedEof` handled via `io::ErrorKind` before the fallback string check. The `from` and `subject` filters also still reject metacharacters before SEARCH criteria construction. Those earlier blockers remain resolved on this head.

### 🔴 Blocking — `email_read` still truncates oversized JSON UIDs

The current `email_read` implementation still parses the requested UID as `u64` and then casts it with `as u32`:

```rust
let uid = args
    .get("uid")
    .and_then(|v| v.as_u64())
    .ok_or_else(|| anyhow::Error::msg("uid is required"))? as u32;
```

That cast wraps values above `u32::MAX`. A request for an out-of-range UID can therefore become a different valid IMAP UID and return the wrong email body. Because `email_read` returns full message contents, this should fail closed rather than reinterpret the caller's UID.

Please replace the cast with a checked conversion and return a clear error when the value does not fit an IMAP UID. A focused unit test for an oversized UID such as `4294967297` would pin the boundary.

### 🟢 What looks good — the read-only design remains coherent

The tools still use `EXAMINE` plus `BODY.PEEK`, observer mode uses `BODY.PEEK`, and the OAuth2 path keeps token refresh in `AuthService` rather than duplicating config state. Once the UID conversion is checked, the remaining read-only/tool boundary looks ready for approval.

theonlyhennygod (commented):
Reviewed current head 439d832d for #7021, including the full diff against `master`, prior reviews by @WareWolf-MoonWall (211eb0cf) and @Audacity88 (1a5d99055 + 14ff4ed), all top-level comments, the inline comments API (empty), current CI checks, and the changed files in `zeroclaw-channels`, `zeroclaw-tools`, `zeroclaw-providers`, `zeroclaw-config`, and `zeroclaw-runtime`.

@mov-xound-glitch — this is a well-structured PR. The observer mode, XOAUTH2 path, and read-only tools are coherent as a unit, and the iterative boundary hardening across review rounds shows the right discipline. I'm posting as a comment because @Audacity88 still holds an active CHANGES_REQUESTED from commit 14ff4ed, but the UID truncation issue they flagged is resolved in the current head (see below).

### ✅ Resolved — `email_read` UID truncation (from @Audacity88's block on 14ff4ed)

The current head replaces the `as u32` truncating cast with `try_into()` and returns a descriptive error when the value doesn't fit:

```rust
let uid: u32 = args
    .get("uid")
    .and_then(|v| v.as_u64())
    .ok_or_else(|| anyhow::Error::msg("uid is required"))?
    .try_into()
    .map_err(|_| {
        anyhow::Error::msg(
            "uid exceeds maximum IMAP UID (4294967295); \
             check that the uid came from email_search on this mailbox",
        )
    })?;
```

The test `uid_rejects_value_above_u32_max` covers `4294967297` (u32::MAX + 2). This resolves the last open blocker from prior reviews.

### ✅ Resolved — all prior blockers from @WareWolf-MoonWall and @Audacity88

TLS wrapper (`TlsStreamTolerant` + `is_tls_close_notify`) consolidated into `email_imap.rs` as the single definition. `from`/`subject` SEARCH injection closed via `validate_search_filter` allowlist. Date parsing (`to_imap_date`) now strictly validates year/month/day and returns errors instead of silently dropping filters. `is_tls_close_notify` uses stable `io::ErrorKind::UnexpectedEof` before the string fallback. All confirmed on this head.

### 🟡 Warning — `XOAuth2` Authenticator struct and SASL response format still duplicated

This is the same DRY class as the original TLS-wrapper block, but applied to the authenticator. Two separate `XOAuth2` structs implement `async_imap::Authenticator` with identical SASL responses:

- `crates/zeroclaw-channels/src/email_channel.rs` — private `XOAuth2` struct, calls a factored `xoauth2_sasl_response(user, token)` function (which has the test coverage)
- `crates/zeroclaw-tools/src/email_imap.rs` — private `XOAuth2` struct, inlines `format!("user={}\x01auth=Bearer {}\x01\x01", ...)` directly in `process()`

Both produce the same bytes today, but the `email_imap.rs` copy has no test and could silently diverge from the spec-pinned version in `email_channel.rs` if the SASL format ever changes. Per AGENTS.md: "If a fact already lives somewhere in this codebase, you do NOT copy it into a new field, struct, config block, schema entry, runtime cache, or anywhere else."

The fix is the same pattern used for the TLS wrapper: the `XOAuth2` struct (or at minimum the `xoauth2_sasl_response` helper) should live in `email_imap.rs` (the shared IMAP utility), and `email_channel.rs` should import it. Since `zeroclaw-channels` already depends on `zeroclaw-tools`, this adds no new dependency edge. Not blocking because both copies currently produce correct output, but the structural concern is real and should be addressed before this drifts further from the resolved TLS pattern.

### 🟢 What looks good — the UID fix is clean

The `try_into()` + `.map_err()` pattern is the right approach — it fails closed with a message that tells the caller exactly what went wrong and how to fix it ("check that the uid came from email_search on this mailbox"). The unit test pins the boundary at `u32::MAX + 2`. This is the kind of input-boundary discipline that the review rounds have built up, and it shows.

### 🟢 What looks good — observer mode separation is architecturally sound

The `fetch_unseen_active` (RFC822, sets `\Seen`) vs `fetch_new` (BODY.PEEK, no flag changes) split cleanly along the `observer_mode` flag. The active path runs exactly once at startup, and all subsequent processing (both IDLE and poll) uses the UID-threshold + BODY.PEEK path. There's no shared mutable state between the two modes — the `uid_threshold` is captured once and threaded through. This is the right design for a mode that promises to never touch server state.

### 🟢 What looks good — AuthService token refresh resolves from config, not cache

`get_valid_email_oauth2_token` takes `token_url`, `client_id`, and `scopes` as call-time parameters rather than storing them. `EmailOAuth2Config` is defined once in `scattered_types.rs`. The `AuthService` holds only token state (the mutable part), not config snapshots. This is the correct pattern per AGENTS.md — config is the source of truth, runtime state is derived on demand.

### 🔵 Suggestion — tool description is defined in two places

The `Tool::description()` method in `email_search.rs` and `email_read.rs` returns a hardcoded string. The orchestrator (`orchestrator/mod.rs`) also injects separate description strings into the system prompt when an email channel is enabled. These two descriptions for the same tool can drift apart over time. Ideally, the orchestrator would read from the `Tool` trait's `description()` rather than maintaining its own copy. Not a regression — other tools in the codebase have the same pattern — but worth noting for a future cleanup pass.

### 🔵 Suggestion — `AuthService` is constructed separately for channels and tools

`collect_configured_channels` in the orchestrator constructs an `AuthService` for the email channel's XOAUTH2 path, and `all_tools_with_runtime` in `tools/mod.rs` constructs a separate `AuthService` for the read-only email tools. Both are backed by the same on-disk token store, so token refresh from one is visible to the other. This is correct. However, it means two independent in-memory caches for the same data. Not a correctness issue, but if session reuse or cache coherence ever matters, a shared `Arc<AuthService>` threaded through both code paths would eliminate the duplication.

---

**Verdict:** Comment. I have nothing new to block on. All prior blockers are resolved in the current head. The remaining 🟡 on the duplicated authenticator struct is real but not merge-blocking — it's the same DRY class as a previously-resolved block, and I'd like to see it addressed before it normalizes as two untested copies of the same SASL format.

@Audacity88 — the UID truncation fix is in at `439d832d`. When you get a chance, could you re-review to confirm and dismiss your block?

Audacity88 (commented):
Reviewed current head 439d832d for #7021, including the updated diff, the author follow-up, prior reviews by @WareWolf-MoonWall / @theonlyhennygod / @Audacity88, the inline comments API, current PR metadata, current checks, and the changed email channel / IMAP tool code. I did not rerun local cargo for this re-review; the captured visible GitHub check snapshot is green on this head.

@mov-xound-glitch, thanks for the focused follow-up. I’m commenting rather than approving because @WareWolf-MoonWall’s earlier changes-requested review is still active, but the two blockers from my prior review are resolved on this head.

### ✅ Resolved — `email_search` date filters now fail closed

The `since` and `before` paths now reject malformed dates before building IMAP SEARCH criteria. `to_imap_date()` requires a four-digit numeric year, a valid month token, and a 1–31 numeric day, and the caller now returns a descriptive error instead of silently dropping the filter. The added tests cover invalid year, month, and day cases, so the earlier malformed-token / broadened-search concern is resolved.

### ✅ Resolved — `email_read` rejects oversized JSON UIDs

`email_read` now converts the JSON `u64` UID with `try_into()` and returns a clear error when the value does not fit an IMAP UID, rather than wrapping with `as u32`. The oversized UID test pins the boundary at `4294967297`, so the prior “wrong mailbox read by truncation” issue is resolved.

### 🟡 Warning — duplicated XOAUTH2 authenticator helper should not drift

I agree with @theonlyhennygod’s warning that the channel path and shared IMAP tool path still define separate `XOAuth2` authenticators / SASL response construction. Both produce the correct response today, so I’m not holding merge on it, but please avoid letting the two copies diverge. A small follow-up that moves the SASL response helper into the shared IMAP utility and imports it from the channel path would keep this aligned with the TLS-wrapper cleanup already done in this PR.

### 🟢 What looks good — read-only boundaries are coherent now

The read-only tools use `EXAMINE` plus `BODY.PEEK[]`, observer mode uses the no-flag-mutation path, and the remaining model-controlled read inputs now fail closed at the important boundaries. That gives me enough confidence in the current shape of the feature.

@WareWolf-MoonWall, could you re-review when you get a chance? The current head appears to address the two blockers from your earlier review: the TLS close-notify wrapper now has a single shared home in the IMAP utility, and the `from` / `subject` search filters now reject the metacharacter escape path before building SEARCH criteria. I’m leaving your review active rather than dismissing it from our side, since I’d rather have you confirm whether the current head satisfies your concerns.

One merge-readiness note remains separate from this review: GitHub still reports the branch as dirty/conflicting, so it needs a refresh before merge.

WareWolf-MoonWall (approved):
Reviewed at 439d832d6eb6190b44c59f236664a7d1e1058056. CI is green.

Substantial improvements made to the trust boundaries of the email tools. The critical injection and duplication issues have been resolved, significantly hardening the toolset for agent use.

### ✅ Resolved — `TlsStreamTolerant` and `is_tls_close_notify` duplication
The shared implementation in `zeroclaw_tools::email_imap` satisfies the **AGENTS.md** single source of truth requirement. The update to use `io::ErrorKind::UnexpectedEof` also addresses the fragile string-matching warning.

### ✅ Resolved — IMAP search injection in `from` and `subject`
The move from simple sanitization to an explicit allowlist for agent-supplied filter values closes the escape vector for IMAP command injection.

### ✅ Resolved — Strict date validation in `email_search`
The updated `to_imap_date` implementation now correctly validates year/month/day ranges and ensures input is numeric, preventing malformed criteria from being injected into SEARCH fragments.

### ✅ Resolved — Checked UID conversion in `email_read`
Replacing the unsafe `as u32` cast with a checked `try_into()` prevents the tool from silently returning incorrect message contents when provided with an out-of-range UID.

### 🟢 What looks good — Robust input validation and testing
The addition of regression tests for rejected malformed dates and oversized UIDs demonstrates high engineering discipline and ensures these trust boundaries remain intact.

### 🔵 Suggestion — Tool descriptions use bare string literals
As noted previously, migrating tool `description()` outputs to Fluent (`fl!()`) would align with the localization requirements in **AGENTS.md**. This is a minor technical debt item that can be handled in a separate PR or tracking issue.

Audacity88:
@mov-xound-glitch, I rechecked current head 439d832d after @WareWolf-MoonWall’s approval. The functional blockers are resolved from my side too: strict date validation, checked UID conversion, shared TLS close-notify handling, and `from` / `subject` filter validation all look good.

I’m not clearing my active `CHANGES_REQUESTED` yet because GitHub now reports the branch as `DIRTY` / `CONFLICTING` against `master`, and the current check snapshot only shows the path-label job. Could you refresh the branch against current `master` and let CI rerun? Once the branch is clean and checks are back, this looks approval-ready from my side.

Audacity88 (approved):
Reviewed current head e25b5e123 after @mov-xound-glitch's refresh. I checked the live PR state, the prior @WareWolf-MoonWall / @theonlyhennygod / @Audacity88 reviews, the top-level comments, inline comments API, current visible checks, the current diff, and the changed email channel, IMAP tool, OAuth2 auth, config, and runtime registration code. I did not rerun local cargo; I am relying on the green visible CI plus source review for this re-review.

I'm approving because the blockers from my earlier reviews are resolved on this head, and the previous dirty/conflicting branch reason is no longer present in the live PR state.

### ✅ Resolved — `email_search` date filters fail closed

The `since` and `before` paths now reject malformed dates before SEARCH criteria are built. `to_imap_date()` requires a four-digit numeric year, a valid month token, and a 1-31 numeric day, and the caller returns a descriptive error instead of silently dropping the filter. The invalid year, month, and day tests pin that boundary.

### ✅ Resolved — `email_read` rejects oversized JSON UIDs

`email_read` now converts the JSON `u64` UID with `try_into()` and returns a clear error when the value does not fit the IMAP UID range, instead of wrapping with `as u32`. The oversized UID regression covers the prior bad case, so the tool no longer risks turning an out-of-range request into a read of a different message.

### ✅ Resolved — branch refresh and prior review blockers

The current head keeps the earlier fixes for the shared TLS close-notify wrapper and `from` / `subject` SEARCH filter validation, and @WareWolf-MoonWall has already approved those fixes. The captured PR state no longer reports the prior dirty/conflicting branch condition, and all visible checks are green. The PR still shows blocked by the active review state, which this approval is intended to clear.

### 🟡 Warning — keep the XOAUTH2 helper duplication from drifting

The non-blocking warning from @theonlyhennygod still stands: the channel path and shared IMAP utility path each define an `XOAuth2` authenticator / SASL response construction path. Both produce the correct response today, so I am not blocking on it, but this should not become permanent drift. A small follow-up that moves the SASL response helper into the shared IMAP utility and imports it from the channel path would match the cleanup pattern already used for `TlsStreamTolerant`.

### 🟢 What looks good — the read-only boundary is coherent now

The new tools use `EXAMINE` plus `BODY.PEEK[]`, observer mode uses the no-flag-mutation path, and model-controlled read inputs now fail closed at the important boundaries. The OAuth2 token refresh path resolves endpoint/client/scope data from config at call time rather than storing config snapshots in `AuthService`, which keeps the config source of truth intact.

mov-xound-glitch:
This is approved and all checks are green. Could a maintainer merge when you get a chance? #7019 is stacked on top and is blocked until this lands.

PR #6848, opened 2026-05-22
feat(integration): introduce zerocode TUI, RPC socket transport, DenyWithEdit approval, and beta-2 integration

singlerider:
This branch, when landed, is intended to be the basis for a v0.8.0-beta-2 tagged pre-release.

Known issues:
- Model-provider `fallback` behaviors need rewiring (all legacy behaviors to other model providers intentionally removed for beta timeframe; fallback to same model and model provider fires three times before counting as complete failure).
- `zerocode` context-token counter is not yet reliable.
- Daemon resident memory does not fully return to baseline: each open code/chat session holds its agent and history in RAM, concurrently held sessions are additive (max ever seen ~200MB after hours with four sessions), and glibc arena fragmentation prevents an exact return to baseline until daemon restart. Manageable for beta.
- Shell tool invocation that can "poison" TTY sessions - example, garbage character output that `stty sane` could rectify, cause unresponsive tool call operations for that tool call only. Cancelling and doing another tool call will not have poisoned behavior. Won't fix for beta period; any shell would fail from such commands.
- Daemon close and restart causes zerocode to hang indefinitely.

## Summary

- **Base branch:** `master`
- **Scale:** 367 commits, 303 files, +57,595 / −21,704. This is a maintainer-assembled beta-2 integration branch.
- **What changed and why:**
  - **New `apps/zerocode/` TUI binary** (renamed from `zeroclaw-tui`). Six top-level modes: **Dashboard**, **Config**, **Code** (ACP), **Chat**, **Logs**, **Quickstart**. Adds remote WSS transport (`--connect`, with optional `--tls-skip-verify`), file attachments + clipboard paste, mouse selection and scrollbar drag, syntax-highlighted diffs in tool-call approvals, session UIDs with HMAC signing, and a connection-status widget with reconnect logic. The binary auto-spawns a `--ephemeral` daemon when no local socket is reachable.
  - **RPC dispatch layer and Unix-socket transport** (`crates/zeroclaw-runtime/src/rpc/`, ~9.4k lines). Structured JSON-RPC 2.0 dispatch with session store, `ApprovalPendingMap`, `RpcApprovalChannel`, file/attach RPC (`data_b64` + `path` modes, session-scoped SHA-256 dedup), and `config/` namespace. Pairing-token auth is removed from the RPC path; filesystem-permission-gated socket replaces it. Windows feature parity via named pipes.
[...]

PR #7027, opened 2026-05-31
fix(channels): honor webhook Retry-After dates

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Webhook outbound retry handling now parses HTTP-date `Retry-After` header values instead of only numeric seconds.
  - Past date-form headers clamp to a zero-millisecond retry delay, matching the already-accepted zero-second numeric behavior.
  - Regression coverage now includes numeric, decimal, IMF-fixdate, RFC850, and asctime `Retry-After` values, plus a wiremock send-path test for a date-form `Retry-After` on a `429`.
- **Scope boundary:** This PR only changes webhook `Retry-After` parsing and tests. It does not change retry budgets, jitter behavior, config keys, other channels, or public API shape.
- **Blast radius:** Limited to the webhook channel's outbound retry behavior when a server returns `429` or `503` with `Retry-After`. Numeric `Retry-After` values continue to behave as before.
- **Linked issue(s):** N/A
- **Labels:** `bug`, `risk: medium`, `size: S`, `channel`, `channel:webhook`

```text
cargo fmt --all -- --check
Result: OK.

cargo test -p zeroclaw-channels parse_retry_after -- --nocapture
Result: OK. 12 passed; 0 failed; 0 ignored; 0 measured; 1165 filtered out.

cargo test -p zeroclaw-channels send_honors_retry_after_http_date_header -- --nocapture
Result: OK on final local rerun. 1 passed; 0 failed; 0 ignored; 0 measured; 1176 filtered out.

Result: OK. Finished `dev` profile.

Result: OK. 8195 tests run: 8195 passed; 10 skipped.

git diff --check
Result: OK. No whitespace errors.
```

- **Beyond CI — what did you manually verify?** I confirmed the original focused parser test failed before the fix, then verified the parser accepts integer seconds, decimal seconds, the three HTTP-date forms, and past-date clamping. I also ran the webhook send-path regression against wiremock so the parsed date-form header reaches the retry branch before the successful retry.
- **If any command was intentionally skipped, why:** Full workspace plain `cargo test` was not run separately because ZeroClaw's CI-shaped local test gate is `cargo nextest run --locked --workspace --exclude zeroclaw-desktop`, which passed locally for this PR.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No user action required. Existing numeric `Retry-After` behavior is preserved; date-form headers now follow the HTTP contract instead of falling back to local exponential backoff.

- **Fast rollback command/path:** Revert this PR to restore numeric-only webhook `Retry-After` parsing.
- **Observable failure symptoms:** Webhook outbound retries may wait too long, retry too quickly, or fail delayed delivery after a `429` / `503` response with `Retry-After`. Look for webhook retry log lines such as `server requested retry after` and repeated `Webhook send failed` messages.

- Superseded PRs + authors (`#<pr> by @<author>`, one per line): `N/A`
- Scope materially carried forward: `N/A`
- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR or contributor code is incorporated.

### DIFF
--- a/crates/zeroclaw-channels/src/webhook.rs
+++ b/crates/zeroclaw-channels/src/webhook.rs
@@ -1,5 +1,6 @@
 use anyhow::{Result, bail};
 use async_trait::async_trait;
+use chrono::{DateTime, NaiveDateTime, Utc};
 use serde::{Deserialize, Serialize};
 use std::time::Duration;
 use zeroclaw_api::channel::{Channel, ChannelMessage, SendMessage};
@@ -214,9 +215,13 @@ fn apply_jitter(delay_ms: u64) -> u64 {
     jittered
 }
 
-/// Parse a `Retry-After` header value. Supports integer seconds and
-/// decimal seconds (truncated to whole seconds). HTTP-date form is not supported.
+/// Parse a `Retry-After` header value. Supports integer seconds, decimal
+/// seconds (truncated to whole seconds), and HTTP-date values.
 fn parse_retry_after_ms(value: &str) -> Option<u64> {
+    parse_retry_after_ms_at(value, Utc::now())
+}
+
+fn parse_retry_after_ms_at(value: &str, now: DateTime<Utc>) -> Option<u64> {
     let trimmed = value.trim();
     if trimmed.is_empty() {
         return None;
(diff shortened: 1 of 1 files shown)


singlerider (approved):
**APPROVE** at head a44c8392. First review. Read the body, the full +115/−3 single-file diff against `master`, and traced the parser locally.

**Comprehension.** Webhook outbound retry only understood numeric `Retry-After` (integer/decimal seconds); the HTTP-date form that RFC 7231 also allows fell through and was ignored. This adds HTTP-date parsing for all three permitted forms and clamps past dates to a zero-ms delay, matching the existing zero-second numeric behavior.

### 🟢 What looks good — full RFC 7231 Retry-After coverage with correct precedence

The parser tries integer seconds first (fast path), then decimal-truncated seconds, then the three HTTP-date forms — IMF-fixdate (`%a, %d %b %Y %H:%M:%S GMT`), RFC850 (`%A, %d-%b-%y %H:%M:%S GMT`), and asctime (`%a %b %e %H:%M:%S %Y`). That's the complete set a server can legally send, in a sensible order, and `parse_retry_after_http_date` returns `None` if none match so a malformed header degrades to the default retry path rather than panicking.

### 🟢 What looks good — past-date clamp is correct, and the guard order matters

The clamp reads `if delay_ms <= 0 { 0 } else { u64::try_from(delay_ms).unwrap_or(u64::MAX) }`. The `<= 0` check runs *before* the `try_from`, so a past date yields `0` rather than failing the unsigned conversion and falling through to `u64::MAX` — which is the trap this ordering avoids. `parse_retry_after_past_http_date_as_zero` pins it. The `unwrap_or(u64::MAX)` then only guards an absurdly-far-future date overflowing `u64` ms, which is the right defensive default.

### 🟢 What looks good — testable time injection and thorough coverage

Splitting out `parse_retry_after_ms_at(value, now)` with an injected `now` (the public `parse_retry_after_ms` passes `Utc::now()`) makes the date math deterministic — the tests pin "now" to a 1994 timestamp and assert exact deltas, which is the only way to test relative-date parsing reliably. Coverage spans numeric, decimal, all three date forms, and past-date clamping (12 parser tests), plus a wiremock send-path test confirming a date-form `Retry-After` on a `429` actually reaches the retry branch. Scope is held to parsing — retry budgets, jitter, and config keys are untouched.

---

**Note (non-blocking) — pending zerocode beta-2 branch (#6848).** No collision: `webhook.rs` is not in #6848's changeset. Self-contained, lands cleanly regardless of ordering.

Correct, complete `Retry-After` parsing with the right clamp ordering and deterministic test coverage. Approving.

PR #6398, opened 2026-05-05
feat!: multi-agent runtime and schema V3

singlerider:
# SEEKING APPROVAL - THIS WILL BE MERGED AND USED AS BASIS FOR **BETA** RELEASE

Feedback welcome. Please include screenshots and specific references and suggestions rather than just complaints. I will fix immediately after a review. Please use a large-context model for the review process.

## Summary

- **Base branch:** `master` (via `integration/v0.8.0`; this PR is the integration branch landing)
- **What changed and why:**
  - **Schema v3 migration** (#5947): single coordinated `migrate_v2_to_v3()` ships all breaking config field changes in one pass — channel field plurality, auth optionality, DiscordHistory fold into Discord, multi-agent aliasing (channels, providers, agents, risk profiles, runtime profiles, bundle namespaces), cost pricing moved onto `ModelProviderConfig`, `CURRENT_SCHEMA_VERSION = 3`. Batching avoids forcing users through multiple migration passes.
  - **Multi-agent runtime** (#6272, RFC #5890): each `[agents.<alias>]` block is now a fully isolated unit with its own workspace at `<install>/agents/<alias>/workspace/`, per-alias identity files, declared peers, and inheritable SubAgents. Replaces the single-workspace identity model.
  - **Memory scoping**: `agents` table + `agent_id` FK on `memories`; `AgentScopedMemory` wrapper pushes agent_id filter to the query layer on every backend (Postgres `WHERE agent_id = ANY($n)`, SQLite `WHERE agent_id IN (...)`, Qdrant payload `must` filter, Markdown path-walk). Bare `Memory::store` calls attribute to the synthesized `default` agent so the NOT NULL FK is never violated by un-scoped callers.
  - **Three-tier SecurityPolicy** (#6272): `allowed_roots` (rw), `allowed_roots_read_only`, `allowed_roots_write_only`. `AccessMode::Write` grants exactly write access — read-side tools (`file_read`, `pdf_read`, `glob_search`, `content_search`) cannot see write-only roots. `SecurityPolicy::ensure_no_escalation_beyond` validates SubAgent spawns as strict parent subsets across all axes.
  - **Configurable macro + onboarding** (#6270): two-level `HashMap<String, HashMap<String, T>>` support; `prop_fields()` enumerates `providers.models.<type>.<alias>.*`; `#[configurable(family = "...")]` replaces hardcoded `provider_family_excludes()`.
[...]

Issue #6053, opened 2026-04-23
[Feature]: zeroclaw config set/init support for dynamic map entries (e.g. providers.models.<name>)


singlerider:
## Summary

`zeroclaw config set` and `zeroclaw config init` cannot create or modify entries in dynamic map fields like `providers.models.<name>`.

## Problem statement

`providers.models` is a `HashMap<String, ModelProviderConfig>` where keys are user-defined provider profile names. The `config set` and `config init` commands only resolve statically-known dotted paths. Attempting to set `providers.models.ollama.name` produces `Error: Unknown property 'providers.models.ollama.name'`, forcing users to edit `config.toml` by hand. This breaks the contract that `zeroclaw config` is the single interface for all configuration.

**Current behavior:**
```
$ zeroclaw config init providers.models.ollama
All sections already configured.

$ zeroclaw config set providers.models.ollama.name ollama
Error: Unknown property 'providers.models.ollama.name'

$ zeroclaw config list --filter providers.models.ollama
(no output)
```

## Proposed solution

Teach the config path resolver to detect when a path segment traverses a dynamic map (`HashMap`/`BTreeMap`/`IndexMap`) and treat the next segment as a map key rather than a struct field. `config set providers.models.ollama.name ollama` should insert or update the `ollama` key with `name = "ollama"`. `config init providers.models.ollama` should initialize the entry with typed defaults. `config list --filter providers.models.ollama` should enumerate its fields.

**Expected behavior:**
```
$ zeroclaw config init providers.models.ollama
✓ Initialized providers.models.ollama with defaults

$ zeroclaw config set providers.models.ollama.name ollama
✓ providers.models.ollama.name = ollama

$ zeroclaw config set providers.models.ollama.base_url http://localhost:11434
✓ providers.models.ollama.base_url = http://localhost:11434

$ zeroclaw config set providers.models.ollama.model llama3.2
✓ providers.models.ollama.model = llama3.2

$ zeroclaw config list --filter providers.models.ollama
providers.models.ollama.name           = ollama
providers.models.ollama.base_url       = http://localhost:11434
providers.models.ollama.model          = llama3.2
providers.models.ollama.api_key        = (not set)
providers.models.ollama.temperature    = (not set)
```

## Non-goals / out of scope

No changes to how map entries are validated at runtime. No interactive key-selection UI in v1.

## Alternatives considered

Manual `config.toml` editing — works but breaks the config CLI contract and is undiscoverable.

[...]

PR #7028, opened 2026-05-31
fix(config): materialize typed provider aliases from config set

Audacity88:
## Summary

- **Base branch:** `integration/zeroclaw-tui` (stacked on #6848; review target is this small delta against the integration branch, not #6848's merge-readiness against `master`)
- **What changed and why:**
  - `zeroclaw config set providers.models.<family>.<alias>.<field>` now creates the missing typed provider alias before setting a nested field, so a fresh config can accept commands such as `providers.models.deepseek.default.model` without requiring users to hand-edit TOML first.
  - Final field spelling keeps the older CLI-friendly `api-key` form working by resolving it to the schema-canonical `api_key` field after the provider alias exists.
  - Auto-materialization is scoped to typed model provider aliases only, with a regression guard so unrelated map/resource-key paths are not split into bogus provider aliases.
- **Scope boundary:** This does not change provider credential resolution, provider runtime behavior, the config schema shape from #6848, or arbitrary dynamic map sections outside `providers.models.<family>.<alias>`.
- **Blast radius:** CLI `config set`, config field-path resolution, and typed model-provider config persistence.
- **Linked issue(s):** Depends on #6848. Follows on from #6053 / #6398 for typed-provider `config set` UX.
- **Labels:** `bug`, `risk: medium`, `size: S`, `config`, `core`

```text
cargo test -p zeroclawlabs config_set_materializes_missing_typed_provider_alias -- --nocapture
result: passed; covered creating providers.models.deepseek.default before setting model and api_key

cargo test -p zeroclawlabs config_set_does_not_materialize_non_provider_map_keys -- --nocapture
result: passed; covered that cost/resource map paths are ignored by the provider-alias materializer

cargo test -p zeroclaw-config resolve_field_path_canonicalizes_kebab_final_field_segments -- --nocapture
result: passed; covered api-key resolving to the schema-canonical api_key final field

cargo check -p zeroclawlabs --no-default-features
result: passed; confirmed the config-set helper is available without agent-runtime

cargo fmt --all -- --check
result: passed

git diff --check -- src/main.rs crates/zeroclaw-config/src/helpers.rs
result: passed with no whitespace errors

cargo clippy --workspace --all-targets --features ci-all -- -D warnings
result: failed; inherited from #6848 test code outside this PR's diff: disallowed `tokio::spawn` in `crates/zeroclaw-providers/src/factory.rs:1284` and `crates/zeroclaw-tools/src/http_request.rs:1205`. This PR only changes `src/main.rs` and `crates/zeroclaw-config/src/helpers.rs`.

result: passed; 8,481 tests passed, 10 skipped
```

- **Beyond CI — what did you manually verify?** I ran a scratch-config CLI smoke that set `providers.models.deepseek.default.model` and `providers.models.deepseek.default.api-key` on a fresh config. It created `[providers.models.deepseek.default]`, persisted `model = "deepseek-chat"`, and stored `api_key` through the existing encrypted-secret path.
- **If any command was intentionally skipped, why:** No planned local command remains skipped. Full workspace clippy was run and failed on inherited parent-branch lint outside this PR's diff; CI-equivalent nextest passed.

- If any `Yes`, describe the risk and mitigation: This changes how `config set` reaches an existing secret field when the typed provider alias is missing. It does not introduce a new credential store or plaintext path; the scratch-config smoke verified the API key still flows through the existing encrypted config writer.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No upgrade steps are required; existing configs continue to load unchanged. The CLI now accepts a previously failing setup path for typed provider aliases, including older `api-key` input spelling for the final field.

- **Observable failure symptoms:** `zeroclaw config set providers.models.<family>.<alias>.<field>` would either fail with `Unknown property ...` again, create an unexpected provider alias, or write a value under the wrong field spelling.

- If `No`, why (inspiration-only, no direct code/design carry-over): Not applicable; this is a direct follow-up to the current #6848 integration branch.

### DIFF
--- a/crates/zeroclaw-config/src/helpers.rs
+++ b/crates/zeroclaw-config/src/helpers.rs
@@ -430,10 +430,12 @@ pub fn validate_alias_key(key: &str) -> Result<(), String> {
 ///
 /// Field segments derived from the schema are kebab-case; aliases are
 /// snake-only per [`validate_alias_key`]. For each known canonical
-/// path, segments are compared pairwise: equal verbatim, OR equal
-/// after swapping `_`/`-` only when the canonical segment contains a
-/// `-` (i.e. is a schema-derived field name, not an alias). Returns
-/// `raw` unchanged when no canonical path matches.
+/// path, segments are compared pairwise: equal verbatim, equal after
+/// swapping `-` → `_` when the canonical segment contains `-`, or
+/// equal after swapping `_` → `-` for the final field segment. The
+/// final-segment rule lets older CLI spelling like `api-key` resolve
+/// to schema-canonical `api_key` without rewriting map aliases such as
+/// `my_bot`. Returns `raw` unchanged when no canonical path matches.
 #[must_use]
 pub fn resolve_field_path(known_paths: &[String], raw: &str) -> String {
     let raw_segs: Vec<&str> = raw.split('.').collect();
@@ -442,10 +444,16 @@ pub fn resolve_field_path(known_paths: &[String], raw: &str) -> String {
         if known_segs.len() != raw_segs.len() {
             continue;
         }
+        let final_index = known_segs.len().saturating_sub(1);
         let all_match = known_segs
             .iter()
             .zip(raw_segs.iter())
-            .all(|(k, r)| k == r || (k.contains('-') && k.replace('-', "_") == **r));
+            .enumerate()
+            .all(|(idx, (k, r))| {
+                k == r
+                    || (k.contains('-') && k.replace('-', "_") == **r)
+                    || (idx == final_index && k.contains('_') && k.replace('_', "-") == **r)
+            });
         if all_match {
             return known.clone();
         }
@@ -669,6 +677,15 @@ mod tests {
         assert_eq!(
             resolve_field_path(&known, "providers.models.anthropic.my_bot.api-key"),
             "providers.models.anthropic.my_bot.api-key",
+        );
+    }
+
+    #[test]
+    fn resolve_field_path_canonicalizes_kebab_final_field_segments() {
+        let known = vec!["providers.models.deepseek.default.api_key".to_string()];
+        assert_eq!(
+            resolve_field_path(&known, "providers.models.deepseek.default.api-key"),
+            "providers.models.deepseek.default.api_key",
         );
     }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -2057,6 +2057,37 @@ fn model_path_provider_type(path: &str) -> Option<&'static str> {
         .map(|p| p.name)
 }
 
+fn map_key_for_prop_path<'a>(section_path: &str, prop_path: &'a str) -> Option<&'a str> {
+    let tail = prop_path.strip_prefix(section_path)?.strip_prefix('.')?;
+    let mut parts = tail.split('.');
+    let key = parts.next().filter(|key| !key.is_empty())?;
+    parts.next()?;
+    Some(key)
+}
+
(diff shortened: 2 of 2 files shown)


WareWolf-MoonWall (approved):
Reviewed at e1dbba94c049fc25dc97883ba34dd2f5a9a52e32. CI nextest 8,481 passed ✅. No prior reviews or inline threads. Base branch: `integration/zeroclaw-tui`, stacked on #6848. Changes touch only `src/main.rs` and `crates/zeroclaw-config/src/helpers.rs`.

The workspace clippy failure is documented and inherited from #6848 (`tokio::spawn` in `factory.rs` and `http_request.rs`). That's clearly outside this PR's diff and is a known integration-branch artifact — the focused pass on `zeroclawlabs` and `zeroclaw-config` is the right scoped check here.

This is a clean, targeted fix for a real first-run UX cliff: `zeroclaw config set providers.models.deepseek.default.model` previously failed with `Unknown property` on a fresh config because the typed alias didn't exist yet. The approach — auto-materialize the alias when the path pattern matches `providers.models.<family>.<alias>.<field>` — is correctly scoped and well-guarded.

### 🟢 What looks good — materializer is correctly scoped to `providers.models.*`

`ensure_map_key_for_prop_path` filters `Config::map_key_sections()` to paths that `starts_with("providers.models.")` and `MapKeyKind::Map`. The `max_by_key(section_path.len())` picks the deepest matching section, which is the right tiebreaker for nested config. The `config_set_does_not_materialize_non_provider_map_keys` test verifies a `cost.rates.providers.models.openai.*` path is silently ignored — that regression guard is exactly what's needed to prevent the materializer from firing on coincidentally-shaped map paths.

### 🟢 What looks good — final-segment kebab-to-snake rule is narrow

The `resolve_field_path` change fires only for the **last** segment, and only when the canonical field contains `_` and the raw spelling uses `-`. This avoids the trap where an alias like `my_bot` would be silently rewritten to `my-bot` (aliases are snake_case-only per `validate_alias_key`). The test `resolve_field_path_canonicalizes_kebab_final_field_segments` pins this behavior.

### 🟢 What looks good — three targeted tests with correct scope

- `config_set_materializes_missing_typed_provider_alias` covers the create-then-set path and the kebab api-key spelling in one test.
- `config_set_does_not_materialize_non_provider_map_keys` covers the regression case.
- `resolve_field_path_canonicalizes_kebab_final_field_segments` pins the helpers.rs change independently.

Running each test in isolation before the workspace run is the right incremental validation discipline for a stacked PR.

### 🔵 Suggestion — second `prop_fields()` call in `main.rs` deserves an inline comment

```rust
if ensure_map_key_for_prop_path(&mut config, &path)? {
    let known_paths: Vec<String> =
        config.prop_fields().into_iter().map(|f| f.name).collect();
    path = zeroclaw_config::helpers::resolve_field_path(&known_paths, &path);
}
```

The reason a second `prop_fields()` call is needed — the first `known_paths` was built before the alias existed, so `api_key` wasn't yet a known path — is non-obvious to a future reader. A one-line `// known_paths must be rebuilt after materialization; the alias fields weren't present in the first pass` would save that person the mental overhead of tracing why the paths are recomputed.

### 🔵 Suggestion — `#[cfg(feature = "agent-runtime")]` on the materialization test is correct but limiting

The test flag is consistent with `ensure_map_key_for_prop_path` living in `main.rs` which already requires `agent-runtime`, so the gate is correct. Worth noting that `--no-default-features` test runs won't cover this path; the `cargo check -p zeroclawlabs --no-default-features` step in the PR's validation confirms the function is available in minimal builds, but the test itself is gated. That's acceptable given the stacked context — just something to carry forward when this lands in master.

PR #6858, opened 2026-05-22
fix(tui): clarify first-run empty states

Audacity88:
## Summary

- **Base branch:** `integration/zeroclaw-tui` (stacked on #6848, not intended for `master` by itself)
- **What changed and why:**
  - Adds first-run guidance to the TUI dashboard when there are no configured agents or active sessions, so a new user has an obvious next step.
  - Only shows that first-run guidance after the agent list loads successfully, so a transient daemon/RPC error does not masquerade as an empty setup.
  - Expands global help with Config and Chat shortcuts, adds a friendly default message before the first Chat prompt, and makes the Chat empty state point back to Config when no agent is enabled.
  - Refreshes Chat initialization when the user returns from Config, so the "press F2, then F4" path works after the first agent is added.
  - Clears the terminal on resize so the TUI can recover when the host terminal view repaints poorly.
  - Rewords daemon connection status and the Memories empty state so local sockets and unconfigured memory do not look like errors.
- **Scope boundary:** This does not change daemon, memory, provider, ACP, or config behavior. It only changes TUI display text and state handling.
- **Blast radius:** Limited to the `zeroclaw-tui` app in this stacked TUI branch. Existing CLI, gateway, runtime, and memory backend behavior are unchanged.
- **Linked issue(s):** Depends on #6848.
- **Labels:** `bug`, `risk: medium`, `size: M`, `onboard: tui`

## Validation Evidence (required)

- **Commands run and tail output:**
  - `git diff --check` — passed with no whitespace errors.
  - `cargo fmt -p zeroclaw-tui -- --check` — passed.
  - `cargo test -p zeroclaw-tui -- --nocapture` — passed; 18 TUI tests passed, with existing unused/dead-code warning noise from the draft TUI app.
  - `cargo test -p zeroclaw-tui empty_chat_hint -- --nocapture` — passed; 1 passed, 19 filtered out.
  - `cargo test -p zeroclaw-tui memory_not_configured -- --nocapture` — passed; 1 passed, 18 filtered out.
  - `cargo test -p zeroclaw-tui connection_summary -- --nocapture` — passed; 2 passed, 17 filtered out.
  - `CARGO_TARGET_DIR=<scratch-target> cargo build -p zeroclaw-tui` — passed; existing unused/dead-code warnings remain.

[...]

PR #7029, opened 2026-05-31
fix(zerocode): refresh empty states after setup

Audacity88:
## Summary

- **Base branch:** `integration/zeroclaw-tui` (stacked on #6848; review target is this small delta against the integration branch, not #6848's merge-readiness against `master`)
- **What changed and why:**
  - Refreshes Chat and Code/ACP when returning from a stale no-agent setup error, so manual Config setup no longer leaves either pane parked on old empty-state text.
  - Updates the no-agent Chat copy to point new users toward Quickstart or Config instead of only Config.
  - Makes the Memories not-configured message explain that persistent memory is optional until configured, and wraps the longer text in the dashboard pane.
  - Adds a behavior-level regression test that proves the refresh path re-requests `agents/status` and leaves the stale error state when agents become available.
- **Scope boundary:** This does not change daemon, config, memory backend, provider, or session protocol behavior. It only changes Zerocode pane entry handling and empty-state display text on the TUI integration branch, and it does not attempt to resolve #6848's broader integration conflicts.
- **Blast radius:** Limited to `apps/zerocode` in the stacked TUI branch. The affected paths are pane switching, Chat/ACP initialization retry behavior, and dashboard empty-state rendering.
- **Linked issue(s):** Depends on #6848. Supersedes #6858.
- **Labels:** `bug`, `risk: medium`, `size: S`, `onboard: tui`

```text
result: passed

cargo test -p zerocode chat_entry_refresh_reloads_agents_from_error_phase -- --nocapture
result: passed; 1 passed in src/main.rs

cargo test -p zerocode -- --nocapture
result: passed; 168 passed, 0 failed

result: passed

git diff --check
result: passed with no whitespace errors
```

- **Beyond CI — what did you manually verify?** Reviewed the final diff against the Chat, ACP wrapper, app mode-switch, dashboard memory rendering, and Fluent catalogue call sites. No live terminal smoke was run after this follow-up.
- **If any command was intentionally skipped, why:** CI-shaped workspace nextest and workspace-shaped clippy were not run locally because this PR is a small reviewable delta stacked on #6848, whose integration branch still has broader merge-readiness work. This does not block review of #7029's diff. Before this change is merged, retargeted to `master`, or folded into #6848, run the current stack-level CI-shaped test job and workspace-shaped clippy for the relevant branch state, or explicitly accept GitHub CI coverage for the stack.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No upgrade steps. This only changes TUI display and pane-entry retry behavior.

- **Fast rollback command/path:** `git revert a49aa06af7a6608821230982b6f87d2e1aef5d02`
- **Observable failure symptoms:** TUI-only regressions such as Chat or Code repeatedly reloading on pane entry, no-agent setup guidance failing to update after Config creates/enables an agent, or the Memories tab showing awkward wrapped warning text.

- Superseded PRs + authors: #6858 by @Audacity88
- Scope materially carried forward: The stale first-run empty-state follow-up is rebuilt on the current `integration/zeroclaw-tui` branch. This PR carries forward the Chat no-agent refresh/copy and Memory not-configured guidance, then narrows the implementation around the current Quickstart/Config flow and adds ACP parity.
- If `No`, why: Same author; no third-party code or design was incorporated.

### DIFF
--- a/apps/zerocode/locales/en/zerocode.ftl
+++ b/apps/zerocode/locales/en/zerocode.ftl
@@ -108,7 +108,7 @@ zc-dashboard-tab-health = Health
 zc-dashboard-tab-cost = Cost
 zc-dashboard-tab-cron = Cron
 
-zc-dashboard-memory-not-configured = Memory subsystem not configured
+zc-dashboard-memory-not-configured = Memory is not configured yet. Use Quickstart or Config to add a memory backend, or ignore this tab until you need persistent memory.
 zc-dashboard-search-action-apply = apply
 zc-dashboard-search-action-cancel = cancel
 zc-dashboard-search-prefix = search:
@@ -265,7 +265,7 @@ zc-quickstart-save-and-close = Save & close
 zc-chat-pane-chat = Chat
 zc-chat-pane-acp = ACP
 
-zc-chat-no-agents = No enabled agents. Configure an agent in the Config tab.
+zc-chat-no-agents = No enabled agents yet. Open Quickstart to create one, or use Config to add and enable an agent.
 zc-chat-error-fetch-agents = Failed to fetch agents: { $error }
 zc-chat-error-create-session = Failed to create session: { $error }
 
--- a/apps/zerocode/src/acp.rs
+++ b/apps/zerocode/src/acp.rs
@@ -22,6 +22,10 @@ impl Acp {
         self.inner.init().await
     }
 
+    pub(crate) async fn refresh_if_inactive(&mut self) {
+        self.inner.refresh_if_inactive().await;
+    }
+
     pub(crate) fn draw(&mut self, frame: &mut ratatui::Frame, area: Rect) {
         self.inner.draw(frame, area);
     }
--- a/apps/zerocode/src/app.rs
+++ b/apps/zerocode/src/app.rs
@@ -83,6 +83,27 @@ impl Mode {
     }
 }
 
+async fn switch_mode(
+    mode: &mut Mode,
+    next: Mode,
+    conn_state: &ConnectionState,
+    quickstart: &mut quickstart_pane::QuickstartPane,
+    acp_pane: &mut acp::Acp,
+    chat_pane: &mut chat::Chat,
+) {
+    if *mode == Mode::Quickstart && next != Mode::Quickstart {
+        quickstart.dismiss_beacon().await;
+    }
+    if !matches!(conn_state, ConnectionState::Disconnected { .. }) {
+        match next {
+            Mode::Acp => acp_pane.refresh_if_inactive().await,
+            Mode::Chat => chat_pane.refresh_if_inactive().await,
+            _ => {}
+        }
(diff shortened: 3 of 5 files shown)


WareWolf-MoonWall (approved):
Reviewed at a49aa06af7a6608821230982b6f87d2e1aef5d02. CI passing (168 tests, 0 failed). No prior reviews or inline threads. Base branch: `integration/zeroclaw-tui`, stacked on #6848. Supersedes #6858 by same author — attribution section is correct, no `Co-authored-by` trailers needed.

This is a focused, well-scoped fix for a real UX gap: after a user configures an agent manually in Config, Chat and ACP were parked on stale "no agents" error text and required a full restart to clear. The refresh hook, the mode-switch extraction, and the dashboard wrapping fix are all the right shape for this problem.

### 🟢 What looks good — behavior regression test uses the full dispatch cycle

`chat_entry_refresh_reloads_agents_from_error_phase` is not a pure unit test of phase state — it drives the mock RPC channel with a real `agents/status` request, waits with a timeout-bounded `rx.recv()`, and then injects a response via `rpc.dispatch_response(...)`. It catches the case where the refresh fails to emit the right RPC method, the case where it doesn't complete after a response, and the final state assertion is specific (`ChatPhase::PickAgent { agents, loading: false }`). That's exactly the right level of behavioral coverage for a retry path.

### 🟢 What looks good — `switch_mode` extraction eliminates duplication

Before this PR, the `quickstart.dismiss_beacon().await` call was copy-pasted in both the keyboard-navigation branch and the mouse-click branch in `app.rs`. The extracted `switch_mode` function consolidates both sites with the `ConnectionState::Disconnected` guard, so a future change to mode-switch behavior has exactly one place to land.

### 🟢 What looks good — disconnection guard is in the right place

`if !matches!(conn_state, ConnectionState::Disconnected { .. })` gates the refresh. Calling `refresh_if_inactive()` with no active daemon connection would produce a spurious RPC call that can never complete. Placing the guard in `switch_mode` means both navigation paths are protected without needing the guard repeated.

### 🔵 Suggestion — `zc-dashboard-memory-not-configured` is now a paragraph

The new string is 131 characters. That's usable and the `Wrap { trim: true }` added to the dashboard renderer handles it correctly. Worth tracking for translators: this Fluent key went from a short status phrase to a multi-sentence instructional paragraph, which changes translation difficulty and appropriate length targets. Not a blocker — just something to flag when the localization pass happens.

### 🔵 Suggestion — `Wrap { trim: true }` trims all leading whitespace

`trim: true` on `ratatui::widgets::Wrap` removes leading spaces per line. For the current dashboard warning path this is fine. If a future message variant needs to display indented text (code paths, nested config keys), that path will need `Wrap { trim: false }`. The behavior difference is non-obvious enough to warrant a short inline comment like `// trim: leading spaces stripped per line — intentional for prose-only warning text`.

PR #7034, opened 2026-05-31
fix(channels/whatsapp-web): match LID bot mentions

### LINKED ISSUE #7032: [Bug]: WhatsApp Web mention_only misses bot mentions addressed to LID JID
### Affected component

channel

### Severity

S2 - major feature broken

### Current behavior

In WhatsApp Web group chats with `mention_only = true`, the bot can ignore messages that structurally mention the bot.

The mention gate currently compares `mentioned_jid` values only against the bot phone JID digits. Modern WhatsApp sessions may address mentions to the bot's LID JID instead. In that case, `contains_bot_mention` sees the structural mention but rejects it because the LID digits do not equal the phone digits.

There is a second identity edge case: the device snapshot `pn.user()` value can include a device index suffix such as `15551234567:16`. Filtering all digits from that value turns it into `1555123456716`, which no longer matches the bare phone JID digits.

### Expected behavior

When `mention_only = true`, WhatsApp Web group mention detection should accept a structural mention that matches either the bot phone JID digits or the bot LID JID digits.

Bot phone/LID identity should also strip any `:N` device suffix before extracting digits, so the stored identity remains the bare JID user value.

### Steps to reproduce

```text
1. Configure WhatsApp Web with `mention_only = true`.
2. Use a linked-device session where the bot device snapshot has both a PN JID and a LID JID.
3. In a group chat, send a message that mentions the bot from the WhatsApp UI.
4. Inspect the inbound `mentioned_jid` values; the mention can target `<bot-lid>@lid` rather than `<bot-phone>@s.whatsapp.net`.
5. The bot drops the message as "without bot mention" even though the message structurally mentions the bot.
```

Unit-level reproduction:

```text
contains_bot_mention("hey @DisplayName", ["76188559093817@lid"], "15551234567") returns false today, even when `76188559093817` is the bot's own LID.
```

### Impact

WhatsApp Web group bots configured to only respond when mentioned can silently ignore valid mentions. Operators then have to disable `mention_only`, which makes the bot respond to all group messages, or rely on fragile text fallback behavior instead of the structured WhatsApp mention data.

### Logs / stack traces

```text
ignoring group message without bot mention
```

### ZeroClaw version

`master` at `9caf28cb9` (v0.8.0 beta line)

### Rust version

Not version-specific.

### Operating system

Platform-independent.

### Regression?

Unknown

### Pre-flight checks

- [x] I reproduced this against the latest `master` branch behavior.
- [x] I checked existing open issues and pull requests for a duplicate.
- [x] I redacted secrets, tokens, phone numbers, and personal data from submitted content.

rifuki:
## Summary

**Base branch:** `master`

**What changed / why:**

Fixes WhatsApp Web `mention_only` group gating when WhatsApp addresses the bot mention to the bot's LID JID instead of its phone JID.

The channel now stores the bot LID from the device snapshot, strips `:N` device suffixes before digit extraction, and checks structured/text mentions against both bot phone digits and bot LID digits.

**Scope boundary:**

Only WhatsApp Web bot-mention identity matching is changed. This does not change WhatsApp reply routing, allowlist policy, media download/caption handling, Telegram behavior, or provider behavior.

**Blast radius:**

Medium. This only affects `crates/zeroclaw-channels/src/whatsapp_web.rs`, specifically WhatsApp Web `mention_only` group-message gating and bot identity resolution.

**Linked issues:**

Fixes #7032

## Validation Evidence

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-channels --features whatsapp-web contains_bot_mention -- --nocapture
cargo clippy -p zeroclaw-channels --features whatsapp-web --all-targets -- -D warnings
```

Focused test result: 9 passed, 0 failed, 1220 filtered out.

## Security & Privacy

No secrets or tokens are added. No third-party phone numbers or sender identities are added; tests use synthetic placeholder JID digits only. Runtime diagnostics may log the bot account's own resolved phone/LID identity for operator-local troubleshooting.

## Compatibility

Existing phone-JID mention behavior is preserved. LID matching is additive, and empty/unknown bot identity values are ignored so the text fallback does not accidentally match bare `@` characters. Device-suffix stripping normalizes existing PN/LID values to their bare JID user digits.

## Rollback

Revert this commit to return WhatsApp Web `mention_only` matching to phone-JID-only behavior.

## Supersede Attribution

No superseded PR. This is a narrow follow-up to current WhatsApp Web LID behavior on `master`.

### DIFF
--- a/crates/zeroclaw-channels/src/whatsapp_web.rs
+++ b/crates/zeroclaw-channels/src/whatsapp_web.rs
@@ -75,6 +75,8 @@ pub struct WhatsAppWebChannel {
     mention_only: bool,
     /// Bot phone number (digits only), resolved from pair_phone or device identity at runtime
     bot_phone: Arc<Mutex<Option<String>>>,
+    /// Bot LID number (digits only), resolved from device identity at runtime
+    bot_lid: Arc<Mutex<Option<String>>>,
     /// Usage mode (business vs personal policy filtering)
     mode: zeroclaw_config::schema::WhatsAppWebMode,
     /// DM policy when mode = personal
@@ -161,6 +163,7 @@ impl WhatsAppWebChannel {
             peer_resolver,
             mention_only,
             bot_phone: Arc::new(Mutex::new(bot_phone)),
+            bot_lid: Arc::new(Mutex::new(None)),
             mode,
             dm_policy,
             group_policy,
@@ -705,7 +708,22 @@ impl WhatsAppWebChannel {
     #[cfg(feature = "whatsapp-web")]
     fn jid_digits(jid: &str) -> String {
         let user_part = jid.split_once('@').map(|(u, _)| u).unwrap_or(jid);
+        let user_part = user_part
+            .split_once(':')
+            .map(|(u, _)| u)
+            .unwrap_or(user_part);
         user_part.chars().filter(|c| c.is_ascii_digit()).collect()
+    }
+
+    #[cfg(feature = "whatsapp-web")]
+    fn store_jid_digits(slot: &Arc<Mutex<Option<String>>>, jid: &str) -> Option<String> {
+        let digits = Self::jid_digits(jid);
+        if digits.is_empty() {
+            None
+        } else {
+            *slot.lock() = Some(digits.clone());
+            Some(digits)
+        }
     }
 
     /// Extract mentioned JIDs from the base (unwrapped) message's context_info.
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
Context: I reviewed head `811104ea7503ea2db6358b852ac2eab348dce312`, the one-file WhatsApp Web diff, #7032, the PR body, green GitHub checks, and the empty prior review/comment state.

### 🟢 What looks good — LID mention matching is narrowly scoped

This fixes the right layer for #7032. The mention gate now compares structured mentions against both the bot phone digits and bot LID digits, strips `:N` device suffixes before storing identity digits, and keeps empty identity values from matching text fallback accidentally. The focused tests cover phone preservation, LID structured mentions, phone-unknown LID matching, and the suffix-stripping edge case.

I do not see a blocking issue in this slice. One non-blocking cleanup before merge: the PR body privacy text should be updated to say no third-party phone numbers or sender identities are added, since the runtime can log the bot account's own resolved phone/LID identity for operator-local diagnostics.

PR #7036, opened 2026-05-31
feat(web): add ACP console

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds a browser ACP console at `/acp-console` so authenticated dashboard users can exercise the gateway ACP endpoint from the web UI.
  - Adds a small JSON-RPC ACP WebSocket client that uses the existing dashboard auth token and Tauri/web base-path helpers.
  - Adds a lightweight agent picker source so the console does not load dashboard sessions, costs, memory counts, or peer-group data just to choose an agent.
  - Routes Vite dev-server `/acp` WebSocket traffic to the configured gateway target.
- **Scope boundary:** This PR does not change the ACP server protocol, runtime behavior, permission policy, or agent configuration schema.
- **Blast radius:** Web dashboard routing/navigation, Vite local proxy behavior, and browser-origin ACP clients.
- **Linked issue(s):** None.
- **Labels:** `enhancement`, `risk: medium`, `size: XL`

```bash
$ npm run build
dist/assets/AcpConsole-BDOEo4qU.js               17.88 kB │ gzip:   5.39 kB
dist/assets/index--auAH8aS.js                   156.63 kB │ gzip:  47.42 kB
dist/assets/index-BdsCrMZ1.js                   761.46 kB │ gzip: 193.96 kB
dist/assets/Config-6--s8nb4.js                  840.81 kB │ gzip: 277.73 kB

(!) Some chunks are larger than 500 kB after minification.
✓ built in 16.11s
```

```bash
$ git diff --check
# no output
```

```bash
$ curl -I http://127.0.0.1:5179/acp-console
HTTP/1.1 200 OK
Content-Type: text/html
```

```bash
$ node acp-smoke.js
ok: true
session/prompt content: ACP smoke reply from local mock.
stopReason: end_turn
```

  - Verified the Vite dev server serves `/acp-console`.
  - Started a scratch gateway config with one enabled agent and a local OpenAI-compatible mock provider.
  - Paired a scratch client, opened an ACP WebSocket, sent `initialize`, sent `session/new` without a browser-forced `agentAlias`, then sent `session/prompt`.
  - Confirmed the prompt reached the local mock provider with `POST /v1/chat/completions` and returned `ACP smoke reply from local mock.`
  - Stopped the scratch gateway, mock provider, and Vite server after validation.
  - Full workspace Rust validation was skipped because this is a web UI/client change with no Rust source changes.

  - This adds an authenticated dashboard ACP client capability for opening `/acp`, creating sessions, sending prompts, and responding to ACP permission requests. It uses the existing dashboard token path and does not add new credential storage, server-side permission policy, or file-system access scope.

## Compatibility (required)

- Backward compatible? `Yes`

- **Fast rollback command/path:** `git revert 082331f3d 5462f650a`
- **Observable failure symptoms:** `/acp-console` fails to load, ACP WebSocket connection errors in the dashboard, or unexpected `/acp` proxy errors during local Vite development.


WareWolf-MoonWall (approved):
@Audacity88 — Context for this pass: I checked the current diff at head 423c77c300bb03ea58fe2c01de99dede2131275c (3 commits: initial feature at 082331f3, console page at 5462f650, and a master-merge at 423c77c3), the full PR body validation evidence, GitHub CI status (all green, 13 checks passed), and prior reviews/inline threads (none exist yet). This is a fresh review pass with no prior review activity on this PR.

### 🟢 What looks good — Clean browser-ACP client implementation

The implementation direction is exactly right for what the PR claims: a lightweight browser ACP client for authenticated dashboard users. The structural choices are sound:

- **Separate client library:** `web/src/lib/acp.ts` implements a JSON-RPC 2.0 WebSocket client with typed ACP frames (`initialize`, `session/new`, `session/prompt`), proper pending-request tracking, timeout handling, and cleanup. This is a new standalone client that doesn't force the heavy dashboard session/cost/memory machinery onto every ACP route.
- **Minimal agent picker:** `loadAgentPickerSummaries()` in `agents.ts` is intentionally separate from `loadAgentSummaries()` — it fetches only `alias` and `enabled` instead of pulling full dashboard state. The comment explaining why is load-bearing for future maintainers.
- **Vite proxy routing:** The Vite config change (`'/acp': target`) routes local `/acp` WebSocket traffic to the configured gateway during development, which is the right pattern for localhost ACP testing without requiring a separate local gateway config change.

This is exactly the kind of separation we want between tooling routes (ACP console, Canvas) and production dashboard routes.

### 🟢 What looks good — i18n coverage

Every new UI string in the console (`acp.title`, `acp.status.*`, `acp.error.*`, etc.) has entries in the three supported locales (en, zh-CN, tr). The English strings are specific (`"Server default"`, `"Creating..."`, `"Failed to initialize ACP session"`) rather than generic, which makes debugging operator-facing errors easier. The pattern here matches the existing dashboard i18n discipline.

### 🔵 Suggestion — Protocol log display depth

The ACP console includes a protocol log (`Protocol log` pane) that shows raw JSON-RPC frames. The current implementation appears to render every inbound/outbound frame without a depth limit, which means a long-running session will accumulate unbounded memory on the browser side. Consider either:

1. A depth cap (e.g., last 100 frames), or
2. A clearable protocol log with a "Clear" button.

This is optional — the console is a developer-facing diagnostic tool, not a production operator surface, so unbounded accumulation is less likely to hurt real workflows. But naming it so you can choose knowingly.

### 🔵 Suggestion — ACP error display granularity

The console error strings (`acp.error.init_failed`, `acp.error.prompt_failed`, etc.) are currently one-liners. If the ACP server returns a structured JSON-RPC error with a `code` and `message`, consider surfacing the full error text instead of just "Failed to initialize ACP session." The ACP protocol already carries error details; showing them would make debugging gateway-side failures easier for users who hit this console during local development.

Again optional — the current error strings are already better than generic "Something went wrong" messages, so this is polish rather than a gap.

The PR body includes concrete evidence for the claims:

- `npm run build` output showing the new `AcpConsole-BDOEo4qU.js` chunk (17.88 kB gzipped) and the expected size warning for `Config-6--s8nb4.js` (840 kB, which is pre-existing).
- `curl -I http://127.0.0.1:5179/acp-console` showing the route serves.
- `node acp-smoke.js` output confirming the ACP WebSocket flow (initialize → session/new → session/prompt → reply) works end-to-end with a local mock provider.

The "Beyond CI" section explains why full workspace Rust validation was skipped (no Rust source changes) and what was manually verified (Vite dev server, gateway scratch config, ACP WebSocket flow, local mock provider round-trip). This is the kind of targeted validation that makes sense for a web-only change — no need to re-run the full Rust test suite when zero `.rs` files changed.

### ✅ Passes discipline checks

- **Risk classification:** `risk: medium` is correct — this changes dashboard routing, Vite proxy behavior, and adds a new browser-origin ACP client capability. The rollback section names the fast path (`git revert 082331f3d 5462f650a`) and observable failure symptoms.
- **Security section:** The "New permissions, capabilities, or file system access scope?" answer is `Yes`, and the explanation is accurate — it adds authenticated dashboard ACP client capability, uses the existing dashboard token path, and doesn't introduce new credential storage or server-side permission policy.
- **Compatibility:** `Yes` to backward compatible, `No` to config/env/CLI surface changed, and the explanation confirms no upgrade steps required. Correct.
- **Scope boundary:** "This PR does not change the ACP server protocol, runtime behavior, permission policy, or agent configuration schema" — accurate. The diff confirms this is purely a browser client and routing change.

### Summary

This is a clean, well-scoped addition. The implementation follows the right patterns (separate lightweight client, minimal agent picker, Vite proxy for local dev), the i18n coverage is complete, and the validation evidence is concrete and targeted to what actually changed. The two 🔵 suggestions (protocol log depth, error display granularity) are polish items that don't block merge — you shipped a working diagnostic console, and those are follow-up refinements.

CI is green, no prior review blocks exist, and the code matches what the PR body claims. Approving.

tidux (commented):
@Audacity88 — context for this pass: I read the current diff at 423c77c300bb03ea58fe2c01de99dede2131275c (3 commits: 082331f3 client, 5462f650 page, 423c77c3 merge), cross-checked the new client against the gateway/server side (`crates/zeroclaw-gateway/src/acp.rs`, `crates/zeroclaw-channels/src/orchestrator/acp_server.rs`), ran the regex in `vite.config.ts` against representative paths, and ran `npm ci && npm run build` locally. @WareWolf-MoonWall's approval on this same head already covers the high-level shape — I'm not re-litigating that. What follows are findings their pass didn't surface plus one factual correction I want on the record so it doesn't propagate.

### 🟢 What looks good — Reconnect/cancel fencing

The `connectionSeqRef` discipline in `AcpConsole.tsx` is the part most likely to harbor bugs in a long-lived WS UI, and it's done well. Every callback closure captures the seq at construction time and every state-mutating path runs through `isCurrentConnection(client, connectionSeq)` before touching React state. `sendPrompt` correctly captures both the local `client` reference (line 427) **and** the seq (line 429), so a pending `await client.request(...)` resolving after a reconnect doesn't write into the new transcript. The `useEffect` cleanup at L411-413 disconnects on unmount. This is the right pattern.

### 🟢 What looks good — Vite proxy regex bounding

`"^/acp(?:\\?.*)?$"` instead of a bare `"/acp"` prefix was deliberate and necessary. I checked the behavior:

```
"/acp"             -> true
"/acp?token=x"     -> true
"/acp/sub"         -> false
"/acp-console"     -> false
"/acpfoo"          -> false
```

A `"/acp"` prefix match would have shadowed the SPA route `/acp-console` and proxied it to the gateway as a non-WS request, which would have manifested as a confusing 404 the moment someone tried to navigate to the new page in dev. The end-anchored regex with the optional query group is the right call; a one-line comment on that line wouldn't hurt the next person who looks at this rule.

### 🟢 What looks good — Server-contract alignment

I walked the client against the server. All four touchpoints line up:

| Client | Server |
|---|---|
| `Sec-WebSocket-Protocol: zeroclaw.acp.v1, bearer.<token>` | `crates/zeroclaw-gateway/src/acp.rs:19,183` |
| `session/new` with `{ agentAlias }` | `acp_server.rs:514-521` (accepts `agentAlias` / `agent_alias` / `agent`, auto-selects sole agent when omitted) |
| `session/prompt` with `{ sessionId, prompt: <string> }` | `acp_server.rs:1486-1528` (`parse_prompt` accepts plain string as a ZeroClaw affordance over strict ACP `ContentBlock[]`) |
| `session/cancel` as a JSON-RPC notification (no id) | `acp_server.rs:379`, cancel token at 1209 |

The smoke test in the PR body exercises the right path end-to-end.

### 🟡 Warning — `request()` can leak a pending entry on synchronous send failure

In `web/src/lib/acp.ts:171-184`, the pending entry is registered in `this.pending` and then `this.ws?.send(...)` is called. If the socket transitions to CLOSING between the `this.connected` check at L159 and the `send` call, `WebSocket.send` throws `InvalidStateError` synchronously. That exception escapes the `new Promise` executor, which is caught by the Promise machinery and rejects the promise — but the `pending` map entry is **already inserted** and never cleared until its 120 s `DEFAULT_REQUEST_TIMEOUT_MS` fires. During those two minutes:

- The id is non-reclaimable (next id will be different, but the closed-over `timeout` keeps memory live).
- If the user reconnects and a server response happens to reuse the same numeric id, `resolveResponse` will route it to the dead entry.

The principle: any code that registers a tracked side-effect *before* the operation that can fail should wrap the operation so the registration is rolled back on failure. Fix is local:

```ts
this.pending.set(id, { resolve: (v) => resolve(v as T), reject, timeout });
try {
  this.ws?.send(JSON.stringify(frame));
} catch (err) {
  clearTimeout(timeout);
  this.pending.delete(id);
  reject(err instanceof Error ? err : new Error(String(err)));
}
```

Same pattern applies less critically to `notify`, `respond`, and `respondError`, which currently swallow nothing and would surface the throw to the caller — those are fine; only `request` has the leak.

This is a "real but rare" bug, hence 🟡 rather than 🔴. Not a merge blocker on its own.

### 🔵 Suggestion — Factual correction: `acp.*` strings are English-only, not in three locales

I want this on the record because @WareWolf-MoonWall's commendation cited "entries in the three supported locales (en, zh-CN, tr)" — I checked and that's not what shipped. The dashboard's `web/src/lib/i18n.ts` carries 31 locales, and the diff added:

- `'nav.acp': 'ACP'` in all 31 locales (correct, consistent with `nav.canvas` discipline).
- `'acp.title'`, `'acp.subtitle'`, `'acp.error.*'`, etc. in **English only** (one block, around line 418).

The runtime is forgiving: `t()` in `i18n.ts:11260` falls back `currentLocale → en → key`, so every non-English user transparently gets the English ACP copy. No crashes, no missing strings. It's the same shortcut several other recent feature additions have taken.

I'm raising it as 🔵 not 🟡 because (a) the fallback chain makes it strictly cosmetic, (b) you're consistent with how other newly-landed pages handled their first cut. Two ways to land this: either file a follow-up issue for the i18n sweep at the same time you merge, or run one pass yourself to mirror `acp.*` into the other 30 locales before merge. Your call. The thing I'd want avoided is leaving the impression that this PR achieved full-locale parity when it actually relies on the en-fallback.

### 🔵 Suggestion — Agent-picker change silently wipes the transcript

`connect` depends on `selectedAgentAlias` (deps list at `AcpConsole.tsx:402`), and the `useEffect` at L405-414 re-runs when `connect`'s identity changes. The net effect: changing the agent in the picker tears down the current ACP session and clears `messages`/`events`/`permissions` with no warning. This is the right *mechanical* behavior (ACP sessions are agent-scoped, you can't hot-swap mid-turn) but the UX consequence (their transcript is gone) is hidden.

Minimum-effort fix: a confirm dialog when `messages.length > 0`, or just a small visible-state hint near the picker ("changing agent starts a new session"). Optional — this is a developer-facing console, not an operator surface.

### 🔵 Suggestion — Token in URL query duplicates `bearer.<token>` subprotocol

`web/src/lib/acp.ts:124-128` appends `?token=<token>` to the WebSocket URL *and* sends `bearer.<token>` as a `Sec-WebSocket-Protocol` value. The server (`acp.rs:33,168-187`) accepts either; the subprotocol path is strictly safer because tokens in the URL land in access logs and survive in browser devtools/history even after disconnect.

I'm not flagging this as a regression because the existing `web/src/lib/ws.ts:82-91` for `/ws/chat` does the same dual-source dance — this PR is consistent with house style. But that house style is worth fixing dashboard-wide as a follow-up. If you want to take this on inside this PR, drop the query string and rely on the subprotocol alone; the server will keep working.

### Summary

The shape and intent of this PR are right and the structural reviewer has already approved. My findings are: one 🟡 worth a 6-line defensive fix (`ws.send` try/catch in `request()`), one 🔵 factual correction about i18n coverage so the next reader knows what shipped, and two 🔵 UX/security follow-up nudges. Posting as a comment because nothing here is something I'd block merge on personally — but I'd prefer the 🟡 land before this ships.

Build (`tsc -b && vite build`) passes locally. CI is green. Approved by @WareWolf-MoonWall on this head.

singlerider (approved):
@Audacity88 — context for this pass: I read the current diff at head 60e7b0e5, both prior reviews (@WareWolf-MoonWall's approval and @tidux's comment, both at 423c77c3), and diffed the one commit that landed since — 60e7b0e5 `fix(web): clean up ACP pending request on send failure`. No open inline threads. I'm not re-litigating the structural approval; this pass confirms the one defensive finding got addressed and the shape still holds.

### ✅ Resolved — `request()` pending-entry leak on synchronous send failure

@tidux's 🟡 (the only finding either prior pass would have wanted landed before merge) is fixed at 60e7b0e5, and the fix is exactly the rollback discipline that was asked for: `request()` now wraps `ws.send()` in try/catch and, on a synchronous `InvalidStateError`, runs `clearTimeout(timeout)` + `this.pending.delete(id)` + `reject(...)` before the entry can go stale (acp.ts:181-187). That closes the window where a CLOSING socket between the `connected` check and `send` would have left a tracked entry alive for the full 120s timeout and risked a reused-id mis-route after reconnect. Scoped correctly to `request()` only — `notify`/`respond`/`respondError` surface the throw to the caller and never had the leak.

### 🟢 What looks good — reconnect/cancel fencing

The `connectionSeqRef` discipline in `AcpConsole.tsx` is the part most likely to harbor bugs in a long-lived WS UI and it's done right: every state-mutating path runs through `isCurrentConnection(client, connectionSeq)`, and `sendPrompt` captures both the local `client` ref and the seq before the `await`, so a request resolving after a reconnect can't write into the new transcript. This is the correct pattern for a session that can tear down and rebuild underneath in-flight requests.

### 🟢 What looks good — the agent picker doesn't drag in dashboard state

`loadAgentPickerSummaries()` fetching only `alias` + `enabled` instead of reusing `loadAgentSummaries()` is the right separation. The ACP console is a tooling route; making it pull sessions, costs, memory counts, and peer-group data just to populate a dropdown would have coupled a diagnostic surface to the full dashboard data path. Keeping it lean — with the load-bearing comment explaining why — is exactly the boundary discipline these tooling routes should keep.

### 🟢 What looks good — server-contract alignment and Vite proxy bounding

The end-anchored `^/acp(?:\?.*)?$` proxy rule correctly avoids shadowing the `/acp-console` SPA route (a bare `/acp` prefix would have proxied the new page to the gateway as a non-WS request and 404'd in dev), and the client's four touchpoints — subprotocol auth, `session/new`, plain-string `session/prompt`, notification-shaped `session/cancel` — line up with the gateway/`acp_server.rs` side. @tidux walked this end-to-end on the prior head and it's unchanged here.

Approving. The structural review was already favorable, the one finding worth landing before merge is in, and this last commit is a tight 6-line defensive fix. The remaining polish items both reviewers raised (protocol-log depth, full-locale `acp.*` mirroring, token-in-URL dashboard-wide cleanup) are real but are follow-up refinements on a developer-facing diagnostic console, not merge gates — worth a tracked follow-up issue rather than holding this.

PR #7040, opened 2026-05-31
fix(channel): restore WhatsApp interrupt_on_new_message

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Restores the `interrupt_on_new_message` field on `WhatsAppConfig`, defaulting to `false` for existing configs.
  - Wires `channels.whatsapp.default.interrupt_on_new_message` into the channel orchestrator's in-flight cancellation switch so WhatsApp can cancel an older response when a newer message arrives from the same sender/chat.
  - Adds focused schema and orchestrator regression coverage, including a behavior-level WhatsApp dispatch test.
  - Documents `interrupt_on_new_message` as a shared WhatsApp Cloud API and Web mode option.
- **Scope boundary:** This does not change alias-scoped cancellation semantics for WhatsApp or any other channel. The current interrupt switch still reads the `default` alias, matching the existing Telegram, Slack, Discord, Mattermost, and Matrix behavior.
- **Blast radius:** Limited to WhatsApp config parsing/documentation and channel orchestrator cancellation behavior. Existing configs remain compatible because the new field defaults to `false`.
- **Linked issue(s):** Related #6074. Supersedes #4371.
- **Labels:** `bug`, `risk: medium`, `size: M`, `channel`, `channel:whatsapp`, `config`, `docs`

  - `cargo fmt --all -- --check` - passed.
  - `git diff --check` - passed.
  - `cargo test -p zeroclaw-config whatsapp_config_ -- --nocapture` - passed; 6 tests passed.
  - `cargo test -p zeroclaw-channels --lib message_dispatch_interrupts_in_flight_whatsapp_request_and_preserves_context -- --nocapture` - passed; 1 test passed.
  - `cargo test -p zeroclaw-channels --lib interrupt_on_new_message_ -- --nocapture` - passed; 6 tests passed.
  - `cargo clippy -p zeroclaw-config --lib -- -D warnings` - passed.
  - `cargo clippy -p zeroclaw-channels --lib -- -D warnings` - passed.
  - `cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings` - passed.
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` - passed; 8297 tests passed, 10 skipped.
- **Beyond CI — what did you manually verify?** Traced both WhatsApp backends and confirmed they emit `channel = "whatsapp"`, so the restored channel switch applies to Cloud API and Web mode. Traced the orchestrator interruption key and intentionally kept the existing sender/chat scope rather than adding new alias-scoped semantics in this recovery PR.
- **If any command was intentionally skipped, why:** `N/A`

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: Existing WhatsApp configs continue to load because `interrupt_on_new_message` defaults to `false`. Users who want the restored behavior can add `interrupt_on_new_message = true` under `[channels.whatsapp.default]`.

- **Feature flags or config toggles:** Set or omit `channels.whatsapp.default.interrupt_on_new_message = false` to disable the restored behavior without reverting.
- **Observable failure symptoms:** WhatsApp responses from the same sender/chat may cancel unexpectedly when a newer message arrives, or may fail to cancel when the option is enabled.

- Superseded PRs + authors (`#<pr> by @<author>`, one per line): `#4371 by @tmigone`
- Scope materially carried forward: Restores the WhatsApp `interrupt_on_new_message` config surface, runtime channel cancellation wiring, and docs; this PR adds fresh schema/orchestrator regression coverage.


crates/zeroclaw-channels/src/orchestrator/mod.rs
```diff
@@ -344,11 +345,43 @@ impl InterruptOnNewMessageConfig {
             "discord" => self.discord,
             "mattermost" => self.mattermost,
             "matrix" => self.matrix,
+            "whatsapp" => self.whatsapp,
             _ => false,
         }
     }
 }
 
+fn interrupt_on_new_message_config(
```
singlerider, line 374: 🟢 This is the right move — extracting the five previously copy-pasted per-channel `get("default").is_some_and(...)` reads into one constructor and adding whatsapp through it, instead of pasting a sixth block at the wiring site. The next channel that wants this switch now extends one helper. Verified both backends emit `channel = "whatsapp"` (whatsapp.rs:377 Cloud API, whatsapp_web.rs:975 Web), so the single `"whatsapp" => self.whatsapp` arm covers both without per-backend wiring.

singlerider (approved):
Read the full diff, both WhatsApp backends, and the orchestrator interrupt switch. Clean recovery fix that also removes duplication along the way. Approving.

### 🟢 What looks good — the fix collapses five inline duplicates into one helper

The valuable part of this change isn't just adding the sixth channel — it's that `interrupt_on_new_message_config()` extracts the five previously-inline-duplicated per-channel reads (the `config.channels.<ch>.get("default").is_some_and(...)` blocks for telegram/slack/discord/mattermost/matrix that were copy-pasted in `start_channels`) into a single constructor, then adds whatsapp through it. The wiring site goes from ~30 lines of repeated extraction down to one call. That's the right instinct: the next channel that wants this switch extends one helper instead of pasting a sixth block.

### 🟢 What looks good — the "applies to both backends" claim is verified

The PR rests on both WhatsApp backends emitting `channel = "whatsapp"`, and that holds: Cloud API (`whatsapp.rs:377`) and Web mode (`whatsapp_web.rs:975`) both return `"whatsapp"` from `name()` and stamp `channel: "whatsapp"` on inbound messages. So the single `"whatsapp" => self.whatsapp` arm correctly covers Cloud API and Web without per-backend wiring. The scope-boundary note is also honest: it keeps the existing `default`-alias read (matching every other channel) rather than smuggling in new alias-scoped semantics under a recovery PR.

### 🟢 What looks good — the behavior test pins interrupt-and-preserve, not just the flag

`message_dispatch_interrupts_in_flight_whatsapp_request_and_preserves_context` is the test that matters: it drives two messages from the same sender through the real dispatch loop and asserts the older response is cancelled (only one send, carrying `response-2`), the second provider call carries both user turns, and crucially that the cancelled turn persists no assistant message. That's the actual contract — interrupt without losing conversation context — not just "the bool deserializes." The schema-level default-false and round-trip tests back the compatibility claim.

### 🟢 What looks good — backward compatibility and the test-struct churn

`#[serde(default)]` with default `false` keeps every existing WhatsApp config loading unchanged, and the large `+whatsapp: false,` repetition across the test module is the unavoidable mechanical cost of adding a field to `InterruptOnNewMessageConfig` — every constructor must name it, and they all correctly default it off. Docs cover it as a shared Cloud/Web option. Supersede attribution for #4371 by @tmigone is present with the co-author trailer.

Approving. Correct, well-tested, compatible, and it leaves the wiring cleaner than it found it.

PR #6148, opened 2026-04-27
feat(hardware): smart-room ESP32 demo with Telegram + simulator

Rhoahndur:
## Summary                                                                                                                                                                                                   
                  
Hackathon project — built an end-to-end demo: phone → Telegram bot → ZeroClaw daemon → MiniMax M2.7 → tool call → simulated ESP32 (`esp32_sim` example) → live browser SVG visualization.                    
                                          
This single branch bundles **5 upstream-shippable fixes** + a new example crate that makes hardware-free peripheral testing accessible to anyone. Happy to **split into separate PRs** if maintainers prefer one-concern-per-PR.                                                                                                                                                                                          
  
  ## What's in here                                                                                                                                                                                            
                  
  ### Fixes                                                                                                                                                                                                    
  1. **`src/main.rs`** — `register_peripheral_tools_fn` was only wired in the `Daemon` command handler. The `Agent` command got "Unknown tool: gpio_write" silently. 6-line fix: same registration block, 
  applied in the Agent handler.               
  2. **`channels/orchestrator/mod.rs`** — channel-driven sessions (Telegram, Discord, etc.) weren't loading peripheral tools either. New public helper `load_peripheral_tools()` in `runtime/agent/loop_` + a
  7-line wire-up at the orchestrator's tool-registry build site.
  3. **Slim `agent-runtime` build** — split 25 channel features into a `channels-all` aggregate. Default behaviour unchanged. New: `cargo build --no-default-features --features agent-runtime,hardware`       
  produces a ~30% smaller image with no chat-channel deps. Used by the demo Dockerfile.
  4. **`hardware/dev-sim` feature** — extends serial allow-list with `/tmp/zc-sim-*` paths so the simulator's pty pair is connectable without weakening the production allow-list. Off by default.             
  5. **`hardware_capabilities` tool** — was stripping the `pin_devices` map from the firmware's capabilities response before it reached the LLM. Surfaces it now.
                                                                                                                                                                                                               
  ### New                                 
  - **`crates/zeroclaw-hardware/examples/esp32_sim.rs`** — ~280-LOC tokio binary that speaks the same JSON-over-pty protocol as `firmware/esp32`. Spawns socat for the pty pair, runs an axum HTTP+WebSocket   
  server on :8080.                                                                                                                                                                                             
  - **`crates/zeroclaw-hardware/examples/esp32_sim_frontend.html`** — single-file SVG smart-room UI; live state pushes via WebSocket.                                                                          
[...]

PR #7045, opened 2026-05-31
feat(hardware): add smartroom named-device tools (set_device / read_device) + peripheral wiring support

Rhoahndur:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Added a public `load_peripheral_tools()` helper in `zeroclaw-runtime`.
  - Wired peripheral tool registration in the `zeroclaw agent` command, where it previously differed from daemon startup.
  - Wired peripheral tools into the channels orchestrator so Telegram, Discord, and other channel-driven sessions can use hardware tools when `[peripherals]` is configured.
  - Added `smartroom.rs` with `set_device` and `read_device` tools that let the model control devices by name, such as `reading_lamp` and `fan`, instead of guessing raw pin numbers.
  - Automatically registers the smartroom tools for boards named `esp32` or `esp32-sim`.
- **Scope boundary:** This PR delivers the peripheral tools wiring plus the first small piece of the adapted demo work: named-device smartroom tools. It does **not** include the full `esp32_sim` example, demo harness, `dev-sim` feature, or the `pin_devices` capabilities change from the original bundle.
- **Blast radius:** Small and isolated to configurations that use hardware peripherals.
- **Linked issue(s):** Extracted from #6148.
- **Labels:** `enhancement`, `size: S`, `risk: high`, `core`, `agent`, `channel`, `runtime`

```bash
cargo fmt --all -- --check
cargo clippy --all-targets --features "agent-runtime,hardware" -- -D warnings
cargo test --features "agent-runtime,hardware"
```

- New permissions, capabilities, or file system access scope? Yes. Agents and channel sessions with `[peripherals]` configured can now see hardware tools in the paths this PR wires.
- If any `Yes`, describe the risk and mitigation: The new capability is limited to explicit `[peripherals]` configuration and uses the existing peripheral tool factory. `load_peripheral_tools()` returns an empty tool set when the factory is absent or returns an error, so peripheral initialization failure does not crash agent or channel startup.

## Compatibility (required)

- Backward compatible? Yes.
- Config / env / CLI surface changed? No new config or CLI flags. Existing `[peripherals]` config now affects the `zeroclaw agent` command and channel-driven sessions as intended.
- If `No` or `Yes` to either: Existing users without `[peripherals]` configured are unaffected. Users with configured peripherals may now see hardware tools in agent and channel sessions; this is the intended behavior.

- **Feature flags or config toggles:** Remove or disable the `[peripherals]` section to avoid registering hardware tools.
- **Observable failure symptoms:** Agent or channel sessions unexpectedly expose or call peripheral tools; hardware sessions log peripheral tool loading or serial connection warnings around startup.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -7765,6 +7765,23 @@ pub async fn start_channels(
         );
         let mut built_tools = all_tools_result_ch.tools;
         let delegate_handle_ch = all_tools_result_ch.delegate_handle;
+
+        // Wire peripheral tools (gpio_read/gpio_write etc.) so channel-driven
+        // sessions (Telegram, Discord, etc.) can actuate hardware when
+        // [peripherals] is configured. Mirrors the agent loop wiring.
+        // The helper is safe to call unconditionally (returns empty when
+        // no peripherals are wired).
+        let peripheral_tools =
+            zeroclaw_runtime::agent::loop_::load_peripheral_tools(config.peripherals.clone()).await;
+        if !peripheral_tools.is_empty() {
+            ::zeroclaw_log::record!(
+                INFO,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_attrs(::serde_json::json!({"count": peripheral_tools.len()})),
+                "Peripheral tools added (channels orchestrator)"
+            );
+            built_tools.extend(peripheral_tools);
+        }
         let reaction_handle_ch = all_tools_result_ch.reaction_handle;
         let ask_user_handle_ch = all_tools_result_ch.ask_user_handle;
         let poll_handle_ch = all_tools_result_ch.poll_handle;
--- a/crates/zeroclaw-hardware/src/peripherals/mod.rs
+++ b/crates/zeroclaw-hardware/src/peripherals/mod.rs
@@ -17,6 +17,8 @@ pub mod capabilities_tool;
 #[cfg(feature = "hardware")]
 pub mod nucleo_flash;
 #[cfg(feature = "hardware")]
+pub mod smartroom;
+#[cfg(feature = "hardware")]
 pub mod uno_q_bridge;
 #[cfg(feature = "hardware")]
 pub mod uno_q_setup;
@@ -133,6 +135,23 @@ pub async fn create_peripheral_tools(config: &PeripheralsConfig) -> Result<Vec<B
                         &format!("Arduino upload tool added (port: {})", path)
                     );
                 }
+
+                // Smart-room named device tools (ESP32 / ESP32 simulator).
+                // Lets the model use device names instead of guessing pin numbers.
+                if board.board == "esp32" || board.board == "esp32-sim" {
+                    let transport = p.transport();
+                    tools.push(Box::new(smartroom::SetDeviceTool {
+                        transport: transport.clone(),
+                    }));
+                    tools.push(Box::new(smartroom::ReadDeviceTool { transport }));
+                    ::zeroclaw_log::record!(
+                        INFO,
+                        ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                            .with_attrs(::serde_json::json!({"board": board.board})),
+                        "Smart-room device tools added (set_device, read_device)"
+                    );
+                }
+
                 ::zeroclaw_log::record!(
                     INFO,
                     ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
--- a/crates/zeroclaw-hardware/src/peripherals/serial.rs
+++ b/crates/zeroclaw-hardware/src/peripherals/serial.rs
@@ -77,7 +77,7 @@ pub struct SerialTransport {
 const SERIAL_TIMEOUT_SECS: u64 = 5;
 
 impl SerialTransport {
-    async fn request(&self, cmd: &str, args: Value) -> anyhow::Result<ToolResult> {
+    pub(crate) async fn request(&self, cmd: &str, args: Value) -> anyhow::Result<ToolResult> {
         let mut port = self.port.lock().await;
         let resp = tokio::time::timeout(
             std::time::Duration::from_secs(SERIAL_TIMEOUT_SECS),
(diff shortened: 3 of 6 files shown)


WareWolf-MoonWall (approved):
Reviewed at e5fa2424c46799efdca8b526e002ba1934fead4e. CI 13/13 ✅. No prior reviews or inline threads. Extracted from #6148. Touches `crates/zeroclaw-runtime/src/agent/loop_.rs` (high-risk tier per AGENTS.md).

Three distinct but related fixes land together here: (1) `zeroclaw agent` was missing the peripheral tool registration that `zeroclaw daemon` had — this PR adds it; (2) channel-driven sessions (Telegram, Discord, etc.) couldn't use hardware tools even when `[peripherals]` was configured — this PR wires it; (3) the smartroom named-device tools provide the ESP32-specific abstraction that lets the LLM address devices by name.

### 🟢 What looks good — `load_peripheral_tools` is safe-by-default

```rust
pub async fn load_peripheral_tools(config: PeripheralsConfig) -> Vec<Box<dyn Tool>> {
    if let Some(f) = PERIPHERAL_TOOLS_FN.get() {
        f(config).await.unwrap_or_default()
    } else {
        Vec::new()
    }
}
```

The `unwrap_or_default()` means a peripheral initialization failure returns an empty tool set rather than propagating an error up through the agent or channel startup path. Hardware is optional; startup should never hard-fail because a board isn't connected.

### 🟢 What looks good — `zeroclaw agent` registration mirrors daemon exactly

The new `#[cfg(feature = "hardware")]` block in `src/main.rs` is structurally identical to the equivalent daemon-side registration — same closure pattern, same `create_peripheral_tools` call. This eliminates a silent behavioral difference that was causing `zeroclaw agent` to start without hardware tools even when a peripheral was configured.

### 🟢 What looks good — `SerialTransport::request` visibility is `pub(crate)` only

The method is exposed within the `zeroclaw-hardware` crate for `smartroom.rs` to use, but not leaked to external crates. `pub(crate)` is the minimum necessary visibility.

### 🟢 What looks good — orchestrator wiring is guarded against empty peripheral set

```rust
if !peripheral_tools.is_empty() {
    // ... log and extend ...
    built_tools.extend(peripheral_tools);
}
```

The log event only fires when tools are actually loaded. An operator with no peripherals configured won't see a confusing "0 peripheral tools added" message on every channel start.

### 🟡 Warning — hardcoded pin map in `smartroom.rs` will drift if firmware wiring changes

`output_pin()` hardcodes reading_lamp→12, overhead_light→13, heater→14, fan→2, and `input_pin()` hardcodes motion_sensor→5. The firmware knows the actual physical wiring. The companion PR #7047 surfaces `pin_devices` from the firmware capabilities response, which is exactly the dynamic mapping that would eliminate this table. As written, if the demo board is rewired, both the firmware and `smartroom.rs` must be updated in sync — that's a drift surface.

For v1 this is acceptable (the mapping is a known constant for the ESP32 demo board), but it's worth a comment in `output_pin()` noting that this table is intentionally hardcoded for the specific demo wiring, and referencing `hardware_capabilities` as the path toward dynamic resolution if the wiring changes.

### 🟡 Warning — missing PR template sections for a high-risk PR

The Validation Evidence is present. The Security, Compatibility, and Rollback sections are absent. Since this touches the runtime agent loop and the channel orchestrator startup path — both in the high-risk tier — those sections matter. For example: what happens if `load_peripheral_tools` panics (it can't under `unwrap_or_default`, but stating that explicitly helps reviewers and operators)? And the compatibility question: agents that previously ran without hardware tools now pick them up when `[peripherals]` is configured — is there any agent that shouldn't get hardware tools? These are answerable questions but they should be in the body.

### 🔵 Suggestion — `ToolKind::Plugin` is the current best fit but worth tracking

`SetDeviceTool` and `ReadDeviceTool` use `Role::Tool(ToolKind::Plugin)` for attribution. There's no `ToolKind::Hardware` variant in the current API. That's fine for now, but as the hardware ecosystem grows, a dedicated attribution kind would make hardware tool attribution distinguishable from user-installed plugins in logs and observability. Worth a follow-up issue.

### 🔵 Suggestion — PeripheralsConfig is snapshotted at channel startup

```rust
let peripheral_tools =
    zeroclaw_runtime::agent::loop_::load_peripheral_tools(config.peripherals.clone()).await;
```

The `config.peripherals.clone()` here creates a one-time snapshot. If the peripherals config changes after channels start (e.g., a new board is added via `zeroclaw config set` + reload), channel sessions will continue with the old tool set until the orchestrator restarts. This matches how other startup-time tools work, and it's fine for hardware (a new board connection requires reconnect anyway), but it's worth a comment so the behavior is explicit.

WareWolf-MoonWall:
@JordanTheJet — milestone alignment needed: feat(hardware) smartroom tools + runtime peripheral wiring, no scope match against open milestones. Hardware/peripheral scope — please advise on placement.

Audacity88 (approved):
I reviewed current head ff8f62c for #7045, including the PR body, current diff, prior @WareWolf-MoonWall approval, top-level comments, inline comments, related #6148 split context, related #7047 context, live labels, and current checks. I did not rerun local cargo; all visible GitHub checks are green.

@Rhoahndur, thanks for splitting this down from #6148. This is now the right-sized slice: it wires existing peripheral tool registration into the missing agent/channel paths and adds the named smartroom tools without pulling in the full simulator/demo bundle.

### ✅ Resolved — hardcoded demo pin map is now explicit

The current head adds the missing comment above the smartroom pin mapping. That makes the important boundary visible: this table is intentionally tied to the hackathon ESP32 demo wiring, and dynamic named-device resolution should come from the `hardware_capabilities` / `pin_devices` path when that data is available.

### 🟢 What looks good — hardware remains opt-in and startup-safe

The new runtime helper keeps peripheral loading behind the registered factory and returns an empty tool set when no factory is registered or the factory reports an error. The `zeroclaw agent` command now mirrors daemon registration, and the channel orchestrator only extends the registry when peripheral tools actually load. That keeps no-peripheral configs quiet while fixing the real missing-tool path for configured hardware users.

### 🟢 What looks good — the smartroom tool surface avoids raw-pin guessing

`set_device` and `read_device` give the model a constrained named-device API instead of asking it to pick GPIO pins. The schemas are narrow, the serial transport visibility increase stays `pub(crate)`, and the tool attribution uses the current best-fit `ToolKind::Plugin` until we have a dedicated hardware attribution kind.

I’m approving this on the current diff and green CI.

PR #7046, opened 2026-05-31
feat(hardware): add dev-sim feature with /tmp/zc-sim-* serial allowlist

Rhoahndur:
**Extracted from #6148** (hackathon ESP32 demo bundle, split per AGENTS.md one-concern rule as requested in the review by @theonlyhennygod and @Audacity88).

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Added a new opt-in `dev-sim` Cargo feature.
  - When enabled, the serial peripheral allow-list is extended with `/tmp/zc-sim-*` paths for the ESP32 simulator and similar hardware-free testing.
  - Updated the serial-path rejection message to surface the simulator paths as a hint when the feature is active.
- **Scope boundary:** This PR only adds the `dev-sim` feature and allow-list extension. It does **not** include the `esp32_sim` example, the demo harness, smartroom tools, or any other pieces from the original bundle.
- **Blast radius:** Small and defensive. The feature is off-by-default and only affects builds compiled with `dev-sim`.
- **Linked issue(s):** Extracted from #6148
- **Labels:** `enhancement`, `size: XS`, `risk: medium`, `dependencies`

## Validation Evidence (required)

```bash
# 1. Formatting
cargo fmt --all -- --check
Result: ✅ Passed (clean, exit code 0)

# 2. Clippy (strict)
cargo clippy --all-targets --features "agent-runtime,hardware,dev-sim" -- -D warnings
Result: ✅ Passed (no warnings or errors)

# 3. Tests with the feature
cargo test --features "agent-runtime,hardware,dev-sim" --test '*'
Result: ✅ Passed (relevant tests green)

# 4. Slim build check
cargo check --no-default-features --features "agent-runtime,hardware,dev-sim"
Result: ✅ Passed (compiles cleanly)
```

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? `Yes` — builds compiled with `dev-sim` allow serial paths under `/tmp/zc-sim-*` so hardware-free simulator ptys can be used.
- New external network calls? `No`
- Secrets / tokens / credentials handling changed? `No`
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No`
- Risk and mitigation: the new paths are compile-time gated behind the opt-in `dev-sim` feature. Production builds without the feature keep the existing serial allow-list.

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `Yes` — this adds a new opt-in Cargo feature.
[...]

PR #7047, opened 2026-05-31
fix(hardware): surface pin_devices and description in hardware_capabilities tool

Rhoahndur:
**Extracted from #6148** (hackathon ESP32 demo bundle, will split per AGENTS.md one-concern rule).

## Summary

- **Base branch:** `master`
- **What changed and why:** The `hardware_capabilities` tool was silently dropping the `pin_devices` map and `description` from the firmware response. These fields are important for the LLM to reason about named devices (e.g. `reading_lamp`) instead of guessing raw pin numbers.
- **Scope boundary:** Only updates output formatting in `capabilities_tool.rs`. No behavior changes for boards that don't report `pin_devices`.
- **Blast radius:** Very small and isolated to the capabilities tool.
- **Linked issue(s):** Extracted from #6148
- **Labels:** `bug`, `size: XS`, `risk: medium`, `hardware`

## Validation Evidence (required)

Local validation is the signal CI cannot replace. Run the full battery and paste literal output (tails, failures, warnings — not "all passed").

```bash
cargo fmt --all -- --check
cargo clippy --all-targets --features "agent-runtime,hardware" -- -D warnings
cargo test --features "agent-runtime,hardware" --test '*'
```

- **Commands run and tail output:**
  - `cargo fmt --all -- --check` → ✅ Passed (clean)
  - `cargo clippy --all-targets --features "agent-runtime,hardware" -- -D warnings` → ✅ Passed (no warnings)
  - `cargo test --features "agent-runtime,hardware" --test '*'` → ✅ Passed

- **Beyond CI — what did you manually verify?** Confirmed that when firmware reports `pin_devices` and `description`, they now appear in the tool output (pretty-printed JSON for the map). Boards without those fields are unaffected.
- **If any command was intentionally skipped, why:** N/A — used targeted features matching the change; full default-feature matrix is covered by the broader CI runs.

## Security & Privacy Impact (required)

Yes/No for each. Answer any `Yes` with a 1–2 sentence explanation.

- New permissions, capabilities, or file system access scope? (`No`)
- New external network calls? (`No`)
- Secrets / tokens / credentials handling changed? (`No`)
- PII, real identities, or personal data in diff, tests, fixtures, or docs? (`No`)
- If any `Yes`, describe the risk and mitigation: N/A

## Compatibility (required)

- Backward compatible? (`Yes`)
- Config / env / CLI surface changed? (`No`)
[...]

PR #7048, opened 2026-05-31
feat(hardware): add esp32_sim example (simulator binary + web frontend)

Rhoahndur:
**Extracted from #6148** (hackathon ESP32 demo bundle, will split per AGENTS.md one-concern rule as requested in the review by @theonlyhennygod and @Audacity88). Rebased after #7046 landed (dev-sim feature deduped).

## Summary

- **Base branch:** `master`
- **What changed and why:** Added the `esp32_sim` example (tokio + axum binary + single-file SVG/JS frontend). It implements the same JSON-over-serial protocol as real ESP32 firmware, spawns socat for a `/tmp/zc-sim-*` pty pair (enabled by the `dev-sim` feature), and serves a live-updating WebSocket visualizer. 

  Later commits (addressing review): made HTTP bind address overridable via optional first CLI arg (default `0.0.0.0:8080` for Docker; safe `127.0.0.1:...` for laptops); switched from raw `tokio::spawn` to `zeroclaw_spawn::spawn!` macro to satisfy `clippy -D warnings` (the disallowed-method lint surfaced after rebase onto current master); added `zeroclaw-spawn` to dev-dependencies for the example.

- **Scope boundary:** This PR only adds the simulator example + its frontend under `crates/zeroclaw-hardware/examples/`. It does **not** include the full `demo/` harness, Docker setup, additional docs, or any production peripheral/smartroom changes (those landed in #7045/#7046).
- **Blast radius:** Very small — new example binary only, gated behind `hardware` + `dev-sim` features. No effect on default builds, the main `zeroclaw` binary, or any runtime behavior.
- **Linked issue(s):** Extracted from #6148
- **Labels:** `enhancement`, `size: L`, `risk: medium`, `dependencies`, `hardware`

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-hardware --features "hardware,dev-sim" --example esp32_sim -- -D warnings
cargo check -p zeroclaw-hardware --features "hardware,dev-sim" --example esp32_sim
```

```
$ cargo fmt --all -- --check
$ echo $?
0

$ cargo clippy -p zeroclaw-hardware --features "hardware,dev-sim" --example esp32_sim -- -D warnings
    Checking ...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.72s
$ echo $?
0

$ cargo check -p zeroclaw-hardware --features "hardware,dev-sim" --example esp32_sim
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3.18s
$ echo $?
0

$ cargo clippy -p zeroclaw-hardware --all-targets --features "hardware,dev-sim" -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.06s

$ cargo test -p zeroclaw-hardware --features "hardware,dev-sim"
test result: ok. 136 passed; 0 failed; ...
```
- **Beyond CI — what did you manually verify?** Ran the binary with explicit bind override to confirm the CLI path works and server starts on the requested address (no "address in use" or bind failures). The pty symlinks are created, the HTTP frontend serves, and WS updates reflect pin flips from the protocol. Also exercised the motion-sensor convenience (pin 5 always 1) path.
- **If any command was intentionally skipped, why:** Full workspace `cargo clippy --all-targets -- -D warnings` and `cargo test` (without package filter) are slow and disk-heavy on the dev laptop (storage-constrained MacBook Air); focused validation on the changed crate + example (which exercises the new code paths) plus the required fmt/clippy for the example. CI will run the complete matrix.

- New external network calls? (`Yes`) — The example runs a local HTTP + WebSocket server (axum) to host the SVG visualizer and push live GPIO state. This is a demo/dev tool only; no calls to external services.
- If any `Yes`, describe the risk and mitigation: The server binds to `0.0.0.0:8080` by default (for `docker-compose` port mapping convenience in the original demo harness). When run directly on a dev machine this would expose the unauthenticated visualizer + manual controls on all interfaces. The HTTP surface consists of: GET / (demo page), GET /state (current snapshot), POST /manual (JSON {pin, value} — only mutates the in-memory simulated GPIO state for the demo visualizer/manual buttons; never real hardware and never injected into the pty/"firmware" command stream), and WebSocket /ws (broadcast of state changes). The actual command input to the simulated firmware always arrives over the local pty pair (from a ZeroClaw agent using the dev-sim serial path). Mitigated by: (1) explicit first CLI arg for bind address so `127.0.0.1:8080` is the easy/safe default for laptop use, (2) fully documented as demo-only, (3) the `dev-sim` + `hardware` features are off-by-default, (4) no path from the network endpoints to real peripherals or to the pty command channel.

## Compatibility (required)

- Config / env / CLI surface changed? (`Yes`) — only for the example binary: an optional first positional argument now overrides the HTTP bind address (e.g. `cargo run --example esp32_sim --features "hardware,dev-sim" -- 127.0.0.1:8080`).
- If `No` or `Yes` to either: exact upgrade steps for existing users: N/A for new example. Existing users of the (previously unreleased) example who relied on the hardcoded const can continue to omit the arg; default behavior is unchanged.

- **Fast rollback command/path:** `git revert <merge-sha>` (or manually: delete `crates/zeroclaw-hardware/examples/esp32_sim.rs` and `esp32_sim_frontend.html`, remove the `[[example]]` stanza and the `zeroclaw-spawn` dev-dep line from `crates/zeroclaw-hardware/Cargo.toml`).
- **Feature flags or config toggles:** Entirely behind the `dev-sim` + `hardware` features (both off by default). The example is never built for normal `cargo build` or releases.
- **Observable failure symptoms:** N/A — if the example is broken it simply fails to `cargo run --example` under the features; the main daemon, serial layer, and all production peripherals are unaffected.

(none)


WareWolf-MoonWall (approved):
Reviewed at 9265cc1031ffebc0afe81e2fb87324e800c9463f. CI 13/13 ✅. No prior reviews or inline threads. Extracted from #6148.

The simulator binary is a well-designed dev tool: a tokio + axum process that speaks the same JSON-over-serial protocol as real ESP32 firmware, creates a socat pty pair at `/tmp/zc-sim-*`, and serves a live WebSocket visualization of virtual GPIO state. The SVG frontend + state broadcast architecture is appropriate for a hardware-free demo harness.

### 🟡 Warning — workspace `Cargo.toml` adds `dev-sim` feature that #7046 also adds; one will conflict at merge

Both this PR and #7046 add this line to the workspace `Cargo.toml`:

```toml
dev-sim = ["dep:zeroclaw-hardware", "zeroclaw-hardware/dev-sim"]
```

Whichever PR merges second will hit a merge conflict on that line. The fix is straightforward: once #7046 lands, rebase this branch and drop the duplicate workspace feature entry (since #7046 already owns it). Please add a `Depends on #7046` note to the PR body to make the ordering explicit.

Similarly, `crates/zeroclaw-hardware/Cargo.toml` — this PR adds `dev-sim = []` to the feature table and the `[dev-dependencies]` + `[[example]]` entries; #7046 adds `dev-sim = []` too. The `dev-sim = []` line will conflict; the `[dev-dependencies]` and `[[example]]` entries are unique to this PR and won't.

### 🟡 Warning — missing PR template sections

Security, Compatibility, and Rollback sections are absent. For this PR: security is relevant because the simulator binds `HTTP_BIND = "0.0.0.0:8080"` — intentional for Docker port mapping per the comment, but an operator running the example on a networked machine exposes the dashboard on all interfaces. The Rollback section should note this is an example binary that can be disabled by removing the `[[example]]` entry or reverting the PR — no production code path is affected.

### 🟢 What looks good — pty pair architecture is correct

Spawning `socat` to create two named symlink ends (`/tmp/zc-sim-esp32` for ZeroClaw, `/tmp/zc-sim-firmware` for the simulator) is the standard approach for pty-based hardware simulation. ZeroClaw's serial layer sees the same `/tmp/zc-sim-*` path that the `dev-sim` allowlist permits; the simulator sees its end transparently. The symlink cleanup at startup (removing stale files from prior runs) prevents "address already in use" style failures.

### 🟢 What looks good — virtual motion sensor pin returns 1 for lively demos

```rust
async fn read_pin(&self, pin: u8) -> u8 {
    if pin == 5 {
        return 1;
    }
    // ...
}
```

Hardcoding pin 5 (motion sensor) to always return "motion detected" keeps demo conversations interesting without requiring physical movement. This is clearly a demo convenience, not a bug.

### 🟢 What looks good — WebSocket broadcast uses `let _ =` for send errors

`let _ = self.tx.send(self.snapshot().await)` — dropping the send result is correct here. If there are no WebSocket subscribers, the broadcast channel returns an error that should be ignored. The actual GPIO state is still updated in `self.pins` regardless of whether anyone is watching.

### 🔵 Suggestion — `HTTP_BIND` could be a CLI argument for security-conscious use

`"0.0.0.0:8080"` is right for the Docker demo workflow but exposes the dashboard on all interfaces when run outside Docker. A simple `--bind` CLI flag (defaulting to `"127.0.0.1:8080"` with `"0.0.0.0:8080"` as opt-in) would let developers run the simulator safely on laptops without needing to remember to firewall port 8080.

WareWolf-MoonWall:
@JordanTheJet — milestone alignment needed: feat(hardware) esp32_sim example, no scope match against open milestones. Hardware/peripheral scope — please advise on placement. Note: this PR depends on #7046 and will need a rebase after #7046 lands to resolve the duplicate dev-sim workspace feature entry.

Audacity88 (changes requested):
Reviewed current head 6d42ff59995e5c480f87572343637e8f578e4bd0, the prior approval at 9265cc1031ffebc0afe81e2fb87324e800c9463f, the current PR body/check state, and the changed files. I also ran the PR's focused example check locally against the current head. This PR already had a helpful approval review on the earlier simulator shape, but the later CLI bind-address update introduced a compile blocker in the example itself.

### 🔴 Blocking — pass the configured bind address into `run_http_server`

The new bind-address argument is parsed in `main()` as `http_bind`, but the spawned server still calls `run_http_server(http_state).await`, and `run_http_server()` still only accepts `state: AppState`. Inside that function, the listener now tries to bind `http_bind`, which is out of scope:

```rust
async fn run_http_server(state: AppState) -> Result<()> {
    // ...
    let listener = tokio::net::TcpListener::bind(http_bind).await?;
```

That means the PR's own validation command fails now:

```bash
cargo check -p zeroclaw-hardware --features "hardware,dev-sim" --example esp32_sim
```

The local result was:

```text
error[E0425]: cannot find value `http_bind` in this scope
   --> crates/zeroclaw-hardware/examples/esp32_sim.rs:329:50
```

The fix should be small: pass the parsed bind address into the task, for example by changing the call to `run_http_server(http_state, http_bind)` and the function signature to accept `&str`/`String` or a parsed socket address with the right lifetime for the spawned task.

### 🟡 Warning — close the validation code fence in the PR body

The PR body now includes the Security, Compatibility, and Rollback sections, which is good, but the validation block starts with ```bash and is not closed before `## Security`. GitHub will render those sections as part of the code block, which makes the public evidence hard to read and can hide the required template sections.

### 🟢 What looks good — the requested bind-address concern is directionally addressed

The change to make the HTTP bind address configurable is the right response to the earlier review suggestion. Once the compile issue is fixed and the PR body fence is closed, the remaining simulator scope still looks cleanly isolated to the hardware example/demo path.

Rhoahndur:
@Audacity88 — addressed the 🔴 blocking item:

- The bind address parsed from CLI (`let bind_addr = ...`) is now correctly passed into `run_http_server` (via a dedicated `bind_for_http` clone + `zeroclaw_spawn::spawn!(async move { ... })` form).
- Also fixed the post-rebase clippy failure (raw `tokio::spawn` is disallowed; switched both spawns to the attributed `zeroclaw_spawn::spawn!` macro and added the workspace dev-dep).
- Rebased onto current `upstream/master` (post #7046 merge) so the `dev-sim` feature declarations are no longer duplicated — only the example + dev-dep bits remain in the diff.
- Updated the PR body: closed the validation fence, added fresh literal output from `cargo fmt` / `clippy ... --example` / `check ... --example` / package tests (136 passed), aligned sections to the current template, expanded the Security/Compat/Rollback with the Yes/No details.

All focused validation commands now exit 0 locally with no warnings. Pushed the fix commit; CI should be happy on the new head. 

(Also updated the "depends on #7046" note since it has landed and we rebased.)

Ready for re-review when you have a moment. Thanks!

Rhoahndur:
Polite ping @Audacity88 @JordanTheJet @theonlyhennygod —

@Audacity88: We addressed the 🔴 blocking item you called out (CLI bind address now correctly threaded into `run_http_server` + switched to the attributed `zeroclaw_spawn::spawn!` macro to satisfy clippy -D warnings). Also rebased cleanly after #7046 (dev-sim deduped) and updated the PR body with fresh validation evidence + proper fence. The focused example checks now pass locally with no warnings. Latest head has the fixes.

@JordanTheJet / @theonlyhennygod: Still have the open review requests, plus the earlier milestone alignment comment from WareWolf-MoonWall. Any advice on where hardware example work like this should land?

These two PRs (#7047 + this one) are the last blockers before we can open the grand finale demo-harness PR. The stacked end-to-end (sim + agent + visualizer + the exact "It's getting dark and chilly..." vignette that exercises `read_device` + `set_device(reading_lamp, "on")` etc. with the browser updating) is ready locally with the low-storage host scripts. Extracted from #6148 per the original split request.

Appreciate any time you can give — excited to get the full story landed!

Audacity88 (changes requested):
Reviewed current head fbfd84e677eecab8576d341c6a0e471a51fe617f, the prior reviews from @WareWolf-MoonWall and me, the author’s latest response, the current PR body/check state, and the changed files. The original compile blocker is fixed, but I found one security/body mismatch that still needs to be corrected before I can clear my review block.

### ✅ Resolved — bind address is now passed into the HTTP server task

The previous blocker is addressed. `main()` now clones the parsed bind address into `bind_for_http`, moves it into the spawned HTTP task, and passes it to `run_http_server(http_state, bind_for_http)`. `run_http_server()` now accepts that `String` and binds `&bind_addr`, so the old out-of-scope `http_bind` compile failure is gone.

### ✅ Resolved — the required PR-body sections now render normally

The validation fence is closed, and the Security, Compatibility, and Rollback sections are visible as normal Markdown now.

### 🔴 Blocking — update the Security section to match the `/manual` HTTP write endpoint

The Security section says the simulator “never accepts commands over the network — it only emits snapshots and receives pin commands over the local pty from ZeroClaw.” That is not true at the current head. The axum server registers `POST /manual`, and `manual_flip()` mutates simulator pin state from the JSON request body:

```rust
.route("/manual", post(manual_flip))
```

The frontend calls that route for the manual override buttons:

```js
await fetch('/manual', {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({ pin, value }),
});
```

I do not think the manual endpoint is inherently unacceptable for this dev-only simulator. It only changes in-memory simulated GPIO state, and the PR already documents that the example is behind `hardware` + `dev-sim`. But the public security analysis needs to say what the network surface actually is, especially because the default bind remains `0.0.0.0:8080`.

Please either remove/loopback-constrain the network write path, or update the Security section to explicitly say that the HTTP server exposes an unauthenticated `/manual` endpoint for flipping simulated pins, that it cannot control real hardware, and that direct laptop runs should use the `127.0.0.1:...` bind override unless intentionally exposing the demo.

### 🟢 What looks good — the simulator remains isolated to the example/dev-sim surface

The rebase dropped the duplicate workspace `dev-sim` feature conflict from #7046, and the remaining diff is limited to the hardware example, the single-file frontend, the example registration, dev-dependencies, and the lockfile entry. The WebSocket path itself is still state-broadcast-only; the only remaining issue is making the separate HTTP manual-control route visible in the security record.

Rhoahndur:
@Audacity88 — thank you for the re-review and the detailed breakdown!

✅ The two prior items you marked resolved (bind address threading + spawn! + the PR body sections now rendering cleanly) are great to have confirmed.

🔴 The security description accuracy issue is now addressed:

* Updated the "If any `Yes`" mitigation paragraph in the **Security & Privacy Impact** section to explicitly enumerate the actual HTTP surface (`GET /`, `GET /state`, `POST /manual`, `WS /ws`), what `/manual` does (JSON-driven in-memory simulated pin flips for the demo page only — never real hardware, never forwarded to the pty/"firmware" command path), and the isolation from the pty side.
* Added source comments right at the `.route("/manual"...)` registration and inside `manual_flip()` explaining the demo-only, in-memory nature.
* New commit on the branch: 00439e285 (plus the body edit is live).

The default bind remains `0.0.0.0:8080` only for the containerized demo workflow (as documented); the first CLI arg continues to make safe localhost binding trivial for direct runs.

All other CI remains green. This (together with #7047) unblocks opening the final stacked demo-harness PR with the end-to-end vignette evidence.

Ready for another look whenever convenient. Thanks again!

Audacity88 (approved):
Reviewed current head 00439e285f73652d038d278c54c885e634eef7f4, the prior reviews from @WareWolf-MoonWall and me, the author response, the current PR body, CI state, and the changed files. This clears my previous block.

### ✅ Resolved — `/manual` is now described as part of the HTTP surface

The prior request-changes review blocked because the Security section said the simulator never accepted commands over the network, while the example registered unauthenticated `POST /manual` and used it from the browser controls.

The current PR body now lists the actual HTTP surface: `GET /`, `GET /state`, `POST /manual`, and WebSocket `/ws`. It also describes `/manual` as a demo-only JSON endpoint that mutates in-memory simulated GPIO state, does not touch real hardware, and is not forwarded into the pty/firmware command path. That matches the current implementation.

### ✅ Resolved — the code now marks the same boundary at the route and handler

The new comments beside `.route("/manual", post(manual_flip))` and inside `manual_flip()` make the demo-only boundary visible in the source as well as in the PR body. That is the right place to leave the warning, because the default bind is still `0.0.0.0:8080` for the containerized demo path.

### 🟢 What looks good — the remaining diff stays isolated to the simulator example

The diff is still limited to the `esp32_sim` example, its frontend, the example registration/dev-dependencies, and the lockfile update. The simulator remains behind `hardware` + `dev-sim`, the pty command path stays separate from the browser manual-control path, and the current CI run is green.

Milestone placement still looks like the separate routing question already raised in the thread, not a reason for me to keep this review blocked.

PR #7049, opened 2026-05-31
fix(providers): omit temperature for kimi-k2 models in compatible.rs (#7022)

### LINKED ISSUE #7022: [Bug]: kimi-k2.6 fails with 400 invalid temperature because compatible.rs always sends baseline 0.7
### Affected component

provider

### Severity

S1 - workflow blocked

### Current behavior

`OpenAiCompatibleModelProvider` (used by Moonshot/Kimi via `build_kimi_code_compat`) always serializes `temperature: f64` into the request body. When no `temperature` is configured in `config.toml`, it falls back to `BASELINE_TEMPERATURE = 0.7`.

Moonshot's `kimi-k2.6` (and `kimi-k2.5`) enforce a **fixed temperature** per mode:
- Thinking mode → `temperature = 1.0`
- Instant mode → `temperature = 0.6`

Any other value results in a non-retryable HTTP 400:

```
400 invalid temperature: only 1 is allowed for this model
```

This means `kimi-k2.6` is **unusable out-of-the-box** with ZeroClaw unless the user manually sets `temperature = 1` in their config.

### Expected behavior

ZeroClaw should detect model-level temperature constraints (like it already does for Anthropic `claude-opus-4-7` via `anthropic_model_omits_temperature`) and either:
1. Force the correct fixed temperature for `kimi-k2.*` models, or
2. Skip sending `temperature` entirely and let the backend choose the model-appropriate default.

### Steps to reproduce

```bash
# 1. Configure Moonshot kimi-k2.6 without explicit temperature
# ~/.zeroclaw/config.toml
[providers.models.moonshot.default]
model = "kimi-k2.6"
api_key = "sk-..."

# 2. Start daemon and trigger any chat/agent turn
zeroclaw daemon
zeroclaw agent -a bmad_orchestrator -m "hello"

# 3. Observe 400 error in logs
```

### Impact

- **Affected users:** Anyone using Moonshot `kimi-k2.5` / `kimi-k2.6` with default settings.
- **Frequency:** Always.
- **Consequence:** Complete workflow blockage until user discovers the workaround.

### Logs / stack traces

```
400 Bad Request: invalid temperature: only 1 is allowed for this model
```

### ZeroClaw version

v0.8.0-beta-1+85 (f6015ab4b)

### Rust version

rustc 1.94.1

### Operating system

Windows 10 / MINGW64

### Regression?

No, first-time setup

### Pre-flight checks

Yyukan:
## Summary

- **Base branch:** `master`
- **What changed and why:** `OpenAiCompatibleModelProvider` always sends `temperature: 0.7` (from `BASELINE_TEMPERATURE`), which breaks Moonshot `kimi-k2.5` and `kimi-k2.6` — they enforce fixed temperatures per mode (1.0 thinking, 0.6 instant) and reject any other value with HTTP 400. Fix follows the `anthropic_model_omits_temperature` / `bedrock_model_omits_temperature` precedent: add `compatible_model_omits_temperature(model)` helper, change both request structs to `Option<f64>` with `skip_serializing_if`, update 7 call sites to omit only when user did not configure temperature AND model requires omission; explicit user config passes through.
- **Scope boundary:** Only `compatible.rs` temperature handling for `kimi-k2.*` models. No changes to config schema, other providers, or top_p/penalty handling.
- **Blast radius:** All `OpenAiCompatibleModelProvider` callers (Kimi, OpenRouter-compatible, etc.). Non-kimi models unaffected — same behavior as before.
- **Linked issue(s):** Fixes #7022

```bash
$ cargo fmt --all -- --check
ok (all files formatted correctly)

ok (no issues found)

$ cargo test --package zeroclaw-providers
ok 890 passed, 1 ignored

$ cargo test --package zeroclaw-providers compatible_model_omits_temperature
ok 2 passed
```

- **Beyond CI:** Manually verified diff — 7 call sites updated identically, 2 internal fallback calls cleaned up, 7 test struct literals updated, 2 new tests added matching anthropic/bedrock pattern.
- **Skipped commands:** None.

## Compatibility (required)

- Backward compatible? Yes — only omits temperature for models that currently fail with 400. Non-kimi models unchanged.

- **Fast rollback command/path:** `git revert <sha>` — single-file change in `compatible.rs`, no config migration.
- **Feature flags or config toggles:** None — behavioral change only.
- **Observable failure symptoms:** Kimi K2.5/K2.6 users will see 400 `invalid temperature` errors again; other providers unaffected.

## Files Changed

| File | What | Lines |
|------|------|-------|
| `crates/zeroclaw-providers/src/compatible.rs` | Helper function + 2 struct fields + 7 call sites + 2 internal calls + 7 test literals + 2 new tests | +75 / -22 |

### DIFF
--- a/crates/zeroclaw-providers/src/compatible.rs
+++ b/crates/zeroclaw-providers/src/compatible.rs
@@ -633,11 +633,20 @@ impl OpenAiCompatibleModelProvider {
     }
 }
 
+/// Kimi K2.5/K2.6 models enforce fixed temperatures per mode (1.0 thinking,
+/// 0.6 instant) and reject any other value with HTTP 400. Omit `temperature`
+/// for kimi-k2.* models so the backend chooses the correct mode default.
+/// Substring match covers k2.5, k2.6, and future k2.x variants.
+fn compatible_model_omits_temperature(model: &str) -> bool {
+    model.contains("kimi-k2")
+}
+
 #[derive(Debug, Serialize)]
 struct ApiChatRequest {
     model: String,
     messages: Vec<Message>,
-    temperature: f64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    temperature: Option<f64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     stream: Option<bool>,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -936,7 +945,8 @@ struct Function {
 struct NativeChatRequest {
     model: String,
     messages: Vec<NativeMessage>,
-    temperature: f64,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    temperature: Option<f64>,
     #[serde(skip_serializing_if = "Option::is_none")]
     stream: Option<bool>,
     /// Mirrors `ApiChatRequest::stream_options`. Without this, tool-enabled
@@ -1690,7 +1700,7 @@ impl OpenAiCompatibleModelProvider {
         effective_messages: &[ChatMessage],
         tools: Option<Vec<serde_json::Value>>,
         model: &str,
-        temperature: f64,
+        temperature: Option<f64>,
         allow_user_image_parts: bool,
     ) -> NativeChatRequest {
         let has_tool_entries = tools.as_ref().is_some_and(|tools| !tools.is_empty());
@@ -2184,7 +2194,11 @@ impl ModelProvider for OpenAiCompatibleModelProvider {
         model: &str,
         temperature: Option<f64>,
     ) -> anyhow::Result<String> {
-        let temperature = temperature.unwrap_or(self.default_temperature());
+        let temperature = if temperature.is_none() && compatible_model_omits_temperature(model) {
+            None
+        } else {
+            Some(temperature.unwrap_or(self.default_temperature()))
+        };
         let credential = self.credential.as_deref();
 
         // Normalize image markers (e.g. local file paths from channel
@@ -2297,7 +2311,11 @@ impl ModelProvider for OpenAiCompatibleModelProvider {
         model: &str,
         temperature: Option<f64>,
     ) -> anyhow::Result<String> {
-        let temperature = temperature.unwrap_or(self.default_temperature());
+        let temperature = if temperature.is_none() && compatible_model_omits_temperature(model) {
+            None
+        } else {
+            Some(temperature.unwrap_or(self.default_temperature()))
+        };
         let credential = self.credential.as_deref();
 
         let normalized = Self::normalize_messages_for_upstream(messages).await?;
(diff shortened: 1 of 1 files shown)


WareWolf-MoonWall (approved):
Reviewed at a151e45259406413d902da181f30afd2e87455ba. CI 13/13 ✅. No prior reviews or inline threads. Fixes #7022.

This is a clean, well-scoped fix following the established `anthropic_model_omits_temperature` / `bedrock_model_omits_temperature` precedent in the providers crate. Kimi K2.5 and K2.6 enforce fixed temperatures per mode (1.0 thinking, 0.6 instant) and reject any other value with HTTP 400; the right fix is to omit the field rather than guess a valid value.

### 🟢 What looks good — logic respects explicit user configuration

```rust
let temperature = if temperature.is_none() && compatible_model_omits_temperature(model) {
    None
} else {
    Some(temperature.unwrap_or(self.default_temperature()))
};
```

The gate fires only when the caller passed `None` (no user preference) **and** the model requires omission. If a user explicitly configures a temperature — even 0.6 or 1.0 — that value passes through unchanged. The omission is purely a default-path behavior correction, which is the right contract.

### 🟢 What looks good — all 7 call sites updated identically, two fallback calls cleaned

Every site that previously did `temperature.unwrap_or(self.default_temperature())` (producing a bare `f64`) now goes through the same conditional. The two internal `chat_with_history` fallback calls that previously wrapped an already-unwrapped `f64` back into `Some(temperature)` are simplified to pass `temperature` directly — no behavior change, but cleaner.

### 🟢 What looks good — `ApiChatRequest` and `NativeChatRequest` updated consistently

Both request structs now carry `temperature: Option<f64>` with `#[serde(skip_serializing_if = "Option::is_none")]`. The field is omitted from the serialized JSON when `None`, which is exactly what Kimi's API requires and what the test vector confirms.

### 🟢 What looks good — two targeted tests cover positive and negative cases

`compatible_model_omits_temperature_matches_kimi_k2` tests kimi-k2, kimi-k2.5, kimi-k2.6, and kimi-k2.7 (demonstrating forward-coverage for the `k2.x` family). `compatible_model_omits_temperature_skips_other_models` tests kimi-k1, kimi-latest, gpt-4o, claude-sonnet-4-6, and llama-3.1-70b. The negative cases are the important ones — they confirm the helper does not accidentally widen scope.

### 🔵 Suggestion — `contains("kimi-k2")` would also match a hypothetical `kimi-k20` or `kimi-k2-turbo`

The substring match is intentional per the doc comment ("future k2.x variants") and is the same approach used for anthropic (`model.starts_with(...)`) and bedrock. Moonshot's kimi-k2 family does consistently enforce fixed temperatures, so matching any `kimi-k2*` model is a reasonable forward assumption. Worth keeping in mind if Moonshot ever releases a `kimi-k2`-prefixed model that accepts standard temperature — it would be silently omitted. Low risk for now.

PR #7061, opened 2026-06-01
fix(providers): retry empty completions instead of returning a blank turn

mn13:
## Problem

A 2xx provider response carrying **no text, no tool calls, and no reasoning** is treated by `ReliableModelProvider` as a valid final answer and returned as-is. Downstream (e.g. the agent loop's "no tool calls → final response" branch) then surfaces that blank turn to the caller.

Concrete cases where this is user-visible:
- **Non-streaming chat** (e.g. the webhook channel, which posts a single HTTP reply) → an empty reply, indistinguishable from a dropped request.
- **Subagent delegation** (`crates/zeroclaw-runtime/src/tools/delegate.rs`, via `chat_with_system`) → an empty delegation result.

Empty completions are almost always transient — a truncated stream, a content-filter soft block, or a 0-token sample. The provider layer already classifies transient failures as retryable; an empty 2xx simply wasn't one of them.

## Fix

Treat an empty completion like a retryable error and re-roll it, reusing the existing retry machinery:

- Bounded by `max_retries`, with the same exponential backoff progression.
- On the final attempt the empty response **falls through and is returned unchanged** — so behaviour is never worse than before, only better when a re-roll succeeds.
- Applied to **all four** `ReliableModelProvider` entry points:
  - `chat()` and `chat_with_tools()` → `ChatResponse`, checked via `is_empty_completion()` (empty text **and** no tool calls **and** no reasoning; prompt-guided tool calls live in `text`, so they're never misclassified).
  - `chat_with_history()` and `chat_with_system()` → `String`, checked via `resp.trim().is_empty()`.
- The shared `push_failure` + warn + sleep + backoff tail is factored into a private `backoff_after_empty_completion` helper, so the four call sites only carry the (per-return-type) emptiness check.

**Out of scope:** live token-streaming responses are assembled outside these methods, so an empty *stream* isn't covered here — that would need a separate guard at the point where streamed deltas are collected.

## Testing

- `cargo test -p zeroclaw-providers` — all green, including 6 new tests:
  - retry-then-recover on `chat`, `chat_with_tools`, `chat_with_history`, and `chat_with_system`
  - persistent-empty exhausts retries and returns a blank response **without erroring**
  - a non-empty response triggers **no** extra calls
- `cargo clippy -p zeroclaw-providers --all-targets` — clean
- `cargo fmt -p zeroclaw-providers -- --check` — clean

## Risk

Low; no public signatures change. Worst case for a persistently-empty provider is `max_retries` extra re-rolls before returning the same blank result that was returned before this change.

## Rollback

Revert this commit — the change is contained to `crates/zeroclaw-providers/src/reliable.rs`.

### DIFF
--- a/crates/zeroclaw-providers/src/reliable.rs
+++ b/crates/zeroclaw-providers/src/reliable.rs
@@ -354,6 +354,24 @@ fn push_failure(
     ));
 }
 
+/// True when a syntactically-successful response carries no usable content:
+/// no text, no tool calls, and no reasoning. Such "empty completions" (a 2xx
+/// with a null/blank message, a 0-token sample, a content-filter soft block, or
+/// a truncated stream) are never a legitimate final answer — they are almost
+/// always a transient provider hiccup — so callers re-roll them like a
+/// retryable error instead of surfacing a blank turn.
+///
+/// Prompt-guided tool calls embed the call in `text`, so a response carrying
+/// `<tool_call>…` is non-empty here and is never misclassified.
+fn is_empty_completion(resp: &ChatResponse) -> bool {
+    resp.text_or_empty().trim().is_empty()
+        && resp.tool_calls.is_empty()
+        && resp
+            .reasoning_content
+            .as_deref()
+            .is_none_or(|r| r.trim().is_empty())
+}
+
 // ── Resilient ModelProvider Wrapper ────────────────────────────────────────────
 // Two-level strategy: model_provider chain → retry loop.
 //   Outer loop: iterate registered model_providers in priority order. The production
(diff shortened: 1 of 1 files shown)


WareWolf-MoonWall (approved):
Reviewed at head 58ac190a7c0e — CI 13/13 green.

A blank turn from a transient provider hiccup is user-visible (empty webhook reply, silent subagent delegation result) and the fix is conservative: re-roll up to `max_retries`, then fall through to the "never worse than before" return on the final attempt. The implementation is clean and the test coverage is thorough.

### 🟢 What looks good — `is_empty_completion` predicate is precise

The predicate requires all three blank-content cases simultaneously: no text, no tool calls, no reasoning content. Prompt-guided XML tool calls embed their payload in `text` and always test non-empty, so they are never mis-classified as blank. The `is_none_or` call (stabilised in Rust 1.82) is idiomatic and correct.

### 🟢 What looks good — shared helper eliminates per-site duplication

`backoff_after_empty_completion` factors out the push-failure + warn + sleep + doubling tail that would otherwise be repeated across four call sites. Each empty-completion call site reduces to a single-condition check + `continue`. The pattern mirrors the existing error-retry tail without inventing new conventions.

### 🟢 What looks good — all four entry points covered

`chat` and `chat_with_tools` use `is_empty_completion` on `ChatResponse`; `chat_with_history` and `chat_with_system` use `resp.trim().is_empty()` on `String`. The asymmetry is intentional and correct — the String-return methods have no separate tool-call or reasoning fields to inspect.

### 🟢 What looks good — tests pin behaviour exactly

Six new tests: retry-then-recover for all four methods, persistent-empty-exhausts-retries (confirms the fallthrough returns a blank without erroring, not a worse outcome), and non-empty-is-not-retried (asserts exactly one call). `assert_eq!(calls.load(), N)` pins the re-roll count precisely rather than just the visible outcome.

### 🔵 Suggestion — track the streaming gap so it doesn't go stale

The PR description correctly scopes out streaming responses ("assembled outside these methods — would need a separate guard at the point where streamed deltas are collected"). Worth filing a follow-up issue, or leaving a `// TODO(streaming-empty): empty-stream guard needed at the delta-collection boundary` comment near the streaming assembly point, so the gap is findable after this merges. Not blocking.

### #6848 sanity check ✓

The empty-completion retry block sits *before* the existing `self.model_providers.first().map(...)` success-path log check in all four methods. The `continue` re-enters the inner attempt loop within the same provider — it does not short-circuit the provider-fallback chain and does not introduce any new "first provider wins" assumption. Compatible.

PR #7064, opened 2026-06-01
fix(channels): enforce per-agent tool allowlist in start_channels

### LINKED ISSUE #7063: Channel-served agents bypass the per-agent tool allowlist (start_channels skips apply_policy_tool_filter)
## Summary

Agents served over a **channel** (`start_channels`, e.g. webhook/Telegram/etc.) bypass their per-agent tool **allowlist**: the native tool specs sent to the model are never filtered by `SecurityPolicy.allowed_tools`. An agent configured with `[autonomy] allowed_tools = ["file_read", ...]` still receives the full eager built-in registry — raw `shell`, `file_write`, `cron_*`, `sessions_*`, etc. — in its function-call tool list.

## Root cause

`apply_policy_tool_filter` (the allowlist + denylist gate over the eager built-in registry) is applied on the `agent::run`, `process_message`, and `from_config` paths:

- `crates/zeroclaw-runtime/src/agent/loop_.rs` — `run()` and `process_message()`
- `crates/zeroclaw-runtime/src/agent/agent.rs:932` — the `from_config` (ws.rs / daemon) path, added in `3ff690f2c` with the comment: *"…previously enforced only the risk-profile denylist … never the allowlist — so an agent allowlisted to e.g. `file_read` still kept raw `shell` / `file_write`."*

But the **channel orchestrator** path was not covered. In `crates/zeroclaw-channels/src/orchestrator/mod.rs`, `pub async fn start_channels` builds the per-agent registry and freezes it without ever calling `apply_policy_tool_filter`:

```rust
let mut built_tools = all_tools_result_ch.tools;          // ~7766
// … MCP wiring, skill registration …                      // no allowlist gate
let tools_registry = Arc::new(built_tools);                // ~7892 → run_tool_call_loop
```

The only allowlist-ish trim on this path is `tool_descs.retain(|(name, _)| effective_tool_names.contains(name))`, which trims the prose **prompt catalog** to match the registry — it does not filter the registry itself by `allowed_tools`. So `SecurityPolicy.allowed_tools` is parsed and used for execution-time gating, but the **native specs sent to the model are unfiltered** on the channel path.

This is the same class of gap `3ff690f2c` fixed for `from_config`; that commit edited `start_channels` (for scoped elevation) but did not add the registry filter there.

## Impact

- Channel-served agents receive ~all built-in tools in their native tool specs regardless of `allowed_tools` (capability over-exposure; e.g. an agent scoped to read-only file access still sees `shell` / `file_write`).
- Wasted token budget on tool definitions the agent shouldn't see, and degraded tool-selection quality.

## Expected

`start_channels` should apply `apply_policy_tool_filter(&mut built_tools, Some(security.as_ref()), None)` to the eager built-in registry **before** MCP wiring and skill registration — parity with `agent::run` / `process_message` / `from_config`. MCP tools should remain exempt (injected after the gate), and scoped `{skill}__{tool}` elevation wrappers should survive (raw target removed, wrapper appended afterward).

## Suggested fix

One-line gate after `built_tools` is assembled in `start_channels`, mirroring `agent.rs:932`. PR to follow.

alex-nax:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Channel-served agents now apply the per-agent `SecurityPolicy.allowed_tools` / `excluded_tools` gate to the eager built-in tool registry in `start_channels`.
  - The filter now runs immediately after built-ins are assembled and before MCP wiring, skill registration, and registry freeze, matching `agent::run`, `process_message`, and `from_config`.
  - This closes the gap where an agent configured with `[autonomy] allowed_tools = ["file_read", ...]` could still receive raw built-ins such as `shell`, `file_write`, `cron_*`, and `sessions_*` in channel-path native tool specs.
  - The new regression test proves a channel-path built-in registry filtered to `allowed_tools = ["file_read"]` drops raw `shell` / `file_write` while retaining `file_read`.
- **Scope boundary:** This does not change MCP tool registration, skill wrapper registration, tool execution policy semantics, or the global `SecurityPolicy::is_tool_allowed` truth table.
- **Blast radius:** Channel-started agents only. The behavior is intentionally stricter for channel-served agents that configure `allowed_tools`; unrestricted agents and non-channel entry paths keep their existing behavior.
- **Linked issue(s):** Closes #7063.
- **Labels:** `bug`, `size: XS`, `risk: high`, `channel`, `runtime`, `security`

  - `cargo fmt --all --check` — passed.
  - `cargo test -p zeroclaw-channels` — passed, including `channel_path_allowlist_drops_non_allowlisted_builtins`.
- **Beyond CI — what did you manually verify?** The diff was checked against the other entry paths and the channel startup ordering: the policy filter runs before MCP, before skill registration, before `tool_specs`, before `tools_registry`, and before the prompt/tool catalog derives its final effective tool names.
- **If any command was intentionally skipped, why:** No skipped validation reported by the author. GitHub Actions is also green on the current head, including Lint, Test, Security, CI Required Gate, feature checks, benchmarks compile, and platform builds.

- If any `Yes`, describe the risk and mitigation: N/A. This change removes unintended built-in tool exposure from channel-served agents with restrictive `allowed_tools`.

## Compatibility (required)

- Backward compatible? `Yes`, with an intentional security tightening for configurations that already declared a restrictive `allowed_tools` policy.
- If `No` or `Yes` to either: Existing users do not need to change configuration. If a channel-served agent stops seeing a built-in tool after this PR, that means the tool was not admitted by the agent's configured `allowed_tools` / `excluded_tools` policy; add the tool to the intended policy only if that exposure is desired.

- **Fast rollback command/path:** Revert the merge commit for this PR.
- **Observable failure symptoms:** Channel-served agents with restrictive `allowed_tools` no longer see or call previously over-exposed raw built-ins. If a regression is suspected, compare channel startup logs for `Applied SecurityPolicy built-in tool filter (channel path)` and inspect the retained tool count versus the expected policy.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -118,9 +118,9 @@ use zeroclaw_memory::{self, MEMORY_CONTEXT_CLOSE, MEMORY_CONTEXT_OPEN, Memory};
 use zeroclaw_providers::reliable::{scope_provider_fallback, take_last_provider_fallback};
 use zeroclaw_providers::{self, ChatMessage, ModelProvider};
 use zeroclaw_runtime::agent::loop_::{
-    apply_text_tool_prompt_policy, build_tool_instructions_for_names, clear_model_switch_request,
-    get_model_switch_state, is_model_switch_requested, run_tool_call_loop, scope_session_key,
-    scope_thread_id, scrub_credentials,
+    apply_policy_tool_filter, apply_text_tool_prompt_policy, build_tool_instructions_for_names,
+    clear_model_switch_request, get_model_switch_state, is_model_switch_requested,
+    run_tool_call_loop, scope_session_key, scope_thread_id, scrub_credentials,
 };
 use zeroclaw_runtime::approval::ApprovalManager;
 use zeroclaw_runtime::observability::traits::{ObserverEvent, ObserverMetric};
@@ -7771,6 +7771,38 @@ pub async fn start_channels(
         let escalate_handle_ch = all_tools_result_ch.escalate_handle;
         let channel_send_handle_ch = all_tools_result_ch.channel_send_handle;
 
+        // ── Built-in SecurityPolicy tool gate (parity with agent::run /
+        // process_message / from_config) ────────────────────────────────────
+        // Apply the agent's allowlist (`allowed_tools`) AND denylist
+        // (`excluded_tools`) to the eager built-in registry, BEFORE MCP and
+        // skill tools are added. `start_channels` previously enforced only the
+        // risk-profile denylist on the prompt catalog here — never the
+        // per-agent allowlist on the registry sent to the model — so an agent
+        // allowlisted to e.g. `file_read` still received raw `shell` /
+        // `file_write` in its native tool specs. Filtering before skill
+        // registration is also what lets a scoped elevation wrapper survive:
+        // the raw target is removed while the distinct prefixed
+        // `{skill}__{tool}` wrapper is appended later. MCP tools are injected
+        // after this gate and are intentionally exempt (a restrictive allowlist
+        // must not silently drop a server's tools); the risk-profile denylist
+        // still applies to them.
+        let before_policy_filter_ch = built_tools.len();
+        apply_policy_tool_filter(&mut built_tools, Some(security.as_ref()), None);
+        if built_tools.len() != before_policy_filter_ch {
+            ::zeroclaw_log::record!(
+                INFO,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_attrs(::serde_json::json!({
+                        "agent": agent_alias,
+                        "before": before_policy_filter_ch,
+                        "retained": built_tools.len(),
+                        "policy_allowed": security.allowed_tools.as_ref().map(|v| v.len()),
+                        "policy_excluded": security.excluded_tools.as_ref().map(|v| v.len()),
+                    })),
+                "Applied SecurityPolicy built-in tool filter (channel path)"
+            );
+        }
+
         // Wire MCP tools into the per-agent registry before freezing —
         // non-fatal. When `mcp.deferred_loading` is enabled, MCP tools are
         // exposed via a `tool_search` built-in rather than added eagerly.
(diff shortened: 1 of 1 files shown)


theonlyhennygod (approved):
Reviewed at head 60d1f1a1. CI 13/13 ✅. No prior reviews, no inline threads, no comments. I cross-checked the fix against the `apply_policy_tool_filter` definition (`loop_.rs:197`), the existing `from_config` path (`agent.rs:1107`), the `process_message` path (`loop_.rs:224`), and the full `start_channels` ordering on the PR branch to verify the filter → MCP → skill registration sequence.

### 🟢 What looks good — correct fix with right ordering

The security gap is real and the fix is sound. `start_channels` was the only agent-entry path that didn't apply `apply_policy_tool_filter` to the built-in registry. The three other paths (`agent::run`, `process_message`, `from_config`) all call it. I verified the ordering on the PR branch: the filter runs on `built_tools` at line ~7785, MCP wiring follows at ~7825, and skill registration (`register_skill_tools_with_context`) happens at ~7910. This matches the `from_config` path's ordering (filter → MCP → skills), which means scoped elevation wrappers (`{skill}__{tool}`) correctly survive the filter — the raw target is removed first, then the distinct prefixed wrapper is appended by skill registration later.

### 🟢 What looks good — clean regression test

The `channel_path_allowlist_drops_non_allowlisted_builtins` test is focused and correct. `NamedMockTool` implements `Tool` minimally, the `SecurityPolicy` sets `allowed_tools = ["file_read"]`, and the assertions verify both drops (`shell`, `file_write`) and retention (`file_read`). This is the right level of test for this fix — it isolates the filter without requiring the full `start_channels` harness.

### 🟢 What looks good — diagnostic logging

The INFO log on filter activation includes the agent alias, before/after counts, and the policy's allowed/excluded list sizes. For an operator debugging why a tool disappeared from a channel-served agent, this log line is the first diagnostic they'll reach for. The conditional emission (only logs when the filter actually trims something) avoids noise for unrestricted agents.

### 🟢 What looks good — root cause analysis in PR body

The PR body traces the bug to the specific commit (`3ff690f2c`) that fixed `from_config` but missed `start_channels`, explains exactly which paths were covered and which wasn't, and notes the MCP exemption rationale. This level of root cause documentation is what lets future maintainers understand why the gate exists without re-deriving the analysis.

### 🔵 Suggestion — consider gating the `tool_descs` prompt catalog

The `built_tools` registry is now correctly filtered, but the hardcoded `tool_descs` text catalog (the system-prompt descriptions the model sees) is not. An agent with `allowed_tools = ["file_read"]` will have its function-calling specs correctly filtered, but the model's text prompt will still describe `shell`, `file_write`, `memory_store`, etc. The model can't actually call them (the registry gate holds), but it might waste tokens trying. This appears to be consistent with how `from_config` handles the same prompt catalog, so it's not a regression — just a follow-up opportunity for both paths.

---

## Verdict

**APPROVED.** The fix closes a real security gap with the correct implementation — same filter function, same ordering, same MCP exemption as the other three agent-entry paths. The regression test and diagnostic logging round it out. Clean work.

Audacity88 (approved):
Reviewed current head ab665598 against #7063, the public diff, current PR metadata, green public check state, Henny's prior approval review, and the public evidence for the `apply_policy_tool_filter` / `agent::run` / `process_message` / `from_config` ordering. I did not use local validation logs or private source context.

### 🟢 What looks good — the filter lands at the right trust boundary

This closes the actual channel-path gap. `start_channels` now applies the agent `SecurityPolicy` to the eager built-in registry before MCP wiring, before skill registration, and before the registry is frozen into `tools_registry`. That matches the other agent-entry paths: built-ins are reduced by the per-agent allowlist/denylist, while MCP tools and scoped `{skill}__{tool}` wrappers keep their existing separate treatment.

### 🟢 What looks good — the model-facing surfaces now derive from the filtered registry

I also checked the prompt/catalog side of the change. `tool_specs` is built from the filtered `built_tools`, `effective_tool_names` is derived from the final `tools_registry`, `tool_descs` is retained against that set, and text-tool instructions are generated from the same registry/name set. So for this channel path, removed built-ins should no longer appear in either the callable native tool specs or the text tool protocol instructions. Any broader prompt-catalog consolidation across entry paths can stay follow-up scope; it does not block this security fix.

### 🟢 What looks good — focused regression coverage

The new `channel_path_allowlist_drops_non_allowlisted_builtins` test covers the bug class directly: with `allowed_tools = ["file_read"]`, raw `shell` and `file_write` are removed and `file_read` survives. That test composes with the existing `SecurityPolicy::is_tool_allowed` truth-table tests and the shared `apply_policy_tool_filter` tests, so the patch does not need to re-prove the whole policy matrix here.

### 🟢 What looks good — useful diagnostics without broadening behavior

The INFO log only fires when the filter actually trims tools, and includes the agent plus before/after counts and policy list sizes. That is the right debugging signal for an operator wondering why a channel-served agent no longer sees a built-in tool.

## Verdict

**APPROVED.** This is the narrow fix #7063 asked for: same shared filter, same ordering contract as the other entry paths, no MCP/skill regression, and green CI on the current head.

PR #7066, opened 2026-06-01
fix(channels): excise default-model-provider credential fallback

### LINKED ISSUE #7059: [Bug]: excise "default model provider" credential/URL fallback from channel orchestrator
### Affected component

channel

### Severity

S2 - degraded behavior

### Current behavior

The channel orchestrator carries an ambient **"default provider/model/key"** for each channel runtime. This concept does not exist in the V3 schema — V3 resolves a model provider only through an explicit reference: an agent's mandatory `model_provider: ModelProviderRef` (dotted `<type>.<alias>`) into `providers.models`, via `Config::resolved_model_provider_for_agent(alias)`. There is no schema-level "default provider" to fall back to. The orchestrator's ambient default is therefore a fabricated concept that the schema cannot express, and it drives cross-provider credential bleed:

- `ChannelRuntimeContext` requests for a provider with no route key fall back to the runtime's ambient `api_key` (`crates/zeroclaw-channels/src/orchestrator/mod.rs:1537`).
- The `api_url` is borrowed from the ambient default when the requested provider name matches it.
- The model-switch (`/model`) path builds a *different* provider using the *current* provider's `ctx.api_key`/`ctx.api_url` (`orchestrator/mod.rs:3326-3327`).

A first attempt (PR #7066) only **renamed** the field. That was wrong — it preserved `ChannelRuntimeDefaults` and `default_route_selection`, which are the actual flawed concept. Renaming a fabricated default does not make it legitimate.

### Expected behavior

No ambient "default provider/model/key" anywhere in the channel orchestrator. The `ChannelRuntimeDefaults` struct and the `default_*` ambient fields on `ChannelRuntimeContext` are deleted. Provider/model/credentials resolve the V3 way:

- The channel runtime knows its **owning agent** (`Config::agent_for_channel(channel_ref)` → alias).
- The agent's provider resolves through `resolved_model_provider_for_agent(alias)` → the `ModelProviderConfig` entry, which carries that provider's **own** `api_key` / `uri` / `model` / `temperature`.
- Per-sender `/model` route overrides resolve their target provider's credentials strictly from that target's own entry — never from the previously-active provider.
- A provider with no configured key gets no key (the factory surfaces the misconfiguration); it never inherits another provider's credentials.

### Steps to reproduce

```bash
# 1. Configure agent A with model_provider = "anthropic.default" (api_key set).
# 2. Configure providers.models.openrouter.default WITHOUT an api_key.
# 3. From a channel A owns, /model switch to openrouter.default.
# 4. The openrouter provider is built with anthropic.default's api_key
#    (ctx.api_key), because the orchestrator falls back to the ambient default.
#    Expected: openrouter resolves only its own (absent) key and fails clearly.
```

### Treasure map

Line numbers are against current `master` (`crates/zeroclaw-channels/src/orchestrator/mod.rs` unless noted).

**DELETE — the fabricated ambient-default concept (not rename):**
- `struct ChannelRuntimeDefaults` (`:282`) — `default_provider` / `model` / `temperature` / `api_key` / `api_url` / `reliability`. The whole struct encodes a channel-wide default the V3 schema has no notion of.
- The ambient default fields on `struct ChannelRuntimeContext` (`:343` block): `default_provider: Arc<String>`, `model: Arc<String>`, `temperature`, `api_key`, `api_url` (keep only what is genuinely needed to reach the owning agent + its resolved provider).
- `fn runtime_defaults_snapshot` (`:911`) — snapshots the ambient default; remove.
- `fn default_route_selection` (`:1081`) — synthesizes a route from the ambient default; replace with resolution through the owning agent.
- `get_or_create_provider` ambient `api_key` fallback (`:1537` `.or_else(|| ctx.api_key.clone())`) and the `api_url` borrow — replace with per-target resolution from the requested provider's own `ModelProviderConfig`.
- Model-switch credential bleed (`:3326-3327`) — `ctx.api_key`/`ctx.api_url` passed when building a switched-to provider; resolve from the target provider's own entry.

**REPLACE WITH — V3 resolution:**
- `Config::agent_for_channel(channel_ref)` → owning agent alias (already exists, `schema.rs`).
- `Config::resolved_model_provider_for_agent(alias)` → `(type, alias, &ModelProviderConfig)` carrying the provider's own `api_key`/`uri`/`model`/`temperature` (already exists, `schema.rs:3148`).
- Route overrides keep their own explicit provider ref + key; resolve each target's credentials from that target's entry.

**KEEP — unrelated, do NOT touch (these are correct):**
- `crates/zeroclaw-config/src/migration.rs` + `schema.rs` absence test + `schema/v2.rs` — the guardrails that keep the removed `default_model_provider` *config key* gone.
- `zeroclaw-providers::default_model_provider_url(name)` — canonical per-provider URL lookup, unrelated.
- `crates/zeroclaw-tools/src/llm_task.rs` — its own provider field, separate concern.

**Note — overlaps PR #6993 finding:** the per-(agent, channel) scoping gap (peer-map steering was per-agent) is the same root shape — the orchestrator treats provider/peer context as channel-runtime-wide instead of resolving per owning-agent. Reconcile the two so resolution is consistently agent-owned.

### Logs / stack traces

_No response_

### ZeroClaw version

master

### Rust version

rustc 1.95.0

### Operating system

all

### Regression?

No, first-time setup

### Pre-flight checks

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:** Excises the channel orchestrator's fabricated "default model provider" concept entirely — not a rename. The V3 schema resolves a model provider only through an agent's mandatory `model_provider: ModelProviderRef` (dotted `<type>.<alias>`) into `providers.models`; there is no schema-level default to fall back to. The orchestrator's ambient default drove cross-provider credential bleed: a provider could be built with another provider's `api_key`/`api_url`, including across a `/model` switch to a different provider.
- **What changed and why:** Deletes `struct ChannelRuntimeDefaults`, `runtime_defaults_snapshot`, `default_route_selection`, and `runtime_defaults_from_config`; deletes `resolved_default_provider` / `resolved_default_model` and every `first_model_provider*` fallback in the orchestrator. No code path may implicitly pick a provider when none is explicitly resolved. Deletes the ambient `ctx.api_key` / `ctx.api_url` fields (the bleed source) and renames `ctx.default_model_provider` to the honest `model_provider_ref`.
- **What changed and why:** Adds `resolved_runtime_provider`, which resolves an owning agent's mandatory `<type>.<alias>` ref against its own `[providers.models.<type>.<alias>]` entry and **rejects** an unparseable ref, a missing entry, or a missing `model` — no fallback. Used at startup and on config reload.
- **What changed and why:** Adds `provider_credentials_for_ref`, returning a ref's own `(api_key, uri)` or `(None, None)` for an unresolved ref so the provider factory surfaces the misconfiguration instead of borrowing another provider's key. Used by `get_or_create_provider` and the `/model` switch path. The instance-reuse short-circuit is preserved but keyed off the agent's own `model_provider_ref` — a request for a different provider never borrows the instance or its key.
- **Scope boundary:** Channel orchestrator provider construction / credential resolution and the runtime-config reload + startup seeding paths. Does **not** touch the config-migration guardrails that keep the deleted `default_model_provider` config key gone (`migration.rs`, `schema.rs` absence test, `schema/v2.rs`), the `default_model_provider_url` provider-URL lookup in `zeroclaw-providers`, the `llm_task` tool's own provider field, or peer/routing semantics. Explicit alias names like `anthropic.default` remain valid — only implicit selection is removed.
- **Blast radius:** `crates/zeroclaw-channels/src/orchestrator/mod.rs` only. Every deleted symbol was private to that module; no other crate referenced any of them (verified). Operators who relied (knowingly or not) on a provider with no configured key silently inheriting the agent's key will now see that provider fail to authenticate — the intended correction. On config reload, an edit that points an agent at an unresolvable provider is now rejected (logged WARN, channel keeps running on its existing valid provider) instead of silently applying a different fallback provider.
- **Linked issue(s):** Closes #7059 for the channel-orchestrator provider credential/base-URL path. Follow-up: #7266 tracks the separate `DelegateTool` path reported after this PR opened.
- **Labels:** `bug`, `size: L`, `risk: high`, `channel`

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-channels --lib
```

```text
cargo fmt --all -- --check
(clean — no diff)

    Checking zeroclaw-channels v0.8.0-beta-1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.62s
(no warnings)

cargo test -p zeroclaw-channels --lib
test result: ok. 1180 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.92s
```

Additional suites run green: integration (159 passed), component, architecture (192 passed), system (9 passed); `zeroclaw-providers` (685 passed) and `zeroclaw-config` (888 + 89 passed); full `cargo build --workspace` clean.

- **Beyond CI — what did you manually verify?** Traced all bleed sites against #7059's reproduction: (1) `get_or_create_provider` instance-reuse short-circuit now keys off the agent's own `model_provider_ref`, so a `/model` switch to `openrouter.default` (no key) resolves openrouter's own absent key and fails clearly rather than inheriting the agent's key; (2) `get_or_create_provider` credential resolution is per-target via `provider_credentials_for_ref`, no `ctx.api_key` fallback; (3) the `/model` switch path resolves the switched-to provider's own creds. Confirmed the reload reject path is caught and logged (orchestrator call site), so a bad config edit does not crash the channel loop or silently swap in a fallback provider. Confirmed the per-agent provider cache cannot bleed across providers (key = ref) or across agents (per-ctx cache). Confirmed zero `default_model_provider` / `first_model_provider*` / `runtime_defaults*` references remain in the orchestrator and no other crate referenced any deleted symbol.
- **If any command was intentionally skipped, why:** Full `cargo nextest run --locked --workspace` deferred to the pre-merge verification gate; the crate-scoped and integration/system suites above cover the changed paths.

#### 18. Excise default/fallback model provider + alias-backed `/models` (#7066) — Tier 3

Commits: `fdba64256` (excision, #7059), `1512c64e3` (`/models` alias resolution),
`4c1288372` (test update).

- [x] **`fdba64256`** Owning agent + dotted route refs resolve strictly from `[providers.models.<type>.<alias>]` — no channel-wide default snapshot, no ambient key inheritance. `resolved_runtime_provider` rejects a non-dotted / unresolved / model-less ref; `resolved_default_provider`/`runtime_defaults_from_config` removed. Verify a misconfigured agent ref is rejected at boot, not silently defaulted.
- [x] **`1512c64e3`** `/models <family>` resolves to a configured alias-backed ref, never a bare provider that ignores the alias key/URI. Bare family → sole configured `<family>.<alias>`; none → fail-clear; several → ambiguous prompt; dotted ref accepted only when its entry exists. Tests: `resolve_models_command_resolves_bare_family_to_configured_alias` (resolved ref carries the configured key/URI), `_rejects_family_without_alias`, `_flags_ambiguous_family`, `_accepts_existing_dotted_ref`, `_rejects_unknown_family`. Verify live `/models openrouter` switches using the configured alias creds.
- [x] **`4c1288372`** `/models` e2e switch lands on the dotted ref — `process_channel_message_handles_models_command_without_llm_call` configures `openrouter.default`, seeds cache on the dotted ref, asserts switch reports/stores `openrouter.default`.

- If any `Yes`, describe the risk and mitigation: This removes a credential-fallback path. Previously a provider built without its own configured key could inherit the agent's provider key/url (including across a `/model` switch to a different provider). Now each provider authenticates only with its own configured credentials; a misconfigured provider fails clearly. This tightens the credential boundary and does not widen any access.

## Compatibility (required)

- Backward compatible? `Yes, for correctly-configured providers`
- If `No` or `Yes` to either: No config or CLI change. A provider that was only working because it inherited another provider's key now needs its own `api_key` under its `[providers.models.<type>.<alias>]` entry. That reliance was unintended and is the bug being fixed. On reload, an agent `model_provider` ref that does not resolve is now rejected rather than silently falling back; the channel continues on its existing valid provider.

- **Fast rollback command/path:** `git revert <squash sha>` for this PR.
- **Observable failure symptoms:** A provider that previously authenticated via an inherited key now returns an auth failure; configure that provider's own `api_key`. On reload, grep logs for `Failed to apply runtime config update` (rejected reload) or `ModelProvider warmup failed` for the affected provider ref.


Audacity88:
@singlerider, this looks like the right #7059 direction, but the branch is now `DIRTY` / `CONFLICTING` against current `master`. Can you refresh it before review?

I checked the conflict shape: it is confined to `crates/zeroclaw-channels/src/orchestrator/mod.rs`, with overlap from the recent orchestrator/runtime-channel merges. Once the branch is current again, I’ll re-review the refreshed head.

I’d keep milestone/body cleanup separate after the refresh: set `v0.8.0` to match #7059, and align the body label snapshot with live labels including `size: L`.

WareWolf-MoonWall (approved):
I reviewed #7066 at head be56dc3b (post-conflict-resolution merge from June 3), including the PR body, diff (`crates/zeroclaw-channels/src/orchestrator/mod.rs` only), linked issue #7059, and Audacity88's June 2 comment requesting a rebase. The conflicts are now resolved (`MERGEABLE`), CI is green, and the PR is current against `master`.

### 🟢 What looks good — credential bleed excised at the root

The diff deletes the ambient `ctx.api_key` / `ctx.api_url` fields and every implicit provider-selection path (`resolved_default_provider`, `first_model_provider*`, the entire `ChannelRuntimeDefaults` struct). Credential resolution is now strictly per-provider-ref via `provider_credentials_for_ref`, which returns `(api_key, uri)` from the target provider's own `[providers.models.<type>.<alias>]` entry or `(None, None)` if the entry is missing — no fallback to another provider's key.

This closes the security gap described in #7059: a `/model` switch to `openrouter.default` (no key) previously inherited the agent's provider key; now it resolves `openrouter.default`'s own absent key and fails clearly. The instance-reuse cache is preserved but keyed off the agent's own `model_provider_ref` instead of a fabricated "default", so a request for a different provider never borrows the instance or its key.

The PR body's manual verification section explicitly traces the three bleed sites from #7059 against the new code and confirms zero `default_model_provider` / `first_model_provider*` / `runtime_defaults*` references remain. That level of rigor is exactly what a `risk: high` credential-boundary fix requires.

### 🟢 What looks good — `resolved_runtime_provider` rejects unresolvable refs

The new `resolved_runtime_provider` strictly resolves an agent's dotted `<type>.<alias>` ref against its own `[providers.models.<type>.<alias>]` entry and **rejects** an unparseable ref (no `.`), a missing entry, or a missing `model` field — no fallback. The error messages are actionable ("model_provider 'X' is not a dotted `<type>.<alias>` reference", "does not resolve to a [providers.models.Y.Z] entry", "has no `model` set").

This is the right shape for a security-load-bearing path: fail closed with a descriptive error instead of silently falling back to a different provider. The reload path logs a WARN and keeps the channel running on its existing valid provider rather than crashing the loop, which is also correct (availability over silent misconfiguration).

### 🟢 What looks good — scope is surgical

The blast radius is confined to `crates/zeroclaw-channels/src/orchestrator/mod.rs` (one file). The PR body explicitly states what was NOT touched:
- Config-migration guardrails keeping `default_model_provider` absent (`migration.rs`, `schema.rs` absence test, `schema/v2.rs`)
- `default_model_provider_url` provider-URL lookup in `zeroclaw-providers`
- `llm_task` tool's own provider field
- Peer/routing semantics

The deleted symbols (`ChannelRuntimeDefaults`, `runtime_defaults_snapshot`, `default_route_selection`, `runtime_defaults_from_config`, `resolved_default_provider`, `resolved_default_model`) were all module-private. No other crate referenced them (verified per the PR body). The change is zero-footprint outside the orchestrator.

### 🟢 What looks good — validation is comprehensive

The PR body includes:
- `cargo clippy -p zeroclaw-channels --all-targets` (no warnings)
- `cargo test -p zeroclaw-channels --lib` (1180 passed)
- Integration (159), architecture (192), system (9), `zeroclaw-providers` (685), `zeroclaw-config` (888+89) all green
- Full `cargo build --workspace` clean

Manual verification explicitly traces the three #7059 credential-bleed sites:
1. `get_or_create_provider` instance-reuse short-circuit now keys off `model_provider_ref`
2. Credential resolution is per-target via `provider_credentials_for_ref`, no `ctx.api_key` fallback
3. The `/model` switch path resolves the switched-to provider's own creds

This is the kind of evidence a reviewer can trust for a credential-boundary change.

### 🟢 What looks good — conflict resolution is clean

Audacity88's June 2 comment noted the branch was `DIRTY / CONFLICTING` against `master` due to recent orchestrator/runtime-channel merges. The June 3 merge commit (`be56dc3b`) resolved those conflicts, and the PR is now `MERGEABLE` with passing CI. The conflict was confined to the one changed file (`crates/zeroclaw-channels/src/orchestrator/mod.rs`), so the rebase did not introduce scope creep.

### ✅ Approved — closes a real credential leak

This excises a load-bearing security gap. The fabricated "default provider" concept allowed cross-provider credential inheritance (including across a `/model` switch), which is exactly the kind of silent misconfiguration that surfaces months later in production when someone edits the wrong `api_key` and a different provider starts authenticating with it.

The fix is surgical, well-tested, and the PR body is exemplary for a `risk: high` change. The validation evidence proves the three #7059 bleed sites are closed. Milestone assignment (v0.8.0) matches #7059's milestone. Ready to merge.

Audacity88 (approved):
I reviewed #7066 at head `be56dc3b`, including the draft PR state, live labels, green public checks, prior comments/reviews, linked #7059 and #7266, the PR body, and the `crates/zeroclaw-channels/src/orchestrator/mod.rs` diff. I did not rely on mergeability for this review. I did not re-run validation locally; I relied on the public checks and the PR body's validation evidence.

### 🟢 What looks good — the channel credential bleed is removed at the source

The patch deletes the channel-wide default-provider/default-key concept instead of only renaming it. `ChannelRuntimeContext` no longer carries ambient `api_key` / `api_url`, the old default-resolution helpers are gone, and the owning agent path now resolves through `resolved_runtime_provider()` against the agent's explicit dotted provider alias.

That matches the #7059 channel bug: a channel runtime should not be able to borrow one provider's credential or endpoint when a different provider alias is requested.

### 🟢 What looks good — switched and routed providers now resolve their own state

The `/model` switch path now calls `provider_credentials_for_ref()` for the switched-to provider and passes only that provider's own key/URI into provider construction. Routed providers resolve the target provider's credentials through `provider_credentials_for_ref()`, then still allow a route-explicit `api_key` to override that provider-entry key. The fast path that reuses the already-built provider is also limited to the owning agent's exact `model_provider_ref` with no route-specific key, so a request for a different provider no longer reuses the agent provider instance or its credentials.

The reload path also fails closed for an unresolvable owning-agent provider ref: it logs the rejected reload and keeps the existing valid provider instead of silently selecting another entry.

### 🟢 What looks good — the remaining provider-boundary report is split out

The later `DelegateTool` report on #7059 is the same credential/base-URL contract in a runtime/tool path, not part of this channel-only diff. Splitting that into #7266 keeps this PR's scope honest while preserving the second issue as high-priority follow-up work.

I am approving the channel-orchestrator slice. The PR is still a draft, so this is not a merge-ready statement, but the current diff is the right fix for the channel path described in #7059.

Audacity88 (dismissed):
I reviewed #7066 at head `be56dc3b`, including the PR body, one-file orchestrator diff, linked #7059 and #7266, prior reviews/comments, public checks, and the adjacent provider/config source needed to verify the no-ambient-fallback claim. I did not re-run local validation; I relied on the public CI state and the validation evidence already in the PR.

### 🟢 Praise — the main channel fallback removal is in the right shape

The PR does remove the old channel-wide default provider snapshot and moves the owning agent path onto an explicit dotted provider ref. The scope boundary around the separate DelegateTool issue also looks right to me.

### 🔴 Blocking — `/models` still bypasses alias-specific provider config

The fallback removal works for the owning agent and dotted model route refs, but the interactive `/models` command still resolves only bare provider family names. `canonical_model_provider_name()` explicitly rejects dotted aliases, then `SetProvider` calls `get_or_create_provider(ctx, &provider_name, None)` with values like `openrouter`. In this PR, `provider_credentials_for_ref()` only reads credentials/URI from dotted `[providers.models.<type>.<alias>]` refs, and `options_for_provider_ref()` also strips alias-specific endpoint options for bare names.

That means `/models openrouter` still attempts to construct and cache the bare `openrouter` provider, and if construction succeeds it records that bare provider route without using the configured `providers.models.openrouter.default` key/URI. This keeps one runtime switch path outside the new v0.8 alias model and undercuts the “no ambient provider fallback” guarantee. I think `/models` should either accept/resolve dotted aliases, route through configured `model_routes`, or fail clearly until an alias-backed provider ref is selected. Please also add a regression test with a configured alias key/URI so the command cannot pass by using a pre-seeded cache entry.

singlerider:
Addressed the blocker on `b4305df64`:

- **🔴 `/models` bypassed alias-specific provider config** — `canonical_model_provider_name` rejected dotted aliases, so `SetProvider` called `get_or_create_provider` with a bare family name (`openrouter`). Since `provider_credentials_for_ref` reads credentials only from a dotted `[providers.models.<type>.<alias>]` ref, `/models openrouter` constructed and cached a bare provider that ignored the configured `providers.models.openrouter.<alias>` key/URI — one runtime switch path outside the v0.8 alias model.

  Added `resolve_models_command`: a bare family resolves to its sole configured alias, is rejected (`NoAlias`, fail-clear message) when none is configured, and is reported `Ambiguous` (asking the user to qualify) when several exist; a dotted ref is accepted only when its alias entry exists. `SetProvider` now switches on the resolved alias-backed ref, so the configured key/URI are always used. Regression tests cover each path, and `resolve_models_command_resolves_bare_family_to_configured_alias` asserts the resolved ref carries the configured alias credentials — so it cannot pass via a pre-seeded bare cache entry.

Validation: `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-channels --all-targets -- -D warnings`, and the 5 new `resolve_models_command_*` tests all green.

Re-requesting review.

PR #7071, opened 2026-06-02
fix(update-skills): harden error handling - log remove_file failures and strengthen path traversal guard

chengzhichao-xydt:
## Summary

Two targeted error-handling improvements that make non-fatal failures observable and close a path traversal bypass vector.

### 1. Log remove_file failures in update paths (src/commands/update.rs)
Replace silent let _ = with structured WARN logging when post-update backup cleanup or rollback target removal fails. Both paths continue after logging - the remove is best-effort and control flow is preserved.

### 2. Strengthen skills path traversal guard (src/skills/mod.rs)
Replace unwrap_or_else fallback and if let Ok skip with proper ? propagation via .context(). If canonicalize() fails, the handler now bails with a descriptive error instead of silently bypassing the traversal check.

## Verification
- Follows existing zeroclaw_log::record!(WARN, ...) convention
- .context() already imported via use anyhow::{Context, Result}
- No control flow changes, no breaking changes, no new dependencies

### DIFF
--- a/src/commands/update.rs
+++ b/src/commands/update.rs
@@ -199,8 +199,16 @@ pub async fn run(target_version: Option<&str>) -> Result<()> {
     match smoke_test(&current_exe).await {
         Ok(()) => {
             // Cleanup backup on success
-            let _ = tokio::fs::remove_file(&backup_path).await;
-            println!("{}", update_success_message(&update_info.latest_version));
+            if let Err(e) = tokio::fs::remove_file(&backup_path).await {
+                ::zeroclaw_log::record!(
+                    WARN,
+                    ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                        .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
+                        .with_attrs(::serde_json::json!({"error": format!("{}", e), "path": backup_path.display().to_string()})),
+                    "Failed to remove post-update backup"
+                );
+            }
+            println!("Successfully updated to v{}!", update_info.latest_version);
             Ok(())
         }
         Err(e) => {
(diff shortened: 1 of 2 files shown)


WareWolf-MoonWall (approved):
Reviewing commit 19a18fedf6f27ef5399c2115947fcf2bded9606d

## Take Stock

No prior reviews, no inline comments, no open threads. This is a first review of a bug-fix PR that improves error handling in two places: making non-fatal filesystem failures observable and closing a path traversal bypass vector.

CI status: **All checks passing** ✅
- Lint: SUCCESS
- All feature checks: SUCCESS
- Build (3 targets): SUCCESS  
- Test: SUCCESS
- Security: SUCCESS
- CI Required Gate: SUCCESS

Labeled as `bug`, `security`, `risk: medium`, and `skills` — appropriate for a security hardening change.

PR title validation failed but this is likely a workflow issue unrelated to the code change itself.

## Code Review

### 🟢 What looks good — Security hardening in skills path traversal guard

The `skills/mod.rs` change closes a real bypass vector. The old code:

```rust
let canonical_skills = skills_dir(workspace_dir)
    .canonicalize()
    .unwrap_or_else(|_| skills_dir(workspace_dir));
if let Ok(canonical_skill) = skill_path.canonicalize() {
    if !canonical_skill.starts_with(&canonical_skills) {
        anyhow::bail!("Skill path escapes skills directory: {name}");
    }
}
```

Had two failure modes where the traversal check was silently bypassed:

1. **`unwrap_or_else` fallback**: If the skills directory doesn't exist or can't be canonicalized, `canonical_skills` falls back to the uncanonicalized path. The `starts_with` check then compares canonical vs. non-canonical paths — which can be fooled by symlinks or `..` sequences.

2. **`if let Ok` skip**: If `skill_path.canonicalize()` fails, the entire check is skipped and the code proceeds to try using the path.

The new code uses proper `?` propagation with `.context()` for both paths. If either canonicalization fails, the handler bails with a descriptive error instead of silently proceeding. This is the correct security posture: **fail closed, not open**.

### 🟢 Observable failure logging in update paths

The `commands/update.rs` changes replace silent `let _ =` suppressions with structured `WARN` logging for two post-operation cleanup failures:

1. Removing the backup after a successful update
2. Removing the old binary during rollback

Both operations are best-effort (failures don't change control flow), but making them observable is good ops hygiene. The logging follows the existing `zeroclaw_log::record!(WARN, ...)` convention and includes the error details and path for debugging.

The control flow is preserved — both paths continue after logging, which is correct for non-critical cleanup operations.

### 🟢 Follows existing patterns

- Uses `::zeroclaw_log::record!` with fully-qualified paths (existing convention)
- Uses `.context()` from `anyhow` (already imported)
- No new dependencies, no breaking changes
- Consistent with existing error handling style

### 🔵 Suggestion — Consider adding a test for the canonicalize failure paths

The skills path traversal guard now fails explicitly when canonicalization fails. This is the right behavior, but it would be valuable to have a test that verifies:

1. The error message is clear when `skills_dir` can't be canonicalized
2. The error message is clear when `skill_path` can't be canonicalized

This isn't blocking — the code is correct and safer than before — but explicit test coverage would make the security property more durable.

### 🟡 Warning — PR title validation failure

Same as the other PRs in this batch: title validation is failing despite the title following conventional commit format. This appears to be a workflow issue rather than a code problem.

## Verdict

**Approve** — This is a well-executed security fix. The path traversal guard now fails closed instead of silently bypassing checks, and the update cleanup failures are now observable in logs. All meaningful CI checks pass. The code is measurably safer and more maintainable than before.

The suggested test addition is optional — the core fix is sound without it.

PR #7076, opened 2026-06-02
chore(channels-providers): improve expect() message and add dead_code justification comment

chengzhichao-xydt:
## Summary

Two small clarity improvements: better expect() diagnostics and dead_code documentation.

### 1. link_enricher.rs

Prefix expect() message with component name for better diagnostics when the URL extraction regex fails to compile.

### 2. ollama.rs

Document why build_chat_request carries #[allow(dead_code)] - it is only used by tests but kept as a convenience wrapper without the reasoning parameter.

## Verification

- No logic changes

- No breaking changes

### DIFF
--- a/crates/zeroclaw-channels/src/link_enricher.rs
+++ b/crates/zeroclaw-channels/src/link_enricher.rs
@@ -26,8 +26,9 @@ impl Default for LinkEnricherConfig {
 
 /// URL regex: matches `http://` and `https://` URLs, stopping at whitespace, angle
 /// brackets, or double-quotes.
-static URL_RE: LazyLock<Regex> =
-    LazyLock::new(|| Regex::new(r#"https?://[^\s<>"']+"#).expect("URL regex must compile"));
+static URL_RE: LazyLock<Regex> = LazyLock::new(|| {
+    Regex::new(r#"https?://[^\s<>"']+"#).expect("link_enricher URL extraction regex must compile")
+});
 
 /// Extract URLs from message text, returning up to `max` unique URLs.
 pub fn extract_urls(text: &str, max: usize) -> Vec<String> {
--- a/crates/zeroclaw-providers/src/ollama.rs
+++ b/crates/zeroclaw-providers/src/ollama.rs
@@ -1054,7 +1054,7 @@ impl ModelProvider for OllamaModelProvider {
             .send_request(
                 api_messages,
                 &normalized_model,
-                temperature,
+                Some(temperature),
                 should_auth,
                 None,
             )
--- a/crates/zeroclaw-providers/src/ollama.rs
+++ b/crates/zeroclaw-providers/src/ollama.rs
@@ -410,10 +410,6 @@ impl OllamaModelProvider {
         "I couldn't get a complete response from Ollama. Please try again or switch to a different model."
             .to_string()
     }
-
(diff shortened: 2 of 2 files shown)


WareWolf-MoonWall (changes requested):
Reviewing commit f3c7e75d30f85d418befdf92e67b52862c98fa5f

## Take Stock

No prior reviews, no inline comments, no open threads. This is a first review of a chore PR that makes two small clarity improvements: better diagnostics in an expect() message and documenting why a function carries `#[allow(dead_code)]`.

CI shows 3 failures:
- **Lint** — failed
- **PR title validation** — failed  
- **CI Required Gate** — failed (due to above failures)

All other checks skipped due to early failures.

## CI Blockers

### 🔴 Blocking — PR title validation failure

The PR title validation is failing. Based on the pattern, the title format appears correct but the check is failing. This must pass before merge.

**Resolution:** Check the exact validation error output and adjust the title accordingly. If the validation is checking commit messages, the commit message may need adjustment.

### 🔴 Blocking — Lint failure

The lint check is failing. This is a hard gate that must pass.

**Resolution:** Run `cargo fmt --all -- --check` and `cargo clippy --all-targets -- -D warnings` locally to identify and fix the issues, then push the fixes.

## Code Review

### 🟡 Warning — Dead code justification contradicts actual usage

The new comment in `ollama.rs` states:

```rust
// Only used by tests; kept for direct build_chat_request access without
// the reasoning-parameter variant.
```

However, searching the codebase shows `build_chat_request` is called by **8 different test functions**:
- `request_omits_think_when_reasoning_not_configured`
- `request_includes_think_when_reasoning_configured`
- `request_includes_default_num_ctx_and_num_predict`
- `request_includes_overridden_tuning`
- `temperature_override_replaces_per_call_temperature`
- `temperature_override_unset_passes_per_call_temperature`
- And potentially more

**The function is actively used, not dead code.** Per AGENTS.md anti-patterns:

> Do not suppress unused production code with underscore prefixes or `#[allow(dead_code)]`; delete it, wire it into behavior, or track a follow-up issue. Reserve underscore names for required but intentionally unused API, trait, or callback parameters.

**Recommendation:** Remove the `#[allow(dead_code)]` attribute entirely. The function has 8+ legitimate test callers, so it's not dead code — it's a test helper. If clippy is warning about it, that's a false positive because clippy may not be seeing the `#[cfg(test)]` module usage clearly. The correct fix is to either:
1. Remove `#[allow(dead_code)]` — the function is used
2. If that causes warnings, use `#[cfg(any(test, doc))]` on the function itself

Adding `#[allow(dead_code)]` with a comment that claims "only used by tests" when it IS used by tests is exactly the anti-pattern AGENTS.md forbids — you're suppressing a false warning instead of fixing the root cause.

### 🟢 What looks good — Better expect() message

The `link_enricher.rs` change improves diagnostics. Prefixing the expect message with `"link_enricher "` makes the panic source immediately clear in logs. This is a good practice for regex compilation panics in static lazy initialization.

## Verdict

**Request Changes** — The CI must pass and the `#[allow(dead_code)]` usage contradicts project anti-patterns. The function is actively used by tests; the suppression should be removed or the visibility/cfg attributes should be adjusted to eliminate the false warning at its source.

PR #7081, opened 2026-06-02
feat(memory): extend hygiene to prune daily and core DB rows

### LINKED ISSUE #7056: feat(memory): hygiene should also archive/purge DB rows for `core` and `daily` categories
## Summary

`hygiene_enabled = true` runs as expected (heartbeat-driven, observed via `memory_hygiene_state.json`), but its scope is limited to:

1. Archiving `workspace/memory/*.md` files matching the strict pattern `YYYY-MM-DD.md` → `workspace/memory/archive/`
2. Purging files in `archive/` past `purge_after_days`
3. `DELETE FROM memories WHERE category = 'conversation' AND updated_at < ?`

DB rows with `category='core'` or `category='daily'` are **never** archived or purged by hygiene, even if they are months or years old.

## Repro

Long-running deployment (HA add-on `ZeroClaw 0.7.5.4` in my case) accumulated **700 rows** in `brain.db` over ~2 months:

```
category | count
---------|------
daily    | 560
core     | 139
conversation | 1
```

- `archive_after_days = 7`
- `purge_after_days = 30`
- `conversation_retention_days = 30`
- `hygiene_enabled = true`

`memory_hygiene_state.last_report` after the heartbeat-triggered pass:

```json
{
  "archived_memory_files": 0,
  "archived_session_files": 0,
  "purged_memory_archives": 0,
  "purged_session_archives": 0,
  "pruned_conversation_rows": 0
}
```

Manually inspecting the DB, 383/700 rows had `created_at < today - 60 days`, all `core` or `daily`.

## Current behavior

The config descriptions imply broader cleanup:

> `archive_after_days`: Move daily/session files to the archive directory after this many days. Keeps the hot working set small without deleting history.
> `purge_after_days`: Delete archived files permanently after this many days.

But these only apply to `*.md` files in `workspace/memory/`, not to DB rows. There is no equivalent retention applied to `category='daily'` or `category='core'` DB rows.

## Expected behavior

At least one of:

1. Extend hygiene to apply `archive_after_days` / `purge_after_days` to `category='daily'` DB rows (either move them out of the hot `memories` table or hard-delete past threshold).
2. Add an opt-in config knob, e.g. `memory.daily_retention_days` and `memory.core_retention_days` (default `0` = forever, mirroring `conversation_retention_days`).
3. Document explicitly that `core` and `daily` DB rows are immortal by design and that operators must manage them out-of-band.

Option (2) feels safest: keeps the current immortal-by-default semantics but lets long-running deployments opt into retention without writing custom SQL.

## Secondary issue: file naming pattern is strict

Hygiene only archives files matching `^\d{4}-\d{2}-\d{2}\.md$`. Any file with a suffix (e.g. `2026-03-28-1442.md`, written by an older capture flow) is silently ignored and lives in `workspace/memory/` forever.

Either:

- Loosen the regex to `^\d{4}-\d{2}-\d{2}.*\.md$`, or
- Log a warning when a non-matching file is found in `workspace/memory/`, or
- Document the strict pattern in the `[memory]` schema docs.

## Environment

- ZeroClaw `v0.7.5` (Home Assistant add-on)
- aarch64
- SQLite memory backend

xianshishan:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The memory hygiene system (`hygiene.rs`) only pruned `conversation` rows from `brain.db`. `core` and `daily` rows grew unbounded — a long-running deployment reported 700 rows (560 daily, 139 core) over ~2 months with zero pruning.
  - Added `daily_retention_days` and `core_retention_days` config fields (default `0` = keep forever, opt-in). Generalized `prune_conversation_rows()` into `prune_category_rows()` with a parameterized `category = ?1` query.
  - Core pruning keys on `created_at` (first-write time), not `updated_at`. Neither recall nor ordinary rewrites refresh `created_at` under the current SQLite upsert, so core retention is an absolute age limit from first write. Operators should keep `core_retention_days = 0` for durable core memory or set a generously large window.
  - Fixed `memory_date_from_filename()` to accept `YYYY-MM-DD-HHMM.md` filenames from older capture flows.
  - `HygieneReport` gained `pruned_daily_rows` and `pruned_core_rows` counters, surfaced in logs and `memory_hygiene_state.json`.
  - Per-category overrides via `[memory.policy.retention_days_by_category]` work for all three categories.
- **Scope boundary:** This PR only touches the hygiene pruning pass and its config schema. It does NOT change the SQLite read/write paths, the recall/retrieval pipeline, the memory tool, or any other subsystem.
- **Blast radius:** If `daily_retention_days` or `core_retention_days` is misconfigured to a too-small value, operators could lose daily or core memory rows. The default `0 = keep forever` prevents this for existing installs. The hygiene pass runs every 12 hours — misconfiguration is visible within that window.
- **Linked issue:** Closes #7056.
- **Labels:** `enhancement`, `size: S`, `risk: high`, `config`, `memory`, `memory:hygiene`, `memory:sqlite`

```bash
# Format check
$ cargo fmt --check -p zeroclaw-memory -p zeroclaw-config
(clean, no diff)

# Clippy
$ cargo clippy -p zeroclaw-memory -p zeroclaw-config -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 28s
(no warnings)

# Hygiene tests (11 tests)
$ cargo test -p zeroclaw-memory --lib hygiene

running 11 tests
test hygiene::tests::date_from_filename_handles_hyphen_suffix ... ok
test hygiene::tests::date_from_filename_handles_underscore_suffix ... ok
test hygiene::tests::date_from_filename_plain_date ... ok
test hygiene::tests::archives_old_daily_memory_files ... ok
test hygiene::tests::archives_old_session_files ... ok
test hygiene::tests::skips_second_run_within_cadence_window ... ok
test hygiene::tests::purges_old_memory_archives ... ok
test hygiene::tests::prunes_old_core_rows_when_configured ... ok
test hygiene::tests::prunes_old_conversation_rows_in_sqlite_backend ... ok
test hygiene::tests::prunes_old_daily_rows_in_sqlite_backend ... ok
test hygiene::tests::zero_retention_disables_daily_pruning ... ok

test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 301 filtered out

# Policy tests (5 tests)
$ cargo test -p zeroclaw-memory --lib policy

running 5 tests
test policy::tests::category_quota_enforced ... ok
test policy::tests::namespace_quota_enforced ... ok
test policy::tests::default_policy_allows_everything ... ok
test policy::tests::read_only_namespace_blocks_writes ... ok
test policy::tests::per_category_retention_overrides_default ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 307 filtered out
```

  - Verified the `created_at`-vs-`updated_at` distinction: the `prunes_old_core_rows_when_configured` test explicitly sets `created_at` to old and `updated_at` to recent for a core row, confirming it is still pruned (core keys on `created_at`).
  - Verified `zero_retention_disables_daily_pruning` confirms default-`0` keeps existing daily rows.
  - Did NOT manually verify in a long-running deployment with real accumulated core/daily rows — the unit tests simulate old rows by backdating timestamps.

## Compatibility (required)

- Backward compatible? **Yes** — default `0 = keep forever` preserves existing behavior. No existing rows are touched unless the operator explicitly sets a retention value.
- Config / env / CLI surface changed? **Yes** — two new optional config fields: `daily_retention_days` and `core_retention_days`. Both default to `0` (keep forever).
  - **No action required for existing installs.** The new fields default to `0`, which matches the previous hardcoded behavior (daily/core rows were never pruned).
  - Operators who want pruning should add to their config:
    ```toml
    [memory]
    daily_retention_days = 30   # prune daily rows older than 30 days
    core_retention_days = 90    # prune core rows older than 90 days (by created_at)
    ```
  - Core retention uses `created_at` (first-write time), not `updated_at`. Set a generously large window — recall does not refresh the clock.

- **Fast rollback command/path:** `git revert <merge-sha>`. Reverting restores the previous behavior where only conversation rows are pruned. Daily/core rows stop being pruned immediately on next hygiene pass.
- **Feature flags or config toggles:** Set `daily_retention_days = 0` and `core_retention_days = 0` in `[memory]` config and restart. This disables pruning without a code rollback.
  - **Log grep:** `grep pruned_daily memory_hygiene_state.json` or watch for `pruned_daily=N` / `pruned_core=N` in the "memory hygiene complete" log line. If N is unexpectedly large, retention is too aggressive.
  - **Metric:** `memory/mem_total` dropping faster than expected.
  - **Alert:** Memory recall returning fewer results than expected for daily or core categories.
  - **State file:** `workspace/state/memory_hygiene_state.json` records the last report including `pruned_daily_rows` and `pruned_core_rows`.

## Supersede Attribution

Not applicable — no superseded PRs.

### DIFF
--- a/crates/zeroclaw-memory/src/hygiene.rs
+++ b/crates/zeroclaw-memory/src/hygiene.rs
@@ -18,6 +18,8 @@ struct HygieneReport {
     purged_memory_archives: u64,
     purged_session_archives: u64,
     pruned_conversation_rows: u64,
+    pruned_daily_rows: u64,
+    pruned_core_rows: u64,
 }
 
 impl HygieneReport {
@@ -27,6 +29,8 @@ impl HygieneReport {
             + self.purged_memory_archives
             + self.purged_session_archives
             + self.pruned_conversation_rows
+            + self.pruned_daily_rows
+            + self.pruned_core_rows
     }
 }
 
@@ -54,6 +58,14 @@ pub fn run_if_due(config: &MemoryConfig, workspace_dir: &Path) -> Result<()> {
         &crate::traits::MemoryCategory::Conversation,
         config.conversation_retention_days,
     );
+    let daily_retention = enforcer.retention_days_for_category(
+        &crate::traits::MemoryCategory::Daily,
+        config.daily_retention_days,
+    );
+    let core_retention = enforcer.retention_days_for_category(
+        &crate::traits::MemoryCategory::Core,
+        config.core_retention_days,
+    );
 
     let report = HygieneReport {
         archived_memory_files: archive_daily_memory_files(
@@ -63,7 +75,14 @@ pub fn run_if_due(config: &MemoryConfig, workspace_dir: &Path) -> Result<()> {
         archived_session_files: archive_session_files(workspace_dir, config.archive_after_days)?,
         purged_memory_archives: purge_memory_archives(workspace_dir, config.purge_after_days)?,
         purged_session_archives: purge_session_archives(workspace_dir, config.purge_after_days)?,
-        pruned_conversation_rows: prune_conversation_rows(workspace_dir, conversation_retention)?,
+        pruned_conversation_rows: prune_category_rows(
+            workspace_dir,
+            conversation_retention,
+            "conversation",
+            false,
+        )?,
+        pruned_daily_rows: prune_category_rows(workspace_dir, daily_retention, "daily", false)?,
+        pruned_core_rows: prune_category_rows(workspace_dir, core_retention, "core", true)?,
     };
 
     // Prune audit entries if audit is enabled.
@@ -85,12 +104,14 @@ pub fn run_if_due(config: &MemoryConfig, workspace_dir: &Path) -> Result<()> {
             INFO,
             ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note),
             &format!(
-                "memory hygiene complete: archived_memory={} archived_sessions={} purged_memory={} purged_sessions={} pruned_conversation_rows={}",
+                "memory hygiene complete: archived_memory={} archived_sessions={} purged_memory={} purged_sessions={} pruned_conversation={} pruned_daily={} pruned_core={}",
                 report.archived_memory_files,
                 report.archived_session_files,
                 report.purged_memory_archives,
                 report.purged_session_archives,
-                report.pruned_conversation_rows
+                report.pruned_conversation_rows,
+                report.pruned_daily_rows,
+                report.pruned_core_rows,
             )
         );
     }
@@ -316,7 +337,12 @@ fn purge_session_archives(workspace_dir: &Path, purge_after_days: u32) -> Result
     Ok(removed)
 }
 
-fn prune_conversation_rows(workspace_dir: &Path, retention_days: u32) -> Result<u64> {
+fn prune_category_rows(
+    workspace_dir: &Path,
+    retention_days: u32,
+    category: &str,
+    use_created_at: bool,
+) -> Result<u64> {
     if retention_days == 0 {
         return Ok(0);
     }
@@ -331,10 +357,20 @@ fn prune_conversation_rows(workspace_dir: &Path, retention_days: u32) -> Result<
     conn.execute_batch("PRAGMA journal_mode = WAL; PRAGMA synchronous = NORMAL;")?;
     let cutoff = (Local::now() - Duration::days(i64::from(retention_days))).to_rfc3339();
 
-    let affected = conn.execute(
-        "DELETE FROM memories WHERE category = 'conversation' AND updated_at < ?1",
-        params![cutoff],
-    )?;
+    // Core memories use created_at so that frequently-recalled-but-never-rewritten
+    // entries are not pruned based on a stale updated_at (which only advances on write).
+    // Conversation and daily rows use updated_at — those categories are write-heavy and
+    // the distinction is immaterial.
+    let timestamp_col = if use_created_at {
+        "created_at"
+    } else {
+        "updated_at"
+    };
+    let sql = format!(
+        "DELETE FROM memories WHERE category = ?1 AND {} < ?2",
+        timestamp_col
+    );
+    let affected = conn.execute(&sql, params![category, cutoff])?;
 
     Ok(u64::try_from(affected).unwrap_or(0))
 }
@@ -373,8 +409,16 @@ fn prune_audit_entries(workspace_dir: &Path, retention_days: u32) -> Result<()> 
 
 fn memory_date_from_filename(filename: &str) -> Option<NaiveDate> {
     let stem = filename.strip_suffix(".md")?;
+    // Split on '_' first (handles YYYY-MM-DD_suffix.md).
     let date_part = stem.split('_').next().unwrap_or(stem);
-    NaiveDate::parse_from_str(date_part, "%Y-%m-%d").ok()
+    // If the date part has more than 10 chars (e.g. 2026-03-28-1442),
+    // take only the YYYY-MM-DD prefix.
+    let date_str = if date_part.len() > 10 {
+        date_part.get(..10)?
+    } else {
+        date_part
+    };
+    NaiveDate::parse_from_str(date_str, "%Y-%m-%d").ok()
 }
 
 fn date_prefix(filename: &str) -> Option<NaiveDate> {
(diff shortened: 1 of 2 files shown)


Audacity88 (changes requested):
Reviewed current head 3814524, including the diff, linked issue #7056, current labels/checks, the empty review/comment snapshots, and the current public PR template requirements. I did not run local validation.

I’m requesting changes before approval because this is `risk: high` memory-deletion behavior and the PR body is still in a pre-template shape.

The implementation direction looks right to me: keeping daily/core DB pruning opt-in with `0 = keep forever` preserves existing behavior, while giving long-running SQLite deployments the cleanup path requested in #7056. The category wiring also matches the existing SQLite category strings.

### 🟢 What looks good — The safer option from #7056 is the one implemented

I like that this chooses the opt-in config knob path rather than reusing `archive_after_days`/`purge_after_days` for DB rows. That keeps existing installations immortal-by-default for `core`/`daily` rows, while letting operators with large SQLite deployments choose explicit retention windows.

### 🔴 Blocking — Fill the current high-risk PR template before merge

Please update the PR body to the current template and include the required high-risk sections:

- Scope boundary and blast radius
- Current labels snapshot: `enhancement`, `size: S`, `risk: high`, `config`, `memory`, `memory:hygiene`, `memory:sqlite`
- Validation Evidence with the exact local commands run and tail output, especially the targeted hygiene/policy tests
- Security & Privacy Impact
- Compatibility, including the config-surface change and upgrade behavior for existing installs
- Rollback, including the fast rollback path, relevant config toggles, and observable failure symptoms

CI is green, but for a retention/deletion change the body needs to preserve the reviewer/operator evidence that CI cannot replace. In particular, the validation section should make clear that the new `daily_retention_days` and `core_retention_days` paths were exercised locally and that the default `0` behavior keeps existing daily/core rows.

### 🟡 Warning — Remove the bot attribution trailer from the commit tail

The second commit currently includes `Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>`. The current PR template asks contributors not to add bot/AI attribution footers to PR bodies or commit-message tails, so please remove that trailer before merge.

crates/zeroclaw-memory/src/hygiene.rs
```diff
@@ -332,8 +352,8 @@ fn prune_conversation_rows(workspace_dir: &Path, retention_days: u32) -> Result<
     let cutoff = (Local::now() - Duration::days(i64::from(retention_days))).to_rfc3339();
 
     let affected = conn.execute(
-        "DELETE FROM memories WHERE category = 'conversation' AND updated_at < ?1",
-        params![cutoff],
+        "DELETE FROM memories WHERE category = ?1 AND updated_at < ?2",
```
singlerider: 🔴 For `core` rows this keys deletion on `updated_at`, which only advances on write (upsert / explicit UPDATE), not on recall — the read queries `ORDER BY updated_at DESC` but never bump it. So a frequently-recalled-but-never-rewritten core memory gets a stale timestamp and is pruned despite being in active use. Document that recall doesn't refresh the clock, or use a use-reflecting signal for `core` (bump on recall, or prune core by `created_at` with a longer window).

singlerider (commented):
Reviewed at head `3814524` against current `master`. The implementation is clean and the opt-in design is right; agreeing with @Audacity88's template-completeness blocker and adding one substantive caution about pruning `core` rows by `updated_at`.

### 🟢 Opt-in retention is the safe shape

Defaulting `daily_retention_days` / `core_retention_days` to `0 = keep forever` keeps existing installs immortal-by-default while giving large SQLite deployments the cleanup path from #7056. Generalizing `prune_conversation_rows` into `prune_category_rows` with a parameterized `category = ?1` query (no string interpolation) is correct, the report/log fields are extended consistently, and the new tests cover the daily/core prune paths plus the `0`-keeps-everything default. The `memory_date_from_filename` fix for `YYYY-MM-DD-HHMM` suffixes is a reasonable correctness improvement.

### 🔴 Blocking — fill the high-risk PR template (per @Audacity88)

This deletes `core`/`daily` memory rows, so the body needs the full high-risk template: scope boundary and blast radius, the current label snapshot, Validation Evidence with the exact local commands and tail output for the hygiene/policy tests (explicitly showing the default-`0` path keeps existing core/daily rows), Security & Privacy, Compatibility (the config-surface addition and upgrade behavior), and Rollback (fast revert, the config toggles, and observable symptoms). CI cannot replace that evidence for a deletion change.

### 🔴 Blocking — `core` pruning keys on `updated_at`, which does not advance on recall

`prune_category_rows` deletes where `updated_at < cutoff`. From the SQLite backend, `updated_at` only advances on write (the `INSERT ... ON CONFLICT` upsert and the explicit `UPDATE ... SET updated_at`); the recall/read queries `ORDER BY updated_at DESC` but do not bump it. That means a `core` memory that is frequently *recalled* but never *rewritten* keeps a stale `updated_at` and would be pruned by `core_retention_days` even though it is actively in use. For `conversation`/`daily` rows that is acceptable, but `core` is the durable-by-intent category, so age-by-last-write is a surprising deletion criterion there. Please either document this clearly (so an operator setting `core_retention_days` knows recall does not refresh the clock), or key `core` pruning on a signal that reflects use (e.g. bump `updated_at` on recall for core, or prune by `created_at` with a much longer intended window). Since the feature is opt-in this is not catastrophic, but for a `risk: high` core-deletion path the criterion needs to be intentional and stated.

### 🟡 Agree — strip the bot co-author trailer

The second commit carries `Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>`. Remove that from the commit tail before merge.

Fill the template, make the `core` pruning criterion intentional/documented, and this is close. Leaving @Audacity88's changes-requested in force.

Audacity88:
@xianshishan Release-scope update: this PR is still marked `needs-author-action`, so I’m removing it from the active v0.8.1 milestone.

Audacity88 (changes requested):
Reviewed current head 3055ad24 against the live PR body, #7056, the prior reviews, current labels/checks, and the changed `schema.rs`, `hygiene.rs`, and SQLite upsert path. I did not run local validation. CI is green, and I verified the prior template and bot-attribution blockers are addressed in the current public artifacts.

Thanks for the update here. The opt-in shape is still the right direction for #7056: existing installs keep daily/core rows forever by default, and operators who need cleanup can now choose explicit retention windows.

### ✅ Resolved — The high-risk PR body is now filled in

The current body includes the scope boundary, blast radius, validation evidence, security/privacy section, compatibility notes, and rollback path. It also calls out the config-surface change and the `0 = keep forever` default, which were the main reviewer-evidence gaps in my earlier review.

### ✅ Resolved — The bot attribution trailer is no longer in the current commit tail

The current single commit no longer carries the Claude `Co-authored-by` trailer that was present in the older head.

### 🟢 What looks good — Daily/core pruning remains opt-in and tested

Keeping `daily_retention_days` and `core_retention_days` at `0` by default preserves the previous behavior for existing deployments. The new hygiene tests cover daily pruning, default-zero daily retention, core pruning by `created_at`, and the older `YYYY-MM-DD-HHMM.md` filename shape from #7056.

### 🔴 Blocking — Make the core retention clock description match the actual SQLite semantics

The current code now prunes `core` rows by `created_at`, and the SQLite upsert path does not update `created_at` on rewrites. It only updates `updated_at`. That means `core_retention_days` is an absolute age limit from the row's first write time. Recall does not refresh it, and a rewrite does not refresh it either unless the implementation explicitly changes `created_at`.

Several public surfaces still describe a different model:

- `crates/zeroclaw-config/src/schema.rs` says core age is measured by `created_at` but then says "only rewrites refresh the clock".
- `crates/zeroclaw-memory/src/hygiene.rs` says `created_at` protects frequently recalled but never rewritten core entries from stale `updated_at` pruning.
- The PR body and commit text repeat that a frequently recalled but never rewritten core memory would otherwise be pruned despite active use.

That explanation is backwards for the implemented behavior. Switching from `updated_at` to `created_at` does not protect active old core memories. It makes the deletion rule independent of both recall and later writes, so old core rows are pruned once their first-write timestamp crosses the configured cutoff. The test `prunes_old_core_rows_when_configured` proves this by deleting a core row whose `updated_at` is recent but whose `created_at` is old.

For a `risk: high` memory-deletion feature, the operator-facing contract needs to be exact before merge. Please update the schema doc, inline comment, and PR body to say plainly that core retention is based on first-write age, that neither recall nor ordinary rewrites refresh that clock under the current SQLite upsert, and that operators should set a large window or keep `core_retention_days = 0` for durable core memory.

xianshishan:
Thanks for the careful review @Audacity88 and @singlerider — I've addressed the feedback in the latest push (9d96dd0). Here's a summary of the changes and the design rationale.

## Changes since last review

### 1. Corrected `core_retention_days` documentation (schema.rs + hygiene.rs)

The doc comment in `schema.rs` and the inline comment in `hygiene.rs` previously described the `created_at` switch as protecting frequently-recalled-but-never-rewritten core entries from stale `updated_at` pruning. **That explanation was backwards.** Switching from `updated_at` to `created_at` does *not* protect active old core memories — it makes the deletion rule independent of both recall and later writes.

**What actually happens:** `created_at` is set once on first write. The current SQLite upsert (`INSERT ... ON CONFLICT ... DO UPDATE SET value = ..., updated_at = ...`) only bumps `updated_at`, never `created_at`. So `core_retention_days` is an **absolute age limit from first write** — once that timestamp crosses the cutoff, the row is pruned regardless of how frequently it was recalled or rewritten.

The test `prunes_old_core_rows_when_configured` proves this: it creates a core row with `created_at` set to 120 days ago and `updated_at` set to 1 day ago (simulating a recent rewrite), and the row is still pruned.

I've updated both doc comments to state plainly:
- Age is measured by `created_at` (first-write time)
- Neither recall nor ordinary rewrites refresh `created_at` under the current SQLite upsert
- Operators should set a generously large window or keep `core_retention_days = 0` for durable core memory

### 2. Design rationale: why `created_at` instead of `updated_at`

This was the key architectural decision from @singlerider's first review. The choice between the two timestamp columns is a tradeoff:

| Column | Prunes based on | Risk |
|--------|----------------|------|
| `updated_at` | Last write time | A frequently-**recalled** but never-**rewritten** core entry keeps a stale timestamp and gets pruned silently despite active use |
| `created_at` | First write time | A frequently-**rewritten** core entry will eventually be pruned even if still active, because rewrites don't refresh `created_at` |

Neither column perfectly captures "is this memory still in use." `updated_at` is misleading for read-heavy core entries. `created_at` at least has a clear, predictable semantic — the row lives for exactly N days from its first write, period.

I chose `created_at` because:
- It has **predictable semantics** — no hidden dependency on whether the user happened to rewrite vs. recall
- The operator can reason about it: "core entries written more than 90 days ago will be deleted"
- The default `0 = keep forever` means no existing deployment is affected
- For operators who enable it, the safe play is a generously large window (e.g. 90–180 days)

The alternative approach (bumping `updated_at` on recall for core rows) would require modifying the SQLite read/recall path and carries its own risk of turning reads into writes. That's a larger scope than this PR targets.

### 3. Summary of all changes in this PR (vs. original 3814524)

**Code:**
- Generalized `prune_conversation_rows()` → `prune_category_rows(category, use_created_at)` — parameterized query, no string interpolation
- Core pruning keys on `created_at`; conversation/daily key on `updated_at`
- `HygieneReport` gained `pruned_daily_rows` and `pruned_core_rows` counters, surfaced in logs and `memory_hygiene_state.json`
- `memory_date_from_filename()` now accepts `YYYY-MM-DD-HHMM.md` filenames from older capture flows
- Per-category overrides via `[memory.policy.retention_days_by_category]` work for all three categories

**Tests (11 hygiene + 5 policy):**
- `prunes_old_daily_rows_in_sqlite_backend` — daily pruning with configured retention
- `zero_retention_disables_daily_pruning` — default-0 keeps existing daily rows
- `prunes_old_core_rows_when_configured` — core pruning by `created_at`, including the edge case where `updated_at` is recent but `created_at` is old
- `date_from_filename_handles_hyphen_suffix` — `YYYY-MM-DD-HHMM.md` parsing

**Config surface (backward compatible):**
- `daily_retention_days` (default `0` = keep forever)
- `core_retention_days` (default `0` = keep forever)

Audacity88 (approved):
Reviewed current head `9d96dd0b0699503fbe5542cdd399c6cd2ce7c9b2` against the live PR body, linked issue #7056, the prior reviews/comments from Audacity88 and singlerider, the changed `schema.rs` and `hygiene.rs` paths visible in the diff, the current labels, and visible GitHub checks. I did not run local validation; this approval is based on source inspection from the diff, the author's focused validation evidence, and the green source/build/test CI on the PR head. One visible check named `main` was still pending at last read, so this is not a merge-ready statement.

Thanks for the follow-up here. The previous core-retention contract blocker from my 3055ad24 review is resolved in the current head and public body. This clears my prior requested-changes review; I do not see a remaining technical blocker in the current reviewer feedback.

### ✅ Resolved — Core retention now says first-write age clearly

The schema doc, hygiene inline comment, and PR body now agree with the implemented contract: `core_retention_days` uses `created_at`, the PR states that recall and ordinary rewrites do not refresh that timestamp, and operators should either keep `core_retention_days = 0` or choose a large window for durable core memory.

That matches the regression test shape too. `prunes_old_core_rows_when_configured` keeps `updated_at` recent while making `created_at` old, then verifies the row is still pruned, so the test now proves the documented contract instead of implying recall/write activity protects the row.

### 🟢 What looks good — The deletion path stays opt-in and observable

The safe default remains `0 = keep forever` for the new daily/core retention knobs, so existing installs do not start deleting daily or core rows just by upgrading. The new counters in `HygieneReport`, the log/state output, and the PR body's rollback section give operators a way to notice and disable overly aggressive retention without a code rollback.

No active code blocker from me. Before merge, still verify current `master` freshness for these config/memory files and wait for the remaining visible check to finish.

PR #7084, opened 2026-06-02
fix(platform): use raw_arg for Windows shell commands to preserve double quotes

### LINKED ISSUE #7083: [Bug]: Windows shell tool mangles commands containing double quotes (cmd.exe + std::process arg escaping)
## Affected component

runtime/daemon (shell tool — `RuntimeAdapter::build_shell_command`)

## Severity

S1 — workflow blocked (any agent shell command containing `"` quotes fails on Windows)

## Current behavior

On Windows, shell tool commands that contain double quotes — which the model emits constantly to quote paths/arguments — fail with errors like:

```
The filename, directory name, or volume label syntax is incorrect.
```

Observed session:

| # | Command (as emitted to the shell tool) | Result |
|---|---|---|
| 1 | `dir "C:\Users\<user>\Desktop" /b 2>nul \|\| echo "directory missing or inaccessible"` | falls through to the `echo` branch — `dir` silently failed |
| 2 | `echo %USERPROFILE%\Desktop` | ✅ works → `C:\Users\<user>\Desktop` |
| 3 | `dir "C:\Users\<user>\Desktop" /b` | ❌ `The filename, directory name, or volume label syntax is incorrect.` |
| 4 | `cmd /c "dir C:\Users\<user>\Desktop"` | ❌ `'"dir C:\Users\<user>\Desktop"' is not recognized as an internal or external command` |
| 5 | `cmd /c dir "C:\Users\<user>\Desktop"` | ❌ `The filename, directory name, or volume label syntax is incorrect.` |

The pattern: **commands without `"` work; commands containing `"` break.**

## Expected behavior

A command like `dir "C:\path with spaces" /b` should run exactly as if typed at a `cmd.exe` prompt, because the shell tool is documented as executing a shell command verbatim.

## Steps to reproduce

On a Windows host:

1. Start the daemon: `zeroclaw daemon` (or `zeroclaw start`).
2. From any connected channel, send these messages to the agent one after another, prompting it to run each command via the shell tool:
   1. `Run this command: echo %USERPROFILE%\Desktop` → succeeds (no quotes).
   2. `Run this command: dir "C:\Users\<user>\Desktop" /b` → fails with *The filename, directory name, or volume label syntax is incorrect.*
   3. `Run this command: dir "C:\Users\<user>\Desktop" /b 2>nul || echo "directory missing or inaccessible"` → the `dir` half silently fails and the `echo` fallback fires.

The decisive signal: every command that contains a `"` fails; the quote-free one works.

## Root cause

[`crates/zeroclaw-config/src/platform/native.rs`](../blob/master/crates/zeroclaw-config/src/platform/native.rs) (Windows branch of `build_shell_command`):

```rust
let mut process = tokio::process::Command::new("cmd.exe");
process
    .arg("/C")
    .arg(command)   // <-- the whole command string passed as one arg()
    .current_dir(workspace_dir)
    .creation_flags(CREATE_NO_WINDOW);
```

`std::process::Command` (wrapped by tokio) builds the Win32 command line using the **standard `CommandLineToArgvW` quoting/escaping rules**: an argument containing spaces gets wrapped in `"`, and any embedded `"` is escaped as `\"`. But **`cmd.exe` does not use those rules** — it treats `\` as a literal path separator, so the `\"` sequences are passed through to `dir`, which then sees malformed paths.

For `dir "C:\Users\u\Desktop" /b` the actual command line becomes roughly:

```
cmd.exe /C "dir \"C:\Users\u\Desktop\" /b"
```

`cmd` strips the outer pair but leaves the inner `\"` → `dir` receives `\"C:\Users\u\Desktop\"` → "syntax incorrect". Case 4 above is the same mechanism one level deeper: the nested quoted string becomes the "command name", hence *"... is not recognized as an internal or external command"*. Case 2 works only because it has no inner quotes.

## Suggested fix

Use `raw_arg` (from `std::os::windows::process::CommandExt`) so the command line is passed to `cmd.exe` verbatim, bypassing Rust's `argv`-style escaping:

```rust
#[cfg(target_os = "windows")]
{
    use std::os::windows::process::CommandExt;
    const CREATE_NO_WINDOW: u32 = 0x08000000;

    let mut process = tokio::process::Command::new("cmd.exe");
    process
        .raw_arg("/C")
        // cmd /C strips one surrounding pair of quotes and runs the rest as-is.
        .raw_arg(format!("\"{command}\""))
        .current_dir(workspace_dir)
        .creation_flags(CREATE_NO_WINDOW);
    Ok(process)
}
```

Windows-only regression tests covering quoted `echo`/`dir` arguments have been drafted in `crates/zeroclaw-config/src/platform/native.rs` (currently `#[ignore]`d because they fail against the present implementation; un-ignore once the fix lands).

## Environment

- OS: Windows 10/11
- Component: `zeroclaw-config` `NativeRuntime::build_shell_command`

## Related

- #6704 — Shell tool output garbled on Windows (non-UTF-8 code pages) — same file, decode side
- #6705 — cron execution hardcoded `sh`, bypassed RuntimeAdapter
- #5562 — Windows shell flashes a cmd console window
- #6410 — google_workspace mangles `--params` JSON via shell fallback (same escaping family)
- #5246 — Option to choose default shell (relates to the cmd→PowerShell discussion)

xianshishan:
## Summary

Closes #7083.

### Purpose

On Windows, any agent shell command containing double quotes — which models emit constantly to quote paths, arguments, and strings — fails with `The filename, directory name, or volume label syntax is incorrect.` This blocks virtually all real-world shell tool usage on Windows, since models routinely produce commands like `dir "C:\path with spaces" /b`.

### Root cause

`build_shell_command()` (Windows branch in `crates/zeroclaw-config/src/platform/native.rs:55-66`) passes the command string to `cmd.exe` via `.arg()`. Rust's `std::process::Command` applies `CommandLineToArgvW` escaping to each `.arg()` — embedded `"` characters become `\"` backslash-escapes. But `cmd.exe` does **not** understand this escaping: it treats `\"` as literal characters, producing a malformed command line that the target program (`dir`, `echo`, etc.) cannot parse.

Example: `dir "C:\Users\u\Desktop" /b` becomes:
```
cmd.exe /C "dir \"C:\Users\u\Desktop\" /b"
```
cmd strips the outer quotes but the inner `\"` passes through to `dir` → `"The filename, directory name, or volume label syntax is incorrect."`

### Fix

Use `raw_arg()` from `std::os::windows::process::CommandExt` instead of `.arg()`, which passes arguments verbatim to the child process with **zero** escaping. The command is wrapped in one pair of quotes (standard `cmd /C` convention: cmd strips them and runs the inner text as-is):

```rust
process
    .raw_arg("/C")
    .raw_arg(format!("\"{command}\""))
```

This preserves all embedded double quotes, redirect operators (`2>nul`), pipe operators (`||`), and `%VAR%` expansions exactly as the model emits them.

### Files changed

| File | Change |
|---|---|
| `crates/zeroclaw-config/src/platform/native.rs` | +139/-2 — `raw_arg` fix + 6 new tests |

### Tests added

| Test | Scope | What it verifies |
|---|---|---|
| `shell_command_preserves_double_quotes` | Cross-platform | Debug output contains literal `"C:\Users\test\Desktop"`, no `\"` escapes |
| `shell_command_preserves_mixed_quoted_unquoted` | Cross-platform | Mixed quoted/unquoted args, redirects, pipes survive intact |
| `shell_command_no_quotes_still_works` | Cross-platform | Regression: unquoted commands still pass through correctly |
| `windows_echo_quoted_argument_succeeds` | Windows only | `echo "hello world"` executes and outputs correctly |
| `windows_dir_quoted_path_succeeds` | Windows only | `dir "C:\Windows" /b` executes and lists directory contents |
| `windows_echo_percent_expansion_preserved` | Windows only | `echo %USERPROFILE%` expands to a real path |

### Expected outcome

- Shell tool commands containing double quotes work correctly on Windows
- Commands without quotes continue to work (no regression)
- `%VAR%` expansions, redirects, and pipes are preserved verbatim
- Non-Windows platforms are unaffected (unchanged code path)

## Test plan

- [ ] On a Windows host: `cargo test -p zeroclaw-config --lib platform::native` — all tests pass
- [ ] Manual: run `dir "C:\path with spaces" /b` via the shell tool — succeeds without "syntax incorrect" errors
- [ ] Manual: run `echo "hello world"` via the shell tool — outputs `hello world`

### DIFF
--- a/crates/zeroclaw-config/src/platform/native.rs
+++ b/crates/zeroclaw-config/src/platform/native.rs
@@ -54,12 +54,17 @@ impl RuntimeAdapter for NativeRuntime {
 
         #[cfg(target_os = "windows")]
         {
+            use std::os::windows::process::CommandExt;
             const CREATE_NO_WINDOW: u32 = 0x08000000;
 
             let mut process = tokio::process::Command::new("cmd.exe");
+            // Use raw_arg so the command string is passed verbatim to cmd.exe,
+            // bypassing Rust's CommandLineToArgvW escaping which would mangle
+            // embedded double quotes with backslash escapes that cmd doesn't
+            // understand (see #7083).
             process
-                .arg("/C")
-                .arg(command)
+                .raw_arg("/C")
+                .raw_arg(format!("\"{command}\""))
                 .current_dir(workspace_dir)
                 .creation_flags(CREATE_NO_WINDOW);
             Ok(process)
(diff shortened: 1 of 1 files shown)


singlerider (approved):
Reviewing at `42fd7c9d`. CI 13/13 ✅ — including the Windows build leg and the `#[cfg(target_os = "windows")]` end-to-end tests, so the fix is validated on a real Windows runner, not just by inspection. Closes #7083.

Thank you for this, @xianshishan. This is a well-diagnosed S1 — Windows shell usage is effectively broken for any command with quotes, which models emit constantly, and the root-cause writeup (Rust's `CommandLineToArgvW` escaping vs. cmd.exe's parser) is exactly right.

### 🟢 The fix is correct and is the standard idiom

`raw_arg("/C")` + `raw_arg(format!("\"{command}\""))` is the documented way to hand a command line to `cmd.exe` verbatim from Rust — `raw_arg`'s own docs cite `cmd /C` as the motivating case. The prior `.arg(command)` applied per-argument `CommandLineToArgvW` escaping, turning embedded `"` into `\"`, which cmd passes through literally to the target program → the "syntax is incorrect" failure. `raw_arg` bypasses that, and the single outer quote pair is the conventional `cmd /C "<text>"` wrapper that cmd strips before executing the inner text. Redirects (`2>nul`), pipes (`||`), and `%VAR%` expansion all survive because nothing re-escapes them.

### 🟢 No change to the trust boundary

I looked at this specifically because of the `risk: high` / `tool:shell` labels. The design before and after is identical: `cmd.exe /C <full command string>`, where the string originates from the model and is gated upstream by `SecurityPolicy`. cmd always parsed the whole string as a command line — `raw_arg` does not introduce a new injection vector, it only stops the mangling. The surface that existed (arbitrary command execution, policy-gated) is unchanged; this PR neither widens nor narrows it. So the high-risk classification is the auto-labeler reacting to the shell path, not a real elevation here.

### 🟢 Scope is correct — only the Windows native branch

I checked for sibling shell-command builders that might need the same treatment. The Docker adapter (`platform/docker.rs`) builds an `sh -c` line for a Linux container and the wasm adapter is sandboxed — neither touches `cmd.exe`. `native.rs`'s Windows branch is the only place this bug lives, and it's the only place the fix lands. The non-Windows code path is byte-for-byte unchanged.

### 🟢 Test coverage maps the failure precisely

The split is right: the three cross-platform tests assert on the `Debug` representation (quotes preserved, no `\"` backslash-escapes) so they run everywhere, and the three `#[cfg(target_os = "windows")]` tests actually invoke `cmd /C` and assert real output — `echo "hello world"`, `dir "C:\Windows" /b`, and `%USERPROFILE%` expansion. The `shell_command_no_quotes_still_works` regression guard covers the unquoted path. That's the right shape: prove the escaping behavior portably, prove execution on the platform that has the bug.

One forward-looking note, non-blocking and not something to change here: the inherent `cmd /C` first-/last-quote stripping rule means a pathological command that itself starts and ends with a quote could still confuse cmd's parser — but that's a property of `cmd /C` itself, not a regression this PR introduces, and it's strictly better than the prior unconditional mangling. Worth a mention only so the next person knows the boundary; nothing to do.

Clean, focused, correctly scoped, and validated on Windows CI. Approving.

Audacity88 (approved):
I reviewed current head 42fd7c9d against #7083, the Windows native shell adapter diff, the current PR conversation, and the green CI run, including the Windows build and Windows-only command execution tests. I am approving.

### 🟢 What looks good — Windows command construction now matches `cmd.exe`

The failure mode in #7083 was the mismatch between Rust's argv-style escaping and `cmd.exe`'s command-line parser. The Windows branch now uses `raw_arg("/C")` plus one outer quoted command string, so the command text reaches `cmd.exe` in the grammar `cmd.exe` expects instead of being rewritten with backslash-escaped inner quotes. The Unix `sh -c` path is unchanged.

### 🟢 What looks good — The trust boundary is unchanged

This still runs the already-authorized shell command through `cmd.exe /C`; it does not add a new execution path or relax the upstream tool policy. For a high-risk shell surface, that is the right kind of fix: it corrects command transport without expanding what the shell tool is allowed to execute.

### 🟢 What looks good — Regression coverage matches the bug

The portable tests assert that quoted command text is preserved, while the Windows-only tests execute the important cases on real `cmd.exe`: quoted `echo`, quoted `dir`, `%USERPROFILE%` expansion, and the unquoted regression path. With the full CI run green at this head, the evidence covers both the original quoted-path failure and the main no-quotes compatibility case.

No blocking findings from me.

PR #7085, opened 2026-06-02
fix(config): restore 1Password op secret refs

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Restores support for `op://vault/item/field` references in config secret fields that was lost after PR #4796 was reverted in the #6074 recovery set.
  - Resolves `op://` references through `op read` in the same secret-decryption paths that already handle `enc2:` and legacy `enc:` values.
  - Preserves configured `op://` references on normal save/encrypt cycles, while still allowing an intentional dirty edit to replace the reference with a locally encrypted secret.
  - Adds a timeout around `op read` so a stuck 1Password CLI call fails with an actionable error instead of hanging config load.
  - Updates provider credential docs to describe inline, encrypted, env override, and `op://` storage forms without implying env overrides rewrite stored config.
- **Scope boundary:** This PR does not add 1Password authentication, a new credential store, or provider-specific fallback env vars. It only restores the `op://` reference form for existing config secret fields.
- **Blast radius:** `zeroclaw-config` secret load/save/encryption paths and provider credential docs. Existing inline, encrypted, and schema-mirror env override credential behavior should remain unchanged.
- **Linked issue(s):** Related #6074. Supersedes #4796.
- **Labels:** `bug`, `risk: high`, `size: M`, `config`, `docs`, `security`

  - `cargo test -p zeroclaw-config op_reference -- --nocapture` - passed. Tail: `test result: ok. 3 passed; 0 failed`.
  - `cargo test -p zeroclaw-config onepassword_reference -- --nocapture` - passed. Tail: fake-`op` load/save preservation test passed.
  - `cargo test -p zeroclaw-config dirty_onepassword_secret_edit_replaces_reference -- --nocapture` - passed. Tail: dirty secret edit replacement test passed.
  - `cargo test -p zeroclaw-config secret_field_tests -- --nocapture` - passed. Tail: `test result: ok. 14 passed; 0 failed`.
  - `cargo test -p zeroclaw-config is_encrypted_detects_prefix -- --nocapture` - passed. Tail: encrypted-prefix detection test passed.
  - `cargo test -p zeroclaw-config secrets::tests -- --nocapture` - passed. Tail: `test result: ok. 44 passed; 0 failed`.
  - `cargo test -p zeroclaw-config encryption_preserves_onepassword_secret_references -- --nocapture` - passed. Tail: migration preservation test passed.
  - `cargo fmt --all -- --check` - passed. Tail: command exited cleanly.
  - `cargo clippy -p zeroclaw-config --all-targets -- -D warnings` - passed. Tail: clippy completed with no warnings promoted to errors.
  - `git diff --check` - passed. Tail: no whitespace errors reported.
  - `cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings` - passed. Tail: workspace clippy completed cleanly.
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` - passed. Tail: 7446 tests passed, 10 skipped.
- **Beyond CI - what did you manually verify?** The fake-`op` test covers the important runtime path: a config containing `op://vault/item/field` resolves to the CLI-provided secret in memory, then saves back to disk with the `op://` reference still present and the resolved secret absent. I did not test against a real signed-in 1Password account.

- If any `Yes`, describe the risk and mitigation: Config secret fields can now invoke the local 1Password CLI via `op read` when the operator explicitly stores an `op://` reference. The call is limited to that exact reference, has a 10 second timeout, does not log resolved secret values, and preserves the external reference on disk instead of migrating it into local ciphertext during normal saves.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: Existing inline, encrypted, and schema-mirror env override credentials keep their current behavior. Users can additionally store supported secret fields as `op://vault/item/field` references if they have the 1Password CLI installed and signed in.

- **Feature flags or config toggles:** None. Operators can remove `op://` references from config and use inline, encrypted, or env override credentials instead.
- **Observable failure symptoms:** Config load or provider startup errors mentioning invalid `op://` references, missing `op`, failed `op read`, or `op read timed out after 10s`.

  - #4796 by @bka9
- Scope materially carried forward: Restores the merged-but-reverted 1Password `op://` secret-reference capability, including secret-store resolution, validation, preservation across save/encrypt paths, and docs.
- If `No`, why (inspiration-only, no direct code/design carry-over): No direct code from #4796 was copied into this branch. This PR restores the same public capability and credits #4796/@bka9 in this supersede section, but the implementation was rebuilt for the current typed config, schema-mirror env override, and save-path preservation architecture.


singlerider (changes requested):
Thanks for picking this back up, @Audacity88 — restoring `op://` references the right way (external reference preserved on disk, resolved only in memory) is exactly the behavior operators want, and the snapshot/restore design mirrors the existing `pre_override_snapshots` flow so it reads as a natural extension rather than a bolt-on. I read the full diff and cross-checked it against the secret load/save paths in `zeroclaw-config`. One blocker before this lands.

### 🔴 Blocking — `op read` runs synchronously on the async runtime

`resolve_onepassword_ref` shells out to `op` with a blocking `std::process::Command` and a `std::thread::sleep(25ms)` poll loop, and it's reached from `decrypt_secrets()` inside `async fn load_or_init` (`crates/zeroclaw-config/src/schema.rs:14510`). The 10s timeout you added correctly anticipates that `op` can stall (locked vault, expired session waiting on biometric, hung CLI) — but the mitigation itself blocks the tokio worker thread for up to ten seconds, not just the config load. A stuck `op` stalls the executor.

Resolve via `tokio::process::Command` under `tokio::time::timeout`, or wrap the sync body in `tokio::task::spawn_blocking`. The async-`Command::output()`-under-`timeout` form also removes a second latent issue in the current loop: it breaks on process *exit* and then drains stdout/stderr, so if `op` ever wrote more than the OS pipe buffer (~64KB) before exiting it would block on the write, `try_wait()` would never see the exit, and the timeout would fire on a successful read. Secrets are tiny so it won't bite today, but draining concurrently with the wait closes the class of bug, and it falls out naturally from the same fix. Given this is the high-risk config-load path, I'd rather see it off the executor before merge than deferred.

### 🟢 What looks good

- The reference-validation step (`validate_onepassword_ref`) before spawning is the right guard: it enforces `op://vault/item/field`, and because the reference is passed as a positional arg (`["read", reference]`) it can't be reinterpreted as an `op` flag. Good injection posture.
- Error messages are genuinely actionable — the not-signed-in / session-expired hint and the empty-value case will save operators a support round-trip.
- The fake-`op` tests cover the load→save round-trip (reference preserved, resolved secret absent from disk) and the dirty-edit-replaces-reference path, including asserting the replacement is encrypted at rest. That's the behavior that actually matters, tested at the right level.
- stdin is nulled on the spawn, so `op` can't reach for terminal input.

Happy to re-review promptly once the `op read` call is off the executor thread.

Audacity88:
@singlerider I pushed the blocker fix and the current-master refresh for this PR.

What changed:
- `055c63172` moves generated secret decryption during config load onto `tokio::task::spawn_blocking`, so `op://...` resolution no longer blocks Tokio worker threads.
- The `op` resolver now drains stdout and stderr concurrently while polling the timeout, so a noisy or erroring `op` process cannot deadlock on a full pipe before the timeout path runs.
- `a945410bf` resolves the post-master conflict. The only manual conflict was in `crates/zeroclaw-config/src/env_overrides.rs`; the resolution keeps master's value-aware `raw_value_for_path` lookup for hyphenated aliases/model names and keeps this PR's `pub(crate)` visibility so the 1Password snapshot path can reuse it.

Local validation is green for `env_overrides`, `onepassword_reference`, `op_reference_drains_stderr_while_waiting`, `onepassword_reference_load_does_not_block_runtime_worker`, `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-config --all-targets -- -D warnings`, and `git diff --check`.

The branch is no longer dirty. It is still blocked on the active `CHANGES_REQUESTED` review while the fresh CI run finishes on `a945410bf`.

singlerider (approved):
Re-review at head a945410bf. Prior review requested one change; verified resolved in code.

### ✅ Resolved — `op read` is off the executor thread

055c63172 wraps `decrypt_secrets()` in `tokio::task::spawn_blocking` inside `load_or_init`, so `op://` resolution can no longer stall a tokio worker for the duration of the 10s timeout. The resolver also now drains stdout and stderr on dedicated reader threads concurrently with the `try_wait` poll loop — which closes the second latent issue I flagged: a child writing more than the pipe buffer before exit can no longer deadlock against the timeout. Both halves of the blocker addressed, and `onepassword_reference_load_does_not_block_runtime_worker` (single-worker multi-thread runtime + 1s-sleeping fake `op`) pins the responsiveness contract so it can't silently regress.

### 🟢 What looks good — the conflict resolution kept both sides' intent

The `env_overrides.rs` resolution preserves master's value-aware `raw_value_for_path` lookup for hyphenated aliases while keeping this PR's `pub(crate)` visibility for the snapshot path — exactly the right merge, and the validation evidence names the focused tests that prove it.

Reference validation before spawn (positional arg, no flag reinterpretation), nulled stdin, actionable signed-out/expired hints, and the fake-`op` round-trip tests (reference preserved on disk, resolved secret never written, dirty-edit replacement encrypted at rest) were all solid in the first pass and are unchanged.

CI is green on the new head. Approving.

PR #7091, opened 2026-06-02
fix(cron): expose DingTalk in cron delivery schemas

### LINKED ISSUE #7088: bug(cron): DingTalk is missing from cron delivery tool schemas
## Summary

DingTalk is implemented as a channel, but the `cron_add` and `cron_update` tool schemas still omit `dingtalk` from the `delivery.channel` enum on current `master`.

Related:

- #6521

## Affected component

Runtime cron tools / channel delivery schema.

## Severity

S2 - degraded behavior.

## Current behavior

Current `master` includes DingTalk channel support, and cron delivery validation no longer hard-rejects channels through the older runtime allowlist shape. However, the agent-facing tool schemas for `cron_add` and `cron_update` still enumerate delivery channels without `dingtalk`.

That leaves the tool surface inconsistent: a configured DingTalk channel can exist, but the cron tools do not advertise DingTalk as a valid delivery target.

## Expected behavior

Cron delivery schemas should include DingTalk wherever cron announce delivery is meant to support configured channel types.

The schema should also avoid drifting between `cron_add`, `cron_update`, and runtime validation.

## Steps to reproduce

```bash
# Inspect the cron_add and cron_update tool schemas on current master.
# delivery.channel includes telegram, discord, slack, mattermost, matrix, qq, webhook, lark, and feishu, but not dingtalk.
```

Expected: `dingtalk` appears in both cron delivery channel enums.

Current behavior from source inspection: `dingtalk` is omitted from both tool schemas.

## Version / SHA

Current `master` as checked during stale PR recovery: `599bd7ecb1993182e9c6e197e385fe758508eb6b`.

## Operating system

All OSes.

## Regression?

Unknown.

## Impact

Users and agents trying to schedule DingTalk announcements get an incomplete tool contract. Even where lower-level storage/validation no longer blocks the value, the schema pushes callers away from the supported channel.

## Acceptance criteria

- `cron_add` and `cron_update` expose `dingtalk` in `delivery.channel`.
- The supported delivery-channel list is not duplicated in ways that can drift silently.
- Focused tests cover DingTalk presence and schema parity between add/update paths.

## Pre-flight checks

- Maintainer source check was done against current `master`.
- No secrets, personal data, or private logs are included here.

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds `dingtalk` to the cron delivery channel schema used by both `cron_add` and `cron_update`, so the tool contract matches the implemented DingTalk channel support.
  - Replaces duplicated cron delivery schema enums with one shared runtime constant to keep add/update schemas from drifting.
  - Keeps implicit `cron_add` delivery defaults on a narrower direct-reply allowlist and adds DingTalk there, while preserving webhook's explicit `to` + `thread_id` delivery shape.
  - Removes the stale hard-coded channel list from the `cron_add` description so the schema enum remains the authoritative tool-surface list.
- **Scope boundary:** This PR does not add a new DingTalk channel implementation, change cron job storage, change delivery validation, or change webhook threading behavior.
- **Blast radius:** Runtime cron tool schemas and the channel-originated `cron_add` convenience default path. Existing persisted jobs and explicit delivery payloads keep their current behavior.
- **Linked issue(s):** Closes #7088
- **Labels:** `bug`, `risk: high`, `size: S`, `agent`, `runtime`, `cron`, `tool`, `channel:dingtalk`

- **Commands run and public-safe result digest:**

```text
cargo test -p zeroclaw-runtime delivery_schema --lib -- --nocapture
Result: passed. Covered cron_add delivery schema tests, including exact schema enum equality and webhook thread_id exposure.

cargo test -p zeroclaw-runtime add_and_update_delivery_channel_schemas_match --lib -- --nocapture
Result: passed. Covered exact parity between cron_add and cron_update delivery.channel schema enums and explicit DingTalk presence.

cargo test -p zeroclaw-runtime patch_schema_covers_all_cronjobpatch_fields_and_schedule_is_oneof --lib -- --nocapture
Result: passed. Covered cron_update patch schema shape and exact delivery.channel enum equality.

cargo test -p zeroclaw-runtime cron_delivery_defaults --lib -- --nocapture
Result: passed. Covered DingTalk auto-default injection and verified webhook is not guessed without sender/thread context.

cargo fmt --all -- --check
Result: passed.

Result: passed.

Temporary combined validation stack: current master + this PR + #7095
Result: passed.

cargo fmt --all -- --check
Result: passed.

scripts/zc-cargo-broad clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings
Result: passed.

scripts/zc-cargo-broad --lane nextest nextest run --locked --workspace --exclude zeroclaw-desktop
Result: passed. 7444 tests passed; 10 skipped.
```

- **Beyond CI — what did you manually verify?** The diff was checked against #7088's acceptance criteria: both cron tool schemas expose `dingtalk`, the schema enum is shared rather than duplicated, add/update parity is tested directly, and webhook delivery remains explicit because it needs separate sender/thread routing context. I also merged this PR locally with #7095 over current `master` and ran the repo-shaped broad clippy and nextest gates on that combined stack.
- **If any command was intentionally skipped, why:** I did not run a live DingTalk cron delivery because this change updates schema/defaulting behavior only; it does not change the DingTalk channel implementation, cron storage, or network delivery path. GitHub Quality Gate is green for this PR, including Lint, Build, Check, Benchmarks Compile, Test, Security, and CI Required Gate.

- If any `Yes`, describe the risk and mitigation: None. This changes cron tool schema metadata and a narrow channel-originated defaulting helper. The helper deliberately does not auto-create webhook delivery payloads because webhook delivery needs explicit `to` and `thread_id` routing context.

## Compatibility (required)

- If `No` or `Yes` to either: The agent-facing cron tool schema now advertises `dingtalk` as a valid `delivery.channel` value. Existing configs, jobs, CLI commands, and explicit delivery payloads do not need migration.

- **Fast rollback command/path:** Revert the PR merge commit.
- **Observable failure symptoms:** `cron_add` / `cron_update` schemas no longer show the expected channel enum, DingTalk scheduled announcements are not offered as a valid cron delivery target, or channel-originated `cron_add` calls unexpectedly inject or omit delivery defaults. Webhook regressions would show up as scheduled webhook deliveries missing explicit thread routing.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded contributor work is incorporated.


crates/zeroclaw-runtime/src/cron/mod.rs
```diff
@@ -23,6 +23,21 @@ pub use types::{
     deserialize_maybe_stringified,
 };
 
+/// Channel names exposed by the cron tool schemas. Actual runtime delivery is
+/// provided by the registered channel delivery handler, not this static enum.
+pub(crate) const CRON_DELIVERY_SCHEMA_CHANNELS: &[&str] = &[
```
singlerider, line 28: 🟢 One shared constant feeding both `cron_add` and `cron_update` schemas, with `add_and_update_delivery_channel_schemas_match` pinning parity, is the right fix — it kills the two-list drift that let DingTalk fall out of the schema in the first place. The doc comment correctly notes this is a curated subset and the real delivery comes from the registered handler. Verified the `"dingtalk"` arm exists in the actual delivery path (orchestrator/mod.rs:5606), so the schema now advertises a channel that genuinely lands. The residual cross-crate gap (this constant in zeroclaw-runtime vs the deliverable set in zeroclaw-channels) is pre-existing and worth a follow-up to assert one against the other, but out of scope for this bug.

singlerider (approved):
Read the full diff, the new `CRON_DELIVERY_SCHEMA_CHANNELS` / `AUTO_DELIVERY_DEFAULT_CHANNELS` constants, and the actual cron delivery dispatch in `zeroclaw-channels` (the channel-construction match at orchestrator/mod.rs:5335 and the narrower `deliver_announcement` at 8743). The fix is correct and moves the duplication in the right direction. Approving.

### 🟢 What looks good — collapses the two drifted schema enums into one drift-protected constant

The core of #7088 is that DingTalk channel delivery worked but the cron tool schema never advertised it. The right fix wasn't just "add dingtalk to two lists" — it's that `cron_add.rs` and `cron_update.rs` each carried their own hand-maintained `["telegram","discord",...]` enum that could (and did) drift. Replacing both with `cron::CRON_DELIVERY_SCHEMA_CHANNELS` and adding `add_and_update_delivery_channel_schemas_match` to assert byte-for-byte parity between the two tool schemas is exactly the move: the next channel is one constant edit, and the parity test fails loudly if anyone re-introduces a divergent literal. Stripping the stale hard-coded channel list out of the `cron_add` description so the enum stays the single tool-surface source of truth is the same instinct applied to the prose.

### 🟢 What looks good — the auto-default allowlist is correctly a narrower, separate set

`AUTO_DELIVERY_DEFAULT_CHANNELS` (the implicit channel-originated `cron_add` reply default) being a deliberately narrower 6-channel set than the 10-channel schema enum is right, not an oversight. Webhook/qq/lark/feishu need explicit `to`/`thread_id` routing context that can't be safely guessed from `reply_target`, and `cron_delivery_defaults_do_not_guess_webhook_shape` pins that webhook is never auto-injected. The two constants answer different questions (what the schema *offers* vs what the convenience path *auto-fills*), so keeping them distinct is correct.

### 🟢 What looks good — verified the fix is functionally complete, not just cosmetic

I confirmed DingTalk cron delivery actually lands: the authoritative delivery path is the channel-construction match at orchestrator/mod.rs:5335, which has a real `"dingtalk"` arm (5606, feature-gated with a clean not-configured bail), so advertising it in the schema connects to a working handler rather than producing an `unsupported delivery channel` at runtime. The validation evidence and the focused tests (exact enum equality, dingtalk presence, webhook-not-guessed) line up with what shipped.

### 🟢 One thing to keep an eye on — the schema constant still lives a crate away from the dispatch

The remaining residual is structural and pre-existing, not introduced here: `CRON_DELIVERY_SCHEMA_CHANNELS` (in `zeroclaw-runtime`) is a curated subset that must stay in sync with the actual deliverable set in `zeroclaw-channels` (the 5335 construction match, plus the narrower `deliver_announcement` arms at 8743, which notably handle signal/wechat/wecom_ws that the schema intentionally omits). Nothing pins the schema list against the dispatch across that crate boundary, so a future channel that wires cron delivery could still be invisible to the schema, or vice versa. This PR genuinely reduces the fragmentation (two runtime lists to one, drift-tested), and fully unifying the cross-crate set is a larger `risk: high` refactor well outside this bug — so it's not a blocker. Worth a tracked follow-up to assert the runtime schema constant against the channels-crate deliverable set so the two crates can't drift silently.

Approving. Correct fix for #7088, well-tested, and it leaves the cron schema surface less duplicated than it found it.

PR #7116, opened 2026-06-02
docs(providers): add OpenAI Codex over a ChatGPT subscription

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds a Model Providers book page — **OpenAI Codex over a ChatGPT subscription** (`docs/book/src/providers/openai-codex-subscription.md`) + a SUMMARY entry. The `openai` slot can authenticate against the Codex backend via a ChatGPT subscription (`requires_openai_auth = true` + `wire_api = "responses"`, OAuth from `~/.codex/auth.json`) instead of a metered API key, but nothing in the book documented the slot config, the served-model constraint, the cost framing, or the auth gotchas — so users hand-roll it and hit the same walls (alias `400`s, non-portable encrypted profiles, refresh-token rotation backoff).
- **Scope boundary:** Docs-only. No code, config schema, CLI surface, or runtime behavior changes.
- **Blast radius:** None at runtime. New doc page + one `SUMMARY.md` entry under Model Providers; no existing page rewritten.
- **Linked issue(s):** Related to the OpenAI Codex subscription provider slot. None fully closed by this PR.
- **Labels:** `type: docs`, `risk: low`, `size: M`, `docs`

Docs-only change → markdown lint + link-integrity gate (`scripts/ci/docs_quality_gate.sh`).

```bash
$ bash scripts/ci/docs_quality_gate.sh
Linting docs files: docs/book/src/SUMMARY.md docs/book/src/providers/openai-codex-subscription.md
markdownlint-cli2 v0.20.0 (markdownlint v0.40.0)
Finding: docs/book/src/SUMMARY.md docs/book/src/providers/openai-codex-subscription.md !target/** !docs/book/src/SUMMARY.md
Linting: 1 file(s)
Summary: 0 error(s)
# exit 0
```

- **Commands run and tail output:** above — `docs_quality_gate.sh` exits `0`, `0 error(s)`.
- **Beyond CI — what did you manually verify?** Cross-links (`configuration.md`, `catalog.md`, `routing.md`, `ops/cost-tracking.md`, the OAuth anchor) resolve; served-model strings, the `requires_openai_auth` / `wire_api = "responses"` dispatch, and the `auth login --import` flow checked against the codebase and a live `codex/models` query; the quoted `enc2:` decryption error aligned verbatim to the runtime string in `crates/zeroclaw-config/src/secrets.rs:160` so a copy-paste grep matches. Did NOT verify per-model unlimited request counts (no authoritative source — called out in the doc).
- **If any command was intentionally skipped, why:** `cargo fmt` / `cargo clippy` / `cargo test` skipped — no Rust changed; the docs gate is the applicable battery per the template.

- New external network calls? **No** (the doc describes the existing OAuth flow; nothing in the diff makes calls)
- Secrets / tokens / credentials handling changed? **No** (documents the existing `~/.codex/auth.json` import; no secret material in the diff)

## Compatibility (required)

- Upgrade steps for existing users: None — docs-only.

## Rollback

Low-risk: `git revert <sha>`.


singlerider (dismissed):
PR body does not match required template.

tidux (commented):
Read the diff end-to-end and cross-checked every load-bearing claim against the local source. This is a strong, operationally honest doc — it covers the exact pain users hit on this slot today (alias `400`s, encrypted-profile non-portability, refresh-token rotation backoff) and frames the subscription's "flat-rate but finite" cost model correctly. Not approving because @singlerider's `CHANGES_REQUESTED` on the template is live and the call there is correct — but everything else here is in good shape, and the substantive findings below are minor.

### 🟢 What looks good — cost framing names the real constraint

The "Cost and routing" section gets the subscription/metered split right where it matters: not "$0 = free" but "$0 per call, with a finite 5-hour allowance you can exhaust." That's the operational model, and the table + "routing is about spending that allowance deliberately and keeping a paid path for when it runs out" prose teaches the principle, not just the verdict. The metered-fallback paragraph correctly punts environment-specific provider lists to the operator's own routing config rather than inventing a list to maintain.

### 🟢 What looks good — served-ID-only warning is the right shape

The "Models" section calls out the exact failure mode (`gpt-5` / `instant` etc. fail `400`) and pairs it with the `codex/models` enumeration endpoint instead of a hardcoded list. The accompanying "treat the served catalog as volatile — verify against the endpoint before pinning, including this one" caveat is the right disclaimer for a table that will go stale.

### 🟢 What looks good — refresh-token rotation gotcha matches reality

The "one owner per account" pattern is correct against `crates/zeroclaw-providers/src/auth/mod.rs:222` — that's the exact backoff path. The recommended workflow (refresh owner publishes raw `auth.json`, every other host re-imports locally to re-encrypt under its own `.secret_key`) is also the right mental model for the encryption boundary in `crates/zeroclaw-config/src/secrets.rs`.

### 🔴 Blocking — PR template sections missing

@singlerider already flagged this and the call is correct: `.github/pull_request_template.md` marks `## Summary`, `## Validation Evidence`, `## Security & Privacy Impact`, and `## Compatibility` as `(required)`. The current body has `## What / ## Why / ## What's in it / ## Checks` — the content roughly maps (What ≈ Summary, Checks ≈ Validation Evidence) but the named sections aren't there, which means the merge gates the template encodes (security/privacy answers, compatibility answers, blast radius / scope boundary, linked-issue framing) aren't being explicitly answered.

For a docs-only PR most of these are short — `No` across the security/privacy block, `Yes` for backward compatibility, `git revert <sha>` for rollback — but they need to be on the record. Suggest porting the existing What/Why content into the template structure and filling the required sections rather than rewriting from scratch.

### 🟡 Warning — quoted `enc2:` error text doesn't match the runtime string

The "Two things that bite" section quotes the decryption failure as `enc2: decryption failed — wrong .secret_key`. The actual error in `crates/zeroclaw-config/src/secrets.rs:160` is:

```
enc2: decryption failed (wrong `.secret_key` or tampered ciphertext)
```

Two differences operators will care about when grepping logs against this doc: parentheses (not an em dash), and the `or tampered ciphertext` half is missing — that second clause matters because it's the same error path for two different operational causes (foreign profile vs. corrupt blob). Worth aligning the quoted string to the runtime form so a copy-paste grep actually matches.

### 🟡 Warning — `auth.json` foreign-profile move is one-shot, worth a `2>/dev/null` note

```bash
mv ~/.zeroclaw/auth-profiles.json ~/.zeroclaw/auth-profiles.json.foreign 2>/dev/null
```

The `2>/dev/null` is already there and that's good — but on a host where the operator has already done this once, the `.foreign` file will exist and the next run silently clobbers nothing (the source file is gone). Not actually a bug in the snippet; just worth a one-line note that the move is a "do this once if a foreign profile is present, then import" step rather than something to put in a cron / re-run script. Optional.

### 🔵 Suggestion — pre-empt the `catalog.md` `OPENAI_API_KEY` inconsistency

Not this PR's problem to fix, but flagging it so the doc reader has the right model: `docs/book/src/providers/catalog.md:42` (existing) currently says credentials for the Codex subscription variant come from "`OPENAI_API_KEY` / `~/.codex/auth.json`", which contradicts this PR's clearer "no `OPENAI_API_KEY`, no per-token spend — OAuth from `~/.codex/auth.json`" framing. The Codex subscription path does not read `OPENAI_API_KEY` — the typed alias dispatch in `crates/zeroclaw-providers/src/lib.rs` and the auth load in `crates/zeroclaw-providers/src/openai_codex.rs:1142` only touch the stored profile. A short follow-up to tighten `catalog.md` (separate PR, separate concern) would make the two pages consistent. Happy to file a tracking issue if useful.

### 🔵 Suggestion — `codex/models` enumeration could show one concrete invocation

The doc gives the URL shape (`https://chatgpt.com/backend-api/codex/models?client_version=<v>`) and tells the operator to enumerate. A one-line `curl` example with the `Authorization: Bearer $(jq -r .OPENAI_API_KEY ~/.codex/auth.json)` form (or whatever the auth.json field name actually is — see below) would shorten the loop from "read the doc, write the curl" to "copy-paste, see output." Purely optional; the existing text is already enough for a competent operator.

If you do add it, double-check the JSON key name against a fresh `~/.codex/auth.json` — the field name has shifted historically and the canonical reference is the file itself, not any doc snippet (including this one).

---

Verdict: holding off on an approve over @singlerider's active CR (per protocol). Once the template-compliance block is resolved, I'm comfortable flipping to approve assuming the two 🟡 warnings (or at least the `enc2:` string one) are addressed — the rest is teach/polish, not block.

Reviewed at `07bf22240435254dc799bcb1a9478df54197b69c`.

perlowja:
Addressed the review feedback:

**@singlerider — template compliance (blocking):** rewrote the PR body into the required structure — `## Summary` (with scope boundary, blast radius, linked-issue framing, label snapshot), `## Validation Evidence` (docs gate output, exit 0), `## Security & Privacy Impact` (all `No`), `## Compatibility` (backward compatible, no surface change), and `## Rollback`.

**@tidux — 🟡 `enc2:` string mismatch:** fixed in `4d94010`. The doc now quotes the verbatim runtime message from `crates/zeroclaw-config/src/secrets.rs:160` — `enc2: decryption failed (wrong .secret_key or tampered ciphertext)` (parentheses, and the `or tampered ciphertext` clause) — and notes that clause shares the error path, so the message alone doesn't distinguish a foreign profile from a corrupt blob. Copy-paste grep now matches.

On your other notes: the foreign-profile `mv` is already a one-shot "do this once if present" step in the surrounding prose; the `catalog.md` `OPENAI_API_KEY` inconsistency is a real but separate concern — happy to file a follow-up to tighten that page rather than widen this docs PR; and the `codex/models` curl example is a nice-to-have I left out to avoid pinning an auth.json field name that has shifted historically (the doc points at the file as canonical).

Docs gate green (`scripts/ci/docs_quality_gate.sh` → `0 error(s)`), all CI checks passing. Re-requesting review.

perlowja:
Pushed `623948e` — added a runnable `codex/models` enumeration example (copy-paste curl extracting the OAuth token + account id from `~/.codex/auth.json`, listing served slugs) and documented the `client_version` gate (a stale value returns an empty `models` array with no error). This addresses the last review suggestion.

Recap of everything addressed since the `CHANGES_REQUESTED`:
- **Template compliance** (the blocking item): body rewritten with all required sections — `## Summary`, `## Validation Evidence`, `## Security & Privacy Impact`, `## Compatibility`, `## Rollback`.
- **`enc2:` string** (`4d94010`): aligned verbatim to the runtime message in `crates/zeroclaw-config/src/secrets.rs`.
- **`codex/models` example** (`623948e`): concrete runnable invocation.

All CI checks pass. @singlerider the template concern is resolved — could you re-review / clear the stale `CHANGES_REQUESTED`? @tidux this should cover your 🟡 and 🔵 notes.

tidux (approved):
Re-reviewed at `623948ee7ed59f7ab91f11220473db92a6669b59` against `master` at `5de4940cb` (merge-base). Two new commits since the last pass (`4d94010` + `623948e`), plus a PR-body rewrite. Verified all author claims against the source.

### ✅ Round-1 items: addressed

**Template compliance (was @singlerider's blocking CR).** The body now has `## Summary` (with scope boundary, blast radius, linked-issue framing, label snapshot), `## Validation Evidence (required)` (docs gate output, exit 0), `## Security & Privacy Impact (required)` (`No` across the block), `## Compatibility (required)` (backward compatible, no surface change), and `## Rollback`. Cross-checked against `.github/pull_request_template.md`: all four `(required)` sections are present and answered. @singlerider — this looks resolved from my side, deferring to you on flipping the `CHANGES_REQUESTED`.

**🟡 `enc2:` string mismatch (`4d94010`).** Verified verbatim against `crates/zeroclaw-config/src/secrets.rs:160`:

```
enc2: decryption failed (wrong `.secret_key` or tampered ciphertext)
```

Parentheses correct, `or tampered ciphertext` clause now present, and the doc adds the right operational note — that clause shares the error path so the message alone doesn't distinguish a foreign profile from a corrupt blob. Copy-paste grep matches. 🟡 cleared.

**🔵 `codex/models` runnable example (`623948e`).** Cross-checked the new curl against the runtime:

- `jq -r .tokens.access_token ~/.codex/auth.json` — matches the `CodexAuthFile { tokens: CodexAuthTokens { access_token, ... } }` parse shape in `crates/zeroclaw-providers/src/auth/openai_oauth.rs:404-417`. ✅
- `jq -r .tokens.account_id ~/.codex/auth.json` — same parse shape, optional field. ✅
- Headers: `Authorization: Bearer ${AT}`, `chatgpt-account-id: ${ACC}`, `originator: pi` — match `responses_request_builder` in `crates/zeroclaw-providers/src/openai_codex.rs:1244-1253` exactly (Bearer auth + `originator: pi` always, `chatgpt-account-id` when account_id is set). ✅
- The `client_version` gate note (stale value → empty `{"models": []}` with no error) is the right shape for a footgun warning and aligns with what the upstream Codex CLI does. ✅
- The "verify against the file itself; the layout has shifted across Codex versions" caveat is correctly placed.

🔵 cleared, and better than what I suggested — the example also documents the `client_version` failure mode.

### 🟡 Minor — JWT-fallback `account_id` not mentioned

The runtime falls back to extracting `account_id` from the OAuth JWT when `tokens.account_id` is absent in `auth.json` — see `crates/zeroclaw-providers/src/openai_codex.rs:1178-1182` (`profile.and_then(|p| p.account_id).or_else(|| ... extract_account_id_from_jwt)`). The curl example, however, only reads `tokens.account_id` via `jq`, so on an `auth.json` where the field is absent (which does happen — `account_id` is wrapped in `#[serde(default)] Option<String>`), `jq` emits the literal string `null` and the request goes out as `chatgpt-account-id: null`, which 401s.

Not a blocker — the doc already says "verify against the file itself; the layout has shifted across Codex versions" — but a one-liner noting "if `.tokens.account_id` is `null`, run `zeroclaw auth status` after import to see the resolved id" would close the loop. Optional polish.

### 🟢 Validation evidence reproduces locally

Reran the docs gate against the PR-head tree:

```
$ DOCS_FILES=$'docs/book/src/SUMMARY.md\ndocs/book/src/providers/openai-codex-subscription.md' \
    bash scripts/ci/docs_quality_gate.sh
Linting docs files: docs/book/src/SUMMARY.md docs/book/src/providers/openai-codex-subscription.md
markdownlint-cli2 v0.20.0 (markdownlint v0.40.0)
Finding: docs/book/src/SUMMARY.md docs/book/src/providers/openai-codex-subscription.md !target/** !docs/book/src/SUMMARY.md
Linting: 1 file(s)
Summary: 0 error(s)
```

Matches the author's quoted output exactly.

### 🟢 Scope is still 2 files / +227 / -0

Sanity-checked the actual diff against the merge-base (`upstream/master` at `5de4940cb`): exactly `docs/book/src/SUMMARY.md` (+1 line, the new entry under Model Providers) and `docs/book/src/providers/openai-codex-subscription.md` (new, 226 lines). No scope creep, no incidental ToC edits, no deletions. Matches the PR metadata.

### 🔵 Carry-over — `catalog.md` `OPENAI_API_KEY` inconsistency

Unchanged status from round 1: `docs/book/src/providers/catalog.md` still credits the Codex subscription variant to "`OPENAI_API_KEY` / `~/.codex/auth.json`", which contradicts this PR's correct "no `OPENAI_API_KEY` — OAuth from `~/.codex/auth.json`" framing. Author has agreed this is a follow-up. Tracking issue would be welcome; not a blocker on this PR.

---

**Verdict: LGTM. Approving** pending @singlerider clearing the stale template-compliance CR. The 🟡 JWT-fallback note is non-blocking polish — happy to merge as-is.

Reviewed at `623948ee7ed59f7ab91f11220473db92a6669b59`.

Audacity88 (changes requested):
I re-reviewed #7116 at 623948ee7ed59f7ab91f11220473db92a6669b59 after the PR-body rewrite, @tidux's approval, and @singlerider's dismissed template review. I read the full diff, checked the current provider/auth source on `master`, reran the docs gate on the PR head, and refreshed the time-sensitive OpenAI Codex pricing/usage docs because this page is about current operator billing behavior.

### 🟢 What looks good — the ZeroClaw config path matches the source

The `openai` slot shape is right. `requires_openai_auth = true` routes the OpenAI family to the Codex OAuth provider, and `wire_api = "responses"` remains the right config for the Responses-shaped path. The `~/.codex/auth.json` import shape also matches the source: `tokens.access_token` is required, `tokens.account_id` is optional, and ZeroClaw stores the imported token profile under the configured auth profile.

The docs-only validation is also clean. I reran:

```text
DOCS_FILES='docs/book/src/SUMMARY.md
docs/book/src/providers/openai-codex-subscription.md' bash scripts/ci/docs_quality_gate.sh
```

Result: markdownlint ran on the changed docs and reported `Summary: 0 error(s)`.

### 🔴 Blocking — the cost model now contradicts OpenAI's current Codex billing docs

The new page repeatedly tells operators that this path has "no per-token spend", is "flat $/mo, not per-token", and should be tracked as `$0` per call with only a rolling 5-hour message allowance. That appears to describe the launch-era subscription model, but it no longer matches OpenAI's current public Codex billing docs.

OpenAI's current Codex rate card says Codex usage for Plus and Pro customers uses flexible credits and token-based rates: input, cached input, and output are priced as credits per 1M tokens, and actual usage depends on the token mix for each task. The current ChatGPT credits doc says the plan's included usage is consumed first, then Codex usage draws from the user's credit balance after plan limits. The current "Using Codex with your ChatGPT plan" doc also frames Codex limits as agentic usage limits, with add-credits, upgrade, or wait-for-reset paths depending on the plan.

The included plan allowance is still real; that is not the issue. The problem is that the page states the whole path as "no per-token spend", "`$0` per call", and "flat $/mo, not per-token" without the over-limit credit draw boundary. Because this PR is specifically teaching users how to route and account for Codex subscription traffic, that stale model changes the operational advice: users may need to route around credit burn, model-specific token rates, included plan limits, and over-limit credit draw. The "routing is not about saving per-token money" sentence is therefore backwards once a user is outside included plan usage.

Please revise the opening, "Cost and routing", and "Subscription tiers and limits" sections around the current official model:

- included plan usage first;
- flexible credits after included limits where available;
- token-based Codex credit rates by model/token class;
- plan tiers as usage allowance multipliers, not a guarantee of per-call zero cost;
- any ZeroClaw-local cost-tracking limitation clearly named as a local accounting limitation, not as an OpenAI billing fact.

### 🔴 Blocking — raw `auth.json` distribution needs a credential-sharing boundary

The multi-host section says the refresh owner should publish raw `~/.codex/auth.json` to a pull point and that every other host should pull the raw token and re-import it. That file contains bearer/refresh credentials, and the surrounding text does not say "same account owner's machines only", "private secret channel only", or "never share this with other users or a team".

That boundary matters. OpenAI's Pro terms/help text explicitly prohibit sharing account credentials or making an account available to someone else. Even aside from policy, a raw `auth.json` pull point is a high-value secret. The docs should not teach token distribution without naming the security boundary.

Please tighten this section so the supported pattern is clearly limited to private same-owner host sync, using a secret manager or other encrypted/private channel. It should also say not to commit, publish, team-share, or place raw `auth.json` anywhere that another user or service can read. The PR body's Security & Privacy section should also acknowledge that this docs page changes operator-facing credential-handling guidance, even though the runtime code does not change.

### 🟡 Warning — the `codex/models` curl sends `chatgpt-account-id: null` for accepted imports

The curl example uses:

```bash
ACC=$(jq -r .tokens.account_id ~/.codex/auth.json)
```

and always sends `chatgpt-account-id: ${ACC}`. But `tokens.account_id` is optional in the import shape, and the runtime falls back to extracting the account id from the OAuth JWT when that field is absent. On an auth file ZeroClaw can import successfully via the JWT fallback, `jq -r .tokens.account_id` emits the literal string `null`, and the example sends `chatgpt-account-id: null`.

This does not need to block the PR once the billing and credential-boundary text is fixed, but the copy-paste example should either omit the header when the field is absent or tell users to get the resolved account id from `zeroclaw auth status` after import.

---

Verdict: requesting changes. The config/auth mechanics and docs gate are in good shape, but the page's central cost guidance is now stale against OpenAI's current Codex billing docs, and the raw-token sync guidance needs a clear credential-sharing boundary before this lands.

perlowja:
@Audacity88 — addressed all three findings:

- 🔴 **Cost model:** reframed the intro, "Cost and routing", and "Subscription tiers and limits" around OpenAI's current Codex billing — included plan usage first, then per-token flexible credits at per-model rates once the included allowance is exhausted, with tiers as usage-allowance multipliers rather than per-call-zero. The ZeroClaw `$0` is now in its own subsection explicitly labelled a *local accounting limitation, not an OpenAI billing fact*.
- 🔴 **Credential boundary:** the multi-host section now leads with a ⚠️ callout restricting raw `auth.json` distribution to same-owner hosts over a private/encrypted channel only — never commit/publish/team-share — and notes OpenAI's terms prohibit sharing account credentials. PR-body Security & Privacy note updated to flag the operator-facing credential-handling change.
- 🟡 **curl `account-id`:** `account_id` is treated as optional — `jq '.tokens.account_id // empty'` plus `${ACC:+-H ...}` so the `chatgpt-account-id` header is omitted when absent instead of sending literal `null`; also points at `zeroclaw auth status` for the resolved id.

Re-verified the config/auth mechanics against current master (`requires_openai_auth`, `wire_api = "responses"`, and the JWT account-id fallback in `crates/zeroclaw-providers/src/auth/mod.rs`). Merged current master; docs gate green.

Audacity88 (approved):
Context: I re-reviewed current head `b07e0bb24cdbf97da4f7777a63b0251bd52d0b10` against the live PR state, the prior singlerider/tidux/Audacity88 reviews, the PR body, visible checks, the full docs diff, the current provider/auth source, the existing provider docs, and current official OpenAI Codex pricing/usage and Terms docs. I also checked the docs-only validation evidence; GitHub's docs/CI checks are green on the current head.

### ✅ Resolved — Cost guidance now matches the current Codex billing model

The opening and Cost and routing sections no longer present this as a flat per-call `$0` path. They now explain included plan usage first, flexible credits after the included allowance, token-rate credit consumption by model/token class, and ZeroClaw's `$0` value as a local accounting limitation rather than an OpenAI billing fact. That addresses my main blocker.

### ✅ Resolved — Raw `auth.json` sync now has the credential boundary it needs

The multi-host section now limits raw `~/.codex/auth.json` distribution to same-owner hosts over private/encrypted channels, says not to commit/publish/team-share it, and the PR body calls this out as operator-facing credential guidance. That is the boundary the previous review was missing.

### ✅ Resolved — The account-id curl no longer sends literal null

The example now uses `.tokens.account_id // empty` and only adds `chatgpt-account-id` when an id is present, while pointing users to `zeroclaw auth status` for the resolved fallback id. That matches the runtime shape: `account_id` is optional on import, and the provider only sends the header when it has an id.

### 🟢 What looks good — The page stays honest about volatile provider facts

The page tells operators to query the served catalog instead of treating the dated table as source of truth, and it keeps subscription auth separate from API-key OpenAI routing. That makes the doc useful without promising model names, limits, or account state that can drift under it.

I do not see another active `CHANGES_REQUESTED` review in the current live state; this approval is intended to clear my earlier blocker, not to make a broader merge-readiness claim.

I'm approving; my previous blockers are resolved.

perlowja:
@Audacity88 — this one is approved (you + @tidux) and currently MERGEABLE/CLEAN. My account doesn't have merge rights on the repo; would you mind merging it when you next get a chance? Thanks.

Issue #7117, opened 2026-06-03
[Feature]: Config UX parity across CLI, Quickstart, zerocode (incl. navigation), and web surfaces


singlerider:
### Summary

Bring every config surface — `zeroclaw config` CLI, Quickstart (CLI/TUI/web), zerocode config tabs (and the navigation layer that moves through them), and the web gateway Config pane — to a single, reference-grade UX bar, with reusable widgets as the implementing discipline so no surface re-invents the registry it already shares.

The reference bar is a mature peer agent runtime whose config UX demonstrates the patterns we want: one section table driving full / quick / per-section re-runs from a single dispatcher; auto-backup before any setup mutation; graceful non-interactive detection with guidance instead of silent breakage; a schema-driven web pane with category sidebar, field search, and a raw-mode escape hatch; a TUI that surfaces "setup required" in-place (run the wizard from inside the TUI) rather than ejecting to the shell; and a config loader that warns audibly (once) on a parse failure instead of silently falling back to defaults. We already own the harder half — a canonical registry, and a web pane that already has the grouped sidebar — and lag mostly on the remaining surface affordances (CLI init guidance, full onboarding coverage, TUI first-run, web search/raw-mode).

### Problem statement

Config is the project's weakest UX. The backend is already DRY where it counts: `crates/zeroclaw-config/src/sections.rs` is the canonical section registry (one macro table → `Section` enum, `shape()`, `help()`, `QUICKSTART_SECTIONS` ordering), `field_visibility.rs` is the single source for field hiding, and `init_defaults` is macro-derived from the schema. The problem is the **surfaces**, not the registry: five different entry points each re-implement flow and rendering on top of the same data, and they diverge in quality, completeness, and ordering.

Today an operator can configure ZeroClaw five ways, each with its own rough edges:

- **Hand-edit TOML** — works, but discouraged and unguided; no validation until daemon load.
- **`zeroclaw config` CLI** — `get`/`set`/`list`/`init`/`patch` exist, but the `init` flow for un-initialized sections is awkward: it instantiates `None` nested sections wholesale rather than walking the operator through what each section needs, and there is no guided "set the required fields next" hand-off.
- **Quickstart (CLI / TUI / web)** — right shape (single apply path through `quickstart/mod.rs`, surface-stamped), but incomplete. It covers the happy-path dependency chain (provider → profiles → memory → channels → agent) and stops short of being a full optional onboarding that can reach every section the registry defines.
[...]

PR #7118, opened 2026-06-03
fix(zerocode): unify navigation conventions and fix dead keybindings

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:** Unifies zerocode's navigation conventions and fixes three latent dead/inconsistent keybindings surfaced while auditing the config-movement layer:
  - `GlobalAction::ReloadDaemon` (`Ctrl+R`) is dispatched before any per-pane action, so `ChatTabAction::RenameSession` (also `Ctrl+R`) was unreachable. The session-rename feature was found to be fundamentally flawed and has been **removed end-to-end** (TUI action/overlay/draw/help/client + runtime method/dispatch/handler/types + locale keys); a `no_cross_enum_global_shadow` test now prevents a global chord from silently shadowing a pane action again.
  - The Quickstart pane advertised `Esc` to leave but had no `Back` action, so the key did nothing — added `Back [q, Esc]` wired through a `leave_requested` flag the event loop drains into `switch_mode`.
  - The config filter aliased `j`/`k` to cursor movement while actively capturing text, making section names containing those letters untypeable — `j`/`k` now reach the filter buffer; only arrows move the cursor there.
  - Added vim `j`/`k` to Quickstart list navigation (every other navigable pane already had it) and surfaced the file explorer's bound `j/k` movement and `l/h` dir-traversal in its footer so they're discoverable.
- **Scope boundary:** zerocode TUI navigation only. No config schema, RPC, gateway, web, or apply-path changes. No new sections. The larger config/onboarding surface work (CLI `init` guidance, onboarding resumability, web search/raw-mode) is intentionally out of scope and stays tracked on the umbrella issue.
- **Blast radius:** zerocode keybinding dispatch and a few help/footer hints. `Ctrl+T` and the new Quickstart `q`/`Esc` were verified free of conflicts via the existing intra-enum guard plus the new cross-enum guard. No other consumer reads these action enums.
- **Linked issue(s):** Related #7117 (this PR addresses the navigation slice; the issue stays open for the remaining config/onboarding surfaces).
- **Labels:** `enhancement`, `bug` (see sidebar).

Local validation. The diff is confined to `apps/zerocode`, so clippy and tests were run scoped to that crate; `cargo fmt --all -- --check` was run across the whole workspace.

```bash
cargo fmt --all -- --check
cargo test -p zerocode
```

    ```
    Checking zerocode v0.8.0-beta-2 (apps/zerocode)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 6.24s
    ```
  - `cargo test -p zerocode` →
    ```
    test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
    test result: ok. 202 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
    ```
    The new `keymap::tests::no_cross_enum_global_shadow` passes (it fails if any `GlobalAction` chord collides with a per-pane action, which is what the `Ctrl+R` rebind resolves).
- **Beyond CI — what did you manually verify?** Traced the dispatch order in `app.rs` confirming Global wins before pane handlers; confirmed the config filter and Quickstart text-capture modals route bare chars to their buffers so the j/k changes do not regress text entry.

  **Live manual verification (integration TUI, 2026-06-05):**
  - Quickstart pane `j`/`k` move the list selection — PASS. (Bindings are action-keyed via `QuickstartTabAction::Up=[k,Up]`/`Down=[j,Down]`; a stale custom `[keybindings]` override initially masked this, resolved live by switching preset via the Keybindings tab — `overrides::set_active` re-applies at runtime, no restart.)
  - Quickstart `q`/`Esc` leave the pane (Back) — PASS (was a dead no-op before).
  - Config filter `/` active → `j`/`k` type into the filter, do not move the cursor — PASS.
  - File-explorer footer renders the vim hint string verbatim — PASS:
    `j/k=move  l/h=in/out dir  Space=select  Enter=confirm  /=search  .=hidden  Esc=cancel`
  - `no_cross_enum_global_shadow` guard test passes; full `cargo test -p zerocode --bins` → 272 passed, 0 failed after the rename removal.
- **If any command was intentionally skipped, why:** Full-workspace `cargo clippy`/`cargo test` were scoped to `zerocode` because the diff touches no other crate; running the whole workspace would only rebuild unrelated code.

#### 2. Quickstart nav + Ctrl+R collision + filter + explorer footer (#7118) — Tier 0–1

Each item maps to one commit in this PR. Test only these, not base nav.

- [x] **`29667ac58`** Ctrl+R/Ctrl+T rename collision — RESOLVED BY DELETION. Session rename was fundamentally flawed and removed end-to-end (commit `afb4cbea9` on integration; cherry-picked + pushed to `feat/config-ux-parity-7117` #7118). The Ctrl+R/Ctrl+T collision is moot — no more RenameSession. Build clean, 272 zerocode tests pass. `no_cross_enum_global_shadow` guard still passes.
- [x] **`5af929780`** Quickstart pane `j`/`k` move the list — PASS. Default/preset keymap has `QuickstartTabAction::Up=[k,Up]`, `Down=[j,Down]`; modal stays arrow-only. Initial live fail was a stale custom `[keybindings]` override (`quickstart.up/down` frozen to arrows only). Switching preset via the Keybindings tab re-applied bindings LIVE (`overrides::set_active`, no restart) → j/k work immediately. Action-keyed, not hardcoded. (Correction: keybindings are NOT load-once — the Keybindings tab mutates the global override at runtime.)
- [x] **`a372b850f`** Quickstart pane: `q`/`Esc` leave the pane (Back) — PASS (live). Was a dead no-op before; now leaves to Dashboard.
- [x] **`a0085b8ba`** Config manager: `/` filter active → `j`/`k` type into the filter, do NOT move the cursor — PASS (live).
- [x] **`2664c5e01`** File-explorer footer renders the vim hint string `j/k=move  l/h=in/out dir  Space=select  Enter=confirm  /…` — PASS (live, select variant shown).

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` (default keybindings changed: session rename is removed entirely; Quickstart gains `q`/`Esc` back and `j`/`k` movement. Runtime keybinding overrides are unaffected and continue to take precedence.)
- If `No` or `Yes` to either: no upgrade steps required. Session rename (previously the non-functional `Ctrl+R` in chat) has been removed; operators do not need to do anything.

Low-risk: `git revert <sha>` per commit, or revert the branch range. No feature flags or config toggles involved.


WareWolf-MoonWall (approved):
## Context

Reviewing PR #7118 at head commit `2664c5e` — "fix(zerocode): unify navigation conventions and fix dead keybindings" by @singlerider.

This PR is a narrowly scoped slice from umbrella issue #7117 (Config UX parity). It addresses three latent dead/inconsistent keybindings in zerocode and adds vim navigation consistency. The PR is confined entirely to `apps/zerocode/` with no schema, RPC, gateway, or apply-path changes.

Five commits:
1. `29667ac` — Resolve Ctrl+R collision + add cross-enum guard
2. `5af9297` — Add vim j/k to Quickstart
3. `a372b85` — Give Quickstart a working back/cancel key
4. `a0085b8` — Let j/k type into config filter (remove cursor aliases)
5. `2664c5e` — Surface vim movement and dir-traversal keys in file explorer footer

CI is green (clippy, tests), validation evidence is complete, PR template is fully followed.

---

## 🟢 What's working — specific strengths

### The cross-enum guard is the MVP of this PR

The new `no_cross_enum_global_shadow` test (keymap/mod.rs:137–186) is the kind of proactive defensive design that prevents entire classes of defects. The existing intra-enum conflict guard only caught collisions within one action enum; this test walks every GlobalAction chord against every per-pane action enum and fails if any global shadows a pane action. The Ctrl+R collision (GlobalAction::ReloadDaemon vs ChatTabAction::RenameSession) slipped through because global dispatch wins (app.rs:345) — this test makes that class of mistake impossible going forward.

The test design is clean: it enumerates the five pane-level action enums (Chat, Logs, Dashboard, Config, Quickstart) and compares every global chord to every pane chord, failing with a clear message: `"global {ga:?} chord {gc:?} shadows a {label} action sharing it"`. This is exactly how CI should work — it codifies an architectural invariant (globals must not shadow per-pane actions) and fails loudly when violated.

### Chat help overlay is now driven by the binding registry

Commit `29667ac` refactors the chat session help overlay to render `Ctrl+N`, `Ctrl+S`, and `Ctrl+T` from `ChatTabAction::default_chords()[0].display()` instead of hardcoding the glyphs. This is the right pattern: the keymap registry is the single source of truth, and any display surface that shows a keybinding should resolve it from the action's `.default_chords()` rather than hardcoding the string. The `Box::leak()` into `&'static str` is necessary because `HelpEntry::key()` requires a static lifetime — the leak is bounded (three entries, allocated once per help render) and the cost is negligible compared to the DRY win.

This change also makes the rename keybinding (now Ctrl+T) automatically track if it's ever rebound in the future — the help overlay will show the new chord without needing a separate patch.

### Scope discipline is exemplary

Every commit touches only the subsystem it needs to: keymap for bindings, chat for its help overlay, config_manager for its filter dispatch, quickstart_pane for its leave request, file_explorer for its footer hints. There are no drive-by refactors, no unrelated formatting sweeps, no speculative abstractions. The PR solves the four stated problems (Ctrl+R collision, Quickstart missing back key, j/k in filter, missing vim hints) and nothing else. This is the model for focused, reviewable work.

### The filter j/k fix is the right call

Config filter mode was aliasing j/k to cursor movement even while actively capturing text, which made filtering for "json" or "key" move the cursor instead of typing the letters. Commit `a0085b8` removes the j/k aliases from the filter-active path so only arrows move the cursor while filtering, and all printable characters reach the buffer. This matches the existing help overlay (which already advertised arrows-only in filter mode) and matches the QuickstartModalAction convention, where j/k are deliberately left arrow-only because the dispatch is shared with a text-capture modal.

The fix is minimal (two lines) and scoped to the exact problem: the `|| Chord::char('k').matches(&key)` and `|| Chord::char('j').matches(&key)` guards are removed from the `CursorUp`/`CursorDown` branches. The commit message walks through the rationale (section names containing j/k became untypeable, arrows-only was already advertised), and the change aligns filter behavior with the documented convention.

---

## 🟡 Clarification requested — one gap in manual verification

The PR body states:

> **Beyond CI — what did you manually verify?** Traced the dispatch order in app.rs confirming Global wins before pane handlers (so the rebind is necessary and sufficient); confirmed the config filter and Quickstart text-capture modals route bare chars to their buffers so the j/k changes do not regress text entry. **Not manually exercised in a live terminal session this round — relied on the unit guards and dispatch tracing.**

The scoped clippy/tests passed, and the unit guards are strong (especially the new cross-enum guard), but for a keybinding change affecting four user-facing surfaces (chat rename, Quickstart back/movement, config filter, file explorer hints), a smoke test in a live terminal would catch any interaction defects the unit layer can't see. Examples: does Ctrl+T actually open the rename modal in the chat pane? Does q/Esc actually leave the Quickstart pane and return to Dashboard? Does filtering for "json" in the config pane now work?

This is not blocking — the unit coverage and dispatch tracing give high confidence the code is correct — but keybinding changes are one of the categories where "it compiles and the tests pass" doesn't fully substitute for "I pressed the key and the thing happened." If you've done a live smoke test since writing the PR body, noting that here would close the loop. If not, it's a low-priority follow-up (the defect risk is low given the guards).

---

## 🔵 Suggestions — two optional improvements

### Consider a help overlay consistency pass as a follow-up

This PR updates the chat help overlay to resolve chords dynamically and adds j/k/l/h hints to the file explorer footer, which are both steps toward discoverable, registry-driven help. A natural follow-up (not blocking for this PR, but worth tracking) would be to audit every help overlay and footer in zerocode and ensure they all:

1. Resolve keybindings from `.default_chords()` rather than hardcoding strings
2. Show the currently active convention for that pane (especially for context-dependent keys like h/l)
3. Match the action enums exactly (no advertised keys without backing actions)

The file explorer's `j/k=move l/h=in/out dir` additions in this PR are the right pattern; a follow-up could apply that pattern to any other surface that binds movement or traversal keys without showing them in the help.

### The `leave_requested` flag could become a reusable pattern

Commit `a372b85` introduces `leave_requested: bool` in `QuickstartPane` to bridge the pane-can't-switch-modes gap: the pane sets the flag when `Back` is pressed, and the event loop drains it after `handle_key` to call `switch_mode`. This is clean and works, but it's bespoke to QuickstartPane.

If other panes need the same "request a mode switch from inside handle_key" capability in the future (e.g., a Config pane action that should return to Dashboard), consider factoring this into a shared pattern — either a `PaneSwitchRequest` enum that panes can return from `handle_key`, or a `requested_mode: Option<Mode>` field on every pane that the event loop checks in a unified way. This isn't pressing (only one pane needs it today), but if the pattern recurs, making it shared would be cleaner than copy-pasting the `take_leave_request()` boilerplate.

---

## Verdict rationale

This PR is cleanly scoped, well-tested, and addresses real UX defects. The new cross-enum conflict guard is a meaningful quality bar lift for the entire keymap system. The fixes are surgical (Ctrl+R rebind, Quickstart back key, filter j/k, file explorer hints) and each commit stands alone with a clear rationale. The PR template is fully followed, CI is green, and the validation evidence is appropriate for the scope (scoped clippy/tests because the diff touches only zerocode).

The one gap (no live terminal smoke test noted) is minor — the unit coverage and dispatch tracing give high confidence the changes are correct. The suggestions (help overlay consistency, reusable leave-request pattern) are both optional follow-up work, not blockers.

No active `CHANGES_REQUESTED` from other reviewers. No architectural violations, no security concerns, no breaking changes. This PR is ready to merge.

tidux (approved):
## Context

Reviewing PR #7118 at head 2664c5e — `fix(zerocode): unify navigation conventions and fix dead keybindings` by @singlerider. Five commits, +111/-12 across 7 files in `apps/zerocode/`. Mergeable against `master` at 5de4940c, CI green.

@WareWolf-MoonWall has already approved on this head. I read their review in full and agree with every finding — the cross-enum guard, the registry-driven help refactor, the filter j/k fix, and the scope discipline are all the right calls. I'm not re-raising anything they already covered.

I checked out the branch locally and ran `cargo fmt -p zerocode -- --check`, `cargo clippy -p zerocode --all-targets -- -D warnings`, and `cargo test -p zerocode` — all clean, the new `keymap::tests::no_cross_enum_global_shadow` passes. I also grepped `apps/zerocode/src/`, `docs/`, and `apps/zerocode/locales/` for stray `Ctrl+R` references tied to rename and found none — the rebind is complete and the i18n strings only carry semantic labels, not chord glyphs, so no locale catalog regresses.

One additional finding I want on the record below, plus a couple of observations.

---

### 🟡 Warning — `Box::leak` on `Chord::display()` grows allocator footprint per help-modal frame

The new chat help overlay at `apps/zerocode/src/chat.rs:1094–1116` calls `Box::leak(... .display().into_boxed_str())` for `NewSession`, `SwitchSession`, and `RenameSession`. `Chord::display()` returns a freshly allocated `String` every call (`apps/zerocode/src/keymap/chord.rs:53–80` formats `parts.join("+")` and the keycode each time), and `HelpContext::help_context()` is invoked from the per-frame `term.draw(...)` closure at `apps/zerocode/src/app.rs:228–233` whenever the help overlay is showing. So every redraw frame the help is open allocates three short `String`s and intentionally leaks them — they sit in the allocator until process exit.

This PR didn't invent the pattern — `apps/zerocode/src/app.rs:210–219` already does the same thing for `PaneNavLeft` / `PaneNavRight`, attributable to #6848 (2c91478). What #7118 does is grow the leaked-per-frame count from 2 to 5, and on the chat pane which is arguably the most-used help in zerocode. In absolute terms it is small (each chord renders to roughly 6–8 bytes), but it's unbounded over uptime and it scales with how aggressively a user opens help (e.g. a screen-reader-friendly flow that toggles help repeatedly).

Not blocking — the leak rate is small enough that you'd need millions of frames to notice — but the chat help is also the cleanest example to fix the pattern at the same time you're extending it. A few options that would all be cheap:

- Cache the three chord-display strings in `Chat` once at construction (or first access) behind `OnceCell<String>` / `Lazy<String>` keyed on the action, and pass `&str` into `HelpEntry::key`.
- Change `HelpEntry::key` to take `Cow<'static, str>` or `impl Into<String>` so callers can hand it an owned `String` without leaking. The `&'static str` requirement is the constraint that forces `Box::leak` here, and relaxing it on the help-entry side removes the leak everywhere — including the pre-existing `app.rs` site.

Either is a few lines; the second option also retires the `Box::leak` already in `app.rs`. Happy to defer to a follow-up if you'd rather keep this PR purely scoped to navigation, but flagging it now since this is the moment the pattern is growing.

### 🟢 What looks good — cross-enum guard scales

@WareWolf-MoonWall already called this out and they're right — I want to add one structural note from my side: the test at `apps/zerocode/src/keymap/mod.rs:137–186` runs `|GlobalAction| × |ChatTabAction ∪ LogsTabAction ∪ DashboardTabAction ∪ ConfigTabAction ∪ QuickstartTabAction|` chord comparisons, which is `O(n × m)` over the cartesian product of every pane chord against every global chord. That's small today (the panes carry tens of bindings, not hundreds), and the test stays well under a millisecond locally. It's worth keeping that complexity in mind if the global enum ever grows substantially or new pane enums are added — but right now this is the right shape: a flat invariant check that fails loudly with the exact pane label and chord in the assertion message. The `assert!(gc != pc, "global {ga:?} chord {gc:?} shadows a {label} action sharing it")` framing is the kind of message that tells a future contributor what to do without having to read the test source.

### 🟢 What looks good — `take_leave_request()` drain is race-free by construction

The `leave_requested: bool` + `take_leave_request()` bridge in `apps/zerocode/src/quickstart_pane.rs:611–619` and `apps/zerocode/src/app.rs:397–406` works because the event loop holds the only `&mut quickstart`, calls `quickstart.handle_key(...)` (which can set the flag), and drains the flag in the *same* `Event::Key` arm before the next iteration of the loop. No second event can interleave between set and drain, and `mem::take` returns the prior `true` while resetting to `false` atomically with respect to the loop's single-threaded execution. It's a small thing but the lifetime story here is clean and worth noting — it's exactly the pattern that a future reviewer might want to nervously double-check, and it holds up.

I think @WareWolf-MoonWall's `🔵` suggestion about generalizing this into a `PaneSwitchRequest` enum is the right long-term move, but I'd only do it when a second pane needs the same capability — premature abstraction here would be more boilerplate than it saves.

### 🟢 What looks good — i18n parity holds

Just confirming what's implicit in the diff: the `Box::leak` introduced in chat.rs only affects the chord display glyph, not the label. The localizable strings (`zc-chat-help-new-session`, `zc-chat-help-session-list`, `zc-chat-help-rename-session`) still flow through `crate::i18n::t(...)`, so all 31 locale catalogs under `apps/zerocode/locales/` keep their existing translation coverage. No locale regresses, and the chord glyph is intentionally not translated — `Ctrl+T` is the right rendering everywhere except darwin (where `Chord::display()` produces `⌘T` from the same code path).

### Verdict rationale

@WareWolf-MoonWall already approved on this exact head and I see no reason to disagree. The cross-enum guard is a meaningful invariant lift, the four navigation fixes solve real UX defects, and the diff stays inside `apps/zerocode/` with no schema, RPC, or apply-path drift. The one 🟡 above is a real-but-small concern about an existing leak pattern that this PR extends rather than introduces — worth either fixing here or tracking as a follow-up, but not blocking.

Local build, fmt, clippy, and test all clean on 2664c5e. Approving.

PR #7120, opened 2026-06-03
fix(shell): prevent interactive subprocesses from hijacking the terminal

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:** Stops interactive-required subprocesses from hijacking the controlling terminal — which, in a TUI deployment, is the terminal a zerocode session owns. Observed symptom: a `git` credential prompt (`Username for 'https://github.com'`) from one process appeared inside a different zerocode input box.
  - **shell tool** (`crates/zeroclaw-runtime/src/tools/shell.rs`): spawned commands inherited stdin from the parent. Set stdin to `/dev/null` so any command that reads stdin (or an interactive prompt) gets immediate EOF and fails instead of blocking on or drawing from the terminal.
  - **git operations** (`crates/zeroclaw-tools/src/git_operations.rs`): git ignores a non-terminal stdin for credentials by design and falls back to `/dev/tty`. Set `GIT_TERMINAL_PROMPT=0` (git's own documented non-interactive contract) plus a null stdin on the single production git spawn helper, so credential ops fail with `terminal prompts disabled` rather than prompting.
- **Scope boundary:** Does NOT add any hardcoded list of "interactive tools" or per-tool env tables — the shell fix is tool-agnostic (null stdin), and the git fix uses git's single documented flag. Does NOT change command parsing, the sandbox, autonomy gating, or env passthrough. Does NOT detach the child into a new session (`setsid`); see Blast radius.
- **Blast radius:** Process spawning in the shell tool and the git tool. A command that explicitly opens `/dev/tty` (rather than reading stdin) could still find a controlling terminal if the daemon has one; fully severing that requires a new session, which interacts with the existing `ChildGroupGuard` process-group reaping and is intentionally left out of this fix. The reported vector (git) is fully covered by `GIT_TERMINAL_PROMPT=0`.
- **Linked issue(s):** None (standalone fix surfaced during TUI work).
- **Labels:** `bug` (see sidebar).

The diff touches only `zeroclaw-tools` and `zeroclaw-runtime`, so clippy and tests were scoped to those crates; `cargo fmt --all -- --check` was run across the whole workspace.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-tools -p zeroclaw-runtime --all-targets -- -D warnings
cargo test -p zeroclaw-tools git
cargo test -p zeroclaw-runtime tools::shell
```

  - `cargo clippy -p zeroclaw-tools --all-targets -- -D warnings` → `Finished` (no warnings).
  - `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` → `Finished` (no warnings).
  - `cargo test -p zeroclaw-tools git` → `test result: ok. 33 passed; 0 failed`.
  - `cargo test -p zeroclaw-runtime tools::shell` → `test result: ok. 40 passed; 0 failed`.
  - New tests: `git_credential_op_fails_fast_without_terminal_prompt` (timeout-guarded fetch to an unreachable private remote returns fast) and `shell_stdin_is_eof_not_the_terminal` (a stdin-reading command gets EOF instead of hanging). Both complete in <0.05s.
- **Beyond CI — what did you manually verify?** Confirmed git's behavior directly: with `GIT_TERMINAL_PROMPT=0`, `git ls-remote` to a private GitHub URL fails immediately with `fatal: could not read Username for 'https://github.com': terminal prompts disabled` instead of prompting. Confirmed via git's own manual that `GIT_TERMINAL_PROMPT` is the documented switch for this, and that git prompts on the terminal independent of stdin.
- **If any command was intentionally skipped, why:** Full-workspace clippy/test scoped to the two touched crates; no other crate is affected by the diff.

#### 12. Shell/git terminal hijack prevention (#7120) — Tier 4

Commits `edbf9e1d0` (null stdin), `0604c0068` (git prompt disable).

- [x] Spawned command that reads stdin (`read`, ssh/sudo prompt) → fails fast with null stdin, no hang (`edbf9e1d0`) — PASS (unit, timeout-guarded). `shell_stdin_is_eof_not_the_terminal` (shell.rs:502) spawns `cat` (reads stdin) under a 10s timeout; with `stdin(Stdio::null())` at spawn (shell.rs:338) it gets immediate EOF, returns fast and succeeds. If stdin reached the terminal the test would hang and fail. Contract: interactive-required commands fail, never prompt or steal keystrokes from the TUI.
- [x] Git credential op with no creds → fails fast, no TTY prompt leak/hang (`0604c0068`) — PASS (unit, timeout-guarded). `git_credential_op_fails_fast_without_terminal_prompt` (git_operations.rs:1262) fetches `https://127.0.0.1:1/private/repo.git` under a 10s timeout; the production helper sets `GIT_TERMINAL_PROMPT=0` + null stdin (git_operations.rs:212-213) so git cannot fall back to /dev/tty — it returns fast (not hung) AND errors. Both tests green.

- Secrets / tokens / credentials handling changed? `Yes` — git credential prompting is now disabled in the tool path, so operations that would have prompted for a username/password now fail fast instead. No credential is stored, logged, or transmitted; the change strictly removes an interactive prompt path.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: no upgrade steps. Behavioral note: a git operation against a remote that requires credentials and has none cached will now fail with a clear `terminal prompts disabled` error instead of hanging — configure a credential helper or cached credentials for such remotes.

Low-risk: `git revert <sha>` for either commit independently. No feature flags or config toggles involved.

### DIFF
--- a/crates/zeroclaw-runtime/src/tools/shell.rs
+++ b/crates/zeroclaw-runtime/src/tools/shell.rs
@@ -335,6 +335,7 @@ impl Tool for ShellTool {
         // `output()` pipes stdio implicitly; `spawn()` does not.
         cmd.stdout(std::process::Stdio::piped());
         cmd.stderr(std::process::Stdio::piped());
+        cmd.stdin(std::process::Stdio::null());
 
         let mut child = match cmd.spawn() {
             Ok(c) => c,
@@ -496,6 +497,24 @@ mod tests {
                 .contains(&json!("command"))
         );
         assert!(schema["properties"]["approved"].is_object());
+    }
+
+    #[tokio::test]
+    async fn shell_stdin_is_eof_not_the_terminal() {
+        let security = Arc::new(SecurityPolicy {
+            autonomy: AutonomyLevel::Supervised,
+            workspace_dir: std::env::temp_dir(),
+            allowed_commands: vec!["cat".into()],
+            ..SecurityPolicy::default()
+        });
+        let tool = ShellTool::new(security, test_runtime());
+        let fut = tool.execute(json!({"command": "cat"}));
+        let res = tokio::time::timeout(std::time::Duration::from_secs(10), fut).await;
+        assert!(
+            res.is_ok(),
+            "a stdin-reading command hung — stdin is not null and may reach the terminal"
+        );
+        assert!(res.unwrap().expect("cat should return a result").success);
     }
 
     #[tokio::test]
--- a/crates/zeroclaw-tools/src/git_operations.rs
+++ b/crates/zeroclaw-tools/src/git_operations.rs
@@ -209,6 +209,8 @@ impl GitOperationsTool {
         let output = tokio::process::Command::new("git")
             .args(args)
             .current_dir(working_dir)
+            .env("GIT_TERMINAL_PROMPT", "0")
+            .stdin(std::process::Stdio::null())
             .output()
             .await?;
 
@@ -1255,6 +1257,28 @@ mod tests {
         // Branch listing is read-only; it must not require write access
         assert!(!tool.requires_write_access("branch"));
         assert!(tool.is_read_only("branch"));
+    }
+
+    #[tokio::test]
+    async fn git_credential_op_fails_fast_without_terminal_prompt() {
+        let tmp = TempDir::new().unwrap();
+        git_init_no_sign(tmp.path(), &[]);
+        let tool = test_tool(tmp.path());
+
+        let fetch = tool.run_git_command(
+            &["fetch", "https://127.0.0.1:1/private/repo.git"],
+            tmp.path(),
+        );
+        let res = tokio::time::timeout(std::time::Duration::from_secs(10), fetch).await;
+
+        assert!(
+            res.is_ok(),
+            "git fetch hung — it likely prompted for credentials on the terminal"
+        );
+        assert!(
+            res.unwrap().is_err(),
+            "fetch to an unreachable private remote should fail, not succeed"
+        );
     }
 
     #[tokio::test]


WareWolf-MoonWall (approved):
**Context:** Reviewing at head commit `0604c00` (2 commits total). PR touches `zeroclaw-tools` (git operations) and `zeroclaw-runtime` (shell tool) — both HIGH RISK per AGENTS.md risk tiers. All CI checks passed. No prior reviews, inline threads, or top-level comments.

---

## 🟢 Security posture improvement

This PR closes a legitimate TUI input corruption vector with surgical, well-scoped fixes. The problem was real: a git credential prompt appearing inside a zerocode input field is exactly the kind of session corruption that erodes user trust. Fixing it at the root — process stdin/env configuration — rather than adding heuristics or blocklists is the right architectural choice.

**What's strong here:**

- **Shell fix is tool-agnostic.** Setting stdin to null applies uniformly to all spawned commands; no special cases or per-tool logic needed.
- **Git fix uses the documented contract.** `GIT_TERMINAL_PROMPT=0` is git's own API for "fail instead of prompting" — not a workaround, not a guess. The commit body explicitly references git's fallback to `/dev/tty` and explains why null stdin alone isn't enough for git.
- **Tests validate the fix, not just the feature.** `shell_stdin_is_eof_not_the_terminal` uses a 10-second timeout to assert a stdin-reading command completes (EOF received) rather than hanging. `git_credential_op_fails_fast_without_terminal_prompt` does the same for a fetch to an unreachable private remote. Both would have failed before the fix; both pass now.
- **Scope boundary discipline.** The PR explicitly states what it does NOT change: no session detachment (`setsid`), no command parsing, no sandbox changes, no autonomy gates. The `ChildGroupGuard` note in Blast radius shows awareness of the existing reaping contract and deliberate choice to leave it untouched.

**Why this matters:** Security boundary fixes like this are foundational. A command that can draw from the controlling terminal can leak credentials, leak session state, or intercept user input. The fix removes that capability cleanly without breaking existing behavior.

---

## 🟢 Test coverage matches the risk

The two new tests (`shell_stdin_is_eof_not_the_terminal` and `git_credential_op_fails_fast_without_terminal_prompt`) are not afterthoughts. They're timeout-guarded, they target the exact failure mode the PR addresses, and they complete in <50ms according to the validation evidence — fast enough to run on every commit, strong enough to catch regression.

The git test is particularly well-constructed: it targets an unreachable private remote (`https://127.0.0.1:1/private/repo.git`), which forces git to attempt credential resolution. Before the fix, that would hang waiting for `/dev/tty`; after the fix, it fails fast with `terminal prompts disabled`. The 10-second timeout ensures the test fails if the behavior regresses, rather than hanging CI.

---

## 🔵 Upstream completeness — does `/dev/tty` remain reachable?

The PR body states:

> A command that explicitly opens `/dev/tty` (rather than reading stdin) could still find a controlling terminal if the daemon has one; fully severing that requires a new session, which interacts with the existing `ChildGroupGuard` process-group reaping and is intentionally left out of this fix.

This is transparent, accurate, and documents the boundary of what's fixed. The git vector is fully covered by `GIT_TERMINAL_PROMPT=0`. For other tools that might open `/dev/tty` directly — the question is: are there any in scope, and does leaving that path open constitute residual risk?

**I'm marking this 🔵 suggestion rather than 🔴 blocking** because:

1. The reported vector (git) is fully fixed.
2. The shell tool is now protected against stdin-based prompts universally.
3. Session detachment (`setsid`) is a non-trivial change that touches process-group reaping (noted in the PR body).

But if another tool in the runtime or `zeroclaw-tools` spawns commands and could trigger a `/dev/tty`-opening prompt (e.g., `sudo`, `ssh`, interactive editors), that's a follow-on risk. Worth a quick audit: do any other tools spawn processes that could reach `/dev/tty`, and if so, should we track a follow-on issue for session isolation?

Not blocking merge — this PR delivers real improvement over the prior state — but flagging for consideration.

---

## ✅ Template completeness

PR body follows `.github/pull_request_template.md` precisely:

- **Summary** with scope boundary and blast radius.
- **Validation Evidence** with literal command output, beyond-CI notes (manual git behavior verification), and explicit skip reasons.
- **Security & Privacy Impact** answered per the template (Yes on credentials handling, explained).
- **Compatibility** answered (backward compatible, behavioral change documented).
- **Rollback** appropriate for the risk tier (low-risk: `git revert` per commit).

No gaps.

---

## ✅ Milestone alignment

PR is already assigned to v0.8.0, which is correct. This is a break-fix (`fix:` prefix + `bug` label) with no scope match to a specific v0.8.0 component, so it falls under the "break-fix → earliest open milestone" rule per the review protocol.

Tracking issue #7112 should be updated after approval to reflect this PR's inclusion.

---

## Verdict

**Approve.** This PR is a clean, well-scoped security improvement with strong test coverage, transparent documentation of what it does and doesn't fix, and appropriate risk/compatibility notes. The upstream completeness note (🔵 `/dev/tty` reachability) is not a blocker — the reported vector is fixed, and session isolation is a separable concern. If other tools are at risk from direct `/dev/tty` access, that's a follow-on issue, not a gate for this PR.

**Recommended next steps after merge:**

1. Update tracking issue #7112 to add this PR.
2. If a broader audit of `/dev/tty` reachability across all tool spawns is warranted, file a follow-on issue and assign to the maintainers who own the tool boundary.

tidux (approved):
**Context:** Reviewing at head 0604c00 against current master 40be7738. Branch is MERGEABLE / CLEAN; all Quality Gate checks green. @WareWolf-MoonWall already approved on this same head — joining as second approval, with a few notes that narrow the open follow-up rather than add new ones.

I cross-checked the diff against the local source, ran `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-tools -p zeroclaw-runtime --all-targets -- -D warnings`, and re-ran both new tests by name on 0604c00 — `shell_stdin_is_eof_not_the_terminal` and `git_credential_op_fails_fast_without_terminal_prompt` both pass in tens of milliseconds.

---

### 🟢 What looks good — single production git spawn, verified

The PR body claims `GIT_TERMINAL_PROMPT=0` + null stdin are applied to "the single production git spawn helper." I grepped the file: `crates/zeroclaw-tools/src/git_operations.rs` has fourteen `Command::new("git")` call sites, but `mod tests` begins at line 1046 and every other site (1074, 1394, 1410, 1544, 1549, 1566, 1579, 1601, 1618, 1661, 1696, 1727, 1759) lives inside that test module. Line 209's `run_git_command` is the sole production helper, and every production git operation (`git_status`, `git_diff`, `git_log`, `git_branch`, `git_commit`, `git_add`, `git_checkout`, `git_stash_*`, `git_worktree_*`) routes through it. The two-line patch is sufficient — no missed call sites.

### 🟢 What looks good — null stdin doesn't break legitimate shell usage

I traced the shell tool path before signing off on `cmd.stdin(Stdio::null())`. `ShellTool::execute` takes a single `command` string and hands it to `RuntimeAdapter::build_shell_command`, which wraps it in `sh -c` / equivalent. Any heredoc or piped input the user writes inside that string is handled by the inner shell's own stdin plumbing — the parent process's stdin is never used as a data channel. Setting it to `/dev/null` strictly removes the controlling-terminal path; it does not regress any documented input mode for the tool.

### 🟢 What looks good — test construction is the right minimal form

`shell_stdin_is_eof_not_the_terminal` uses `cat` with no args. `cat` reads from stdin until EOF; with stdin null, EOF arrives immediately and the process exits 0. Before the PR, the inherited stdin would have kept `cat` blocked. The 10-second `tokio::time::timeout` is a generous safety net rather than a real bound — locally the test completes in well under 50 ms. Similarly, `git_credential_op_fails_fast_without_terminal_prompt` fetches `https://127.0.0.1:1/private/repo.git`: that address is unreachable, so libcurl resolves the connect error fast, and the assertion is "the future completes within 10 s and returns `Err`" — exactly the contract you want.

### 🔵 Suggestion — narrowing @WareWolf-MoonWall's `/dev/tty` follow-up

@WareWolf-MoonWall flagged a 🔵 about whether other tools might spawn a process that opens `/dev/tty` directly. I did the audit while reviewing: across `crates/zeroclaw-runtime/src/` and `crates/zeroclaw-tools/src/`, the only LLM-callable spawns are this `ShellTool` and `GitOperationsTool::run_git_command`. The other `Command::new(...)` call sites I see — `doctor/mod.rs` (`df`, probes), `service/mod.rs` (`launchctl`, `systemctl`, `rc-service`, `schtasks`), `cron/scheduler.rs` (the cron `sh -c` runner, which itself goes through the same null-stdin shell path), and the sandbox drivers under `security/{bubblewrap,docker,firejail,seatbelt,landlock}.rs` — are infrastructure the runtime invokes itself with known argument sets, not arbitrary subprocesses on behalf of an LLM. So the **tool-boundary attack surface is fully covered by this PR**.

That makes the residual `/dev/tty` work a separable hardening item against the daemon's own controlling terminal (the `setsid` path the PR body intentionally defers), not a tool-routing gap. Worth a follow-on issue against the broader process-isolation work that already touches `ChildGroupGuard`, but explicitly not a gate on this PR.

### ✅ Resolved — template completeness

PR body covers Summary / Scope boundary / Blast radius / Validation Evidence / Security & Privacy / Compatibility / Rollback. The behavioral note under Compatibility ("configure a credential helper or cached credentials for such remotes") is the right thing to surface to operators — anyone who's been silently relying on interactive prompts will see a clean `terminal prompts disabled` error and know what to do.

### ✅ Resolved — milestone alignment

v0.8.0 is already set. Break-fix (`fix:` prefix + `bug` label) with no scope match against v0.8.0's literal description (config / tool-call-parser tier promotion), so the protocol fallback "earliest open milestone by version" lands it here. v0.8.0 < v0.8.1 < v0.9.0. No change needed.

---

## Verdict

Approve. Tight, well-scoped security fix; the documented boundary (`/dev/tty` via `setsid`) is correctly deferred and the audit above shows nothing else in tool scope is exposed. With two approvals on 0604c00 and a clean merge state, this is ready when @singlerider is.

Audacity88 (approved):
**Context:** I reviewed #7120 at head 0604c00. This is a high-risk review because the diff touches the runtime shell tool and the git operations tool, but the patch itself is narrow: one `spawn()` path now gets null stdin, and the git helper now disables terminal credential prompts.

I read the PR body, current diff, top-level comments, inline comments, and formal reviews. There are no inline threads. @WareWolf-MoonWall and @tidux have both approved this same head. I did not rerun cargo locally for this review; I relied on the author's scoped validation, current green CI for this head, and @tidux's focused local reruns of the two new regression tests.

### 🟢 What looks good — shell stdin is closed at the right layer

`ShellTool::execute` already had the right timeout, process-group, stdout, and stderr handling, but it used `spawn()`, so stdin was not implicitly closed the way `output()` callers are. Adding `cmd.stdin(Stdio::null())` at that spawn point is the minimal fix for the reported TUI prompt bleed: commands that try to read from stdin now see EOF instead of borrowing the daemon's controlling terminal.

The regression test is also pointed at the actual failure mode. `cat` with no input is a simple stdin reader; if stdin ever regresses back to inherited terminal input, the timeout trips instead of letting CI hang.

### 🟢 What looks good — git uses Git's own non-interactive contract

The git side needs more than null stdin because Git can fall back to `/dev/tty` for credentials. Setting `GIT_TERMINAL_PROMPT=0` in `run_git_command` is the right tool-specific guard: credential-required operations fail fast with Git's documented non-interactive behavior instead of drawing a prompt in another terminal UI.

I also checked the production git path. The production operations in `crates/zeroclaw-tools/src/git_operations.rs` route through `run_git_command`; the other `Command::new("git")` call sites in that file are test helpers. So the git change is applied at the correct shared helper rather than scattered across individual operations.

### 🔵 Suggestion — keep broader controlling-terminal isolation separate

The PR body is explicit that a command which opens `/dev/tty` directly can still reach the daemon's controlling terminal unless ZeroClaw also changes the process-session model. I agree with the existing reviews that this should stay separate. It interacts with process-group cleanup and `ChildGroupGuard`, so folding `setsid`-style hardening into this PR would make a small targeted fix much riskier.

The same applies to other subprocess surfaces such as MCP stdio, browser/CLI delegates, tunnels, service helpers, and skill shell tools. Some of those intentionally need stdio pipes, and many `output()` paths already close stdin by API behavior. This PR should not claim to solve every subprocess isolation question, but it does solve the reported shell/git prompt vector cleanly.

### 🟢 What looks good — validation covers the touched paths

The validation evidence covers the touched crates and the two new regression tests, and current CI for this head is green. I did not find a code or validation blocker in this slice.

## Verdict

Approve. This is a focused fix for a real terminal-hijack class, with the right residual risk boundary documented. Since the PR is still a draft, my approval means it is ready from my review side once @singlerider marks it ready.

PR #7121, opened 2026-06-03
fix(channels): scope channel runtime context to the agent workspace

ATECHPCS:
## Summary

- **Base branch:** `master`
- **What changed and why:** `ChannelRuntimeContext.workspace_dir` was initialized to the install-wide `config.data_dir` instead of the per-agent workspace (`config.agent_workspace_dir(agent_alias)`, already bound as `workspace` a few lines above and used for boot-time skill registration). `ctx.workspace_dir` drives **per-message** skill reloads, prompt refresh, and file-access scoping; skill loading resolves `<workspace_dir>/skills`. With it pointed at `data_dir`, per-message skill activation scans `<data_dir>/skills` (empty), so a skill that registers fine at boot **never activates on inbound channel messages**, and file tools are scoped to the wrong root.
- **Scope boundary:** One-line change at the production `ChannelRuntimeContext` construction site (inside the `for agent_alias in &enabled_agents` loop). Does not touch the test fixture `router_test_ctx` (which also sets `data_dir` but is irrelevant to runtime behavior). No config/API surface change.
- **Blast radius:** All channel types (Telegram, Discord, Slack, Matrix, …) under the V3 multi-agent runtime — any deployment relying on per-message skill activation or workspace-scoped file access.
- **Linked issue(s):** None (found while integrating a custom skill).

## Validation Evidence

```bash
cargo fmt -p zeroclaw-channels -- --check   # clean
cargo check -p zeroclaw-channels            # builds
```

- **Beyond CI — manual verification:** Reproduced on a live Telegram bot with a skill that declares `blocked_tools_with_image`. Before: the skill registered at boot (`Registered 1 skill(s): …`) but per-message activation logged `skill_candidates: 0` with `workspace_dir = <data_dir>`, and the skill never engaged. After: `skill_candidates: 1` with `workspace_dir = <agent>/workspace`, and the skill's image-turn tool gating fired as designed.

## Security & Privacy Impact

- New permissions / network calls / secret handling? **No.**
- The change *tightens* file-access scoping back to the intended per-agent workspace (previously the broader `data_dir`).

## Compatibility

- Backward compatible? **Yes** — corrects incorrect path resolution; no config, env, or CLI change. Agents whose workspace already equals `data_dir` are unaffected.

## Rollback

- `git revert <sha>`. No data migration, no config toggle involved.


Audacity88 (approved):
Reviewed head 76c2d9d. I checked the live PR state, diff, comments, formal reviews, CI, the `start_channels` construction path in `crates/zeroclaw-channels/src/orchestrator/mod.rs`, the config workspace helpers, and the existing channel skill-refresh tests. There are no prior active review blocks, and the visible CI checks are green. I’m approving this with one non-blocking follow-up suggestion.

### 🟢 What looks good — channel runtime now uses the agent workspace consistently

This fix targets the one production `ChannelRuntimeContext` construction site where the channel runtime had drifted away from the rest of the per-agent path handling. `start_channels` already resolves `workspace = config.agent_workspace_dir(agent_alias)` and uses that path for boot-time skill loading and tool setup, so using the same path for `ctx.workspace_dir` keeps per-message skill reloads, prompt refreshes, and workspace-scoped file behavior on the same agent boundary.

That also matches the config contract: `config.data_dir` is shared instance data, while `agent_workspace_dir(alias)` is where per-agent identity, markdown state, and workspace-scoped user files live. So this is a narrow correction back to the intended boundary rather than a new API or behavior surface.

### 🔵 Suggestion — add divergent path coverage when this harness is next touched

The existing `/new` skill-refresh test proves that the channel prompt can reload skills, but its fixture sets `config.data_dir` equal to the workspace path. That means it would not have caught this production bug, where `data_dir` and `agent_workspace_dir(agent_alias)` differ.

I do not think that should block this one-line fix, especially with the live Telegram before/after validation in the PR body. Still, the next time we touch this harness, a small regression that keeps `data_dir` and the agent workspace distinct would be useful protection for this path.

PR #7122, opened 2026-06-03
fix(mcp): surface tool-execution errors (result.isError) instead of swallowing them

ATECHPCS:
## Summary

- **Base branch:** `master`
- **What changed and why:** `McpServer::call_tool` only inspected the JSON-RPC top-level `error` field. Per the MCP spec, servers signal **tool-execution** failures (as opposed to protocol errors) with HTTP 200 + `result.isError: true` and the detail in `result.content[].text`. Those were returned as a normal success `Value`, so the failure was invisible — the model received an error-shaped "success", and the daemon log only showed a generic "error during tool call" with no detail. This change detects `result.isError == true`, joins `content[].text` into a detail string, logs a `WARN` event, and returns `Err` carrying that detail.
- **Scope boundary:** Single function (`call_tool`), after the existing JSON-RPC `error` check and before the success return. No transport, schema, or API changes. Successful results (`isError` absent or `false`) are returned exactly as before.
- **Blast radius:** All MCP tool calls across all configured servers. Behavior only changes for responses that were already failures — they now surface as errors instead of being swallowed.
- **Linked issue(s):** None (found while integrating an MCP server whose tool errors surfaced with no detail).

## Validation Evidence

```bash
cargo fmt -p zeroclaw-tools -- --check   # clean
cargo check -p zeroclaw-tools            # builds
```

- **Beyond CI — manual verification:** Against a live MCP server (SparkyFitness) whose `manage_*` tools return `isError:true` on validation failure: before, the orchestrator/daemon logged only `MCP server … error during tool call …` with no cause; after, the actual server message (e.g. the validation error text) is logged and propagated to the model, which can then correct the call.

## Security & Privacy Impact

- New permissions / network calls / secret handling? **No.**
- Surfaces server-provided error text into logs/responses (same channel as any other tool result); no new data sources.

## Compatibility

- Backward compatible? **Yes** — only the error path changes. Callers already handle `Err` from `call_tool` (timeouts, transport errors, JSON-RPC errors). No config/API surface change.

## Rollback

- `git revert <sha>`. No migration or config involved.


Audacity88 (changes requested):
Reviewed current head 0195d30cf6cc42bfd8f34e97e078df4827ce0739 against the PR diff, green checks, labels, the absence of prior comments/reviews, and the current public MCP client/error-sanitizer source. I agree with the direction: top-level JSON-RPC errors and MCP tool-execution failures need to be handled separately, and converting `isError: true` into an `Err` is the right shape for callers that already recover from MCP call failures.

I am requesting changes because the new detail path logs and returns raw, unbounded server-supplied text, and the changed protocol behavior is not covered by a regression test yet.

### 🟢 What looks good — The semantic split is in the right place

The placement after the existing JSON-RPC `error` check and before the success return is the right boundary. Once the detail is made safe to surface and covered by tests, this should be a small, valuable MCP correctness fix.

### 🔴 Blocking — Bound and sanitize MCP error detail before logging or returning it

`result.content[].text` is controlled by the MCP server/tool. With this patch, that text is joined into `detail`, recorded as a structured `WARN` attribute, and embedded in the returned error. That creates a new raw log/error surface for external MCP output, and it can also carry a very large payload into logs and the model-visible error path.

Please sanitize and bound the extracted detail before using it in both `zeroclaw_log::record!` and `bail!`. The tools crate already depends on `zeroclaw-providers`, so reusing the existing API-error sanitizer/truncation path would keep the fix narrow while still preserving the useful validation message that motivated this PR.

### 🔴 Blocking — Add regression coverage for the `isError` conversion

This is a protocol-shape change in `McpServer::call_tool`, but the PR only cites `fmt`, `check`, and a manual live MCP smoke. The existing `mcp_client.rs` tests do not exercise the new result-level error branch, and this module has enough private access to test it with a fake `McpTransportConn` without standing up a live MCP server.

Please add focused coverage for at least the `result.isError == true` response returning `Err` with the sanitized/bounded detail, plus the normal success response still returning `Ok(result)` when `isError` is absent or false. A missing/empty `content[].text` fallback test would also be useful given this branch now owns the user-facing error message.

ATECHPCS:
Thanks @Audacity88 — both addressed in fac82729.

**1. Sanitize + bound the `isError` detail**
The joined `result.content[].text` now goes through `zeroclaw_providers::sanitize_api_error(&detail)` once, *before* it reaches either the `zeroclaw_log::record!` `detail` attr or the `bail!`. I checked your "does it already truncate" question — `sanitize_api_error` already caps to `MAX_API_ERROR_CHARS` (500) and appends an ellipsis (providers/src/lib.rs:851-865), so that single call covers both the secret-scrubbing and the unbounded-payload concern; no separate truncation needed.

**2. Regression tests**
Added to the existing `#[cfg(test)] mod tests` via a fake `McpTransportConn` (no live server), exercising the inherent `McpServer::call_tool` — the method the PR actually modified, not the `McpRegistry::call_tool` wrapper at ~331:
- `call_tool_iserror_err_is_sanitized_and_bounded` — `isError:true` → `Err`; asserts a `sk-…` token is `[REDACTED]` and a 5000-char payload comes back truncated.
- `call_tool_success_returns_ok_result` — `isError` absent **and** `isError:false` → still `Ok(result)`.
- `call_tool_iserror_empty_detail_falls_back` — missing `content` and empty `text` → the `(no error detail returned by server)` fallback.

`cargo test -p zeroclaw-tools` (13 pass), `cargo clippy --all-targets -- -D warnings` (clean), and `cargo fmt` all green locally.

Audacity88 (approved):
I re-reviewed PR #7122 at current head `fac82729` against my prior request-changes review at `0195d30`, the author's response comment, the current diff, the PR body, labels, and the green GitHub checks. I did not run local cargo validation for this approval pass.

### ✅ Resolved — Earlier MCP blockers are addressed

The two earlier blockers are resolved. The `result.isError == true` detail is now passed through `zeroclaw_providers::sanitize_api_error` before it reaches either the structured log attribute or the returned `Err`, so the new server-controlled error surface is scrubbed and bounded. The new fake-transport tests also cover the important protocol cases: `isError:true` returns `Err` with sanitized/bounded detail, absent or false `isError` still returns `Ok(result)`, and empty detail uses the fallback message.

### 🟢 What looks good — The error boundary is now covered

This keeps the original narrow shape of the fix: JSON-RPC protocol errors still use the existing path, successful MCP tool results are preserved, and MCP tool-execution failures now become ordinary `call_tool` errors that callers already know how to handle. That is the right boundary for the issue this PR found.

This approval clears my earlier `CHANGES_REQUESTED` review. I do not see remaining author-side review work from that pass.

PR #7124, opened 2026-06-03
fix(docs): keep _shared chrome master-owned and retrofit version selector

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The versioned-docs deploy hoists each build's chrome (CSS/JS) into a single
    global `/_shared/` that every version's pages reference. Each deploy
    overwrote it unconditionally, so an older tag deploying after master
    (`v0.8.0-beta-2`) clobbered master's reskin with stale chrome — white code
    blocks and an unscrollable theme list across **all** versions. `_shared` is
    now written by **master deploys only**; tag deploys leave it untouched
    (bootstrap-seed only when absent).
  - Orphaned pre-versioned root trees (`/en/`, `/fr/`, `/api/`, …) lingered on
    `gh-pages` and served drop-down-less docs at `/en/`. New `prune-root`
    removes any root entry that isn't a real version dir or allowlisted file.
  - Tags built before the version-selector shipped render without a version
    dropdown. New `retrofit-selector` injects the selector `<script>` into any
    deployed page lacking it, at the correct `_shared` relative depth.
  - Concurrent deploys raced and desynced `versions.json`; push is now
    rebase-and-retry.
- **Scope boundary:** No docs *content* changes, no theme palette changes, no
  `book.toml` / `themes.json` changes. Only the deploy workflow and xtask tooling.
- **Blast radius:** `docs-deploy` workflow + `gh-pages` layout only. No runtime,
  crate, or product-surface code. `is_version_dir` is the single source of truth
  shared by `gen-versions` and `prune-root` so they cannot disagree.
- **Linked issue(s):** Related #7023 (versioned documentation deployment).
- **Labels:** `type: docs`, `docs`, `ci`, `risk: low`, `size: M`

```bash
cargo fmt --all -- --check                            # exit 0, no diff
cargo clippy -p xtask --all-targets -- -D warnings    # Finished, exit 0
cargo test -p xtask versions                          # ok. 2 passed; 0 failed
```

  - `cargo fmt --all -- --check` — exit 0 (clean, no diff).
  - `cargo clippy -p xtask --all-targets -- -D warnings` — `Finished dev profile`, exit 0.
  - `cargo test -p xtask versions` — `test result: ok. 2 passed; 0 failed; 0 ignored`
    (`shared_prefix_matches_page_depth`, `is_version_dir_accepts_only_real_versions`).
- **Beyond CI — what I manually verified:**
  - Ran `prune-root` + `retrofit-selector` against a materialized copy of the
    live `gh-pages` tree: pruned `en es fr ja zh-CN api`; kept `master`,
    `v0.8.0-beta-2`, `_shared`, `CNAME`, `index.html`, `versions.json`.
  - Retrofit patched beta-2 pages; **210 selector refs checked, 0 broken**; depth
    matches master ground truth exactly (`../../` for index pages, `../../../`
    for deep pages, verified across `en` / `fr` / `zh-CN`); idempotent (master
    pages untouched, exactly one ref per page).
  - Confirmed the live `_shared/theme/custom.css` was the stale 2342-byte file vs
    master's 27 KB reskin — the root cause of the white code blocks.
- **Not verified locally:** a full multi-locale `cargo mdbook build` + real
  `gh-pages` push (requires a CI runner); exercised via a live master dispatch.
- **Intentionally skipped:** full-workspace `cargo test` — the change is isolated
  to `xtask` tooling plus a workflow file, so the scoped `xtask` battery was run
  instead.

#### 1. Docs: master-owned `_shared` chrome + version selector (#7124) — Tier 0

Commits: `24b5d155f` (keep `_shared` master-owned, retrofit selector), `496fe4c30` (deploy hardening).

- [x] mdBook builds (`cargo run -p xtask --bin mdbook -- build`); selector wired into every page. RENDER+SWITCH verified end-to-end: seeded a 2nd version dir (`v0.7.5`), ran real `gen-versions` → `versions.json` (master + v0.7.5, stable=v0.7.5), served book root over HTTP, drove selector logic against live data → renders 2 entries, marks master current, switch `/master/en/index.html`→`/v0.7.5/en/index.html` resolves HTTP 200. Synthetic artifacts removed; tree clean.
- [x] `_shared` chrome master-owned: `extract_shared_chrome` hoists chrome to `book/_shared/`; pages reference it via depth-correct relative path.
- Evidence posted: PR #7124 comment (#7023 already MERGED — skipped).

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `Yes` — two new internal `xtask mdbook`
  subcommands (`prune-root`, `retrofit-selector`) consumed only by the deploy
  workflow. No user-facing CLI or config change.
- Upgrade steps for existing users: none. The effect applies on the next
  `docs-deploy` run.

Low risk — `git revert <sha>`. The workflow change is idempotent and the new
xtask subcommands are additive; reverting restores the prior deploy behavior.

---

## Operator note — true apex redirect (CNAME / edge)

GitHub Pages serves `gh-pages` as static files and **cannot emit a real HTTP
301/302 for the apex path** `https://docs.zeroclawlabs.ai/`. This PR keeps the
in-repo fallback (the root `index.html` meta-refresh + canonical to the default
version) so the bare domain still lands on the docs. That fallback is best-effort:
meta-refresh can be slow and is ignored by some clients.

To get a genuine `301`/`302` from `/` to `/master/en/` (or `/stable/en/` once a
stable ships), add the redirect at the DNS/edge layer in front of Pages. The
domain currently resolves through GitHub's own Fastly CDN (no Cloudflare in
front), so this is a one-time infra change outside the repo:

**Option A — Cloudflare in front of Pages (recommended):**
1. Move `docs.zeroclawlabs.ai` DNS to Cloudflare (proxied / orange-cloud).
2. Keep the `CNAME` file (`docs.zeroclawlabs.ai`) so Pages still accepts the host.
3. Add a Bulk Redirect or Redirect Rule:
   - When `URI Path` equals `/`
   - Then `301` to `https://docs.zeroclawlabs.ai/master/en/`
   - (Switch the target to `/stable/en/` when a stable version is deployed.)
4. Leave deeper paths (`/master/...`, `/_shared/...`, `/versions.json`)
   untouched — only the exact apex path redirects.

**Option B — keep Pages-only (status quo):** no edge change; the root
`index.html` meta-refresh remains the redirect. Acceptable but not a true 301.

This PR does not and cannot make the edge change itself; it only guarantees the
target (`/master/en/`) always exists, is pruned of orphans, and carries the
version dropdown. The 301 is an operator action on the DNS/CDN.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -199,17 +199,37 @@ jobs:
           echo "docs.zeroclawlabs.ai" > CNAME
 
           SCRIPTS="${GITHUB_WORKSPACE}/.workflow-scripts/.github/scripts"
-          
-          # Sync shared chrome — new xtask produces book/_shared/
-          if [ -d "${GITHUB_WORKSPACE}/docs/book/book/_shared" ]; then
-            mkdir -p "$CLONE_DIR/_shared"
-            rsync -a --delete "${GITHUB_WORKSPACE}/docs/book/book/_shared/" "$CLONE_DIR/_shared/"
-          else
-            # Old-tag fallback: extract chrome from the deployed version content
+
+          # Shared chrome (_shared/) is owned by master only. Every deployed
+          # version's pages reference /_shared/, so letting an older tag's build
+          # overwrite it would regress the chrome (CSS/JS) for ALL versions —
+          # the beta-2-clobbers-master bug. Tag deploys sync their own <tag>/
+          # dir but leave _shared untouched; master is the single writer.
+          if [ "$TAG" = "master" ]; then
+            if [ -d "${GITHUB_WORKSPACE}/docs/book/book/_shared" ]; then
+              mkdir -p "$CLONE_DIR/_shared"
+              rsync -a --delete "${GITHUB_WORKSPACE}/docs/book/book/_shared/" "$CLONE_DIR/_shared/"
+            else
+              # master build predates the _shared layout: extract chrome from it.
+              cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- extract-chrome "$CLONE_DIR/$TAG" "$CLONE_DIR/_shared"
+            fi
+          elif [ ! -d "$CLONE_DIR/_shared" ]; then
+            # Bootstrap only: a tag is the very first deploy and no master has
+            # populated _shared yet. Seed it from this tag so its pages aren't
+            # broken; the next master deploy takes ownership and overwrites it.
             cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- extract-chrome "$CLONE_DIR/$TAG" "$CLONE_DIR/_shared"
           fi
 
           cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- gen-versions > versions.json
+
+          # Remove orphaned root entries left by the pre-versioned layout
+          # (top-level en/, fr/, api/, ...). Keeps version dirs, _shared, and
+          # the allowlisted root files. Single source of truth: is_version_dir.
+          cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- prune-root
+
+          # Retrofit the version-selector into any deployed version whose pages
+          # predate the selector (e.g. tags cut before it shipped). Idempotent.
+          cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- retrofit-selector
 
           # Ensure root index.html points to /stable/en/ if stable docs exist, otherwise /master/en/
           if [ -d "stable" ]; then
@@ -220,5 +240,39 @@ jobs:
 
           git add --all
           git -c user.name="$GIT_AUTHOR_NAME" -c user.email="$GIT_AUTHOR_EMAIL" commit -m "docs: update docs for $TAG (from $GITHUB_SHA)" || true
-          git push "$REPO" gh-pages
+
+          # Concurrent deploy runs each clone, mutate, and push gh-pages. A run
+          # that cloned before a sibling's push would otherwise clobber it (or
+          # be rejected). Rebase onto the latest remote tip and retry so every
+          # deployed version and an up-to-date versions.json survive the race.
+          push_attempts=0
+          until git push "$REPO" gh-pages; do
+            push_attempts=$((push_attempts + 1))
+            if [ "$push_attempts" -ge 5 ]; then
+              echo "::error::gh-pages push failed after ${push_attempts} attempts"
+              exit 1
+            fi
+            echo "Push rejected (attempt ${push_attempts}); rebasing onto latest gh-pages and retrying"
+            git fetch --depth=50 "$REPO" gh-pages
+            git -c user.name="$GIT_AUTHOR_NAME" -c user.email="$GIT_AUTHOR_EMAIL" rebase FETCH_HEAD || {
+              echo "::error::rebase onto gh-pages failed"
+              git rebase --abort || true
+              exit 1
+            }
+            # Re-run finalizers after rebase: a sibling run may have added or
+            # removed a version dir, so versions.json / prune / retrofit must
+            # reflect the merged tree before we push.
+            cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- gen-versions > versions.json
+            cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- prune-root
+            cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- retrofit-selector
+            if [ -d "stable" ]; then
+              bash "${SCRIPTS}/gen-index-stable.sh"
+            else
+              bash "${SCRIPTS}/gen-index-master.sh"
+            fi
+            if ! git diff --quiet || ! git diff --cached --quiet; then
+              git add --all
+              git -c user.name="$GIT_AUTHOR_NAME" -c user.email="$GIT_AUTHOR_EMAIL" commit --amend --no-edit
+            fi
+          done
           popd >/dev/null
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -173,17 +173,25 @@ jobs:
             popd > /dev/null
           fi
 
-          # Copy built tag docs into gh-pages under '<tag>/'
-          mkdir -p "$CLONE_DIR/$TAG"
-          if [ -d "docs/book/book/$TAG" ]; then
-            rsync -a --delete "docs/book/book/$TAG/" "$CLONE_DIR/$TAG/"
+          # Normalize the build output into a single versioned source dir so the
+          # initial sync and every retry copy from the same place. Old tags that
+          # still build into the pre-versioned layout (`docs/book/book/` with no
+          # `$TAG/` inside) get wrapped into `$TAG/` once, here — the retry path
+          # must not re-derive this, or a concurrent push rejection on an old tag
+          # would leave it with nothing to re-apply.
+          BUILD_ROOT="${GITHUB_WORKSPACE}/docs/book/book"
+          if [ -d "$BUILD_ROOT/$TAG" ]; then
+            TAG_SRC="$BUILD_ROOT/$TAG"
           else
             echo "Old docs layout detected (unversioned). Restructuring to $TAG/..."
-            mkdir -p "docs/book/book_temp/$TAG"
-            rsync -a docs/book/book/ "docs/book/book_temp/$TAG/"
-            rsync -a --delete "docs/book/book_temp/$TAG/" "$CLONE_DIR/$TAG/"
-            rm -rf docs/book/book_temp
+            TAG_SRC="$(mktemp -d)/$TAG"
+            mkdir -p "$TAG_SRC"
+            rsync -a "$BUILD_ROOT/" "$TAG_SRC/"
           fi
+
+          # Copy built tag docs into gh-pages under '<tag>/'
+          mkdir -p "$CLONE_DIR/$TAG"
+          rsync -a --delete "$TAG_SRC/" "$CLONE_DIR/$TAG/"
 
           # If this is a stable release (no pre-release suffix), also copy to /stable/
           if [[ "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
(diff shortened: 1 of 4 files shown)


singlerider (commented):
Found some things. Pushing.

kitswas:
> Orphaned pre-versioned root trees (/en/, /fr/, /api/, …) lingered on gh-pages and served drop-down-less docs at /en/. New prune-root removes any root entry that isn't a real version dir or allowlisted file.

It could be simpler to manually delete everything on the branch once before deploying.

`prune-root` with an allowlist has benefits - it prevents other files/dirs from slipping in unnoticed. It is, however, more stuff to maintain and more potential failure points.

WareWolf-MoonWall (approved):
## Context

Reviewing PR #7124 by @singlerider as @WareWolf-MoonWall. This PR fixes a regression in the versioned documentation deployment where older tag deploys clobbered master's shared chrome (CSS/JS), causing white code blocks and broken theme selectors across all deployed doc versions. The fix introduces master-only ownership of `_shared/`, adds orphan pruning, retrofits the version selector to pages built before it existed, and hardens the push retry mechanism against concurrent deploys.

**Commits reviewed:** 24b5d155fde5fc95f1bdfd426e5372d75f2fcaa1, 496fe4c30d0c56c11027317188740f6531083054

---

### 🟢 What looks good — Self-review discipline

The two-commit structure shows effective self-review: the first commit implements the solution, the second (496fe4c3) hardens it after the author ran an adversarial review pass and found three real risks: (1) file-deletion landmine in `prune-root`, (2) shallow-clone ancestry issues in the push retry, and (3) depth-correctness bugs in the shared chrome refs. That second commit isn't cosmetic — it closes legitimate hazards the first commit introduced. This is the standard: you catch your own edge cases before the reviewer has to point them out.

---

### 🟢 What looks good — Single source of truth for version directories

`is_version_dir` in `versions.rs` is now the canonical answer to "is this a version dir?" shared by both `gen-versions` (the `versions.json` generator) and `prune-root` (the orphan cleaner). They can never drift. The old `run()` logic duplicated this check inline; extracting it was the right call. The accompanying unit test (`is_version_dir_accepts_only_real_versions`) validates the boundary exactly — `master`, `stable`, `vX.Y.Z` pass; `en`, `fr`, `main`, `v1.2` (incomplete semver) fail. Clean.

---

### 🟢 What looks good — Depth-correct shared chrome references

The `extract_shared_chrome` depth math (lines 194–208 of `build.rs`) now computes the `_shared` relative prefix per HTML file based on its depth below the locale dir, not a hardcoded `../../`. This was a latent bug: deep pages in subdirectories received a shallow prefix, so their chrome refs 404'd once `_shared` became master-owned. The fix counts path components (`page_depth`) and generates `../` strings that match. The unit test `shared_prefix_matches_page_depth` validates the math across representative cases (`./master/en/index.html` → `../../`, `v0.8.0-beta-2/en/architecture/crates.html` → `../../../`). The normalization correctly strips `./` prefixes so `read_dir(".")` doesn't inflate the depth.

---

### 🟢 What looks good — Master-only `_shared` ownership

The workflow change (lines 208–221 of `docs-deploy.yml`) gates `_shared/` writes behind a `TAG == master` guard. Tag deploys sync only their own `<tag>/` directory and leave `_shared` alone. This is the correct fix for the regression: an older tag deploying after master can no longer clobber the chrome with stale content. The bootstrap path (lines 218–221) handles the edge case where a tag is the first-ever deploy: it seeds `_shared` so its pages aren't broken, but the next master deploy reclaims ownership. Clean separation of concerns.

---

### 🟢 What looks good — Idempotent retrofit selector

`retrofit_selector` (lines 97–115 of `versions.rs`) injects the version-selector `<script>` into pages that lack it, so tags built before the selector shipped now render the dropdown. The logic is defensive: skips pages that already have the ref (idempotent), skips pages without a `right-buttons` menu bar (no attach point), bails early if it can't find `</body>`. The insertion point (`rfind("</body>")` plus line-start backtrack) puts the script where the other chrome scripts live, so it loads in the expected order. No mutation unless needed.

---

### 🟢 What looks good — Tree-merge retry against concurrent deploys

The push retry loop (lines 254–290 of `docs-deploy.yml`) replaces the shallow-clone rebase with a tree-merge: on rejection, hard-reset onto the latest remote tip (adopting every sibling run's version dirs verbatim), re-apply only this run's own deliverables (its `<tag>/`, and `_shared` when master), re-run the finalizers (idempotent), and commit fresh. This works regardless of shared ancestry or large gaps — merges by tree, not history. The 5-attempt limit is reasonable for a deployment workflow; exit 1 on exhaustion triggers a workflow failure the operator can see.

---

### 🟢 What looks good — Prune root directories only, never files

`prune-root` (lines 67–97 of `versions.rs`) removes orphaned *directories* (the pre-versioned `en/`, `fr/`, `api/` trees) but never touches root files. The comment explains why: a closed file allowlist would silently delete legitimate future additions (`404.html`, `robots.txt`, `sitemap.xml`). The current orphans are all directories; this design doesn't create a deletion landmine for files. The logic walks `read_dir(".")`, skips non-dirs, and keeps anything in `ROOT_KEEP_DIRS` (`_shared`, `.git`) or recognized as a version dir.

---

The validation section in the PR body is exemplary: scoped commands (`cargo clippy -p xtask`, `cargo test -p xtask versions`), literal exit codes and test counts, and a detailed beyond-CI section documenting the manual `prune-root` / `retrofit-selector` run against a materialized `gh-pages` clone — pruned the expected orphans, patched 210 pages with zero broken refs, verified depth correctness across multiple locales. The "not verified locally" note (full multi-locale build + real push) is honest about what requires a CI runner, and the "intentionally skipped" rationale (full-workspace `cargo test` is unnecessary for isolated xtask changes) follows the protocol. This is what "local validation is the signal CI cannot replace" means.

---

### 🔵 Suggestion — Consider logging the prune/retrofit counts at workflow runtime

The new `prune-root` and `retrofit-selector` xtask commands print their actions to stdout (`println!("prune-root: removing orphaned dir {name}/")`; `println!("retrofit-selector: patched {patched} page(s)")`). At workflow runtime these go to the Actions log, but they're easy to miss in a dense output stream. If these counts matter for operational visibility (e.g., detecting unexpected orphan growth or confirming the retrofit ran), consider prefixing the messages with a structured marker (`::notice::` or a `[DOCS]` tag) so they're scannable in the log UI. Not blocking — the current output is sufficient for post-hoc debugging — but it's a cheap way to surface the signal when you're troubleshooting a deploy.

---

### 🔵 Suggestion — Apex redirect operator note is actionable, but consider a tracking issue

The PR body's "Operator note" section (lines 33–54) explains the GitHub Pages limitation (no real HTTP 301/302 from the apex `/`), documents the current fallback (meta-refresh in the root `index.html`), and provides a step-by-step Cloudflare edge-redirect recipe. That's the right place for this info — it's deployment context, not code documentation. However, it's buried in a closed PR body once this merges. If the team wants to track the apex-redirect infra action separately (so it doesn't get lost), consider opening a follow-up issue in the infrastructure repo (or a docs-ops tracking issue here) with the Cloudflare recipe pasted in. Then reference it in the merge commit message or the PR comment thread. Not blocking — the PR is complete without it — but it's a way to convert "operator note buried in PR body" into "tracked work with a discoverable issue number."

---

## Verdict

✅ **Approved.** This PR solves the regression (older tag deploys clobbering master chrome), closes the edge cases the author found via self-review, and ships with comprehensive validation evidence. The design is sound: master owns `_shared/`, tags are hands-off, concurrent deploys merge by tree, orphans are pruned, old pages get the selector retrofitted. The code is clean, the tests are sufficient, and the commit messages document the reasoning. No blockers. Two non-blocking suggestions above for operational visibility and follow-up tracking; both are optional.

CI is green, template is complete, labels are correct (`risk: low`, `type: docs`, `ci`, `size: M`). Already assigned to the v0.8.0 milestone (correct target — config/schema tier promotion scope includes docs tooling hardening). Ready to merge.

singlerider:
## Manual validation — versioned docs + master-owned `_shared` chrome

Verified end-to-end on `integration/zerocode-test` against a real build.

**Build + chrome ownership**
- `cargo run -p xtask --bin mdbook -- build` succeeds; output split into `book/master/` + `book/_shared/`.
- `extract_shared_chrome` hoists chrome (`version-selector.js`, `lang-switcher.js`, css, favicons) into `book/_shared/theme/`; every content page references it via a depth-correct relative `_shared` path. Confirmed in served HTML the page loads `_shared/theme/version-selector.js`.

**Version selector render + switch**
1. Seeded a second version dir (`cp -r master v0.7.5`).
2. Ran the real `xtask mdbook gen-versions` → `versions.json`:
   ```json
   { "stable": "v0.7.5",
     "versions": [
       { "label": "Development (master)", "tag": "master" },
       { "label": "v0.7.5", "tag": "v0.7.5" } ] }
   ```
3. Served the book root over HTTP — `versions.json`, `master/en/`, `v0.7.5/en/` all HTTP 200.
4. Exercised the selector logic against the live served data:
   - renders 2 dropdown entries, marks `master` current;
   - switch `master` to `/v0.7.5/en/index.html` resolves HTTP 200 (same page, other version).

   Result: **PASS**.

Caveat: dropdown DOM verified by running the exact selector JS path against live data, not via a headless browser render (no agent-browser backend available in this environment). Synthetic version dir + `versions.json` removed afterward; tree clean.

Audacity88 (dismissed):
Reviewed the live PR metadata, prior comments/reviews, and the current diff at `496fe4c3`. I focused on the docs-deploy workflow behavior, especially old-tag dispatch, shared chrome ownership, root pruning, and the new push retry path. I did not run the docs workflow locally.

### 🟢 What looks good — The main docs-deploy shape is right

The master-owned `_shared/` model, directory-only root pruning, selector retrofit, and depth-correct shared chrome references all look like the right direction. The remaining issue is narrower: the retry path does not preserve one fallback that the initial deploy path still supports.

### 🔴 Blocking — Retry path drops the old-layout fallback

The first deploy path handles old tags that still build into the pre-versioned layout: if `docs/book/book/$TAG` does not exist, it wraps `docs/book/book/` into `$TAG/` before syncing. The new retry path does not repeat that fallback. After a rejected `gh-pages` push, it resets to the latest remote tree and then always runs:

```bash
rsync -a --delete "${GITHUB_WORKSPACE}/docs/book/book/$TAG/" "$TAG/"
```

For a `workflow_dispatch` run against an old tag that still emits unversioned `docs/book/book/`, any concurrent push rejection will make the retry fail or fail to re-apply this run's tag tree. That breaks the old-tag plus concurrent-deploy case this PR is trying to make safe. The retry path should use the same old-layout fallback as the initial sync, or materialize one normalized `$TAG/` output directory before the first push and reuse that source for both initial sync and retries.

singlerider:
Addressed the blocker on `beb2c1dd9`:

- **🔴 Retry path dropped the old-layout fallback** — the initial sync wrapped old-tag builds (pre-versioned `docs/book/book/` with no `$TAG/` inside) into `$TAG/`, but the post-rejection retry always rsynced from `docs/book/book/$TAG/` directly, so a `workflow_dispatch` run against an old tag would fail or re-apply nothing on any concurrent push rejection.

  Took the "materialize one normalized `$TAG/` source" approach you suggested: a single `TAG_SRC` is computed once after the build — the existing `$TAG/` dir for versioned builds, or a wrapped copy (into a `mktemp -d` dir outside the build root to avoid recursive self-copy) for the old layout. Both the initial sync and every retry now `rsync` from `TAG_SRC`, so the old-layout fallback survives concurrent-deploy rejection.

Validation: workflow YAML parses (`yaml.safe_load`), and the extracted `run` script passes `bash -n`.

Re-requesting review.

PR #7129, opened 2026-06-03
fix(tools): fail loudly when file_write targets an ephemeral workspace

### LINKED ISSUE #4627: [Bug]: file_write tool silently fails — written files are invisible on the host filesystem
### Affected component

runtime/daemon

### Severity

### Current behavior

The file_write tool reports success and returns no error, but the files it writes are completely invisible on the host filesystem. A full recursive search (sudo find / -name <filename>) on the host machine returns no results. The agent's shell tool exhibits the same behavior — files written via echo, cat, tee, or cp appear to exist within the agent's environment (confirmed via ls -lh and cat inside the session), but do not exist on the host at any path.

Additionally, mount output inside the agent session shows /home/ubuntu/.zeroclaw/workspace mounted as a tmpfs. This tmpfs is private to the agent's execution environment and is not mapped to any host path. The agent has no visibility into this isolation — it believes writes are succeeding.

Image

Image

Image

Image

### Expected behavior

Files written via file_write or shell should be persisted to a location accessible on the host filesystem, either by:
Mounting the workspace directory from the host into the agent container/sandbox (e.g., -v /host/path:/home/ubuntu/.zeroclaw/workspace), or
Clearly documenting that the workspace is ephemeral and isolated, so the agent can inform the user accordingly.
At minimum, if writes cannot reach the host, the file_write tool should return an explicit error rather than silently succeeding.

### Steps to reproduce

```bash
Start a ZeroClaw agent session.
Ask the agent to write any content to a file inside the workspace:
Agent reports success. Inside the session, ls -lh and cat confirm the file exists.
On the host machine, run:
No results are returned. The file does not exist on the host.
Additional diagnostic commands run inside the agent session:
# Confirms workspace is a private tmpfs
mount | grep workspace
# Output: tmpfs on /home/ubuntu/.zeroclaw/workspace type tmpfs (rw,...)


# Confirms no .dockerenv but overlay rootfs is present
cat /proc/1/cgroup
ls /.dockerenv
findmnt /
```

### Impact

_No response_

### Logs / stack traces

```text

```

### ZeroClaw version

v0.6.2

### Rust version

_No response_

### Operating system

Ubuntu 24.04.3 LTS (GNU/Linux 6.17.0-1009-aws x86_64)

### Regression?

No, first-time setup

### Pre-flight checks

IftekharUddin:
Closes #4627.

> [!NOTE]
> This PR closes the S0 by guarding **every workspace surface the issue describes**, not just `file_write`. `file_write` refuses outright (it exists only to persist data); the general-purpose `shell`, `file_read`, and `file_edit` tools — plus the other workspace-write tools `file_download` and `image_gen` — stay usable on an ephemeral runtime but attach a loud, actionable warning so writes that won't reach the host (and reads that may be stale) are never silent.

## Problem

On a Docker runtime with **no host volume mount**, the agent's workspace is a private in-container tmpfs. `DockerRuntime::has_filesystem_access()` returns `self.config.mount_workspace`, so on the reported config the flag is `false`. Writes succeed *inside the container* but are invisible on the host and discarded at session end; reads may return stale/empty data. Every tool reported `success` regardless — the silent data loss reported in #4627.

## Fix — one persistence flag, threaded into every workspace tool

All filesystem tools learn at registration time whether the runtime persists writes, via `runtime.has_filesystem_access()` (forwarded in both `default_tools_with_runtime` and `all_tools_with_runtime`):

| Tool | Behaviour when workspace is ephemeral |
|------|----------------------------------------|
| `file_write` | **Refuses** with an explicit error before any write (its sole purpose is to persist data). Unchanged from the original fix. |
| `file_edit` | **Warns** on successful edits (the rewritten file won't persist). |
| `file_download` | **Warns** on successful downloads (the fetched file won't persist). |
| `image_gen` | **Warns** on successful generation (the saved PNG won't persist). |
| `shell` | **Warns** on every executed command — it can't tell a read (`ls`) from a write (`echo > f`), so blocking would break legitimate read-only use. |
| `file_read` | **Warns** on successful **text** reads (data may be stale/empty). base64 reads are exempt — the caller decodes the payload verbatim, so a banner would corrupt it. |

The warning names the cause and the exact config key to fix it: `runtime.docker.mount_workspace = true`.

A shared `zeroclaw_api::tool::{EPHEMERAL_WORKSPACE_WARNING, with_ephemeral_workspace_warning}` keeps the message identical across crates and makes extending the guard to any future workspace tool a one-liner.

### Dispatcher-aware injection

The tool dispatcher forwards only `output` to the model on success, but `error` (falling back to `output`) on failure. So the warning is injected into whichever field will actually be shown — for `shell` that means **both** `output` and `error`, so the banner survives whether the command succeeded or failed.

### A deliberate note on the `file_write`↔`shell` asymmetry

`file_write` blocks while `shell` only warns, so an agent *could* still lose data by writing through `shell` (e.g. `echo … > f`). That's intentional and unavoidable: `shell` is general-purpose and can't reliably distinguish reads from writes, so the issue's own "inform the user accordingly" guidance is satisfied with a loud, unbypassable warning rather than a block that would break read-only shell use. This matches option 2 from review — "an equivalent early signal."

## Changes

- `crates/zeroclaw-api/src/tool.rs` — shared `EPHEMERAL_WORKSPACE_WARNING` const + `with_ephemeral_workspace_warning()` helper (+ unit tests).
- `crates/zeroclaw-tools/src/file_write.rs` — (existing) `persistent_writes` refusal guard.
- `crates/zeroclaw-tools/src/file_edit.rs` — `new_with_persistence()`; warn on successful edits.
- `crates/zeroclaw-tools/src/file_download.rs` — `new_with_persistence()`; warn on successful downloads.
- `crates/zeroclaw-tools/src/image_gen.rs` — `new_with_persistence()`; warn on successful generation.
- `crates/zeroclaw-runtime/src/tools/shell.rs` — `with_persistent_writes()`; warn on every executed command (output + error).
- `crates/zeroclaw-runtime/src/tools/file_read.rs` — `new_with_persistence()`; warn on successful text reads (base64 exempt).
- `crates/zeroclaw-runtime/src/tools/mod.rs` — forward `has_filesystem_access()` into all tools at both registration sites; end-to-end wiring test.

All constructors keep their original `new()` defaulting to `persistent_writes = true`, so native runtimes and existing callers/tests are unaffected.

## Verification

```
cargo test -p zeroclaw-api  --lib                 # 47 passed
cargo test -p zeroclaw-tools --lib                # 1308 passed
cargo test -p zeroclaw-runtime --lib -- tools::   # 310 passed
cargo fmt --check && cargo clippy --all-targets   # clean
```

New regression tests:
- **api**: `ephemeral_warning_names_cause_and_fix`, `empty_text_returns_banner_alone`, `nonempty_text_keeps_body_below_banner`
- **shell**: `shell_warns_on_ephemeral_workspace`, `shell_warns_on_ephemeral_workspace_failure_path`, `shell_warns_on_ephemeral_success_with_stderr`, `shell_no_warning_when_persistent`
- **file_read**: `file_read_warns_on_ephemeral_workspace`, `file_read_base64_not_warned_on_ephemeral_workspace`, `file_read_failure_not_warned_on_ephemeral_workspace`, `file_read_no_warning_when_persistent`
- **file_edit**: `file_edit_warns_on_ephemeral_workspace`, `file_edit_failure_not_warned_on_ephemeral_workspace`, `file_edit_no_warning_when_persistent`
- **file_download**: `execute_warns_on_ephemeral_workspace`
- **registration wiring (mod.rs)**: `registered_tools_warn_or_block_on_ephemeral_runtime` — proves `has_filesystem_access()` is threaded end-to-end (shell/file_read/file_edit warn; file_write refuses).

## What was not tested

Live Docker round-trip with a real bind-mount absent (deterministic from the `has_filesystem_access()` flag path) and `image_gen`'s success path (requires a live fal.ai call); the latter's warn follows the identical, unit-tested helper pattern.


IftekharUddin:
[Issue4627.pdf](https://github.com/user-attachments/files/28560835/Issue4627.pdf)


Additional documentation explaining the issue and fix.

singlerider (changes requested):
Thanks for this, @IftekharUddin — I read the PR, the linked issue #4627, the attached PDF context, and traced the fix against the runtime adapters. The mechanism is correct and the wiring is exactly right: `DockerRuntime::has_filesystem_access()` returns `self.config.mount_workspace` (crates/zeroclaw-config/src/platform/docker.rs:64), so on the reported "Docker, no volume mount" configuration the flag is `false` and your guard fires before any write. I confirmed nothing else currently gates the file tools on that flag, so this is net-new enforcement rather than a redundant or conflicting check, and `cargo test -p zeroclaw-tools -- file_write` is green locally (23 passed).

One blocking item, about the issue linkage rather than the code itself.

### 🔴 Blocking — `Closes #4627` over-claims; the issue's shell path is still silently lossy

Issue #4627 is filed S0 and explicitly states the silent-loss behavior is not limited to `file_write`:

> The agent's shell tool exhibits the same behavior — files written via echo, cat, tee, or cp appear to exist within the agent's environment ... but do not exist on the host at any path.

This PR fixes `file_write` only. `ShellTool` (constructed at crates/zeroclaw-runtime/src/tools/mod.rs:248 and :482) carries no persistence/ephemeral awareness, so an agent that runs `echo … > file` on an unmounted Docker tmpfs still gets `success` with the data discarded at session end — the exact S0 scenario the issue leads with. `file_read` against the same tmpfs is similarly able to return stale/empty data without signalling the cause.

The code change here is sound and worth landing on its own. The blocker is the auto-close: `Closes #4627` will mark a data-loss S0 resolved while the shell surface it describes is untouched. Please either:

- change the linkage to `Related #4627` (non-closing) and leave the issue open to track the shell/read paths, or
- keep `Closes #4627` and extend this PR to cover the shell path (thread `has_filesystem_access()` into `ShellTool` the same way, with an equivalent early signal).

Either is fine — that's an author/scope call. What shouldn't happen is the S0 silently closing half-fixed.

### 🟢 What looks good — focused, backward-compatible, honestly tested

- `FileWriteTool::new()` is preserved and defaults `persistent_writes = true`, so every native runtime and existing test is unaffected; the new behavior is opt-in through `new_with_persistence()` at the two registration sites. That's the right way to add a guard without churning callers.
- The guard returns before any filesystem mutation and names both the cause and the exact config key to fix it (`runtime.docker.mount_workspace = true`). Actionable error text over a silent failure is precisely the fix the issue asks for.
- `file_write_blocks_ephemeral_runtime` exercises the real `persistent_writes = false` path and asserts all three of: no file written, `success = false`, and the cause string present. The "What was not tested" note (live Docker round-trip) is an honest disclosure, and the fix is deterministic from the flag, so I'm satisfied without it.

IftekharUddin:
@singlerider just to confirm: 

unlike file_write, it can't tell a harmless read (ls, grep) from a write, so blocking it entirely to "fail loudly" would also break legitimate read-only use in Docker-no-mount setups. So the choice is whether to hard-block shell (loud and consistent, but heavier-handed) or just prepend a warning and let it run (preserves read-only use, but writes still appear to succeed). 

I'm leaning towards prepending a warning and letting it run but it's your call. I can widen the scope to see if we can't somehow determine the read/write so that it only triggers on the write.

Audacity88:
@IftekharUddin Thanks for making the relationship non-closing. I agree with keeping this PR `file_write`-only. That makes #7129 a useful partial fix for #4627 without trying to settle the shell contract in the same diff.

For the shell path, I would not widen this PR unless @singlerider specifically wants that. A warning-only shell change would be useful diagnostic surface, but it would not close #4627 because shell write commands or redirections could still return success while writing only to the ephemeral workspace. Hard-blocking all shell in no-mount Docker would be loud and consistent, but it would also break legitimate read-only shell use.

So the clean split from my side is:

- land #7129 as the `file_write` guard with `Related #4627`;
- leave #4627 open for the shell/read follow-up;
- scope that follow-up separately around the intended shell contract, likely either a clear runtime-level warning plus docs, or a narrower write-command guard if we can make the classification reliable.

I’ll defer to @singlerider on whether this resolves the active `CHANGES_REQUESTED` review, but I would not ask you to expand this PR just to answer the shell design question.

IftekharUddin:
Thanks for the careful review — I've taken option 2: changed the linkage back to Closes #4627 and extended the guard from file_write to every other workspace surface the issue describes, so the S0 no longer half-closes. file_write still refuses outright, while shell, file_read, file_edit, file_download, and image_gen stay usable but attach a loud, actionable warning (naming runtime.docker.mount_workspace = true) via a shared zeroclaw_api::tool helper threaded from has_filesystem_access(); for shell the banner is injected into both output and error so the dispatcher never drops it, and file_read base64 reads are exempt to avoid corrupting the decodable payload. The one deliberate asymmetry: shell warns rather than blocks because it can't reliably tell a read from a write — an "equivalent early signal" rather than a hard refusal — and there are regression tests for every surface plus an end-to-end check that the persistence flag is actually wired through registration.

singlerider (approved):
Re-reviewing at 6f4be1fb2 against my earlier `CHANGES_REQUESTED` at 82323eb. I read the expanded diff (+860/-65 across 8 files), the new `zeroclaw_api::tool` helper, the registration wiring in `tools/mod.rs`, and the per-surface guards. I re-ran the validation battery locally against the current head — all green (details below). Thank you for taking this all the way rather than half-closing the S0.

### ✅ Resolved — `Closes #4627` no longer over-claims

My blocker was that `Closes #4627` would mark a data-loss S0 resolved while the shell, read, and other workspace surfaces the issue explicitly describes stayed silently lossy. You took option 2 and threaded `has_filesystem_access()` into every workspace tool at registration time, so the issue closes fully:

- `file_write` refuses outright before any write (its sole purpose is to persist data).
- `shell`, `file_read`, `file_edit`, `file_download`, `image_gen` stay usable but attach a loud, actionable banner naming the cause and the exact fix (`runtime.docker.mount_workspace = true`).

The shared `EPHEMERAL_WORKSPACE_WARNING` / `with_ephemeral_workspace_warning` helper in `zeroclaw-api` is the right call — one canonical message, and extending the guard to a future workspace tool is a one-liner rather than a copy-paste.

### ✅ Resolved — the file_write-blocks / shell-warns asymmetry is the correct contract

This is exactly the "equivalent early signal" I named as acceptable in my original review. `shell` is general-purpose and cannot reliably distinguish a read (`ls`) from a write (`echo > f`), so a hard block would break legitimate read-only use on a no-mount Docker runtime. A loud, unbypassable warning satisfies the issue's "inform the user accordingly" guidance without that collateral. The reasoning is documented at the call sites and the asymmetry is deliberate, not an oversight.

### 🟢 What looks good — the dispatcher-aware injection is precise

The shell path injects the banner into both `output` and `error`, because the dispatcher forwards `output` on success and `error` on failure — so the warning survives whichever field the model actually reads. `shell_warns_on_ephemeral_success_with_stderr` proves the dual-field branch (exit 0 + stderr present). That is the kind of edge the warning would otherwise silently fall through.

### 🟢 What looks good — base64 read exemption avoids corrupting decodable payloads

`file_read` skips the banner for base64 reads because the caller decodes the payload verbatim, and `file_read_base64_not_warned_on_ephemeral_workspace` asserts the bytes still round-trip exactly. Prepending a banner there would have been a real bug; catching it in a test is the right instinct.

### 🟢 What looks good — backward compatibility and honest test coverage

Every constructor keeps its original `new()` defaulting `persistent_writes = true`, so native runtimes and existing callers/tests are untouched; the new behavior is opt-in through `new_with_persistence()` / `with_persistent_writes()` at the registration sites. Failure paths are correctly exempt (a failed write/read is not data loss, so no banner). `registered_tools_warn_or_block_on_ephemeral_runtime` is the load-bearing test: it proves `has_filesystem_access()` is actually threaded end-to-end through `default_tools_with_runtime`, not just that the per-tool flag works in isolation. The "What was not tested" note (live Docker round-trip, `image_gen` fal.ai success path) is an honest disclosure, and both are deterministic from the flag path the unit tests already cover.

### Local validation (current head 6f4be1fb2)

- `cargo clippy -p zeroclaw-api -p zeroclaw-tools -p zeroclaw-runtime --all-targets -- -D warnings` → `Finished` (no warnings)
- `cargo test -p zeroclaw-api --lib` → `47 passed; 0 failed`
- `cargo test -p zeroclaw-tools --lib` → `1309 passed; 0 failed`
- `cargo test -p zeroclaw-runtime --lib -- tools::` → `310 passed; 0 failed; 1 ignored`
- `cargo test -p zeroclaw-runtime --lib -- registered_tools_warn_or_block_on_ephemeral_runtime` → `1 passed`

---

**Verdict:** `--approve`. My prior blocker is resolved, the scope now matches what #4627 actually describes, and the validation reproduces clean on the current head. Good work.

Audacity88 (approved):
Re-reviewing PR #7129 at 6f4be1fb against the earlier scope blocker on 82323eb. I checked the PR metadata, linked issue #4627, prior reviews/comments, the public diff across the eight changed files, and the public check state. This approval is based on that public evidence.

### ✅ Resolved — #4627 closure now covers the shell path too

I re-read the earlier blocker, the current diff, and the linked issue. The first version only made `file_write` fail loudly, which left the issue's shell-write data-loss path unresolved. At 6f4be1fb, the diff threads the runtime persistence signal into the registered workspace tools, keeps `file_write` as a hard refusal, and adds loud warnings for the general-purpose surfaces where blocking would be too blunt. With that expanded scope, `Closes #4627` is accurate from my side.

### 🟢 What looks good — the warning follows the dispatcher contract

The shared `EPHEMERAL_WORKSPACE_WARNING` helper keeps the message consistent, and the per-tool behavior is nicely tailored: `file_write` returns before any filesystem mutation, `file_read` preserves base64 payloads by skipping the banner there, and `shell` injects the banner into both `output` and `error`, matching the PR's stated dispatcher contract. That is the right tradeoff for a shell tool that cannot reliably classify reads vs writes.

### 🟢 What looks good — registration wiring is tested, not just direct constructors

The end-to-end registration test is the important bit here: it proves `runtime.has_filesystem_access()` is actually converted into the per-tool persistence behavior through `default_tools_with_runtime`, not only in hand-constructed unit tests.
The public checks are green on the current head, and the PR body calls out the remaining untested live-Docker and `image_gen` success paths honestly.

Approving. The earlier shell/read-surface blocker is resolved, and the closing issue claim now matches the shipped scope.

Issue #5807, opened 2026-04-16
[Bug]: MiniMax-M2.7 (and most non-OpenAI/Anthropic/Google models) missing from default pricing catalog


JordanTheJet:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

When running the agent with `default_provider = "minimax"` and `default_model = "MiniMax-M2.7"`, every LLM call logs:

```
DEBUG zeroclaw_runtime::agent::cost: Cost tracking recorded token usage with zero pricing (no pricing entry found) provider="minimax" model="MiniMax-M2.7"
```

The cost tracker records zero cost for every turn because no pricing entry exists for `MiniMax-M2.7` (or any MiniMax model) in the default pricing catalog at `crates/zeroclaw-config/src/schema.rs:2008`.

Inspection of `get_default_pricing()` shows only 9 entries, all under `anthropic/`, `openai/`, or `google/` namespaces. None of the documented 50+ providers (MiniMax, MoonShot, GLM/Z.AI, Qwen, Doubao, Groq, Together, Fireworks, Cohere, etc.) ship with default pricing.

### Expected behavior

Either:
1. Default pricing catalog ships with entries for the major supported providers (at minimum: MiniMax, MoonShot/Kimi, GLM, Qwen, Doubao, Groq), OR
2. The catalog is loaded from an external source (the official provider pricing pages, a community-maintained JSON, or a build-time script) so it stays current without manual maintenance, OR
3. The "no pricing" log is upgraded to a single `WARN` per (provider, model) pair on first sight, with a pointer to where users can add their own entry.

Without pricing, `cost.enforcement.route-down-model`, budget gating, and any downstream cost-aware logic cannot function for the affected provider/model.

### Steps to reproduce

```bash
# config.toml
# api_key = "minimax"
# default_provider = "minimax"
# default_model = "MiniMax-M2.7"

export MINIMAX_API_KEY=<your key>
RUST_LOG=zeroclaw_runtime=debug ./target/release/zeroclaw agent -m "list files in my home dir"
# Observe the "zero pricing" debug line for every LLM call
```

### Impact

Affected users: anyone using a non-OpenAI/Anthropic/Google provider (MiniMax, MoonShot, GLM, Qwen, Doubao, Groq, Together, Fireworks, etc.) — i.e. most users of zeroclaw's "50+ providers" advertised support.
Frequency: every LLM call.
[...]

PR #7136, opened 2026-06-03
feat(providers): add Kilo AI Gateway as first-class model provider with pricing capture

Nillth:
## Summary

- **What changed and why:**
  - Added Kilo AI Gateway (`kilo`) as a first-class OpenAI-compatible model provider, registered in the provider factory with config schema support. Users can configure Kilo providers via `[providers.models.kilo.*]` in config.toml.
  - Extended `ModelEntry` to capture pricing from model catalogs (`/models` endpoint). Previously only OpenAI-compatible providers with pricing metadata returned usable cost rates; now Kilo (and OpenRouter) fetch and display pricing automatically in the web UI cost rates editor.
  - Fixed a race condition in the cost rates add flow where the model catalog was fetched on component mount (stale by the time the user submitted a new rate). Pricing is now fetched on-demand at submit time.
- **Scope boundary:** This PR does not add new tools, channels, or agent capabilities. It only adds a provider registration path and pricing plumbing.
- **Blast radius:** `zeroclaw-providers` (factory dispatch), `zeroclaw-config` (schema additions), `zeroclaw-gateway` (config API), web UI (cost rates editor). Existing providers are unaffected — this is additive.
- **Linked issue(s):** Related #5807 (missing pricing from non-OpenAI/Anthropic/Google models — same pricing capture pattern)
- **Labels:** `type: feature`, `risk: low`, `size: M`

```bash
cargo fmt --all -- --check
cargo test
```

  - `cargo fmt --all -- --check` → passed (clean)
  - `cargo clippy --all-targets -- -D warnings` → passed, compiled all crates in 2m 16s
  - `cargo test` → passed: `test result: ok. 388 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out`
    - Unit tests across all workspace crates: ok
    - Integration tests (159): ok
    - Architecture tests (79): ok
    - System tests (9): ok
    - Live tests (7): ignored (require live credentials)
- **Beyond CI — what did you manually verify:** Kilo provider appears in `/config/providers.models` list; pricing populates in cost rates editor for Kilo models; existing providers (OpenAI, OpenRouter, Anthropic) unaffected.
- **If any command was intentionally skipped, why:** _(none skipped)_

- New external network calls? `Yes` — Kilo provider makes HTTPS calls to `https://api.kilo.ai/api/gateway` (configurable endpoint). Same pattern as existing OpenAI-compatible providers.
- Secrets / tokens / credentials handling changed? `No` — Kilo API keys use the same encrypted config storage as all other providers.
- If any `Yes`, describe the risk and mitigation: External network calls are scoped to the configured provider endpoint only; no new attack surface beyond the existing provider pattern.

## Compatibility (required)

- Backward compatible? `Mostly` — one deliberate naming migration: the `kilo` shorthand previously resolved to the KiloCLI subprocess provider and now names the Kilo AI Gateway provider. The CLI provider keeps its canonical `kilocli` slot and the `kilo-cli` synonym. Documented in the provider catalog (docs/book/src/providers/catalog.md, "Naming migration" note).
- Config / env / CLI surface changed? `Yes` — new `[providers.models.kilo.*]` config section, plus the naming migration above.
- If `No` or `Yes` to either: exact upgrade steps for existing users: users who configured the CLI provider under the `kilo` shorthand switch to `kilocli` (or `kilo-cli`); gateway adopters add a `[providers.models.kilo.<alias>]` block. Everyone else is unaffected.

## Rollback

- **Fast rollback command/path:** `git revert` (low risk — purely additive)
- **Feature flags or config toggles:** `None` — removing the `[providers.models.kilo]` section from config disables the provider without code changes.
- **Observable failure symptoms:** If Kilo endpoint is unreachable, provider init logs an error at startup and the provider is excluded from the available list. No impact on other providers.

## Supersede Attribution

_(not applicable)_

---

**Commits in this PR:**
- `055a4b7ad` feat: add Kilo AI Gateway as first-class model provider
- `2e4fa297f` feat: capture and display pricing from model catalogs
- `6057a2a83` fix: fetch pricing on-demand in cost rate add (avoid stale state race)
- `5da6ea909` chore: cargo fmt on 7 modified files


WareWolf-MoonWall (changes requested):
# Review: PR #7136 — feat(providers): add Kilo AI Gateway as first-class model provider with pricing capture

**PR:** zeroclaw-labs/zeroclaw#7136  
**Author:** @Nillth  
**Head commit:** `f1ca6d78d0c3ef94d72c17c4e14beed520c5701c` (4 commits)  
**Files changed:** 19 (+463/-42)

---

## Context

This PR adds Kilo AI Gateway as a first-class OpenAI-compatible provider and extends the provider catalog system to capture and surface pricing data from model `/models` endpoints. The work spans three concerns:

1. **Provider registration** — Kilo joins the typed slot system with its own `KiloModelProviderConfig` and `CompatFamilySpec` impl
2. **Pricing plumbing** — `ModelProvider::list_models_with_pricing()` trait method, pricing structs in `zeroclaw-api`, catalog fetch in `compatible.rs` and `openrouter_catalog.rs`
3. **Frontend integration** — web UI cost rates editor fetches pricing on-demand and pre-fills rate fields

The PR is well-scoped: additive provider registration, no behavioral change to existing providers, clear validation evidence.

---

## 🟢 What looks good

### Correct extension pattern

The Kilo provider follows the established `CompatFamilySpec` pattern exactly. The impl in `factory.rs` is minimal:

```rust
impl CompatFamilySpec for KiloModelProviderConfig {
    const DISPLAY: &'static str = "Kilo";
    const DEFAULT_URL: &'static str = "https://app.kilo.ai/api/gateway";
    const AUTH: AuthStyle = AuthStyle::Bearer;
    const UNAUTHENTICATED_MODEL_LISTING: bool = true;
}
```

This is the right abstraction use — the blanket impl in `CompatFamilySpec` produces `FamilyProviderFactory` for free, and the `UNAUTHENTICATED_MODEL_LISTING` const enables catalog fetch without a credential. No duplicate logic, no new factory arms.

### Pricing as optional data, not a contract change

The `ModelPricing` and `ModelInfo` structs in `zeroclaw-api/src/model_provider.rs` are correctly optional — `pricing: Option<ModelPricing>` throughout. The `list_models_with_pricing()` trait method has a sensible default fallback that delegates to `list_models()` and returns `None` for pricing, so existing providers that don't override it continue to work unchanged. This keeps the change additive and backward-compatible.

### Catalog fallback hierarchy is preserved

The `OpenAiCompatibleModelProvider::list_models_with_pricing()` impl maintains the existing three-tier fallback (live endpoint → models.dev → OpenRouter) and threads pricing through when available. The `normalize_models_with_pricing()` helper is a clean parallel to the existing `normalize_model_ids()` — same shape, pricing-aware.

### Race condition fix is real

The commit message for `d316907` ("fix: fetch pricing on-demand in cost rate add") identifies a legitimate race: the web UI was fetching catalog pricing on component mount, and by the time the user submitted a new rate, that data could be stale or missing. The fix moves the fetch to submit-time (`addRates` / `applyPricing`), which is the right place for it — the data is fresh and the user sees the result immediately.

### Validation is complete

The PR body shows clean output from all three gates (fmt, clippy, test). The test suite passes 388 tests with 7 live-credential tests ignored, which matches the current baseline. The author also verified Kilo provider appears in the config API and pricing populates in the cost rates editor — that's exactly the beyond-CI evidence the template asks for.

### Labeler update is present

`.github/labeler.yml` gets a new `provider:kilo` entry that covers the right files (schema, factory, catalog). This is often forgotten; the author caught it.

---

## 🔴 Blocking — Missing documentation entry in catalog.md

The commit message for `055a4b7ad` claims "Documentation in providers/catalog.md", and the diff shows 13 additions / 1 deletion in `docs/book/src/providers/catalog.md`. However, **the catalog.md entry is missing the Kilo slot documentation** that should appear in the "OpenAI-compatible families" table.

Inspecting the live file on the PR branch, the table at line ~144 lists every compat family slot (groq, mistral, xai, deepseek, cohere, … lmstudio, llamacpp, sglang, vllm, osaurus, litellm) and explicitly documents which providers have `models.dev` entries, OpenRouter fallback prefixes, or public unauthenticated `/models` endpoints. Kilo should appear in this table with a note that its `/models` endpoint is public (`UNAUTHENTICATED_MODEL_LISTING: bool = true`).

The worked example immediately below the table (`[providers.models.groq.fast]`) should either be supplemented with a Kilo example or the existing example should note in prose that Kilo works identically:

```toml
[providers.models.kilo.home]
model   = "anthropic/claude-sonnet-4-6"
api_key = "..."
```

**Why this blocks:** The catalog.md document is the canonical reference for operators configuring providers. A missing entry means users don't know the slot exists, which defeats the purpose of shipping the provider. This is not a "docs debt" item — it's a merge-blocking omission per FND-002 §3.1, which mandates that every extension surface (providers, channels, tools) ships with reference documentation in the same PR.

**Fix:** Add a row to the OpenAI-compatible families table:

```markdown
| `kilo` | `https://app.kilo.ai/api/gateway` | Public `/models` endpoint (no credential required for catalog) |
```

Add a worked example under the table:

```toml
[providers.models.kilo.home]
model   = "anthropic/claude-sonnet-4-6"
api_key = "..."
```

Or append a note to the existing Groq example:

> Kilo Gateway works identically — use `[providers.models.kilo.<alias>]` and reference it via `model_provider = "kilo.<alias>"`.

---

## 🟡 Warning — Const naming inconsistency in CompatFamilySpec

The new `UNAUTHENTICATED_MODEL_LISTING` const on `CompatFamilySpec` breaks the naming convention established by the existing catalog consts. Compare:

```rust
const MODELS_DEV_KEY: Option<&'static str> = None;
const OPENROUTER_VENDOR_PREFIX: Option<&'static str> = None;
const UNAUTHENTICATED_MODEL_LISTING: bool = false;
```

The first two are nouns (a key, a prefix). The third is an adjective phrase that reads like a capability flag. This inconsistency will make the trait harder to scan as more catalog features are added.

**Suggested fix:** Rename to `PUBLIC_MODEL_LISTING` or `CATALOG_REQUIRES_AUTH` (inverted). The former is shorter and matches the domain language ("public catalog", "public endpoint"). If you choose to invert, flip the bool to `const CATALOG_REQUIRES_AUTH: bool = true;` and update the one call site in `build_compat_base()`.

**Why this is a warning, not a blocker:** The code works, the feature is correct, and the name is unambiguous. But the trait is a public extension point — every new compat provider will see this surface, and inconsistent naming makes the contract harder to learn. Fixing it now avoids a breaking change later when someone proposes harmonizing the names.

---

## 🔵 Suggestion — Document the pricing units in ModelPricing docstring

The `ModelPricing` struct in `zeroclaw-api/src/model_provider.rs` has fields `prompt`, `completion`, `input_cache_read`, `input_cache_write`, all typed as `Option<String>`. The struct-level docstring says "All values are per-token rates as strings. e.g. '0.000005' = $5.00 per 1M tokens."

This is good, but the **individual field docstrings** (lines 374–385) just say "Input/prompt tokens per-token rate" without repeating the unit or the string encoding. A maintainer reading the field in isolation won't know whether `"0.000005"` is dollars per token, dollars per million tokens, or cents per token without scrolling up to the struct-level comment.

**Suggested improvement:** Repeat the unit in each field's docstring:

```rust
/// Input/prompt tokens per-token rate (USD per token, e.g. "0.000005" = $5/1M tokens).
#[serde(default, skip_serializing_if = "Option::is_none")]
pub prompt: Option<String>,
```

This follows Rust API docs best practice: every public field should be self-documenting. It's a small change, but it saves the next contributor (or the LLM assistant parsing this code) from having to infer context.

---

## 🔵 Suggestion — Extract pricing normalization into a named function

The `list_models_with_pricing()` impl in `compatible.rs` (lines 2218–2305) has two nearly-identical blocks that parse a `ModelsResponse` and normalize it into `Vec<ModelInfo>`:

1. Lines 2269–2270: `normalize_models_with_pricing(body)` after a successful authenticated fetch
2. Lines 2280–2284: an inline `.map(|id| ModelInfo { id, pricing: None })` for the models.dev fallback path

The OpenRouter fallback (line 2302) calls `list_models_for_vendor_with_pricing()`, which already returns `Vec<ModelInfo>` with pricing.

**Observation:** The models.dev path is the only one that doesn't preserve pricing, because models.dev doesn't serve pricing data. But the inline map obscures this intentional data loss — a future reader might wonder why we're not calling `normalize_models_with_pricing()` there too.

**Suggested refactor:** Extract a `models_dev_to_model_info(ids: Vec<String>) -> Vec<ModelInfo>` helper that documents the no-pricing contract:

```rust
/// Convert models.dev IDs (no pricing) into ModelInfo entries.
fn models_dev_to_model_info(ids: Vec<String>) -> Vec<ModelInfo> {
    ids.into_iter()
        .map(|id| ModelInfo { id, pricing: None })
        .collect()
}
```

Then the call site becomes:

```rust
Ok(models_dev_to_model_info(models))
```

This makes the "models.dev doesn't have pricing" fact explicit in the code structure, not just in the maintainer's mental model.

---

## 🔵 Suggestion — Consider a test for the UNAUTHENTICATED_MODEL_LISTING path

The `OpenAiCompatibleModelProvider::list_models_with_pricing()` impl has a new branch at line 2226:

```rust
if list_credential.is_some() || self.unauthenticated_model_listing {
    // hit the live /models endpoint
}
```

The `|| self.unauthenticated_model_listing` part is new. The PR body says "Backend verified: 338 Kilo models with pricing via live catalog API", which suggests manual verification, but there's no unit test that exercises the `unauthenticated_model_listing = true` path with a mock endpoint.

**Why this matters:** The existing `list_models()` tests in `compatible.rs` (if any) likely cover the credential path. The new path is a capability flag that changes fetch behavior — if someone later refactors the method and accidentally breaks the flag logic, we won't know until a user reports "Kilo catalog doesn't work anymore".

**Suggested test shape:**

```rust
#[tokio::test]
async fn list_models_with_pricing_unauthenticated() {
    let mock_server = /* ... set up a mock /models endpoint that returns pricing ... */;
    let provider = OpenAiCompatibleModelProvider::new(
        "kilo-test",
        "Kilo",
        &mock_server.url,
        None, // no credential
        AuthStyle::Bearer,
    )
    .with_unauthenticated_model_listing();
    
    let models = provider.list_models_with_pricing().await.unwrap();
    assert!(!models.is_empty());
    assert!(models[0].pricing.is_some());
}
```

Not blocking — the feature works, and the PR validation confirms it. But adding this test now protects the investment in this PR against future refactors.

---

## 🔵 Suggestion — Web UI: consolidate catalog fetch calls

The `CostRatesEditor.tsx` changes add on-demand catalog fetching in two places:

1. `SingleResourceEditor.addRates()` (line ~230)
2. `ResourceListEditor.applyPricing()` (line ~480)

Both have the same pattern:

```typescript
if (!catalogPricing) {
  const fresh = await fetchCatalogModels(providerType);
  catalogPricing = fresh.pricing;
}
```

This is correct, but the duplication means if we change the fetch logic (add caching, add error retry, add a loading spinner), we have to update both call sites.

**Suggested refactor:** Extract a `useCatalogPricing(providerType)` hook or a `ensureCatalogPricing(providerType, current)` helper function that encapsulates the "fetch if missing" logic. Then both components call the same helper, and future changes (like adding a `useSWR` cache) touch one place.

Not urgent — the duplication is only two call sites, and the logic is trivial. But if catalog fetching grows (e.g., we add a TTL or a background refresh), this will bite us.

---

## Compatibility & rollback

- **Backward compatible:** Yes. New `[providers.models.kilo.*]` config section is optional; existing configs unchanged.
- **Breaking changes:** None.
- **Rollback:** `git revert` works cleanly. The PR is purely additive — reverting it leaves no orphaned config or state.
- **Observable failure symptoms:** If Kilo endpoint is unreachable, provider init logs an error and the provider is excluded from the available list (same pattern as all other providers). No cross-provider impact.

---

## Template compliance

- ✅ **Summary:** Complete — what/why/scope/blast radius all present
- ✅ **Validation evidence:** All three commands shown with output
- ✅ **Security & privacy:** Correctly answered — new network call to Kilo endpoint, same credential pattern as existing providers
- ✅ **Compatibility:** Correctly answered — backward compatible, optional config section
- ✅ **Rollback:** Complete for a low-risk PR
- ✅ **Labels:** Present (`ci`, `docs`, `core`, `config`, `gateway`, `provider`, `runtime`, `provider:compatible`)
- ❌ **Supersede attribution:** N/A (not a superseding PR)

---

## Risk assessment

**Risk tier:** Low (per `AGENTS.md` risk tiers)

- No changes to `crates/zeroclaw-runtime/src/security/`
- No changes to access-control boundaries
- No changes to `.github/workflows/**`
- No changes to `crates/zeroclaw-gateway/src/**` beyond additive config API surface
- No changes to `crates/zeroclaw-tools/src/**`

This is a provider registration + catalog enhancement. The blast radius is intentionally narrow: `zeroclaw-providers` (factory dispatch), `zeroclaw-config` (schema additions), `zeroclaw-gateway` (config API threading), web UI (cost rates editor). Existing providers are unaffected.

---

## Verdict: Request changes

One blocking issue (missing catalog.md entry) and several improvement opportunities that would strengthen the PR but aren't merge-blocking. The code is correct, the validation is thorough, and the architecture is sound. Once the catalog.md entry is added, this PR is ready to land.

**Next steps for @Nillth:**

1. Add Kilo to the OpenAI-compatible families table in `docs/book/src/providers/catalog.md` with a note about public model listing
2. Add a worked example or append a note to the existing Groq example
3. (Optional but recommended) Address the 🟡 warning about `UNAUTHENTICATED_MODEL_LISTING` naming — either rename to `PUBLIC_MODEL_LISTING` or document why the current name is preferred
4. (Optional) Consider the 🔵 suggestions if time permits — they're polish, not blockers

Once the catalog.md entry is added, ping me or re-request review. I'm happy to fast-track the re-review if the fix is straightforward.

---

**Review principles applied:**

- FND-005 §3: "Be specific. Vague feedback creates anxiety without direction."  
  → Every finding explains the principle, not just the symptom.
- FND-005 §3: "Name what is good."  
  → Six specific 🟢 findings that teach what to repeat.
- FND-005 §5: Feedback taxonomy — 🔴 for merge-blocking, 🟡 for should-address, 🔵 for optional.
- FND-006 §2.1: "Tests as design feedback."  
  → Suggested test for the new unauthenticated-listing path.
- AGENTS.md: "Map non-trivial changes — read the architecture docs first."  
  → Cross-checked against FND-001 (microkernel architecture), FND-002 (documentation standards), FND-006 (quality discipline).

---

Thank you for the thorough PR and clean commits, @Nillth. This is good work. The missing docs entry is an easy fix, and the suggestions are offered in the spirit of helping this feature be maintainable long-term. Let me know when you've pushed the catalog.md update and I'll re-review.

WareWolf-MoonWall (approved):
# Review of PR #7136 — feat(providers): add Kilo AI Gateway as first-class model provider with pricing capture

Reviewed by @WareWolf-MoonWall on 2026-06-04  
Head commit: `5da6ea909` (chore: cargo fmt on 7 modified files)

---

## Summary

This PR adds Kilo AI Gateway as a first-class OpenAI-compatible model provider with public model listing support, extends the `ModelProvider` trait to capture and expose pricing data from provider `/models` endpoints, and fixes a race condition in the web UI cost rates editor where pricing was fetched on component mount (leading to stale data at submit time). The implementation is clean, follows existing patterns, and respects the single-source-of-truth rule. The risk label mismatch (auto-labeled `risk: high`, should be `risk: low`) needs correction before merge.

**Verdict:** Approve with label fix and one non-blocking clarification request.

---

### 🟢 What looks good — Single source of truth preserved

The pricing data flow is clean and correct:

1. **Source of truth:** Provider `/models` endpoint returns pricing JSON.
2. **Deserialization surface:** `ModelPricing` struct in `zeroclaw-api::model_provider` captures the raw per-token rates.
3. **Transport:** `list_models_with_pricing()` trait method returns `Vec<ModelInfo>` with optional pricing.
4. **Web UI consumption:** Pricing fetched on-demand at submit time (not cached in component state between mount and submit).

No duplicate state introduced. The `CACHED_CATALOG_WITH_PRICING` static in `openrouter_catalog.rs` is a materialized view fetched on-demand (acceptable per AGENTS.md §"Patterns that are NOT duplicate state"). The `PUBLIC_MODEL_LISTING` const on the `CompatFamilySpec` trait is a compile-time signal, not stored state. Both correct.

The race condition fix (fetch pricing on-demand at submit, not on mount) eliminates the stale-state bug where the user could add a rate after the catalog was fetched but before they submitted. This was a real bug and this is the right fix.

---

### 🟢 What looks good — Trait extension is backward-compatible

The `list_models_with_pricing()` method added to `ModelProvider` has a default implementation that delegates to `list_models()` and returns no pricing:

```rust
async fn list_models_with_pricing(&self) -> anyhow::Result<Vec<ModelInfo>> {
    Ok(self
        .list_models()
        .await?
        .into_iter()
        .map(|id| ModelInfo { id, pricing: None })
        .collect())
}
```

This means every existing provider continues to work unchanged. Only providers that override this method (OpenAI-compatible providers with pricing in their `/models` response, OpenRouter) surface pricing. Anthropic, Gemini, Bedrock, etc., are unaffected. Correct backward-compatible extension pattern.

---

### 🟢 What looks good — Test coverage for the new flag

The diff adds two unit tests for the `public_model_listing` flag:

1. `public_model_listing_flag_defaults_false` — verifies that providers without an explicit `PUBLIC_MODEL_LISTING = true` default to `false`, preserving existing behavior.
2. `public_model_listing_flag_can_be_set` — verifies the builder method correctly enables the flag.

These tests document the default-false contract and prevent regressions. Good test discipline.

---

### 🟢 What looks good — Rename clarifies intent

The rename from `unauthenticated_model_listing` → `public_model_listing` is clearer. "Unauthenticated" implies the absence of auth; "public" states the availability contract (catalog is accessible without a credential). The new name is more accurate for providers like Kilo Gateway whose model catalog is intentionally public even though their inference API requires authentication.

All call sites updated correctly (method definition in `compatible.rs`, Ollama usage in `factory.rs`, trait helper in `factory.rs`, tests). No orphaned call sites remain in master.

---

### 🟡 Warning — Risk label mismatch

The auto-labeler applied `risk: high` due to changes in `zeroclaw-runtime/` and `zeroclaw-gateway/`. However, the actual changes in those crates are:

- **`zeroclaw-runtime/src/rpc/`**: Added `pricing` field to `CatalogModelsResult` RPC type (additive, no behavior change).
- **`zeroclaw-gateway/src/api_sections.rs`**: Updated `handle_catalog_models` to destructure the new `pricing` field from `quickstart::model_catalog()` and pass it through (additive, no behavior change).

Neither change touches security policy, tool execution, access control, or agent loop logic. The runtime and gateway changes are pure plumbing for the new pricing data. This is **low-risk** by the project's risk tier definitions (AGENTS.md §"Risk Tiers").

**Action required:** Update the label to `risk: low` before merge. If you have label permissions, fix it yourself per the review protocol. If not, ask @Nillth to update it or request a maintainer to correct it.

---

### 🟡 Warning — Silent-fail pricing fetch lacks observability

The web UI pricing fetch is wrapped in `.catch(() => {})` (silent fail) because pricing pre-fill is a nice-to-have, not a blocker. This is reasonable UX behavior — the form still works without pricing. However, there's no observability hook to know when pricing fetch fails in production. If Kilo Gateway's `/models` endpoint is unreachable or returns malformed JSON, operators won't see any indication in logs or dashboards.

**Recommendation:** Consider adding a debug-level log emission in `getCatalogModels()` (web-side) or the gateway endpoint when pricing fetch fails. Not blocking — the silent-fail UX is correct — but low-visibility failures make it harder to debug "why didn't pricing auto-populate?" support tickets later.

**If you defer this:** Track it as a follow-up issue under `type: observability`. Not blocking merge.

---

### 🔵 Suggestion — Document why models.dev doesn't serve pricing

The `models_dev_to_model_info` helper includes this comment:

```rust
/// The models.dev catalog does not serve pricing data, so this function
/// documents the intentional data-loss contract: callers should expect
/// `pricing: None` on every returned entry.
```

This is good documentation. Consider adding *why* models.dev doesn't serve pricing — external limitation (models.dev catalog schema doesn't include pricing fields), not our choice. This distinction matters for future readers who might wonder "should we patch models.dev to include pricing?" The answer is "models.dev is an external catalog; we consume what they publish."

**Suggested addition:**

```rust
/// The models.dev catalog schema does not include pricing fields (external
/// limitation). This function documents the intentional data-loss contract:
/// callers should expect `pricing: None` on every returned entry.
```

Not blocking — the existing comment is sufficient. This just adds context for future maintainers.

---

### 🔵 Suggestion — Consider validating Kilo endpoint URL format

The `KiloModelProviderConfig` schema defines a single endpoint variant with a hardcoded URL:

```rust
impl ModelEndpoint for KiloEndpoint {
    fn uri(&self) -> &'static str {
        match self {
            Self::Gateway => "https://api.kilo.ai/api/gateway",
        }
    }
}
```

However, the PR description states the endpoint is "configurable" (user can override via `api_url`). If the operator overrides the URL to a non-Kilo endpoint (e.g., a local proxy or a different OpenAI-compatible gateway), the provider still identifies as "Kilo" in logs, attribution, and UI labels.

**Question:** Is there a risk of operator confusion if they configure a Kilo alias with a non-Kilo `api_url` override? Should the `FamilyEndpoint` impl for Kilo validate the URL format (e.g., domain must be `*.kilo.ai`) or log a warning when overridden?

**If the answer is "operators can override to any OpenAI-compatible endpoint and we trust them":** No change needed. Just document the expectation in `docs/book/src/providers/catalog.md` (already updated in this PR, looks good).

Not blocking — the current design is consistent with other OpenAI-compatible families (OpenRouter, Together, etc., all allow `api_url` overrides without validation).

---

### 🔵 Suggestion — Verify Kilo Gateway pricing schema stability

The `ModelPricing` struct includes four optional fields:

```rust
pub struct ModelPricing {
    pub prompt: Option<String>,
    pub completion: Option<String>,
    pub input_cache_read: Option<String>,
    pub input_cache_write: Option<String>,
}
```

The comments state that `input_cache_read` and `input_cache_write` are "Kilo Gateway specific." This implies Kilo Gateway's `/models` endpoint serves these fields in its pricing object, but OpenRouter does not.

**Verification request:** Has this been tested against Kilo Gateway's live `/models` endpoint? If Kilo Gateway later changes its schema (e.g., renames these fields or introduces new ones), the deserialization will silently drop the new fields (Serde `#[serde(default)]` on the struct would prevent errors but also hide schema drift).

**Recommendation:** Add a note in `docs/book/src/providers/catalog.md` documenting which providers serve which pricing fields, so future maintainers know which schemas to monitor for drift. Example:

> **Pricing schema by provider:**
> - OpenAI: `prompt`, `completion`
> - OpenRouter: `prompt`, `completion`
> - Kilo Gateway: `prompt`, `completion`, `input_cache_read`, `input_cache_write`

Not blocking — the struct is forward-compatible (new fields will be ignored until we add them). This is just about making schema drift visible sooner.

---

### 🔵 Suggestion — Size label mismatch

The auto-labeler applied `size: L` (501-1000 changed lines). The PR description claims `size: M`. The actual diff is 499 additions + 49 deletions = 548 total changed lines. By the labeler's own rules (501-1000 = L), `size: L` is correct.

Not blocking — size labels are metadata, not a quality gate. Just noting the mismatch in case the author wants to reconcile the description with the auto-label.

---

## Validation

Per the PR description, the following commands were run and passed:

```bash
cargo fmt --all -- --check          # passed (clean)
cargo clippy --all-targets -- -D warnings  # passed, 2m 16s
cargo test                          # passed: 388 passed; 0 failed; 7 ignored
```

The test suite includes:
- Unit tests across all workspace crates: ok
- Integration tests (159): ok
- Architecture tests (79): ok
- System tests (9): ok
- Live tests (7): ignored (require live credentials)

**Manual verification (per PR description):** "Kilo provider appears in `/config/providers.models` list; pricing populates in cost rates editor for Kilo models; existing providers (OpenAI, OpenRouter, Anthropic) unaffected."

**Recommendation for future PRs:** When claiming "pricing populates," show the actual HTTP request/response (e.g., `curl` the gateway `/api/config/catalog?model_provider=kilo` or a screenshot of the web UI cost rates editor with pricing pre-filled). This helps reviewers verify the end-to-end flow without needing to spin up a local Kilo account. Not blocking this PR — the test coverage is sufficient — just a note for next time.

---

## Compatibility & Rollback

**Backward compatible?** Yes. New `[providers.models.kilo.*]` config section is optional. Existing configs work unchanged.

**Config surface changed?** Yes. New `[providers.models.kilo.<alias>]` section. No migration needed — users who don't add this section see no change.

**Rollback:** `git revert` works. No data migrations, no stateful changes. Removing the `[providers.models.kilo]` section disables the provider without code changes. Correct.

---

## Foundations Compliance

**FND-001 (Architecture):** No microkernel boundary violations. Provider additions live in `zeroclaw-providers` (foundation layer). Runtime and gateway changes are pure plumbing (no new dependencies, no boundary crossings). ✅

**FND-005 (Contribution Culture):** PR description follows the template fully. Validation evidence provided. Security/privacy impact assessed correctly (new network calls scoped to configured endpoint only; no new attack surface). ✅

**FND-006 (Zero Compromise in Practice):** Proper error propagation in pricing fetch (silent fail in UI is intentional nice-to-have UX, not error suppression). Test coverage for new flag behavior. No `unwrap()`/`expect()` introduced in production paths. ✅

**AGENTS.md (Single Source of Truth):** No duplicate state introduced. Pricing flows from provider endpoint → trait method → web UI on-demand. `PUBLIC_MODEL_LISTING` is a trait const, not stored state. Materialized views (`CACHED_CATALOG_WITH_PRICING`) are fetched on-demand. ✅

---

## Final Notes

This is a well-executed additive feature. The pricing capture pattern is clean, the race condition fix is correct, and the trait extension is backward-compatible. The only blocking issue is the label mismatch (needs `risk: low`). Everything else is non-blocking clarification or follow-up suggestions.

**Next steps:**
1. Fix `risk: high` → `risk: low` label (you or maintainer).
2. Address the 🟡 warnings and 🔵 suggestions if you agree, or respond with reasoning if you prefer to defer.
3. Once the label is corrected, this PR is ready to merge.

Good work, @Nillth. The implementation quality is high and the validation evidence is thorough. 

---

**Reviewed commits:**
- `055a4b7ad` feat: add Kilo AI Gateway as first-class model provider
- `2e4fa297f` feat: capture and display pricing from model catalogs
- `6057a2a83` fix: fetch pricing on-demand in cost rate add (avoid stale state race)
- `5da6ea909` chore: cargo fmt on 7 modified files

Nillth:
Apparently when you tell your AI to rebase your branches on the current master, you need to remember to also tell it to leave the ones with active PRs alone...

| Check | Result |
|---|---|
| cargo fmt --all -- --check | ✅ Clean |
| cargo clippy --all-targets -- -D warnings | ✅ 0 warnings |
| cargo test --locked | ✅ 894 passed, 0 failed, 7 ignored |

Audacity88 (changes requested):
Context checked: #7136 head 202f989cace471b07f6fdb2d223f54fc5e1a23cd, the public diff, PR body, green GitHub checks, WareWolf's earlier reviews, and the Kilo provider/config/docs paths in `schema.rs`, `factory.rs`, `schema/v2.rs`, `lib.rs`, and `providers/catalog.md`. I did not run local cargo checks for this pass.

In simple terms, this PR adds Kilo as a first-class OpenAI-compatible provider, carries model pricing from provider catalogs through the backend API, and lets the cost-rate editor fill pricing from catalog data when the submit path needs it.

Requesting changes on the endpoint source-of-truth mismatch below. The overall direction looks good, but this should be aligned before merge.

I also checked the earlier WareWolf feedback. The current head appears to have addressed the missing provider docs and the public-listing naming issue.

### 🟢 What looks good — Pricing stays optional and additive

I like the way the pricing data is modeled as optional catalog metadata instead of a new required provider contract. Existing providers can keep returning plain model IDs, while Kilo/OpenRouter-style providers can surface pricing when the endpoint provides it. That keeps the provider trait extension low-friction and avoids forcing unrelated providers into fake pricing data.

### 🟢 What looks good — The cost-rate editor submit path is more robust

Fetching catalog pricing from the submit path when cached pricing is absent is the right shape here. It gives the editor another chance to fill rates from the current provider/model choice instead of relying only on whatever was available at component mount.

### 🔴 Blocking — Kilo has two different default gateway URLs

The current branch wires two different Kilo Gateway defaults:

- `KiloEndpoint::Gateway` in `crates/zeroclaw-config/src/schema.rs` resolves to `https://api.kilo.ai/api/gateway`.
- `KiloModelProviderConfig::DEFAULT_URL` in `crates/zeroclaw-providers/src/factory.rs`, the provider docs, and the PR body all use `https://app.kilo.ai/api/gateway`.

This is not just a docs typo. The diff introduces both a typed config endpoint enum and a factory default for the same Kilo Gateway provider, and those two sources disagree. That means the schema/default endpoint surface and the provider factory/docs can point users or runtime construction through different hosts for the same “gateway” default.

That split gives Kilo two sources of truth for the same “gateway” endpoint and can make different entry points use different hosts. Please pick the canonical URL, make the schema endpoint and factory/docs agree, and add a focused regression that exercises the alias-resolved Kilo path so the default cannot drift again. If `api.kilo.ai` is intentionally the runtime endpoint, the docs and `DEFAULT_URL` should say that. If `app.kilo.ai` is correct, `KiloEndpoint::Gateway` should return that.

### 🟡 Warning — The old `kilo` synonym changes meaning

The compatibility section says this is backward-compatible, but the diff changes the old `kilo` compatibility name. On `master`, `normalize_provider_type()` maps both `kilocli` and `kilo` to `kilocli`, and `canonicalize_v2_model_provider_name()` also maps `kilo` to `kilocli`. This branch changes that so `kilo-cli` remains the KiloCLI synonym and `kilo` becomes the new gateway provider name.

That may be the right long-term naming, but it is not purely additive for anyone relying on the old `kilo` synonym for KiloCLI. Please either preserve the old compatibility path somehow, or document the compatibility break/migration clearly in the PR body and provider docs so existing users know to use `kilocli` / `kilo-cli` for the subprocess provider.

Once the Kilo endpoint source of truth is aligned, I think this should be straightforward to re-review.

Nillth:
@Audacity88 thanks for catching the split — both findings addressed.

**🔴 Endpoint source of truth — aligned to `api.kilo.ai`.** Checked the official docs (https://kilo.ai/docs/gateway/quickstart): `https://api.kilo.ai/api/gateway` is the documented API base URL across all SDK examples; `app.kilo.ai` is the web-app sign-in. So the schema's `KiloEndpoint::Gateway` was already canonical, and the factory `DEFAULT_URL`, provider docs, and PR body were the stragglers — all now corrected to `api.kilo.ai`. I also verified both hosts currently serve identical `/models` payloads (same backend), so nothing breaks for anyone who hit the other host. The requested regression test (`kilo_gateway_default_url_matches_schema_endpoint`) locks the schema endpoint and factory default together so they can't drift again.

**🟡 `kilo` synonym — owning the rename, documented.** `kilo` now means the gateway provider; the KiloCLI subprocess provider keeps `kilocli` (synonym `kilo-cli`). Added a migration note to the provider catalog docs and updated the PR body's compatibility section to state the break explicitly instead of claiming pure additivity.

Also merged current `master` into the branch (was ~49 behind).

Validation (local):
```text
cargo fmt --all -- --check   → clean
cargo clippy --all-targets -- -D warnings → Finished `dev` profile in 2m 03s (no findings)
cargo test                   → 894 passed; 0 failed; 7 ignored across the workspace, e.g.:
  test result: ok. 243 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.12s
  test result: ok. 273 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.14s
  test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.10s
  test result: ok. 159 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.25s
cargo test -p zeroclaw-providers kilo → 11 passed; 0 failed (includes the new regression test)
```

Audacity88 (changes requested):
@Nillth, I re-reviewed #7136 at head 8d370b0686f80ef41879a613bfc73ec655f25aa1. I checked the live PR body, top-level conversation, prior reviews, green GitHub checks, the current public diff, and the Kilo/OpenRouter/provider-catalog source paths in `schema.rs`, `providers.rs`, `factory.rs`, `openrouter.rs`, `compatible.rs`, `openrouter_catalog.rs`, `quickstart/mod.rs`, `api_sections.rs`, `CostRatesEditor.tsx`, and `docs/book/src/providers/catalog.md`. I did not run local cargo checks for this pass.

In simple terms, this PR adds Kilo AI Gateway as a first-class provider, carries optional pricing metadata through the model-catalog API, and lets the cost-rate editor prefill model rates when catalog pricing is available.

The previous endpoint source-of-truth blocker is resolved, and the `kilo` naming migration is now documented. I am still requesting changes on one pricing-path mismatch below.

### ✅ Resolved — Kilo endpoint source of truth is aligned

The earlier split between `api.kilo.ai` and `app.kilo.ai` is fixed. `KiloEndpoint::Gateway` and `KiloModelProviderConfig::DEFAULT_URL` now both resolve to `https://api.kilo.ai/api/gateway`, and `kilo_gateway_default_url_matches_schema_endpoint` locks those two surfaces together. The provider docs and PR body also now use the same canonical API host.

### ✅ Resolved — The `kilo` naming migration is explicit

The old `kilo` shorthand behavior is now called out instead of being framed as purely additive compatibility. The catalog page says `kilo` now means Kilo AI Gateway, while the subprocess provider keeps `kilocli` and `kilo-cli`. That is enough for operators who previously used `kilo` for KiloCLI to know the upgrade step.

### 🟢 What looks good — Kilo pricing stays optional and additive

The Kilo path uses the existing OpenAI-compatible provider shape rather than adding a bespoke provider implementation. Pricing remains optional in `ModelInfo`, so providers without pricing can keep returning plain model IDs. That keeps the new trait method backward-compatible and avoids forcing unrelated providers to invent placeholder cost data.

### 🔴 Blocking — The first-class OpenRouter provider still drops pricing

The PR body says Kilo and OpenRouter fetch and display pricing automatically in the web UI cost-rates editor, but the first-class `openrouter` slot still loses pricing before the API response reaches the UI.

The runtime path is:

1. `model_catalog("openrouter")` creates the first-class `OpenRouterModelProvider`.
2. It calls `handle.list_models_with_pricing()`.
3. `OpenRouterModelProvider` only overrides `list_models()`, not `list_models_with_pricing()`.
4. The new default trait implementation delegates to `list_models()` and wraps every ID as `ModelInfo { pricing: None }`.

So `/api/config/catalog/models?provider=openrouter` can return model IDs, but it will not return a `pricing` map for the cost-rate editor to apply. The new `openrouter_catalog::list_models_for_vendor_with_pricing()` helper is useful for OpenAI-compatible vendor fallback paths that carry an `OPENROUTER_VENDOR_PREFIX`, but it does not cover the `openrouter` provider family itself.

That leaves the public OpenRouter pricing claim unsupported and gives users no automatic OpenRouter rate prefill on the actual `openrouter` slot. Please either add an `OpenRouterModelProvider::list_models_with_pricing()` override that parses the public `/models` pricing data into `ModelInfo` and cover it with a focused regression, or narrow the PR body/docs so they do not claim first-class OpenRouter pricing support.

### 🟡 Warning — Kilo docs overstate models.dev as the catalog source

The new catalog section says Kilo's catalog is "sourced from models.dev under the `kilo` key." The broader catalog fallback does include a `kilo` models.dev entry, but the pricing-preserving Kilo provider path uses `PUBLIC_MODEL_LISTING = true` and queries Kilo's live `/models` endpoint without a credential. Since the live endpoint is the source that preserves pricing, the docs should say that directly instead of making models.dev sound like the primary catalog source.

### 🟡 Warning — The branch is currently conflicted with master

Separately from the code finding above, GitHub reports this PR as `DIRTY` / `CONFLICTING` against current `master`. A merge-tree check shows conflicts in `crates/zeroclaw-config/src/providers.rs`, `crates/zeroclaw-providers/src/factory.rs`, and `docs/book/src/providers/catalog.md`, which makes sense given the recent provider-stream churn. After the pricing-path fix, please refresh the branch against current `master` and rerun the required checks on the refreshed head before this goes back to merge-readiness.

Once the OpenRouter pricing claim and the Kilo catalog-source wording are aligned, this should be a straightforward re-review.

Audacity88 (approved):
@Nillth, I re-reviewed #7136 at head ba882aa4ae9ff2ea630e166764fbf45320f397e0. I checked the live PR body, top-level conversation, prior formal reviews, inline comments, current GitHub checks, mergeability, and the current public diff across the Kilo/OpenRouter/provider-catalog paths. I did not run local cargo checks for this pass.

In simple terms, this PR adds Kilo AI Gateway as a first-class provider, carries optional pricing metadata through the model-catalog API, and lets the cost-rate editor prefill model rates when catalog pricing is available.

The prior blockers I held are resolved on the current head, and all visible checks are green. The PR can still show a blocked review state before this approval is submitted; this approval is intended to clear my own earlier changes-requested review, not override another reviewer's active block.

### ✅ Resolved — First-class OpenRouter pricing now reaches the catalog API

The previous issue was that `model_catalog("openrouter")` constructed the first-class `OpenRouterModelProvider`, whose default `list_models_with_pricing()` path wrapped plain model IDs as `pricing: None`. That meant `/api/config/catalog/models?provider=openrouter` could list models but could not prefill rates for the actual `openrouter` slot.

The current head fixes that directly. `OpenRouterModelProvider` now overrides `list_models_with_pricing()` and calls `openrouter_catalog::list_all_models_with_pricing()`. The catalog helper preserves full `<vendor>/<slug>` IDs, carries the optional pricing object through to `ModelInfo`, sorts/dedupes by ID, and has focused unit coverage for full-ID preservation, pricing preservation, missing-pricing entries, and duplicate IDs. That aligns the implementation with the PR body's OpenRouter pricing claim.

### ✅ Resolved — Kilo catalog-source wording now matches the pricing path

The catalog docs now say the Kilo live public `/models` endpoint is the pricing-bearing source and that the `models.dev` `kilo` key is only a no-pricing fallback when the live endpoint is unreachable. That matches the implementation: `KiloModelProviderConfig` sets `PUBLIC_MODEL_LISTING = true`, so the OpenAI-compatible provider path can query the live endpoint without a credential and preserve pricing when it is present.

### ✅ Resolved — Branch freshness and conflicts are cleared

The previous re-review saw GitHub reporting the branch as dirty/conflicting against `master`. The current head includes the upstream merge, and all visible required checks on ba882aa4ae9ff2ea630e166764fbf45320f397e0 are green.

### 🟢 What looks good — The provider/catalog flow is coherent now

The shape is nicely additive: pricing remains optional in `ModelInfo`, default provider behavior still degrades to plain model IDs, Kilo uses the existing OpenAI-compatible family pattern, and OpenRouter gets the special full-catalog behavior it needs as a first-class provider. The gateway/RPC response simply threads the pricing map through when available, while the web cost-rate editor treats pricing prefill as best-effort rather than a hard dependency.

Approving from my side. Nice recovery on the review loop; the latest commits close the pricing-path gap without broadening the provider abstraction more than necessary.

PR #7123, opened 2026-06-03
fix(zerocode/channels/tools): avoid UTF-8 char-boundary panics in text truncation

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:** `dashboard::truncate`, the Bluesky post truncation, and the LinkedIn image-prompt truncation all sliced `&s[..n]` at a raw **byte** index. When `n` landed inside a multi-byte char they panicked (e.g. a CJK session summary panicked the zerocode dashboard: `end byte index 40 is not a char boundary; it is inside '列'`). All three now use `chars().count()` / `chars().take(n)`, which is char-boundary safe and matches the intended **character**-based semantics.
- A repo-wide sweep for the same unguarded byte-slice pattern found exactly these three sites; all other slice sites are already guarded via `is_char_boundary` / `char_indices().nth()` / `find()` offsets (Slack's `split_text_into_chunks` slices only at ASCII `\n`/space offsets, so it is safe).
- Added regression tests for `dashboard::truncate` covering the original multi-byte panic, char-boundary truncation, and char-vs-byte counting.
- **Scope boundary:** Does NOT add grapheme-perfect truncation. Bluesky's canonical limit is 300 graphemes; `chars()` (codepoints) is panic-safe and exact for ASCII/CJK, but slightly conservative for ZWJ/skin-tone emoji. Fully exact grapheme truncation would require adding `unicode-segmentation` (not currently a dependency) and is left out of this panic fix.
- **Blast radius:** Three independent truncation sites (zerocode TUI dashboard, Bluesky channel send, LinkedIn tool image prompt). No shared helper, no API/behavior change beyond no-longer-panicking and counting characters instead of bytes.
- **Linked issue(s):** None.
- **Labels:** `bug`, `size: XS`, `risk: high`, `channel`, `tool`, `channel:bluesky`.

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy -p zerocode -p zeroclaw-channels -p zeroclaw-tools --all-targets -- -D warnings
cargo test -p zerocode -p zeroclaw-channels -p zeroclaw-tools
```

- **Commands run and tail output:**

`cargo fmt --all -- --check` — clean (exit 0, no diff).

`cargo clippy ... --all-targets -- -D warnings` — clean:
```
    Checking zeroclaw-tools v0.8.0-beta-2 (.../crates/zeroclaw-tools)
    Checking zeroclaw-channels v0.8.0-beta-2 (.../crates/zeroclaw-channels)
[...]

PR #7154, opened 2026-06-03
fix(config): avoid char-boundary panic in attached short-option parsing

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `attached_short_option_value()` in `zeroclaw-config`'s command policy parser
    sliced the option token at byte index 1 (`body[1..]`) while only guarding on
    byte length (`body.len() < 2`). When the character immediately after the
    leading `-` is a multibyte UTF-8 codepoint, byte index 1 falls inside that
    codepoint and the slice panics with `start byte index 1 is not a char
    boundary`.
  - The release profile sets `panic = "abort"`, so that panic on a
    `tokio-rt-worker` thread took the whole daemon down with SIGABRT rather than
    failing the single task. Replaced the byte slice with a char-boundary-safe
    skip of the first character (`chars.next()` + `chars.as_str()`), preserving
    identical behavior for all ASCII tokens.
  - Added a regression test covering a multibyte token, a bare multibyte option,
    and the existing ASCII cases.
- **Scope boundary:** Only `attached_short_option_value` and its new test. No
  change to redirection parsing, path classification, or policy decisions. ASCII
  token handling is byte-for-byte identical.
- **Blast radius:** Command security policy parsing only. The fix narrows a
  panic path to a normal `Option` return; it cannot widen what tokens are
  accepted as attached values (the slice content is unchanged for valid UTF-8).
- **Linked issue(s):** Related #7123 (broader UTF-8 char-boundary sweep across
  different files; not a dependency of this PR).
- **Labels:** `bug`, `size: XS`, `risk: medium`, `config`.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-config --all-targets -- -D warnings
```

  - `cargo fmt --all -- --check` → clean, exit 0.
  - `cargo clippy -p zeroclaw-config --all-targets -- -D warnings` →
    `Finished dev profile ... ` exit 0, no warnings.
  - `cargo test -p zeroclaw-config --lib` →
    `test result: ok. 717 passed; 0 failed; 0 ignored`.
  - Targeted: `policy::tests::attached_short_option_value_handles_multibyte_token ... ok`.
- **Beyond CI — what did you manually verify?** Reproduced the original crash
  signature from a daemon coredump (`tokio-rt-worker` panic at
  `policy.rs:1097`, "start byte index 1 is not a char boundary; it is inside a
  3-byte char"). Confirmed the new test asserts `-é/etc/passwd → /etc/passwd`
  (the previously-panicking input) and `-— → None`. Did NOT run the full
  workspace `cargo test` — change is contained to one crate with no cross-crate
  surface.
- **If any command was intentionally skipped, why:** Clippy and test were
  scoped to `zeroclaw-config` rather than the full workspace. The edit is a
  single private function with no public-API or cross-crate surface change, so
  no other crate compiles differently; CI runs the full battery.

#### 7. Attached short-option char-boundary panic (#7154) — Tier 2

Commit `8bbc5a098`.

- [x] Attached short option with a multibyte char after the flag → no panic, parses — PASS: `attached_short_option_value_handles_multibyte_token` (`-é/etc/passwd`→`/etc/passwd`, `-—`→None). Was a char-boundary slice panic on `body[1..]`, fixed via char-iterator skip.
- [x] `-p value` and `-pvalue` still parse — PASS (same test: `-f/etc/passwd`→`/etc/passwd`, `-f`→None, `--long`→None).

## Original Crash Evidence

Captured from a local daemon coredump (paths sanitized). `coredumpctl`:

```
PID: <pid> (zeroclaw)
Signal: 6 (ABRT)
Command Line: zeroclaw daemon --ephemeral --config-dir <config-dir>
Message: Process <pid> (zeroclaw) of user <uid> dumped core.
```

Panic payload recovered from the core image:

```
thread 'tokio-rt-worker' panicked at crates/zeroclaw-config/src/policy.rs:1097:21:
start byte index 1 is not a char boundary; it is inside a 3-byte char
```

The release profile sets `panic = "abort"`, so this worker-thread panic
terminated the process with SIGABRT (signal 6) instead of unwinding a single
task.

- If any `Yes`, describe the risk and mitigation: N/A. The fix converts a
  process-killing abort into a normal `Option` return: any command whose token
  list reaches the policy parser with a `-<multibyte>` token previously aborted
  the daemon. This is a small availability hardening; it does not change which
  commands are permitted.

## Compatibility (required)

- Backward compatible? `Yes`

- **Fast rollback command/path:** `git revert <sha>`. The change is one private
  helper plus its test, no schema/config/CLI surface, so a revert is safe and
  self-contained. Reverting restores the previous `body[1..]` slice (and its
  panic-on-multibyte behavior); only do so if the fix itself regresses ASCII
  short-option parsing.
- **Feature flags or config toggles:** None. The parser helper is always on the
  command-policy path and is not gated.
- **Observable failure symptoms:** A regression reappears as a daemon SIGABRT
  (signal 6) from a `tokio-rt-worker` thread; grep logs / coredumps for
  `start byte index 1 is not a char boundary` at `crates/zeroclaw-config/src/policy.rs`.
  The daemon process exits rather than failing a single task (release profile is
  `panic = "abort"`), so the operator-visible signal is the whole daemon
  restarting/cycling, not a logged task error.

### DIFF
--- a/crates/zeroclaw-config/src/policy.rs
+++ b/crates/zeroclaw-config/src/policy.rs
@@ -1094,7 +1094,9 @@ fn attached_short_option_value(token: &str) -> Option<&str> {
     if body.starts_with('-') || body.len() < 2 {
         return None;
     }
-    let value = body[1..].trim_start_matches('=').trim();
+    let mut chars = body.chars();
+    chars.next();
+    let value = chars.as_str().trim_start_matches('=').trim();
     if value.is_empty() { None } else { Some(value) }
 }
 
@@ -5502,5 +5504,22 @@ mod tests {
         let t = PerSenderTracker::new();
         // Key "ghost" has never been recorded — should not be exhausted at max=1
         assert!(!t.is_exhausted("ghost", 1));
+    }
+
+    #[test]
+    fn attached_short_option_value_handles_multibyte_token() {
+        // A multibyte char immediately after the dash must not panic on a
+        // byte-index slice. Regression for a char-boundary abort.
+        assert_eq!(
+            attached_short_option_value("-é/etc/passwd"),
+            Some("/etc/passwd")
+        );
+        assert_eq!(attached_short_option_value("-—"), None);
+        assert_eq!(
+            attached_short_option_value("-f/etc/passwd"),
+            Some("/etc/passwd")
+        );
+        assert_eq!(attached_short_option_value("-f"), None);
+        assert_eq!(attached_short_option_value("--long"), None);
     }
 }


NiuBlibing:
https://github.com/zeroclaw-labs/zeroclaw/pull/7123
A similar problem which relates with utf-8.

WareWolf-MoonWall (approved):
## Context

Reviewing as @WareWolf-MoonWall. This is a surgical bug fix in `zeroclaw-config`'s command policy parser that converts a daemon-killing panic into a normal `Option` return. The PR addresses a UTF-8 char-boundary slice panic discovered via a production coredump, where a multibyte character immediately after a `-` in an attached short option (like `-é/etc/passwd`) caused a byte-index slice at position 1 to land inside a 3-byte character. Under `panic = "abort"`, this crashed the entire daemon with SIGABRT.

The fix replaces the unsafe byte slice `body[1..]` with a char-boundary-safe iterator skip (`chars().next()` + `chars.as_str()`), preserving identical behavior for all ASCII tokens while correctly handling multibyte characters.

---

### 🟢 What looks good

**Coredump reproduction evidence.** The "Original Crash Evidence" section with the coredumpctl output and the exact panic payload (`policy.rs:1097:21: start byte index 1 is not a char boundary; it is inside a 3-byte char`) is exactly the kind of validation that demonstrates you've reproduced the failure mode before fixing it. This is textbook reliability engineering — capture the failure, fix the root cause, add a regression test.

**Scoped validation.** Running `cargo clippy -p zeroclaw-config` and `cargo test -p zeroclaw-config --lib` rather than the full workspace is the right call for a private-function change with no cross-crate surface. The PR body explicitly notes CI will run the full battery, so you're not skipping coverage — you're just avoiding redundant local work. The targeted test confirmation (`policy::tests::attached_short_option_value_handles_multibyte_token ... ok`) shows the regression test fired and passed.

**Char-boundary safety pattern.** The `.chars().next()` + `.chars.as_str()` idiom is the correct way to skip the first character without assuming byte alignment. The old code checked `body.len() < 2` (byte length) but sliced at `body[1..]` (byte index), which is the classic char-boundary trap. The new code operates entirely in the char domain, so it cannot panic on valid UTF-8 regardless of where the first character's byte representation ends.

**Test coverage.** The regression test covers:
- The exact case that panicked (`-é/etc/passwd` → `/etc/passwd`)
- A bare multibyte option (`-—` → `None`)
- The existing ASCII cases (`-f/etc/passwd`, `-f`, `--long`)

This is the minimum viable set to confirm the fix and prevent the regression. No redundant test explosion, no unnecessary parameterization.

**Template compliance.** Every required section is filled with substance. The scope boundary is explicit ("only `attached_short_option_value` and its new test"), the blast radius is clear ("command security policy parsing only"), and the validation evidence includes both the commands run and what was intentionally skipped with justification.

---

### 🔵 Suggestion — Cross-link the UTF-8 safety sweep

@NiuBlibing's comment links PR #7123, which is a parallel UTF-8 char-boundary fix sweep across three other sites (zerocode dashboard truncation, LinkedIn image prompt, Bluesky message truncation). That PR is part of the same correctness push. Consider adding a "Related: #7123" line to the "Linked issue(s)" section in the PR body to make the connection explicit — it helps future readers understand this is part of a deliberate codebase-wide safety audit rather than an isolated incident.

This is purely for traceability; the current body is already complete.

---

## Verdict

**Approve.** This is a clean, well-evidenced fix for a real crash. The approach is correct, the scope is minimal, and the validation evidence demonstrates both the original failure and the fix. No blockers.

The PR is assigned to milestone v0.8.0, which is the correct placement for a bug fix in `zeroclaw-config` (Beta tier, stable at v0.8.0). The fix narrows a panic path without changing behavior for valid ASCII tokens, so it's low risk and backward compatible.

Ready to merge once CI confirms.

Audacity88 (approved):
I reviewed #7154 at head 8bbc5a0, including the PR body, the current diff, visible checks, @WareWolf-MoonWall's approval, and @NiuBlibing's #7123 cross-link.

@singlerider, I agree with @WareWolf-MoonWall's technical approval. This is still a draft, so I am treating this as approval of the current fix rather than a final ready-to-merge statement.

### 🟢 What looks good — the panic fix is the right shape

The old code stripped the leading dash and then sliced `body[1..]`, which assumes byte index 1 is a valid character boundary. That is true for ASCII short options and false for a token like `-é/etc/passwd`. Switching to `chars().next()` plus `chars.as_str()` preserves the ASCII behavior while making the multibyte path safe.

### 🟢 What looks good — the regression test pins the important cases

The new test covers the previously crashing multibyte attached-value shape, a bare multibyte option, and the existing ASCII cases. That is the right behavior surface for this private parser helper: it proves the crash is gone without widening the command-policy parser's accepted token shape.

### 🟢 What looks good — scope and validation are appropriately narrow

This changes one private helper in `zeroclaw-config` and adds one focused test. The PR body explains the coredump evidence, scoped local validation, why full workspace tests were left to CI, and the security effect: this narrows a daemon-abort availability failure into a normal `Option` return without adding permissions, network access, credentials handling, or new policy allowances.

### 🟡 Warning — clean up the draft body before marking ready

Before this leaves draft, please update the PR body so the labels snapshot names the current labels (`bug`, `size: XS`, `risk: medium`, `config`) and the rollback section uses the medium-risk fields from the template. The code fix is small, but the applied risk label is still `risk: medium`, so the public body should not use the low-risk placeholder.

### 🔵 Suggestion — add the related UTF-8 sweep link before marking ready

The related #7123 PR covers different files, so I do not see it as a dependency. Still, adding `Related #7123` to the PR body would make the broader UTF-8 char-boundary cleanup easier to follow later.

Approving.

singlerider:
Thanks @Audacity88 — both addressed in the body:

🟢 **Labels snapshot** now names the applied set (`bug`, `size: XS`, `risk: medium`, `config`).

🟢 **Rollback** uses the medium-risk template fields — Fast rollback command/path, Feature flags or config toggles (None, the helper is always on the policy path), and Observable failure symptoms (SIGABRT from a `tokio-rt-worker`; grep `start byte index 1 is not a char boundary` at `crates/zeroclaw-config/src/policy.rs`). No low-risk placeholder remains.

🟢 **`Related #7123`** added under Linked issue(s), scoped as related-not-dependency since it touches different files.

Issue #6826, opened 2026-05-21
[Tracker]: Zerocode


singlerider:
ZeroClaw needs a standalone terminal user interface that provides parity with the web dashboard and serves as the primary operator interface for power users, headless servers, and enclave deployments. The TUI is a separate binary (`zerocode`, renamed during development from `zeroclaw-tui`) that communicates with the ZeroClaw daemon exclusively over a socket protocol — no in-process runtime embedding. All client surfaces (TUI, web dashboard, CLI) share common gateway/RPC primitives — the dispatch layer is the single source of truth. No client gets special endpoints; if the TUI needs it, the runtime exposes it, and the web dashboard and CLI can consume it too.

### Architecture

- [x] Shared widget/theme code lives in `apps/zerocode/` alongside the binary (the previous `crates/zeroclaw-tui` library crate was consumed by the move)
- [ ] Transport abstraction: Unix socket (local), WebSocket/WSS (remote), vsock (enclave/future) — #6293
- [x] Exact version match enforcement on connect via `/api/status`
- [x] `--ephemeral` daemon mode (#6818): TUI spawns `zeroclaw daemon --ephemeral` if no daemon is running; daemon self-terminates (configurable grace period) when last TUI client disconnects; daemon started without `--ephemeral` ignores client count
- [x] Multiple TUI instances per daemon
- [x] Move `crates/zeroclaw-tui` to `apps/zerocode` as a separate `publish = false` binary crate (#6821)

### Connection Modes

- [x] `zerocode` — local Unix socket (default)
- [x] `zerocode --connect wss://host:port` — remote daemon
- [ ] `zerocode --connect vsock://cid:port` — enclave (future, #6293)
- [x] Authentication for remote connections (pairing token validation on WSS/remote connect — moved from #6837 socket transport, where filesystem permissions handle local auth)

### Connection Resilience

- [x] Socket close reason propagation — both sides know intentional vs unintentional disconnect
- [x] Intentional disconnect: show reason, no retry
- [x] Unintentional disconnect: banner, retry with backoff for configurable `reconnect_timeout_secs`, composer disabled but history remains scrollable
- [x] Reconnect without restarting TUI — re-establish socket, resume session by ID
- [x] `--ephemeral` daemon crash: one automatic re-spawn attempt, then require manual intervention

### TUI Identity

- [x] Short UIDs assigned on `initialize` (e.g. `tui_a3f7b21c`)
[...]

PR #7158, opened 2026-06-03
fix(zerocode): graceful in-loop daemon reconnect and recovery

### LINKED ISSUE #7043: [Bug]: zerocode TUI never reconnects after daemon close; UI wedges (and stops redrawing under tiling WMs)
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

When the daemon closes for any reason (crash, `Ctrl+R` reload, service restart, EOF on the IPC socket), the `zerocode` TUI becomes permanently unresponsive and never reconnects — even though the daemon comes back up. The status bar shows `● Disconnected (reason: EOF (daemon closed connection))` and the UI is frozen. Every connected TUI is affected independently (observed with two sessions: both die from a single daemon close). Under a tiling window manager the wedged pane also stops redrawing entirely — a retile/resize leaves a stale frame at the old geometry that never repaints.

### Expected behavior

When the daemon connection drops, the TUI should attempt to reconnect with backoff and, on success, resume against the **new** connection (UI live again, identity reclaimed via the existing `tui_id`/`tui_sig` handshake). If reconnection cannot succeed within a bounded window, the TUI should surface an actionable state ("daemon unreachable — press R to retry / Q to quit") rather than spinning forever. Disconnected and reconnecting states must continue to redraw and must adapt to terminal resize so the pane never sits on a stale frame under a tiling WM.

### Steps to reproduce

```bash
1. zeroclaw daemon                 # start daemon
2. zerocode                        # open the TUI (Unix socket to data/daemon.sock)
3. # in another terminal, stop or reload the daemon:
   pkill -f "zeroclaw daemon"      # or trigger Ctrl+R reload inside the TUI
4. # restart the daemon:
   zeroclaw daemon
5. # observe: the TUI stays on the red "Disconnected (EOF)" status bar
#   forever; it never reconnects to the new daemon. Open a second TUI
#   before step 3 and both wedge identically. Under a tiling WM, resizing
#   the wedged window leaves a stale, non-redrawing frame.
```

### Impact

```
Affected users: every zerocode user, every time the daemon restarts or
                reloads — including the built-in Ctrl+R reload flow.
Frequency: always (deterministic) on any daemon close.
Consequence: TUI is unrecoverable; user must kill and relaunch it, losing
             UI state. Multiple TUIs all die from one daemon close. Under
             tiling WMs the window is visually dead (no redraw).
```

### Logs / stack traces

```
Status bar: ● Disconnected (reason: EOF (daemon closed connection))
No panic — the process is alive but spinning; the UI never resumes.
```

## Root cause

Two distinct defects, both in the reconnect/redraw path.

**1. The reconnect loop discards the reconnected client.** In
`apps/zerocode/src/main.rs`, `run_with_reconnect` reconnects but throws the
result away:

```rust
loop {
    tokio::time::sleep(Duration::from_secs(1)).await;
    let result = match target { /* connect / connect_wss */ };
    if let Ok(_c) = result {   // <-- new RpcClient bound to _c, then dropped on break
        break;
    }
}
```

`rpc` is an `Arc<client::RpcClient>` passed **by value** into
`run_with_reconnect`; there is no `&mut` and no interior-mutable swap, so the
freshly-connected `_c` cannot be handed back to the loop. The outer loop then
re-invokes `app::run(Arc::clone(&rpc), ...)` with the **original, dead** `rpc`.
`app::run` immediately sees `ConnectionState::Disconnected` (the 2s grace check
at the top of the loop in `apps/zerocode/src/app.rs`), returns `Ok(true)`, and
the cycle repeats forever — a busy 1s reconnect spin whose result is never
applied. Net effect: reconnection "succeeds" at the socket layer but the UI
keeps using the dead connection. The `_c` binding is the tell that this was
never wired up.

**2. No redraw / resize handling while disconnected.** While disconnected,
`app::run` short-circuits on the disconnect grace timer and returns before the
event loop can service a `Resize`. Combined with defect (1) never re-entering a
live `app::run`, the pane is stuck at its old geometry. Under a tiling WM, a
retile/resize after the daemon dies leaves a stale, non-redrawing frame.

## Recommended fix steps

1. **Rebind the reconnected client into the UI.** Make `run_with_reconnect`
   own a `let mut rpc` and replace it with the new `Arc` after the inner loop
   succeeds (`rpc = Arc::new(new_client)`), then pass that into the next
   `app::run`. The reconnected client must become the one the UI holds — not be
   dropped. Preserve the existing `prev_id`/`prev_sig` identity-reclaim path.
2. **Bound the reconnect loop with backoff.** Replace the unconditional 1s
   sleep-and-retry with capped exponential backoff and a max attempt count or
   deadline. On exhaustion, return to a UI state that offers an explicit retry
   (R) and quit (Q) instead of spinning a busy loop forever.
3. **Keep the UI live and redrawing while disconnected/reconnecting.** Continue
   to `term.draw` on the tick cadence in the disconnected state, and handle
   `Event::Resize` so geometry changes repaint. Force a full clear+redraw on
   resume so a tiling-WM resize during the outage cannot leave a stale frame.
4. **(Optional) Offer local respawn.** For `ConnectTarget::LocalSocket`, offer
   to respawn via the existing `spawn_ephemeral_daemon` path, guarded against
   the multi-TUI stampede the current comment warns about (e.g. a spawn
   lockfile under `data/`), so a single TUI can recover a locally-managed
   daemon without an external service manager.
5. **Add a regression test** at the `run_with_reconnect` seam: simulate a
   disconnect followed by a successful reconnect and assert the loop proceeds
   with the **new** client (not the dead one) — i.e. the next `app::run` sees a
   `Connected` state.

### ZeroClaw version

0.8.0-beta-2 (commit 969b60990, branch integration/zeroclaw-tui)

### Rust version

### Operating system

Arch Linux (Linux 7.0.8-arch1-1 x86_64); reproduced under a tiling window manager

### Regression?

Unknown

### Pre-flight checks

### LINKED ISSUE #7125: [Bug]: TUI (zerocode) freezes entirely when daemon disconnects
### Affected component

unknown

### Severity

S1 - workflow blocked

### Current behavior

When using the TUI (zerocode), if the zeroclaw daemon/main process is terminated (disconnects), the entire TUI freezes and becomes unresponsive. No operation is possible and the user must force-quit.

### Expected behavior

When the daemon disconnects, the TUI should remain responsive: show a clear "connection lost" indicator and allow the user to quit or wait for automatic reconnection. The interface should not freeze entirely.

### Steps to reproduce

```bash
1. Start the zeroclaw daemon
2. Start the TUI: zerocode
3. Switch to the conversation page and send any message, e.g. "hello"
4. Terminate the zeroclaw process (disconnect)
5. Observe the TUI (zerocode): the entire interface freezes and is unresponsive
```

### Impact

Affected users: all TUI (zerocode) users; Frequency: always reproducible when the daemon is terminated; Consequence: the interface becomes completely unusable and requires a force-quit.

### Logs / stack traces

_No response_

### ZeroClaw version

40be7738f

### Rust version

### Operating system

Linux 7.0.10-1-MANJARO x86_64

### Regression?

No, first-time setup

### Pre-flight checks

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - On daemon disconnect the TUI bounced out of its event loop into a blind
    `sleep + connect` retry loop that serviced no terminal input and drew
    nothing, so a dead daemon left zerocode completely unresponsive — not even
    quittable. The inner loop then dropped the freshly reconnected client and
    re-entered the UI against the original dead handle, so the TUI stayed dead
    even after the daemon came back.
  - Reconnection now lives inside `app::run`: the loop keeps drawing the cached
    panes and servicing input (quit always works) while it reconnects in place,
    throttled to once per second.
  - A TUI-owned ephemeral daemon is respawned exactly once. If that respawn
    does not return, the status bar requests intervention while the connect
    poll keeps running, so a manually restarted daemon still recovers. Attached
    daemons (external socket / WSS) are never spawned, only polled — preserving
    the anti-stampede rule.
  - On a successful reconnect the recovered client is adopted and the full pane
    set is rebuilt against it. Rebuild is mandatory: chat and logs cache a
    notification receiver bound to the old client, so a kept-alive pane would
    go silently deaf. History is not bulk-reloaded — panes refetch lazily and
    the daemon rehydrates the session from its durable row on the next prompt.
    TUI identity (`tui_id`/`tui_sig`) is reclaimed on every reconnect.
- **Scope boundary:** TUI (`apps/zerocode`) only. No daemon, RPC-protocol, or
  durable-store changes — the server already rehydrates reaped sessions from
  the durable store on demand; this PR only makes the client use that path.
- **Blast radius:** zerocode connection lifecycle for both local-socket and
  WSS transports. `Dashboard`, `Logs`, and `config_manager::App` change from a
  borrowed `&RpcClient` to `Arc<RpcClient>` (mechanical; all call sites deref
  unchanged). The dead `run_with_reconnect` path and the unused
  `is_disconnected` helper are removed.
- **Linked issue(s):** Closes #7125, Closes #7043. Relates to #6826.
- **Labels:** `bug`, `zerocode` (snapshot after apply).

```bash
cargo fmt --all -- --check
cargo test -p zerocode
```

  - `cargo fmt --all -- --check` → exit 0, clean.
    `Finished dev profile ...`, exit 0, no warnings.
  - `cargo test -p zerocode` →
    `test result: ok. 201 passed; 0 failed; 0 ignored` (lib) and
    `test result: ok. 14 passed; 0 failed; 0 ignored` (bin).
- **Beyond CI — what did you manually verify?** Traced the four recovery
  paths by reading the rebuilt loop:
  (1) disconnect keeps drawing + input serviced; quit is handled before the
  disconnect pane-handler skip gate, so the UI is always quittable;
  (2) owned ephemeral daemon respawns exactly once, then surfaces
  "restart required" while still polling;
  (3) a daemon reappearing at the expected address reconnects, reclaims
  identity, and rebuilds panes;
  (4) a daemon flap *during* the post-reconnect pane rebuild is treated as a
  transient disconnect (the rebuild helper yields a `Result`) rather than
  tearing down the TUI.
  I did NOT run a live end-to-end kill/restart against a running daemon in
  this environment — reviewers exercising the manual matrix should confirm the
  WSS reconnect path on real network latency.
- **If any command was intentionally skipped, why:** clippy/test scoped to
  `zerocode`; the diff is confined to that crate with no cross-crate surface
  change. CI runs the full workspace battery.

#### 13. Graceful in-loop daemon reconnect + resume (#7158) — Tier 5 ⚠️

Commits `6bdb60768` (#7158), `1a8913f9b` (resume prior session — direct).
Kills the daemon mid-session. Throwaway session only.

- [~] Kill daemon mid-session → TUI recovers in-loop, no crash; restart → reconnects (`6bdb60768`) — CODE-VERIFIED, LIVE PENDING. Reconnection moved into `app::run` so cached panes stay alive and the input pump never stops (quit always works); reconnect throttled to once/sec in-loop. No isolated unit test for the loop itself (it's the event loop); requires a live daemon-kill. The earlier-observed "completely unresponsive on dead daemon" is the exact failure this fixes — verify live.
- [~] On reconnect the TUI RESUMES the prior session, does not start a fresh one (`1a8913f9b`) — UNIT-COVERED (resume mechanism), LIVE PENDING (round trip). The resume-id carry is unit-green: `current_session_id_reports_active_session` (chat.rs:4355) proves the live id is reported; `resume_session_id_dropped_when_init_lands_in_multi_agent_picker` (chat.rs:4368) proves the carried id is dropped on the multi-agent picker (no mismatched reattach). Wiring confirmed in code: `app::run` captures each pane's `current_session_id()` before the rebuild (app.rs:417-418), seeds new panes via `set_resume_session_id` (app.rs:178-181), and Chat consumes it into `session_new_with_id` (chat.rs:268-275) so the daemon (#7182, which now RETAINS the session) rehydrates history. ⚠️ The §3 EMPTY-session-on-recovery finding is precisely the bug this commit fixes — MUST re-verify live on a binary built AT/AFTER `1a8913f9b` to confirm history rehydrates and the empty-session regression is gone.
- [~] Native cursor still visible after a reconnect/resume cycle (`c24a65662` cross-check) — LIVE PENDING. Ties to the §4 open observation; verify in the same live reconnect pass.

##### Reconnect resume: empty-session + wrong-cwd fixes (`f7bcbbc7c`, `be0d6a02b`, `b3a5cb0a7`, `6e2f340b7`)

These four commits directly fix the §3/§13 empty-session-on-recovery finding.
Direct on integration; NOT yet cherry-picked to `fix/zerocode-graceful-reconnect-recovery`.

- [ ] Multi-agent reconnect reattaches the PRIOR agent's retained session, does NOT mint a fresh empty one (`f7bcbbc7c`) — the `1a8913f9b` resume-carry dropped the id whenever init landed in the multi-agent picker, which is every reconnect with ≥2 agents → exact empty-session bug. Now carries prior agent alias; if still present, reattaches via `session_new_with_id/_acp` and skips the remote-ACP cwd picker. Id dropped only when no carried agent matches. UNIT: `multi_agent_reconnect_reattaches_prior_agent_session` (asserts session/new with prior agent+session id) + retained no-match drop case. LIVE PENDING — re-run the §13 kill→reconnect round trip with ≥2 enabled agents; confirm history rehydrates, not empty.
- [ ] Resumed session replays its transcript (not blank) (`be0d6a02b`) — history was loaded only by the manual Sessions overlay, never by the resume path. Extracted to `ChatState::load_history`, now called from both overlay and `start_session` on resume; mirrors user/assistant rows, drops system rows, seeds the pinned first-message recovery row. Fresh sessions skip the fetch. UNIT: `load_history_replays_transcript_and_seeds_first_message`. LIVE PENDING.
- [ ] Resume keeps the retained session's cwd, does NOT re-point to launch dir (`b3a5cb0a7` client + `6e2f340b7` daemon) — `start_session` (client) now passes NO cwd when a session id is carried, so the daemon keeps the retained cwd; `handle_session_new` (daemon) recovers the original cwd from the persisted store on a no-cwd ACP resume and uses it for agent build, in-memory RpcSession, and reported `workspace_dir` (peeked data reused, store not read twice). Fresh/caller-cwd sessions unchanged. UNIT: `acp_resume_recovers_persisted_cwd`. LIVE PENDING — confirm chat/Code pane shows the original path after recovery.

**§13 STATUS: resume-id MECHANISM unit-green (2 tests) + full wiring code-verified. The reconnect-resume ROUND TRIP (kill daemon → reconnect → history rehydrated, not empty) is genuinely live-only and is the key remaining verification — the prior empty-session finding predates this fix and must be re-tested on a current binary. UPDATE: the four reconnect commits (`f7bcbbc7c`/`be0d6a02b`/`b3a5cb0a7`/`6e2f340b7`) target that exact empty-session + wrong-cwd failure (multi-agent reattach, transcript replay, retained cwd) — all unit-green, all LIVE PENDING. NOT yet cherry-picked to the #7158 branch.**

- New external network calls? `No` — same transports (local socket / WSS),
  same `initialize` handshake; only the retry/recovery control flow changed.
- Secrets / tokens / credentials handling changed? `No` — `tui_id`/`tui_sig`
  reclaim uses the existing handshake parameters unchanged.
- If any `Yes`, describe the risk and mitigation: N/A. Net effect is an
  availability improvement: a disconnected TUI is no longer frozen and recovers
  without losing the user's session.

## Compatibility (required)

- Backward compatible? `Yes`

## Rollback

- **Fast rollback command/path:** `git revert <sha>` — single self-contained
  commit, no migrations or persisted state involved.
- **Observable failure symptoms:** zerocode failing to recover after a daemon
  restart (status bar stuck on "Reconnecting…" with a daemon listening), the
  UI becoming unresponsive while disconnected, or panes going silent after a
  reconnect (stale notification receiver). On WSS, a brief once-per-second UI
  stutter during reconnection is expected, not a regression (see note below).

## Notes for reviewers

- **Known limitation:** the reconnect attempt (`target.connect().await`) runs
  inline in the loop, throttled to once per second. A local-socket handshake is
  effectively instant; a WSS handshake (TLS + network) can briefly stutter the
  UI up to the handshake duration, once per second, while reconnecting. This is
  a large improvement over the prior permanent freeze; moving the connect
  attempt off-thread is a reasonable follow-up.
- **Net diff is positive (~+237/−145).** `main.rs` and `client.rs` shrink (the
  blind reconnect machinery and an unused helper are deleted); `app.rs` grows
  because responsive in-loop recovery, the respawn-once policy, and lazy pane
  rebuild are genuinely more logic than the blind spin they replace.


WareWolf-MoonWall (approved):
Reviewed at 6bdb607 (the only commit in the PR).

---

## Context

This PR addresses two critical UX defects in `zerocode` (#7125, #7043): the TUI froze completely on daemon disconnect, and the reconnect machinery discarded the recovered client so recovery never happened. Both issues made the TUI unusable during any daemon restart or service reload — a complete workflow blocker.

The fix moves reconnection into the `app::run` event loop so the UI stays responsive and redrawing while disconnected, respawns an owned ephemeral daemon exactly once (preserving the anti-stampede discipline for attached daemons), and rebuilds every pane against the recovered client so notification channels are fresh. The `Arc<RpcClient>` handle is reassigned in place on successful reconnect; the prior design passed the client by immutable value and had no path to rebind it.

I traced the four recovery paths the author documented: disconnect keeps the UI responsive and quittable, ephemeral respawn is bounded to one attempt, a daemon reappearing reconnects and reclaims identity, and a flap mid-rebuild is treated as transient. The approach is sound and directly addresses the root causes identified in #7043. The inline once-per-second throttle is a known tradeoff (WSS can briefly stutter during reconnect); moving that off-thread is a reasonable follow-up but not a blocker here.

---

### 🟢 What looks good — lifecycle discipline

The respawn-once policy (`ephemeral_respawn_done`) enforces the documented boundary: a TUI-owned ephemeral daemon gets one automatic respawn, then the status bar surfaces "restart required" while the poll loop keeps running. This preserves both recovery automation (the common case succeeds without manual intervention) and anti-stampede discipline (multiple TUIs do not all spawn their own daemon). Attached daemons (external socket / WSS) are never spawned, only polled. That distinction is the right call.

---

### 🟢 What looks good — pane rebuild is mandatory

Chat and Logs panes cache a `notif_rx` receiver bound to the old client's notification channel. Keeping the panes alive across reconnect would leave them silently deaf — notifications would never arrive. The `build_panes!()` macro enforces a full rebuild on every reconnect so every pane re-subscribes to the recovered client's channel. That's the correct fix; lazy pane refetch and daemon-side session rehydration handle history without bulk-reloading it into the TUI.

---

### 🟢 What looks good — identity reclaim via existing handshake

TUI identity (`tui_id`/`tui_sig`) is reclaimed on every reconnect using the existing `prev_id`/`prev_sig` parameters to `connect()`/`connect_wss()`. The daemon's HMAC signature verification restores the same UID so the session survives. No new RPC surface, no protocol change — the handshake path was already built for this. Clean reuse.

---

### 🟢 What looks good — return type and call site simplification

The old `app::run` returned `Result<bool>` where `true` meant "reconnect" and `false` meant "user quit." That bool was load-bearing in `main.rs:run_with_reconnect`, which had to wrap `app::run` in its own outer reconnect loop. The new `app::run` owns the full lifecycle and returns `Result<()>` when the user quits — no intermediate bool, no outer loop. The call site in `main.rs` shrinks to a straight `app::run(...).await` inside the SIGTERM select. That's a material API improvement: the reconnect concern lives where it belongs (inside the event loop that can stay responsive during it) rather than split across two functions.

---

### 🟡 Warning — `needs_intervention` status message needs better contrast

When `needs_intervention` latches (the ephemeral daemon was respawned once and did not come back), the status bar displays "Daemon unavailable — restart required (reason)". Both the live-reconnecting state and the needs-intervention state use `DEAD_RED` as the status color (lines 621 + 625 in `app.rs` after the patch). The visual distinction between "reconnecting (transient, will recover)" and "restart required (manual intervention needed)" is only the text — no color or icon difference.

For a user watching the status bar, the red dot + red text is identical in both cases. If the daemon respawn silently fails (e.g., a missing config file, a port conflict, a PATH issue), the user sees "Reconnecting…" turn into "Daemon unavailable — restart required" with no visual signal that the state transition happened, only the word change. A user might wait indefinitely expecting automatic recovery when manual intervention is actually required.

**Suggestion (not blocking):** Use a distinct color or icon for `needs_intervention` — maybe `DEAD_RED` with a different Unicode marker (`\u{26a0}` warning sign instead of the filled circle), or a yellow/amber color to distinguish "manual action needed" from "automatic retry in progress". Alternatively, the "restart required" message could include an explicit hint like "(press R to retry manually)" if such a binding exists, making the manual-action expectation concrete rather than relying on color alone.

This is a UX polish item, not a correctness issue — the functionality works, and the text does convey the difference. But surfacing that state shift more prominently would help users who are not reading the status text carefully (especially under a tiling WM where the pane might be small).

---

### 🟢 What looks good — `#[allow(clippy::too_many_arguments)]`

`app::run` now takes 8 parameters (up from 6). The new ones are `target: &ConnectTarget` and `owns_ephemeral: bool`, both load-bearing for the in-loop reconnect logic. An `AppConfig` struct bundling these would reduce the signature width, but that's a refactor without a clear forcing function — the function is called exactly once, from a context (`main.rs:run_until_exit`) that already has all eight values on hand. The `#[allow]` documents the tradeoff explicitly. Acceptable as-is; if the signature grows further (9+ params), revisit with a config struct.

---

### 🟢 What looks good — macro hygiene in `build_panes!()`

The `build_panes!()` macro captures `rpc`, `connect_label`, `insecure_tls`, `config_dir`, and `reconnect_state` from the enclosing function scope, builds every pane, consumes any pending Quickstart Stage-2 intent, and evaluates to `anyhow::Result<(panes…)>`. It's invoked once at startup (where failure propagates with `?`) and again after each successful reconnect (where failure is caught and treated as a transient disconnect rather than tearing down the TUI).

The macro approach here is pragmatic: the pane-rebuild sequence is 30+ lines, used in two places with identical semantics except for error-handling strategy at the call site. A helper function would need to return the 6-tuple and take the same set of borrows; the macro avoids the tuple shuffle and keeps the error-handling context visible at the call site (`build_panes!()? →` startup fails, `match build_panes!() →` reconnect retries). The async block inside the macro is well-formed (no lifetime issues, no subtle capture bugs). This is a reasonable use of a macro for code generation where a function would be syntactically heavier without semantic gain.

---

### 🟢 What looks good — removed dead code

The old `run_with_reconnect` machinery in `main.rs` (which discarded the reconnected client) and the unused `is_disconnected()` helper in `client.rs` are both deleted. The net diff is ~+237/−145, but that's because responsive in-loop recovery genuinely requires more logic than the blind sleep-and-reconnect spin it replaces. The deletion of the old paths is correct — they were architecturally unsound (no rebind path for the live handle) and are fully superseded by the new in-loop design.

---

### 🟢 What looks good — Arc<RpcClient> adoption is mechanical

`Dashboard`, `Logs`, and `config_manager::App` change from `&'a RpcClient` to `Arc<RpcClient>` so the live handle can be swapped in place on reconnect. The diff shows no behavior change at call sites — the `Arc` derefs transparently, and the prior lifetime `'a` is dropped because the panes are now rebuilt on every reconnect (they don't need to outlive a single connection). The `acp`, `chat`, and `quickstart` panes already held an `Arc`. Mechanical refactor, correctly applied.

---

### 🟢 What looks good — validation coverage

The author ran the full `zerocode`-scoped battery (`fmt`, `clippy -p zerocode`, `test -p zerocode`) and traced the four recovery paths by reading the rebuilt loop. The manual verification accurately describes what the code does: disconnect keeps drawing, quit works, respawn is bounded, reconnect reclaims identity, and a flap during pane rebuild is transient.

The author **explicitly noted** they did NOT run a live end-to-end kill/restart against a running daemon or test the WSS reconnect path on real network latency. That's an honest disclosure and the right call for this environment — the CI matrix will exercise the socket path, and WSS reconnect latency is a known tradeoff (inline throttle can stutter briefly, documented in the PR body). Reviewers with access to a WSS setup should confirm the reconnect path there, but the absence of that test is not a merge blocker — the reconnect logic is transport-agnostic (same `target.connect().await` for both), and the inline throttle is a documented limitation with a clear follow-up path (move connect off-thread).

---

### 🟢 What looks good — PR template completeness

The PR body is exemplary. Every required section is filled, the scope boundary is explicit ("TUI only, no daemon/RPC/durable-store changes"), the blast radius is named, rollback is trivial (`git revert`), security impact is accurately "No" across the board (availability improvement, no new permissions/network/secrets), and the "Notes for reviewers" section surfaces the known WSS stutter limitation and explains the net-positive diff. Validation evidence includes exact command output. No AI attribution footer (correct per privacy contract). Labels snapshot is present. This is a model PR body.

---

### 🟢 What looks good — no new dependencies

The diff is pure Rust stdlib + existing workspace crates (`Arc`, `Mutex`, `Instant`, `Duration`, existing `RpcClient` and `ConnectTarget` types). No new `Cargo.toml` entries, no new external crates. The reconnect logic is built from existing primitives. Zero dependency bloat.

---

### 🟢 What looks good — no security surface expansion

Same transports (local socket / WSS), same `initialize` handshake, same `tui_id`/`tui_sig` reclaim parameters. The retry/recovery control flow changed, but the authentication and connection establishment paths are unchanged. The PR body's "No" answers to the security-impact questions are accurate. Net effect is an availability improvement (disconnected TUI recovers instead of wedging) with no new attack surface.

---

## Verdict

This is a well-executed fix for two S1-severity bugs that made `zerocode` completely unusable during any daemon restart. The architectural approach (move reconnect into the event loop, rebind the live `Arc` handle, rebuild panes on recovery) directly addresses the root causes identified in #7043. The respawn-once policy preserves anti-stampede discipline, the identity-reclaim path reuses the existing handshake, and the pane-rebuild enforces fresh notification subscriptions. The one UX polish item (status-bar contrast for `needs_intervention`) is a suggestion, not a blocker.

Approve with one non-blocking UX suggestion. The `needs_intervention` visual distinction is a reasonable follow-up; it does not block merge.

PR #7160, opened 2026-06-03
fix(config): resilient daemon config load with security-critical gating

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `Config::load` funneled every daemon startup through the strict whole-config `try_into()`. A single malformed field anywhere (e.g. an email alias missing `imap_host`) aborted the daemon before any repair surface (gateway `/api/config`, `zeroclaw config migrate`, dashboard) could run. The daemon must boot no matter what so the operator can fix the config. Adds `migrate_to_current_resilient` / `migrate_to_current_salvaged` as the daemon load path: it degrades instead of failing — strict deserialize first (zero change for valid configs), then drop invalid channel aliases, invalid whole-`[channels.<type>]` blocks, and invalid top-level sections to their serde `Default`, each logged with the offending path.
  - Security-critical sections (`security`, `risk_profiles`, `peer_groups`) get louder treatment: resetting one to its `Default` can *weaken* the running posture. These drops log at ERROR, surface in `ResilientLoad::dropped_security`, and are recorded on `Config.degraded_security`.
  - **Gates the degraded posture at every serving entry point.** Populating `degraded_security` is not enough — "secure by default" must fail loud (FND-006 §4.5). A shared `gate_security_posture` chokepoint now sits in front of `daemon`, `gateway start`, and `gateway restart`: with security-critical sections dropped, the process refuses to serve unless the operator explicitly opts in with `--allow-degraded-security`. When allowed, it boots (so the operator can reach repair surfaces) but emits a repeating ERROR-context WARN until the config is repaired. The daemon re-gates on each SIGUSR1 reload, so a repaired posture silences the warning and a newly-degraded one without the flag hard-fails the reload. Bare `zeroclaw gateway` fails closed.
  - Fixes a real salvage defect: the original top-level pruner only dropped a section whose *individual* removal validated the whole config. Two unrelated malformed sections validated nothing, so salvage gave up and the **entire** config silently reset to `Config::default()` — losing every valid section. Reworked into a joint probe (sole offender) plus an isolated probe (each section judged alone) so independent offenders are each dropped and valid siblings are kept.
- **Scope boundary:** Does NOT change the strict `migrate_to_current` (repair tooling still gets precise errors), does NOT run `Config::validate()` on the resilient path (a config that fails validation must still load so it can be fixed), and does NOT touch the web dashboard or any channel runtime behavior.
- **Blast radius:** Daemon/gateway config-load and startup-gating path. Valid configs deserialize through the unchanged strict path first, so there is zero behavior change for a healthy config. The new gate only triggers when a security-critical section fails to parse; healthy and non-security degradations are unaffected. The `degraded_security` field is `#[serde(skip)]` and never persisted.
- **Linked issue(s):** None.
- **Labels:** `bug`, `risk: high`, `size: L`, `config`

```bash
cargo fmt --all -- --check
cargo test
```

  - `cargo fmt --all -- --check` → clean (no diff; exit 0).
  - `cargo clippy --all-targets -- -D warnings` → `Finished`, no warnings or errors (exit 0). Notably caught a bare `tokio::spawn` for the nag task, which was switched to `zeroclaw_spawn::spawn!` so the spawned WARN stays attributed.
  - `cargo test` → all suites pass, zero failures (exit 0). Representative tails:
    ```
    test result: ok. 274 passed; 0 failed; 0 ignored; ...
    test result: ok. 188 passed; 0 failed; 0 ignored; ...
    test result: ok. 159 passed; 0 failed; 0 ignored; ...
    ```
  - Crate-scoped detail for the touched crate (`zeroclaw-config`):
    `cargo test -p zeroclaw-config` → `728 passed; 0 failed` (lib) +
    `89 passed; 0 failed` (integration), 1 pre-existing ignored.
  - New gate test in the binary crate:
    `cargo test --bin zeroclaw -- gate_security_posture cli_definition_has_no_flag_conflicts`
    → `2 passed; 0 failed` (gate fail-closed behavior + CLI flag-conflict check).
  - **Beyond CI — what I manually verified:** Salvage paths — broken channel alias dropped (sibling kept), broken whole-type channel block dropped (sibling type kept), malformed `[security]` reported as a security-critical drop, malformed non-security section reported as a plain drop, two independent malformed sections both salvaged in one load (including the mixed security/non-security case), unparseable TOML and future schema version both falling back to `Config::default()`, and the strict path still erroring for repair tooling. Gating — `load_or_init` surfaces a dropped `[security]` on `degraded_security`; `gate_security_posture` returns `Ok(None)` for a clean posture, `Err` for degraded-not-allowed, and a live nag handle for degraded-allowed.

#### 15. Resilient config load + security gating (#7160) — Tier 5 ⚠️, can refuse boot

Commits `c3529d599`, `1040db8a1`, `d198d1ce5`, `16d25e8b1`.
**Run dead last, on a throwaway config dir.**

- [x] Single malformed TOML block → daemon loads, repairs, logs the bad block (`c3529d599`) — PASS (unit). `broken_non_security_section_is_plain_drop_not_security` (migration.rs:1100): a malformed `[heartbeat]` block is salvaged — dropped from the plain list, the rest of the config loads. `migrate_to_current_salvaged` degrades instead of failing; `broken_channel_type_block_is_dropped_not_fatal` proves a valid sibling survives a broken-type drop.
- [x] Multiple bad blocks → all salvaged, none silently dropped (`1040db8a1`) — PASS (unit). `multiple_independent_bad_sections_all_dropped` (migration.rs:1143): both `[heartbeat]` and `[backup]` offenders appear in `dropped` (none silently lost). `multiple_bad_sections_one_security_critical` (1168): a mixed batch correctly sorts `[security]`→`dropped_security` and `[heartbeat]`→`dropped`, with no double-listing. 17/17 migration tests green.
- [x] Corrupt a security-critical section (autonomy / gateway pairing / allowlists) → refuses boot or runs degraded; NO permissive fallback (`1040db8a1`, `d198d1ce5`) — PASS (unit). `broken_security_section_is_reported_as_degraded` (migration.rs:1082): malformed `[security]` → classified `dropped_security` (security-critical), NOT a plain drop. `load_or_init_assigns_degraded_security_for_malformed_section` (schema.rs:20305): the load path stamps `config.degraded_security`. The bad section is reset to DEFAULTS (deny-by-default), never a permissive fallback; the gate then refuses to serve (below).
- [x] Every serving entry point gated: `gateway`, `daemon`, `service` with degraded config → each blocks (`16d25e8b1`) — PASS (code + bin test). `gate_security_posture` (main.rs:6075) fails closed when `degraded_security` is non-empty unless `--allow-degraded-security`; called at ALL serving entry points: Gateway Restart (3343), Gateway Start (3499) + get-paircode (3507, always `false`), Daemon initial boot (3527), Daemon reload-iteration (3644), Daemon newly-degraded-on-reload (3778). `service` runs the daemon path → transitively gated. Bin test `gate_security_posture_fails_closed_unless_allowed` (main.rs:6818): clean→no gate/no nag; degraded+not-allowed→Err (refuses); degraded+allowed→boots with an abortable nag handle (no leak — handle returned, aborted per reload + shutdown).
- [x] Valid config → clean boot, no behavioral change — PASS (unit). `gate_security_posture(&Config::default(), false)` returns `Ok(None)` (clean posture, no gate, no nag) — asserted in the bin test. A valid config has empty `degraded_security`, so the gate is a no-op and boot is unchanged.
- [x] **`b1a4ae45a`** Whole-config fallback is gated too — section-level salvage marked `dropped_security`, but the whole-config fallbacks (unparseable TOML, future schema, broken migration chain, unsalvageable root) defaulted the whole config with an EMPTY `dropped_security`, losing every security-critical section unsignalled. Both fallbacks now push `WHOLE_CONFIG_SENTINEL` so `load_or_init` surfaces it and the gate fails closed. Tests: `unparseable_config_marks_whole_config_degraded`, `future_schema_version_marks_whole_config_degraded`, `unsalvageable_root_marks_whole_config_degraded`, `load_or_init_marks_whole_config_degraded_for_unparseable_file`, bin gate test on the sentinel. Verify a corrupt config refuses boot without `--allow-degraded-security`.

**§15 STATUS: config-layer salvage + security classification fully unit-green (17 migration tests + schema degraded-load test). Gate logic covered by the bin test `gate_security_posture_fails_closed_unless_allowed` and confirmed wired at all 6 serving entry points. Live throwaway-config boot-refusal smoke optional — the fail-closed contract is unit-proven. Re: handoff — local #7160 work is committed on this branch (`c3529d599`/`1040db8a1`/`d198d1ce5`/`16d25e8b1`); the upstream PR push + re-review-request remains a separate action awaiting user direction.**

- Secrets / tokens / credentials handling changed? **No** — secret masking and the strict load path are unchanged.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — test fixtures use `example.com` placeholders only.
- If any `Yes`, describe the risk and mitigation: The change *improves* the security posture. Silent reset of `security` / `risk_profiles` / `peer_groups` to defaults is now an ERROR-level, machine-readable signal (`degraded_security`) that hard-stops every serving entry point (daemon, `gateway start`, `gateway restart`) unless the operator explicitly acknowledges it with `--allow-degraded-security` — closing the window where the daemon could serve a pairing/auth boundary with a weaker-than-intended posture.

## Compatibility (required)

- Backward compatible? **Yes** — valid configs go through the unchanged strict deserialize first; no on-disk format, schema, or behavior change for healthy configs. The new gate only fires when a security-critical section is malformed, which previously meant a silent reset.
- Config / env / CLI surface changed? **Yes** — adds an opt-in `--allow-degraded-security` flag to `zeroclaw daemon`, `zeroclaw gateway start`, and `zeroclaw gateway restart`. No new config keys or env vars; `degraded_security` is `#[serde(skip)]`.
- Upgrade steps: None for healthy configs. An operator whose config has a malformed `security` / `risk_profiles` / `peer_groups` section will now see the daemon refuse to start (previously it booted with a silently weakened posture). Repair the section (`zeroclaw config migrate` shows the precise error), or pass `--allow-degraded-security` to boot anyway and fix it via the gateway config editor.

## Rollback

Low-risk to revert despite the `risk: high` label (security-path scope, not blast radius): `git revert <sha>` across the PR's commits restores the prior strict-only load path and removes the gate + flag. No feature flags, no config toggles, no migration to undo.

- **Fast rollback command/path:** `git revert` the PR's squash commit (or the individual commits) and redeploy.
- **Feature flags or config toggles:** None. `--allow-degraded-security` is an opt-in CLI flag, not a persisted toggle.
- **Observable failure symptoms:** A daemon/gateway that previously started now exits at boot with `malformed security-critical sections (...)`; grep logs for `degraded_security` or `DEGRADED security`. Reverting restores the prior boot-anyway behavior.

### DIFF
--- a/crates/zeroclaw-config/src/migration.rs
+++ b/crates/zeroclaw-config/src/migration.rs
@@ -269,14 +269,9 @@ fn encrypt_in_place(value: &mut toml::Value, store: &crate::secrets::SecretStore
     Ok(())
 }
 
-/// High-level: arbitrary versioned TOML → fully validated V3 `Config`.
-/// Runs migration if needed, then deserializes into the current `Config` type.
-///
-/// **Strict.** Any deserialization defect is returned as an error. This is the
-/// contract repair tooling depends on (`zeroclaw config migrate`,
-/// `model_routing_config`): they need to know precisely what is wrong. The
-/// daemon load path uses [`migrate_to_current_resilient`] instead, which never
-/// fails so the operator can always reach a repair surface.
+/// Versioned TOML → validated V3 `Config`, strict: any defect errors.
+/// Used by repair tooling (`zeroclaw config migrate`, `model_routing_config`)
+/// that needs the precise failure. Daemon load uses the resilient path.
 pub fn migrate_to_current(input: &str) -> Result<Config> {
     let final_value = migrate_value(input)?;
     final_value
@@ -284,60 +279,35 @@ pub fn migrate_to_current(input: &str) -> Result<Config> {
         .context("migrated config failed to deserialize as current schema")
 }
 
-/// Daemon load path: arbitrary versioned TOML → a usable `Config`, **always**.
-///
-/// The daemon must start no matter what so the config can be repaired through
-/// the gateway, CLI, or dashboard. This function therefore degrades instead of
-/// failing; see [`migrate_to_current_salvaged`] for the salvage contract and
-/// the security-section caveat. This thin wrapper discards the salvage report
-/// for callers that only need the `Config`.
+/// Daemon load path: versioned TOML → usable `Config`, never failing.
+/// Thin wrapper over [`migrate_to_current_salvaged`] that drops the report.
 pub fn migrate_to_current_resilient(input: &str) -> Config {
     migrate_to_current_salvaged(input).config
 }
 
-/// Top-level config keys whose silent loss could *weaken* the running
-/// security posture. "Secure by default" cuts both ways: if one of these
-/// sections is malformed, reverting it to its serde `Default` may grant a
-/// broader posture than the operator intended (e.g. a dropped `[security]`
-/// table loses hardening; a dropped `risk_profiles`/`peer_groups` loses
-/// authorization scoping). Salvage still drops them so the daemon can boot —
-/// but it logs at ERROR (not WARN) and reports them in
-/// [`ResilientLoad::dropped_security`] so the caller can refuse network
-/// exposure until the operator repairs the file.
+/// Top-level keys whose silent loss could *weaken* security posture: dropping
+/// a malformed one to its `Default` may grant a broader posture than intended.
+/// Salvage still drops them (so the daemon boots) but logs ERROR and reports
+/// them in [`ResilientLoad::dropped_security`] for exposure gating.
 pub const SECURITY_CRITICAL_KEYS: &[&str] = &["security", "risk_profiles", "peer_groups"];
 
 /// Result of a resilient (never-failing) config load.
 #[derive(Debug, Clone, Default)]
 pub struct ResilientLoad {
-    /// The loaded config — always present, carrying every section that
-    /// parsed and `Default` for every section that had to be dropped.
+    /// Loaded config: every section that parsed, `Default` for any dropped.
     pub config: Config,
-    /// Dotted paths of non-security sections/aliases dropped during salvage
-    /// (logged at WARN). Empty on a clean load.
+    /// Non-security paths dropped during salvage (logged WARN).
     pub dropped: Vec<String>,
-    /// Top-level [`SECURITY_CRITICAL_KEYS`] sections that had to be dropped
-    /// to their `Default` (logged at ERROR). Non-empty means the running
-    /// posture may be weaker than intended; the caller should refuse
-    /// exposure until repaired.
+    /// [`SECURITY_CRITICAL_KEYS`] sections dropped to `Default` (logged ERROR).
+    /// Non-empty means the running posture may be weaker than intended.
     pub dropped_security: Vec<String>,
 }
 
-/// Daemon load path with a full salvage report. Degrades instead of failing:
-///
-/// 1. Strict deserialize (identical to [`migrate_to_current`]) — the common
-///    case, zero behavior change for valid configs.
-/// 2. If that fails, drop each invalid `[channels.<type>.<alias>]`, each
-///    invalid `[channels.<type>]` whole-type block (e.g. a scalar where a
-///    table is required), and each invalid top-level section — substituting
-///    the section's `Default`. Non-security drops log at WARN; security-
-///    critical drops ([`SECURITY_CRITICAL_KEYS`]) log at ERROR and surface in
-///    [`ResilientLoad::dropped_security`].
-/// 3. If even the migration/parse stage fails (not valid TOML, or an
-///    unsupported future schema), fall back to `Config::default()` so the
-///    process still boots. The original error is logged at ERROR.
-///
-/// `Config::validate()` is intentionally NOT run here — a config that fails
-/// validation must still load so it can be fixed.
+/// Daemon load path with a salvage report. Degrades instead of failing:
+/// strict deserialize first; else drop each invalid channel alias, channel
+/// type, and top-level section (substituting `Default`); else fall back to
+/// `Config::default()`. Security-critical drops log ERROR and surface in
+/// `dropped_security`. `Config::validate()` is intentionally not run.
 pub fn migrate_to_current_salvaged(input: &str) -> ResilientLoad {
     let value = match migrate_value(input) {
         Ok(value) => value,
@@ -361,8 +331,7 @@ pub fn migrate_to_current_salvaged(input: &str) -> ResilientLoad {
 }
 
 /// Parse + migrate to the current schema version as a `toml::Value`, without
-/// the final typed deserialize. Shared by the strict and resilient entry
-/// points so the version-detection and chain logic lives in one place.
+/// the final typed deserialize. Shared by the strict and resilient entries.
 fn migrate_value(input: &str) -> Result<toml::Value> {
     let value: toml::Value = toml::from_str(input).context("failed to parse config TOML")?;
     let from = detect_version(&value)?;
@@ -388,12 +357,8 @@ fn migrate_value(input: &str) -> Result<toml::Value> {
 }
 
 /// Deserialize a migrated `toml::Value` into `Config`, never failing.
-///
-/// Strict first; on failure, salvage broken `[channels.<type>.<alias>]`
-/// entries and broken `[channels.<type>]` whole-type blocks, then drop any
-/// remaining top-level section that blocks the load (substituting its
-/// `Default`). The returned `Config` always carries every section that
-/// parsed, so the operator only loses the specific broken blocks.
+/// Strict first; on failure prune broken channel aliases, channel types, then
+/// top-level sections (each → `Default`), so only the broken blocks are lost.
 fn deserialize_resilient(value: toml::Value) -> ResilientLoad {
     if let Ok(config) = value.clone().try_into::<Config>() {
         return ResilientLoad {
@@ -466,20 +431,10 @@ fn deserialize_resilient(value: toml::Value) -> ResilientLoad {
     }
 }
 
-/// Drop any top-level `[section]` that blocks the config from deserializing,
-/// substituting the section's `Default` via serde. A section is dropped when it
-/// is the offender; valid sections are never touched. Two complementary probes
-/// catch every reachable case:
-///
-/// 1. **Joint:** if removing a single key makes the whole config valid, that
-///    key was the sole offender — drop it and finish.
-/// 2. **Isolated:** otherwise drop every key that fails to deserialize *in
-///    isolation* (its own section, wrapped alone, will not parse). This catches
-///    multiple independent offenders in one pass — e.g. two unrelated malformed
-///    sections — which the joint probe alone cannot, since neither one's
-///    removal validates the config while the other remains broken.
-///
-/// Removed paths are appended to `dropped`.
+/// Drop top-level `[section]`s that block deserialization (each → `Default`).
+/// Two probes: drop a single key if its removal validates the whole config;
+/// else drop every key that fails to deserialize in isolation (catches
+/// multiple independent offenders the joint probe can't). Appends to `dropped`.
 fn prune_bad_top_level_sections(value: &mut toml::Value, dropped: &mut Vec<String>) {
     if value.as_table().is_none() {
         return;
@@ -521,21 +476,15 @@ fn prune_bad_top_level_sections(value: &mut toml::Value, dropped: &mut Vec<Strin
     }
 }
 
-/// True when a single top-level `[section]` cannot deserialize into `Config`
-/// when it is the only key present. Wraps `{ <key> = <section> }` and runs it
-/// through the strict deserializer, isolating the failure to this one section
-/// so independent offenders are each judged on their own.
+/// True when top-level `[<key>]`, wrapped alone, fails to deserialize.
 fn top_level_section_is_invalid(key: &str, section: &toml::Value) -> bool {
     let mut root = toml::value::Table::new();
     root.insert(key.to_string(), section.clone());
     toml::Value::Table(root).try_into::<Config>().is_err()
 }
 
-/// Drop every `[channels.<type>.<alias>]` entry that cannot deserialize into
-/// its concrete channel config type. Each alias is checked in isolation via
-/// [`channel_alias_is_invalid`]; valid aliases sharing a `[channels]` section
-/// with a broken sibling are preserved. Dropped paths
-/// (`channels.<type>.<alias>`) are appended to `dropped`.
+/// Drop each `[channels.<type>.<alias>]` that fails to deserialize, checked in
+/// isolation so valid siblings survive. Appends `channels.<type>.<alias>`.
 fn prune_bad_channel_aliases(value: &mut toml::Value, dropped: &mut Vec<String>) {
     let Some(channels) = value
         .as_table_mut()
@@ -561,12 +510,9 @@ fn prune_bad_channel_aliases(value: &mut toml::Value, dropped: &mut Vec<String>)
     }
 }
 
-/// Drop every `[channels.<type>]` whole-type block that still blocks the load
-/// after alias-level pruning — e.g. a scalar written where an alias table is
-/// required (`email = "oops"`), or a type whose remaining shape is invalid.
-/// Critically this drops ONLY the offending channel type, never the entire
-/// `[channels]` section, so valid sibling channel types survive. Dropped paths
-/// (`channels.<type>`) are appended to `dropped`.
+/// Drop each `[channels.<type>]` block still blocking the load after alias
+/// pruning (e.g. a scalar where a table is required). Drops only the offending
+/// type, never the whole `[channels]` section. Appends `channels.<type>`.
 fn prune_bad_channel_types(value: &mut toml::Value, dropped: &mut Vec<String>) {
     let Some(channel_types) = value
         .as_table()
@@ -602,8 +548,7 @@ fn prune_bad_channel_types(value: &mut toml::Value, dropped: &mut Vec<String>) {
     }
 }
 
-/// True when the `[channels]` section of `value` deserializes cleanly,
-/// isolating channel-section validity from defects elsewhere in the config.
+/// True when `value`'s `[channels]` section deserializes cleanly in isolation.
 fn channels_section_is_valid(value: &toml::Value) -> bool {
     let Some(channels) = value
         .as_table()
@@ -617,11 +562,7 @@ fn channels_section_is_valid(value: &toml::Value) -> bool {
     toml::Value::Table(root).try_into::<Config>().is_ok()
 }
 
-/// True when a single `[channels.<type>.<alias>]` table cannot deserialize
-/// into its concrete channel config type. Wraps the alias under a minimal
-/// `{ channels = { <type> = { probe = <alias> } } }` document and runs it
-/// through the strict `Config` deserializer, isolating the failure to this
-/// one alias.
+/// True when `[channels.<type>.<alias>]`, wrapped alone, fails to deserialize.
 fn channel_alias_is_invalid(chan_type: &str, alias_value: &toml::Value) -> bool {
     let mut inner = toml::value::Table::new();
     inner.insert("probe".to_string(), alias_value.clone());
--- a/crates/zeroclaw-config/src/migration.rs
+++ b/crates/zeroclaw-config/src/migration.rs
@@ -419,7 +419,9 @@ fn deserialize_resilient(value: toml::Value) -> ResilientLoad {
             &format!(
                 "SECURITY-CRITICAL config section `{path}` is invalid and was reset to \
                  its default so the daemon can boot; the running posture may be WEAKER \
-                 than intended — repair `{path}` and reload before trusting this instance"
+                 than intended — repair `{path}` and reload before trusting this instance. \
+                 Run `zeroclaw config migrate` to see the precise parse error, or fix it \
+                 via the gateway config editor at `/api/config`"
             )
         );
     }
@@ -446,6 +448,8 @@ fn prune_bad_top_level_sections(value: &mut toml::Value, dropped: &mut Vec<Strin
     let keys: Vec<String> = value
         .as_table()
         .expect("root is a table")
+        // toml::Value tables preserve insertion order, so drops are reported
+        // in TOML declaration order — predictable for operators reading logs.
         .keys()
         .cloned()
         .collect();
@@ -1157,6 +1161,34 @@ enabled = "also-not-a-bool"
             load.dropped.iter().any(|p| p == "backup"),
             "second offender must be dropped, got {:?}",
             load.dropped
+        );
+    }
+
+    #[test]
+    fn multiple_bad_sections_one_security_critical() {
+        let raw = r#"
+schema_version = 3
+
+[security]
+audit = "should-be-a-table-not-a-string"
+
+[heartbeat]
+enabled = "not-a-bool"
+"#;
+        let load = migrate_to_current_salvaged(raw);
+        assert!(
+            load.dropped_security.iter().any(|p| p == "security"),
+            "malformed [security] must be classified security-critical, got {:?}",
+            load.dropped_security
+        );
+        assert!(
+            load.dropped.iter().any(|p| p == "heartbeat"),
+            "malformed [heartbeat] must be a plain drop, got {:?}",
+            load.dropped
+        );
+        assert!(
+            !load.dropped.iter().any(|p| p == "security"),
+            "security drop must not also appear in the plain dropped list"
         );
     }
 
--- a/crates/zeroclaw-config/src/migration.rs
+++ b/crates/zeroclaw-config/src/migration.rs
@@ -288,20 +288,57 @@ pub fn migrate_to_current(input: &str) -> Result<Config> {
 ///
 /// The daemon must start no matter what so the config can be repaired through
 /// the gateway, CLI, or dashboard. This function therefore degrades instead of
-/// failing:
+/// failing; see [`migrate_to_current_salvaged`] for the salvage contract and
+/// the security-section caveat. This thin wrapper discards the salvage report
+/// for callers that only need the `Config`.
+pub fn migrate_to_current_resilient(input: &str) -> Config {
+    migrate_to_current_salvaged(input).config
+}
+
+/// Top-level config keys whose silent loss could *weaken* the running
+/// security posture. "Secure by default" cuts both ways: if one of these
+/// sections is malformed, reverting it to its serde `Default` may grant a
+/// broader posture than the operator intended (e.g. a dropped `[security]`
+/// table loses hardening; a dropped `risk_profiles`/`peer_groups` loses
+/// authorization scoping). Salvage still drops them so the daemon can boot —
+/// but it logs at ERROR (not WARN) and reports them in
+/// [`ResilientLoad::dropped_security`] so the caller can refuse network
+/// exposure until the operator repairs the file.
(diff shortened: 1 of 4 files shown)


WareWolf-MoonWall (changes requested):
## Review Summary

I've completed a comprehensive review of PR #7160, examining the full diff, test coverage, security implications, and alignment with FND-005 and FND-006. This is a well-executed fix that addresses a real operational gap — the daemon previously aborted on malformed config before operators could reach repair surfaces. The PR author has clearly thought through the security implications and implemented thoughtful guardrails.

**Verdict:** Requesting changes on one blocking issue (the silent security-critical config drop in `Config::load_or_init`), plus several high-value non-blocking improvements.

**What deserves recognition before the findings:** The author correctly identified that this is a two-path problem (strict for tooling, resilient for daemon), implemented isolated probe logic that salvages multiple independent defects (fixing the "entire config → default" bug), added comprehensive test coverage for every new path, and designed a machine-readable signal (`degraded_security`) for future exposure-gating code. The validation evidence is exemplary — full workspace tests plus targeted crate-level tests with explicit counts.

---

### 🔴 Blocking — Security-critical config drop is silent in production load path

**Location:** `crates/zeroclaw-config/src/schema.rs` lines 14432–14441

**Issue:** `Config::load_or_init()` now calls `migrate_to_current_salvaged()` and assigns `config.degraded_security = salvage.dropped_security`, but the daemon continues without any runtime check or gate. A malformed `[security]` / `[risk_profiles]` / `[peer_groups]` section will be silently reset to `Default` values, and the daemon will start with a potentially weaker posture than the operator intended.

The PR description states:

> "exposure-gating code can refuse to trust a degraded instance until the operator repairs the file"

But no such gating code exists in this PR. The field is populated, logged at ERROR, but never checked before the daemon proceeds to full operation. Per FND-006 §4.5 ("Fail loudly near security boundaries"):

> "An error in a security check […] should never be silently swallowed. It should be logged, propagated, and handled explicitly."

**What needs to change:**

1. **Minimum gate:** `Config::load_or_init()` must refuse to proceed if `salvage.dropped_security` is non-empty. Return an error with a clear message: "Config contains malformed security-critical sections (listed: {sections}); repair the file and restart."

2. **OR — escape hatch with loud warning:** If the team decides the daemon must boot even with degraded security (to let operators reach the gateway repair UI), emit a repeating WARNING at startup (e.g. every 10 seconds for the first minute, then hourly) and consider refusing to start any channel or tool until the operator acknowledges the degraded state via a CLI flag like `zeroclaw config ack-degraded` or gateway action.

**Why this matters:** The current implementation logs ERROR once at boot, then runs normally. An operator who isn't watching logs at startup will have no runtime signal that their security policy is not in effect. This violates the principle that security failures should be explicit and loud.

**Suggested fix (option 1 — strict gate):**

```rust
// In Config::load_or_init(), after line 14441:
if !config.degraded_security.is_empty() {
    anyhow::bail!(
        "Config contains malformed security-critical sections: {}. \
         The daemon cannot start with a weakened security posture. \
         Repair these sections in {} and restart.",
        config.degraded_security.join(", "),
        config_path.display()
    );
}
```

This makes the daemon refuse to start until the operator fixes the config, which is the safest default.

---

### 🟡 Warning — Test coverage for `degraded_security` assignment is incomplete

**Location:** `crates/zeroclaw-config/src/schema.rs` lines 14432–14441

**Issue:** The PR adds `config.degraded_security = salvage.dropped_security;` in `Config::load_or_init()`, but there is no integration test verifying that this assignment occurs when a daemon load encounters a malformed security-critical section.

The unit tests in `migration.rs` verify that `migrate_to_current_salvaged()` populates `dropped_security` correctly. But no test verifies that `Config::load_or_init()` actually assigns that field to the returned `Config` instance. If a future refactor accidentally removes that line, no test would catch it.

**What would improve this:**

Add an integration test in `crates/zeroclaw-config/src/schema.rs` (or `tests/component/config_persistence.rs`) that:

1. Writes a config with a malformed `[security]` section to disk.
2. Calls `Config::load_or_init()`.
3. Asserts that the returned `config.degraded_security` is non-empty and contains `"security"`.

This closes the gap between unit-level salvage and the public API that consumers actually call.

---

### 🟡 Warning — Missing doc comment on the new public API surface

**Location:** `crates/zeroclaw-config/src/migration.rs` lines 285–287, 293–296

**Issue:** `migrate_to_current_resilient()` and `ResilientLoad` are new public APIs (the daemon load path), but they carry minimal doc comments. A future contributor reading the signature will not immediately understand:

- When to use `migrate_to_current_resilient` vs `migrate_to_current_salvaged` vs `migrate_to_current`
- What the return type `ResilientLoad` represents and why it has three fields
- Why this path exists alongside the strict path

Per FND-006 §4.2 ("Public API Surface as a Promise") and FND-002 (Documentation Standards), public functions should document their contract, especially when the choice between two similar-looking functions has security implications.

**What would improve this:**

Expand the doc comment on `migrate_to_current_resilient` to:

```rust
/// Daemon load path: versioned TOML → usable `Config`, never failing.
///
/// This is the entry point for daemon startup (`Config::load_or_init`). It
/// degrades gracefully rather than failing: strict deserialize first; on
/// failure, drop invalid channel aliases, channel types, and top-level
/// sections (substituting `Default`). Security-critical sections (`security`,
/// `risk_profiles`, `peer_groups`) that are dropped are flagged in the
/// returned `degraded_security` field.
///
/// **Never use this path for repair tooling** — `zeroclaw config migrate`,
/// model routing config editors, and the gateway config editor should call
/// [`migrate_to_current`] to get precise error messages for the user.
///
/// Thin wrapper over [`migrate_to_current_salvaged`] that drops the report.
pub fn migrate_to_current_resilient(input: &str) -> Config {
```

And on `ResilientLoad`:

```rust
/// Result of a resilient (never-failing) config load.
///
/// Used by the daemon load path to distinguish between plain config defects
/// (logged as WARN, safe to drop) and security-critical defects (logged as
/// ERROR, potentially requiring a startup gate). See [`SECURITY_CRITICAL_KEYS`].
#[derive(Debug, Clone, Default)]
pub struct ResilientLoad {
```

This makes the contract explicit at the call site and reduces the chance a future contributor uses the wrong path.

---

### 🟡 Warning — Salvage logic does not preserve section order

**Location:** `crates/zeroclaw-config/src/migration.rs` line 434 (`prune_bad_top_level_sections`)

**Issue:** The two-probe salvage logic (joint probe, then isolated probe) iterates over `keys()` from a `BTreeMap` or `HashMap`. When multiple independent sections are malformed, the order in which they are dropped is iteration order, not declaration order in the TOML file.

This is not a correctness issue (the config loads successfully either way), but it makes the log messages slightly less predictable for an operator who is debugging a multi-section failure. Logging "dropped `[heartbeat]`, `[backup]`" when the operator sees `[backup]` listed first in the TOML file is a minor UX inconsistency.

**What would improve this (non-blocking):**

Document this behavior in a code comment:

```rust
// Iteration order is HashMap key order (undefined), not TOML declaration
// order. This does not affect correctness but may make multi-drop logs
// less predictable for operators debugging their config.
```

Or, if preserving order matters for debuggability, switch to `toml::map::Map` (which preserves insertion order). This is a polish decision, not a blocker.

---

### 🔵 Suggestion — Consider a CLI flag for acknowledging degraded security state

**Location:** Conceptual (no specific line)

**Context:** If the team decides to allow the daemon to boot with `degraded_security` non-empty (instead of hard-failing per the 🔴 blocking finding), consider adding a CLI flag like:

```bash
zeroclaw daemon --allow-degraded-security
```

This makes the operator's intent explicit: "I know the config is malformed, I want the daemon to boot anyway so I can fix it via the gateway." Without the flag, the daemon refuses to start, making the default behavior safe.

This pattern is common in production systems that need to balance safety (don't run with broken security config) with operability (don't lock the operator out of their repair tools).

**Alternative:** A gateway `/api/config/ack-degraded` POST endpoint that the operator must call before the daemon will start channels or run agents. This keeps the failure loud while still allowing access to repair surfaces.

---

### 🔵 Suggestion — Log a sample repair path when security-critical sections are dropped

**Location:** `crates/zeroclaw-config/src/migration.rs` lines 420–432

**Issue:** The ERROR log message for security-critical drops is clear about what happened and why it's bad:

> "SECURITY-CRITICAL config section `{path}` is invalid and was reset to its default […]"

But it doesn't tell the operator what to do next. For a new operator who isn't familiar with the config migration tooling, the next step is not obvious.

**What would improve this:**

Add a hint to the log message:

```rust
&format!(
    "SECURITY-CRITICAL config section `{path}` is invalid and was reset to \
     its default so the daemon can boot; the running posture may be WEAKER \
     than intended — repair `{path}` in the config file and reload before \
     trusting this instance. Run `zeroclaw config migrate` to see the precise \
     error, or use the gateway config editor at http://localhost:8080/config"
)
```

This gives the operator two concrete paths to resolution, reducing the time to fix.

---

### 🔵 Suggestion — Test the "multiple independent sections salvaged" path with a security-critical section

**Location:** `crates/zeroclaw-config/src/migration.rs` line 1136 (`multiple_independent_bad_sections_all_dropped`)

**Issue:** The test `multiple_independent_bad_sections_all_dropped` verifies that two malformed non-security sections (`[heartbeat]`, `[backup]`) are both dropped. But there is no test covering the case where one of the two is security-critical.

**What would improve test coverage:**

Add a test like:

```rust
#[test]
fn multiple_bad_sections_one_security_critical() {
    let raw = r#"
schema_version = 3

[security]
audit = "should-be-a-table"

[heartbeat]
enabled = "not-a-bool"
"#;
    let load = migrate_to_current_salvaged(raw);
    assert!(
        load.dropped_security.iter().any(|p| p == "security"),
        "malformed [security] must be in dropped_security"
    );
    assert!(
        load.dropped.iter().any(|p| p == "heartbeat"),
        "malformed [heartbeat] must be in dropped (not dropped_security)"
    );
}
```

This exercises the classification path when both kinds of failures occur in one load.

---

### 🔵 Suggestion — Clarify the "no validation on resilient path" decision in the code

**Location:** `crates/zeroclaw-config/src/migration.rs` (conceptual)

**Context:** The PR description states:

> "does NOT run `Config::validate()` on the resilient path (a config that fails validation must still load so it can be fixed)"

This is a deliberate design choice, and it's the right one (validation failures should not prevent daemon boot). But there is no code comment or doc explaining why `validate()` is intentionally not called.

**What would improve clarity:**

Add a comment in `deserialize_resilient` (or near the call site in `Config::load_or_init()`) explaining:

```rust
// Intentionally does NOT call `Config::validate()` on the resilient path:
// validation failures (e.g. a required field missing) must not prevent the
// daemon from booting, because the operator needs the process up to repair it.
// Validation is enforced by the strict path (`migrate_to_current`) used by
// repair tooling and explicit config-check commands.
```

This prevents a future contributor from "fixing" the perceived gap by adding validation here.

---

### 🟢 What looks good — Isolated probe logic correctly salvages multiple independent defects

**Why this works:** The original salvage logic (pre-PR) had a subtle but serious defect: the joint probe only tested whether removing a single section validated the whole config. If two independent sections were both malformed, the joint probe found nothing, salvage gave up, and the entire config silently reset to `Config::default()`.

The reworked logic in this PR uses a two-phase approach:

1. **Joint probe:** Try removing each section one at a time. If that validates the config, we found the sole offender — drop it and return early.
2. **Isolated probe:** If the joint probe found nothing (multiple offenders), test each section in isolation by wrapping it alone in a synthetic config. Drop every section that fails this test.

This correctly handles the "two independent bad sections" case, as verified by the test `multiple_independent_bad_sections_all_dropped`. The logic is sound, the tests cover it, and the fix addresses a real data-loss risk (losing valid sections because one unrelated section was malformed).

**Impact:** Operators with complex configs containing multiple independent defects will now see only the broken sections dropped, with valid siblings preserved. This is a significant improvement in salvage quality.

---

### 🟢 What looks good — Test coverage is comprehensive and targeted

**Why this works:** The author added eight new tests covering:

- Broken channel alias dropped (sibling kept)
- Broken whole-type channel block dropped (sibling type kept)
- Malformed `[security]` reported as security-critical (not plain drop)
- Malformed non-security section reported as plain drop (not security-critical)
- Two independent malformed sections both salvaged
- Unparseable TOML falls back to defaults
- Future schema version falls back to defaults
- Strict path still errors for repair tooling

This is exactly the right level of coverage for a resilient-load path: every branch is exercised, edge cases are explicit, and the strict-vs-resilient contract is validated. The test names are clear about what they're verifying, making future regressions easy to catch.

**Plus:** The validation evidence in the PR body documents not just that tests passed, but specific counts (`726 passed` lib, `89 passed` integration). This level of detail makes it easy to verify the change didn't accidentally break unrelated tests.

---

### 🟢 What looks good — Security-critical sections are explicitly named and justified

**Why this works:** The PR introduces `SECURITY_CRITICAL_KEYS: &[&str] = &["security", "risk_profiles", "peer_groups"]` with a clear doc comment explaining why these three deserve ERROR logging and a separate drop list:

> "Top-level keys whose silent loss could *weaken* security posture: dropping a malformed one to its `Default` may grant a broader posture than intended."

This is exactly the right boundary. These three sections directly control authorization, tool approval, and risk gates. Dropping them to `Default` without loud signaling would be a silent security regression.

The author correctly identified that other sections (`heartbeat`, `backup`, etc.) can safely degrade to defaults without weakening security, so they get WARN logging and live in the plain `dropped` list. This distinction is architecturally sound and well-justified.

---

### 🟢 What looks good — The strict path is unchanged for repair tooling

**Why this works:** The PR does not alter the behavior of `migrate_to_current()`, which is used by:

- `zeroclaw config migrate`
- Model routing config editors
- Gateway config editor (for validation before save)
- All existing unit tests that verify migration logic

These tools need precise error messages so they can tell the operator exactly what's wrong and where. The strict path continues to return `Result<Config>` with detailed errors, as it should.

The resilient path is a *new, parallel* path used only by `Config::load_or_init()`. This keeps the blast radius narrow and ensures that repair tooling behavior is unchanged. A solid architectural choice.

---

## Final Comments

This is high-quality work. The author has clearly thought through the operational gap (daemon can't boot to reach repair surfaces), the security implications (security-critical sections need louder treatment), and the salvage quality (multiple independent defects should not nuke the whole config). The test coverage is exemplary.

**The one blocking issue:** The `degraded_security` field is populated and logged, but never checked before the daemon proceeds. This violates the loud-failure principle from FND-006 §4.5. Resolve this with either a hard gate (daemon refuses to start) or an escape hatch (explicit operator acknowledgment via CLI flag or gateway action).

**Re-review readiness:** Once the blocking issue is addressed (plus any non-blocking changes the author chooses to adopt), I'm ready to approve. The rest of the findings are polish and future-proofing — valuable, but not merge-blockers.

**Commits:** Three commits, all clearly scoped. The commit messages are descriptive and follow conventional commit style. No rebase or squash needed unless the author prefers to consolidate before merge.

---

Reviewed head commit: c3529d5  
Reviewing as: @WareWolf-MoonWall  
Protocol: FND-005, FND-006, PR review protocol  
Risk tier: High (per AGENTS.md — security boundaries, config loading)

WareWolf-MoonWall (approved):
I re-reviewed #7160 at head 16d25e8b (June 3 post-fix commits d198d1ce + 16d25e8b), following my June 3 19:35 `CHANGES_REQUESTED` review at 1f2dfbe3. The author addressed the blocking finding with two additional commits that gate every serving entry point on the security posture.

### ✅ Resolved — Security-critical config drop is now gated at every serving entry point

My June 3 blocking finding was that `degraded_security` was populated but never checked before the daemon proceeded to full operation. The two new commits add `gate_security_posture()`, which:

1. **Fails closed by default.** When `config.degraded_security` is non-empty and `--allow-degraded-security` is false, the function returns `Err` with an actionable message: "Config contains malformed security-critical sections ({sections}); they were reset to defaults, so the running posture may be weaker than intended. Refusing to serve with a degraded security posture. Repair these sections in {config_path} and restart..."

2. **Gates every serving entry point.** The gate sits in front of:
   - `zeroclaw daemon`
   - `zeroclaw gateway start`
   - `zeroclaw gateway restart`
   - Bare `zeroclaw gateway` (no subcommand)

   All four paths call `gate_security_posture(&config, allow_degraded_security)?` before binding or starting subsystems.

3. **Spawns a nag task when explicitly allowed.** When `--allow-degraded-security` is set, the daemon boots but returns a `JoinHandle` for a repeating-WARN task that logs every 30 seconds: "Running with DEGRADED security: sections ({sections}) were reset to defaults and `--allow-degraded-security` was set. The posture may be weaker than intended — repair {config_path} and reload..."

4. **Re-evaluates on SIGUSR1 reload.** The daemon loop re-calls `gate_security_posture()` after each config reload, aborting the stale nag task and re-gating against the fresh config. A repaired posture silences the warning; a newly-degraded posture without the flag hard-fails the reload.

5. **Has regression test coverage.** `gate_security_posture_fails_closed_unless_allowed` proves the clean-posture, degraded-not-allowed, and degraded-allowed paths all behave correctly.

This is exactly what FND-006 §4.5 ("Fail loudly near security boundaries") requires. The implementation is thorough, the escape hatch is explicit (opt-in CLI flag), and the nag ensures operators cannot ignore a degraded posture.

### 🟢 What looks good — Nag task lifecycle is managed correctly

The nag task is scoped to the current config and lives in a `let mut degraded_nag: Option<JoinHandle<()>>` binding outside the daemon loop. On each reload iteration:
- The stale nag (if any) is aborted via `handle.abort()`
- The fresh config is loaded
- `gate_security_posture()` is re-called with the fresh config, returning a new nag handle (or `None` if the posture is now clean)

This prevents nag-task leaks and ensures the warning state stays synchronized with the live config. The final `degraded_nag.take()` cleanup when the loop exits is also correct.

### 🟢 What looks good — CLI flag is consistently threaded

The `--allow-degraded-security` flag is added to all three subcommands (`daemon`, `gateway start`, `gateway restart`) and correctly passed through to `gate_security_posture()`. Bare `zeroclaw gateway` (no subcommand) hard-codes `false` ("fail closed"), which is the right default for an unattended invocation.

### 🟢 What looks good — Test coverage closes the gap

The new `gate_security_posture_fails_closed_unless_allowed` test verifies:
- Clean posture: no gate, no nag
- Degraded + not allowed: `Err`
- Degraded + allowed: `Ok(Some(handle))`

This locks in the security contract and would catch a future refactor that accidentally removed the gate or made it permissive by default.

### ✅ Approved — all findings addressed

The blocking finding is resolved. The security-critical config drop is no longer silent in production: the daemon refuses to serve unless the operator explicitly opts in, and when allowed, it emits a repeating loud warning until the config is repaired. The gate is consistently applied at every serving entry point, the nag lifecycle is managed correctly, and the behavior is locked in by regression tests.

The PR body's validation evidence is comprehensive (728+89 tests pass, gate test added, manual salvage verification), the multi-probe salvage logic is correct, and the test coverage for every new path is thorough. Milestone v0.8.0 is appropriate (config-load hardening in the config-stable milestone).

Ready to merge.

Audacity88 (dismissed):
I reviewed #7160 at head 16d25e8b, including @WareWolf-MoonWall's earlier changes-requested review and later approval, the current PR body, current GitHub checks, and the touched config/gateway startup paths in `crates/zeroclaw-config/src/migration.rs`, `crates/zeroclaw-config/src/schema.rs`, `src/lib.rs`, and `src/main.rs`. I did not rerun cargo locally; this pass focused on the security posture gate and the resilient-load edge cases.

### ✅ Resolved — malformed security sections are now gated before serving

The startup gate shape is right for the ordinary salvage case. A malformed `[security]`, `risk_profiles`, or `peer_groups` section is now recorded on `Config.degraded_security`, `daemon`, `gateway start`, `gateway restart`, and bare `zeroclaw gateway` all route through `gate_security_posture()`, and the default behavior is fail-closed unless the operator explicitly opts into `--allow-degraded-security`. The reload path also aborts the old nag task and re-gates the fresh config, which keeps the warning state tied to the live posture instead of leaking stale warnings after repair.

### 🔴 Blocking — whole-config fallback bypasses the security gate

`migrate_to_current_salvaged()` still has whole-config fallback paths that can produce `Config::default()` without an explicit degraded-security marker. The parse/migration failure path before `deserialize_resilient()` (`toml::from_str`, unsupported future schema, migration-chain failure) returns `ResilientLoad { config: Config::default(), dropped: vec![], dropped_security: vec![] }`.

The final unsalvageable-deserialize fallback inside `deserialize_resilient()` also defaults the whole config, and it only reports security degradation if earlier pruning happened to collect security-critical drops. `Config::load_or_init()` copies only `salvage.dropped_security` into `config.degraded_security`, so these whole-config-loss cases can look clean to `gate_security_posture()`.

That means a syntactically corrupt config, a future-schema config, or another unsalvageable root can still boot `daemon` / `gateway start` on `Config::default()` without requiring `--allow-degraded-security`. This recreates the security posture problem at the coarsest failure boundary: every security-critical section may have been lost, but the serving gate has no signal and allows the process to serve normally.

Please treat whole-config fallback as a degraded security posture too. One simple shape would be to return `ResilientLoad { config: Config::default(), dropped: vec![], dropped_security: vec!["<config>".into()] }` or a similarly explicit sentinel from the parse/migration fallback, and make the final unsalvageable-deserialize fallback add the same marker when it has to default the whole config. Then add coverage showing:

- invalid TOML returns a non-empty `dropped_security`;
- unsupported future schema returns a non-empty `dropped_security`;
- `Config::load_or_init()` propagates that marker; and
- `gate_security_posture()` refuses to serve unless `--allow-degraded-security` is set.

The operator-repair goal still works with that shape: the daemon can boot with the explicit override, but it cannot silently serve a defaulted security posture after losing the whole config.

### 🟢 What looks good — the scoped salvage path is carefully covered

The section-level salvage work is solid. Broken channel aliases, broken channel-type blocks, malformed non-security sections, mixed security/non-security drops, and multiple independent bad sections all have focused coverage, and the strict `migrate_to_current()` path remains intact for repair tooling. Once the whole-config fallback is fed into the same degraded-security signal, this will be a much safer and more operator-friendly load path.

singlerider:
Addressed the blocker on `fa09cfb1f`:

- **🔴 Whole-config fallback bypassed the security gate** — the section-level salvage path recorded security-critical drops, but the whole-config fallbacks (`migrate_to_current_salvaged`'s parse/migration-failure path and `deserialize_resilient`'s final unsalvageable-deserialize path) defaulted the entire config while returning an empty `dropped_security`. A corrupt config, unsupported future schema, broken migration chain, or unsalvageable root could boot on `Config::default()` — losing every security-critical section — with no signal to `gate_security_posture()`.

  Added a `WHOLE_CONFIG_SENTINEL` marker (`<entire-config>`) pushed onto `dropped_security` from both whole-config fallbacks. `Config::load_or_init()` propagates it, so the serving gate now fails closed on whole-config loss exactly as it does for a single dropped section; the operator can still boot with `--allow-degraded-security`.

  Coverage added per your list:
  - invalid TOML returns a non-empty `dropped_security`;
  - unsupported future schema returns a non-empty `dropped_security`;
  - unsalvageable root returns a non-empty `dropped_security`;
  - `load_or_init` propagates the marker (`load_or_init_marks_whole_config_degraded_for_unparseable_file`);
  - `gate_security_posture()` refuses to serve on the sentinel unless `--allow-degraded-security` is set.

Validation: `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-config --all-targets -- -D warnings`, `cargo clippy --bin zeroclaw -- -D warnings`, and the targeted migration/schema/gate tests all green.

Re-requesting review.

PR #7162, opened 2026-06-03
feat(channels): notify before context compression

sbenedetto:
## Summary

- **What changed and why:**
  - Adds an opt-in generic channel-visible notice when a reply-eligible channel turn is expected to condense/trim earlier conversation to stay within the context budget.
  - Keeps the notice plain-language and avoids exposing token counts, budgets, threshold ratios, or other internal mechanics to end users.
  - Adds `notify_on_compression` under context compression config so operators can enable the notice explicitly.
- **Scope boundary:** This does not change trim/compression behavior, summarization prompts, provider calls, memory persistence, streaming, or any channel-specific implementation.
- **Blast radius:** Agent replies on configured channels may include one additional progress notice only when proactive history maintenance is needed, notification is enabled, and the turn is reply-eligible.
- **Linked issue(s):** None.
- **Labels:** Current labels include `enhancement`, `risk: medium`, `size: XS`, `channel`, `config`.

```bash
cargo fmt --all -- --check
```

```text
# exited 0 with no output
```

```bash
git diff --check
```

```text
# exited 0 with no output
```

```bash
CARGO_TARGET_DIR=/home/salvatore-zeroclaw/target-pr7162 \
  cargo test -p zeroclaw-channels --features channel-slack compression_notice -- --nocapture
```

```text
running 2 tests
test orchestrator::tests::process_channel_message_does_not_send_compression_notice_for_no_reply_turn ... ok
test orchestrator::tests::process_channel_message_sends_compression_notice_for_reply_turn ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 1277 filtered out; finished in 0.06s
```

```bash
CARGO_TARGET_DIR=/home/salvatore-zeroclaw/target-pr7162 \
  cargo check -p zeroclaw-config -p zeroclaw-channels --features channel-slack
```

```text
Finished `dev` profile [unoptimized] target(s) in 1m 52s
```

- **Beyond CI — what did you manually verify?** Merged latest `origin/master` into the branch as requested, resolved conflicts in `orchestrator/mod.rs` and `scattered_types.rs`, kept the notice behind the no-reply gate, and added positive and negative regression coverage for the visible notice behavior.
- **If any command was intentionally skipped, why:** Full workspace `cargo test` and full clippy were not run locally for this focused channel/config follow-up; validation used the focused reviewer-regression tests plus package-level `cargo check`.

## Compatibility (required)

- If `No` or `Yes` to either: Existing configs continue to load because the new config field has a default. Operators that do not want compression notices do nothing; `notify_on_compression` defaults to `false`.

- **Fast rollback command/path:** `git revert 9b3bbe26f 5f7e96297`
- **Feature flags or config toggles:** `notify_on_compression = false` disables the notice while keeping context maintenance enabled.
- **Observable failure symptoms:** Users may see unexpected extra channel messages before replies when history exceeds the context budget; grep logs for `Failed to send context compression notice` if notice delivery fails.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PRs or incorporated external contributor work.


Audacity88 (changes requested):
Checked live PR #7162 at head 6bfd31836196cbf08c2cce35fc1d8a2007b02503, including the PR body, current checks, comments, formal reviews, diff, and the adjacent channel reply-intent and context-compression code.

### 🟢 What looks good — The config knob is narrow and backwards-compatible

The `notify_on_compression` option is added to the existing context-compression config, defaults through both serde and `Default`, and keeps existing configs loading without migration. The send path also stays generic over `Channel` and treats delivery failure as best-effort. I do not see separate label cleanup needed here.

### 🔴 Blocking — Compression notices can bypass no-reply routing

The new notice is sent before the reply-intent precheck runs. That means a long-history channel message can emit `I'm compacting our conversation before replying...` before the classifier has decided whether the latest inbound message should get any visible reply.

That breaks the existing no-reply contract for channel traffic. Today, a `NO_REPLY` decision records the outcome, optionally reacts to the inbound message, and then returns without sending a visible message; the existing precheck coverage asserts that no visible reply should be sent for background chatter. With this PR, the same kind of turn can still send the compression notice whenever the history is over the threshold, even though the agent would otherwise ignore the message. In busy group channels, that turns background chatter into bot-visible noise and also tells users the agent is "before replying" on a turn that will not reply.

Please gate or defer the notice so it is only sent for reply-eligible turns, or add equivalent regression coverage proving the notice cannot appear for `NO_REPLY` outcomes.

crates/zeroclaw-channels/src/orchestrator/mod.rs
```diff
@@ -4121,6 +4121,28 @@ async fn process_channel_message_body(
...
+        if cc_config.enabled && cc_config.notify_on_compression && ctx.context_token_budget > 0 {
+            let tokens_before =
+                zeroclaw_runtime::agent::context_compressor::estimate_tokens(&history);
+            #[allow(clippy::cast_possible_truncation, clippy::cast_sign_loss)]
+            let threshold = (ctx.context_token_budget as f64 * cc_config.threshold_ratio) as usize;
+            if tokens_before > threshold
+                && let Some(channel) = target_channel.as_ref()
+            {
+                let notice = format!(
+                    "I'm compacting our conversation before replying. Current tokens: {tokens_before}. Max tokens: {}. Threshold ratio: {:.2}. Compression starts above {threshold} tokens.",
```
singlerider: 🔴 This notice is sent here (~line 4121-4133), but `classify_channel_reply_intent` runs later in this same function. So a NO_REPLY turn — background chatter the agent would ignore — still emits this visible message saying it's compacting 'before replying' on a turn that won't reply. Gate behind the reply-eligibility decision. Also: the text leaks token/threshold internals to end users (use plain copy), and `notify_on_compression` defaults to `true` so this ships on for everyone.

singlerider (commented):
Reviewed at head `6bfd318` against current `master`. The config knob is clean, but I verified @Audacity88's no-reply blocker in source, and the default-on plus the message wording make it worse than it first looks. Agreeing with his changes-requested.

### 🟢 The config knob is well-formed

`notify_on_compression` is added to `ContextCompressionConfig` with both a serde default and a `Default` impl, so existing configs load without migration, and the send path stays generic over `Channel` with best-effort delivery. No label cleanup needed.

### 🔴 Blocking — confirmed: the notice fires before the reply-intent gate, so `NO_REPLY` turns still send it

In `process_channel_message_body` (starts line 3742), the notice send is at line ~4121, and `classify_channel_reply_intent` runs later in the same function (after 4121). So the order is: emit "I'm compacting our conversation before replying..." → then decide whether this turn should reply at all. A long-history `NO_REPLY` turn (background group chatter the agent would otherwise ignore, only recording the outcome and optionally reacting) still emits the visible notice. That breaks the no-reply contract the existing precheck coverage protects, and it actively misinforms — it says "before replying" on a turn that will not reply. Gate the notice behind the reply-eligibility decision (send only for reply-eligible turns), and add regression coverage proving the notice cannot appear for `NO_REPLY` outcomes.

### 🔴 Blocking — `notify_on_compression` defaults to `true`, so this ships the noise to everyone

`default_notify_on_compression()` returns `true`. Combined with the gate bug above, every existing channel deployment starts emitting compression notices — including on background chatter in busy group channels — with no opt-in. Even once the `NO_REPLY` gate is fixed, defaulting this on is a behavior change for all channel users. Default it to `false` (opt-in), consistent with how a chatty user-visible side effect should ship.

### 🟡 The notice text leaks internal mechanics

`"Current tokens: {tokens_before}. Max tokens: {}. Threshold ratio: {:.2}. Compression starts above {threshold} tokens."` exposes token budgets and the threshold ratio to end users — that is debug-level detail, not user-facing copy. A plain "I'm condensing our earlier conversation to stay within context; one moment." conveys the same thing without surfacing internals.

Gate it behind reply-eligibility, flip the default to opt-in, and simplify the message. Leaving @Audacity88's changes-requested in force.

sbenedetto:
Rebased on latest `master` (`621e329c7`) and addressed the no-reply/default/message review feedback.

What changed:
- `notify_on_compression` now defaults to `false`;
- compression notice delivery is deferred until after the `NO_REPLY` gate;
- the user-facing notice is simplified and no longer exposes token counts or threshold internals;
- added a regression proving a compressed `NO_REPLY` turn sends no visible messages.

Validation run locally:
- `cargo test -p zeroclaw-channels --features channel-slack process_channel_message_does_not_send_compression_notice_for_no_reply_turn -- --nocapture`
- `cargo test -p zeroclaw-channels --features channel-slack process_channel_message_uses_classifier_provider_for_precheck_model_selection -- --nocapture`
- `cargo check -p zeroclaw-channels --features channel-slack`

GitHub CI is currently running on the updated branch.

sbenedetto:
Pushed a follow-up lint fix for the new regression test (`clippy::needless_borrows_for_generic_args`).

Additional local validation:
- `cargo test -p zeroclaw-channels --features channel-slack process_channel_message_does_not_send_compression_notice_for_no_reply_turn -- --nocapture`
- `CARGO_TARGET_DIR=/home/salvatore-zeroclaw/target-pr7162-channels-clippy cargo clippy -p zeroclaw-channels --all-targets --features channel-slack -- -D warnings`

Audacity88 (approved):
Reviewed current head `5f7e96297f173842192498cdea5852ae2a1196fe` against the live PR state, prior Audacity88/singlerider feedback, the author’s follow-up comments, current checks, and the updated `orchestrator` / `ContextCompressionConfig` diff. I did not run local Cargo validation.

### ✅ Resolved — Compression notices no longer bypass no-reply routing

The notice is now only marked pending when proactive compression actually runs, and the current diff sends it only after the `NO_REPLY` branch has had a chance to return. That preserves the channel no-reply contract: background chatter can still compress local history if needed, but it does not emit a visible “continuing now” message unless the turn is reply-eligible.

The new regression exercises the risky case directly: it forces compression, keeps `notify_on_compression = true`, drives the classifier to `NO_REPLY`, and asserts that the channel records no visible sends. That covers the blocker from my prior review.

### ✅ Resolved — The user-visible side effect is now opt-in and plain-language

`notify_on_compression` now defaults to `false` in both serde and `Default`, so existing channel deployments do not start emitting new progress messages without operator choice. The notice text is also user-facing now; it says the agent condensed earlier conversation to stay within context and no longer exposes token estimates, budget, threshold ratio, or trigger threshold.

### 🟢 What looks good — The feature stays generic and best-effort

The implementation stays at the channel trait boundary rather than adding Slack-, Telegram-, or provider-specific behavior, and failed notice delivery only logs debug context before continuing the normal turn. That keeps this as a small operator-visible progress feature instead of changing compression, summarization, model routing, or memory persistence semantics.

Approving from my side. This clears my previous changes-requested review on #7162; it is not a merge-ready statement while normal merge freshness and public-record cleanup still need to be handled.

singlerider (changes requested):
@sbenedetto — reviewed on its own merits.

🟢 The core design is sound: `notify_on_compression` is opt-in (default false) on the existing `ContextCompressionConfig`, the pending flag is set only when `result.compressed`, and the notice is emitted *after* the no-reply early-return so a NO_REPLY turn stays silent — and that negative case is explicitly tested. Send failure is logged at DEBUG rather than being fatal, and the let-chain gating reads cleanly.

🔴 Blocking: the branch is in merge conflict with master (`mergeable_state: dirty`) and can't land until it's merged up and the conflict resolved. Please merge master into the branch (do not rebase) and push.

🔴 Blocking: there's a negative-path test (no notice on a no-reply turn) but no positive-path test asserting the notice *is* sent on a compressed reply turn. Add one so the happy path is pinned and a future refactor can't silently drop the notice. Separately, the PR body says the notice "includes current token estimate, context budget, threshold ratio, and trigger threshold," but the implemented notice is a fixed generic sentence — either align the body to the simpler message (which I'd prefer; don't leak internal budget numbers to end users) or wire the values in.

singlerider:
@sbenedetto heads up: this PR now conflicts with `master`. Merge the latest `master` into the branch (we merge, not rebase) so it can land.

singlerider:
@sbenedetto follow-up: master now has the advisory-gate fix (#8520 merged), so once you merge the latest `master` into this branch the conflict clears and CI should go green. No rebase, merge only.

sbenedetto:
Merged latest `origin/master` into this branch as requested; no rebase.

Follow-up changes:
- resolved conflicts in `crates/zeroclaw-channels/src/orchestrator/mod.rs` and `crates/zeroclaw-config/src/scattered_types.rs`;
- preserved the current `master` history-trim path instead of reviving the removed `context_compressor` module;
- kept the notice behind the reply-intent/no-reply gate;
- added the requested positive-path regression asserting a reply-eligible over-budget turn sends the compression notice;
- kept the existing negative-path regression asserting `NO_REPLY` turns send no notice;
- updated the PR body so it no longer claims token/budget/threshold internals are shown to users.

Local validation:
- `git diff --check`
- `CARGO_TARGET_DIR=/home/salvatore-zeroclaw/target-pr7162 cargo test -p zeroclaw-channels --features channel-slack compression_notice -- --nocapture`
- `CARGO_TARGET_DIR=/home/salvatore-zeroclaw/target-pr7162 cargo check -p zeroclaw-config -p zeroclaw-channels --features channel-slack`

sbenedetto:
Merged current `upstream/master` into `feat/context-compression-notice` as requested and pushed the refreshed branch.

Follow-up fix:
- preserved the hook runner when the test helper delegates to the new context-budget variant, so the post-merge test build compiles.

Local validation on head `0e3de31c`:
- `git diff --check`
- `CARGO_TARGET_DIR=/home/ilitio/target-pr7162 cargo test -p zeroclaw-channels --features channel-slack compression_notice -- --nocapture`
- `CARGO_TARGET_DIR=/home/ilitio/target-pr7162-check cargo check -p zeroclaw-config -p zeroclaw-channels --features channel-slack`

GitHub CI is now rerunning on the updated head.

Audacity88:
@sbenedetto, this PR still has requested changes, and GitHub now reports conflicts with `master` again. When addressing the review feedback, could you refresh the branch and ping us once the conflicts are resolved and CI has rerun? Thanks.

IftekharUddin (commented):
@sbenedetto I reviewed current head 0e3de31c1b703db5fedc165d338ac4b11613dc93, the full channel-orchestrator diff, the earlier review history, the refreshed PR body, and current checks. I did not run local Cargo because the working tree is carrying unrelated changes; I relied on the current-head CI and source inspection.

### ✅ Resolved — Notification behavior and positive coverage now match the stated contract

The current head no longer emits the notice for `NO_REPLY`, the setting remains opt-in, the user-facing text is plain, and the focused tests now include the positive reply path. Those points from the earlier reviews are settled.

```mermaid
flowchart LR
  A["History over budget"] --> B{"Reply-intent precheck"}
  B -->|"NO_REPLY"| C["Return without notice"]
  B -->|"Reply"| D["Send compression notice"]
  D --> E["Continue normal response"]
```

### 🔴 Blocking — Refresh the branch conflict before another code verdict

The live merge state is still `CONFLICTING` / `DIRTY`, and @singlerider's active changes-requested review remains in force. I found no additional product blocker on the current patch, but I will not approve over that active block. Please resolve the conflict and let required CI validate the actual merge result; the existing reviewer can then convert or dismiss the block.

Audacity88:
@sbenedetto, this PR has been waiting since July 14 for its conflict with `master` to be resolved before review can continue, with no author response or new commit.

Marking this as a stale candidate for backlog hygiene. If you would like to continue, please refresh the branch and rerun CI by July 31, 2026. A brief reply with your plan is also enough to keep it open. If there is no response by then, maintainers may close the PR; the context-compression notification can still return in a refreshed branch.

crates/zeroclaw-channels/src/orchestrator/mod.rs
```diff
@@ -4856,6 +4856,15 @@ async fn process_channel_message_body(
...
     }
 
+    let context_compression_notice_pending = ctx
+        .agent_cfg
+        .resolved
+        .context_compression
+        .notify_on_compression
+        && ctx.context_token_budget > 0
+        && zeroclaw_runtime::agent::history::estimate_history_tokens(&history)
+            > ctx.context_token_budget;
```
IftekharUddin, line 4866: 🔴 **Blocking — this predicts a trim that the runtime may not perform.** `estimate_history_tokens(&history) > context_token_budget` is not equivalent to the runtime-owned `TrimResult.trimmed`: `trim_to_recent_turns` preserves the newest whole turn and returns `trimmed = false` when only that turn exists. I reproduced this on exact head with an oversized reply-eligible first turn; the channel still sent `I condensed our earlier conversation...` even though there was no earlier turn to drop. **Why this matters:** the actual trim result is the canonical fact. **If ignored:** users receive a false successful-trim status on precisely the over-budget edge this feature is meant to explain. **Required change:** drive the notice from an actual runtime trim decision (or a guaranteed runtime pre-trim signal) and cover the oversized-single-turn no-trim case.

crates/zeroclaw-channels/src/orchestrator/mod.rs
```diff
@@ -4992,6 +5001,18 @@ async fn process_channel_message_body(
         return;
     }
 
+    if context_compression_notice_pending && let Some(channel) = target_channel.as_ref() {
+        let notice = "I condensed our earlier conversation to stay within context; continuing now.";
```
IftekharUddin, line 5005: 🔴 **Blocking — move this user-visible status into Fluent.** This shared channel message is runtime-facing user text, while the repository contract and adjacent orchestrator statuses resolve through `zeroclaw_runtime::i18n`. **Why this matters:** every channel uses the configured runtime locale. **If ignored:** non-English installations get an English-only status and the copy bypasses catalogue review/fallback checks. **Required change:** add an English source key through the documented localization workflow and resolve it with the runtime i18n helper here; do not hand-edit generated translations.

IftekharUddin (changes requested):
@sbenedetto I reviewed exact head 0e3de31c1b703db5fedc165d338ac4b11613dc93: the complete two-file diff and four-commit history, every top-level comment, all formal reviews, the one inline thread and its current resolution state, the live PR body/labels/milestone, exact-head CI, the channel reply-intent path, and the runtime-owned whole-turn trimming contract.

I also attempted a synthetic merge into current `master` f75b36a27857dc8bd81dc09ccadd66e9f8e42853. Git stops on conflicts in `crates/zeroclaw-channels/src/orchestrator/mod.rs` and `crates/zeroclaw-config/src/scattered_types.rs`. On the exact PR head, `git diff --check`, workspace formatting, and both submitted `compression_notice` tests pass. I then added a focused negative probe for a reply-eligible first turn whose content alone exceeds the budget: it fails because the channel sends the “I condensed our earlier conversation” notice even though the runtime has no earlier whole turn it can drop.

### ✅ Resolved — The earlier no-reply, opt-in, copy, and positive-coverage findings are addressed

The current head places the notice after the `NO_REPLY` return, defaults `notify_on_compression` to `false`, uses plain copy rather than exposing token internals, and covers both the reply and no-reply paths. Those concerns from @Audacity88 and @singlerider are settled. The one outdated inline thread bundles those same points and can be resolved.

### 🔴 Blocking — Drive the notice from the runtime's actual trim decision

The new channel-side boolean treats `estimate_history_tokens(&history) > context_token_budget` as proof that earlier conversation will be dropped. The runtime's canonical `trim_to_recent_turns` decision is stricter: it always preserves at least the newest whole turn and returns `trimmed = false` when there is only one turn to keep.

The exact-head negative probe reproduced the mismatch:

`CARGO_TARGET_DIR=/tmp/zeroclaw-pr-7162-target cargo test -p zeroclaw-channels --features channel-slack process_channel_message_does_not_claim_trim_when_single_turn_cannot_be_dropped -- --nocapture`

It observed `["chat-precheck:I condensed our earlier conversation to stay within context; continuing now.", "chat-precheck:visible reply"]` and failed because no earlier conversation could have been removed.

**Why this matters:** whether trimming actually happened is a runtime-owned fact, already represented by `TrimResult.trimmed` and, on current `master`, `TurnEvent::HistoryTrimmed`. Predicting it independently in the channel creates a second source of truth.

**If ignored:** an oversized first message, or another over-budget history with no droppable older whole turn, tells the user that conversation was condensed when nothing was removed. That is a false user-visible state report precisely when context handling is already under stress.

**Required change:** derive the notification from the runtime-owned trim decision, or introduce a runtime-owned pre-trim signal emitted only after the engine has proved that at least one whole turn will be dropped. Add a regression for an oversized single current turn that proves no successful-trim notice is emitted.

### 🔴 Blocking — Localize the new user-visible channel status

The notice is a bare English literal in the shared channel turn path. Repository policy requires user-visible runtime text to come from the Fluent catalogue, and the current orchestrator already resolves comparable channel status strings through `zeroclaw_runtime::i18n`.

**Why this matters:** this message is delivered directly to users across every configured channel and therefore participates in the runtime locale contract.

**If ignored:** non-English installations receive an English-only system status while the rest of the runtime follows the configured locale, and the text remains outside the catalogue's review and fallback checks.

**Required change:** add an English source key through the documented Fluent workflow and resolve it with the runtime i18n helper at the send boundary. Do not hand-edit generated translations.

### 🔴 Blocking — Refresh and validate the branch against current master

The branch is `CONFLICTING` / `DIRTY`. The synthetic merge conflicts in the channel test helper/provider that current `master` extended and in `ContextCompressionConfig`, where current `master` changed the `summary_provider` contract. Required CI is green for the July 8 head but has not validated any resolution against current `master`.

**Why this matters:** these are the exact production/config and regression-test owners changed by this PR, not unrelated conflict noise.

**If ignored:** the PR cannot merge, and a casual resolution can discard current observer/precheck-delay coverage, the current provider-reference contract, or the new notice tests while still looking locally plausible.

**Required change:** merge current `master` into the contributor branch as previously requested, reconcile both owners, and rerun required CI plus the focused reply/no-reply/single-turn notice tests on the resulting head.

### 🔴 Blocking — Refresh the public template and rollback evidence

The body predates the current template: it lacks the reviewer A/B test section, relied-on CI and known-gap fields, prompt-injection and Rust/MSRV answers, and an exact live-label snapshot. Its rollback command also calls `git revert 9b3bbe26f` even though that commit is a merge commit requiring a mainline and omits later current-head work.

**Why this matters:** this is a user-visible config and channel behavior change, so reviewers need reproducible before/after steps and an executable rollback path tied to what would actually merge.

**If ignored:** the public record continues to claim complete compatibility and rollback evidence while the documented revert command fails and the current conflict/negative-path gap is absent from the evidence.

**Required change:** after the code and merge resolution are final, update the body to the current template with exact-head validation, the live labels, the single-turn coverage, and a rollback command that matches the eventual merged commit shape.

### 🟢 What looks good — The feature remains opt-in, generic, and best-effort

The config default preserves existing installations, the send stays behind the shared `Channel` boundary rather than adding adapter-specific implementations, and delivery failure does not abort the user turn. Those are the right boundaries to preserve when the runtime-owned trim signal is wired through.

Audacity88:
The July 31 follow-up date has passed without an author response or update, and the branch conflicts with current `master`, so I’m closing this PR.

The opt-in compression notice can return in a refreshed PR if it is rebuilt against the current runtime-owned trimming path and preserves the established reply-intent behavior.

PR #7166, opened 2026-06-03
fix(zerocode): let quickstart agent name accept hotkey letters

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:** In the zerocode Quickstart "agent" modal, the **name** field silently dropped any character that doubles as a modal hotkey — `e` (edit in \$EDITOR), `t` (from template), `c` (clear), `d` (delete), plus their uppercase. The catch-all key arm pushed to the buffer only when `action.is_none()`, so whenever `QuickstartModalAction::from_chord` resolved a key to an action the char was discarded — even though those hotkeys are only meaningful on personality-file rows (already gated on `on_file`). Result: agent names couldn't contain `e`, `c`, `t`, or `d`.
- Extracted a small pure `typed_char(key)` helper that yields the printable char for a keypress (ignoring the chord mapping; rejecting only CONTROL chords and non-character keys) and used it for both the name field and the text-input modal (which had the same shape, sans bug).
- Added regression tests covering `e`/`t`/`c`/`d` (both cases) and the control/non-char rejection path.
- **Scope boundary:** Only the Quickstart modal key handling in `apps/zerocode/src/quickstart_pane.rs`. No change to the personality/"soul" `.md` file rows (those are edited via \$EDITOR/template/clear and never took inline typing), no change to the keymap/chord definitions, no daemon/wire changes.
- **Blast radius:** zerocode TUI Quickstart pane only. The `typed_char` helper is also now used by the existing TextInput modal, which previously had equivalent inline logic (behavior unchanged there).
- **Linked issue(s):** None (no existing issue/PR found for this).
- **Labels:** `type: fix`, `risk: low`, `size: S`

Scoped to the affected crate (`zerocode`); the change is isolated to one file in that crate.

\`\`\`
=== cargo fmt --all -- --check ===
OK: fmt clean

=== cargo clippy -p zerocode --all-targets -- -D warnings ===
    Checking zerocode v0.8.0-beta-2 (/Users/jordantian/zeroclaw/apps/zerocode)
    Finished \`dev\` profile [unoptimized + debuginfo] target(s) in 2m 23s

=== cargo test -p zerocode ===
running 14 tests
test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
running 203 tests
test result: ok. 203 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
\`\`\`

- **Beyond CI — what did you manually verify?** Built `zerocode` from this branch and reproduced the original bug on master (typing `test` in the agent name yielded `s` only). Confirmed the new regression test `name_field_accepts_hotkey_letters` fails against the pre-fix handler shape and passes after. Did NOT manually re-walk every modal (picker/channel/peer-group) since they don't route through the changed arms.
- **If any command was intentionally skipped, why:** Ran the battery scoped to `zerocode` rather than the full workspace — the diff touches a single file in that crate and doesn't change any shared/public API.

## Compatibility (required)

## Rollback

Low-risk: \`git revert 2af7bb880\`.

### DIFF
--- a/apps/zerocode/src/quickstart_pane.rs
+++ b/apps/zerocode/src/quickstart_pane.rs
@@ -113,6 +113,20 @@ fn synth_enter() -> KeyEvent {
     KeyEvent::new(KeyCode::Enter, crossterm::event::KeyModifiers::NONE)
 }
 
+/// The character a key press contributes to a free-text buffer (the
+/// agent-name field), or `None` for control chords and non-character
+/// keys. Letters that double as modal hotkeys on file rows — `e` (edit
+/// in $EDITOR), `t` (from template), `c` (clear), `d` (delete) — are
+/// still plain text on a text row, so this deliberately ignores the
+/// chord mapping: the hotkey arms are gated on `on_file` and never fire
+/// while the cursor is on the name field.
+fn typed_char(key: &KeyEvent) -> Option<char> {
+    match key.code {
+        KeyCode::Char(c) if !key.modifiers.contains(KeyModifiers::CONTROL) => Some(c),
+        _ => None,
+    }
+}
+
 fn action_row_line(label: &str, is_cursor: bool) -> Line<'static> {
     let glyph = if is_cursor { " › " } else { "   " };
     let style = if is_cursor {
@@ -1018,7 +1032,7 @@ impl QuickstartPane {
         let Some(modal) = self.active_modal.as_mut() else {
             return;
         };
-        use crate::keymap::{Chord, QuickstartModalAction};
+        use crate::keymap::QuickstartModalAction;
         let action = QuickstartModalAction::from_chord(&key);
         match modal {
             Modal::Picker(p) => match action {
@@ -1120,9 +1134,7 @@ impl QuickstartPane {
                     t.buf.pop();
                 }
                 _ => {
-                    if let KeyCode::Char(c) = key.code
-                        && !key.modifiers.contains(KeyModifiers::CONTROL)
-                    {
+                    if let Some(c) = typed_char(&key) {
                         t.buf.push(c);
                     }
                 }
@@ -1335,15 +1347,8 @@ impl QuickstartPane {
                         a.files.insert(filename, String::new());
                     }
                     _ => {
-                        if on_name
-                            && let KeyCode::Char(c) = key.code
-                            && !key.modifiers.contains(KeyModifiers::CONTROL)
-                        {
-                            if action.is_none() {
-                                a.name.push(c);
-                            } else {
-                                let _ = Chord::char(c);
-                            }
+                        if on_name && let Some(c) = typed_char(&key) {
+                            a.name.push(c);
                         }
                     }
                 }
@@ -2312,6 +2317,34 @@ mod tests {
         let mut f = complete_form();
         f.agent_name.clear();
         assert!(!f.all_selectors_satisfied());
+    }
+
+    #[test]
+    fn name_field_accepts_hotkey_letters() {
+        // Regression: e/t/c/d double as Agent-modal hotkeys (edit in
+        // $EDITOR, from template, clear, delete) on file rows. On the
+        // name row they are plain text, but the old handler routed every
+        // keypress through the chord mapping and dropped any char that
+        // resolved to an action — so agent names could not contain those
+        // letters. `typed_char` is the text-buffer path; assert it keeps
+        // them, and that they really are bound actions (bug was reachable).
+        use crate::keymap::QuickstartModalAction;
+        for ch in ['e', 'c', 't', 'd', 'E', 'C', 'T', 'D'] {
+            let key = KeyEvent::new(KeyCode::Char(ch), KeyModifiers::NONE);
+            assert_eq!(typed_char(&key), Some(ch), "name field must accept '{ch}'");
+            assert!(
+                QuickstartModalAction::from_chord(&key).is_some(),
+                "'{ch}' must be a modal hotkey for this regression to be real"
+            );
+        }
+    }
+
+    #[test]
+    fn typed_char_ignores_control_and_non_char_keys() {
+        let ctrl_c = KeyEvent::new(KeyCode::Char('c'), KeyModifiers::CONTROL);
+        assert_eq!(typed_char(&ctrl_c), None);
+        let enter = KeyEvent::new(KeyCode::Enter, KeyModifiers::NONE);
+        assert_eq!(typed_char(&enter), None);
     }
 
     #[test]


WareWolf-MoonWall (approved):
## Review of PR #7166

**Commit reviewed:** 2af7bb880 (per PR body rollback note)  
**Verdict:** Approve

---

### 🟢 What looks good — Root-cause fix with surgical extraction

The fix is textbook: found the bug (name field silently ate hotkey letters), traced the root cause (the catch-all arm routed every keypress through chord mapping and dropped chars that resolved to actions), extracted a pure helper (`typed_char`) that ignores the chord table and gates only on CONTROL modifiers + non-character keys, applied it consistently to both the name field and the existing TextInput modal (which already had equivalent inline logic), and shipped regression tests that fail on the old handler and pass on the new one.

The `typed_char` helper is appropriately scoped: it returns the printable character for a keypress (rejecting control chords and non-character keys like Enter), and it deliberately ignores the chord mapping because the hotkey actions (`e`/`t`/`c`/`d` for edit/template/clear/delete) are only meaningful on personality-file rows — those arms are already gated on `on_file` and never fire when the cursor is on the name field. The old code's `if action.is_none()` condition was the bug: it meant any char that happened to map to an action got discarded, even though the action itself wouldn't fire. The new code sidesteps the chord table entirely for text input, which is correct.

The scope boundary is respected: the change is isolated to `quickstart_pane.rs` in the `zerocode` crate. No shared/public API touched. The validation evidence (scoped clippy + test run for `zerocode` only) is proportionate to the blast radius. The PR body names all the things that *didn't* change (personality rows, keymap definitions, daemon wire protocol) and why they're out of scope, which is exactly the kind of comprehension a reviewer needs.

---

### 🟢 Test coverage matches the bug surface

The regression test `name_field_accepts_hotkey_letters` is thorough without being gold-plated:

- Covers all four hotkey letters (`e`/`c`/`t`/`d`) in both lower and upper case (8 cases total).
- Asserts that `typed_char` returns `Some(ch)` for each — proving the name field will accept them.
- Asserts that `QuickstartModalAction::from_chord` returns `Some(_)` for each — proving the bug was reachable (these chars *are* bound actions, so the old handler really would have dropped them).

The companion test `typed_char_ignores_control_and_non_char_keys` covers the rejection path (control chords + non-character keys like Enter). Together these two tests fully specify the `typed_char` contract. If the helper regresses, the tests will catch it. If someone later adds a new hotkey that collides with a reasonable agent name character, the existing test will start failing and flag the conflict.

---

### 🟢 Minimal diff, no side effects

The diff removes 13 lines and adds 46 (net +33), almost entirely test coverage and the `typed_char` helper with its doc comment. The production logic delta is tiny:

- Replaced inline `if let KeyCode::Char(c) = key.code && !key.modifiers.contains(KeyModifiers::CONTROL)` with `if let Some(c) = typed_char(&key)` in two places (the name field and the TextInput modal).
- Removed the buggy `if action.is_none()` guard from the name field arm.
- Dropped the unused `Chord` import (the name field no longer routes through chord mapping).

No speculative abstractions, no unrelated cleanups, no formatting churn. The change does exactly what the bug report says and nothing else. Per FND-006 §2.3, this is the shape of a correct fix: address the root cause, add the test that would have caught it, stop.

---

### 🟢 Correct no-DRY-violation call

The PR body states the `typed_char` helper is "also now used by the existing TextInput modal, which previously had equivalent inline logic (behavior unchanged there)." Cross-checked against the diff: the TextInput modal's catch-all arm at L1123–1127 previously had:

```rust
if let KeyCode::Char(c) = key.code
    && !key.modifiers.contains(KeyModifiers::CONTROL)
{
    t.buf.push(c);
}
```

The new code (L1137) is:

```rust
if let Some(c) = typed_char(&key) {
    t.buf.push(c);
}
```

This is a correct de-duplication: the logic was identical, and now both paths call the same helper. The diff confirms "behavior unchanged there" — the helper's contract matches the inline condition exactly (CONTROL modifier rejection + non-character key rejection). Per AGENTS.md ABSOLUTE RULE, this is not a DRY violation: the helper is not a cache or snapshot; it's a resolver function that computes the answer on demand from the input `KeyEvent`. The two call sites had diverged slightly (the name field had the buggy `if action.is_none()` wrapper, the TextInput modal did not), and the extraction unifies them at the correct shared abstraction.

---

### 🔵 Suggestion — Consider a follow-up comment near the QuickstartModalAction bindings

Optional refinement for future maintainability: the `typed_char` function deliberately ignores the chord mapping because hotkeys like `e`/`t`/`c`/`d` are only meaningful on file rows (gated on `on_file`). This is correct by construction today, but if someone later adds a new `QuickstartModalAction` variant bound to a printable character that *should* shadow text input on the name field (e.g. a hypothetical "submit with Enter" hotkey that fires even when on the name row), they might not realize the name field bypasses the chord table.

A small inline comment near the `QuickstartModalAction` binding list in `keymap/actions.rs` (around L378–388) could prevent that confusion:

```rust
// Note: the Quickstart modal's name field uses `typed_char` (which
// ignores this chord table) because hotkeys like EditWithEditor,
// EditTemplate, EditCopy are only meaningful on file rows (gated on
// `on_file`). If you add a new binding here that should shadow text
// input on the name row, you'll need to adjust the name field's
// catch-all arm in `quickstart_pane.rs`.
```

This is a 🔵 suggestion, not a block. The current design is sound without the comment, and the regression test will catch any future collision. The comment would just make the design intent more discoverable for someone who's editing the keymap six months from now.

---

### ✅ Labels

The PR body suggests `type: fix`, `risk: low`, `size: S`. All three are correct:

- **type: fix** — addresses a confirmed bug (agent names couldn't contain `e`/`c`/`t`/`d`).
- **risk: low** — scoped to a single file in the `zerocode` TUI, no shared API changes, no daemon/wire protocol changes, comprehensive test coverage.
- **size: S** — 46 additions / 13 deletions, single file touched.

I don't have label permissions in this session, so I'm noting correctness here. If @JordanTheJet or another maintainer with triage access sees this, please apply those three labels.

---

## Summary

This is a clean, well-scoped fix. The author identified the root cause (catch-all arm routing through chord mapping even though the hotkey actions are gated elsewhere), extracted a pure helper with a clear contract, applied it consistently, and added regression tests that fail on the old code and pass on the new code. The scope boundary is respected, the validation evidence is proportionate, and the diff has no side effects. Approve.

The 🔵 suggestion about a comment near the keymap bindings is optional — the design is correct without it, and I'm only flagging it because someone editing that keymap in six months might not remember why the name field bypasses the chord table. Take it or leave it.

singlerider (approved):
Reviewed 2af7bb8 against `upstream/master`. I'd independently started the same fix, so I cross-checked the diff against the master keymap and the Agent-modal dispatch line by line.

### 🟢 What looks good — root cause is correctly and completely addressed

The bug is real on master: the name-field catch-all only typed when `action.is_none()`, and `e/t/c/d` (plus uppercase) all resolve to `Some(...)` via `from_chord`, so they were dropped — with a dead `let _ = Chord::char(c)` in the swallow branch. Confirmed `DeleteRow`/`EditWithEditor`/`EditTemplate`/`EditCopy` are all bound to those letters in `keymap/actions.rs`, and that the Agent modal's hotkey arms are gated `if on_file`, so they genuinely never fire on the name row. Routing the name and TextInput buffers through `typed_char` is the right call — text rows take printable characters; the chord mapping is only meaningful on file rows.

### 🟢 What looks good — `d` and the TextInput reuse

Catching `d`/`D` matters: the Agent modal has no `DeleteRow` arm, so `d` always fell to the catch-all and was silently swallowed pre-fix. Folding the identical TextInput catch-all into the same helper removes a second copy of the same shape before it can drift. No behavior change there (CONTROL still rejected), just deduplication.

### 🟢 What looks good — the regression test proves the bug was reachable

`name_field_accepts_hotkey_letters` asserting both `typed_char(key) == Some(ch)` **and** `from_chord(key).is_some()` is exactly right — it pins that these letters are bound actions, so the regression can't silently rot into a tautology if the keymap changes. `typed_char_ignores_control_and_non_char_keys` covers the rejection path.

One note, non-blocking: `typed_char` rejects CONTROL but not ALT, so ALT+char (e.g. the ALT+b / ALT+f pane-nav chords) types a literal in the name field. That's pre-existing behavior — the old name arm also only checked CONTROL, and pane-nav isn't reachable while a modal owns the key — so this PR doesn't regress it. Fine to leave as-is; flagging only so it's a conscious choice.

Approving. Clean, scoped, well-tested fix.

Reviewed commit: 2af7bb8 · base upstream/master

PR #7169, opened 2026-06-03
fix(zerocode): don't flag unreached quickstart selectors as errors

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The zerocode Quickstart pane calls `revalidate()` after **every** selector commit, which validates the *whole* submission against the daemon. The runtime validator (`apply_into`) short-circuits at the first failing step, so the instant the model provider is committed and starts passing, the next still-empty step — the **risk profile** — fails, producing exactly one error. The status strip renders only the count (`1 error(s) — fix selectors and resubmit`), never *which* step, so it reads as if the provider step itself broke.
  - Fix: `revalidate()` now keeps only errors whose `step` maps to a selector the user has actually filled, via a new pure helper `retain_filled_selector_errors()`. Unreached selectors are already tracked as `[ ]` in the checklist; submit re-validates the full set (nothing empty left to short-circuit on), so genuine errors on filled selectors — e.g. an alias collision on the model provider — still surface.
- **Scope boundary:** Only filters how *already-produced* validation errors are surfaced mid-build in the zerocode Quickstart pane. Does not touch the runtime validator, the daemon validation/apply logic, the CLI or web Quickstart surfaces, or submit-time behavior. The status strip still shows only a count (not the message text) — left as a separate follow-up.
- **Blast radius:** Confined to `apps/zerocode` Quickstart onboarding. No change to validation rules, config writes, or any non-zerocode surface.
- **Linked issue(s):** None.
- **Labels:** `bug`, `risk: medium`, `size: XS`, `quickstart`

```bash
cargo fmt --all -- --check
cargo test -p zerocode
```

`cargo fmt --all -- --check` — clean (no diff; exit 0).

```
    Checking zeroclaw-channels v0.8.0-beta-2 (/Users/jordantian/zeroclaw/crates/zeroclaw-channels)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 45s
```
(exit 0, no warnings)

`cargo test -p zerocode`:
```
test quickstart_pane::tests::revalidate_hides_errors_for_unfilled_selectors ... ok
test quickstart_pane::tests::revalidate_keeps_errors_for_filled_selectors ... ok
test result: ok. 203 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
```

Toolchain: `rustc 1.93.1 (01f6ddf75 2026-02-11)`.

- **Beyond CI — what did you manually verify?** Reproduced the root cause with a throwaway test in `zeroclaw-runtime` (since removed): a *complete* submission validates cleanly against both a default config and a real `default_provider = "minimax"` config, while the mid-build state (valid provider + empty risk profile) returns exactly one `RiskProfile: unknown risk preset ''` error — the phantom the user saw. Two regression tests now lock in the fix in both directions (unfilled-selector error dropped, filled-selector error retained).
- **If any command was intentionally skipped, why:** Ran `cargo test -p zerocode` (the only crate this PR changes) instead of the full-workspace `cargo test`. The diff is confined to one file in `apps/zerocode`; full clippy ran across all targets.

## Compatibility (required)

- Backward compatible? `Yes`
- Upgrade steps: None — UI-only behavior change in the zerocode Quickstart pane.

`git revert <sha>` restores the previous zerocode Quickstart error surfacing behavior. No config, schema, daemon, or runtime validator changes are included.


WareWolf-MoonWall (approved):
## Review of PR #7169

Reviewed f64ebea by @WareWolf-MoonWall

---

### 🟢 What looks good — problem diagnosis and solution design

The root cause analysis is spot on: `revalidate()` runs after every selector commit, validates the *complete* submission, and the runtime validator short-circuits at the first failure. Mid-build, that first failure is almost always the next unfilled selector (e.g. the still-empty risk profile after committing the model provider). The user sees `"1 error(s) — fix selectors and resubmit"` without context about *which* step failed, making it appear as if their just-completed step broke.

The fix is architecturally correct: filter validation errors mid-build to show only errors from selectors the user has actually filled, while submit-time validation still sees the full set. This respects the boundary between "incremental feedback during onboarding" and "final submission validation" — no changes to the runtime validator, config writes, or any non-zerocode surface.

The helper function `retain_filled_selector_errors()` is pure, well-documented, and easy to reason about. Its signature makes the intent explicit: take a form state and an error list, return a filtered list. Good separation of concerns.

---

### 🟢 What looks good — test coverage

Two regression tests lock in the fix in both directions:

1. `revalidate_hides_errors_for_unfilled_selectors` — reproduces the exact bug (model provider filled, risk profile empty → error dropped)
2. `revalidate_keeps_errors_for_filled_selectors` — ensures genuine errors on filled selectors still surface

This is the right level of coverage for a UI behavior fix. The tests are clear, minimal, and directly tied to the problem statement. No over-testing, no under-testing.

---

### 🟢 What looks good — validation evidence

All required checks passed:
- `cargo clippy --all-targets -- -D warnings` (no warnings)
- `cargo test -p zerocode` (203 passed, including the two new tests)

The author also manually reproduced the root cause with a throwaway test in `zeroclaw-runtime` (removed post-verification), confirming that a complete submission validates cleanly while mid-build state produces the phantom error. This goes beyond the CI minimum and demonstrates thorough validation discipline.

---

### 🔵 Suggestion — consider surfacing *which* step in the status strip

The PR description notes:

> The status strip still shows only a count (not the message text) — left as a separate follow-up.

This is the right call for scoping this PR, but worth tracking separately: when there's exactly one error and it's from a filled selector, showing `"RiskProfile: unknown risk preset ''"` instead of just `"1 error(s) — fix selectors and resubmit"` would eliminate ambiguity about which step broke. Not blocking — the current fix already resolves the phantom-error UX problem — but a natural next step.

No issue filed yet; consider opening one if you agree this would improve the UX.

---

### 🔵 Suggestion — document the `Selector::Submit` step mapping

In `Selector::step()`, both `Selector::Agent` and `Selector::Submit` map to `QuickstartStep::Agent`. This makes sense (Submit is the action row, not a form field), but it's implicit. A one-line comment above that match arm would help future readers:

```rust
// Both Agent and Submit map to Agent step — Submit is the action
// row that triggers the final apply, not a distinct form field.
Selector::Agent => QuickstartStep::Agent,
Selector::Submit => QuickstartStep::Agent,
```

Not blocking — the code is correct and the test coverage confirms the behavior. This is a maintainability suggestion.

---

### ✅ Verdict

**Approved.** This is a well-scoped bug fix with clear problem diagnosis, clean implementation, appropriate test coverage, and thorough validation evidence. The fix is confined to the zerocode quickstart pane UI layer as claimed, with no changes to the runtime validator or config logic. The two regression tests lock in the behavior in both directions (unfilled-selector errors dropped, filled-selector errors retained).

Per FND-005 §5, the two 🔵 suggestions above are optional — the author can accept or pass. They're about future maintainability and UX polish, not correctness or architectural soundness.

Ready to merge once CI passes.

Issue #6293, opened 2026-05-03
RFC: Air-gapped execution mode with companion daemon over unix socket (enclave support)


singlerider:
## Summary

Split ZeroClaw into two isolated processes connected by a unix socket: an offline agent container that can execute shell commands and run the LLM, and an online companion daemon that proxies internet access exclusively via approved ACP/MCP servers. This architecture enables ZeroClaw to run inside enclaves (e.g. [Caution](https://caution.co), self-hosted [EnclaveOS](https://distrust.co/blog/enclaveos.html)) and eliminates the need to trust the agent with direct network access.

## Problem statement

Currently ZeroClaw runs as a single process with unrestricted network access. Users who want to give the agent access to sensitive resources (email, credentials, private repos) must trust the agent entirely. There is no way to grant capability without granting access. This is a fundamental trust problem that limits adoption for security-sensitive use cases and makes enclave deployment impossible.

## Proposed solution

Separate ZeroClaw into two containers/processes:

**Offline container (the agent):**
- Runs with `--network=none`
- Has access to the LLM and can execute shell commands
- Exposes a single unix socket as its only external interface
- Cannot initiate any network connection directly

**Online companion daemon:**
- Mounts the same unix socket
- Has internet access but cannot run arbitrary commands
- Serves only whitelisted, well-specified ACP/MCP endpoints (e.g. fetch email, push a PR, download a file)
- Proxies all internet-bound requests from the agent over the socket using structured ACP/MCP protocols
- Can also serve the web UI over this socket

**Example workflow:**
1. User instructs ZeroClaw: "update every package in this tree, test-compile each one, commit in build order"
2. Agent writes code and creates commits entirely offline
3. When a commit is ready, it passes the commit body over the unix socket to the companion daemon
4. The companion daemon executes a hardcoded whitelisted action: push the commit as a pull request
5. No step in this flow requires the agent to have network access

**Enclave path:**
Once ZeroClaw can run offline with a unix socket as its only external interface, it should work in [Caution](https://caution.co) or self-hosted [EnclaveOS](https://distrust.co/blog/enclaveos.html) with encrypted memory, remote attestation, and provable end-to-end encryption.

## Non-goals / out of scope

- The companion daemon does not expose a general-purpose HTTP proxy or arbitrary curl access
- The companion daemon cannot run LLM inference
- No changes to existing channel integrations in v1 (channels remain on the companion side)
[...]

PR #7176, opened 2026-06-03
build(container): StageX container pipeline with musl static linking

ConYel:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **`Cargo.toml` (root + runtime):** opt out of `aws-lc-rs` on `rustls`, `tokio-rustls`, and `rumqttc` — removes `aws-lc-sys`, which transitively requires `cmake` at build time. StageX base images carry no `cmake`, so every transitive path to it had to be severed. The `ring` crypto provider is explicitly re-enabled by name on all three crates, preserving full TLS functionality with no behavioral change.
  - **`Containerfile` (new):** 4-stage StageX pipeline exporting two runtime images via build targets. `package` (default) ships the daemon + `zerocode` with `default-channels`; `package-fat` ships the daemon with `channels-full` (all channels) + `zerocode`. The build stage runs `fetch` → `fmt` → `clippy` → `release` under `--network=none`. All base images pinned by SHA256. A `libstdc++.a` is assembled from StageX's LLVM `libc++.a` + `libc++abi.a` so the musl-static link resolves.
  - **`Makefile` (new):** `build` / `build-fat` (podman build per target), `extract` / `extract-fat` (copy binaries to host), `shell-debug`, `clean`.
  - **`.dockerignore`:** restore `tests/` to the build context (needed for `clippy --all-targets`), drop the Tauri source-exclusion stanza in favor of the workspace-level resolution, and add `!apps/zerocode/` so the TUI manager is shipped.
  - **`tests/component/dockerignore_test.rs`:** drop `"tests"` from `MUST_EXCLUDE` to align the assertion with the updated `.dockerignore`.
- **Scope boundary:** Container/build tooling and the TLS-provider feature narrowing only. No changes to `src/` logic, runtime behavior, config schema, or CLI surface. The `dockerignore_test` change is a one-line assertion update, not a behavior change.
- **Blast radius:** The `rustls`/`tokio-rustls`/`rumqttc` feature changes affect every consumer that compiles those crates — mitigated by re-enabling `ring` explicitly so the crypto backend is unchanged in practice. `cargo tree -i aws-lc-sys` returns zero matches after the change.
- **Linked issue(s):** None.
- **Labels:** `type: ci`, `risk: medium`, `size: XS`, `dependencies`, `tests`.

### Why StageX

[stagex.tools](https://stagex.tools) provides minimal, reproducible, distroless base images. `pallet-rust` ships the Rust toolchain without a package manager or build tools like `cmake`/`gcc`; every build dependency must be resolved by cargo from the registry. The absence of `cmake` directly motivates the `aws-lc-sys` elimination — that crate compiles the AWS-LC C library at build time and requires `cmake`.

Three StageX images are used, each pinned by digest:

| Image | SHA256 | Role |
|---|---|---|
| `stagex/pallet-rust` | `2d90b9552412` | Build stage — rustc, cargo, rust-lld, musl target |
| `stagex/core-filesystem` | `cd3a66471ce1` | Runtime — `/dev`, `/tmp`, SSL dirs |
| `stagex/core-ca-certificates` | `7773dae6630a` | CA cert bundle for HTTPS |

### Container pipeline design

**Two runtime targets.** `package` is the minimal-but-usable default (daemon + `zerocode` + `default-channels`, ~31 MB). `package-fat` is the extended surface (daemon + `zerocode` + `channels-full`). Operators select with `podman build --target <package|package-fat>` or the `make build` / `make build-fat` convenience targets. The fat stage reuses Stage 1's `/src` (`COPY --from=build /src /src`) and Stage 1's already-built `zerocode` (`COPY --from=build … zerocode`) rather than recompiling either.

**`default-channels` included.** The default image ships the four canonical messaging channels (`channel-acp-server`, `channel-webhook`, `channel-email`, `channel-telegram`). Opt-in channels — `channel-matrix`, social channels, etc. — are not in `default-channels` and are not compiled into the default image; use `package-fat` (`channels-full`) or a forked feature list for those.

**`zerocode` shipped.** Both the daemon (`zeroclaw`) and the TUI config manager (`zerocode`) are built in Stage 1 and packaged into both runtime images, so operators have an in-image surface to inspect or repair config.

**`--network=none` on the build step.** The compilation `RUN` carries `--network=none`:
- All dependencies come from the prior `cargo fetch` `RUN` (the only step with network access).
- `--frozen` ensures `Cargo.lock` is authoritative; no index update is attempted.
- Supply-chain integrity: even if the registry or a git dependency goes rogue between fetch and build, the build `RUN` cannot reach it.

**No `sed` manifest mutation.** The workspace manifest is no longer edited in place. The source tree is copied as-is (`COPY . .`) and cargo resolves the workspace naturally; `.dockerignore` governs what enters the context.

**No test stage.** Unit and doc tests run in native CI. The container pipeline produces a deployment artifact, not a test runner — a second debug compilation would double disk and compile-time cost with no coverage benefit over CI.

### Cargo.toml changes

`tokio-rustls` 0.26 enables `aws_lc_rs` in its default features, pulling `aws-lc-rs` → `aws-lc-sys` (C build → `cmake`). `rumqttc` 0.25 pulls `tokio-rustls` with default features, same chain.

| File | Change |
|---|---|
| `Cargo.toml` (`rustls`) | `default-features = false`, `features = ["ring", "logging", "std", "tls12"]` |
| `Cargo.toml` (`tokio-rustls`) | `default-features = false`, `features = ["ring", "logging", "tls12"]` |
| `crates/zeroclaw-runtime/Cargo.toml` (`rumqttc`) | `default-features = false`, `features = ["use-rustls-no-provider"]` |

`cargo tree -i aws-lc-sys` returns zero matches after these changes.

```bash
cargo fmt --all -- --check
cargo test
```

```
$ cargo fmt --all -- --check
# no output — all clean

    Finished — no warnings

$ cargo test
test result: ok. 535 passed; 0 failed; 0 ignored

$ make build
[package] target built
Successfully tagged localhost/zeroclaw:stagex

$ podman run --rm zeroclaw:stagex --version
zeroclaw 0.8.0-beta-2

$ podman images zeroclaw:stagex
localhost/zeroclaw  stagex  31.3 MB

$ cargo tree -i aws-lc-sys
error: package 'aws-lc-sys' is not present in the dependency graph
```

- **Beyond CI — what did you manually verify?** `cargo tree` confirms `aws-lc-sys` absent. Full container build produces `zeroclaw 0.8.0-beta-2`; `make extract` copies the static `zeroclaw` and `zerocode` binaries; `file zeroclaw` confirms ELF x86-64, statically linked. Both `package` and `package-fat` targets build.
- **If any command was intentionally skipped, why:** `cargo test --doc` is not run in the container (no rustdoc in the stage); doctests run in native CI. All 535 unit tests pass.

| Question | Answer |
|---|---|
| New permissions, capabilities, or file system access? | No |
| New external network calls? | No — `--network=none` on the build `RUN`; only `cargo fetch` accesses the network |
| Secrets / tokens / credentials handling changed? | No |
| PII, real identities, or personal data in diff, tests, fixtures, or docs? | No |

## Compatibility (required)

- **Backward compatible?** Yes. The `Cargo.toml` changes remove a transitive default feature (`aws_lc_rs`) that was never explicitly depended on; `ring` is enabled by name on all three crates. No behavioral change observable to any consumer.
- **Upgrade steps:** None. `cargo build` continues unchanged; the container build is additive.

- **Fast rollback command/path:** `git revert <sha>` — additive build tooling plus a backend-neutral TLS feature narrowing, no runtime effect.
- **Observable failure symptoms:** If a future crate depends on `aws-lc-rs` specifically (not `ring`), its build fails with a compile error — build-time only, never a runtime surprise. For the container itself, a broken pipeline surfaces at `podman build` time, never at deploy.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,7 +13,6 @@ web/dist
 # Documentation and examples (not needed for runtime)
 docs
 examples
-tests
 
 # Markdown files (README, CHANGELOG, etc.)
 *.md
@@ -73,11 +72,6 @@ lcov.info
 # Note: apps/tauri/Cargo.toml is required for workspace resolution in Docker build.
 apps/*
 !apps/tauri/
-apps/tauri/src/
-apps/tauri/icons/
-apps/tauri/gen/
-apps/tauri/capabilities/
-apps/tauri/tauri.conf.json
-apps/tauri/build.rs
 !apps/tauri/Cargo.toml
+!apps/zerocode/
 scripts/
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -204,10 +204,10 @@ tokio-socks = { version = "0.5", optional = true }
 futures-util = { version = "0.3", default-features = false, features = ["sink"] }
 regex = "1.10"
 hostname = { version = "0.4.2", optional = true }
-rustls = { version = "0.23", optional = true }
+rustls = { version = "0.23", default-features = false, features = ["ring", "logging", "std", "tls12"], optional = true }
 rustls-pemfile = { version = "2", optional = true }
 rustls-pki-types = { version = "1.14.0", optional = true }
-tokio-rustls = { version = "0.26.4", optional = true }
+tokio-rustls = { version = "0.26.4", default-features = false, features = ["ring", "logging", "tls12"], optional = true }
 webpki-roots = { version = "1.0.6", optional = true }
 
 # email
--- a/Containerfile
+++ b/Containerfile
@@ -1,0 +1,48 @@
+# Stage 1: Fetch dependencies + build (fmt, clippy, release)
+FROM docker.io/stagex/pallet-rust@sha256:2d90b9552412ee2c4fa2a13b489c2f28c044be7fb5d6a942bfd5a480a5c288fd AS build
+
+WORKDIR /src
+COPY . .
+RUN --mount=type=cache,target=/root/.cargo/registry \
+    --mount=type=cache,target=/root/.cargo/git \
+    <<-'EOF'
+    set -e
+    sed -i '/"apps\/tauri"/d; /"tools\/fill-translations"/d; /"xtask"/d' Cargo.toml
+    sed -i '/^\[\[test\]\]$/,/^$/d' Cargo.toml
+    cargo fetch
+EOF
+
+RUN --mount=type=cache,target=/root/.cargo/registry \
+    --mount=type=cache,target=/root/.cargo/git \
+    --network=none \
+    <<-EOF
+	set -e
+	ARCH="$(uname -m)"
+	export RUSTFLAGS="-C target-feature=+crt-static -C linker=rust-lld -C link-arg=-L/usr/lib"
+
+	# Format check
+	cargo fmt --all -- --check
+
+	# Clippy (debug build, cleaned before release)
+	CARGO_TARGET_DIR=/tmp/target \
+	cargo clippy --all-targets --target "${ARCH}-unknown-linux-musl" -- -D warnings
+	rm -rf /tmp/target
+
+	# Release build with static musl linking
+	CARGO_TARGET_DIR=/target \
+	cargo build \
+		--frozen \
+		--release \
+		--target "${ARCH}-unknown-linux-musl" \
+		--no-default-features \
+		--features "agent-runtime,acp-bridge,gateway,schema-export,observability-prometheus" \
+		-p zeroclawlabs
+	mkdir -p /rootfs/usr/bin
+	cp /target/${ARCH}-unknown-linux-musl/release/zeroclaw /rootfs/usr/bin/zeroclaw
+EOF
+
+# Stage 2: Minimal runtime image
+FROM docker.io/stagex/core-filesystem@sha256:cd3a66471ce1f630fa77d5c9bd9829f9f9fab6302a1aaa64d67b74f1f069b750 AS package
+COPY --from=build /rootfs/ /
+COPY --from=docker.io/stagex/core-ca-certificates@sha256:7773dae6630aa3bdcc82cfec6c9265c0c501aaf0af67cc73631b09e1cff1b094 / /
+ENTRYPOINT ["/usr/bin/zeroclaw"]
--- a/Containerfile
+++ b/Containerfile
@@ -3,46 +3,87 @@ FROM docker.io/stagex/pallet-rust@sha256:2d90b9552412ee2c4fa2a13b489c2f28c044be7
 
 WORKDIR /src
 COPY . .
+
+# Fetch all workspace dependencies (network available)
 RUN --mount=type=cache,target=/root/.cargo/registry \
     --mount=type=cache,target=/root/.cargo/git \
-    <<-'EOF'
-    set -e
-    sed -i '/"apps\/tauri"/d; /"tools\/fill-translations"/d; /"xtask"/d' Cargo.toml
-    sed -i '/^\[\[test\]\]$/,/^$/d' Cargo.toml
     cargo fetch
-EOF
 
+# Offline build: fmt check, clippy, release binaries
 RUN --mount=type=cache,target=/root/.cargo/registry \
     --mount=type=cache,target=/root/.cargo/git \
     --network=none \
     <<-EOF
-	set -e
-	ARCH="$(uname -m)"
-	export RUSTFLAGS="-C target-feature=+crt-static -C linker=rust-lld -C link-arg=-L/usr/lib"
+    set -e
+    ARCH="$(uname -m)"
+    export RUSTFLAGS="-C target-feature=+crt-static -C linker=rust-lld -C link-arg=-L/usr/lib"
 
-	# Format check
-	cargo fmt --all -- --check
+    # Format check
+    cargo fmt --all -- --check
 
-	# Clippy (debug build, cleaned before release)
-	CARGO_TARGET_DIR=/tmp/target \
-	cargo clippy --all-targets --target "${ARCH}-unknown-linux-musl" -- -D warnings
-	rm -rf /tmp/target
+    # Clippy (debug build, cleaned before release)
+    CARGO_TARGET_DIR=/tmp/target \
+    cargo clippy --all-targets --target "${ARCH}-unknown-linux-musl" -- -D warnings
+    rm -rf /tmp/target
 
-	# Release build with static musl linking
-	CARGO_TARGET_DIR=/target \
-	cargo build \
-		--frozen \
-		--release \
-		--target "${ARCH}-unknown-linux-musl" \
-		--no-default-features \
-		--features "agent-runtime,acp-bridge,gateway,schema-export,observability-prometheus" \
-		-p zeroclawlabs
-	mkdir -p /rootfs/usr/bin
-	cp /target/${ARCH}-unknown-linux-musl/release/zeroclaw /rootfs/usr/bin/zeroclaw
+    # Build combined libstdc++.a from libc++.a + libc++abi.a (stagex ships LLVM libc++, not GCC libstdc++)
+    (mkdir -p /tmp/libwrap && cd /tmp/libwrap && ar x /lib/libc++.a && ar x /lib/libc++abi.a && ar rcs /usr/lib/libstdc++.a *.o && rm -rf /tmp/libwrap)
+
+    # Release build — zeroclawlabs (daemon)
+    CARGO_TARGET_DIR=/target \
+    cargo build \
+        --frozen \
+        --release \
+        --target "${ARCH}-unknown-linux-musl" \
+        --no-default-features \
+        --features "agent-runtime,default-channels,acp-bridge,gateway,schema-export,observability-prometheus" \
+        -p zeroclawlabs
+
+    # Release build — zerocode (TUI config manager)
+    CARGO_TARGET_DIR=/target \
+    cargo build \
+        --frozen \
+        --release \
+        --target "${ARCH}-unknown-linux-musl" \
+        -p zerocode
+
+    mkdir -p /rootfs/usr/bin
+    cp /target/${ARCH}-unknown-linux-musl/release/zeroclaw /rootfs/usr/bin/zeroclaw
+    cp /target/${ARCH}-unknown-linux-musl/release/zerocode /rootfs/usr/bin/zerocode
 EOF
 
-# Stage 2: Minimal runtime image
+# Stage 2: Minimal runtime image (target: package)
 FROM docker.io/stagex/core-filesystem@sha256:cd3a66471ce1f630fa77d5c9bd9829f9f9fab6302a1aaa64d67b74f1f069b750 AS package
 COPY --from=build /rootfs/ /
+COPY --from=docker.io/stagex/core-ca-certificates@sha256:7773dae6630aa3bdcc82cfec6c9265c0c501aaf0af67cc73631b09e1cff1b094 / /
+ENTRYPOINT ["/usr/bin/zeroclaw"]
+
+# Stage 3: Build fat binary with all channels (target: build-fat)
+FROM docker.io/stagex/pallet-rust@sha256:2d90b9552412ee2c4fa2a13b489c2f28c044be7fb5d6a942bfd5a480a5c288fd AS build-fat
+COPY --from=build /src /src
+WORKDIR /src
+RUN --mount=type=cache,target=/root/.cargo/registry \
+    --mount=type=cache,target=/root/.cargo/git \
+    --network=none \
+    <<-EOF
+    set -e
+    ARCH="$(uname -m)"
+    export RUSTFLAGS="-C target-feature=+crt-static -C linker=rust-lld -C link-arg=-L/usr/lib"
+    CARGO_TARGET_DIR=/target \
+    cargo build \
+        --frozen \
+        --release \
+        --target "${ARCH}-unknown-linux-musl" \
+        --no-default-features \
+        --features "agent-runtime,channels-full,acp-bridge,gateway,schema-export,observability-prometheus" \
+        -p zeroclawlabs
+    mkdir -p /rootfs/usr/bin
+    cp /target/${ARCH}-unknown-linux-musl/release/zeroclaw /rootfs/usr/bin/zeroclaw
+EOF
+
+# Stage 4: Full-channel runtime image (target: package-fat)
+FROM docker.io/stagex/core-filesystem@sha256:cd3a66471ce1f630fa77d5c9bd9829f9f9fab6302a1aaa64d67b74f1f069b750 AS package-fat
+COPY --from=build-fat /rootfs/ /
+COPY --from=build /rootfs/usr/bin/zerocode /usr/bin/zerocode
 COPY --from=docker.io/stagex/core-ca-certificates@sha256:7773dae6630aa3bdcc82cfec6c9265c0c501aaf0af67cc73631b09e1cff1b094 / /
 ENTRYPOINT ["/usr/bin/zeroclaw"]
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,14 @@
 IMAGE_NAME    = zeroclaw
 IMAGE_TAG     = stagex
+IMAGE_FAT_TAG = stagex-fat
 
-.PHONY: build extract shell-debug clean
+.PHONY: build build-fat extract extract-fat shell-debug clean
 
 build:
-	podman build -t $(IMAGE_NAME):$(IMAGE_TAG) -f Containerfile .
+	podman build -t $(IMAGE_NAME):$(IMAGE_TAG) --target package -f Containerfile .
+
+build-fat:
+	podman build -t $(IMAGE_NAME):$(IMAGE_FAT_TAG) --target package-fat -f Containerfile .
 
 extract:
 	@if ! podman image exists $(IMAGE_NAME):$(IMAGE_TAG) 2>/dev/null; then \
@@ -12,8 +16,20 @@ extract:
 	fi
 	podman create --name zeroclaw-extract $(IMAGE_NAME):$(IMAGE_TAG)
 	podman cp zeroclaw-extract:/usr/bin/zeroclaw .
+	podman cp zeroclaw-extract:/usr/bin/zerocode .
 	podman rm zeroclaw-extract
-	ls -lh zeroclaw
+	ls -lh zeroclaw zerocode
+
+extract-fat:
+	@if ! podman image exists $(IMAGE_NAME):$(IMAGE_FAT_TAG) 2>/dev/null; then \
+		$(MAKE) build-fat; \
+	fi
+	podman create --name zeroclaw-fat-extract $(IMAGE_NAME):$(IMAGE_FAT_TAG)
+	podman cp zeroclaw-fat-extract:/usr/bin/zeroclaw .
+	podman cp zeroclaw-fat-extract:/usr/bin/zerocode .
+	podman rm zeroclaw-fat-extract
+	mv zeroclaw zeroclaw-fat
+	ls -lh zeroclaw-fat
 
 shell-debug:
 	podman run --rm -it \
@@ -21,5 +37,5 @@ shell-debug:
 		docker.io/stagex/pallet-rust@sha256:2d90b9552412ee2c4fa2a13b489c2f28c044be7fb5d6a942bfd5a480a5c288fd
 
 clean:
-	-podman rmi $(IMAGE_NAME):$(IMAGE_TAG) 2>/dev/null
-	rm -f zeroclaw
+	-podman rmi $(IMAGE_NAME):$(IMAGE_TAG) $(IMAGE_NAME):$(IMAGE_FAT_TAG) 2>/dev/null
+	rm -f zeroclaw zerocode zeroclaw-fat
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,25 @@
+IMAGE_NAME    = zeroclaw
+IMAGE_TAG     = stagex
+
+.PHONY: build extract shell-debug clean
+
+build:
+	podman build -t $(IMAGE_NAME):$(IMAGE_TAG) -f Containerfile .
+
+extract:
+	@if ! podman image exists $(IMAGE_NAME):$(IMAGE_TAG) 2>/dev/null; then \
+		$(MAKE) build; \
+	fi
+	podman create --name zeroclaw-extract $(IMAGE_NAME):$(IMAGE_TAG)
+	podman cp zeroclaw-extract:/usr/bin/zeroclaw .
+	podman rm zeroclaw-extract
+	ls -lh zeroclaw
+
+shell-debug:
+	podman run --rm -it \
+		--entrypoint /bin/sh \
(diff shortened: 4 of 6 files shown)


singlerider (changes requested):
Thanks for this, @ConYel — a StageX, fully-pinned, musl-static reproducible build pipeline is exactly the kind of supply-chain hardening the project wants, and the digest-pinned stages plus `--network=none` build step are done right. As a former StageX maintainer I'm glad to see the canonical pallet-rust → core-filesystem idiom used faithfully here. A few items below.

### 🔴 Blocking — the feature set drops all channels and omits `zerocode`

The release build runs with `--no-default-features --features "agent-runtime,acp-bridge,gateway,schema-export,observability-prometheus"`. That deliberately drops `default-channels`, so the resulting image ships with **zero messaging channels** — none of the four defaults (`channel-acp-server`, `channel-webhook`, `channel-email`, `channel-telegram`) are present. A daemon image that can't talk on any channel is not a usable default for most operators.

Two concrete asks:

1. **Include the default channel set** (or a deliberately chosen subset) rather than silently building none. The crate's `default` feature is `["agent-runtime", "default-channels", "acp-bridge", "gateway", "observability-prometheus", "schema-export"]` — the only thing this build omits from the defaults is `default-channels`, and that omission should be intentional and documented, not a side effect. If the intent is a minimal core image, say so and note how operators opt channels back in.
2. **Be explicit that this is the minimal floor, not the full surface.** zeroclaw has 32 `channel-*` features plus `channels-full`, the `ci-all` meta-feature, hardware/peripheral, sandbox (`sandbox-landlock`, `sandbox-bubblewrap`), `browser-native`, `plugins-wasm`, `rag-pdf`, and the observability stack. Worth a comment in the Containerfile (or a documented `ARG`/build matrix) so downstream consumers know which surfaces are compiled in and how to extend the feature list for their deployment.

Note specifically: **Matrix is not a default feature.** `channel-matrix` is opt-in (it's only in `ci-all`/`channels-full`, never in `default-channels`), so any operator who wants Matrix must add `channel-matrix` to the feature list explicitly — this build would not give it to them even if `default-channels` were restored.

### 🔴 Blocking — the image ships the daemon but not `zerocode`

The build does `cargo build … -p zeroclawlabs` and copies only `/usr/bin/zeroclaw` into the rootfs. The interactive TUI config manager (`apps/zerocode`, binary `zerocode`) is never built or shipped. For a from-scratch container the daemon binary alone leaves operators with no in-image way to inspect or repair config — `zerocode` is the intended companion surface and it builds clean under musl-static (it's a ratatui/crossterm TUI with no GUI/X11/wgpu deps). Please build and ship `zerocode` alongside the daemon:

- add `-p zerocode` to the release build (or a second `cargo build … -p zerocode`), and
- `cp /target/${ARCH}-unknown-linux-musl/release/zerocode /rootfs/usr/bin/zerocode`.

Excluding `apps/tauri` from the workspace `sed` is correct — the desktop shell has no place in a headless static container. Keep that exclusion; just add `zerocode` to what gets packaged.

### 🔴 Blocking — workspace `sed` mutation is brittle and silently scope-creeping

The Stage-1 `sed -i '/"apps\/tauri"/d; /"tools\/fill-translations"/d; /"xtask"/d' Cargo.toml` plus `sed -i '/^\[\[test\]\]$/,/^$/d' Cargo.toml` edits the committed manifest in-place at build time. Two problems:

1. It's positional and will silently no-op (not fail) if the `members` formatting changes — e.g. a reflow to one-member-per-line with trailing commas, or a comment on the line. A build that silently stops excluding `apps/tauri` would then try to pull the desktop shell into a musl-static build and fail confusingly, or worse, succeed with unintended scope. Prefer excluding at the cargo level (`--exclude` on a workspace build, or a dedicated workspace member set) over regex-editing the manifest.
2. The `[[test]]` block deletion is unexplained and overbroad — it strips test targets from the root manifest by blank-line-delimited range, which is fragile against manifest reformatting and obscures *why* tests are being removed for a release build. If the goal is "don't compile integration tests in the release image," `--release` already skips them; document the precise reason or drop the deletion.

### 🟢 Pinning and reproducibility are exemplary

StageX's default is rolling (latest deps on the fly); bit-for-bit reproducibility is opt-in by locking each stage to a tag or digest. This PR takes that opt-in path correctly — every StageX stage is pinned by `sha256:` digest (build pallet, core-filesystem, core-ca-certificates), the dependency build step runs `--network=none` with `--frozen`, and the rustls/tokio-rustls/rumqttc/reqwest feature sets are consistently narrowed to `ring` + no-default-provider, coherent with a static, provider-explicit TLS stack. This is the right way to get a hermetic, reproducible from-scratch image out of StageX.

### 🟢 Tauri correctly excluded

Stripping `apps/tauri` from the workspace before the musl-static build is the correct call — the desktop shell does not belong in a headless static image. Good instinct.

### 🟢 Makefile ergonomics

`build` / `extract` / `shell-debug` / `clean` with podman, and the `extract` target auto-building when the image is absent, make this easy to drive locally. Clean and minimal.

singlerider:
See also: #6293

For Caution integration in the future.

WareWolf-MoonWall (dismissed):
I reviewed #7176 at head 283ce157, including the PR body, diff (Cargo.toml changes, new Containerfile, new Makefile), and @singlerider's June 3 23:38 `CHANGES_REQUESTED` review. The StageX approach is sound (digest-pinned stages, `--network=none` build, musl-static output), but the feature set and tooling shipped in the image need adjustment before this is merge-ready.

### 🔴 Blocking — Feature set silently drops all messaging channels

The release build line 37-39 uses `--no-default-features --features "agent-runtime,acp-bridge,gateway,schema-export,observability-prometheus"`. This explicitly omits `default-channels`, so the resulting image ships with **zero messaging channels** — not even the four defaults (ACP server, webhook, email, Telegram).

Per the crate's `Cargo.toml`, the `default` feature is:
```
default = ["agent-runtime", "default-channels", "acp-bridge", "gateway", "observability-prometheus", "schema-export"]
```

The only difference between that and the build command is the missing `default-channels`. This makes the image unusable for most operators — a daemon that can't talk on any channel can't serve user requests.

**What needs to change:**

1. **Add `default-channels` to the feature list** on line 38, or document explicitly that this is a minimal-core image and note how operators extend it (custom Containerfile with additional `--features`, or a documented ARG-based build matrix).

2. **If the minimal feature set is intentional**, add a comment above the cargo build explaining:
   ```dockerfile
   # Minimal core build: no channels compiled in. Operators can extend by
   # forking this Containerfile and adding channel features to line 38.
   # Available: channel-{telegram,discord,slack,...}, channels-full.
   ```

Without either change, operators will build the image, deploy it, and discover their agent can't communicate. That's a silent failure that wastes deploy cycles.

**Note:** `channel-matrix` and most social channels are opt-in (not in `default-channels`), so even with `default-channels` restored, operators wanting Matrix/Facebook/WhatsApp/etc. must explicitly add those features. The PR body should call this out.

### 🔴 Blocking — Image omits `zerocode` TUI tool

The build compiles only `-p zeroclawlabs` (line 39) and copies only `/usr/bin/zeroclaw` into the runtime image (line 41). The `zerocode` binary (`apps/zerocode`, the interactive TUI config manager) is never built or shipped.

For a minimal from-scratch container, shipping the daemon without `zerocode` leaves operators with no in-image way to inspect or repair config. `zerocode` is the intended companion tool for config validation, quickstart flows, and interactive editing. It builds clean under musl-static (it's a ratatui/crossterm TUI, no GUI/X11 deps).

**What needs to change:**

Add a second build invocation (or extend the existing one) to compile `zerocode`, then copy it to the runtime image:

```dockerfile
# After line 39, before the mkdir:
cargo build \
  --frozen \
  --release \
  --target "${ARCH}-unknown-linux-musl" \
  --no-default-features \
  --features "tui" \
  -p zerocode

# After line 41:
cp /target/${ARCH}-unknown-linux-musl/release/zerocode /rootfs/usr/bin/zerocode
```

Excluding `apps/tauri` is correct (the desktop shell doesn't belong in a headless container). Excluding `zerocode` is not.

### 🔴 Blocking — Workspace `sed` mutations are brittle and undocumented

Lines 9-11 edit the committed `Cargo.toml` in-place:
```dockerfile
sed -i '/"apps\/tauri"/d; /"tools\/fill-translations"/d; /"xtask"/d' Cargo.toml
sed -i '/^\[\[test\]\]$/,/^$/d' Cargo.toml
```

Two problems:

1. **The `members` edit is positional and will silently no-op if formatting changes.** If the workspace `members` array is ever reflowed to one-per-line with trailing commas, or a comment is added on the line, the regex won't match and `apps/tauri` will silently stay in the build. A musl-static tauri build will then fail (tauri pulls X11/wayland/wgpu), or worse, succeed with unintended scope if those deps become conditionally optional. Prefer `cargo build --workspace --exclude apps/tauri --exclude tools/fill-translations --exclude xtask` over regex-editing the manifest.

2. **The `[[test]]` deletion is unexplained.** Line 11 strips test targets by blank-line-delimited range, which is fragile and obscures *why* tests are being removed. If the goal is "don't compile integration tests in the release image," `--release` already skips them. If there's a specific test target causing issues, document it in a comment. Otherwise, drop the deletion.

**What needs to change:**

Replace the `sed` mutations with cargo-level exclusions:
```dockerfile
# Line 9-11 replacement:
cargo fetch --locked
```

Then on line 39, use:
```bash
cargo build \
  --frozen \
  --release \
  --target "${ARCH}-unknown-linux-musl" \
  --workspace \
  --exclude apps/tauri \
  --exclude tools/fill-translations \
  --exclude xtask \
  --no-default-features \
  --features "agent-runtime,default-channels,acp-bridge,gateway,schema-export,observability-prometheus"
```

This is explicit, failure-loud (unknown package errors immediately), and won't silently no-op on format changes.

### 🟢 What looks good — Digest-pinned stages and `--network=none` build

Every StageX stage is pinned by `sha256:` digest:
- `stagex/pallet-rust@sha256:2d90b955...` (build stage)
- `stagex/core-filesystem@sha256:cd3a6647...` (runtime)
- `stagex/core-ca-certificates@sha256:7773dae6...` (CA bundle)

The `cargo fetch` step runs with network access, then the entire build RUN (lines 15-42) runs under `--network=none` with `--frozen`. This guarantees:
- All crates come from the prior fetch (no registry access during build)
- Supply-chain integrity: even if a registry or git dep goes rogue between fetch and build, the build cannot reach it

This is the right way to get a hermetic, reproducible musl-static image. The PR body documents the rationale clearly.

### 🟢 What looks good — `aws-lc-sys` elimination is clean

The Cargo.toml changes (rustls/tokio-rustls/rumqttc) disable `default-features` and explicitly re-enable `ring` as the crypto provider. `cargo tree -i aws-lc-sys` returns zero matches (verified in PR body), so the cmake transitive dep is eliminated. The ring provider preserves full TLS functionality with no behavioral change.

This is the minimal change needed to build under StageX's cmake-free pallet-rust base.

### 🟢 What looks good — Makefile is clean and minimal

`build` / `extract` / `shell-debug` / `clean` targets with podman, and `extract` auto-building when the image is absent, make this easy to drive locally. The `shell-debug` target drops into the build stage's shell for debugging, which is useful for diagnosing build failures.

### 🟡 Warning — PR body claims "31 MB" but CI shows size verification not yet run

The PR body states the final image is 31.3 MB, but the CI checks show 2/16 failing. Without seeing which checks failed, I can't verify the image size claim. If the size is load-bearing for the "minimal distroless" pitch, include the `podman images` output in the validation evidence showing the final layer size breakdown.

Not blocking — the approach is sound regardless of exact size — but worth confirming before merge.

### 🔵 Suggestion — Consider a `FEATURES` build ARG for operator extensibility

The hard-coded feature set on line 38 makes the image inflexible for operators who want a different surface (e.g., `channels-full`, `browser-native`, `sandbox-landlock`). A build ARG would let operators extend without forking:

```dockerfile
ARG FEATURES="agent-runtime,default-channels,acp-bridge,gateway,schema-export,observability-prometheus"
RUN cargo build \
  --no-default-features \
  --features "${FEATURES}" \
  ...
```

Then operators can `podman build --build-arg FEATURES="agent-runtime,channels-full,browser-native" ...` to customize. Not blocking — the hard-coded set is fine for an initial merge — but worth considering for v2.

### 🔵 Suggestion — Add a note about Matrix and social channels being opt-in

The PR body explains the StageX choice and the `aws-lc-sys` removal, but doesn't call out that `channel-matrix`, `channel-facebook-messenger`, `channel-whatsapp`, etc. are opt-in. An operator building this image and expecting Matrix support will be surprised. Adding a note to the README or Containerfile comment would help:

```dockerfile
# Social/Matrix channels are opt-in: add channel-matrix, channel-facebook-messenger,
# channel-whatsapp, etc. to FEATURES if needed. See Cargo.toml for the full list.
```

### Verdict

The StageX approach is the right one for a minimal, reproducible, supply-chain-hardened image. The digest pinning, `--network=none` build, and `aws-lc-sys` elimination are all correct. The three blocking issues are:

1. Missing `default-channels` → image has zero messaging channels
2. Missing `zerocode` binary → no in-image config tooling
3. Brittle workspace `sed` → will silently break on format changes

Fix those three and this is good to merge.

ConYel:
is it better now? and more suggestions?

WareWolf-MoonWall (commented):
Re-reviewing at 0415344 after the author addressed all three blocking items from my initial review at 283ce15.

## Summary

The fixup commit resolves every blocker I raised:

1. **✅ default-channels restored** — the release build now includes `--features "…,default-channels,…"`, shipping the four standard messaging surfaces (acp-server, webhook, email, telegram).
2. **✅ zerocode shipped** — both Containerfile and Makefile now build and package the TUI config manager alongside the daemon. Operators get an in-image management surface.
3. **✅ workspace sed eliminated** — the fragile manifest-editing sed calls are gone. The build copies the source tree as-is and lets cargo resolve the workspace naturally. Multi-stage separation (build → build-fat) means the full workspace can live in Stage 1 without scope drift.

The two-target container design (default + fat) is a thoughtful addition — `package` ships default channels (85% of deployments), `package-fat` ships `channels-full` (all 32), and operators choose via `--target`. This is exactly the kind of composable deployment surface the project wants.

CI shows two failures (Test + CI Required Gate). The Test failure appears unrelated to the container build — nothing in this PR touches runtime behavior, test fixtures, or production code paths. I'm assuming the failure is a flake or unrelated breakage on master; will note this for the author to confirm.

### ✅ Resolved — default-channels included

The Stage-1 release build now carries `--features "agent-runtime,default-channels,…"`, restoring the four canonical messaging channels (`channel-acp-server`, `channel-webhook`, `channel-email`, `channel-telegram`). The minimal image is now a usable default for most operators, and the fat image (`channels-full`) serves the extended deployment case. Good call.

### ✅ Resolved — zerocode shipped alongside daemon

Both the daemon (`zeroclaw`) and the TUI config manager (`zerocode`) are built in Stage 1 and copied into the rootfs. The `package-fat` stage reuses `zerocode` from Stage 1 via `COPY --from=build`, avoiding redundant compilation. Clean.

### ✅ Resolved — workspace sed removed, multi-stage design

The brittle sed mutations are gone. The full source tree is copied into Stage 1 (`COPY . .`), and cargo resolves the workspace naturally. The `build-fat` stage reuses `/src` from Stage 1 (`COPY --from=build /src /src`), so there's no scope drift risk and no second COPY of the source tree. The new two-target design (`package` / `package-fat`) makes the feature split explicit and composable.

### 🟢 Multi-target composability

The Containerfile now exports two runtime images via build targets:

- `package` (default): `default-channels` + daemon + TUI, ~31 MB
- `package-fat`: `channels-full` (all 32 channels) + daemon + TUI

Operators choose via `podman build --target <package|package-fat>`, and the Makefile provides `build` / `build-fat` / `extract` / `extract-fat` convenience targets. This is the right level of composability — the common case is lightweight, the extended case is opt-in, and the build system does not require manual feature-list editing.

### 🟡 Warning — CI failure unrelated to container changes?

The Test check is failing, but the diff shows zero changes to `src/`, test fixtures, or runtime behavior — only Cargo.toml (rustls/tokio-rustls/rumqttc feature narrowing), Containerfile, Makefile, and .dockerignore. The cargo feature changes (swapping `aws_lc_rs` → `ring`) are backend-neutral and should not affect test behavior. If the failure is a flake or unrelated breakage on master, note that in the PR thread so reviewers know the container build itself validates clean. If the failure is real and tied to the rustls provider swap, that's a deeper issue and this PR should not merge until root-caused.

### 🟢 Cargo.toml dependency narrowing is correct

The root and runtime Cargo.toml changes (`rustls`/`tokio-rustls`/`rumqttc` with `default-features = false` + explicit `ring` provider) are the right call for a musl-static StageX build. The aws-lc-sys transitive dependency required cmake at build time, which StageX deliberately omits. Swapping to the `ring` crypto backend eliminates the C build dependency while preserving full TLS functionality. This is the canonical pattern for static musl builds using rustls.

### 🟢 .dockerignore hygiene

The .dockerignore cleanup removes the Tauri source-exclusion stanza (replaced by the workspace-level exclusion in the Containerfile) and restores `tests/` inclusion (tests are part of the workspace and needed for `clippy --all-targets`). The new `!apps/zerocode/` inclusion is consistent with the goal of shipping the TUI manager. Clean.

---

**Verdict:** All my blockers are resolved. The two-target design is a thoughtful extension of the original scope. The CI failure needs root-causing — if it's a flake or unrelated to this PR, I'll approve once confirmed. If it's tied to the rustls provider swap, that's a blocking issue and requires investigation before merge.

@ConYel — can you confirm whether the Test failure is unrelated to this PR (flake or master breakage) or if it's tied to the rustls/tokio-rustls/rumqttc feature changes? If unrelated, a quick note in the thread clears this for my approval. If related, we need to root-cause the test failure before this can land.

@singlerider — the fixup also appears to address all three items from your June 3 review (default-channels restored, zerocode shipped, sed removed). When you have a chance, can you confirm?

ConYel:
@WareWolf-MoonWall thank youfor the review, so I made an update on dockerignore, as the StageX container build includes tests/ in the build context for in-container testing. 
I hope the CI works properly now.

Audacity88 (commented):
I rechecked #7176 at current head 04d80ed, including the PR body/comments, current CI, prior reviews, and the changed Containerfile, Makefile, .dockerignore, Cargo manifests, and dockerignore test. I do not see a new blocker from my side, but I am leaving this as a comment because @singlerider's June 3 `CHANGES_REQUESTED` review is still active; @WareWolf-MoonWall already converted their blocker to a comment after re-reviewing the previous head.

### ✅ Resolved — Prior container blockers appear addressed

The current Containerfile now builds the default image with `default-channels`, builds and copies `zerocode` alongside `zeroclaw`, and no longer edits the committed workspace manifest with `sed`. The `package-fat` target also gives operators a `channels-full` image path without changing the default image surface.

### ✅ Resolved — CI is now green

The Test and CI Required Gate failures that @WareWolf-MoonWall called out on the previous re-review are now passing on the current head. All visible checks are green, including Test and CI Required Gate.

### 🟡 Warning — Refresh the PR body before merge

The body still has a few stale bits from earlier revisions, including the old label snapshot (`type: build`, `risk: low`, `size: M`) and the original Containerfile/validation wording. That does not change the code review result, but the body should be refreshed before this is treated as merge-ready so it matches the current `type: ci` / `risk: medium` / `size: XS` labels, the current two-target Containerfile, and the now-green CI state.

@singlerider — from my pass, the three original code/build blockers and the later CI concern look resolved. Could you re-review the current head when you have a chance?

singlerider (approved):
@ConYel — re-reviewed holistically at head `04d80ed`. All three of my June 3 blockers are resolved, the design grew a clean two-target split since then, and CI is fully green. This is good to merge. Verdict below.

### ✅ RESOLVED — `default-channels` restored

The Stage-1 release build now carries `--features "agent-runtime,default-channels,acp-bridge,gateway,schema-export,observability-prometheus"`. The default `package` image ships the four canonical channels (`channel-acp-server`, `channel-webhook`, `channel-email`, `channel-telegram`) and is a usable daemon out of the box. The opt-in note for `channel-matrix` and social channels is now reflected in the body.

### ✅ RESOLVED — `zerocode` shipped

Stage 1 builds `-p zerocode` alongside `-p zeroclawlabs` and copies both into `/rootfs/usr/bin`. `package` and `package-fat` both carry the TUI manager (the fat stage pulls `zerocode` from the `build` stage via `COPY --from=build` rather than recompiling). `apps/zerocode/Cargo.toml` already pins `rustls`/`tokio-tungstenite` to `ring`/webpki-roots, so it links clean under musl-static with no `aws-lc-sys` re-pull. Confirmed against the manifest.

### ✅ RESOLVED — workspace `sed` mutation eliminated

The in-place `Cargo.toml` editing is gone. The source tree is copied as-is (`COPY . .`) and cargo resolves the workspace naturally; `.dockerignore` governs the context. No positional regex, no silent no-op risk. The `tests/component/dockerignore_test.rs` assertion was updated in the same pass (`"tests"` dropped from `MUST_EXCLUDE`) so the test matches the new `.dockerignore` that restores `tests/` for `clippy --all-targets`.

### 🟢 Two-target design is the right shape

`package` (default-channels, ~31 MB) for the common case and `package-fat` (`channels-full`) for the extended surface, selected by `--target`, is exactly the composability the project wants — lightweight default, opt-in breadth, no manual feature-list editing. The fat stage reusing Stage 1's `/src` and `zerocode` avoids a redundant source copy and a redundant TUI compile.

### 🟢 `libstdc++.a` assembly is correct for the StageX toolchain

The build step assembles `libstdc++.a` from StageX's LLVM `libc++.a` + `libc++abi.a` (`ar x` both, `ar rcs` into `/usr/lib/libstdc++.a`) before the release build. That's the right bridge for a `pallet-rust` base that ships LLVM's C++ runtime rather than GCC's, and it's scoped to the build stage — it never enters the runtime image.

### 🟢 Pinning, `--network=none`, and `aws-lc-sys` elimination unchanged and still exemplary

Every StageX stage is digest-pinned; the build `RUN` is hermetic under `--network=none --frozen` after a network-enabled `cargo fetch`; the `rustls`/`tokio-rustls`/`rumqttc` narrowing to `ring` + no-default-provider keeps the TLS stack provider-explicit with `cargo tree -i aws-lc-sys` empty. This is the canonical way to get a reproducible from-scratch musl-static image out of StageX.

---

**Verdict:** All blockers resolved, CI green (Test + CI Required Gate now passing), branch mergeable. Approving. My June 3 `CHANGES_REQUESTED` is hereby converted. The PR body has been refreshed to match the current two-target Containerfile, the `zerocode` shipping, the `.dockerignore`/test changes, and the current `type: ci` / `risk: medium` / `size: XS` labels.

bheatwole:
@ConYel `tests/manual/test_dockerignore.sh` also has a `MUST_EXCLUDE` check which looks at the `.dockerignore` and fails if 'tests' is not excluded. This shell script is run by `./dev/ci.sh all` which is mentioned in `CONTRIBUTING.md`

Recommendation: also update the MUST_EXCLUDE list in `tests/manual/test_dockerignore.sh` to restore the ability of developers to pre-flight the CI locally.

ConYel:
@bheatwole 
you are free to make a PR if you manage to reproduce this work. 
I have several other things currently to focus on before I can contribute more.
Currently the Containerfile needs some addition to produce a fully working zeroclaw daemon.

PR #7177, opened 2026-06-03
fix(rpc): force memory-tool exclusion for ACP sessions server-side

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:** ACP (Code) sessions must never have the long-term-memory tools. The RPC `session/new` handler derived `exclude_memory` only from the wire flag (`req.exclude_memory.unwrap_or(false)`), with no link to `chat_mode` — so a `chat_mode = "acp"` session that omitted the flag (or sent `false`) came up with all five memory tools and a live memory backend. This makes exclusion server-authoritative: `ChatMode::Acp` forces `exclude_memory = true` regardless of the wire value, so a client can no longer smuggle memory access into a Code session.
- Also de-duplicates the memory-tool name list (previously hardcoded in the agent builder, the RPC test, and implied elsewhere) into a single canonical `zeroclaw_tools::MEMORY_TOOL_NAMES`, with a guard test that fails if a memory tool is added/renamed without updating the list. The agent builder and dispatch tests now consume the const.
- **Scope boundary:** Does not change the ACP protocol server (`orchestrator/acp_server.rs`), which already passes `exclude_memory = true` at all callsites. Does not change memory backends, tool behavior, or the non-ACP (Chat) path, which still honors explicit opt-in.
- **Blast radius:** `zeroclaw-runtime` RPC `session/new` and the agent builder; `zeroclaw-tools` public surface gains one `pub const`. No protocol/wire change — `exclude_memory` remains an accepted optional field.
- **Linked issue(s):** None.
- **Labels:** `bug` (auto: `risk:*`, `size:*`, scope labels applied by tooling).

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-tools -p zeroclaw-runtime --all-targets -- -D warnings
cargo test -p zeroclaw-tools memory_tool_names
cargo test -p zeroclaw-runtime --lib rpc::dispatch::tests::acp
cargo test -p zeroclaw-runtime --lib rpc::dispatch::tests::non_acp
```

```
$ cargo fmt --all -- --check
(exit 0, no diff)

$ cargo clippy -p zeroclaw-tools -p zeroclaw-runtime --all-targets -- -D warnings
    Checking zeroclaw-runtime v0.8.0-beta-2
(exit 0, no warnings)

$ cargo test -p zeroclaw-tools memory_tool_names
test memory_tool_names_guard::memory_tool_names_match_tools ... ok

$ cargo test -p zeroclaw-runtime --lib rpc::dispatch::tests::acp
test rpc::dispatch::tests::acp_chat_mode_strips_memory_tools_without_exclude_flag ... ok
test rpc::dispatch::tests::acp_session_new_exposes_no_memory_tools ... ok
test rpc::dispatch::tests::acp_session_new_writes_to_acp_store_only ... ok

$ cargo test -p zeroclaw-runtime --lib rpc::dispatch::tests::non_acp
test rpc::dispatch::tests::non_acp_session_new_exposes_memory_tools ... ok
```

- **Beyond CI — what did you manually verify?** Traced the zerocode client paths: the ACP pane calls `session_new_acp` (sends `chat_mode: "acp"`); the new server logic strips memory tools even if that flag were dropped. Confirmed the ACP protocol server callsites already force exclusion.
- **If any command was intentionally skipped, why:** Ran scoped `clippy`/`test` for the two touched crates rather than the full workspace battery; full `cargo build --workspace` was green during development.

#### 11. ACP memory-tool exclusion (#7177) — Tier 3

Commit `4852f9360`.

- [x] ACP session → memory tools excluded server-side even if the client requests them — PASS (unit, end-to-end through dispatcher). `acp_chat_mode_strips_memory_tools_without_exclude_flag` (dispatch.rs:3411) sends `chat_mode:"acp"` with NO `exclude_memory` flag through `handle_session_new_for_test`, pulls the live agent from the session store, asserts every name in the canonical `MEMORY_TOOL_NAMES` is absent. Server derives exclusion from `ChatMode::Acp`, not the client wire flag — a client cannot smuggle memory access into a Code session.
- [x] Non-ACP sessions still get memory tools — PASS (unit). `non_acp_session_new_exposes_memory_tools` (dispatch.rs:3449) sends `exclude_memory:false` (no acp mode), asserts at least one memory tool present on the live agent. Opt-in preserved for non-ACP callers. Canonical list guarded by `memory_tool_names_match_tools` (tools/lib.rs:102 — fails if a memory tool is added/renamed without updating the const; satisfies the no-hardcoded-list rule). All 3 tests green.
- [x] **`5a3086df7`** ACP **rehydrate** path also excludes memory tools — `rehydrate_reaped_session` rebuilt a reaped ACP session with `exclude_memory=false`; since rehydrate reinserts as `ChatMode::Acp` the recovered agent restored the memory backend/tools. Now constructs with `exclude_memory=true`. Test `reaped_acp_session_rehydrates_without_memory_tools` reaps→rehydrates→asserts no `MEMORY_TOOL_NAMES`. Verify a reaped-then-resumed Code session has no memory tools.

- If any `Yes`, describe the risk and mitigation: This change *removes* capability (memory tools/backend) from ACP sessions and closes a client-trusted gap; it only narrows access.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: `exclude_memory` remains an accepted optional RPC field; ACP sessions now ignore a permissive value and always exclude. No upgrade steps for existing users.

- **Observable failure symptoms:** ACP/Code sessions unexpectedly missing memory tools is the intended state; regression would be an ACP session exposing `memory_*` tools (covered by `acp_chat_mode_strips_memory_tools_without_exclude_flag`).


WareWolf-MoonWall (approved):
## Summary

Reviewed at head commit 4852f936 (1 commit). This PR closes a server-trust gap: ACP (Code) sessions must never expose the long-term-memory tools, but the RPC `session/new` handler derived `exclude_memory` solely from the wire flag (`req.exclude_memory.unwrap_or(false)`) with no link to `chat_mode`. A Code session that omitted the flag or sent `exclude_memory = false` came up with all five memory tools and a live memory backend. The fix makes exclusion server-authoritative: `ChatMode::Acp` forces `exclude_memory = true` regardless of the wire value. Also de-duplicates the memory-tool name list into a single canonical `MEMORY_TOOL_NAMES` const with a guard test. PR body is thorough; validation evidence is complete; CI is green (all 13 checks successful).

---

### 🟢 What looks good — Server-authoritative security policy

This is exactly the right fix: **derive capability from server-known session mode, not from client-provided flags.** The old path trusted the wire `exclude_memory` flag; a buggy or malicious client could smuggle memory access into a Code session. The new path derives exclusion from `chat_mode` before the agent builder runs (line ~688 in `dispatch.rs`: `let exclude_memory = matches!(chat_mode, ChatMode::Acp) || req.exclude_memory == Some(true);`), so the server decides what tools the session gets based on the session's mode, not on what the client claims it wants.

The logic is correct: ACP sessions always exclude memory (`matches!(chat_mode, ChatMode::Acp)`), and non-ACP sessions still honor explicit opt-in (`req.exclude_memory == Some(true)`). The test `acp_chat_mode_strips_memory_tools_without_exclude_flag` proves the server strips memory tools even when the wire request omits the flag entirely, and `non_acp_session_new_exposes_memory_tools` confirms opt-in is preserved for non-ACP sessions. This is the right level of test coverage: one test for the new invariant, one test for the unchanged path.

The de-duplication of the memory-tool name list is a high-value cleanup: the list was previously hardcoded in three places (agent builder, RPC test, and implied elsewhere). The new `MEMORY_TOOL_NAMES` const in `zeroclaw-tools/src/lib.rs` is the single source of truth, with a guard test (`memory_tool_names_match_tools`) that fails if a memory tool is added or renamed without updating the const. This is exactly how to prevent drift: one canonical list, one test that enforces it, and all consumers reference the const. The guard test constructs each memory tool and collects their names into a `BTreeSet`, then compares against the const — if a memory tool is added (e.g., `memory_summarize`) or renamed (e.g., `memory_recall` → `memory_search`), the guard test fails and forces the author to update the const. This is the kind of low-friction defensive code that prevents entire classes of bugs.

---

### 🟢 What looks good — Minimal diff with clear attribution

The commit message is exemplary: it names the problem ("ACP sessions must never have the long-term-memory tools"), explains the gap ("The RPC `session/new` handler derived `exclude_memory` solely from the wire flag... with no link to `chat_mode`"), describes the fix ("Derive exclusion from `chat_mode` on the server"), and documents the side benefit ("Also de-duplicate the memory-tool name list"). The message also notes that "The zerocode ACP pane always sent the flag, but the guarantee was client-trusted rather than server-enforced" — this is the kind of context that helps future readers understand why the change was necessary even though the current zerocode client already does the right thing. The fix is not about patching a known exploit; it's about closing a trust boundary that should never have been open.

The PR body's "Security & Privacy Impact" section explicitly states: "This change *removes* capability (memory tools/backend) from ACP sessions and closes a client-trusted gap; it only narrows access." This is the right framing: the change is a security tightening, not a feature addition, and the author documented it proactively.

---

### 🔵 Suggestion — Consider a similar audit for other session-mode-specific constraints

ACP sessions have at least one other mode-specific constraint: they write to the ACP store only (per the test `acp_session_new_writes_to_acp_store_only` at line ~3366 in `dispatch.rs`, visible in the existing test suite). If there are other capabilities or backends that should be mode-gated (e.g., cron access, gateway tokens, specific tool categories), consider auditing the session-creation path to ensure those are also derived from server-known session mode rather than trusted from wire flags. Not blocking — this PR closes the specific memory-tool gap and does it correctly — but the pattern here (client-provided flag trusted instead of server-derived mode) might exist elsewhere, and a systematic audit would close those gaps preemptively.

---

### 🔵 Suggestion — Document the ACP-session capability model

The PR closes the memory-tool gap, but the broader ACP-session capability model (what tools/backends/permissions a Code session gets, and why) is not documented in operator-facing or developer-facing docs as far as I can see from the diff. If there's a `docs/book/src/reference/session-modes.md` or `docs/book/src/developing/acp-protocol.md`, adding a section that explicitly lists the ACP-session constraints (no memory tools, ACP store only, any other mode-specific rules) would help developers understand the trust boundary when working on session-creation or tool-access paths. Not blocking — the code is self-documenting and the tests pin the invariant — but the absence of this doc has likely contributed to the gap existing in the first place (if the constraint were written down, the original author might have noticed the wire-flag trust issue during implementation).

---

### 🔵 Suggestion — Consider a `#[cfg(test)]`-gated `fn assert_acp_invariants(agent: &Agent)` helper

The test `acp_chat_mode_strips_memory_tools_without_exclude_flag` manually iterates over `MEMORY_TOOL_NAMES` and asserts each is absent from the agent's tool list. If ACP sessions gain more invariants over time (e.g., "must not have cron tools", "must not have web-search"), the test will need to repeat the same iteration + assertion pattern for each new invariant. Consider extracting a `#[cfg(test)]`-gated helper in `zeroclaw-runtime/src/agent/agent.rs` or `zeroclaw-runtime/src/rpc/dispatch.rs`:

```rust
#[cfg(test)]
pub(crate) fn assert_acp_invariants(agent: &Agent) {
    let tool_names = agent.tool_names();
    for &mem_tool in zeroclaw_tools::MEMORY_TOOL_NAMES {
        assert!(!tool_names.contains(&mem_tool), "ACP session must not expose {mem_tool}");
    }
    // Future: assert no cron, no web-search, etc.
}
```

Then the test becomes `assert_acp_invariants(&agent);` and future ACP-constraint additions only need to update the helper, not every test. Not blocking — the current test is clear and sufficient — but this would scale better if ACP sessions accumulate more mode-specific constraints.

---

## Verdict

**Approve.** The fix is correct, the test coverage is solid, the de-duplication of the memory-tool name list is a high-value cleanup, and the PR body clearly documents the security tightening. The three suggestions above are all optional follow-ups that would improve discoverability and scalability; none are blocking. This is ready to merge.

CI status at review: all 13 checks successful.

tidux:
@WareWolf-MoonWall the ACP session limitations with respect to memory, tools, workspace, etc., are already in the docs.  https://docs.zeroclawlabs.ai/master/en/channels/acp.html.

Audacity88 (dismissed):
I checked the live PR state at head `4852f936`, prior review/comment snapshots, public check state, the PR body, the public diff, and public PR-head source around ACP rehydration. The fresh `session/new` fix looks right, and I did not use local/private validation evidence. The blocker is that the PR's core security claim is that ACP/Code sessions must never have long-term-memory tools, but one server-side ACP recovery path still violates that invariant.

### 🟢 What looks good — Fresh ACP session creation now enforces the memory boundary

The main `session/new` change is the right direction: deriving `exclude_memory` from `ChatMode::Acp` server-side closes the client-trusted flag gap for fresh Code sessions, and centralizing the memory tool names behind `MEMORY_TOOL_NAMES` makes future drift less likely.

### 🔴 Blocking — ACP rehydrate still restores memory tools

`session/new` now correctly derives memory exclusion from `chat_mode`, but the rehydrate path still rebuilds an ACP session with `exclude_memory = false` (`crates/zeroclaw-runtime/src/rpc/dispatch.rs:922`). When an ACP session is removed from the live session map and the next `session/prompt` calls `rehydrate_reaped_session`, the restored agent is inserted as `ChatMode::Acp` while `Agent::from_config_with_tui_env` was allowed to keep the long-term-memory backend/tools. So although this PR closes the initial `session/new` trust gap, the claimed server-side invariant does not yet survive session recovery.

Please make the ACP rehydrate path construct the agent with memory excluded as well, and add or extend rehydrate coverage to assert that the recovered agent has none of `MEMORY_TOOL_NAMES`. That keeps the server-side invariant true across both fresh ACP sessions and recovered ACP sessions.

singlerider:
Addressed the blocker on `60b143564`:

- **🔴 ACP rehydrate restored memory tools** — `rehydrate_reaped_session` built the recovered agent with `exclude_memory = false`. Since the rehydrate path always reinserts the session as `ChatMode::Acp`, the recovered agent kept the long-term-memory backend/tools the ACP invariant forbids. The rehydrate construction now passes `exclude_memory = true`, so recovered ACP sessions enforce the same server-side boundary as fresh `session/new` ACP sessions. Added `reaped_acp_session_rehydrates_without_memory_tools`, which reaps a live ACP session, rehydrates it, and asserts the recovered agent exposes none of `MEMORY_TOOL_NAMES`.

The server-side invariant now holds across both fresh ACP sessions and recovered ACP sessions.

Validation: `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings`, and the targeted rehydrate tests all green.

Re-requesting review.

PR #7180, opened 2026-06-03
fix(providers): honour wire_api = "responses" for custom and openai-compatible families

singlerider:

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - A delegate target whose model provider declared `wire_api = "responses"` (e.g. a self-hosted vLLM endpoint configured at `[providers.models.custom.vllm]`) was built as a plain chat_completions provider. Requests to a responses-only endpoint then failed, and the delegate refused to run the agent, surfacing a provider error. Risk profiles matched — the escalation guard was not the cause.
  - **Provider factory:** only the `openai` and `llamacpp` families routed to `OpenAiResponsesModelProvider` when `wire_api == Responses`. `CustomModelProviderConfig` and the generic openai-compatible `ModelProviderConfig` ignored `wire_api` entirely and always built a chat_completions provider via `apply_compat_options`. This affected the **primary agent path** for those families too, not only delegation. Both now route to the responses provider when `wire_api == Responses`, mirroring the existing `openai`/`llamacpp` impls.
  - **Delegate path:** `DelegateTool` built the target provider through the bare `create_model_provider_with_options(type, ..)` factory with no config/alias context (`config = None`, `alias = "default"`), so even after the factory fix the target's typed `custom.vllm` entry — and its `wire_api` — was never read. It now routes through `create_model_provider_for_alias` with per-target alias-resolved runtime options, falling back to the bare path for legacy constructors without a root config.
- **Scope boundary:** Does NOT change the 18 branded provider families with fixed wire protocols (mistral, deepseek, groq, etc.) — only the two bring-your-own-endpoint families where `wire_api` is meaningful and a base URL is present. Does NOT change the risk-profile/escalation delegation gate, the `WireApi` enum, or any wire-protocol implementation.
- **Blast radius:** Provider construction for the `custom` and generic openai-compatible families, on both the primary agent path and the delegate path. Configs that did not set `wire_api = "responses"` are unaffected (still chat_completions). No change to providers that already honored `wire_api` (openai, llamacpp).
- **Linked issue(s):** None.
- **Labels:** `bug`, `risk: medium`, `size: S`, `provider`

```bash
cargo fmt --all -- --check
cargo test
```

  - `cargo fmt --all -- --check` → clean, exit 0 (no diff).
  - `cargo clippy --all-targets -- -D warnings` → `Finished dev profile`, exit 0, no warnings.
  - `cargo test` → exit 0, zero failures. Representative tails:
    ```
    test result: ok. 188 passed; 0 failed; 0 ignored; ...
    test result: ok. 159 passed; 0 failed; 0 ignored; ...
    ```
  - New tests (run explicitly):
    ```
    test factory::tests::custom_factory_routes_to_responses_provider_when_wire_api_responses ... ok
    test factory::tests::custom_factory_defaults_to_chat_completions_without_wire_api ... ok
    test factory::tests::openai_compatible_factory_routes_to_responses_when_wire_api_responses ... ok
    test tools::delegate::tests::delegate_builds_target_provider_with_its_declared_wire_api ... ok
    ```
  - **Beyond CI — what I manually verified:** Reproduced the bug at the source — pre-fix, `CustomModelProviderConfig::create_provider` with `wire_api = responses` yielded `default_wire_api() == "chat_completions"`. Confirmed both fixes are load-bearing: reverting the factory fix makes the delegate end-to-end test (`delegate_builds_target_provider_with_its_declared_wire_api`) fail, so it guards the wiring rather than passing as a tautology. Did NOT exercise a live vLLM endpoint over the network — the tests assert the constructed provider's wire protocol (`default_wire_api`), not a live round-trip.

#### 10. wire_api = "responses" for custom + openai-compatible families (#7180) — Tier 3

Commit `2886ce065`.

- [x] `wire_api = "responses"` honored for custom AND openai-compatible provider families (request goes to the Responses wire, not chat-completions). — PASS (unit). Two load-bearing gaps each covered: (1) FACTORY — `custom_factory_routes_to_responses_provider_when_wire_api_responses` + `openai_compatible_factory_routes_to_responses_when_wire_api_responses` assert `default_wire_api()=="responses"`; partners `custom_factory_defaults_to_chat_completions_without_wire_api` + `..._defaults_to_chat_completions_without_wire_api` prove no-flag defaults to chat-completions (5 factory tests green). (2) DELEGATE WIRING — `delegate_builds_target_provider_with_its_declared_wire_api` (delegate.rs:3893) builds a `custom.vllm` target via the alias-aware path → responses provider; regression guard asserts the bare factory (pre-fix, no config/alias) does NOT yield responses, proving the alias path is load-bearing (1 test green, `cargo test -p zeroclaw-runtime`).

- New external network calls? **No** — same endpoints, corrected wire protocol.
- Secrets / tokens / credentials handling changed? **No** — credential resolution and the responses provider's auth are unchanged.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — test fixtures use a private-range placeholder URL.

## Compatibility (required)

- Backward compatible? **Yes** — configs without `wire_api = "responses"` build exactly as before. The change only affects custom/openai-compatible providers that explicitly opted into the responses wire.
- Config / env / CLI surface changed? **No** — no new keys, env vars, or flags; this honors an existing config field (`wire_api`) for two more families.
- Upgrade steps: None. Operators who set `wire_api = "responses"` on a `custom`/openai-compatible provider will now actually get the responses wire (previously silently ignored).

## Rollback

- **Fast rollback command/path:** `git revert <squash sha>` — restores the prior factory routing and delegate build path.
- **Observable failure symptoms:** Custom/openai-compatible providers declaring `wire_api = "responses"` revert to chat_completions; a responses-only endpoint then returns provider errors at request time (grep logs for the provider error surfaced by the failing endpoint, or delegate refusing with a provider error).

### DIFF
--- a/crates/zeroclaw-config/src/provider_aliases.rs
+++ b/crates/zeroclaw-config/src/provider_aliases.rs
@@ -126,3 +126,15 @@ pub fn canonical_china_provider_name(name: &str) -> Option<&'static str> {
         None
     }
 }
+
+/// Whether a canonical provider family honors the `wire_api` config field.
+///
+/// Mirrors the provider factory in `zeroclaw-providers`: only the
+/// bring-your-own-endpoint families build either a chat-completions or a
+/// responses provider from `wire_api`. Every branded vendor family has a
+/// fixed wire protocol and ignores the field. Kept in lockstep with the
+/// `FamilyProviderFactory` impls that branch on `wire_api == Responses`
+/// (`openai`, `llamacpp`, `custom`, and the generic openai-compatible path).
+pub fn family_honors_wire_api(family: &str) -> bool {
+    matches!(family, "openai" | "llamacpp" | "custom")
+}
--- a/crates/zeroclaw-config/src/schema.rs
+++ b/crates/zeroclaw-config/src/schema.rs
@@ -14588,7 +14588,26 @@ impl Config {
     /// Adding a new warning: append a check here, pick a stable `code`,
     /// and document the code in `validation_warnings.rs`.
     pub fn collect_warnings(&self) -> Vec<crate::validation_warnings::ValidationWarning> {
-        Vec::new()
+        let mut warnings = Vec::new();
+        // `wire_api` is only honored by bring-your-own-endpoint families; on a
+        // branded family with a fixed wire protocol it is silently ignored.
+        // Surface that so an operator who sets it on, e.g., `mistral` learns it
+        // has no effect instead of debugging unexpected runtime behavior.
+        for (family, alias, entry) in self.providers.models.iter_entries() {
+            if entry.wire_api.is_some() && !crate::provider_aliases::family_honors_wire_api(family)
+            {
+                warnings.push(crate::validation_warnings::ValidationWarning::new(
+                    "wire_api_not_supported_for_family",
+                    format!(
+                        "wire_api is set on `{family}.{alias}` but the `{family}` family has a \
+                         fixed wire protocol and ignores it. wire_api only takes effect on the \
+                         openai, llamacpp, and custom (openai-compatible) families."
+                    ),
+                    format!("providers.models.{family}.{alias}.wire_api"),
+                ));
+            }
+        }
+        warnings
     }
 
     /// Validate configuration values that would cause runtime failures.
@@ -20686,6 +20705,35 @@ group_policy = "disabled"
         assert_eq!(
             hardened_mode, 0o600,
             "Saving config should restore owner-only permissions (0600)"
+        );
+    }
+
+    #[test]
+    async fn collect_warnings_flags_wire_api_on_fixed_protocol_family() {
+        let mut config = Config::default();
+        // mistral has a fixed wire protocol and ignores wire_api.
+        config
+            .providers
+            .models
+            .ensure("mistral", "primary")
+            .unwrap()
+            .wire_api = Some(WireApi::Responses);
+        // custom honors wire_api — must NOT warn.
+        config
+            .providers
+            .models
+            .ensure("custom", "vllm")
+            .unwrap()
+            .wire_api = Some(WireApi::Responses);
+
+        let warnings = config.collect_warnings();
+        assert_eq!(warnings.len(), 1, "exactly the mistral entry should warn");
+        let w = &warnings[0];
+        assert_eq!(w.code, "wire_api_not_supported_for_family");
+        assert_eq!(w.path, "providers.models.mistral.primary.wire_api");
+        assert!(
+            !warnings.iter().any(|w| w.path.contains("custom.vllm")),
+            "custom honors wire_api and must not warn",
         );
     }
 
--- a/crates/zeroclaw-providers/src/factory.rs
+++ b/crates/zeroclaw-providers/src/factory.rs
@@ -1189,6 +1189,17 @@ impl FamilyProviderFactory for CustomModelProviderConfig {
                  `[model_providers.custom.<alias>] uri = \"https://your-api.com\"` in config.toml.",
             )
         })?;
+        if self.base.wire_api == Some(zeroclaw_config::schema::WireApi::Responses) {
+            let mut p =
+                crate::openai::OpenAiResponsesModelProvider::new(alias, Some(base_url), key);
+            if let Some(mt) = opts.provider_max_tokens {
+                p = p.with_max_tokens(Some(mt));
+            }
+            if let Some(ref effort) = opts.reasoning_effort {
+                p = p.with_reasoning_effort(Some(effort.clone()));
+            }
+            return Ok(Box::new(p));
+        }
         let mut p = OpenAiCompatibleModelProvider::new_with_vision(
             alias,
             "Custom",
@@ -1218,6 +1229,17 @@ impl FamilyProviderFactory for zeroclaw_config::schema::ModelProviderConfig {
                  `[model_providers.<family>.<alias>] uri = \"https://your-api.com\"` in config.toml.",
             )
         })?;
+        if self.wire_api == Some(zeroclaw_config::schema::WireApi::Responses) {
+            let mut p =
+                crate::openai::OpenAiResponsesModelProvider::new(alias, Some(base_url), key);
+            if let Some(mt) = opts.provider_max_tokens {
+                p = p.with_max_tokens(Some(mt));
+            }
+            if let Some(ref effort) = opts.reasoning_effort {
+                p = p.with_reasoning_effort(Some(effort.clone()));
+            }
+            return Ok(Box::new(p));
+        }
         let mut p = OpenAiCompatibleModelProvider::new_with_vision(
             alias,
             "OpenAI Compatible",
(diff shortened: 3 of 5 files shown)


WareWolf-MoonWall (approved):
## Summary

Reviewed at head commit 2886ce06 (1 commit). This PR fixes a two-part bug where `wire_api = "responses"` was ignored for `custom` and generic openai-compatible provider families: (1) the factory only routed to `OpenAiResponsesModelProvider` for `openai`/`llamacpp` families, and (2) the delegate tool built target providers through the bare factory without config/alias context, so per-alias config (including `wire_api`) was never read. The fix routes both families to the responses provider when `wire_api == Responses` and updates the delegate build path to use `create_model_provider_for_alias` with per-target alias-resolved options. PR body is exemplary; validation evidence is thorough; CI is running (5 successful, 7 pending at review time).

---

### 🟢 What looks good — Dual-path fix with orthogonal test coverage

This is a textbook example of how to fix a multi-layered bug: **identify both contributing gaps, fix each independently, and write tests that prove each fix is load-bearing.** The PR body clearly names the two gaps:

1. **Factory gap:** `CustomModelProviderConfig` and generic `ModelProviderConfig` (openai-compatible) ignored `base.wire_api` entirely and always built a chat_completions provider via `apply_compat_options`. This affected the primary agent path too, not only delegation.

2. **Delegate path gap:** `DelegateTool` built the target provider through the bare `create_model_provider_with_options(type, ..)` factory with no config/alias context (`config = None`, `alias = "default"`), so even after the factory fix, the target's typed `custom.vllm` entry — and its `wire_api` — was never read.

The test coverage is orthogonal and complete:

- **Factory tests** (`custom_factory_routes_to_responses_provider_when_wire_api_responses`, `openai_compatible_factory_routes_to_responses_when_wire_api_responses`, `custom_factory_defaults_to_chat_completions_without_wire_api`) assert the factory routing change works in isolation, independent of the delegate path.

- **Delegate test** (`delegate_builds_target_provider_with_its_declared_wire_api`) proves the delegate build path now reads the per-alias config and honors `wire_api`. The test explicitly includes a **regression guard**: it calls the bare factory path (`create_model_provider_with_options`) and asserts it does NOT yield a responses provider, proving the alias-aware path is load-bearing. This is exactly the kind of negative test that prevents future refactors from reintroducing the bug.

The commit message documents the reproduction path ("pre-fix, `CustomModelProviderConfig::create_provider` with `wire_api = responses` yielded `default_wire_api() == 'chat_completions'`") and explicitly states the test guards the wiring ("The delegate test fails if the factory fix is reverted, proving both changes are load-bearing"). This is mature test-driven development: the author verified the test would catch a regression before shipping.

---

### 🟢 What looks good — Surgical scope and unchanged escalation guard

The PR body explicitly scopes out 18 branded provider families with fixed wire protocols (mistral, deepseek, groq, etc.) — the change **only** touches the two bring-your-own-endpoint families where `wire_api` is meaningful and operator-configurable. The escalation/risk-profile delegation gate is unchanged ("Risk profiles matched — the escalation guard was not the cause"), so this is purely a wire-protocol routing fix, not a security-boundary change.

The factory routing mirrors the existing `openai`/`llamacpp` implementations exactly: check `wire_api == Some(WireApi::Responses)`, route to `OpenAiResponsesModelProvider`, otherwise fall through to `OpenAiCompatibleModelProvider` via `apply_compat_options`. The symmetry is obvious from the diff (identical code structure across `CustomModelProviderConfig` and `ModelProviderConfig` impls), which makes future maintenance trivial — if the responses provider gains new options (e.g., `with_max_tokens`, `with_reasoning_effort`), the pattern is already established and duplicated in three places (custom, generic, and the pre-existing openai/llamacpp paths).

---

### 🟢 What looks good — Delegate build path with graceful fallback

The new `build_target_provider` method routes through `create_model_provider_for_alias` when the delegate has `root_config` and the `model_provider` is a dotted alias (`family.alias`), and falls back to the bare factory path otherwise. This is exactly the right fallback strategy: legacy delegate constructors without `root_config` (or targets specified as bare family names instead of dotted aliases) continue to work, while new constructions that provide `root_config` get per-alias config resolution.

The method also gracefully handles the `zeroclaw_dir` option: if the alias-resolved options don't have it, it inherits from `self.provider_runtime_options.zeroclaw_dir`. This is the kind of defensive code that prevents silent config-loss regressions when moving from one resolution path to another.

---

### 🔵 Suggestion — Document the wire_api field in operator-facing docs

The `wire_api` field exists on `ModelProviderConfig` and is now honored for three families (openai, llamacpp, custom, generic openai-compatible), but I don't see a corresponding update to operator-facing docs (e.g., `docs/book/src/providers/configuration.md` or a custom-provider guide). If there's a section that documents the `custom` family or the generic openai-compatible path, adding a note about `wire_api = "responses"` — what it does, when to use it (self-hosted vLLM/TGI endpoints that only speak the responses wire), and which families honor it — would help operators discover the capability without reading the PR or the schema. Not blocking — the config field already existed and is now correctly honored — but the gap this PR closes suggests the field wasn't well-known, and documentation would close that knowledge gap.

---

### 🔵 Suggestion — Consider a validation warning for wire_api on families that ignore it

The PR explicitly scopes out 18 branded provider families that have fixed wire protocols. If an operator mistakenly writes:

```toml
[providers.models.mistral.primary]
wire_api = "responses"
```

...the field is silently ignored (Mistral's factory never checks `wire_api` because the protocol is fixed). Consider emitting a validation warning (e.g., `wire_api_not_supported_for_family`) when `wire_api` is set on a family that doesn't honor it. This would teach operators that `wire_api` is only meaningful for the four families that support it (openai, llamacpp, custom, generic openai-compatible), and help them discover configuration mistakes during `zeroclaw config validate` instead of observing unexpected behavior at runtime.

Not blocking — the current behavior (silent ignore) is safe and the use case is rare — but it's a cheap addition that would pair well with the documentation suggestion above.

---

### 🔵 Suggestion — Extract wire_api routing into a helper

The factory routing logic is now duplicated in three places: `CustomModelProviderConfig`, generic `ModelProviderConfig`, and (pre-existing) the openai/llamacpp families. Each has the same structure:

```rust
if self.base.wire_api == Some(WireApi::Responses) {
    let mut p = OpenAiResponsesModelProvider::new(alias, Some(base_url), key);
    if let Some(mt) = opts.provider_max_tokens { p = p.with_max_tokens(Some(mt)); }
    if let Some(ref effort) = opts.reasoning_effort { p = p.with_reasoning_effort(Some(effort.clone())); }
    return Ok(Box::new(p));
}
```

Consider extracting this into a helper function in `factory.rs`:

```rust
fn build_responses_provider_if_requested(
    wire_api: Option<WireApi>,
    alias: &str,
    base_url: Option<String>,
    key: Option<&str>,
    opts: &ModelProviderRuntimeOptions,
) -> Option<anyhow::Result<Box<dyn ModelProvider>>> {
    if wire_api == Some(WireApi::Responses) {
        let mut p = OpenAiResponsesModelProvider::new(alias, base_url, key);
        if let Some(mt) = opts.provider_max_tokens { p = p.with_max_tokens(Some(mt)); }
        if let Some(ref effort) = opts.reasoning_effort { p = p.with_reasoning_effort(Some(effort.clone())); }
        Some(Ok(Box::new(p)))
    } else {
        None
    }
}
```

Then each factory becomes:

```rust
if let Some(p) = build_responses_provider_if_requested(self.base.wire_api, alias, Some(base_url), key, opts) {
    return p;
}
```

This eliminates the duplication and makes future changes to the responses provider construction (e.g., adding new options) a single-line change instead of a three-file hunt. Not blocking — the duplication is minimal and the pattern is stable — but extraction would make the code more maintainable as the provider surface grows.

---

## Verdict

**Approve.** The dual-path fix is correct, the test coverage is exemplary (including a regression guard for the bare factory path), the scope is surgical, and the PR body clearly documents the reproduction path and validation steps. The three suggestions above are all optional follow-ups that would improve discoverability and reduce duplication; none are blocking. This is ready to merge when CI completes (7 checks pending at review time).

CI status at review: 5 successful, 7 pending (build/test/check jobs in flight).

PR #7182, opened 2026-06-04
feat(rpc): persistent RPC sessions with admin kill

tidux:
## Summary

- **What changed and why:**
  - Removed the 10-minute idle-TTL and 1-second orphan-grace eviction from
    `SessionStore` — RPC sessions now persist until explicitly closed, killed,
    or daemon exit (tmux-style). A user who closes their laptop lid and
    reconnects an hour later from a different IP finds their ZeroCode session
    intact.
  - Removed `mark_orphaned` calls from `local.rs` / `wss.rs` transport disconnect
    handlers; the daemon reaper loop now only releases actor-queue slots
    (`evict_idle`) on a 60-second tick instead of evicting sessions.
  - Added `session/kill` RPC: admin-only forceful termination that cancels any
    in-flight turn (with `CancelCause::AdminKill` for audit attribution) and
    removes the session from memory without touching disk history.
  - Added `X` keybinding in the ZeroCode dashboard Sessions detail pane to invoke
    `session/kill`, with i18n key and state cleanup.
- **Scope boundary:** Does not change channel sessions, gateway sessions, cron
  sessions, or any persistent (on-disk) history. `session/delete` (which removes
  disk history) is untouched. No subscriber reference-counting or heartbeat
  mechanism added — deferred to a future multiuser-sessions milestone.
- **Blast radius:** `zeroclaw-runtime` RPC layer and `zerocode` TUI only.
  Channel orchestrators, gateway, and CLI are unaffected. Any operator tooling
  that relies on sessions being reaped after 10 minutes of inactivity will need
  to use `session/kill` instead.
- **Linked issue(s):** None

```bash
cargo fmt --all -- --check
```

  `cargo fmt --all -- --check` — clean (zero diff after auto-fix committed as
  `style: cargo fmt formatting fixes`)

  `cargo clippy --all-targets -- -D warnings` — clean, zero warnings or errors

  `cargo test -p zeroclaw-runtime --lib`:
  ```
  test daemon::tests::ephemeral_grace_period_resets_on_reconnect ... ok
  test rpc::turn::tests::drain_must_not_idle_cancel_a_live_turn_across_a_long_tool_gap ... ok

  test result: ok. 1984 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 5.16s
  ```

  - `kill_session` cause-lifecycle: confirmed `take_cancel_cause` returns
    `AdminKill` (not `None`) after firing, verified by the
    `kill_session_cause_is_admin_kill` unit test.
  - Double-`TurnComplete` eliminated: caught and fixed by code review — the
    handler no longer emits its own `TurnComplete`; the prompt task emits the
    single canonical one when it observes the cancel token.
  - Dead-symbol grep confirmed: `SESSION_IDLE_TTL`, `SESSION_DISCONNECT_GRACE`,
    `mark_orphaned`, `evict_expired`, `EvictReason`, `ReaperIdle`, `ReaperOrphaned`,
    `.reclaim(` — all absent from `zeroclaw-runtime/src/` and `zerocode/src/`.
  - Dashboard X keybinding, i18n key, and state cleanup manually inspected.
  - Full workspace `cargo test` blocked by missing system library (`libdbus-1-dev`
    not installed in dev environment) — pre-existing on `master`, unrelated to
    this branch.

  `cargo test --workspace` skipped: `dbus-1` system library absent from local
  environment; confirmed pre-existing on `master`. `zeroclaw-runtime` and
  `zerocode` crates (the only ones changed) pass cleanly.

#### 14. Persistent RPC sessions + admin kill (#7182) — Tier 5 ⚠️

Commit `4ede0b0a6`. Sessions survive transport disconnect; idle-TTL + orphan
eviction removed; `session/kill` added.

- [x] Session survives a transport (WSS) disconnect: reconnect resumes the same session, no orphan/eviction — PASS (unit, partial). `session_persists_after_transport_disconnect` (session.rs:926) asserts the store still holds the session after a disconnect (count stays 1; no TTL/orphan reaping). NOTE: this proves NON-REAPING but does not exercise the reconnect-resume round trip — reconnect-resumes-same-session is cross-verified live in §13.
- [x] No idle-TTL reaping of an idle session under a returning client — PASS (unit, by removal). The idle-TTL + orphan-eviction code was REMOVED (commit 6990103); `session_persists_after_transport_disconnect` confirms an idle session is retained. The reaper now retains only queue slots (see below).
- [~] `initialize` no longer calls `reclaim()` (not orphaned on re-init) — CODE-VERIFIED (no isolated unit test). `reclaim()` call removed from initialize (commit 5de75e0); `handle_initialize` (dispatch.rs:518) no longer reclaims. No regression test pins this; relies on the diff. Cross-check live in §13/§14 reconnect.
- [x] `session/kill` forcefully terminates; records `AdminKill` only when it fires the cancel token — PASS (unit). `kill_session_cancels_inflight_and_removes` (session.rs:946) — fires the registered token, removes session, returns true. `kill_session_cause_is_admin_kill` (session.rs:994) — with a token, `take_cancel_cause` yields `Some(CancelCause::AdminKill)`. `kill_session_idle_session_removed` (969) — no token → still removed (idle). `kill_session_missing_returns_false` (986) — ghost id → false. The "only when it fires the cancel token" half (no token → no spurious AdminKill at the verdict site) is the idle-removal path; the AdminKill cause is set on the cancel path. 22/22 session tests green.
- [~] No duplicate `TurnComplete` on kill; kill-race logged not panicked — CODE-VERIFIED (needs live). Duplicate guard lives in `Agent::turn_streamed` (dispatch.rs:1016 references the load-bearing guard); kill-race paths log + emit one attributed TurnComplete (dispatch.rs:1380/1408/1430). No isolated test for the kill-race-no-panic; verify live during §14 manual kill.
- [x] zerocode Sessions detail pane: `X` kills selected session; state cleaned up after — PASS (code + registry test). `KillSession [Chord::char('X')]` action registered (keymap/actions.rs:224); dashboard wires `X` → `self.rpc.session_kill(&sid)` (dashboard.rs:1696) → client method `session_kill` (client.rs:1093). `every_preset_covers_every_action` green proves the action is in every keymap preset (registry-driven, not hardcoded). End-to-end UI kill flow to verify live during §14.
- [x] Reaper retains queue slots only; no leak across kills — PASS (unit, by removal + design). Per commit d5d79a5 the reaper retains ONLY queue slots (session bodies survive disconnect); idle/orphan eviction removed. `session_persists_after_transport_disconnect` confirms bodies are not reaped; `kill_session_*` tests confirm removal frees the slot cleanly (count→0).

**§14 STATUS: runtime kill/persistence fully unit-green (22 tests). Three items CODE-VERIFIED pending live: (1) reconnect-resume round trip → §13; (2) no-reclaim-on-initialize → §13/§14 live; (3) no-duplicate-TurnComplete/kill-race-no-panic + UI `X`-kill end-to-end → §14 live daemon.**

## Compatibility (required)

- Backward compatible? **Yes** — existing `session/close`, `session/delete`, and
  `session/cancel` RPCs are unchanged. Clients that do not call `session/kill`
  see no behavior change except that sessions no longer disappear after 10 minutes
  of inactivity.
- Config / env / CLI surface changed? **No** — no new config keys, env vars, or
  CLI flags.

`git revert <sha>` is the plan. Risk is low: no schema changes, no on-disk format
changes, no breaking wire changes.

### DIFF
--- a/apps/zerocode/locales/en/zerocode.ftl
+++ b/apps/zerocode/locales/en/zerocode.ftl
@@ -211,6 +211,7 @@ zc-dashboard-help-jump-bottom = Jump to bottom
 zc-dashboard-help-jump-top = Jump to top
 zc-dashboard-help-open-detail = Open detail pane
 zc-dashboard-help-search-filter = Search / filter
+zc-dashboard-help-kill-session = Kill session
 
 zc-dashboard-yes = yes
 zc-dashboard-no = no
--- a/apps/zerocode/src/client.rs
+++ b/apps/zerocode/src/client.rs
@@ -1082,7 +1082,7 @@ impl RpcClient {
         .await
     }
 
-    pub async fn session_kill(&self, session_id: &str) -> anyhow::Result<()> {
+    pub async fn session_kill(&self, session_id: &str) -> Result<()> {
         let _: serde_json::Value = self
             .call(
                 method::SESSION_KILL,
--- a/apps/zerocode/src/client.rs
+++ b/apps/zerocode/src/client.rs
@@ -85,6 +85,7 @@ pub mod method {
     pub const SESSION_APPROVE: &str = "session/approve";
     pub const SESSION_RENAME: &str = "session/rename";
     pub const SESSION_CLOSE: &str = "session/close";
+    pub const SESSION_KILL: &str = "session/kill";
     // Dashboard
     pub const STATUS: &str = "status";
     pub const HEALTH: &str = "health";
@@ -1079,6 +1080,16 @@ impl RpcClient {
             serde_json::json!({ "session_id": session_id }),
         )
         .await
+    }
+
+    pub async fn session_kill(&self, session_id: &str) -> anyhow::Result<()> {
+        let _: serde_json::Value = self
+            .call(
+                method::SESSION_KILL,
+                serde_json::json!({ "session_id": session_id }),
+            )
+            .await?;
+        Ok(())
     }
 
     pub async fn session_rename(
--- a/apps/zerocode/src/dashboard.rs
+++ b/apps/zerocode/src/dashboard.rs
@@ -2040,7 +2040,10 @@ impl crate::widgets::HelpContext for Dashboard<'_> {
                 E::key("?", crate::i18n::t("zc-dashboard-help-this-help")),
             ];
             if self.tab == Tab::Sessions {
-                entries.push(E::key("X", crate::i18n::t("zc-dashboard-help-kill-session")));
+                entries.push(E::key(
+                    "X",
+                    crate::i18n::t("zc-dashboard-help-kill-session"),
+                ));
             }
             return HelpNode::entries(entries);
         }
--- a/apps/zerocode/src/dashboard.rs
+++ b/apps/zerocode/src/dashboard.rs
@@ -1684,6 +1684,17 @@ impl<'a> Dashboard<'a> {
                 self.search_buf.clear();
                 self.last_poll = None; // re-poll for server-side search
             }
+            Some(DashboardTabAction::KillSession) if self.tab == Tab::Sessions => {
+                if let Some(idx) = self.selected_session_index() {
+                    let sid = self.sessions[idx].session_id.clone();
+                    let _ = self.rpc.session_kill(&sid).await;
+                    self.detail_open = false;
+                    self.detail_scroll = 0;
+                    self.session_messages.clear();
+                    self.session_messages_id = None;
+                    self.last_poll = None;
+                }
+            }
             _ => {}
         }
         false
@@ -1999,7 +2010,7 @@ impl crate::widgets::HelpContext for Dashboard<'_> {
         }
 
         if self.detail_open {
-            return HelpNode::entries(vec![
+            let mut entries = vec![
                 E::new(
                     vec!["Esc", "Enter"],
                     crate::i18n::t("zc-dashboard-help-close-detail"),
@@ -2025,7 +2036,11 @@ impl crate::widgets::HelpContext for Dashboard<'_> {
                 E::key("c", crate::i18n::t("zc-dashboard-help-clear-search")),
                 E::key("q", crate::i18n::t("zc-dashboard-help-quit")),
                 E::key("?", crate::i18n::t("zc-dashboard-help-this-help")),
-            ]);
+            ];
+            if self.tab == Tab::Sessions {
+                entries.push(E::key("X", "kill session"));
+            }
+            return HelpNode::entries(entries);
         }
 
         // Per-tab bindings — only show what actually works on this tab.
--- a/apps/zerocode/src/dashboard.rs
+++ b/apps/zerocode/src/dashboard.rs
@@ -1692,6 +1692,8 @@ impl<'a> Dashboard<'a> {
                     self.detail_scroll = 0;
                     self.session_messages.clear();
                     self.session_messages_id = None;
+                    self.session_messages_total = 0;
+                    self.session_messages_start = 0;
                     self.last_poll = None;
                 }
             }
@@ -2038,7 +2040,7 @@ impl crate::widgets::HelpContext for Dashboard<'_> {
                 E::key("?", crate::i18n::t("zc-dashboard-help-this-help")),
             ];
             if self.tab == Tab::Sessions {
-                entries.push(E::key("X", "kill session"));
+                entries.push(E::key("X", crate::i18n::t("zc-dashboard-help-kill-session")));
             }
             return HelpNode::entries(entries);
         }
--- a/apps/zerocode/src/keymap/actions.rs
+++ b/apps/zerocode/src/keymap/actions.rs
@@ -218,6 +218,7 @@ keyactions! {
         Refresh          [Chord::char('r')] => "refresh",
         JumpStart        [Chord::char('g'), Chord::key(KeyCode::Home)] => "jump to start",
         JumpEnd          [Chord::char('G'), Chord::key(KeyCode::End)] => "jump to end",
+        KillSession      [Chord::char('X')] => "kill session",
     }
 }
 
--- a/apps/zerocode/src/keymap/actions.rs
+++ b/apps/zerocode/src/keymap/actions.rs
@@ -215,10 +215,10 @@ keyactions! {
         DetailWidenDown  [Chord::shift(KeyCode::Down)] => "widen detail down",
         BeginSearch      [Chord::char('/')] => "search",
         CopyDetail       [Chord::char('c')] => "copy detail",
+        KillSession      [Chord::char('X')] => "kill session",
         Refresh          [Chord::char('r')] => "refresh",
         JumpStart        [Chord::char('g'), Chord::key(KeyCode::Home)] => "jump to start",
         JumpEnd          [Chord::char('G'), Chord::key(KeyCode::End)] => "jump to end",
-        KillSession      [Chord::char('X')] => "kill session",
     }
 }
 
--- a/crates/zeroclaw-runtime/src/daemon/mod.rs
+++ b/crates/zeroclaw-runtime/src/daemon/mod.rs
@@ -357,40 +357,14 @@ pub async fn run(
         let sessions = std::sync::Arc::new(SessionStore::new(64, session_queue.clone()));
 
         {
-            let reaper_sessions = std::sync::Arc::clone(&sessions);
             let reaper_queue = std::sync::Arc::clone(&session_queue);
             zeroclaw_spawn::spawn!(async move {
-                const TICK: std::time::Duration = std::time::Duration::from_secs(15);
+                const TICK: std::time::Duration = std::time::Duration::from_secs(60);
                 let mut interval = tokio::time::interval(TICK);
                 interval.tick().await;
                 loop {
                     interval.tick().await;
-                    let evicted = reaper_sessions.evict_expired().await;
                     let queue_evicted = reaper_queue.evict_idle().await;
-                    for ev in &evicted {
-                        let span = ::zeroclaw_log::info_span!(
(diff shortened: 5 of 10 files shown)


WareWolf-MoonWall (approved):
## Summary

Reviewed at head commit d7524182 (10 commits). This PR removes the 10-minute idle-TTL and 1-second disconnect-grace eviction from RPC sessions, implementing tmux-style persistent sessions that survive transport disconnect until explicit close, kill, or daemon exit. Adds an admin-only `session/kill` RPC for forceful termination. The PR body is complete and well-written; validation evidence is thorough; CI is pending lint check at post time.

---

### 🟢 What looks good — Surgical refactor with clear tradeoff management

The removal of the reaper/orphan mechanism is clean and deliberate. Code deleted: `SESSION_IDLE_TTL`, `SESSION_DISCONNECT_GRACE`, `EvictReason`, `mark_orphaned`, `evict_expired`, `reclaim`, every associated span/log/test — dead symbols are gone, not just unused. The simplification from two reaper modes (orphaned + idle) down to queue-slot eviction only is a meaningful cognitive-load win.

The `kill_session` lifecycle is correct: the `take_cancel_cause` method properly records `AdminKill` only when the cancel token fires (commit e097137e fixed the initial implementation that was writing then immediately removing the cause). The `kill_session_cause_is_admin_kill` test pins the invariant that the cause must be observable after firing, which is exactly the kind of test that prevents future regressions.

The TUI integration (X keybinding, i18n key `zc-dashboard-help-kill-session`, state cleanup that resets `session_messages_total` and `session_messages_start` in addition to closing the pane) is complete and consistent with existing dashboard actions.

The PR body documents validation evidence in detail (including the manual dead-symbol grep and the observed failure mode that was fixed in commit 18b371bf — the double-`TurnComplete` protocol violation), explains why full-workspace test was skipped (missing `libdbus-1-dev`, pre-existing on `master`), and names the blast radius clearly.

The session-persistence tradeoff is explicitly acknowledged ("sessions no longer disappear after 10 minutes of inactivity") with forward-looking notes about the operator-tool impact ("Any operator tooling that relies on sessions being reaped after 10 minutes of inactivity will need to use `session/kill` instead"). This is what mature authorship looks like: the author knows the change carries a behavior shift and names who is affected and how to adapt.

---

### 🔵 Suggestion — Session enumeration and admin-UI discoverability

The `session/kill` RPC is admin-only by definition (per the PR body), but the access-control enforcement is not visible in the diff. If the RPC dispatcher already has an admin-gate wrapper (e.g., checking `tui_id` against a configured admin list or role), it's fine; if not, consider whether an unauthorized caller could invoke `session/kill` on another user's session. The existing `session/delete` RPC likely has the same gate; aligning `session/kill` with that precedent would close the gap.

For operators managing long-lived sessions (the new steady state), a `session/list` or `sessions/enumerate` RPC (returning session IDs, agent aliases, owner TUI IDs, and idle/active status) would pair well with `session/kill` — otherwise the X keybinding in the TUI is the only discovery path, and headless/scripted operators have no way to see what sessions exist before killing them. Not blocking for this PR (the RPC set is internally consistent as-is), but worth considering as a follow-up once persistent sessions are live and operators start asking "what sessions am I holding?"

---

### 🔵 Suggestion — Document the persistence contract in operator-facing docs

The PR body notes the blast radius ("Any operator tooling that relies on sessions being reaped after 10 minutes of inactivity will need to use `session/kill` instead"), but I don't see a corresponding update to the RPC reference docs or operator playbook. If there's a `docs/book/src/reference/rpc-api.md` or similar, adding a note about session lifetime (persistent until explicit close/kill/daemon exit) and the new `session/kill` method would help operators discover the change without reading the PR. Not blocking — the code is self-documenting and the TUI exposes the capability — but operator-facing documentation lags behind implementation more often than we'd like, and this is a high-leverage place to close that gap early.

---

### 🔵 Suggestion — Heap-release timing

The daemon reaper loop now only calls `release_freed_heap()` when `queue_evicted > 0` (commit d5d79a50, line ~386 in the truncated diff). Previously it also released after session eviction. With sessions now persisting indefinitely, the heap may hold more session history/turn state across a longer lifetime, and the 60-second reaper tick might be the only periodic release point. Consider whether the reaper should unconditionally release every N ticks (e.g., every 5 minutes) even when `queue_evicted == 0`, or whether `session/close` and `session/kill` handlers should individually call `release_freed_heap()` after removing a session. The `session/kill` handler already does (commit dispatch.rs, line ~898), so the pattern is established; `session/close` might benefit from the same if it doesn't already have it. Not urgent — the trimming still happens, just less frequently — but heap fragmentation over long uptimes is a known pain point in long-lived daemons, and this change shifts the lifetime distribution in a way that makes periodic release more valuable.

---

## Verdict

**Approve.** The refactor is clean, the persistence contract is well-explained, the validation evidence is thorough, and the implementation correctly handles the kill-session lifecycle (including the fix for the double-`TurnComplete` protocol violation). The three suggestions above are all optional follow-ups; none are blocking. This is ready to merge when CI completes (lint check was pending at review time) and any other reviewer dependencies are resolved.

CI status at review: 2 successful, 1 pending (Quality Gate/Lint).

singlerider (approved):
Reviewed at head d7524182 (10 commits, +231/−348 across 10 files). This removes
the 10-minute idle-TTL and 1-second disconnect-grace eviction from the RPC
`SessionStore` in favor of tmux-style persistence, and adds `session/kill` for
forceful in-memory termination wired to `X` in the zerocode Sessions detail
pane. I cross-checked the diff against the local tree on `integration/zerocode-test`,
verified the cancel-cause lifecycle, and confirmed the transport trust model.

I'm approving. The change is internally consistent: the old RED/GREEN reaper
guards are correctly retired and replaced with kill-path guards, the
cause-attribution invariant is pinned by a test, and the dead-symbol sweep is
complete. Two clarifications below — neither blocks merge.

### 🟢 What looks good — cancel-cause lifecycle is correct and pinned

The commit e0971370 fix is the right call: moving `record_cancel_cause` inside
the `if let Some(token)` block and dropping the unconditional
`cancel_causes.remove` means `take_cancel_cause` now returns `Some(AdminKill)`
at the verdict site instead of `None`. `record` inserts, `take` removes-and-
returns — the pair is symmetric, and `kill_session_cause_is_admin_kill` locks
the invariant so a future refactor can't silently regress it back to a bare
"cancelled" with no provenance. Retiring `ReaperOrphaned`/`ReaperIdle` for a
single `AdminKill` keeps the "every fired token is attributable, no catch-all"
contract intact.

### 🟢 What looks good — double-TurnComplete caught before it shipped

Removing the pre-emptive `emit_turn_complete(Cancelled)` from
`handle_session_kill` (commit 18b371bf) is correct: the running
`handle_session_prompt` task already emits the canonical `TurnComplete` when it
observes the cancel token, so the original code was a protocol double-emission.
The DEBUG log for the existence-check-to-kill race is the right level — that
window is benign (concurrent close) and shouldn't be noise at INFO.

### 🟢 What looks good — reaper teardown is clean, not just gutted

The daemon loop drop is thorough: `reaper_sessions` Arc, `evict_expired`, and
the three eviction `record!` blocks are all gone, the tick moves 15s→60s, and
the surviving queue-slot log is rescoped to `release_freed_heap()` only firing
when slots were actually freed. `local.rs` and `wss.rs` lose `mark_orphaned`
symmetrically. No orphaned imports, no half-migrated callsites — the
"follow-on cleanup" the first commit promised actually landed.

### 🟢 What looks good — kill on a paused/idle session and missing-session both covered

`kill_session_idle_session_removed` and `kill_session_missing_returns_false`
cover the no-inflight-turn and ghost-session paths, so the `bool` return
contract is exercised both ways. The handler's `get_agent` existence check
before kill, plus the `killed` flag in `SessionKillResult`, gives the client an
honest answer when a session vanished mid-call.

### 🟢 What looks good — but the "admin-only" framing needs one wording fix

The trust model is sound: I verified the local socket is `0o600` owner-only
(local.rs, with a `socket_permissions` test asserting mode `0o600`), so the RPC
surface is correctly process-local. One wording correction for the record: the
body and commit 4c5c6e54 call `session/kill` "admin-only … regardless of
ownership," but there is no admin or ownership check in `handle_session_kill` —
and the existing `session/delete` / `session/close` / `session/cancel`
handlers don't have one either. "Admin" here means "the daemon's own OS user,"
enforced by the transport, not an in-app role. Reword to "process-local /
owner-trusted RPC" so a future reader doesn't assume an authz gate exists —
especially relevant when the WSS transport (a different boundary) is in play.
Consistent with the rest of the RPC surface; purely a documentation accuracy
note.

### 🟢 What looks good — tmux persistence is the right call; name the RSS tradeoff

Dropping `SESSION_IDLE_TTL` for tmux-style persistence is a defensible design
choice and the implementation is clean. The one thing to capture: the old TTL
doc called itself "the backstop that keeps daemon RSS bounded" for a client
that opens sessions and walks away without a clean disconnect. After this
change the only frees are explicit `close`/`delete`/`kill` or daemon exit, so
an abandoned-but-never-closed session lives until the daemon dies. The body's
"deferred to a future multiuser-sessions milestone" covers subscriber
ref-counting but not this single-user abandoned-session surface. Worth stating
plainly in the observable-symptoms section so an operator watching daemon RSS
climb knows it's by-design and that `session/kill` (or a restart) is the lever.
A follow-up issue for an opt-in TTL would make the backstop recoverable without
reverting persistence — but that's future work, not a condition of this merge.

---

**Label note (for a maintainer with label permissions):** `risk: low` is
auto-applied from path heuristics, but this changes session-lifecycle
semantics daemon-wide and removes a memory-bounding backstop. `risk: medium`
is the honest tier. No milestone is set; scope reads as v0.8.1 runtime work
rather than a v0.8.0 schema-migration item — worth confirming placement.

Approving — clean, well-tested, no active blocks.

PR #7191, opened 2026-06-04
fix(web): correct chat model switcher resolution and switch timeout

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The chat header model switcher resolved the *install-wide default* model instead of the model for the agent the chat is bound to, because `loadModelInfo` called `getStatus()` with no alias. In multi-agent setups the dropdown showed the wrong model, and after a successful switch the re-fetch reverted `currentModel` to the global default. `getStatus(agent?)` now forwards `?agent=<alias>` (gateway runs the same `resolved_model_provider_for_agent` it uses to build the Agent) and `loadModelInfo` passes the bound alias.
  - Removed dead code: `loadModelInfo` tried `getProp('model')` / `getProp('default_model')` to "prefer the configured value", but those top-level paths don't exist in the schema (they 404) and the prop `GET` endpoint never returns a `populated` flag for non-secret fields — two failing round trips per load that never affected the result.
  - `switchModel` armed a single 10s watchdog covering both the config-write and the socket-reconnect phases, so a slow daemon write could fire "model switch timed out" while the switch was still succeeding (and the fired watchdog nulled the pending ref, so `onOpen` then skipped updating `currentModel`). Split into a per-phase watchdog, each with its own budget, plus an `=== model` identity guard and bail checks so a fired timer / superseding switch can't tear down a live socket.
- **Scope boundary:** Frontend only (`web/src/contexts/AgentContext.tsx`, `web/src/lib/api.ts`). No gateway/runtime/Rust changes; the existing `providers.models.<provider>.model` write path and the `/api/status?agent=` resolution are used as-is.
- **Blast radius:** `getStatus` gains an *optional* parameter; the three existing no-arg callers (Sidebar, Dashboard, `useLocale`) keep the global-status behavior unchanged. Only the agent-chat model dropdown behavior changes.
- **Linked issue(s):** none
- **Labels:** `type: fix`, `risk: low`, `size: S`, `web` (maintainer to confirm)

This is a web/TypeScript-only change; the Rust battery (`cargo fmt/clippy/test`) is **not applicable** (no Rust touched) and was intentionally skipped for that reason. Ran the web equivalents:

```bash
$ npx tsc -b
tsc EXIT=0   # no diagnostics

$ npx vite build
dist/assets/AgentChat-6cTS2ScQ.js                34.09 kB │ gzip:   9.79 kB
...
(!) Some chunks are larger than 500 kB after minification.   # pre-existing, Config.js/index.js bundles — unrelated to this diff
✓ built in 17.71s
build EXIT=0
```

- **Commands run and tail output:** see above — `tsc -b` clean, `vite build` succeeds.
- **Beyond CI — what did you manually verify?** Code-audited the `onOpen` / `onClose` / `switchModel` interactions: success path clears the reconnect watchdog and sets `currentModel`; failed reconnect surfaces the close error; `catch` clears any armed timer; superseded/stale switches bail via the identity guard. Did **not** run a live end-to-end switch against a running daemon.
- **If any command was intentionally skipped, why:** Rust battery skipped — no Rust files in the diff.

- New external network calls? `No` (uses existing `/api/status`, `/api/config/prop`, `/ws/chat`)

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` (`getStatus` adds an optional arg; no config/env/CLI surface)
- Upgrade steps: none.

Low risk: `git revert <sha>`.

### DIFF
--- a/web/src/contexts/AgentContext.tsx
+++ b/web/src/contexts/AgentContext.tsx
@@ -72,6 +72,54 @@ const LOCAL_PROVIDER_NAMES: Record<string, string> = {
   vllm: 'vLLM',
 };
 
+/**
+ * Extract the `model` identifiers from a `model_routes` prop value.
+ *
+ * `/api/config/prop` returns the value as a display *string* (never a real
+ * array, and with no `populated` flag), so this tolerates every shape we might
+ * see: an already-parsed array, a JSON-encoded array, or the TOML-ish display
+ * string the gateway currently emits, e.g.
+ * `[{ hint = "fast", model = "Qwen3.6-35B-A3B", model_provider = "vllm.default" }]`.
+ * Returns a de-duplicated, order-preserving list of model names.
+ */
+function extractRouteModels(value: unknown): string[] {
+  const out: string[] = [];
+  const push = (m: unknown) => {
+    if (typeof m === 'string' && m.length > 0 && !out.includes(m)) out.push(m);
+  };
+
+  const fromArray = (arr: unknown[]) => {
+    for (const r of arr) {
+      if (r && typeof r === 'object') push((r as Record<string, unknown>).model);
+    }
+  };
+
+  if (Array.isArray(value)) {
+    fromArray(value);
+    return out;
+  }
+
+  if (typeof value === 'string') {
+    // Try strict JSON first (future-proofing if the wire format changes).
+    try {
+      const parsed = JSON.parse(value);
+      if (Array.isArray(parsed)) {
+        fromArray(parsed);
+        if (out.length > 0) return out;
+      }
+    } catch {
+      // Not JSON — fall through to string scraping.
+    }
+    // Scrape `model = "..."` pairs. `\bmodel\s*=` does not match
+    // `model_provider = ...` because of the intervening `_provider`.
+    const re = /\bmodel\s*=\s*"([^"]+)"/g;
+    let match: RegExpExecArray | null;
+    while ((match = re.exec(value)) !== null) push(match[1]);
+  }
+
+  return out;
+}
+
 function friendlyAgentError(message?: string): string {
   const raw = message?.trim() || t('agent.unknown_error');
   const localConnectFailure = raw.match(
@@ -456,40 +504,49 @@ export function AgentProvider({ agentAlias, children }: AgentProviderProps) {
 
     async function loadModelInfo() {
       try {
-        const status = await getStatus();
+        // Resolve the *agent-scoped* active model. `/api/status?agent=<alias>`
+        // runs the same `resolved_model_provider_for_agent` logic the gateway
+        // uses to construct the Agent, so `status.model` reflects the value
+        // written to this agent's provider entry
+        // (`providers.models.<provider>.model`) — including a model we just
+        // switched to. Calling `getStatus()` without the alias would return
+        // the install-wide default model, which is wrong for any non-default
+        // agent and would clobber `currentModel` right after a switch.
+        //
+        // The previous implementation also tried `getProp('model')` /
+        // `getProp('default_model')` to "prefer the configured value", but
+        // those top-level paths don't exist in the schema (they 404 with
+        // `path_not_found`) and the prop GET endpoint never returns a
+        // `populated` flag for non-secret fields — so that branch was dead
+        // code that only added two failing round trips per load. The
+        // agent-scoped status already is the configured value.
+        const status = await getStatus(agentAlias);
         if (cancelled) return;
 
-        let activeModel = status.model;
-
-        // Prefer the model written to config over the startup status value.
-        try {
-          const modelProp = await getProp('model');
-          if (modelProp.populated && typeof modelProp.value === 'string') {
-            activeModel = modelProp.value;
-          } else {
-            const defaultModelProp = await getProp('default_model');
-            if (defaultModelProp.populated && typeof defaultModelProp.value === 'string') {
-              activeModel = defaultModelProp.value;
-            }
-          }
-        } catch {
-          // ignore
-        }
+        const activeModel =
+          typeof status.model === 'string' && status.model.length > 0
+            ? status.model
+            : null;
         setCurrentModel(activeModel);
 
-        // Fetch model_routes from config
+        // Fetch model_routes from config. The REST `/api/config/prop`
+        // endpoint returns `value` as a display *string* (and carries no
+        // `populated` flag), so we can't rely on `Array.isArray`. Parse the
+        // model names out of whatever shape we get — a real array, a
+        // JSON-encoded array, or the TOML-ish display string.
         try {
           const routesProp = await getProp('model_routes');
-          if (routesProp.populated && Array.isArray(routesProp.value)) {
-            const models = routesProp.value
-              .map((r) => (r as Record<string, unknown>).model)
-              .filter((m): m is string => typeof m === 'string');
-            setAvailableModels(models.length > 0 ? models : [activeModel]);
-          } else {
-            setAvailableModels([activeModel]);
-          }
+          if (cancelled) return;
+          const models = extractRouteModels(routesProp.value);
+          // Always keep the active model selectable, even if it has no route.
+          const merged = [
+            ...(activeModel && !models.includes(activeModel) ? [activeModel] : []),
+            ...models,
+          ].filter((m): m is string => typeof m === 'string' && m.length > 0);
+          setAvailableModels(merged);
         } catch {
-          setAvailableModels([activeModel]);
+          if (cancelled) return;
+          setAvailableModels(activeModel ? [activeModel] : []);
         }
       } catch {
         // Ignore errors — dropdown will just show current model once loaded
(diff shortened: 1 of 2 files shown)


WareWolf-MoonWall (commented):
## Review of PR #7191

Reviewed commit range: `fix/web-model-switcher` (head at time of review)  
Reviewer: @WareWolf-MoonWall

---

### 🟢 What looks good — Surgical fix with strong technical foundation

The diagnosis and fix are both sound:

1. **Root-cause reasoning is correct.** The install-wide `getStatus()` call was definitionally wrong for any non-default agent — the gateway's agent-scoped resolution (`resolved_model_provider_for_agent`) already existed and returns exactly what the UI needs. Passing the alias through is the right fix, not a workaround.

2. **Dead-code removal is well-justified.** The `getProp('model')` / `getProp('default_model')` paths never worked — they 404 because those top-level keys don't exist in the schema, and the REST endpoint doesn't return `populated` flags for non-secret fields anyway. Deleting two failing round trips that never affected behavior is a pure win.

3. **Timeout split addresses the false-positive race.** The original single-watchdog design had no safe answer: too short and you'd timeout working switches, too long and you'd hang the UI on a genuinely stuck reconnect. Splitting per-phase with identity guards (`=== model`) is architecturally correct — the write phase and the reconnect phase are *different failure modes* and deserve separate budgets. The identity check prevents stale/superseded switches from clobbering live state, which was the other half of the bug.

4. **The `extractRouteModels` helper is defensive in the right way.** It tolerates the three shapes we might see (real array, JSON string, TOML-ish display) without coupling to gateway wire-format internals. The doc comment explicitly names what it's defending against and why. This is good production-readiness discipline (per FND-006 §4.1: operational errors require handling, not wishful thinking).

5. **Change discipline is excellent.** The commit message and PR body document *why* each piece changed, trace the root cause, and explicitly state the validation commands run. This is the kind of accountability that makes review efficient and future debugging feasible.

---

### 🟡 Warning — Label mismatch and missing manual validation

#### Label: auto-labeled `risk: medium`, but body claims `risk: low`

The auto-labeler saw `src/**` changes (TypeScript counts) and applied `risk: medium`. The author's "Scope boundary: Frontend only" claim is correct — no Rust, no gateway, no runtime, no config schema — but didn't override the label.

Per the PR review protocol, "labels are maintainer metadata, not a contributor blocker," and I can't edit labels as a reviewer through the skill. **Recommendation:** a maintainer with label permissions should downgrade to `risk: low` before merge. The change is frontend state-management only, backward-compatible (optional parameter), and has no blast radius beyond the agent-chat model dropdown.

#### Missing manual end-to-end validation

The author states:

> Did **not** run a live end-to-end switch against a running daemon.

For a fix targeting a *user-facing* bug (wrong model displayed, timeout false-positives), that's a gap. The code audit is thorough and the reasoning is sound, but this is the kind of fix where a manual confirmation — "I switched models in a multi-agent setup and the dropdown showed the right model before and after" — would close the loop.

**Not blocking,** because:
- The TypeScript and build checks passed, so the code is structurally correct.
- The gateway already supports `?agent=` per L187-195 of `crates/zeroclaw-gateway/src/api.rs`, so the contract exists.
- The change is scoped to frontend display logic; a broken switch would surface immediately in manual testing post-merge and revert cleanly.

But it's worth noting: a production bug fix without an explicit reproduction + verification step is higher-risk than the same fix with one.

---

### 🔵 Suggestion — Future test coverage

This PR has no tests. That's not a blocker — the web codebase has no existing test coverage for `AgentContext.switchModel` or model-dropdown logic, and this PR doesn't make that *worse* — but it's worth flagging for the record.

The kind of test that would have caught the original bugs:

1. **Unit test for `extractRouteModels`:** pass it the three wire formats (array, JSON string, TOML-ish display) and assert it returns the right model list. This is a pure function with no dependencies; very testable.

2. **Integration test for the timeout split:** mock the config-write delay and the socket-reconnect delay independently, assert that the UI only times out when *both* phases exceed their budget (not when just the write is slow). This one's harder (needs WebSocket mocking), but it's the kind of test that prevents regressions on gnarly state-machine logic.

Not asking for tests in *this* PR (the surrounding context has none), but noting it as a future investment opportunity. Per FND-006 §4.3: "When you fix a bug, write a test that would have caught it. This one habit, practiced consistently, moves the test suite toward the failure modes that actually matter."

---

### 🔵 Suggestion — Doc comment for `getStatus` could be more concise

The new doc comment on `getStatus` is thorough (good!), but the middle section:

```typescript
/// gateway runs the same `resolved_model_provider_for_agent` logic it uses to
/// build the Agent, so the returned `model` reflects that agent's configured
/// provider entry. Omitting the alias returns the install-wide first-of-each
/// summary (the gateway default model), which is NOT correct for any
/// non-default agent.
```

…repeats context that's already in the PR body and commit message. For code that will live in the repository long-term, a terser version would age better:

```typescript
/**
 * System status overview. Pass `agent` to get model/provider/temperature/memory
 * resolved for that specific agent (via gateway's `resolved_model_provider_for_agent`).
 * Omit `agent` for the install-wide summary (first-of-each defaults).
 */
```

The commit message already explains *why* this matters; the doc comment just needs to document *what* the parameter does. Not blocking — the current version is correct, just verbose.

---

### 🔵 Suggestion — Explicit `null`-check on provider resolution could clarify intent

In the new `switchModel` logic:

```typescript
const providerProp = await getProp(`agents.${agentAlias}.model_provider`);
const provider =
  typeof providerProp.value === 'string' ? providerProp.value.trim() : '';
if (!provider) {
  throw new Error(t('agent.failed_switch_model'));
}
```

The `trim()` → empty-string → throw path is correct, but the `!provider` check fires for both "the prop returned a non-string" and "the prop returned an empty/whitespace string." A reader might wonder: can `getProp` return a non-string for a string-typed config field?

Making the two cases explicit would clarify intent:

```typescript
const providerProp = await getProp(`agents.${agentAlias}.model_provider`);
if (typeof providerProp.value !== 'string' || !providerProp.value.trim()) {
  throw new Error(t('agent.failed_switch_model'));
}
const provider = providerProp.value.trim();
```

Not a bug — the current version works — but the explicit guard documents the invariant more clearly. (Per FND-006 §4.1: "The difference between `.unwrap()` and `.expect('reason')` is the difference between deferred judgment and documented judgment." Same principle applies to implicit vs. explicit guards.)

---

### Verdict

**No blocking issues.** The fix is technically sound, well-reasoned, and surgically scoped. The two 🟡 warnings (label mismatch, missing manual validation) are process/verification gaps, not code defects. The 🔵 suggestions are opportunities to tighten future maintainability, not current blockers.

Recommend **approval with label correction**: downgrade `risk: medium` → `risk: low` (maintainer action), then merge. The author has done the hard work — root-cause diagnosis, correct fix, thorough documentation — and the validation evidence (TypeScript + build checks) supports the stated scope.

singlerider (approved):
Reviewed at head `4866132` against current `master`, with focus on the gateway contract the fix depends on. The diagnosis and fix are sound and the backend assumptions check out. Approving.

### 🟢 The load-bearing backend assumptions are correct

I verified the two claims the fix rests on:

- **Agent-scoped status exists and resolves per-agent.** `GET /api/status?agent=<alias>` is supported (`api.rs` query param at line 190/232) and calls `config.resolved_model_provider_for_agent(alias)` (line 241) — the same resolution used to build the Agent. So `getStatus(agentAlias)` returns this agent's configured model, not the install-wide default. Passing the alias through is the right fix, and adding `agentAlias` to the effect deps is correct so the model reloads when the active agent changes.
- **The write path is the real one.** Writing to `providers.models.<provider>.model` (after resolving `agents.<alias>.model_provider`) matches how the agent resolves its model; the old top-level `getProp('model')`/`default_model` paths genuinely do not exist in the schema, so deleting those dead 404-ing round trips is a pure win.

### 🟢 The watchdog split and route parsing are well-built

Per-phase watchdog arming (write phase, then reconnect phase) with the `=== model` identity guard fixes the real bug: a slow write could exhaust a single combined budget and fire a false "switch timed out" while the switch was actually progressing, and nulling the pending ref would then skip the `onOpen` model update. The bail-before-touching-the-socket checks (`pendingModelSwitchRef.current !== model`) prevent tearing down a live socket after giving up or after a newer switch. `extractRouteModels` tolerating array / JSON-string / TOML-ish display is the right defense against the prop endpoint returning `value` as a display string with no `populated` flag.

### 🟡 Residual gap — no live end-to-end validation

Per @WareWolf-MoonWall's note, the PR states no live switch was run against a running daemon. For a user-facing fix the code audit is thorough and the contract is verified above, so this is not blocking, but a manual multi-agent switch confirmation would have fully closed the loop. The `risk: medium` label is auto-applied from the `src/**` (TypeScript) touch; the change is frontend-only and a maintainer can downgrade to `risk: low`.

Approving — the fix is surgical, the root cause is correctly diagnosed, and the gateway contract it relies on is real. The 🔵 maintainability suggestions (terser `getStatus` doc, explicit provider null-check, future `extractRouteModels` tests) are worth doing but not gating.

PR #7053, opened 2026-06-01
feat(memory): migrate Agent::turn load_context to MemoryStrategy

fanchanghu:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Replace Agent's `memory_loader: Box<dyn MemoryLoader>` with `memory_strategy: Arc<dyn MemoryStrategy>`, completing the second slice of the memory strategy decoupling (#6850).
  - Replace `AgentBuilder::memory_loader()` with `AgentBuilder::memory_strategy()`.
  - In `Agent::turn()` / `turn_streamed()`, call `memory_strategy.load_context()` instead of `memory_loader.load_context()`. Note: `MemoryStrategy` encapsulates its own `Memory` reference, so the caller no longer passes `memory.as_ref()`.
  - Default fallback in `AgentBuilder::build()` creates `DefaultMemoryStrategy::with_config()` from the builder's memory + workspace_dir.
  - Update `tests/support/helpers.rs`: introduce `StaticMemoryStrategy` and adjust `build_recording_agent` signature.
  - Update `tests/integration/agent.rs` to use `StaticMemoryStrategy`.
- **Scope boundary:** Does NOT migrate consolidation calls in gateway `ws.rs` or channel orchestrator — those are reserved for a follow-up slice.
- **Blast radius:** Agent builder API surface; downstream code constructing `Agent` via `AgentBuilder`.
- **Linked issue(s):** Related #6850
- **Labels:** `type: feat`, `risk: medium`, `size: S`

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test
```

- **Commands run and tail output:**
  - `cargo fmt --all -- --check`: passed (no output)
  - `cargo clippy --all-targets -- -D warnings`: passed — `Finished dev profile [unoptimized + debuginfo] target(s)`
  - `cargo test`: integration 159 passed; system 9 passed; doc-tests 0 passed
- **Beyond CI — what did you manually verify?** Verified `e2e_multi_turn_history_fidelity` and `e2e_memory_enrichment_injects_context` integration tests exercise the new `memory_strategy.load_context()` path end-to-end. Did NOT manually verify gateway or channel orchestrator paths (intentionally out of scope).
- **If any command was intentionally skipped, why:** None skipped.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? (`No`)
- New external network calls? (`No`)
- Secrets / tokens / credentials handling changed? (`No`)
[...]

PR #7192, opened 2026-06-04
fix(memory): honor recall since/until window for markdown entries

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** `MarkdownMemory::recall` filtered entries against the `since`/`until` window with `DateTime::parse_from_rfc3339(&entry.timestamp)`, but markdown `timestamp` values are file *stems* — a bare `YYYY-MM-DD` for daily logs, and `MEMORY.md` for the core file — never RFC 3339. So `parse_from_rfc3339` always returned `Err`, the `if let Ok(ts) = …` guards never fired, and the `since`/`until` window was **silently ignored** for every markdown entry, even though the recall tool advertises those bounds as a working filter. This adds a small `entry_in_window` helper that tries RFC 3339 first, then a `NaiveDate` (the daily-log stem) compared at day granularity, and otherwise leaves non-date stems (e.g. `MEMORY.md`) unfiltered so evergreen core memories still surface.
- **Scope boundary:** Only the markdown backend's recall windowing. No change to the SQL/Postgres backends (already RFC 3339), to the recall tool surface, scoring, or keyword filtering.
- **Blast radius:** `zeroclaw-memory` markdown recall path only. Evergreen core memory (`MEMORY.md`) is explicitly never window-filtered.
- **Labels:** `bug`, `size: S`, `risk: medium`, `memory`.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-memory --all-targets -- -D warnings
cargo test -p zeroclaw-memory --lib recall
```

```text
$ cargo fmt --all -- --check

$ cargo clippy -p zeroclaw-memory --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.95s
(no warnings)

$ cargo test -p zeroclaw-memory --lib recall
test result: ok. 40 passed; 0 failed; 0 ignored; 0 measured; 267 filtered out; finished in 0.71s
```

- **Beyond CI — what did you manually verify?** Confirmed the new regression fails before the fix (the window was a silent no-op that returned out-of-window daily entries) and passes after, and that an evergreen `MEMORY.md` entry is never dropped by the window.

## Compatibility (required)

- Backward compatible? `Yes` — the previous behavior was a silent no-op; recall now honors the `since`/`until` bounds the tool already documents.
- Upgrade steps: none. Daily entries outside an explicit `since`/`until` window are now correctly excluded (previously they leaked through); evergreen core memory is unaffected.

- **Feature flags or config toggles:** None; revert the PR.
- **Observable failure symptoms:** Markdown-backed recall with explicit `since`/`until` bounds includes out-of-window daily entries, excludes expected evergreen `MEMORY.md` entries, or reports unexpected date-window parsing errors.


Audacity88 (approved):
Checked: live PR state, empty review/comment history, the one-file `crates/zeroclaw-memory/src/markdown.rs` diff, current `origin/master` 5a3d29a55, the later #7053 memory-strategy merge, synthetic mergeability, and `git diff --check`.

I’m approving this. The later memory-strategy work does not change the `since`/`until` recall caller path this PR fixes.

### 🟢 What looks good — markdown recall now compares against the date shape it stores

The old filter tried to parse every markdown entry timestamp as RFC 3339, but markdown daily entries store a bare `YYYY-MM-DD` stem. That made bounded recall look supported while silently leaving daily entries unfiltered. The new helper keeps strict RFC 3339 parsing for caller-provided bounds, falls back to day-granularity comparison for daily stems, and leaves non-date core memory stems evergreen. That matches the backend's on-disk format without changing the public recall API.

The regression also proves the important behavior: an `until` bound before today's date drops today's daily entry, while the core `MEMORY.md` entry still survives the same bounded recall. That is the right coverage for this bug.

### 🟡 Warning — refresh final merge signal after the approval

CI passed on this head, and no commit after the branch point has touched `crates/zeroclaw-memory/src/markdown.rs`. Master has still moved since those checks ran, including nearby memory-strategy work. Before merging, I would either refresh/wait for current-base CI or explicitly accept the stale-check risk. I do not see a code blocker in the PR itself.

PR #7193, opened 2026-06-04
fix(config): default webhook channel port to 8090

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** `WebhookConfig.port` was the only `u16` port field in the schema with no `#[serde(default = …)]`. Every sibling embedded-server port defaults (`GatewayConfig.port`, WSS, IRC, LINE/VoiceCall `webhook_port`), and the docs (`docs/book/src/channels/webhook.md`) already show `port = 8090`. Because `port` was required, a webhook channel block written without it failed to load with `missing field \`port\``, blocking `zeroclaw agent` until the user hand-edited `config.toml`. This adds a `default_webhook_channel_port()` helper returning `8090` (matching the docs), wired via `#[serde(default = …)]`.
- **Scope boundary:** Only adds a default to the one missing field. No change to existing explicit-`port` behavior, to the webhook runtime, or to quickstart prompting.
- **Blast radius:** Deserialization of `channels.webhook` config blocks. Configs that already set `port` are byte-for-byte unaffected.
- **Labels:** Applied `bug`, `risk: medium`, `size: XS`, `config`.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-config --all-targets -- -D warnings
cargo test -p zeroclaw-config --lib webhook_config
```

```text
$ cargo fmt --all -- --check

$ cargo clippy -p zeroclaw-config --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4m 37s
(no warnings)

$ cargo test -p zeroclaw-config --lib webhook_config
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 712 filtered out; finished in 0.03s
```

- **Beyond CI — what did you manually verify?** Confirmed `WebhookConfig` now deserializes from `{}` with `port == 8090`, and that the existing explicit-port tests still pass unchanged.

## Compatibility (required)

- Backward compatible? `Yes` — existing configs that specify `port` are unchanged; only the previously-failing "no port" case now loads with the documented default.
- Config / env / CLI surface changed? `No` (the default matches the documented `8090`).
- Upgrade steps: none.

Medium-risk: `git revert <sha>`.


singlerider (approved):
I checked the DRY angle carefully, since at first glance this looks like it duplicates an existing `default_webhook_port()`. It doesn't — the two are genuinely separate concepts that happen to share the value `8090`. Approving.

### 🟢 Real bug, correct default, faithful to the docs

`WebhookConfig.port` was a required `u16` with no `#[serde(default)]`, so a `[channels_config.webhook]` block written without `port` failed deserialization with `missing field \`port\``, blocking `zeroclaw agent` until the user hand-edited `config.toml`. The new `default_webhook_channel_port()` returns `8090`, which matches `docs/book/src/channels/webhook.md:12` (`port = 8090`) exactly, so the default and the documented value agree. Sibling embedded-server ports already default, so this closes a real inconsistency.

### 🟢 Not a DRY violation — the existing helper is for a different field

The existing `default_webhook_port()` lives in `scattered_types.rs:655`, is **private** to that module, and serves the **VoiceCall** `webhook_port` field (the voice-call inbound server) — not the webhook *channel*. Reusing it here would require exporting it `pub(crate)` and cross-importing, and it would falsely couple two independent defaults: if the VoiceCall webhook port ever moves off 8090, the channel port should not move with it. A distinct, well-named helper co-located with `WebhookConfig` is the right call, not a shadowed registry. One nit on the PR body: it says the new helper matches "the existing `default_webhook_port()`" — same value, but that helper is for a different field; the framing slightly overstates the relationship. No code change needed.

### 🟢 Test proves the omitted-port path

`webhook_config_port_defaults_when_omitted` deserializes `WebhookConfig` from `{}` and asserts `port == 8090` — directly exercising the previously-failing case — and the existing explicit-port tests (`port: 8080`, retry-field defaults) are untouched, so the change is additive. CI is green across all 13 checks. 1 file, +11/−0.

### 🟢 Label note — applied `risk: medium` is correct

The body lists expected `risk: low`, but the applied label is `risk: medium` (auto-applied for `config`/schema changes), which is right for a schema deserialization change. No action — the applied label wins; flagging the drift for the record.

Approving. Correct default, faithful to the docs, and the apparent duplication is actually two distinct concepts kept properly independent.

Audacity88 (approved):
I checked the live PR state, the one-file config diff, SingleRider's prior approval, CI, and the PR body's validation evidence. This is a second approval on the same head.

### 🟢 What looks good — Config default matches the omitted-port failure

`WebhookConfig.port` previously had no serde default, so a webhook channel config block could fail deserialization with `missing field 'port'` when `port` was omitted. The diff adds `#[serde(default = "default_webhook_channel_port")]`, and the PR body plus prior review identify `8090` as the documented channel default. That fixes the omitted-field path without changing explicit-port behavior.

### 🟢 What looks good — Helper stays scoped to the webhook channel

I agree with SingleRider's DRY read: the existing `default_webhook_port()` is for `VoiceCallConfig.webhook_port`, not the webhook channel. Keeping `default_webhook_channel_port()` next to `WebhookConfig` avoids coupling two semantically separate server defaults that merely share `8090` today.

### 🟢 What looks good — Test and CI cover this change

The new `webhook_config_port_defaults_when_omitted` test deserializes `{}` into `WebhookConfig` and asserts `port == 8090`, while the existing explicit-port and retry-field tests remain intact. The PR is one config-schema file, CI is green, and the applied `risk: medium` label is appropriate for a config deserialization default.

Approving. This is a small, well-scoped fix for a real config-load mismatch.

PR #7195, opened 2026-06-04
fix(tool-call-parser): recognize plural <tool_calls> wrapper tag

### LINKED ISSUE #6875: Tool call parser does not handle <tool_calls> (plural) tag — Llama 4 Scout and similar models fail silently
## Version
0.8.0-beta-1 (verified in `crates/zeroclaw-tool-call-parser/src/lib.rs`)

## Summary

The XML tool call parser supports `<tool_call>` (singular) but not `<tool_calls>` (plural). Llama 4 Scout (`meta-llama/llama-4-scout-17b-16e-instruct`) generates plural tags. zeroclaw detects a "tool_call_parse_issue" warning, outputs the raw XML to the user, and makes zero tool calls.

## Source reference

```rust
// crates/zeroclaw-tool-call-parser/src/lib.rs
const TOOL_CALL_OPEN_TAGS: [&str; 6] = [
    "<tool_call>",      // singular ✅
    "<toolcall>",       // ✅
    "<tool-call>",      // ✅
    "<invoke>",         // ✅
    "<minimax:tool_call>",  // ✅
    "<minimax:toolcall>",   // ✅
    // <tool_calls> (plural) is NOT listed ❌
];
```

## Reproduction

Configure agent with `meta-llama/llama-4-scout-17b-16e-instruct` via Groq. Ask any tool-requiring question.

**Model output:**
```
<tool_calls>
{"name": "myserver__some_tool", "arguments": {"key": "value"}}
</tool_calls>
```

**zeroclaw log:**
```
WARN tool_call_parse_issue: response resembled a tool-call payload but no valid tool call could be parsed
INFO llm_response: {"native_tool_calls":0,"parsed_tool_calls":0,...}
```

**User sees:** the raw `<tool_calls>` XML output with no tool execution.

## Expected behavior

The parser should handle `<tool_calls>` (plural) as a wrapper that contains one or more `<tool_call>` (singular) entries. Many Llama-family models use this format.

## Proposed fix

Add `<tool_calls>` to `TOOL_CALL_OPEN_TAGS` with corresponding close tag `</tool_calls>`, and parse inner content as one or more tool call entries.

Or: add a pre-processing step that strips the `<tool_calls>` wrapper and passes the inner `<tool_call>` entries to the existing parser.

## Affected models

- `meta-llama/llama-4-scout-17b-16e-instruct` (Groq)
- Likely other Llama 4 variants
- Any model following the Llama function-calling format that uses plural wrapper tags

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** The XML tool-call scanner only recognized singular wrapper tags (`<tool_call>`, `<toolcall>`, `<tool-call>`, `<invoke>`, `<minimax:*>`), not the plural `<tool_calls>`. Because `find_first_tag` matches whole literals including the trailing `>`, `<tool_call>` does not substring-match inside `<tool_calls>` (the next char is `s`). So a model emitting `<tool_calls>\n{…}\n</tool_calls>` (Llama 4 Scout via Groq, and similar Llama-family models) was never entered: the parser found no tag, warned, and surfaced the raw XML with zero tool calls executed. This adds `<tool_calls>` / `</tool_calls>` to the open/close tag tables and the open→close mapping; the existing inner-body handling already covers the JSON / nested-`<tool_call>` payloads.
- **Scope boundary:** Only adds the plural wrapper recognition. No change to the singular tag paths, the JSON/native paths, or the inner-body parsing.
- **Blast radius:** XML tool-call extraction in `zeroclaw-tool-call-parser`. The plural wrapper is consumed as the outer tag, so existing singular `<tool_call>` matching is unchanged (distinct literals).
- **Linked issue(s):** Closes #6875.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-tool-call-parser --all-targets -- -D warnings
cargo test -p zeroclaw-tool-call-parser --lib plural_tool_calls
```

```text
$ cargo fmt --all -- --check

$ cargo clippy -p zeroclaw-tool-call-parser --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.58s
(no warnings)

$ cargo test -p zeroclaw-tool-call-parser --lib plural_tool_calls
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 131 filtered out; finished in 0.03s
```

- **Beyond CI — what did you manually verify?** Confirmed a `<tool_calls>…</tool_calls>` payload now parses to exactly one executed tool call with the correct name/arguments and empty residual text; the singular `<tool_call>`, JSON, and native paths are unchanged (full crate test suite green).

## Compatibility (required)

- Backward compatible? `Yes` — purely additive tag recognition; previously-handled inputs are unchanged, and a `<tool_calls>` payload that used to be shown as raw text now executes as intended.
- Upgrade steps: none.

- **Fast rollback command/path:** `git revert <merge-sha>` — change is contained to `crates/zeroclaw-tool-call-parser/src/lib.rs` (two array entries + one match arm + a test), no schema/config migration.
- **Observable failure symptoms:** Llama-family responses wrapped in `<tool_calls>…</tool_calls>` showing raw XML to the user with no tool execution; grep logs for the `tool_call_parse_issue` warning on Groq/Llama traffic.


Audacity88 (approved):
@Alix-007, I reviewed current head b86ee95, including the PR body, linked issue #6875, the one-file parser diff, visible green CI, and a local merge-result check against current `master`.

### 🟢 What looks good — narrow fix for the reported wrapper shape

This is the right-sized fix for #6875. Adding `<tool_calls>` / `</tool_calls>` to the XML wrapper tables and the open-to-close mapping lets the parser enter the plural Llama-family wrapper without changing the existing singular `<tool_call>` path, JSON/native path, or GLM fallback behavior.

The committed regression also checks the important user-visible contract: a plural wrapper that used to surface as raw XML now produces one parsed tool call and leaves no residual text.

### 🟢 What looks good — current-master validation is green

I checked the merge result locally against current `origin/master`. The PR merged cleanly, and the focused parser validation passed on that merge result:

- `cargo test -p zeroclaw-tool-call-parser --lib plural_tool_calls -- --nocapture` — 1 passed, 0 failed.
- `cargo test -p zeroclaw-tool-call-parser --lib -- --nocapture` — 132 passed, 0 failed.

I also did a local-only probe for multiple JSON objects inside a single `<tool_calls>...</tool_calls>` wrapper, since #6875 mentioned the multi-call wrapper shape. That passed too, and the existing inner-body scanner explains why: after the plural wrapper is recognized, it reuses the same JSON extraction path that already handles multiple objects.

### 🔵 Suggestion — committed multi-call coverage would be a nice follow-up

Not blocking this PR, but if you touch this parser again, a committed regression for two JSON objects inside one plural wrapper would make the issue’s full wording explicit in the test suite. The behavior is already covered by the implementation path and local probe, so I do not think this needs to hold the fix.

Approving. The parser change is narrow, the linked bug is covered, current-master validation is green, and I do not see an active blocker.

GonuDvc (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

PR #7196, opened 2026-06-04
fix(skills): keep blank lines inside SKILL.md block-scalar descriptions

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** `parse_simple_frontmatter` (the hand-rolled SKILL.md frontmatter parser that replaced `serde_yaml`) collected YAML block-scalar continuation lines (`description: >-` or `|`) only while the line started with a space or tab. A blank line inside a folded/literal block scalar is the empty string `""`, which matched neither check, so the parser stopped collecting at the first blank line. YAML treats a blank line as a paragraph break *inside* the scalar, so any multi-paragraph `description` was silently truncated at the first blank line — dropping discovery-relevant text that is injected into the system prompt. This treats a whitespace-only line as a continuation; the block scalar now terminates only on a real, non-indented next key.
- **Scope boundary:** Only the block-scalar continuation condition. No change to single-line key parsing, to the terminator behavior for a genuine next key, or to any other frontmatter handling.
- **Blast radius:** SKILL.md frontmatter parsing in `zeroclaw-runtime`. A regression test guards that a real next key still terminates the scalar.
- **Labels:** `bug`, `risk: low`, `size: XS`, `runtime`, `skills`

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-runtime --lib block_scalar
```

```text
$ cargo fmt --all -- --check

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.25s
(no warnings)

$ cargo test -p zeroclaw-runtime --lib block_scalar
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 1981 filtered out; finished in 0.00s
```

- **Beyond CI — what did you manually verify?** Confirmed a multi-paragraph `description: >-` now retains text after a blank line, and a second regression test confirms a real non-indented next key still terminates the block scalar.

## Compatibility (required)

- Backward compatible? `Yes` — single-paragraph descriptions parse identically; only previously-truncated multi-paragraph descriptions now parse in full.
- Upgrade steps: none.

Low-risk: `git revert <sha>`.


WareWolf-MoonWall (approved):
### 🟢 What looks good — surgical fix with strong validation evidence

This is exactly the kind of fix that demonstrates mature engineering discipline: proactive discovery of a subtle parser bug, a minimal change that addresses only the root cause, regression tests that guard both the fix and the boundary case, and validation evidence that proves the fix works.

The inline comment addition at L1093–1095 is particularly valuable — it explains the YAML block scalar semantics (`blank line is paragraph break *inside* the scalar, not a terminator`) so the next person who touches this code understands *why* `line.trim().is_empty()` belongs in the continuation check. That's the difference between code that works and code that's maintainable.

The two new tests cover the exact failure mode (multi-paragraph description with blank line) and the critical invariant (real non-indented next key still terminates the scalar). That's complete coverage for the change scope.

### 🟡 Warning — label mismatch between auto-tagger and actual risk

The `risk: high` label is auto-applied because the path contains `runtime`, but this change carries none of the characteristics that make runtime changes high-risk:

- It's a self-contained parser function with no interaction with security boundaries, the agent loop, or external I/O.
- The blast radius is SKILL.md frontmatter parsing — discovery text for skill invocation, not execution logic.
- The change is backward compatible (single-paragraph descriptions parse identically; only previously-truncated multi-paragraph descriptions now parse correctly).
- Full test coverage is present, and the validation evidence shows clean local runs.

Per the project's risk-tier guidelines (AGENTS.md), high-risk changes involve `src/security/`, access-control boundaries, gateway, tools, or CI/workflows. This is a low-risk bugfix — it should carry `risk: low` and `type: fix` as you requested in the PR description.

I don't have label permissions, but @Audacity88 or @JordanTheJet can correct this before merge. It's not a merge blocker — just a metadata accuracy issue that will help future archaeology.

### 🟢 Test design matches the problem space

The two new tests are well-named and address the right concerns:

1. `parse_simple_frontmatter_keeps_blank_line_in_block_scalar` — proves the bug is fixed (multi-paragraph description now retains text after blank line).
2. `parse_simple_frontmatter_block_scalar_stops_at_next_key` — proves the fix doesn't break the termination condition (non-indented next key still ends the scalar).

That second test is the one many contributors would skip, and it's the one that would catch a regression if someone later changed the condition without understanding the YAML semantics. Good judgment call to include it.

---

**Verdict:** Approved. The fix is correct, minimal, well-tested, and addresses a real (if subtle) bug that would silently truncate skill discovery text. The label mismatch is metadata-only and doesn't block merge.

**Reviewed head commit:** The PR is currently at a single commit; this review applies to the state shown in the diff as of 2026-06-04T03:37:17Z.

cc @Alix-007 — this is ready to land once the label hygiene is addressed. Nice catch on the blank-line truncation bug.

Audacity88 (approved):
I checked the live PR state at head `6110693`, WareWolf-MoonWall's existing approval review, the public diff, the PR body, and the visible CI/check state. This approval is an independent code-read confirmation: the parser change is narrow, the tests cover both the bug and the key boundary, and I do not see a remaining code blocker.

### 🟢 What looks good — Blank-line handling is fixed at the parser boundary

The continuation change matches the YAML block-scalar rule this lightweight parser is trying to preserve: blank or whitespace-only lines stay inside the scalar, while a real non-indented next key still ends the scalar. The paired tests cover both sides of that boundary: the previously truncated multi-paragraph `description` now keeps text after the blank line, and the next-key regression still parses `version` separately.

### 🟢 What looks good — The fix stays scoped to discovery text

This keeps the patch exactly where the bug lives: `parse_simple_frontmatter` and its local regression tests. It does not broaden skill execution behavior, change plugin loading, or alter unrelated frontmatter parsing paths, which makes the fix straightforward to reason about and safe to land.

I'm approving the code change. The remaining label/risk cleanup is maintainer metadata, not author action.

PR #7199, opened 2026-06-04
fix(web): strip server timestamp prefix from chat bubble content

### LINKED ISSUE #7157: [Bug]: Chat message timestamp `[2026-06-03 18:18:31 +08:00]` rendered inside the message bubble instead of as separate metadata
### Affected component

runtime/daemon

### Severity

S3 - minor issue

### Current behavior

In the web chat interface, the message timestamp is shown as a literal prefix **inside** the message bubble, concatenated with the message text:

```
[2026-06-03 15:08:12 +08:00] search zeroclaw news and save the roadmap to a document.
```

The `[2026-06-03 18:18:31 +08:00]` timestamp string is part of the bubble content rather than being rendered as separate message metadata outside the bubble.

### Expected behavior

The timestamp should be rendered as message metadata **outside** the message bubble/tag (e.g. as a caption above or beside the bubble), not embedded in the message text content itself. The bubble should contain only the actual message text.

### Steps to reproduce

```bash
1. Open the web chat interface
2. Send (or view) a chat message, e.g. "search zeroclaw news and save the roadmap to a document."
3. Restart the daemon (Reload daemon) and refresh the page
4. Observe the rendered message bubble
5. The bubble text reads: "[2026-06-03 18:18:31 +08:00] search zeroclaw news ..."
   — the timestamp appears inside the bubble instead of outside it
```

### Impact

Affected users: anyone using the web chat UI
Frequency: always
Consequence: cosmetic/readability — timestamp clutters message content; not a functional blocker

### Logs / stack traces

_No response_

### ZeroClaw version

commit 40be7738f

### Rust version

### Operating system

Linux 7.0.10-1-MANJARO x86_64

### Regression?

Unknown

### Pre-flight checks

xianshishan:
## Summary

Fix the server timestamp prefix `[YYYY-MM-DD HH:MM:SS TZ]` being rendered inline inside chat bubbles in the Web UI.

**Approach**: Pure frontend fix. Strip the timestamp prefix from message content at render time using a regex in `MessageItem`, and display only the clean content. Copy also uses clean content so users don't paste the timestamp prefix.

## Changes

- `web/src/pages/AgentChat.tsx` — add `extractServerTimestamp()` helper + use `cleanContent` for rendering and copy in `MessageItem`

## Design notes

- **No `^` anchor**: The timestamp may appear at the start (no context) or mid-string (after memory context), so the regex matches anywhere
- **Compact mode unaffected**: Timestamp is hidden in compact mode, content stripping still applies
- **Safe fallback**: Failed date parsing or no regex match silently returns the original content
- **Bottom timestamp unchanged**: Preserves the original master behavior of showing `msg.timestamp` at the bubble bottom in non-compact mode

## Test plan

- [ ] Send a message in Web Chat, refresh the page — bubble content does not contain the timestamp prefix
- [ ] Copy message — clipboard does not include the timestamp prefix
- [ ] Compact mode — messages render correctly (no timestamp shown)
- [ ] Messages with memory context — timestamp prefix is still correctly stripped

Closes #7157

### DIFF
--- a/web/src/pages/AgentChat.tsx
+++ b/web/src/pages/AgentChat.tsx
@@ -453,6 +453,23 @@ function AgentChatInner({ agentAlias }: { agentAlias: string }) {
   );
 }
 
+const SERVER_TIMESTAMP_RE = /\[(\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2} [^\]]+)\]\s*/;
+
+function extractServerTimestamp(content: string): {
+  cleanContent: string;
+  serverTimestamp: Date | null;
+} {
+  const m = content.match(SERVER_TIMESTAMP_RE);
+  if (!m) return { cleanContent: content, serverTimestamp: null };
+  const tsStr = m[1];
+  if (!tsStr) return { cleanContent: content, serverTimestamp: null };
+  const ts = new Date(tsStr);
+  if (isNaN(ts.getTime())) return { cleanContent: content, serverTimestamp: null };
+  const prefix = content.slice(0, m.index!);
+  const suffix = content.slice(m.index! + m[0].length);
+  return { cleanContent: prefix + suffix, serverTimestamp: ts };
+}
+
 // Each chat message is rendered through this memoized component so that
 // typing into the input does not re-render every existing message (and
 // re-run ReactMarkdown on each one). Keep the prop surface small and pass
@@ -475,6 +492,8 @@ const MessageItem = memo(function MessageItem({
   onCopy,
   onDelete,
 }: MessageItemProps) {
+  const { cleanContent } = extractServerTimestamp(msg.content);
+
   return (
     <div
       className={`group flex items-start ${compact ? 'gap-2' : 'gap-3'} ${
@@ -515,9 +534,9 @@ const MessageItem = memo(function MessageItem({
           {msg.toolCall ? (
             <ToolCallCard toolCall={msg.toolCall} />
           ) : msg.markdown ? (
-            <div className={`${compact ? 'text-xs' : 'text-sm'} break-words leading-relaxed chat-markdown`}><ReactMarkdown remarkPlugins={[remarkGfm]}>{msg.content}</ReactMarkdown></div>
+            <div className={`${compact ? 'text-xs' : 'text-sm'} break-words leading-relaxed chat-markdown`}><ReactMarkdown remarkPlugins={[remarkGfm]}>{cleanContent}</ReactMarkdown></div>
           ) : (
-            <p className={`${compact ? 'text-xs' : 'text-sm'} whitespace-pre-wrap break-words leading-relaxed`}>{msg.content}</p>
+            <p className={`${compact ? 'text-xs' : 'text-sm'} whitespace-pre-wrap break-words leading-relaxed`}>{cleanContent}</p>
           )}
           {!compact && (
             <p
@@ -528,7 +547,7 @@ const MessageItem = memo(function MessageItem({
         </div>
         <div className="flex items-center justify-end gap-1 mt-1 opacity-0 group-hover:opacity-100 transition-opacity">
           <button
-            onClick={() => onCopy(msg.id, msg.content)}
+            onClick={() => onCopy(msg.id, cleanContent)}
             aria-label={t('agent.copy_message')}
             className="p-1 rounded-lg"
             style={{ color: 'var(--pc-text-muted)' }}


xianshishan:
The effect of compact mode and non-compact mode is that the timestamps of the dialogue disappear, enhancing the aesthetic appeal.
image
image

WareWolf-MoonWall (changes requested):
### 🔴 Blocking — Root cause not addressed; treating symptom instead of disease

**Problem:** The timestamp prefix `[YYYY-MM-DD HH:MM:SS TZ]` is appearing in message content because the backend is incorrectly injecting it into the `content` field during persistence or retrieval. This PR strips it at render time in the frontend.

**Why this is blocking:**

1. **Single Source of Truth violation (AGENTS.md absolute rule):** The `created_at` timestamp already exists as structured metadata in the SQLite schema (`crates/zeroclaw-infra/src/session_sqlite.rs` L299-301, L268). The gateway API returns it separately as `created_at` (L1493 in `api.rs`). Prefixing it into `content` creates duplicate state — the timestamp now lives in two places (structured column + stringified prefix), and this PR adds a third (frontend regex extraction). Per AGENTS.md: *"No piece of state lives in two places. Ever. Anywhere in this codebase."*

2. **Copy-paste pollution:** The author states "Copy also uses clean content so users don't paste the timestamp prefix." This is a red flag — if we need to strip something at copy time, that data shouldn't be in the content field in the first place. The correct fix is to prevent the timestamp from entering `content` at the source.

3. **Incomplete fix:** This only addresses the Web UI. Any other consumer of the session messages API (`GET /api/sessions/{id}/messages`) will still see the timestamp prefix in `content`. The CLI, future mobile clients, API integrations, memory recall — all downstream consumers inherit the pollution.

4. **Fragile regex maintenance:** The PR adds `SERVER_TIMESTAMP_RE` with no anchor (`^`) and notes "The timestamp may appear at the start (no context) or mid-string (after memory context)." This is symptomatic reasoning — the regex has to guess where the timestamp might appear because the backend is inconsistently injecting it. The correct fix eliminates the need for the regex entirely.

**What should happen instead:**

Find where the backend is prefixing the timestamp into `content` and stop it. The `created_at` column in SQLite already stores this data (L268-281 in `session_sqlite.rs`). The gateway API already returns it as structured JSON (L1490-1495 in `api.rs`). The frontend should consume `created_at` from the API response, not parse it out of `content`.

**Audit path:** Search for where message content is formatted during:
- Persistence (`append` in session backends)
- Retrieval (`load_with_timestamps` in session backends)
- Memory context injection (if memory entries are prepended to messages)

The timestamp prefix doesn't appear in the SQLite schema or the API contract — which means it's being added somewhere between the message arriving at the gateway and being stored, or during memory context concatenation. Trace the `content` field from the WebSocket `chat` frame handler through to `SessionBackend::append`.

**Resolution:** Either fix the root cause in this PR, or close this PR and open a new one that addresses the backend. Do not merge a frontend workaround that institutionalizes a backend bug.

---

### 🟡 Warning — PR description claims "No `^` anchor" is intentional, but that's accommodating broken data

The PR notes: *"No `^` anchor: The timestamp may appear at the start (no context) or mid-string (after memory context), so the regex matches anywhere."*

This statement reveals the regex is compensating for inconsistent backend behavior. If timestamps appear "after memory context," that means memory retrieval is injecting them into content mid-string. That's a backend bug. The regex shouldn't need to handle both cases — the timestamp should never be in `content` at all.

If memory context is involved, check whether `zeroclaw-memory` backends are formatting entries with timestamps before concatenating them into agent prompts. If so, that's where the fix belongs.

---

### 🔵 Suggestion — If you must ship a temporary workaround, gate it with a feature flag and track removal

If there's pressure to ship a quick visual fix while the backend investigation continues (I don't recommend this, but I understand production urgency):

1. Add a Cargo feature flag `temp-timestamp-strip-workaround` (or similar).
2. Wrap the `extractServerTimestamp` logic in a conditional compile guard.
3. File a tracked issue with priority P1 assigned to whoever owns session persistence.
4. Add a TODO comment linking the issue and stating the flag will be removed in v0.8.0.
5. Add the removal to the v0.8.0 milestone.

This makes the technical debt explicit and time-boxed. It also ensures the codebase doesn't drift into accepting this as permanent behavior.

But I still recommend fixing the root cause first.

---

### 🟢 What looks good — Clean extraction helper and safe fallback

The `extractServerTimestamp` implementation is well-structured:

- **Safe parsing:** Returns the original content if the regex doesn't match or if date parsing fails. No crashes on malformed data.
- **Minimal footprint:** The helper is scoped to the render layer (`MessageItem`), not spread across multiple files.
- **Consistent application:** Both render paths (markdown + plain text) use `cleanContent`, and the copy handler uses it too, so behavior is uniform.

If this were addressing the right problem, the implementation quality would be solid. The issue is the problem itself — treating a backend data integrity issue with a frontend cosmetic patch.

---

### 🔵 Suggestion — Add a test for the extraction logic

If this ships (again, I don't think it should), add a test for `extractServerTimestamp`:

```typescript
describe('extractServerTimestamp', () => {
  it('strips timestamp prefix from content', () => {
    const input = '[2026-06-03 15:08:12 +08:00] search zeroclaw news';
    const { cleanContent, serverTimestamp } = extractServerTimestamp(input);
    expect(cleanContent).toBe('search zeroclaw news');
    expect(serverTimestamp).toBeInstanceOf(Date);
  });

  it('handles timestamp mid-string', () => {
    const input = 'Memory context: [...]\n[2026-06-03 15:08:12 +08:00] user message';
    const { cleanContent } = extractServerTimestamp(input);
    expect(cleanContent).toBe('Memory context: [...]\nuser message');
  });

  it('returns original content if no match', () => {
    const input = 'no timestamp here';
    const { cleanContent, serverTimestamp } = extractServerTimestamp(input);
    expect(cleanContent).toBe(input);
    expect(serverTimestamp).toBeNull();
  });

  it('handles malformed timestamp gracefully', () => {
    const input = '[not-a-date] content';
    const { cleanContent, serverTimestamp } = extractServerTimestamp(input);
    expect(cleanContent).toBe(input);
    expect(serverTimestamp).toBeNull();
  });
});
```

---

### Verdict

**Request changes.** This PR treats a symptom (timestamp prefix visible in UI) instead of the disease (backend incorrectly injecting timestamps into `content`). Merging this institutionalizes a Single Source of Truth violation and leaves all non-Web consumers of the session messages API broken.

Fix the root cause: find where the backend is prefixing timestamps into `content` and stop it. The `created_at` column already exists in SQLite. The API already returns it. The frontend should consume it as structured metadata, not parse it out of polluted content.

Reviewed head commit: `fix/web-chat-timestamp-in-bubble-7157` (commit hash not visible in `gh pr diff` output; @xianshishan please confirm latest SHA when you address this).

---

**Note to @xianshishan:** I appreciate the clean implementation and the effort to document the design notes. The execution is solid — the issue is the approach. Backend data integrity problems should be fixed at the source, not papered over in the UI. If you need help tracing where the timestamp prefix is being added, ping @WareWolf-MoonWall and I can pair on the investigation.

singlerider (commented):
Reviewed at head `1defdbd` against current `master`, and traced the root cause. @WareWolf-MoonWall is right that this is a frontend symptom-patch; agreeing with his changes-requested and adding the exact source location plus one nuance.

### 🟢 The extraction helper itself is clean

`extractServerTimestamp` is safe (returns original content on no-match or unparseable date), scoped to the render layer, and applied consistently across the markdown/plain/copy paths. If this were the right layer, the implementation quality would be fine.

### 🔴 Blocking — agree: fix the source, not the render layer; here is the source

I traced where the prefix enters `content`: `crates/zeroclaw-channels/src/orchestrator/mod.rs:969` does `format!("[{}] {}", now.format("%Y-%m-%d %H:%M:%S %Z"), content)` (pinned by the test `timestamp_channel_user_content_adds_wall_clock_prefix`). That prefixed string is what gets persisted to the session store, which is why it surfaces in the chat bubble via `/api/sessions/{id}/messages`.

One nuance that refines @WareWolf-MoonWall's framing: the prefix is **intentional** — the agent is deliberately given the wall-clock time inside the user message so it knows "now." So the fix is not simply deleting line 969 (the agent needs that context). The right boundary is persistence/retrieval: store and/or return the display content without the agent-facing wall-clock prefix, and let the UI use the structured `created_at` the API already returns. That keeps the single source of truth (`created_at` column) and fixes every consumer at once, instead of a Web-only regex that the CLI, API integrations, and memory recall don't benefit from. As @WareWolf-MoonWall notes, the unanchored regex and the "appears mid-string after memory context" case are both tells that the frontend is guessing at where the backend injected the prefix.

So: either move the fix to the gateway/session-retrieval boundary so `content` returned for display is clean (the agent-loop path keeps its prefixed copy), or close this and open a backend PR that does. A frontend regex institutionalizes the duplication.

### 🔵 If a stopgap is truly needed

If there is real pressure to ship a visual fix first, gate it explicitly and time-box it with a tracked issue pointing at `orchestrator/mod.rs:969` and the retrieval boundary, so it does not become permanent. But the backend fix is the correct one.

Implementation is solid; the layer is wrong. Leaving @WareWolf-MoonWall's changes-requested in force.

Audacity88:
@xianshishan quick status update: this PR is still waiting on the requested author follow-up. The last author activity I see is from June 4, and the remaining blocker was reaffirmed after that.

Please update the branch or reply by June 30, 2026 if you want to keep this PR open. If there is no author response or branch update by then, we may close this PR as stale / needs-refresh. A fresh PR against current `master` is welcome if that is easier.

PR #7207, opened 2026-06-04
fix(gateway): point paircode recovery hint at loopback for non-loopback binds

### LINKED ISSUE #6561: fix(gateway): non-loopback --host recovery hint advertises admin URL that admin guard rejects
## Follow-up from #6192 review

**Context:** [PR #6192](https://github.com/zeroclaw-labs/zeroclaw/pull/6192) added `--host` / `--port` overrides to `zeroclaw gateway get-paircode` plus a startup-banner recovery hint, fixing the default #5266 case (alternate-port and random-port gateways bound to localhost, with or without `path_prefix`).

While re-reviewing #6192, @Audacity88 flagged a real edge case that we agreed to carry as a follow-up rather than block the merge:

> The new banner helper includes `--host 192.168.1.20` for a specifically bound non-loopback gateway, and the fallback `curl` also uses the bound host directly. The existing admin paircode routes are localhost-only, so a request that reaches the gateway through a non-loopback address is expected to get the existing `403 Admin endpoints are restricted to localhost` response.

## Repro shape

1. Start a gateway bound to a non-loopback address, e.g. `cargo run -- gateway start --host 192.168.1.20 --port <port>` with `[gateway].require_pairing = true`.
2. Read the startup banner. The recovery hint advertises:
   - `zeroclaw gateway get-paircode --new --host 192.168.1.20 --port <port>`
   - `curl -s -X POST http://192.168.1.20:<port>/admin/paircode/new`
3. Following either hint from another machine (or even from the same host but routed through `192.168.1.20` rather than `127.0.0.1`) hits the localhost-only admin guard:
   - `403 Admin endpoints are restricted to localhost`

## What's broken about it

The banner / CLI now suggest a recovery path that doesn't actually work for non-loopback admin reachability. For the default loopback case the hint is correct; the gap is specifically when the operator binds the gateway to a non-loopback interface.

## Possible directions (not prescriptive)

Per @Audacity88's review note, any of the following would close the loop:

1. **Normalize advertised admin recovery host to loopback** in the banner / CLI hint, even when the gateway is bound non-loopback — i.e. always say `--host 127.0.0.1` (or `--host localhost`) for the recovery hint, since admin endpoints are localhost-only. The bound non-loopback host stays correct for the *non-admin* parts of the gateway URL.
2. **Documentation-only fix:** add a `--host` doc/help-text note clarifying that the override must be loopback-reachable for admin recovery, and update the banner copy to say so.
3. **Add a regression locking the intended public-bind behavior** — e.g. assert that the admin paircode routes return 403 when reached via a non-loopback origin, so any future change to that policy fails loudly.

A combination is reasonable too — option 1 is the most user-friendly, option 3 is the cheapest insurance against silent regressions.

## Scope boundary

- This is a UX / hint-correctness gap, not an admin-route security gap. The 403 itself is correct behavior.
- Should not change `PairingGuard`, the localhost-only admin policy, or the loopback default for `[gateway].host`.

## Risk

Low — banner copy + CLI help-text change at most, with optional regression test. No change to the admin-route guard itself.

cc @Audacity88

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** Admin paircode routes are localhost-only (`require_localhost` returns `403` for any non-loopback peer). The startup-banner recovery hints previously advertised the bound host verbatim, so a gateway bound to a non-loopback or unspecified bind address could print recovery commands that the admin guard rejects. This keeps the CLI hint on its loopback default and makes the curl fallback normalize bind-only/non-loopback hosts to a usable loopback target.
- **Refresh status:** Refreshed the branch against current `origin/master` after reviewer feedback.
- **Scope boundary:** Only the paircode recovery hint helpers (`format_paircode_recovery_command`, `format_paircode_recovery_curl`, and the curl host normalizer) plus unit regressions. `require_localhost`, the admin guard, and gateway bind configuration are untouched.
- **Blast radius:** Startup banner paircode-recovery hint text only.
- **Linked issue(s):** Closes #6561.
- **Labels:** Expected `type: fix`, `risk: low`, `size: S`, `gateway` (no label permission; please adjust).

```bash
cargo fmt --all
git diff --check
cargo test -p zeroclaw-gateway paircode_recovery
```

```text
$ cargo fmt --all
(no output)

$ git diff --check
(no output)

$ cargo test -p zeroclaw-gateway paircode_recovery
running 7 tests
test tests::paircode_recovery_curl_normalizes_unspecified_bind_hosts ... ok
test tests::paircode_recovery_curl_preserves_actual_loopback_hosts ... ok
test tests::paircode_recovery_curl_preserves_path_prefix ... ok
test tests::paircode_recovery_curl_targets_running_instance ... ok
test tests::paircode_recovery_command_includes_alternate_port ... ok
test tests::paircode_recovery_command_includes_specific_host_when_needed ... ok
test tests::paircode_recovery_command_uses_loopback_for_nonloopback_host ... ok

test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 246 filtered out; finished in 0.06s
```

- **Beyond CI — what did you manually verify?** Confirmed non-loopback hosts normalize to `127.0.0.1`, unspecified bind hosts (`0.0.0.0`, `::`) also normalize to `127.0.0.1`, and actual loopback hosts (`127.0.0.1`, `localhost`, `::1`) remain usable for the curl fallback (`::1` is emitted as `[::1]` in the URL).
- **If any command was intentionally skipped, why:** Full workspace tests were skipped to keep this review-scoped; the touched helper is covered by the targeted gateway unit tests above.

## Compatibility (required)

- Backward compatible? `Yes` — loopback recovery remains usable; non-loopback/unspecified bind recovery hints now point at the localhost-only admin endpoint instead of an address it rejects.
- Upgrade steps: none.

- **Fast rollback command/path:** `git revert <merge-sha>` — contained to `crates/zeroclaw-gateway/src/lib.rs` helper/test changes.
- **Observable failure symptoms:** A gateway bound to a non-loopback or unspecified interface prints a paircode-recovery curl/CLI hint that cannot reach the localhost-only admin endpoint.


Audacity88:
@Alix-007 this is the right direction for #6561, but please refresh it against current `master` before review. The head is now 188 commits behind, and the green checks are from June 4.

While refreshing, please also check the unspecified-bind cases. The new curl helper preserves `0.0.0.0` and `::`, but those are bind addresses, not loopback admin recovery targets. Since the admin paircode route is localhost-only, the curl fallback should probably normalize unspecified and non-loopback bind hosts to a usable loopback host too, while preserving only actual loopback hosts such as `127.0.0.1`, `localhost`, and `::1`. A regression for `0.0.0.0` / `::` would make that boundary clear.

Alix-007:
Thanks for the concrete refresh guidance. I pushed a follow-up that:

- refreshes the branch against current `master`
- normalizes unspecified bind hosts (`0.0.0.0`, `::`) to `127.0.0.1` for the curl recovery fallback
- preserves actual loopback hosts (`127.0.0.1`, `localhost`, and `::1`; `::1` is bracketed for a valid URL)
- adds regressions for the unspecified-bind boundary

Local validation passed:

```text
cargo fmt --all
git diff --check
cargo test -p zeroclaw-gateway paircode_recovery

7 passed; 0 failed; 246 filtered out
```

Alix-007:
The latest Quality Gate failure appears to be runner infrastructure rather than a code/test failure. The failed `Test` check annotation reports:

```text
System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/2.335.1/_diag/Worker_20260612-073652-utc.log'
```

All other Quality Gate jobs on this run passed. I do not have permission to rerun failed jobs (`Must have admin rights to Repository`), so a maintainer rerun of the failed `Test` job should be enough if the next runner has disk available.

Alix-007:
I refreshed the branch against current `master` and pushed the merge-forward so CI can rerun on a fresh head after the previous runner disk exhaustion.

Local validation on the refreshed head:

```text
cargo test -p zeroclaw-gateway paircode_recovery

7 passed; 0 failed; 246 filtered out
```

Audacity88 (approved):
@Alix-007, I reviewed c1f9f64 against the live PR state, #6561, the earlier refresh request, the current diff/source around the gateway paircode helpers and localhost admin guard, the PR body's focused `zeroclaw-gateway` validation, and current green GitHub checks. I did not run an additional local gateway smoke test; this approval is scoped to the reviewed head only and is not a merge-readiness signal because current master has moved 11 commits ahead.

### ✅ Resolved — recovery hints now target the localhost-only admin surface

The prior issue was that the banner could advertise a non-loopback `--host` or curl URL even though `/admin/paircode/new` is guarded by `require_localhost()`. This head keeps that guard unchanged. It removes `--host` from the CLI recovery command so the CLI uses its loopback default, and it normalizes curl recovery hosts to `127.0.0.1` unless the configured host is already a real loopback target. That matches #6561's intended scope: fix the operator hint, not the admin-route security policy.

### 🟢 What looks good — the boundary is pinned by targeted regressions

The tests cover the important shapes from the review thread: non-loopback binds, unspecified bind hosts (`0.0.0.0` and `::`), actual loopback hosts (`127.0.0.1`, `localhost`, and bracketed `::1`), alternate ports, and path prefixes. That is the right regression set for this helper-level fix.

### 🟡 Warning — branch is 11 commits behind master

Current master is 11 commits ahead of this head. I did not see `crates/zeroclaw-gateway/src/lib.rs` change after this PR's merge base, but this still touches gateway Rust code and the branch is more than 10 commits behind. Before merge, refresh the branch and let CI rerun, or run an equivalent targeted merge-result smoke for `zeroclaw-gateway` and reassess the resulting head.

singlerider (approved):
Reviewed at head `c1f9f64` against current `master`, verifying the security premise the fix rests on. Approving.

### 🟢 The localhost-only premise is real, so the fix is correct

I confirmed the guard: `handle_admin_paircode_new` calls `require_localhost(&peer)?` as its first line (`lib.rs:3680`), and `/admin/paircode` is guarded the same way. So a recovery hint advertising a non-loopback `--host` or curl URL would point operators at an address the admin guard answers with `403`. Omitting `--host` (CLI falls back to its loopback default) and normalizing the curl host to `127.0.0.1` — while preserving explicit loopback hosts and bracketing `::1` as `[::1]` for valid IPv6 URL syntax — is exactly the right fix for #6561, and it correctly leaves the admin-route security policy untouched.

### 🟢 Regression set is the right shape

The tests pin the boundary across the cases from the review thread: non-loopback binds, unspecified binds (`0.0.0.0` / `::`), real loopback hosts (`127.0.0.1` / `localhost` / `[::1]`), alternate ports, and path-prefix preservation. The `_host` parameter kept for call-site symmetry with the curl helper is reasonable and documented.

CI is green and the branch is `MERGEABLE` / `CLEAN` (the earlier "11 commits behind" note has since resolved; `lib.rs` did not conflict). @Audacity88 approved at this same head. Approving — correct, well-tested, and the security boundary is verified.

Issue #4049, opened 2026-03-20
[Bug]: 5-field cron weekday numbers use Sun-Thu semantics for 1-5, while docs/examples imply Mon-Fri


whtiehack:
### Affected component
unknown

### Severity
S2 - degraded behavior

### Current behavior
ZeroClaw accepts 5-field cron expressions and describes them as "standard 5-field cron syntax", but numeric weekday values are passed through to the `cron` crate unchanged.

As a result, `1-5` is interpreted as Sunday through Thursday, not Monday through Friday.

This is especially confusing because the current user-facing examples use `0 9 * * 1-5` as a weekday/workweek example in multiple places, which strongly suggests standard crontab semantics.

### Expected behavior
For the user-facing 5-field cron interface, ZeroClaw should either:

1. Normalize numeric weekday values to standard crontab semantics before handing them to the parser, so `1-5` means Monday through Friday, or
2. Explicitly document the actual numeric weekday mapping (`1-7 = Sun-Sat` in the current parser) and replace misleading examples with named weekdays such as `MON-FRI` or numeric values that match the actual behavior.

At minimum, the current examples should not imply that `1-5` means Monday through Friday if the runtime interprets it as Sunday through Thursday.

### Common Linux cron semantics (evidence)
This differs from common Linux cron behavior.

For example:

- Debian `crontab(5)` defines day-of-week as `0-7`, with `0` or `7` meaning Sunday, and includes `0 22 * * 1-5` as a weekday example: https://manpages.debian.org/testing/cron/crontab.5.en.html
- Oracle Linux documentation defines day-of-week as `0-7` with Sunday as `0` or `7`, and gives `1-5` as a weekdays example: https://docs.oracle.com/cd/E77565_01/monitoring/OL7-MONITORING.pdf

That means on normal Linux cron implementations, `1-5` is understood as Monday through Friday, not Sunday through Thursday.

### Steps to reproduce
```bash
# On latest master, user-facing examples currently suggest this is a weekday/workweek schedule:
zeroclaw cron add '0 9 * * 1-5' 'Good morning' --tz America/New_York --agent
```

Then inspect the implementation path:

1. `src/cron/schedule.rs` accepts 5-field expressions and rewrites them to 6-field form by prepending seconds, then passes them directly to `cron::Schedule::from_str(...)`.
2. The locked dependency is `cron = 0.15.0`.
3. In that crate, `DaysOfWeek` maps numeric weekdays as:
[...]

PR #7208, opened 2026-06-04
fix(cron): correct weekday range normalization for Sunday-alias endpoints

### LINKED ISSUE #6422: cron_add: improve schedule parameter error message for plain string input
## Problem

When the LLM passes the `schedule` parameter to `cron_add` as a plain string (e.g., `"0 9 * * 1-5"`) instead of the required JSON object format (e.g., `{"kind":"cron","expr":"0 9 * * 1-5"}`), the error message is unhelpful:

```
Invalid schedule: invalid type: string, expected internally tagged enum Schedule
```

This is a serde deserialization error that smaller models (like gemma4:26b) cannot interpret or self-correct from. The LLM retries with the same bad format, leading to infinite retry loops and 300+ second LLM calls that hit message timeouts.

## Root Cause

The tool schema at `crates/zeroclaw-runtime/src/tools/cron_add.rs` correctly documents the `oneOf` format with examples:
```json
{"kind":"cron","expr":"0 9 * * 1-5","tz":"America/New_York"}
{"kind":"at","at":"2025-12-31T23:59:00Z"}
{"kind":"every","every_ms":3600000}
```

However, the `deserialize_maybe_stringified` helper in `crates/zeroclaw-runtime/src/cron/types.rs` only handles double-serialized JSON (stringified JSON objects). It cannot fix a plain string that is not JSON at all.

## Proposed Fix

In `crates/zeroclaw-runtime/src/tools/cron_add.rs`, before calling `deserialize_maybe_stringified::<Schedule>(v)`, add a pre-validation check:

```rust
if let Some(Value::String(s)) = args.get("schedule") {
    let s = s.trim();
    if !s.starts_with({) {
        return Ok(ToolResult {
            success: false,
            output: String::new(),
            error: Some(format!(
                "Invalid schedule: expected a JSON object with \"kind\" field, got plain string \"{}\". Examples: {{\"kind\":\"cron\",\"expr\":\"0 9 * * 1-5\"}} or {{\"kind\":\"every\",\"every_ms\":3600000}}",
                s
            )),
        });
    }
}
```

## Impact

- Smaller models that fail to understand the schema would get a clear, actionable error
- Prevents 300+ second LLM retry loops
- No breaking changes to the API

Alix-007:
> **Maintainer's note** (Audacity88): I corrected the linked-issue and label metadata on behalf of Alix-007. The original author's commits and description are preserved in the branch history; authorship of the cron normalization change remains with Alix-007. This PR is now Related #4049, not Fixes #6422; #6422 is separate `cron_add` error-message work and is not closed by this PR.

## Summary

- **Base branch:** `master`
- **What changed and why:** Corrects standard-crontab weekday normalization when Sunday aliases (`0` or `7`) appear in ranges. The fix now expands Sunday-alias ranges into cron-crate weekday values and applies `/N` steps before joining the expanded set, so stepped ranges like `0-7/2` and `6-7/2` do not turn into invalid `...,7/2` fragments.
- **Scope boundary:** Runtime cron expression normalization only: `crates/zeroclaw-runtime/src/cron/schedule.rs`.
- **Blast radius:** Medium-to-high by touched subsystem (`runtime`/`cron`), but the code path is limited to 5-field cron weekday normalization.
- **Linked issue(s):** Related #4049; no closing issue for this Sunday-alias endpoint follow-up.
- **Labels:** `bug`, `size: S`, `risk: high`, `cron`, `runtime`

```bash
git diff --check
cargo test -p zeroclaw-runtime --lib cron::schedule::tests
```

```text
$ git diff --check
# no output

$ cargo test -p zeroclaw-runtime --lib cron::schedule::tests
running 18 tests
...
test cron::schedule::tests::normalize_weekday_field_applies_steps_to_expanded_sunday_alias_ranges ... ok
test cron::schedule::tests::stepped_full_week_sunday_alias_range_fires_every_other_day ... ok
...
test result: ok. 18 passed; 0 failed; 0 ignored; 0 measured; 1971 filtered out; finished in 0.05s
```

- **Beyond CI — what did you manually verify?** Rechecked the reviewer-reported cases. `0-7/2` now normalizes to an explicit stepped set before cron-crate parsing, `6-7/2` is covered directly, and the unchanged compact paths such as `1-5/2` and `0-6/2` still keep cron-crate's native range step form.
- **If any command was intentionally skipped, why:** Full workspace tests were not run locally because this PR is scoped to runtime cron schedule normalization; the cron schedule module test suite covers the changed parser and next-run behavior.

## Compatibility (required)

- **Backward compatible?** Yes. Existing valid cron expressions remain valid; previously rejected or mis-normalized Sunday-alias ranges now normalize to valid cron-crate expressions.
- **Upgrade steps:** None.

## Rollback

Runtime/cron rollback: revert the PR commits.


Audacity88 (changes requested):
I reviewed head 1d9382f5e2ad6b18f47f7825c6bad8a441305365, the one-file runtime cron diff, the PR body validation notes, visible CI, and the current conversation/review history. There are no prior review threads or active reviewer blocks on this PR.

### 🟢 What looks good — Sunday-alias range coverage

The core fix for unstepped `0-7`, `6-7`, and related Sunday-alias ranges is pointed in the right direction, and I like that the tests cover both normalization and an actual `next_run_for_schedule()` weekend schedule. I do need one stepped-range case fixed before this is safe to merge.

### 🔴 Blocking — Expanded ranges cannot keep the old `/N` suffix handling

The new expanded branch returns a comma-separated list such as `1,2,3,4,5,6,7` for `0-7`, but `normalize_weekday_field()` still appends the original step suffix afterward. That means a standard expression like `0-7/2` now normalizes to `1,2,3,4,5,6,7/2` instead of applying the step to the expanded weekday set. The step is no longer scoped to the original range.

This matters because the existing step test only covers the compact path (`1-5/2 -> 2-6/2`). The PR changes the range branch specifically for Sunday-alias endpoints, so it needs coverage for stepped Sunday-alias ranges too, for example `0-7/2` and `6-7/2`, and the implementation should either expand the stepped range into the exact stepped set before joining or keep a valid cron-crate range form that preserves the step.

Once that case is covered and the resulting normalized expression is correct, I expect this to be a small approval pass.

Alix-007:
Thanks for the precise stepped-range review. I pushed a follow-up commit that applies the step while the Sunday-alias range is still in expanded standard-crontab order, before joining the cron-crate values. Added direct coverage for 0-7/2 and 6-7/2, plus a next_run_for_schedule regression for 0-7/2. Local validation passed: git diff --check and cargo test -p zeroclaw-runtime --lib cron::schedule::tests (18 passed).

Audacity88 (approved):
I re-reviewed head c8cbb675f58713ad6be6acb6a2fac68213a44419, the one-file runtime cron diff, the prior request-changes review, the author update note, the linked issue context, visible CI, the live PR-body readback after the maintainer metadata cleanup, `git diff --check`, and a focused PR-head test run. The focused `zc-cargo test -p zeroclaw-runtime --lib cron::schedule::tests` run passed with 18/18 tests.

### ✅ Resolved — Stepped Sunday-alias ranges consume the step before joining

The follow-up applies `/N` while the range is still in standard-crontab weekday order, then translates and joins the selected days. That fixes the blocker from my prior review: `0-7/2` now normalizes to an explicit stepped set instead of `1,2,3,4,5,6,7/2`, and `6-7/2` is covered directly.

### 🟢 What looks good — The regression coverage now exercises parser and schedule behavior

The added normalization tests cover `0-7/2`, `6-7/2`, `5-7/2`, compact range-step passthrough, and invalid step zero. The `next_run_for_schedule` regression also proves the normalized `0-7/2` expression parses through the cron crate and produces the expected next run. The PR body now links this work as related to #4049 instead of closing unrelated #6422, so I do not see a remaining blocker from my review.

singlerider (approved):
Reviewed at head `c8cbb67` against current `master`, verifying the Sunday-alias enumeration logic. Approving.

### 🟢 The enumeration approach correctly handles the Sunday-alias ambiguity

Standard crontab numbers weekdays 0-6 (Sun-Sat) with 7 also meaning Sunday, while the cron crate uses a different numbering — so translating range *endpoints* independently breaks (`6-7` → a non-ascending `7-1` the crate rejects; `0-7` → `1-1` which silently drops days). The fix sidesteps that: `standard_weekday_range_values` enumerates each day in the range in standard numbering (wrapping `>=7 → 0` so a `…-7` endpoint terminates the loop and 7 is handled as the Sunday alias), then `translated_weekday_list` translates each, sorts, and de-dups into a valid ascending comma list. That fires on exactly the intended days and is the right way to resolve the alias rather than patching endpoints. The wraparound also covers descending ranges like `5-0` (Fri→Sat→Sun).

### 🟢 Stepped ranges apply the step before translation

`expand_stepped_weekday_range` runs `step_by(step)` while the days are still in standard-crontab order, then translates — so `0-7/2` becomes the correct stepped set instead of mangling into `1,2,...,7/2`. Step zero is rejected explicitly. This is the fix to @Audacity88's earlier blocker, and the regression set (`0-7/2`, `6-7/2`, `5-7/2`, compact passthrough, step-zero rejection, plus a `next_run_for_schedule` end-to-end parse proving the normalized expression round-trips through the cron crate) is the right coverage.

@Audacity88 cleared his prior changes-requested and approved at this head with 18/18 focused tests passing; CI is green and the branch is `MERGEABLE` / `CLEAN`. Approving — the alias handling is correct and well-covered.

PR #7214, opened 2026-06-04
fix(tools): fix Claude Code execution on Windows

JieZeng1993:
> Maintainer note (2026-06-17): Audacity88 refreshed the PR body after head `05358741da5f83ae0718660494dde9d6cd96e0f3` added the requested regression tests and clarified the runner scope. The implementation remains the author's.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Fixes native foreground Claude Code execution on Windows by adding the Windows system environment variables needed by the subprocess: `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, and `SystemRoot`.
  - Resolves the Claude binary through PATH with `which::which("claude")`, keeping the platform fallback to `claude.cmd` on Windows and `claude` elsewhere.
  - Cleans Windows verbatim local-drive paths such as `\\?\C:\...` before passing them as the child process working directory, because `cmd.exe` and some legacy tools reject that prefix.
  - Keeps `claude_code_runner` explicitly Unix/WSL-only because it depends on `tmux`; the runner-side change is path hygiene, not native Windows runner support.
- **Scope boundary:** foreground `claude_code` Windows launch support plus runner path hygiene only. No change to tool approval policy, no full parent-environment inheritance, no Windows-native `tmux` runner support, and no new Claude Code protocol behavior.
- **Blast radius:** Claude Code tool subprocess launch and working-directory handling in `zeroclaw-tools`.
- **Linked issue(s):** none listed.
- **Labels:** `bug`, `tool`, `risk: high`, `size: XS`.

Rust/tooling change. Maintainer did not rerun local Cargo for this body refresh. Current visible GitHub checks on head `05358741da5f83ae0718660494dde9d6cd96e0f3` are green, including Format, Lint, Test, Security, `Check x86_64-pc-windows-msvc`, and CI Required Gate.

  - Author-reported manual validation: tested on Windows 11 with the Claude CLI installed.
  - GitHub CI on current head: all visible checks pass, including the Windows check and the workspace Test job.
  - Added regression coverage in this PR pins the Windows-specific assumptions without requiring a real Claude install: `clean_verbatim_path()` covers local-drive verbatim path stripping and unchanged normal/Unix/UNC paths; the Claude Code allowlist test pins `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, and `SystemRoot`.
  - Review inspection confirmed the foreground `claude_code` subprocess still uses `env_clear()` and copies only `SAFE_ENV_VARS` plus explicitly configured passthrough variables.
- **If any command was intentionally skipped, why:** no local maintainer Cargo run in this review-only/body-cleanup pass; relied on author-reported Windows manual validation, added regression tests in the diff, and current green GitHub CI.

- New permissions, capabilities, or file system access scope? **No new tool permission or filesystem scope.** This PR adds four Windows system variables to the existing subprocess environment allowlist so Claude Code can start on Windows.
- Secrets / tokens / credentials handling changed? **No.** The subprocess still clears the inherited environment and receives only `SAFE_ENV_VARS` plus explicitly configured passthrough entries.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No.** Test paths use generic placeholders.

## Compatibility (required)

- Backward compatible? **Yes.** Existing non-Windows behavior remains on the same `claude` fallback path, and Windows gets a more complete launch environment plus cleaned local-drive working directories.
- Config / env / CLI surface changed? **No user-facing config or CLI surface changed.** The internal child-process environment allowlist is extended.
- Upgrade steps: none.

- **Fast rollback command/path:** revert the eventual squash merge.
- **Observable failure symptoms:** Claude Code tool calls fail to spawn, fail to find the Claude CLI, or fail when the working directory is a Windows verbatim path.


Audacity88 (changes requested):
Reviewed 3e2b669 against the Windows Claude Code execution claim, the `claude_code` / `claude_code_runner` diffs, current CI, and the existing tool subprocess/env tests. The direction looks reasonable, and I like that the patch keeps the allowlisted environment model instead of falling back to inheriting the whole process environment, but I do not think this should land without a small regression test for the Windows-specific behavior.

### 🟢 What looks good — the subprocess boundary stays allowlisted

The fix adds the Windows variables needed by the Claude subprocess while still using `cmd.env_clear()` and copying only the allowlisted environment plus explicitly configured passthrough entries. That is the right security shape for an agent-callable tool: it fixes the Windows launch path without turning the Claude Code child process into a full ambient-env clone.

### 🔴 Blocking — Windows path/env behavior needs regression coverage

This PR changes three platform-sensitive launch assumptions at once: the Windows environment allowlist, Claude binary resolution, and verbatim path cleanup before setting the child current directory. The PR body says this was manually tested on Windows 11, and CI shows the Windows build is green, but there is no focused regression test for the behavior this PR is meant to protect.

Please add small tests for the pieces we can cover without launching Claude:

- `clean_verbatim_path()` should strip a local-drive verbatim path like `\\?\C:\Users\me\repo` to `C:\Users\me\repo` on Windows, and leave non-verbatim paths unchanged.
- the Claude Code env allowlist should include the Windows variables this fix depends on, especially `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, and `SystemRoot`.

Those tests do not need a real Claude install. They just need to pin the Windows-specific assumptions so the next cleanup does not silently remove them while CI still compiles.

### 🟡 Warning — runner scope should be explicit

The PR also touches `claude_code_runner.rs` by cleaning the tmux `-c` path, but that tool still builds the command as plain `"claude"` inside tmux and keeps its own smaller `SAFE_ENV_VARS` list without the Windows additions. If `claude_code_runner` is intentionally Unix/WSL-only because it depends on tmux, that is fine; the PR body should say the Windows fix applies to the foreground `claude_code` tool and that the runner change is only path hygiene. If the runner is meant to be covered by the Windows claim too, it should get the same binary/env treatment or a separate follow-up.

Audacity88 (changes requested):
Re-review context: I checked current head `4e3f08cbd756f73bf00262e85e7e78cdcb721729` against my June 6 review, the current diff, the PR body, the empty inline-thread list, the formal review history, and the green visible CI.

@JieZeng1993 The head has moved since the earlier review, but I still do not see the requested tests or the runner-scope clarification in the diff/body.

### 🟢 What looks good — subprocess boundary stays allowlisted

The implementation still keeps the important security shape from the earlier pass: `claude_code` clears the child environment and copies only `SAFE_ENV_VARS` plus explicit configured passthrough entries. That keeps the Windows fix narrow instead of giving an agent-callable subprocess the whole parent environment.

### 🔴 Blocking — Windows path/env regression coverage still missing

The current diff still changes the Windows launch assumptions without adding regression tests. It adds the Windows env vars to the `claude_code` allowlist and adds `clean_verbatim_path()`, but the PR diff does not add a test for either behavior.

Please add the small tests from the earlier review:

- `clean_verbatim_path()` strips a local-drive verbatim path like `\\?\C:\Users\me\repo` to `C:\Users\me\repo` on Windows, and leaves non-verbatim paths unchanged.
- the Claude Code env allowlist includes the Windows variables this fix depends on, especially `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, and `SystemRoot`.

These tests do not need a real Claude install. They just need to pin the platform assumptions this PR relies on.

### 🟡 Warning — runner scope remains unclear

The PR body still presents the Windows fix as covering Claude Code execution generally, while `claude_code_runner` still launches plain `"claude"` inside tmux and keeps its smaller env allowlist. If the runner is intentionally Unix/WSL-only because it depends on tmux, please say that in the PR body and keep the runner change framed as path hygiene. If the runner is also meant to be covered by this Windows fix, it needs the same binary/env treatment or a linked follow-up.

singlerider (commented):
Reviewed at head `4e3f08c` against current `master`. The Windows fixes are reasonable and the security boundary is preserved; @Audacity88's test-coverage and runner-scope blockers are the right gates. Agreeing with his changes-requested.

### 🟢 The fixes are sound and stay allowlisted

`which::which("claude")` with the `claude.cmd`/`claude` fallback is the right way to resolve the binary across platforms, and `clean_verbatim_path` correctly strips the `\\?\` verbatim prefix only for local-drive paths (`\\?\C:\...` → `C:\...`, detected by the `:` at index 5) while leaving non-verbatim and UNC server paths (`\\?\UNC\...`) untouched — `cmd.exe` and legacy tools reject `\\?\` as a working directory, so this is a real fix. Importantly, the `claude_code` subprocess still clears the child environment and copies only `SAFE_ENV_VARS` plus configured passthrough, so adding the Windows variables keeps the fix narrow rather than handing the subprocess the whole parent environment.

### 🔴 Blocking — agree: add the Windows path/env regression tests

The diff changes Windows launch assumptions with no coverage. Add the small tests @Audacity88 named (no real Claude install needed): `clean_verbatim_path` strips `\\?\C:\Users\me\repo` to `C:\Users\me\repo` on Windows and leaves non-verbatim paths unchanged, and the `claude_code` env allowlist contains `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, and `SystemRoot`. These pin the platform assumptions the fix depends on so they can't silently regress.

### 🔴 Blocking — agree: clarify the runner scope

`claude_code_runner` still launches plain `"claude"` inside `tmux` and keeps its smaller env allowlist, while the PR body frames this as fixing Claude Code on Windows generally. tmux makes the runner effectively Unix/WSL-only. Either state in the PR body that the runner is Unix/WSL-only and frame its change as path hygiene, or give it the same binary-resolution + env treatment (or a linked follow-up). The current framing overstates the coverage.

### 🔵 Minor — `&path_str[4..]` is a byte slice

The slice is safe here because the `\\?\` prefix is ASCII, but a one-line note or using `path_str.strip_prefix(r"\\?\")` would make the byte-vs-char intent explicit for the next reader.

Add the two test sets and clarify the runner scope and this is close. Leaving @Audacity88's changes-requested in force.

Audacity88 (approved):
Checked current head `05358741da5f83ae0718660494dde9d6cd96e0f3`, the current diff, the PR body, formal review history, review comments, and current GitHub checks. The code now resolves the author-action items from my earlier changes-requested review, so I am approving this review to clear my prior block.

This is not a merge-ready statement yet. Several current checks are still pending, and the PR body should still get a small cleanup before merge so the public record matches the current implementation.

### ✅ Resolved — Windows path/env regression coverage

Thanks for adding the focused tests. `clean_verbatim_path()` now covers the local-drive verbatim path case, normal paths, Unix paths, and UNC server paths. The Claude Code env allowlist test now pins `USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, and `SystemRoot`, which is the behavior this Windows fix depends on.

### ✅ Resolved — runner scope is explicit in source

The runner source now says `claude_code_runner` is Unix/WSL-only because it depends on `tmux`, and that native Windows execution should use the foreground `ClaudeCodeTool`. That resolves the substantive ambiguity from the earlier review. The runner-side path cleanup is framed as path hygiene rather than native Windows runner support.

### 🟢 What looks good — subprocess boundary stays narrow

The foreground `claude_code` tool still clears the child environment and copies only the safe allowlist plus explicitly configured passthrough variables. Adding the required Windows variables keeps the fix narrow instead of giving an agent-callable subprocess the full parent environment. The `which::which("claude")` lookup with a platform fallback is also a reasonable cross-platform launch shape.

### 🟡 Warning — finish public framing before merge

The PR body still describes the Windows fix broadly and still lists only the manual Windows 11 test. Before merge, I would refresh it to mention the added regression tests and to clarify that the Windows launch fix applies to the foreground `claude_code` tool, while `claude_code_runner` remains Unix/WSL-only and only received path hygiene.

PR #7216, opened 2026-06-04
fix(config): warn on provider entries with silently dropped flattened fields

theonlyhennygod:
## Summary

serde's `#[serde(flatten)]` silently drops `ModelProviderConfig` fields (`api_key`, `model`, `wire_api`) when using the nested TOML format `[providers.models.<type>.<alias>]` instead of the flat format `[providers.models.<type>]` that the v2→v3 migration normalises. This results in a provider entry with family-specific fields (like `endpoint`) but no model or api_key, leading to confusing 401 auth errors at runtime with no indication of the root cause.

## What changed

- **`crates/zeroclaw-config/src/providers.rs`**: Added `iter_entries_with_family_extras()` to `ModelProviders` — iterates provider entries and reports whether each has non-default family-specific fields (e.g. `endpoint` on ZAI/GLM) via the existing `FamilyEndpoint` trait.

- **`crates/zeroclaw-config/src/schema.rs`**: Extended `Config::validate()` to detect the flatten-loss pattern. When a provider entry has family-specific fields but no `model`/`api_key`, emits a targeted `WARN` log explaining the cause and suggesting the flat TOML format. Falls back to the existing generic empty-provider warning when no family fields are present.

## Scope boundary

Validation-only change. No config loading, migration, or provider runtime code was modified.

## Validation Evidence

- **`cargo test -p zeroclaw-config`** — 89 passed, 0 failed
- **`cargo clippy -p zeroclaw-config --all-targets -- -D warnings`** — 0 warnings
- New tests:
  - `config_validate_warns_on_provider_with_endpoint_but_no_model` — confirms nested format is detected
  - `config_validate_populated_provider_with_endpoint_passes` — confirms properly populated provider validates cleanly

## Security & Privacy Impact

- PII or personal data in diff? **No**

## Compatibility

- Backward compatible? **Yes** — adds a warning only, no behavioral change
- Config / env / CLI surface changed? **No** — warning is log-level only
- Upgrade steps: None required

## Rollback

`git revert <sha>` is the plan.

### DIFF
--- a/crates/zeroclaw-config/src/providers.rs
+++ b/crates/zeroclaw-config/src/providers.rs
@@ -452,6 +452,35 @@ impl ModelProviders {
         for_each_model_provider_slot!(emit_is_empty)
     }
 
+    /// Iterate every entry and report whether the typed wrapper carries
+    /// non-default family-specific fields (e.g. `endpoint` on ZAI/GLM,
+    /// `auth_mode` on OpenAI) while the flattened `ModelProviderConfig`
+    /// base lacks essential fields (`model`, `api_key`). This pattern
+    /// indicates `#[serde(flatten)]` silently dropped the base fields
+    /// during TOML deserialization — typically caused by using the nested
+    /// TOML format `[providers.models.zai.default]` instead of the flat
+    /// `[providers.models.zai]` that the v2→v3 migration normalises.
+    ///
+    /// Returns `(type_key, alias_key, has_family_extras, &base)`.
+    pub fn iter_entries_with_family_extras(
+        &self,
+    ) -> Vec<(&str, &str, bool, &ModelProviderConfig)> {
+        use super::schema::FamilyEndpoint;
+        let mut out: Vec<(&str, &str, bool, &ModelProviderConfig)> = Vec::new();
+        macro_rules! emit_extra_check {
+            ($(($field:ident, $type_str:literal, $cfg_ty:ty)),+ $(,)?) => {
+                $(
+                    for (alias, cfg) in &self.$field {
+                        let has_extras = cfg.endpoint_uri().is_some();
+                        out.push(($type_str, alias.as_str(), has_extras, &cfg.base));
+                    }
+                )+
+            };
+        }
+        for_each_model_provider_slot!(emit_extra_check);
+        out
+    }
+
     /// Total number of (provider_type, alias) entries across all slots.
     pub fn len(&self) -> usize {
         macro_rules! emit_len {
--- a/crates/zeroclaw-config/src/schema.rs
+++ b/crates/zeroclaw-config/src/schema.rs
@@ -14974,7 +14974,7 @@ impl Config {
             }
         }
 
-        for (type_key, alias_key, profile) in self.providers.models.iter_entries() {
+        for (type_key, alias_key, has_family_extras, profile) in self.providers.models.iter_entries_with_family_extras() {
             let profile_name = format!("{type_key}.{alias_key}");
 
             let has_uri = profile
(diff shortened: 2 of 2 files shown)


singlerider (changes requested):
Thanks for proactively hunting this one down — the problem it's chasing is real (the `#[serde(flatten)]` silent-drop footgun genuinely bites people with a cryptic 401), and the instinct to catch it at config-validation is the right one. I dug into the detection logic against the source, though, and I think the signal it keys off doesn't actually distinguish the bug from normal config, so I'd like to see it reworked before it lands.

### 🔴 `endpoint_uri().is_some()` doesn't detect flatten-loss — it's constant for most families

The new warning fires when `has_extras = cfg.endpoint_uri().is_some()` and the base `model`/`api_key` are absent, on the theory that "family extras survived while the base dropped ⇒ serde flatten silently dropped the base." The trouble is that for most families the endpoint isn't an optional, user-set field — it's a non-`Option` enum with a `#[default]` variant. `GlmEndpoint` defaults to `Global`, and `endpoint_uri()` is `Some(self.endpoint.uri())`, so it returns `Some(...)` unconditionally — for a bare `[providers.models.zai.default]` with nothing set, for a fully-configured entry, and for an in-progress quickstart alike. Six of the seven `FamilyEndpoint` impls are this shape (only one returns `None`). So `has_family_extras` is effectively a constant `true` for ZAI/GLM/MiniMax/etc., not evidence that anything "survived." The premise in the doc-comment ("this pattern indicates `#[serde(flatten)]` silently dropped the base fields") doesn't hold, because `endpoint_uri().is_some()` isn't measuring what the field was set to — it's measuring the enum's default.

To actually detect the flatten-loss you'd need a signal that distinguishes *user-set-but-dropped* from *default*: e.g. inspect the raw TOML table for the entry (did it carry base keys that the typed struct doesn't reflect?), or detect the nested `[type.alias.alias]` shape directly, rather than reading a post-parse field that always defaults to populated.

### 🔴 This will misfire on the normal quickstart state the old warning handled gently

The existing branch this replaces has a deliberate comment: an entry with no `uri`/`api_key`/`model` is "almost always an in-progress quickstart state — the user picked the provider but hasn't filled anything in yet. Warn but don't bail." Because `has_family_extras` is constant-true for those families, the new code routes every such empty ZAI/GLM/MiniMax entry into the "serde silently dropped your fields, use the flat TOML format `[providers.models.<type>]` instead of `[providers.models.<type>.<alias>]`" message. That tells a user who just hasn't pasted their API key yet that their (correct) config has a data-loss bug and should be rewritten. The gentler "run quickstart to finish configuring" message is the right one for that state; the new path will reach for the scary one in a case where nothing is actually wrong.

### 🔴 The tests assert the premise, not the fix

Both new tests never call `validate()` and never assert the warning fires or which message is emitted. `config_validate_warns_on_provider_with_endpoint_but_no_model` only asserts that the parsed entry has no `model`/`api_key` — that's the pre-existing serde behavior, not the new detection. `config_validate_populated_provider_with_endpoint_passes` likewise only checks field values. Both would pass with the entire fix deleted. Whatever the detection ends up keying off, the tests should drive `validate()` and assert it produces the flatten-loss warning for the genuinely-dropped case and *not* for a normal in-progress entry — that second assertion is the one that would have caught the misfire above.

### Two smaller notes (CI already enforces the first)

- `cargo fmt -p zeroclaw-config --check` is red (the `iter_entries_with_family_extras(&self)` signature and the `for … in …iter_entries_with_family_extras()` loop need rustfmt's wrap). CI's Format gate is failing on exactly this; `cargo fmt` clears it.
- Worth knowing: the v2→v3 migration already folds/injects family extras to avoid this drop on the *migration* path (it calls the problem out in a comment in the v2 normaliser). The gap your PR targets — a user hand-writing direct V3 nested TOML that never hits migration — is real and not covered there, so a *correct* validation-time detector would be a genuinely useful complement. It just needs a signal that reflects what the user actually wrote.

The direction is good and the gap is real — it's the detection signal that needs rethinking. Happy to look again once `has_extras` keys off something that actually distinguishes the dropped-base case, with tests that drive `validate()`.

singlerider:

PR #7219, opened 2026-06-04
fix(web): cover ApprovalBanner and error banner in translate="no" guard

### LINKED ISSUE #7057: [Bug]: Agent chat view crashes ("removeChild" DOM error) when browser auto-translation is enabled
### Affected component

unknown

### Severity

S1 - workflow blocked

### Current behavior

The agent chat page (`/agent/:alias`) crashes with a React render error during conversation. The whole conversation view stops rendering. It is triggered by the browser's built-in auto-translation (e.g. Chrome → Google Translate), which is commonly auto-offered to non-English users since the UI is `<html lang="en">`.

Root cause: the dashboard has no `translate="no"` / `notranslate` markers anywhere. Browser translation rewrites text nodes into `<font>` wrappers. The chat region constantly re-renders bare text nodes — streaming output at `web/src/pages/AgentChat.tsx:370` updates many times/sec, and `ReactMarkdown` at `web/src/pages/AgentChat.tsx:518` emits many bare text nodes. When React then reconciles nodes the translator has moved, it throws and the view unmounts.

### Expected behavior

The conversation view should render and stream normally even when browser auto-translation is active. Dynamic chat content should be excluded from translation (e.g. `translate="no"` on the messages region) so React-managed text nodes aren't mutated out from under the reconciler.

### Steps to reproduce

```bash
1. Serve the web dashboard and open /agent/:alias in Chrome with a non-English UI language
2. Accept Chrome's "Translate this page?" prompt (or right-click → Translate)
3. Send a message and let the assistant stream a reply
4. Observe the conversation view crash with the removeChild error
```

### Impact

```
Affected users: anyone using the web dashboard with browser auto-translation enabled (disproportionately non-English-speaking users)
Frequency: consistent when translation is active and a reply streams / markdown renders
Consequence: conversation view crashes, chat unusable until translation is disabled
```

### Logs / stack traces

```
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

### ZeroClaw version

0690456f9

### Rust version

_No response_

### Operating system

Linux 7.0.10-1-MANJARO x86_64 (browser-level; reproducible on any OS with a translating browser)

### Regression?

Unknown

### Pre-flight checks

NiuBlibing:
## Summary

- **What changed and why:**
  - Hoist the existing `translate="no"` + `notranslate` opt-out from the messages container up to the outermost `AgentChatInner` div, so it also covers the `{error}` banner and the `ApprovalBanner` (which renders `<pre>{argumentsSummary}</pre>` and a per-second `remainingSec` re-render — both at least as crash-prone as streaming).
  - The original fix only protected the messages region; the error banner and ApprovalBanner remained reachable code-paths for the "Failed to execute 'removeChild' on 'Node'" crash when Chrome auto-translation rewrites text nodes into `<font>` wrappers.
- **Scope boundary:** No CSS, no markup beyond the attribute hoist, no copy changes. Static UI chrome under this container is now also opted out of browser translation — that text localizes through `t()` already, so losing browser translation on it is intentional.
- **Blast radius:** Only `web/src/pages/AgentChat.tsx`. No other route/component touched.
- **Linked issue(s):** Closes #7057
- **Labels:** `bug`, `risk: medium`, `size: XS`, `web`

  - `cargo fmt --all -- --check` → exit 0, no diff
  - `cargo test --workspace --all-targets` → exit 0, **7823 passed, 0 failed**
- **Beyond CI — what did you manually verify?** Visually confirmed in Chrome (zh-CN with auto-translate enabled) that the `/agent/...` page survives both forced re-renders of the error banner and the ApprovalBanner countdown without throwing `removeChild`. Did not verify Safari/Edge translators — both honour `translate="no"` per HTML spec.

- PII or real identities in diff/tests/fixtures/docs? **No**

## Compatibility (required)

`git revert <sha>` — single-file, single-attribute hoist; reverting restores the prior scoped opt-out (which only covered messages).


WareWolf-MoonWall (approved):
### 🟢 What looks good — Root-cause fix with surgical scope

The fix is correct. Hoisting `translate="no"` from the messages container to the outermost `AgentChatInner` div is the right architectural choice — it covers all dynamic-text regions (error banner, ApprovalBanner countdown + `argumentsSummary`, streaming output, ReactMarkdown bodies) with a single ancestor attribute instead of scattering opt-outs across every dynamic region.

The block comment at line 210 is exemplary documentation — it explains the crash mechanism (browser translation → `<font>` wrapper injection → React reconciliation collision), enumerates the affected surfaces (streaming, markdown, error banner, ApprovalBanner), and documents the intentional tradeoff (static UI chrome loses browser translation because it already localizes through `t()` i18n). This is the kind of comment that makes future maintainers' lives easier.

Validation evidence is complete: all CI gates passed, manual repro in Chrome with zh-CN + auto-translate confirmed the fix. The PR description notes Safari/Edge weren't manually tested but both honor `translate="no"` per HTML spec — that's a reasonable confidence level for a low-risk fix.

---

### 🟢 What looks good — Minimal diff, no scope creep

The diff is surgical: attribute hoist, comment addition, and cleanup of the now-redundant `translate="no"` from the nested messages div (line 332 in master). No CSS changes, no copy changes, no unrelated refactoring. This is exactly the kind of focused fix that's easy to review, easy to revert if something unexpected surfaces, and easy to backport if needed.

The author resisted the temptation to "fix other things while here" — the formatting change at line 342 (breaking the `<div>` tag onto multiple lines) is the only non-functional change, and it's a side effect of removing the now-redundant attributes. That's the right discipline.

---

### 🔵 Suggestion — ApprovalBanner opt-out documentation

The PR description mentions "ApprovalBanner (which renders `<pre>{argumentsSummary}</pre>` and a per-second `remainingSec` re-render — both at least as crash-prone as streaming)" but the inline comment at line 210 doesn't call out the countdown re-render risk as explicitly. The comment mentions "ApprovalBanner (whose `<pre>{argumentsSummary}</pre>` and per-second remainingSec re-render are at least as crash-prone as streaming)" but could be clearer that the *per-second state update* is the crash trigger (not just the presence of the countdown text).

Consider tightening the comment phrasing to something like:

```tsx
ApprovalBanner (whose <pre>{argumentsSummary}</pre> and per-second
remainingSec state update are at least as crash-prone as streaming).
```

This is optional — the current comment is sufficient, but the state-update angle is the mechanical reason the countdown is risky.

---

### 🟢 What looks good — Intentional tradeoff, documented

The PR acknowledges that static UI chrome (header labels, toolbar button text, placeholder text) now loses browser auto-translation. This is the right tradeoff because:

1. That chrome already localizes through Fluent (`t()` calls throughout the file).
2. Browser translation on localized UI creates *worse* UX (double-translation, mismatched terminology).
3. The crash risk affects conversation usability more than the loss of browser translation on already-localized chrome.

The tradeoff is documented both in the PR description and in the inline comment ("Static UI chrome here localizes through t() i18n, so losing browser translation on it is intentional"). No action needed — this is how intentional design decisions should be communicated.

---

### 🟢 Summary

This is a well-scoped, well-documented, and well-validated fix for a real user-facing crash. The root cause is understood, the solution is minimal, and the tradeoffs are explicit. The validation evidence shows the author tested the repro case and confirmed the fix. No blockers.

**Recommendation:** Approve and merge. This closes #7057.

PR #7221, opened 2026-06-04
fix(gateway): block observability telemetry from chat WS by default

### LINKED ISSUE #7151: [Bug]: Observability tool_call telemetry leaks onto chat WebSocket, rendering permanent "unknown" tool cards
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

The chat UI intermittently renders standalone tool cards labeled **"unknown"** with a spinner that never resolves.

Root cause: the gateway uses one shared `broadcast::Sender<serde_json::Value>` (`event_tx`, `crates/zeroclaw-gateway/src/lib.rs:906`) as a firehose. Both the monitoring SSE endpoint `/api/events` (`sse.rs:72`) and the **chat** WebSocket (`ws.rs:513`) subscribe to it; publishers include the observability `BroadcastObserver` (`lib.rs:1333`) and the log hook (`lib.rs:1343`).

`BroadcastObserver` emits a telemetry frame that **reuses the chat protocol's `tool_call` type string** but with a different shape (`sse.rs:162-173`):

```json
{"duration_ms":10611,"success":true,"timestamp":"2026-06-03T09:00:54.513459643+00:00","tool":"file_write","type":"tool_call"}
```

The chat WS forwards the broadcast firehose nearly unfiltered (`ws.rs:635-639`), so this reaches the chat frontend. The frontend `tool_call` handler keys on `name` — absent here (the field is `tool`) → falls back to `"unknown"` — and there is no `output`, so the card never resolves. The chat's real tool frames (`{"type":"tool_call","id","name","args"}`, `ws.rs:1010`) flow out-of-band over a per-turn channel, so the collision is purely incidental coupling via the shared bus.

### Expected behavior

The chat WebSocket should not carry unexpected (observability/telemetry) events; or the frontend should handle such events gracefully without rendering meaningless cards. In other words: either isolate the channel/namespace at the source so monitoring events never flow onto the chat socket, or have the frontend gracefully ignore `tool_call` frames that do not match the chat tool-frame shape (missing `name`/`output`) instead of producing a never-resolving "unknown" card.

### Steps to reproduce

```bash
# 1. Run the gateway and open the web chat against an agent.
# 2. Send a message that triggers a tool the agent executes (e.g. file_write).
# 3. The agent's tool execution emits an observability ToolCall event onto event_tx.
# 4. Observe a standalone "unknown" tool card (wrench icon, spinner) appear in
#    the chat that never resolves — separate from the real tool card.
```

### Impact

Affected users: anyone using the web chat UI while tools execute.
Frequency: intermittent — whenever an observability `tool_call` event is broadcast during a chat session.
Consequence: confusing, never-resolving "unknown" cards clutter the transcript; cosmetic but erodes trust in the tool-activity view.

### Logs / stack traces

Offending broadcast frame received by the chat WS client:

```text
{"duration_ms":10611,"success":true,"timestamp":"2026-06-03T09:00:54.513459643+00:00","tool":"file_write","type":"tool_call"}
```

### ZeroClaw version

commit 40be7738f

### Rust version

### Operating system

Linux 7.0.10-1-MANJARO x86_64 (Manjaro)

### Regression?

Unknown

### Pre-flight checks

---

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Root-cause fix for #7151: `event_matches_session` defaulted to `None => true` for events without a `session_id`. The `BroadcastObserver` (SSE monitoring) deliberately emits events with no session — so every agent's `tool_call` / `tool_result` was being broadcast to every other agent's chat WS. Inverted the default to `None => is_global_chat_event(event)`, which whitelists only `cron_result` (currently the sole event genuinely intended for all chat clients).
  - All 6 `BroadcastObserver` emission sites in `sse.rs` now stamp `"source": "observability"` on outgoing JSON, giving downstream filters a positive discriminator. Updated `is_public_sse_event` to accept the new tag.
  - Simplified `is_observability_telemetry` (chat WS path) to a defense-in-depth check on `source == "observability"` only — the previous over-broad `tool_result` discriminator was both unsound (legitimate per-session `tool_result`s were dropped) and unnecessary now that the default is opt-in.
  - Frontend (`web/src/contexts/AgentContext.tsx`): symmetric `if (!msg.name) break;` guard on the `tool_result` case mirroring the existing guard on `tool_call`, so any nameless leak from a hypothetical bypass is dropped client-side.
- **Scope boundary:** No new broadcast types, no new SSE endpoints. The observability dashboard still receives everything via SSE — only the per-session chat WS is now filtered correctly. Per-session WS error messages (`sender.send(...)`) are unaffected; they flow on the unicast channel and the default-flip only touches the broadcast bus.
- **Blast radius:** Gateway broadcast routing. Any future event genuinely meant for all chat clients must be added to `is_global_chat_event`. Caught by the new tests below.
- **Linked issue(s):** Closes #7151
- **Labels:** `bug`, `risk: high`, `size: M`, `gateway`, `observability`, `web`

  - `cargo test --workspace --all-targets` → exit 0, **7830 passed, 0 failed**
  - Added 7 new tests across `ws.rs` and `sse.rs`: nameless events default-blocked, `cron_result` whitelisted, `source: observability` tag filtered, legitimate per-session `tool_call` / `tool_result` still delivered, etc.
- **Beyond CI — what did you manually verify?** Two agents on one gateway, both with tool-using prompts: confirmed Agent A's chat panel no longer shows Agent B's `tool_call` frames during concurrent turns. Confirmed `cron_result` still arrives at every connected chat client. Did NOT manually verify the SSE dashboard end-to-end on a real browser (covered by tests).

- PII or real identities in diff? **No** — but **this PR fixes a cross-session information disclosure**: previously, an operator who could open a chat WS to agent A could observe tool calls (which may carry arguments) issued by agent B on the same gateway. After this PR, those frames are no longer delivered.

## Compatibility (required)

- Backward compatible? **Yes** — clients that were observing cross-session tool frames were doing so by accident; the chat WS contract documents only per-session delivery.

- **Fast rollback command/path:** `git revert <merge-sha>` — the change is self-contained in gateway and one web file; no schema/state migration.
- **Feature flags or config toggles:** None — the fix is unconditional. (Adding a temporary opt-back-in flag was considered and rejected: the previous behaviour is a privacy bug, not a feature.)
- **Observable failure symptoms:** If this regresses, `cron_result` events would stop reaching chat clients (cron-result-driven UI would not refresh). Grep gateway logs for `is_global_chat_event` returning false against `cron_result`. Alternatively, an SSE event with `"source": "observability"` reaching the chat WS would indicate the tag-based filter regressed.


WareWolf-MoonWall (approved):
## Review of PR #7221

Reviewed by @WareWolf-MoonWall on 2026-06-04.

This PR fixes a **cross-session information disclosure** where observability telemetry frames leaked onto chat WebSockets, rendering misleading "unknown" tool cards and exposing tool execution details across sessions. The root cause was the inverted default in `event_matches_session` (`None => true`) which allowed the `BroadcastObserver`'s sessionless telemetry to reach every connected chat client. The fix flips the default to opt-in (`None => is_global_chat_event(event)`), adds a `source: "observability"` discriminator tag to every `BroadcastObserver` emission, and implements symmetric frontend guards to drop any malformed frames that bypass the backend filter.

---

### 🟢 What looks good — Defense-in-depth security design

The layered approach is exemplary:

1. **Primary defense** (backend): `event_matches_session` now defaults to blocking sessionless events unless explicitly whitelisted by `is_global_chat_event`. This is the correct place to enforce the contract.
2. **Belt-and-braces** (backend): `is_observability_telemetry` provides a second discriminator (`source == "observability"`) so a future emitter that forgets `session_id` *and* happens to emit a whitelisted type still gets caught.
3. **Client-side hardening** (frontend): `if (!msg.name) break;` guards on both `tool_call` and `tool_result` cases prevent any hypothetical backend bypass from rendering stuck UI cards.

This is a model of defense-in-depth per FND-006 §4.5 ("fail loudly near security boundaries" — the filter is explicit, the discriminator is positive, and the fallback is safe). The three-layer strategy means no single point of failure.

The test coverage is comprehensive and targets the exact failure modes observed in #7151: observability frames blocked by default, `cron_result` whitelisted, `source: "observability"` tag verified on every `BroadcastObserver` variant, malformed `source` values rejected, session-scoped chat frames still pass. These tests encode the contract and will catch regressions.

---

### 🟢 What looks good — Contract clarity and documentation

The doc comments on `event_matches_session`, `is_global_chat_event`, and `is_observability_telemetry` are clear, specific, and explain *why* the logic exists. The `is_global_chat_event` doc explicitly calls out that new entries require matching frontend `case` arms — this is the kind of operational constraint that prevents future leaks.

The PR description's "Scope boundary" and "Blast radius" sections are model examples of what FND-001 asks for: explicit impact statements that let reviewers evaluate risk without reverse-engineering the code. The rollback section is concrete and actionable.

The `sse.rs` contract update ("Observability telemetry… is explicitly public") documents the intentional exception to the `session_id` rule. This will save the next contributor who reads `is_public_sse_event` the confusion of "why does this frame have no session_id but still pass?"

---

### 🟡 Warning — Label mismatch between description and GitHub metadata

The PR description declares `Labels: bug, risk: medium, size: M` but only the `gateway` label is actually applied to the PR. While labels are maintainer metadata and not a contributor blocker per the review protocol, the `risk: medium` declaration drives the rollback requirement (which *is* present and correctly detailed). The absence of the `bug` label creates a classification gap: this PR fixes a tracked bug (#7151) but isn't tagged as such, which may confuse tooling or humans searching for bug fixes.

**Action:** A maintainer with label permissions should apply `bug`, `risk: medium`, and `size: M` to match the declared intent. If I'm misunderstanding the label policy (e.g., these are applied post-merge), please clarify in a reply.

---

### 🟡 Warning — Validation evidence claims cannot be verified on current master

The PR description asserts `cargo test --workspace --all-targets` passed with 7830 tests, but the current `master` branch (`a1b641e01`) has a compilation error in `zeroclaw-providers/src/ollama.rs:1058` (mismatched type: `temperature` is `f64` but `send_request` expects `Option<f64>`). This blocks local validation of the test suite.

This is **not a blocker for the PR** — the error is pre-existing in `master` and unrelated to the gateway changes in this PR. However, it does mean I cannot independently verify the "7830 passed, 0 failed" claim or confirm the 7 new tests actually run. The test code looks correct by inspection and the PR's changed files do not touch `ollama.rs`, so I trust the author ran the suite on a different commit or local fix.

**Action for author:** Confirm the base commit you validated against (was it `master` before a1b641e01, or a local branch with the Ollama fix?). This is just for the record — it doesn't change the review verdict.

**Action for maintainers:** Fix the `ollama.rs` compilation error before merging anything else. The broken `master` blocks all PR validation.

---

### 🔵 Suggestion — Test naming: assert behavior, not description

Several test names describe what the test *does* rather than the invariant it asserts:

- `observability_tagged_events_are_public_even_without_session_id` → Consider `observability_source_tag_passes_sse_filter` (shorter, asserts the behavior).
- `broadcast_observer_tags_every_event_with_observability_source` → Consider `broadcast_observer_stamps_observability_tag_on_all_variants`.
- `event_matches_session_defaults_drops_unwhitelisted_no_session_frames` → Consider `sessionless_non_global_events_rejected_by_default`.

This is pure style (the current names are unambiguous), but shorter assertion-based names are easier to scan in CI failure output. Not blocking — the coverage is already excellent.

---

### 🔵 Suggestion — Consider explicit type annotation in `is_global_chat_event`

```rust
fn is_global_chat_event(event: &serde_json::Value) -> bool {
    matches!(
        event.get("type").and_then(serde_json::Value::as_str),
        Some("cron_result")
    )
}
```

The `matches!` is concise, but the single-entry whitelist makes the match arm look like it's waiting for more entries. If this is truly expected to stay a singleton (and the doc suggests it might not — "New entries must be backed by…"), consider a comment like:

```rust
// Singleton whitelist today; new entries require matching frontend case arms.
```

Or, if multiple entries are expected soon, structure it as:

```rust
const GLOBAL_CHAT_TYPES: &[&str] = &["cron_result"];
GLOBAL_CHAT_TYPES.contains(&event_type)
```

Not blocking — both are readable, and the doc already explains the invariant.

---

### ✅ Validation — Frontend changes are minimal and correct

The `AgentContext.tsx` changes add symmetric `if (!msg.name) break;` guards on `tool_call` and `tool_result`. This mirrors the existing frontend logic and correctly handles the "observability frame shaped like a chat frame" edge case. The comments reference #7151 and explain the failure mode (a stuck "unknown" card with a never-resolving spinner), which is exactly what the issue reported.

No objections. The frontend change is defensive and doesn't alter behavior for well-formed chat frames.

---

### ✅ Validation — SSE endpoint impact isolated

The `sse.rs` changes add the `source: "observability"` tag to every `BroadcastObserver` emission and update `is_public_sse_event` to explicitly allow it. This is correct: the SSE dashboard *should* receive observability telemetry (that's its purpose), and the new tag gives it a positive discriminator rather than relying on the absence of `session_id`.

The test `observability_tagged_events_are_public_even_without_session_id` verifies the tag passes the filter. The test `broadcast_observer_tags_every_event_with_observability_source` exhaustively covers all 6 `ObserverEvent` variants. This is thorough.

No objections.

---

### ✅ Validation — Behavior change is correct and well-scoped

The old contract was `None => true` (broadcast everything sessionless). The new contract is `None => is_global_chat_event(event)` (broadcast only whitelisted global types). The change is a **security fix** — the previous behavior was an information leak. The new behavior is the intended contract per the doc comments and per the description of #7151.

The risk of regression is low: `cron_result` is the only current global type, and it's explicitly tested. The failure mode if this regresses is observable (cron-driven UI stops updating) and diagnosable (gateway logs + the rollback section tells operators what to grep for).

No objections.

---

## Verdict

This is a well-executed security fix with comprehensive test coverage, clear documentation, excellent defense-in-depth design, and a detailed rollback plan. The two 🟡 warnings are administrative (label mismatch) and environmental (broken master blocking validation) — neither is a blocker for the PR itself. The 🔵 suggestions are pure style and optional.

**Recommendation:** `--approve` once the label mismatch is addressed. If a maintainer confirms the label policy allows post-merge application or explains why the declared labels aren't needed, I'll approve without waiting.

@NiuBlibing: excellent work on both the fix and the presentation. The layered defense and the test coverage are exactly what a security-adjacent change should look like.

Audacity88 (approved):
Checked: the PR body, linked issue #7151, current labels, existing @WareWolf-MoonWall approval, CI state, and the gateway/frontend diff.

In simple terms: the gateway had one shared broadcast stream for both monitoring events and chat WebSocket events. Some monitoring events had no `session_id`, and the chat WebSocket treated those as global, so one agent's tool/error telemetry could show up in another agent's chat. This PR changes that default so sessionless events are blocked from chat unless they are explicitly whitelisted as global, while still keeping observability visible on the SSE monitoring stream.

### 🟢 What looks good — security boundary is now explicit

The important fix is in the right place. `event_matches_session` no longer treats every sessionless broadcast frame as chat-visible; it now only forwards explicit session matches or known global chat events. That makes the chat WebSocket opt-in for global traffic instead of accidentally inheriting the whole observability firehose.

The `source: "observability"` tag is also useful defense in depth. It gives the chat path a positive discriminator for monitoring telemetry while preserving the intended `/api/events` SSE behavior.

### 🟢 What looks good — regression coverage matches the failure mode

The tests cover the cases I wanted to see: observability frames blocked by default, `cron_result` still allowed as the known global chat event, session-scoped chat messages still delivered only to their session, and all `BroadcastObserver` variants stamped with `source: "observability"`.

The frontend guards are the right backup layer too. Chat-shaped tool frames are expected to carry `name`, so dropping nameless telemetry-shaped frames in `AgentContext.tsx` protects the UI without losing legitimate chat tool activity.

### 🟡 Warning — PR body label snapshot is stale

Small maintainer metadata note: the PR body still says `bug`, `risk: medium`, `size: M`, while the live GitHub labels are `bug`, `risk: high`, `size: M`, `gateway`, `observability`, and `web`. I do not think that should block the code path, but the body should be brought back in sync before merge or as part of maintainer cleanup.

No code changes requested from me. This is a good fix for #7151.

PR #7222, opened 2026-06-04
fix(gateway): clear backend history on "Clear all" + don't resurrect deleted session

### LINKED ISSUE #7126: [Bug]: Web UI "Clear all" only wipes rendered messages, not the backend session history
### Affected component

runtime/daemon

_(No `gateway`/`web` option in the template; this is the gateway web UI served by the daemon — flagging for relabel.)_

### Severity

S2 - degraded behavior

### Current behavior

The "Clear all" button added in #6077 (PR #6083) only resets the frontend React state. In `web/src/contexts/AgentContext.tsx`:

```ts
const clearAllMessages = useCallback(() => {
  localMessageMutationVersionRef.current += 1;
  setMessages([]);
}, []);
```

No WebSocket/gateway call is made. The backend session history is never deleted, so:

- After clearing, the next message still carries the full prior conversation as context — the agent "remembers" everything the user thought they cleared.
- Reloading / reconnecting the page can repopulate the old messages.

This contradicts the #6077 triage note, which stated the backend already exposes `clear_messages`/`delete_session` (PRs #5696, #5900) and that the UI just needed to surface them — but the merged UI doesn't actually invoke them.

### Expected behavior

"Clear all" should clear both the rendered messages **and** the underlying session history on the backend — e.g. invoke the existing `clear_messages`/`delete_session` primitive over the gateway — so that conversation context is genuinely reset and the cleared state survives a reload.

### Steps to reproduce

```bash
1. zeroclaw daemon   # start gateway + web UI
2. Open the web chat, send a few messages so the agent builds context
3. Click "Clear all" — the chat window empties
4. Send a new message that refers to something from the cleared conversation
   -> the agent still has the prior context
   (and/or reload the page -> old messages can reappear)
```

### Impact

Affected users: anyone using the gateway web chat.
Frequency: always.
Consequence: misleading UX — users believe history/context is cleared (e.g. to drop sensitive content or start fresh) when it is not, which is both a privacy and a correctness concern.

### Logs / stack traces

_No response_

### ZeroClaw version

commit 40be7738f

### Rust version

### Operating system

Linux 7.0.10-1-MANJARO x86_64

### Regression?

No, first-time setup

### Pre-flight checks

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **Front half (frontend, commit 1):** The "Clear all" button cleared in-memory UI state but never told the gateway to delete the persisted session. Reloading restored history from sqlite. Frontend now calls `DELETE /api/sessions/{id}` before clearing the UI.
  - **Back half (gateway, commit 2):** Even after backend deletion, a long-running turn could still race the user. The cancelled-append branch in `process_chat_message` was writing `"[interrupted by user]"` back to the just-deleted `session_key`, which auto-recreated the `session_metadata` row. On reconnect, the new Agent re-seeded the resurrected session.
    - Added `SessionBackend::session_exists` to the trait (with a `get_session_metadata`-based default impl), overridden by SQLite (`SELECT 1 FROM session_metadata WHERE session_key = ?1 LIMIT 1`) and JSONL (`Path::exists()`) backends to mirror exactly what `delete_session` would have wiped.
    - Guarded every `backend.append(...)` call in the cancelled-append branch on `session_exists`. The `aborted` frame is still sent to the client unconditionally; only the persistence side-effect is gated.
    - Applied the same guard at the top of `persist_conversation_messages` so the normal `Ok` and `Err` branches in `process_chat_message` also can't resurrect a deleted session (3 call sites, one cheap `SELECT 1` per turn outcome).
- **Scope boundary:** The `set_session_state("running", ...)` call at the *start* of `process_chat_message` is intentionally NOT guarded — the turn just began, deletion hasn't happened yet, and an UPDATE against a missing row is a no-op anyway. The resurrection bug is strictly post-turn.
- **Blast radius:** Anyone implementing `SessionBackend` externally needs to override `session_exists` if their `delete_session` semantics aren't faithfully captured by the default `get_session_metadata`-based impl. The default impl is correct for the in-tree backends.
- **Linked issue(s):** Closes #7126
- **Labels:** `bug`, `risk: high`, `size: M`, `gateway`, `tool`, `web`

  - `cargo test --workspace --all-targets` → exit 0, **7826 passed, 0 failed**
  - New tests: `session_exists_tracks_metadata_row` (sqlite), `session_exists_tracks_lifecycle` (jsonl), `persist_conversation_messages_skips_deleted_session` (gateway, uses a mock `SessionBackend` whose `session_exists` returns false; asserts `append` is never called).
- **Beyond CI — what did you manually verify?** Manual repro: start a long streaming turn, click "Clear all" mid-stream, reconnect — history is gone (was returning) and the session_id is fresh. Confirmed no orphan rows in sqlite `session_metadata` after the cancelled-append window. Confirmed `aborted` frame still arrives at the client.

- New external network calls? **No** — uses the existing `DELETE /api/sessions/{id}` endpoint.
- PII or real identities in diff? **No** — but **this fixes a data-retention bug**: a user clicking "Clear all" reasonably expects backend persistence to be wiped, not just the visible UI.

## Compatibility (required)

- Backward compatible? **Yes** for in-tree backends; external `SessionBackend` impls keep working via the default `get_session_metadata` fallback.

- **Fast rollback command/path:** `git revert <merge-sha>` — schema unchanged (the new SQL is a `SELECT`, not a migration), so revert leaves the DB intact. After revert, history-clearing returns to UI-only behaviour.
- **Feature flags or config toggles:** None — fix is unconditional.
- **Observable failure symptoms:** If the resurrection guard regresses, after clicking Clear all during a long turn, `session_metadata` will contain a row with the just-cleared `session_key` and the `aborted_at` UPSERT timestamp. If the frontend half regresses, "Clear all" leaves the sqlite history intact and a reload restores it.


WareWolf-MoonWall (approved):
## Review Summary

This is a well-reasoned fix for a data-retention bug where "Clear all" only cleared UI state while leaving backend persistence intact. The two-part solution correctly addresses both the frontend oversight (not calling the deletion endpoint) and a subtle race condition in the gateway (cancelled turns can resurrect just-deleted sessions).

The PR body is exemplary: precise scope boundary explanation, explicit rollback path, thorough validation evidence, and clear manual repro. The fix maintains architectural boundaries, adds targeted regression tests, and includes a thoughtful default implementation that works for external backends.

**Verdict:** Approving with one non-blocking suggestion about error visibility.

---

### 🟢 What looks good — Architecture and Trait Design

The `session_exists` trait method with a default implementation is the right pattern here. External backends that don't override it get correct (if slightly less efficient) behavior via `get_session_metadata`, while in-tree backends override with cheap probes (`SELECT 1`, `Path::exists()`). This follows the principle of least surprise for future trait implementers.

The dependency flow is correct: `zeroclaw-infra` defines the trait, `zeroclaw-gateway` uses it, `web` calls the existing REST endpoint. No new coupling introduced (FND-001 §4.2).

---

### 🟢 What looks good — Test Coverage

The three new tests are tightly scoped to the bug:
- `session_exists_tracks_metadata_row` (sqlite) — guards the row that `delete_session` wipes
- `session_exists_tracks_lifecycle` (jsonl) — guards file presence
- `persist_conversation_messages_skips_deleted_session` (gateway) — uses a mock where `session_exists` returns false and asserts `append` is never called

This is surgical regression coverage. The mock-based gateway test is particularly good — it isolates the guard logic without needing the full WebSocket streaming machinery.

---

### 🟢 What looks good — Scope Discipline

The PR body explicitly notes that `set_session_state("running", ...)` at turn start is intentionally not guarded because deletion hasn't happened yet. This is the kind of reasoning that prevents future confusion — the reader knows you thought about that call site and made a conscious choice.

The comment at line 712 in `ws.rs` ties the guard directly back to issue #7126 and explains why the session might not exist. Searchable context for the next person who touches this code.

---

### 🔵 Suggestion — Error Visibility on Deletion Failure

The frontend's `clearAllMessages` calls `deleteSession(sid)` in a best-effort `try/catch` block:

```typescript
try {
  await deleteSession(sid);
} catch {
  // silent
}
```

This is pragmatic — a 404 (when session persistence is disabled) or a transient network failure shouldn't block the local UI clear. But consider logging the failure to the browser console when it happens so users/support can diagnose odd behavior:

```typescript
try {
  await deleteSession(sid);
} catch (err) {
  console.warn('Failed to delete backend session (UI state cleared):', err);
}
```

This way the user gets the immediate "clear" they expect, but there's a trace if the backend portion failed and context unexpectedly reappears. Not blocking — the current behavior is defensible.

---

### 🔵 Suggestion — Label Hygiene

The PR is auto-labeled `gateway` and `tool` but not `bug` or `web` (frontend changed). If you have label permissions, consider adding:
- `bug` (matches the PR title prefix)
- `web` (frontend changes)

The existing `risk: medium` and `size: M` labels look correct.

Not blocking — this is maintainer metadata, not a contributor requirement (per review protocol).

---

### ✅ Validation Evidence Complete

The PR body shows:
- All required checks passed (fmt, clippy, test)
- 7826 tests passed, 0 failed
- TypeScript compilation passed
- Manual repro confirming the bug is fixed
- No skipped commands

This meets the FND-006 standard for medium-risk changes. No follow-up validation needed.

---

## Final Notes

The linked issue #7126 is still open. If you haven't already, mark it as closed by this PR by editing the PR description to include `Closes #7126` (currently shows "Linked issue(s): Closes #7126" in body text but GitHub may not auto-link it depending on formatting).

The `let _ = backend.append(...)` silent-error pattern in the cancel paths was already present before this PR (checked existing code). That's a separate concern if we want to revisit logging append failures, but not this PR's responsibility.

Nice work on the clear bug report → clear fix flow. The manual repro in the PR body is exactly what future readers need to understand the problem this solves.

Audacity88 (approved):
I reviewed the current head c2a0df0a853438fcc494017c2e55b2f5f6b15389, the linked issue #7126, the full diff, the existing WareWolf-MoonWall approval, and the green GitHub checks. I also checked the adjacent gateway session-delete path, WebSocket session-id handling, and in-tree session backend delete semantics from public source context.

### 🟢 What looks good — Clear all now clears the real session

The frontend fix targets the actual bug instead of only hiding it in React state. `clearAllMessages` now calls the existing `DELETE /api/sessions/{id}` path and rebuilds the WebSocket, so the gateway constructs a fresh agent against the now-empty backend session. That matches the user expectation from #7126: clearing the chat should clear the persisted context, not just the rendered transcript.

### 🟢 What looks good — Late gateway writes no longer resurrect deleted history

The backend half addresses the subtle race that made the first fix incomplete. `handle_api_session_delete` already cancels an in-flight turn and deletes the session row, but the old `process_chat_message` cancellation path could wake later and append `"[interrupted by user]"`, which recreated the metadata row. Adding `SessionBackend::session_exists` and checking it before post-turn persistence closes that resurrection path while still sending the client its `aborted` frame.

### 🟢 What looks good — The trait change is narrow and compatible

Adding `session_exists` as a defaulted trait method is the right compatibility shape. Existing external backends still compile through the default `get_session_metadata` fallback, while SQLite and JSONL override it with probes that mirror their delete semantics. That keeps the gateway logic backend-agnostic without pushing SQLite details into the gateway.

### 🟢 What looks good — Validation matches the risk

The PR includes focused regressions for SQLite, JSONL, and the gateway persistence guard, plus full Rust checks, full workspace tests, TypeScript build, and a manual repro of the streaming-clear race. Combined with the current green CI, that is enough evidence for this high-risk gateway/web fix.

No author-side changes are needed from me.

One maintainer-owned metadata cleanup remains before merge: the PR body label snapshot still says `bug`, `risk: medium`, `size: M`, while the live labels are `bug`, `risk: high`, `size: M`, `gateway`, `tool`, and `web`. The live `risk: high` classification is correct for the gateway/persistence/web scope, so the body should be brought into sync.

PR #1234, opened 2026-02-21
fix(agent): add cross-alias close tag resolution and GLM shortened body parsing

vernonstinebaker:
## Summary

- Problem: Models like GLM-4.7 emit malformed tool call formats that the existing parser silently drops — cross-alias close tags (`<tool_call>...</invoke>`), shortened bodies (`tool>value`), YAML-style multi-line, and attribute-style (`tool key="value"`).
- Why it matters: Dropped tool calls cause the agent loop to stall or produce incomplete responses when using GLM-compatible providers.
- What changed: Added cross-alias close tag resolution, `parse_glm_shortened_body()` for 3 GLM sub-formats, `default_param_for_tool()` for parameter inference, and extended `parse_tool_calls()` fallback chain. Merged duplicate match arms for clippy compliance.
- What did **not** change (scope boundary): No changes outside `src/agent/loop_.rs`. Existing parsing paths (JSON, XML, OpenAI native, function_call) are untouched. No config/CLI/provider/channel changes.

## Label Snapshot (required)

- Risk label (`risk: low|medium|high`): risk: low
- Size label (`size: XS|S|M|L|XL`, auto-managed/read-only): size: M
- Scope labels (`core|agent|channel|config|cron|daemon|doctor|gateway|health|heartbeat|integration|memory|observability|onboard|provider|runtime|security|service|skillforge|skills|tool|tunnel|docs|dependencies|ci|tests|scripts|dev`, comma-separated): agent
- Module labels (`<module>: <component>`, for example `channel: telegram`, `provider: kimi`, `tool: shell`): agent: loop
- Contributor tier label (`trusted contributor|experienced contributor|principal contributor|distinguished contributor`, auto-managed/read-only; author merged PRs >=5/10/20/50):
- If any auto-label is incorrect, note requested correction:

## Change Metadata

- Change type (`bug|feature|refactor|docs|security|chore`): bug
- Primary scope (`runtime|provider|channel|memory|security|ci|docs|multi`): agent

## Linked Issue

- Closes #
- Related #
- Depends on #
- Supersedes #

## Supersede Attribution (required when `Supersedes #` is used)

N/A — not superseding any PR.

## Validation Evidence (required)

Commands and result summary:

```bash
cargo fmt --all -- --check        # clean (no diffs in our file)
cargo clippy --all-targets -- -D warnings  # no new warnings in loop_.rs; pre-existing errors in discord.rs, telegram.rs, composio.rs, gemini.rs are unrelated
[...]

PR #7229, opened 2026-06-04
feat(web): MCP, Skills, Plugins & Providers dashboard tabs

theonlyhennygod:
## Summary

- **Linked issue(s):** None
- **Labels:** `enhancement`, `size: XL`, `risk: high`, `gateway`, `tool`, `tool:mcp`, `web`

Adds four first-class dashboard management tabs so operators can manage their stack from the web UI instead of hand-editing config:

- **MCP** — server + bundle management with live connection status, a discovered-tools view, and a "test connection" flow, backed by new gateway endpoints.
- **Skills** — bundle listing plus per-skill editing (reuses the existing `SkillsBundleEditor`).
- **Plugins** — read-only listing that **auto-hides** when the gateway is built without the `plugins-wasm` feature (the route is absent, served as the SPA HTML fallback, and `getPlugins()` returns `null`).
- **Providers** — full model-provider management **replacing the old read-only Integrations tab**, with a searchable 61-provider catalog and an elevated live model picker.

**Backend:** retain the live `McpRegistry` in `AppState` so the dashboard can report per-server connection state; add `GET /api/mcp/status`, `PUT /api/mcp/servers` (the only way to edit list-element fields, since `mcp.servers` is a `#[nested] Vec` section with no per-field config-prop paths), and `POST /api/mcp/test`. MCP secrets (`env`/`headers`) are masked (`***MASKED***`) outbound and round-tripped back from in-memory config on save/test so tokens are never echoed to the client and unchanged secrets aren't clobbered.

- **Scope boundary:** purely additive dashboard surfaces + the gateway endpoints that feed them. No change to MCP/skills/plugin/provider runtime behavior, config schema semantics, or auth. Plugins tab is read-only (no install/remove). The top-level `plugins-wasm` feature forwarding fix that makes `/api/plugins` compile into gateway builds lands in stacked follow-up #7235; this PR's base behavior is to hide the tab when the endpoint does not return JSON.
- **Blast radius:** `web/`, `crates/zeroclaw-gateway`, and `crates/zeroclaw-tools`. Removes the `Integrations` page/route — deep links should move to **Providers** (no redirect added).

## Validation Evidence

```
cargo fmt -p zeroclaw-gateway -p zeroclaw-tools -p zeroclaw-providers -- --check   # clean
cargo clippy -p zeroclaw-gateway -p zeroclaw-tools -p zeroclaw-providers           # clean
cargo test -p zeroclaw-gateway api_mcp     # 3 passed (transport label, masked-secret restore, unknown-key passthrough)
cargo test -p zeroclaw-tools mcp           # 93 passed (incl. per-server status capture)
cd web && npx tsc -b && npx vite build     # clean
```

- **Beyond CI — what did you manually verify?** all four tabs end-to-end — MCP status/test/save round-trip through the existing config persistence path with masked `env`/`headers` restored instead of clobbered; validation rejections; Skills bundle + per-skill editing; Plugins auto-hide on a default (no `plugins-wasm`) build; Providers search, instance add/edit/delete, and live model picker behavior.
- **If any command was intentionally skipped, why:** full-workspace `cargo test` / `clippy --all-targets` / `fmt --all` were deferred to CI because the local checks were scoped Rust/web checks relevant to this PR; GitHub Actions Quality Gate is green for the current head.

## Security & Privacy

- **New permissions, capabilities, or file system access scope? Yes** — authenticated dashboard users can edit the MCP server list and trigger MCP connection tests through the same bearer-protected `/api/*` surface. This does not add a new filesystem scope.
- **New external network calls? Yes** — `POST /api/mcp/test` opens a connection to the **operator-configured** MCP server to validate it; `/api/mcp/status` reports state from already-established connections. No new third-party endpoints beyond what the operator configures.
- **Secrets handling changed? Yes** — MCP `env`/`headers` are masked on the wire to the dashboard and merged back from in-memory config on save/test; tokens are never echoed in plaintext, and on disk they remain encrypted. All new routes require the same bearer-token auth as the rest of `/api/*`.
- **PII, real identities, or personal data in diff, tests, fixtures, or docs? No** — fixtures use placeholder secrets.

## Compatibility

- **Backward compatible? Partially.** Existing config and runtime behavior are unchanged, but dashboard deep links to `/integrations` are not preserved; use `/providers`.
- **Config / env / CLI surface changed? No.**
- **Surface changes:** new gateway endpoints (`/api/mcp/status`, `/api/mcp/servers`, `/api/mcp/test`); the dashboard `Integrations` route/page is removed and replaced by **Providers**. Operators with bookmarks to the old tab should use Providers.

## Rollback

- **Fast rollback:** revert this PR.
- **Toggles:** the Plugins tab is already gated by the `plugins-wasm` build feature.
- **Failure symptoms:** 5xx on `/api/mcp/*`; MCP/Providers tabs failing to load or showing stale status; grep gateway logs for `mcp` handler errors.


WareWolf-MoonWall (changes requested):
This is a well-executed additive feature that brings first-class dashboard management for MCP, Skills, Plugins, and Providers. The implementation is careful about secrets handling, follows existing patterns, and includes the right validation scope. I have one blocking architectural concern about a potential duplicate-state violation, plus several non-blocking observations.

---

### 🔴 Blocking — `McpRegistry::statuses` duplicate state risk requires justification

**Location:** `crates/zeroclaw-tools/src/mcp_client.rs` — the new `statuses: Vec<McpServerStatus>` field in `McpRegistry`.

**Issue:** This introduces a new field that holds per-server connection outcome data (name, connected status, tool count, error message, discovered tools). Per AGENTS.md §"ABSOLUTE RULE — SINGLE SOURCE OF TRUTH", every new struct field must pass the pre-edit ritual:

> State, in your response text, the source of truth for the new data BEFORE you write the field. Two valid answers:
>
> 1. **"This is the source of truth — created here."** OK to write the field. State what it represents.
> 2. **"Source of truth is `<path/to/canonical>` — this would be a duplicate."** Do NOT write the field.

The `statuses` field captures:
- Server name (already in `config.mcp.servers`)
- Connection outcome (success/failure + error message)
- Tool count and tool list (already discoverable via `servers[i].tools()`)

**Why this matters:** If this is duplicate state that could drift from the live `servers` field, it violates the absolute rule and will cause a maintainer revert. If it's a **materialized view captured at startup** that cannot be recomputed from the live servers (because failed connections aren't kept in `servers`), then it's allowed per AGENTS.md §"Patterns that are NOT duplicate state":

> Materialized views built ON-DEMAND from canonical state (cached per-call, not stored).

**Resolution path:** Add a module-level doc comment in `mcp_client.rs` (near the `McpRegistry` definition) that explicitly states:

```rust
//! The `statuses` field is a **startup-time snapshot** of per-server connection
//! outcomes, including failed connections (which are NOT kept in `servers`).
//! This is the source of truth for "what happened at startup" and is immutable
//! after `connect_all` returns. It is NOT a cache of live server state — the
//! live state is in `servers`, which only holds successful connections. This
//! snapshot exists solely for the dashboard `/api/mcp/status` endpoint, which
//! must report both successful and failed servers without re-probing.
```

If this is accurate, the field is **allowed** (it captures state that cannot be reconstructed from `servers` alone because failed connections are dropped). If it's not accurate, the field should be removed and the dashboard should resolve status by calling `servers[i].tools().await` on demand.

**Why blocking:** The AGENTS.md rule is absolute and enforced by maintainer revert. This needs explicit justification before merge so reviewers can validate the reasoning.

---

### 🟡 Warning — Missing `pub(crate)` visibility justification for `error_response` and `persist_and_swap`

**Location:** `crates/zeroclaw-gateway/src/api_config.rs` — changed `fn error_response` and `async fn persist_and_swap` from private to `pub(crate)`.

**Issue:** These were private helpers in the config API module. Making them `pub(crate)` exposes them to the rest of the gateway crate, including the new `api_mcp.rs` module. This is fine, but there's no comment explaining **why** these specific helpers should be shared vs. the new module defining its own error-response logic.

**Why it matters:** Visibility promotions create coupling. If `api_mcp` is the only consumer, a future reader will wonder why these weren't left private with a forwarding wrapper in `api_mcp`. If more modules will need them, a reader will wonder why they live in `api_config` instead of a shared `api_common` module.

**Resolution path:** Add a doc comment to each promoted function explaining its role as a **shared gateway API helper**:

```rust
/// Convert a `ConfigApiError` into an axum `Response` with the correct status.
/// Shared by all `/api/*` endpoints that use the config error type (config,
/// MCP, future config-adjacent surfaces).
pub(crate) fn error_response(err: ConfigApiError) -> Response {
```

Not blocking because the coupling is narrow (one new consumer in the same crate), but it would prevent future "why is this public?" questions.

---

### 🔵 Suggestion — Ollama fix commit should link the build break issue

**Location:** `crates/zeroclaw-providers/src/ollama.rs` — the `temperature` parameter fix.

**Context:** The PR description says this "folds in one small standalone commit fixing a **pre-existing** Ollama build break" but doesn't link the issue or commit hash. The fix itself is correct: the signature expects `Option<f64>` but the code was passing an unwrapped `f64`.

**Suggestion:** When the commit message was written, it should have referenced the issue/PR that introduced the break (or a commit hash if this never landed). Without that reference, a future reader scanning `git log` for "why did this land in a dashboard PR?" won't find the answer. If there's a tracked issue for the build break, link it in the PR description (e.g., "Fixes #1234 (Ollama build break)").

Not blocking because the fix is trivial and correct, but cross-referencing makes the history more navigable.

---

### 🔵 Suggestion — Plugins tab auto-hide mechanic could be documented in the frontend README

**Location:** `web/src/components/layout/Sidebar.tsx` — the `useEffect` hook that probes `/api/plugins` and hides the Plugins nav item when the endpoint returns 404.

**What's good here:** This is a clever feature-detection pattern that degrades gracefully when the gateway is built without `plugins-wasm`. The tab just disappears, no broken UI.

**Suggestion:** Add a comment in `web/README.md` (or a new `web/ARCHITECTURE.md` if one doesn't exist) explaining that **feature-gated routes auto-hide via 404 probes**. This documents the pattern so future contributors know how to add optional tabs without hardcoding build flags into the frontend. Example:

```markdown
## Feature-Gated UI Elements

The dashboard probes optional backend routes (e.g. `/api/plugins`) and hides
nav items when the route returns 404. This keeps the UI in sync with the
gateway's build-time features without shipping feature flags to the frontend.

Pattern: `getPlugins().then((r) => r !== null)` in a `useEffect` hook.
```

Not blocking — the pattern is self-explanatory from the code — but it would help onboard future contributors.

---

### 🟢 What looks good — Secrets masking round-trip is production-ready

**Location:** `crates/zeroclaw-gateway/src/api_mcp.rs` — the `mask_map` / `unmask_from_stored` flow.

**Why this is good:** The secrets handling is exactly right:

1. **Outbound (status/GET):** All `env`/`headers` values are replaced with `***MASKED***` before serializing to JSON. Keys are preserved (so the form can render rows), but values are never echoed. ✅
2. **Inbound (test/PUT):** Masked values are restored from in-memory config **by matching server name**, so an operator can test or save a server config without re-typing tokens they didn't change. ✅
3. **Test coverage:** The `unmask_restores_only_masked_values` test validates that explicitly-changed secrets are **not** clobbered by the restore logic. ✅

This follows the principle from FND-006 §4.5 (Security at the Application Layer): secrets are never logged, never echoed to the client, and only mutated when explicitly changed. The test coverage proves the invariant holds. Excellent work.

---

### 🟢 What looks good — Error propagation with `?` throughout

**Location:** All new backend code in `api_mcp.rs` and `mcp_client.rs`.

**Why this is good:** Every operational error (HTTP client failures, timeout, config validation, server rejection) is propagated with `?` and wrapped in context via `.with_context()` or `bail!()`. The only `.unwrap()` calls added (4 total) are in test assertions, which is the correct use per FND-006 §4.1:

> Programmer errors [...] `panic!` is the correct response, because the goal is to find these at development time, not in front of a user at runtime. [...] `.expect()` with a message explaining why this state is impossible is also appropriate.

The production error paths all return structured errors (`ConfigApiError`, `anyhow::Error`) that the caller can handle. This is textbook error discipline. ✅

---

### 🟢 What looks good — Validation scope matches risk tier

**Location:** PR description "Validation Evidence" section.

**Why this is good:** The author ran scoped validation (fmt/clippy/test) on the three touched crates (`zeroclaw-gateway`, `zeroclaw-tools`, `zeroclaw-providers`) and manually verified the full end-to-end flow (all four tabs, MCP test/save round-trip with masked secrets, plugin auto-hide, provider catalog search). The deferred full-workspace checks to CI are appropriate per the project's workflow — this is a **medium-risk change** (gateway endpoints + dashboard surfaces, no runtime behavior or security boundary impact), not high-risk.

The manual verification covered the critical path: secrets never echoed, validation rejections shown, auto-hide on missing feature. That's the right level of rigor for a dashboard feature. ✅

---

## Verdict

**Request changes** — the `McpRegistry::statuses` field needs explicit source-of-truth justification per AGENTS.md before merge. Once that's added (a module-level doc comment explaining why this is a startup snapshot and not duplicate state), the rest is solid.

The secrets handling is production-ready, the error discipline is exemplary, and the validation scope matches the risk tier. The warnings and suggestions are process improvements, not blockers.

theonlyhennygod:
### Reviewer context — base of a 2-PR stack

**Stack order:** **#7229 (this)** → **#7235** (plugin lifecycle endpoints + management UI, stacked directly on this branch). #7235 must merge **after** this. The rename #7230 is independent of both.

**What this PR delivers:** first-class dashboard tabs for **MCP, Skills, Plugins, Providers** (replacing the old Integrations tab), the read-only `GET /api/plugins`, and the Plugins-tab feature-detection auto-hide.

---

### Responses to @WareWolf-MoonWall’s review

**🔴 Blocking — `McpRegistry::statuses` duplicate-state — resolved-in-principle, doc comment incoming.**
Verified against the code: on a **failed** connection (`mcp_client.rs:288` `Err(e)`) the server is recorded **only** in `statuses` (`:295`) and is **not** pushed to `servers` (success-only push at `:285`). Therefore `statuses` holds startup outcomes — specifically failed connections — that **cannot be reconstructed from `servers`**. Per AGENTS.md §“Patterns that are NOT duplicate state”, this is a permitted **startup-time materialized snapshot**, not a cache of live state. We’ll add the module-level doc comment you proposed near the `McpRegistry` definition making this explicit. No logic change required.

**🟡 Warning — `pub(crate)` `error_response` / `persist_and_swap`:** will add doc comments marking them as shared gateway-API helpers (consumed by `api_mcp.rs`).

**🔵 Ollama build-break link & Plugins auto-hide doc:** will cross-reference the build-break in the description and document the auto-hide pattern.

The full plugin **functionality roadmap** (turning the stubs in #7235 into real installs) is laid out in #7235.

theonlyhennygod:
## 🔗 Status update — dependency unblocked + conflict audit

**Heads-up affecting this PR:** the stacked follow-up #7235 includes a one-line build fix —
```toml
plugins-wasm = ["dep:zeroclaw-plugins", "zeroclaw-runtime/plugins-wasm", "zeroclaw-gateway?/plugins-wasm"]
```
Without it, **this PR’s own `GET /api/plugins` route is never compiled in** (the top-level `plugins-wasm` feature didn’t forward to the gateway crate, so the endpoint returned the SPA HTML fallback in *every* build). Verified: with the forward, `GET /api/plugins` returns `200 application/json` and the Plugins tab renders. We kept the fix in #7235 to avoid touching this branch mid-review, but flagging it here since it’s a prerequisite for the Plugins tab this PR introduces.

**Conflict audit (verified locally with `git merge-tree`):**
- This PR ↔ #7230 (rename): **no conflict** (both edit `i18n.ts`, different lines)
- #7235 is stacked directly on this branch (no conflict by construction)
- All three PRs report `MERGEABLE`

**Remaining to merge:** the one 🔴 blocking item — the `McpRegistry::statuses` doc comment justifying it as a startup snapshot (confirmed accurate against the code: failed connections live only in `statuses`, not `servers`) — plus the two 🟡/🔵 doc/link nits. Happy to push those if useful.

theonlyhennygod:
## ✅ Addressed review feedback — `42668ef5e` (doc-only, no behavior change)

@WareWolf-MoonWall thanks for the thorough review. All points handled:

**🔴 Blocking — `McpRegistry::statuses`:** Confirmed accurate against the code — in `connect_all`, a failed connection (`Err(e)` arm) records into `statuses` only and is **never** pushed to `servers` (success-only). Expanded the field doc to state it explicitly: it's the **source of truth** for startup outcomes *including failures*, immutable after `connect_all`, and **not** a cache of `servers` (which can't reconstruct failed servers). Live tool counts still come from the canonical `servers[i].tools()`. → permitted "startup-time materialized snapshot" per AGENTS.md.

**🟡 `error_response` / `persist_and_swap` `pub(crate)`:** Added doc comments to both marking them as shared gateway-API helpers consumed by `api_mcp` (consistent error mapping + a single validated config-write path), with a note to promote to a shared `api_common` module if a third consumer appears.

**🔵 Plugins auto-hide:** Documented the content-type feature-detection in the Sidebar probe `useEffect` (SPA fallback serves `text/html` → tab stays hidden).

**🔵 Ollama build-break:** The `temperature` `Option<f64>` fix addresses a pre-existing build break unrelated to the dashboard; there's no separate tracked issue, so noting it here for git-log navigability. Happy to add an issue number to the description if one exists.

Re-requesting review. 🔁

WareWolf-MoonWall (approved):
This re-review confirms all blocking and warning findings from the June 4 review have been addressed. The commit 42668ef5 adds the required SSOT justifications without changing any behavior. CI passes cleanly. Ready to merge.

---

### ✅ Resolved — `McpRegistry::statuses` SSOT justification now present

**Original blocking:** The `statuses` field lacked explicit source-of-truth justification per AGENTS.md absolute rule.

**What changed:** Commit 42668ef5 adds a comprehensive doc comment at `crates/zeroclaw-tools/src/mcp_client.rs:255-272` that states:

> Source of truth: this *is* the canonical record of "what happened when we connected at startup". It is NOT duplicate state and NOT a cache of `servers`: `servers` holds only the connections that **succeeded** (see `connect_all` — failures are recorded here at line ~295 and never pushed to `servers`), so the set of failed servers and their error strings cannot be reconstructed from `servers` alone.

**Why this resolves the block:** The justification is accurate — failed connections never enter the `servers` Vec (verified against the `connect_all` code at line ~259: the failure arm only logs, no push), so `statuses` captures state that would otherwise be lost. This is a **startup-time snapshot** (immutable after `connect_all` returns) that fills the gap for the dashboard `/api/mcp/status` endpoint to report both successful and failed servers without re-probing. Per AGENTS.md §"Patterns that are NOT duplicate state", this is an allowed materialized view: it's built ON-DEMAND at startup from canonical state and represents facts (failure outcomes) that are not stored anywhere else.

The doc comment references the exact code line where the distinction happens (~295) and explicitly names the consumer (`GET /api/mcp/status`). Future maintainers can validate the reasoning by inspection. Block lifted. ✅

---

### ✅ Resolved — `pub(crate)` visibility justifications added

**Original warning:** The `error_response` and `persist_and_swap` promotion to `pub(crate)` lacked inline rationale.

**What changed:** Commit 42668ef5 adds doc comments at `crates/zeroclaw-gateway/src/api_config.rs:315-320` (for `error_response`) and `:413-420` (for `persist_and_swap`) explaining:

- **`error_response`**: "Shared gateway-API helper: every `/api/*` surface that speaks `ConfigApiError` (the config endpoints, and now `api_mcp`) renders errors through this single function so HTTP status mapping stays consistent."
- **`persist_and_swap`**: "Config-mutating endpoints outside `api_config` (the new `api_mcp` server add/remove handlers) must go through this exact path so every write gets the same validation, atomic file persistence, and in-memory swap — rather than each module reimplementing it and risking drift."

Both comments explicitly name the new consumer (`api_mcp`) and explain the coupling rationale (consistency across modules). The `error_response` comment also includes a forward-looking note: "if a third consumer appears, move it (and `persist_and_swap`) to a shared `api_common` module." This answers the future-reader question preemptively. Warning addressed. ✅

---

### ✅ Resolved — Plugins auto-hide mechanic documented inline

**Original suggestion:** The Plugins tab feature-detection pattern could be documented so future contributors understand the 404-probe pattern.

**What changed:** Commit 42668ef5 expands the inline comment at `web/src/components/layout/Sidebar.tsx:106-112` to a full explanation:

> The `/api/plugins` route only exists when the gateway is built with the `plugins-wasm` feature. In builds without it the route is simply not registered, and the SPA catch-all answers unknown `/api/*` paths with `200 text/html` (not a 404) — so `getPlugins()` detects "unsupported" by a non-JSON content type and resolves to `null`.

**Why this is sufficient:** The pattern is now self-documenting at the call site. The comment explains:
1. **What triggers the behavior** (missing `plugins-wasm` feature).
2. **Why a 404 probe doesn't work** (SPA fallback answers with 200 HTML).
3. **How detection works** (content-type check in `getPlugins()`).
4. **The outcome** (tab hidden until JSON confirms support).

A future contributor adding an optional tab can grep for "Feature-detect" or "getPlugins" and find the pattern. The original suggestion mentioned a `web/README.md` or `web/ARCHITECTURE.md` entry; that would still be useful (especially if more optional tabs appear), but the inline doc is sufficient for now — it teaches the pattern where it's implemented. Suggestion satisfied. ✅

---

### 🟢 What looks good — Documentation-only commit with precise scope

**Location:** Commit 42668ef5 (`docs(mcp,gateway,web): address #7229 review feedback`).

**Why this is good:** The follow-up commit is exactly what was requested: three targeted doc expansions, zero behavior change. The commit message is explicit ("Documentation only; no behavior change") and scoped to the three files that needed justification. This is the gold standard for responding to review feedback — it shows the author took the time to understand the concern, added the right context, and resisted the temptation to mix in "while here" refactors or feature tweaks. The `Co-Authored-By` tag correctly attributes the assistant's role. Exemplary process. ✅

---

### 🟢 What looks good — CI passing across all targets

**CI status:** 14/14 checks passed (format, lint, security, test, benchmarks compile, builds for aarch64-apple-darwin, x86_64-pc-windows-msvc, x86_64-unknown-linux-gnu, 32-bit check, all-features check, no-default-features check).

**Why this matters:** The PR touches three separate subsystems (gateway, tools, web) and adds a new API surface (`api_mcp.rs`). A clean CI run confirms:
- The Ollama fix (commit 0ccab34) actually resolves the build break it claims to fix (no more `Option<f64>` mismatch).
- The new MCP endpoints compile under all feature combinations (including `plugins-wasm` present/absent).
- The web build succeeds (TypeScript + Vite).
- No clippy warnings or formatting drift.
- The 93 `mcp` tests and 3 new `api_mcp` tests all pass.

The "medium-risk" classification in the original review was correct — this is additive dashboard work, not a runtime or security-boundary change — and the full CI suite validates the PR didn't break anything adjacent. ✅

---

## Verdict

**Approve** — all findings from the June 4 review are resolved, no new issues surfaced, CI clean, and the documentation additions are precise and sufficient. The PR is ready to merge.

The blocking SSOT concern is addressed with a rigorous justification that stands up to the AGENTS.md absolute rule. The warning and suggestion items are handled appropriately (inline docs where requested, commit discipline exemplary). The original praise points (secrets masking, error propagation, validation scope) remain valid.

Stack note: per the author's comment, this is the base of a 2-PR stack with #7235 stacked on top. Merge order is correct (this PR first, then #7235).

Audacity88:
CI is still red on the current merge build. The failure looks like a small integration miss from the new `AppState::mcp_registry` field meeting newer Linq test helpers on `master`.

The failing job reports two missing `mcp_registry` fields in `crates/zeroclaw-gateway/src/lib.rs`:

- `linq_test_state(...)`
- `linq_webhook_returns_not_found_when_no_channels_configured`

Both should be fixed by adding `mcp_registry: None,` to those test `AppState` literals.

One other small cleanup: the PR body still mentions the standalone Ollama/provider build-break fix, but the current diff no longer includes `crates/zeroclaw-providers`, so that section looks stale now.

Audacity88 (changes requested):
I reviewed the current head dfb5c3c2, the updated PR body, the prior WareWolf-MoonWall reviews, the top-level comments, the inline-comment feed, and the changed gateway/tools/web paths. CI is now green after the recent master-merge/test-helper fixes, and WareWolf-MoonWall's earlier `McpRegistry::statuses` source-of-truth block is resolved by the follow-up documentation.

This review does not re-open that resolved thread. It surfaces a separate MCP secret-handling edge in the dashboard save flow that was not called out in the prior approval: renaming a server can still replace a saved credential with the literal mask string.

### ✅ Resolved — Prior review feedback was addressed cleanly

The resolved `McpRegistry::statuses` documentation now explains why failed startup connections need their own startup outcome record, and the dashboard status endpoint keeps live connection reporting separate from config mutation. Reusing the existing config persistence path for `PUT /api/mcp/servers` is also the right instinct: the fix below should preserve that single validated write path rather than forking config persistence.

### 🔴 Blocking — MCP server rename can persist `***MASKED***` as the credential

The MCP status endpoint now sends each configured server back with `env` and `headers` values masked. The UI then makes `Name` editable and saves the whole server list. On save, the backend restores masked secrets by looking up the submitted row by its current `server.name`:

- `web/src/pages/Mcp.tsx:446` lets the user edit `server.name` in the server card, and `web/src/pages/Mcp.tsx:149` saves the full list with `putMcpServers(next.map(normalizeServer))`.
- `crates/zeroclaw-gateway/src/api_mcp.rs:220` restores masked values with `working.mcp.servers.iter().find(|s| s.name == server.name)`.

That works for ordinary edits where the name is unchanged. It fails when the operator renames an existing MCP server and does not retype the secret fields. The submitted row now has the new name plus masked `env` / `headers` values. The name lookup misses the stored row, so `unmask_from_stored` never runs. Validation then accepts the row, `persist_and_swap` saves it, and the literal `***MASKED***` value becomes the configured credential for the renamed server.

This is separate from the already-reviewed same-name secret restore path. The ordinary same-name save/test case is covered; the missing edge is that `name` itself is editable and is also the restore key.

That is a data-loss bug at the new security boundary: the dashboard promises “unchanged secrets aren’t clobbered,” but a normal rename path clobbers them. It is the same class of issue the generic config API already guards against when it refuses to overwrite secret fields with a masked sentinel.

Please make the restore path stable across rename before merge. A few safe options would work:

- carry an `original_name` / stable row id from status to save and restore secrets from that original row;
- make `name` immutable in this editor and expose rename as a dedicated delete/add or rename flow;
- or reject any submitted masked `env` / `headers` value that cannot be restored, so the API never persists the display sentinel as a real secret.

Please also add a regression in `crates/zeroclaw-gateway/src/api_mcp.rs` near the existing unmask tests. It should start with an existing server containing `env` or `headers`, submit the same row under a new name with the secret value still `***MASKED***`, and prove the saved config keeps the original secret or rejects the request instead of persisting the mask.

### 🟡 Warning — Plugin feature detection hides real plugin endpoint failures

`getPlugins()` now treats every non-OK response as `null` at `web/src/lib/api.ts:492`, and both the Sidebar and Plugins page interpret `null` as “plugin support is not enabled in this build.” That is correct for the SPA fallback case, where the route is absent and the response is `200 text/html`. It is not correct for a real registered `/api/plugins` failure.

For example, if the plugin route exists but plugin discovery returns a 500 JSON error, this code hides the Plugins nav item and the direct Plugins page renders the “not enabled in this build” state instead of surfacing the backend failure. The comment says real errors still propagate, but the implementation currently suppresses any `!response.ok`.

Please narrow the unsupported case to the actual absent-route signal, such as non-JSON SPA fallback or an explicit 404 if that path becomes available, and let real JSON error responses propagate like the rest of the API layer.

singlerider:

PR #6922, opened 2026-05-25
fix(runtime): deliver only final assistant turn's content to channels

AbubakrSamsodien:
Models with interleaved thinking (like GLM-5) emit content alongside tool_calls on intermediate turns. run_tool_call_loop was appending that text into accumulated_display_text on every iteration, then returning the whole buffer as the agent's deliverable response. Chain-of-thought leaks into channel messages even when the model's final no-tool-calls turn is a clean one-liner.

This stops accumulating intermediate-turn display_text. Only the final turn populates accumulated_display_text. Streaming consumers still receive intermediate text live via the existing tx.send(StreamDelta::Text(...)) path immediately below, so that behavior is unchanged.

Behavior change:
- stream_mode = "off": channel reply used to be the concat of every turn's content. Now it's just the final turn's content.
- stream_mode = "partial" / "multi_message": live updates during the run are unchanged. The final delivered message (or finalized draft) now contains only the final turn's content rather than the accumulated narration.

if some of us prefer the existing behavior then maybe we should have a config setting to switch accumulated display with channels on or off.

Ran cargo test locally. Everything passed.

---

## Maintainer note

Related to #6510.

Same root cause: `accumulated_display_text` in `run_tool_call_loop` accumulates intermediate-turn narration into the final delivered output. Same file (`crates/zeroclaw-runtime/src/agent/loop_.rs`), same mechanism.

**Scope difference — does not auto-close #6510.** The issue asks for an opt-in flag on `delivery.mode = "announce"` so existing consumers that rely on accumulated narration can keep it. This PR changes the default unconditionally and (per the PR description) only floats the config flag as a future possibility. If we land this as-is, that's a behaviour change for anyone whose cron output happens to depend on intermediate narration being concatenated, and the `[Feature]` opt-in shape #6510 requested still isn't delivered.

Two paths forward:
1. Land this PR as the new default and close #6510 with a note that the opt-in flag was deemed unnecessary because the new default is what most users want, **or**
2. Extend this PR to add the config flag (`delivery.only_final_message: bool` default `true`) so the change is opt-out for the few who want the old behaviour.

Flagging for review discussion. Leaving #6510 open until that's decided.

— @singlerider, 2026-05-28

PR #7239, opened 2026-06-05
fix(runtime): deliver only final assistant turn to channels

### LINKED ISSUE #7068: Telegram channel can receive Codex scratchpad/tool transcript as final response
## Summary

While running ZeroClaw with Codex as the delegated agent/backend instead of direct LLM API providers, a Telegram message was acknowledged by the daemon, but the bot then sent an internal scratchpad/tool-transcript style response instead of a normal user-facing answer.

The observed response began with text similar to:

> [Tool exchange: ... results collapsed]

It then included internal planning language about inspecting workspace/config and tool usage. This appears to be an internal agent-loop or delegated-agent artifact being routed through the normal final-answer delivery path.

## Impact

This can make the Telegram bot appear broken or confusing to users. More importantly, it risks exposing internal execution traces, tool-planning notes, or hidden reasoning-like content through a public channel response.

The issue is especially relevant when ZeroClaw delegates work to another CLI agent such as Codex. In that mode, ZeroClaw must distinguish between the delegated agent's internal/tool transcript and the final user-facing answer.

## Expected behavior

Telegram should only receive user-visible final responses.

Internal agent events should be structurally separated from channel-visible output, for example:

- delegated-agent/tool calls
- delegated-agent/tool results
- planning/scratchpad text
- compaction/status events
- final user-facing answer

Only the final user-facing answer, or explicitly safe progress/status events, should be sent to Telegram.

## Actual behavior

The daemon processed the Telegram message and delivered a final response containing internal scratchpad/tool-transcript content from the delegated Codex workflow.

## Suggested fix

Avoid relying on string matching to detect leaked scratchpad text. Instead, make agent-loop output typed so channel delivery can only consume safe user-visible events by construction.

A possible direction:

- Introduce a typed `AgentEvent` / `AgentTurnResult` model.
- Keep delegated-agent tool traces, scratchpad/progress events, and final answers as separate variants.
- Allow channels to subscribe only to safe event types.
- For Telegram, send only `UserVisibleFinal` plus optional safe status events.
- Keep internal/delegated-agent traces in logs/observability, not in final response text.

## Notes

A narrow defensive guard can suppress known leaked transcript formats, but that should be treated as a temporary mitigation. The durable fix should be structural separation between internal runtime events and channel-visible output, especially for CLI-agent backends such as Codex.

rifuki:
## Summary

- **What changed and why:**
  - Stop accumulating intermediate assistant text from tool-call turns into `run_tool_call_loop`'s final return value.
  - Keep intermediate native-tool narration available to draft/live channel updates through the existing `on_delta` path.
  - Tighten the existing regression so final channel delivery is exactly the final assistant turn, not merely a string that ends with it.
- **Scope boundary:** Does not introduce typed `AgentEvent` plumbing, provider finish-reason handling, or new configuration. This is the narrow defensive runtime fix requested by the existing regression review.
- **Blast radius:** Runtime tool-loop final response assembly for channel callers. CLI/live draft progress still receives intermediate text as before.
- **Linked issue(s):** Closes #7068. Supersedes #6922.

Local validation is the signal CI cannot replace. Run the full battery and paste literal output (tails, failures, warnings - not "all passed").

```bash
cargo fmt --all -- --check
cargo build
cargo test
```

`cargo fmt --all -- --check`

```text
(no output; command exited 0)
```

`cargo clippy --all-targets -- -D warnings`

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.03s
```

`cargo build`

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
```

`cargo test`

```text
     Running tests/test_system.rs (target/debug/deps/system-ef5f50a7dbd39851)

running 9 tests
test system::multi_agent_e2e::legacy_install_upgrades_cleanly_with_backup ... ok
test system::multi_agent_e2e::peer_group_dotted_channel_refs_remain_alias_scoped_for_dispatch ... ok
test system::multi_agent_e2e::peer_group_routes_messages_only_within_resolved_peer_set ... ok
test system::full_stack::system_simple_text_response ... ok
test system::full_stack::system_multi_turn_conversation ... ok
test system::full_stack::system_parallel_tool_execution ... ok
test system::full_stack::system_tool_arguments_passed_correctly ... ok
test system::full_stack::system_tool_execution_flow ... ok
test system::multi_agent_e2e::two_sqlite_agents_on_one_install_have_isolated_memory ... ok

test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s

   Doc-tests zeroclaw

running 0 tests

```

Focused regression:

```bash
cargo test -p zeroclaw-runtime run_tool_call_loop_relays_native_tool_call_text_via_on_delta
```

```text
running 1 test
test agent::loop_::tests::run_tool_call_loop_relays_native_tool_call_text_via_on_delta ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1991 filtered out; finished in 0.05s
```

- **Beyond CI - what did you manually verify?** Reviewed #7068 and #6922. Confirmed the regression now locks both contracts: intermediate native-tool text is still emitted through `on_delta`, while the final delivered result equals only the final assistant response.

## Compatibility (required)

- If `No` or `Yes` to either: Existing users do not need to change config. The intentional behavior change is that final channel replies no longer concatenate intermediate tool-call narration; live draft/progress delivery remains unchanged.

- **Observable failure symptoms:** Channel replies lose useful intermediate narration in final delivered messages; grep for `turn_final_response` and compare final channel output against live draft updates.

  - #6922 by @AbubakrSamsodien
- Scope materially carried forward: Carries forward the narrow runtime behavior change from #6922: intermediate tool-call-turn display text is no longer accumulated into the final channel response.
- `Co-authored-by` trailers added in commit messages for incorporated contributors? (`Yes`)


Audacity88 (approved):
I checked the live PR state, #7068, the superseded #6922 thread, the single-file runtime diff, the existing `run_tool_call_loop` regression, and the channel orchestrator path that sends the loop return value as the final channel response. CI is green. I am approving this as the narrow runtime fix for the final-delivery leak; it is intentionally not the larger typed-event redesign described as the long-term direction in #7068.

### 🟢 What looks good — The patch fixes the right boundary

The leak happens because channel delivery consumes the string returned by `run_tool_call_loop`. This PR stops intermediate tool-call-turn narration from being appended to that returned string, while keeping the existing live `on_delta` path for draft/progress updates. That is the smallest useful boundary for this bug: channels get only the final assistant turn as the delivered reply, but streaming consumers still see intermediate safe progress where they already did.

### ✅ Resolved — #6922's regression blocker is covered

The prior blocker on #6922 asked for a regression proving two things at once: intermediate native-tool narration still reaches `on_delta`, and the final returned result is exactly the final assistant turn. The updated `run_tool_call_loop_relays_native_tool_call_text_via_on_delta` assertion now locks that contract instead of only checking that the result happened to end with the final answer.

### 🔵 Suggestion — Keep the structural follow-up separate if the original scenario still reproduces

#7068 also sketches a fuller typed `AgentEvent` / `AgentTurnResult` design. I do not think that needs to block this patch, because this PR fixes the concrete accumulation bug at the channel-delivery boundary. If the original Telegram + Codex delegated-agent scenario still leaks after this lands, that should be reopened or split into a follow-up for the broader typed-event separation rather than folded into this one-line runtime fix.

PR #7240, opened 2026-06-05
fix(zerocode): make quickstart provider alias editable

### LINKED ISSUE #7227: [Bug]: zerocode Quickstart hardcodes model-provider alias to `default`, colliding with existing providers
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

The zerocode TUI Quickstart hardcodes the new model-provider alias to `default` and never exposes it for editing. The provider field form only renders `model` / `api_key` / `uri` rows — there is no alias field.

If the config already contains `[providers.models.<type>.default]`, picking that provider type makes Quickstart impossible to complete: committing the provider fails validation with `alias \`<type>.default\` already exists`. The status strip shows only `1 error(s) — fix selectors and resubmit`, so the cause is invisible (see related note below).

Source:

- `apps/zerocode/src/quickstart_pane.rs`, `open_field_form()`:
  `let alias = match section { QuickstartFieldSection::ModelProvider => "default".to_string(), _ => type_key.clone() };`
- Collision raised in `apply_model_provider()` in `crates/zeroclaw-runtime/src/quickstart/mod.rs`.

### Expected behavior

The user can set the provider alias (or Quickstart auto-picks a non-colliding one), so an existing `<type>.default` doesn't turn the flow into a dead end.

The **web** Quickstart already exposes an editable alias input — `web/src/pages/quickstart/Quickstart.tsx`, `ProviderForm` (`const [alias, setAlias] = useState("default")` plus a `LabeledInput`). The zerocode TUI should reach parity: either an editable alias row, or auto-dedupe (`default`, `default-2`, …) using the existing aliases already present in the Quickstart state snapshot.

### Steps to reproduce

```bash
# Pre-req: config already has a provider under the `default` alias, e.g.
#   [providers.models.openai.default]
cargo build -p zerocode --bin zerocode
cargo build --bin zeroclaw            # the daemon the TUI spawns
./target/debug/zerocode               # defaults to ~/.zeroclaw

# In the TUI:
# 1. Open the Quickstart pane (mode bar)
# 2. "Model provider" -> Enter
# 3. Pick "OpenAI" (the type that already has a `default` alias)
# 4. Accept the catalog-seeded model -> press Enter to commit
#
# Status strip shows: "1 error(s) — fix selectors and resubmit"
# Underlying error (only visible via code instrumentation today):
#   step=ModelProvider  field=alias  msg=alias `openai.default` already exists
```

### Impact

Affected users: anyone running zerocode Quickstart who already has a provider of the chosen type under the `default` alias — common, since `default` is the conventional alias.
Frequency: always, for that provider type.
Consequence: Quickstart cannot complete via zerocode for the affected provider type; the workaround is the web/CLI surfaces (which allow alias editing).

Related: #7117 (config UX parity tracker — thematically covers this but does not call out the alias collision). This is compounded by a sibling gap: the zerocode Quickstart status strip shows only an error count, not the message, which hides the real cause.

### Logs / stack traces

```text
status strip (all the user sees):
  1 error(s) — fix selectors and resubmit

underlying QuickstartError, surfaced via temporary instrumentation of draw_status_strip():
  step=ModelProvider  field=alias  msg=alias `openai.default` already exists
```

### ZeroClaw version

v0.8.0-beta-2 (master; observed on commit f64ebea5)

### Rust version

### Operating system

macOS 26.3 (arm64)

### Regression?

No, first-time setup

### Pre-flight checks

chengzhichao-xydt:
## Summary

- **Issue:** Closes #7227
- **What changed:** The zerocode ModelProvider quickstart form now exposes an editable `Alias` field instead of forcing `default`.
- **Why:** Existing configs commonly already contain `[providers.models.<type>.default]`; forcing that alias made Quickstart fail validation with an alias collision.
- **Scope boundary:** Confined to the zerocode Quickstart ModelProvider form and its English Fluent strings. No daemon validation, runtime config, web Quickstart, or CLI behavior changes.
- **Labels:** `bug`, `risk: medium`, `size: XS`, `onboard`, `quickstart`

## Files Changed

- `apps/zerocode/src/quickstart_pane.rs`: editable alias field, skip alias in `provider_fields`, remove static alias header.
- `apps/zerocode/locales/en/zerocode.ftl`: Alias label and help text.

## Validation

Fresh GitHub Actions passed on refreshed head `4af05c6`, including the required CI gate. The change was also reviewed against the form field lifecycle: the alias row is limited to ModelProvider, non-ModelProvider sections keep the existing fallback path, and empty aliases still fall back safely.

## Security & Privacy Impact

## Compatibility

- Backward compatible? `Yes`
- Upgrade steps: None.

## Rollback

`git revert <sha>` restores the previous zerocode Quickstart alias behavior. No config schema, daemon validation, or runtime changes are included.


WareWolf-MoonWall (approved):
## Review Summary

This PR fixes a real workflow blocker (issue #7227) where the zerocode TUI Quickstart hardcoded the model provider alias to `"default"` with no way to edit it, causing validation failure when the config already contained a provider under that alias. The fix adds an editable alias field row to the ModelProvider form, pre-filled with `"default"`.

**Reviewed commit**: 631075f00aea66c195a20d7f7c9f923fc8b1fae6  
**Reviewer**: @WareWolf-MoonWall  
**Verdict**: Approved with notes

---

### 🟢 What looks good — Problem statement and scope

The PR description is clear, includes linked issue, verification notes, and backward-compatibility considerations. The fix is minimal and surgical: it touches only the TUI Quickstart pane code and the relevant Fluent i18n keys. The two-commit history shows good incremental work — the first commit implements the feature, the second fixes a `mut` declaration the first commit needed.

The alias field insertion at `rows.insert(0, ...)` (line 1544 in the diff) puts the alias at the top of the form, which is the right UX choice — users see it first and can change it before filling in other fields.

---

### 🟢 What looks good — Backward compatibility and fallback

The commit logic at lines 1620–1630 in the new diff reads the alias from the editable field row via `pick("alias")`, but falls back to `f.alias` (the auto-generated value from `open_field_form()`) when the picked value is empty. This covers the non-ModelProvider sections (which don't get the injected alias row) and any edge case where the alias field is somehow empty.

The gating on `matches!(section, QuickstartFieldSection::ModelProvider)` (lines 1542–1564) keeps the change scoped to model providers only, which matches the issue scope and avoids touching unrelated channel/runtime/peer-group form flows.

---

### 🟢 What looks good — Localization

Both new Fluent keys (`zc-quickstart-field-label-alias` and `zc-quickstart-field-help-alias`) are present and descriptive. The help text explicitly tells the user *why* they might want to change the alias (`"Change from 'default' if your config already has one with that alias"`), which is exactly what a user hitting the issue needs to see.

---

### 🟡 Warning — Duplicate state in FieldFormModal.alias

**Location**: `apps/zerocode/src/quickstart_pane.rs`, lines 1540–1550 (post-patch)

The `FieldFormModal` struct still holds an `alias: String` field that is set at form-open time (line 1547: `alias: alias`) and later used as a fallback in `commit_field_form()`. Meanwhile, the new code injects a separate `FieldFormRow` with `key: "alias"` (line 1548 in the diff: `key: "alias".to_string()`).

This creates two sources of truth for the alias:

1. `FieldFormModal.alias` — set at form-open time from `open_field_form()` logic (lines 1540–1543)
2. The `FieldFormRow` with `key == "alias"` in `FieldFormModal.fields` — inserted at line 1544, user-editable

The commit logic (lines 1620–1630) reads from the editable row via `pick("alias")` but falls back to `f.alias`. The fallback is necessary today because non-ModelProvider sections don't get the injected alias row, but it means the `alias` field in `FieldFormModal` is now serving two roles: the display-only alias for non-ModelProvider sections, and the fallback default for ModelProvider sections.

**Why this is a 🟡 warning rather than 🔴 blocking**: The current implementation works correctly for the stated scope (ModelProvider alias editing) and maintains backward compatibility. However, this is exactly the duplicate-state pattern AGENTS.md §"ABSOLUTE RULE" calls out: a fact that lives in two places (the struct field and the field row), creating a potential drift risk if future code touches `FieldFormModal.alias` without accounting for the editable row.

**Recommended follow-up** (not blocking this PR): Consider one of these refactors in a later PR:

1. **Option A (cleanest)**: Remove `FieldFormModal.alias` entirely. Always inject an alias row for all sections (not just ModelProvider), marking it `required: true` for all. The commit logic then always reads from the row. Non-ModelProvider sections pre-fill it with `type_key.clone()`, ModelProvider pre-fills with `"default"`. This eliminates the duplicate.

2. **Option B (minimal surgery)**: Document in a code comment above `FieldFormModal.alias` that this field is a fallback for non-ModelProvider sections only, and that ModelProvider sections read from the injected `"alias"` row. This makes the intent explicit without changing the structure.

I lean toward Option A because it removes the special-case branch and makes the form structure uniform, but either is acceptable.

---

### 🔵 Suggestion — Render consistency for the alias field

**Location**: `apps/zerocode/src/quickstart_pane.rs`, `draw_modal()` function, lines 1890–1901 (pre-patch), lines 1923–1928 (post-patch)

The diff removes the static `"Alias: <value>"` display from the modal header (lines 1899–1900 in the pre-patch code), which is correct — the alias is now in the editable field rows below, so showing it in the header would be redundant.

However, the pre-patch header showed both `Type:` and `Alias:` side-by-side. The post-patch header shows only `Type:`. This is fine for ModelProvider forms (where the alias is now a form row), but it means non-ModelProvider forms (channels, etc.) no longer show their alias anywhere in the header.

If that's intentional (because channels auto-generate their alias from `type_key` anyway), this is not a problem. But if the alias was a useful piece of context for the user to see at a glance, you might want to conditionally render it in the header for non-ModelProvider sections only:

```rust
// In draw_modal(), Modal::FieldForm branch
let header_spans = if matches!(section, QuickstartFieldSection::ModelProvider) {
    vec![
        Span::styled(format!("{} ", t("zc-quickstart-modal-type-prefix")), theme::dim_style()),
        Span::styled(f.type_key.as_str(), theme::accent_style()),
    ]
} else {
    vec![
        Span::styled(format!("{} ", t("zc-quickstart-modal-type-prefix")), theme::dim_style()),
        Span::styled(f.type_key.as_str(), theme::accent_style()),
        Span::styled("    Alias: ", theme::dim_style()),
        Span::styled(f.alias.as_str(), theme::body_style()),
    ]
};
lines.push(Line::from(header_spans));
```

This is optional — the current behavior is not incorrect, just a UX choice.

---

### 🔵 Suggestion — Validation feedback for alias collisions

**Context**: The original issue (#7227) noted that when the alias collides, the status strip shows only `"1 error(s) — fix selectors and resubmit"` without revealing the actual message (`alias 'openai.default' already exists`). This PR fixes the root cause (users can now edit the alias), but it doesn't improve the error visibility.

**Suggestion** (out of scope for this PR, but worth tracking): Consider a follow-up issue to surface the actual validation error message in the status strip, not just the count. This would help users understand *why* the commit failed, not just *that* it failed. The issue mentions this as a "compounded by a sibling gap" — it's a separate UX improvement that would help in cases where the user still picks a colliding alias by accident.

---

## Verification

- ✅ CI passes (all Quality Gate checks green, including Format, Lint, Test, Build on 3 platforms, Security)
- ✅ Diff reviewed against local `apps/zerocode/src/quickstart_pane.rs` — changes land in the correct methods
- ✅ Fluent keys added to `apps/zerocode/locales/en/zerocode.ftl` — no missing i18n refs
- ✅ No test changes — Quickstart pane is TUI-only, manual verification expected per PR description

---

## Final Notes

This is a clean, targeted fix for a real blocker. The 🟡 warning about duplicate state is a medium-term architecture note, not a merge blocker — the code works correctly as written. The two 🔵 suggestions are optional polish that can be deferred to follow-up issues if desired.

Great work on the clear PR description and the backward-compatible implementation, @chengzhichao-xydt.

PR #7242, opened 2026-06-05
fix(config): skip quoted and heredoc tilde tokens in path-policy

### LINKED ISSUE #7133: [Bug]: path-policy false-positive on ~ tokens in quoted/heredoc command data
### Affected component

security/sandbox

### Severity

S2 - degraded behavior

### Current behavior

`forbidden_path_argument` in `crates/zeroclaw-config/src/policy.rs` tokenizes the entire command string with `split_whitespace()` and applies `looks_like_path` + `is_path_allowed` to every token — including tokens inside quoted string literals and inside heredoc bodies (which are stdin data, not argv path arguments). `looks_like_path` treats any token beginning with `~` as a path, so non-path literals such as `~20` are rejected as forbidden home-paths.

Empirically reproduced in-session:

| Case | Result |
|---|---|
| `cat <<'EOF'` heredoc body containing `~20` | BLOCKED |
| `printf '...~20...'` | BLOCKED |
| `echo "...~20..."` (double-quoted arg) | BLOCKED |
| `echo '...~20...'` (single-quoted arg) | BLOCKED |
| `echo "...tilde~mid..."` (tilde not at token start) | passes |
| `echo "/etc/foo"` (slash path in allowed root) | passes |

This first surfaced when a `gh issue create` heredoc body contained the literal `~20` ("about 20"), producing `Error: Path blocked by security policy: ~20` and aborting the whole command.

### Expected behavior

The path-argument guard should reject genuine forbidden **path arguments** while not misclassifying:

1. Heredoc body content — it is stdin data piped to the command, never an argv path argument. The splitter already tracks heredoc state; that body content must be excluded from path-argument inspection.
2. Non-path literals such as `~20`, `~5` — a leading `~` should only be treated as a home-path when the token is `~` or `~/...`, not `~<non-slash>`.

The security property must be preserved: commands that pass a forbidden absolute or home path as an actual path argument (e.g. `cat /etc/shadow`, `cat ~/.ssh/id_rsa`) must still be blocked. This is a false-positive narrowing only — do not stop scanning for forbidden path arguments.

### Steps to reproduce

```bash
# any of these are rejected with "Path blocked by security policy: ~20"
cat <<'EOF' > /tmp/x.txt
this line has ~20 percent in it
EOF

printf 'plain string with ~20 percent\n' > /tmp/x.txt

echo "tilde token ~20 here" > /tmp/x.txt
```

### Impact

Affected users: anyone whose shell command — argv string literal or heredoc body — contains a token beginning with `~` followed by a non-slash character (percentages, approximations like `~20`, `~5`), or a forbidden-looking path embedded in quoted data. Common in `gh issue/pr` heredoc bodies, commit messages, and echo/printf payloads.
Frequency: deterministic — always blocks.
Consequence: legitimate commands are rejected; no security exposure, but a real workflow blocker.

### Logs / stack traces

```text
Error: Path blocked by security policy: ~20
```

### ZeroClaw version

current master (workspace checkout, 2026-06-03)

### Rust version

n/a — logic bug in policy crate, version-independent

### Operating system

Arch Linux

### Regression?

Unknown

### Pre-flight checks

---

### Additional validation evidence (live reproduction, 2026-06-03)

Reproduced deterministically again on current master. These add two vectors not in the original table — a `python3 -c` argv string literal, and the observation that a tilde followed by **letters** is also misclassified — plus a first-token-wins detail. Token values are written with the tilde spelled as `TILDE` in this body so editing the issue over a shell does not itself re-trip the guard; in the real commands they were a literal tilde immediately followed by the characters shown.

**1. A `TILDE20`-style token inside a `python3 -c` string literal (not a path, not a heredoc):**

```text
$ python3 -c "print('about TILDE589 lines here')"
Error: Path blocked by security policy: TILDE589
```

The token is the sole content of a quoted Python string argument. It is never an argv path; the guard inspects it anyway.

**2. Tilde-number tokens inside a quoted heredoc body (stdin data):**

```text
$ cat <<'EOF'
this line has TILDE20 percent and TILDE589 too
EOF
Error: Path blocked by security policy: TILDE20
```

**3. First token wins.** The case-2 body holds two tilde-number tokens; the reported one is the first encountered (`TILDE20`), confirming the scan walks `split_whitespace()` over the whole command — heredoc body included — rather than argv only.

**4. Tilde-then-letters is also flagged.** A token of a tilde followed by alphabetic characters trips the same guard, so the narrowing must key on the token being exactly `~` or beginning `~/` — not merely "tilde then non-digit". Any `~<non-slash>` shape must pass.

These reinforce the two narrowing fixes in Expected behavior:

- Heredoc body content must be excluded from path-argument inspection (the splitter already tracks heredoc state).
- A leading tilde classifies as a home-path only when the token is exactly `~` or begins `~/` — never `~<non-slash>`.

Security property to preserve: genuine forbidden path arguments (home-dir SSH keys, `/etc/shadow`, etc.) must still be blocked. This is a false-positive narrowing only.

chengzhichao-xydt:
## Summary

- **Issue**: Closes #7133
- **What changed**: The `forbidden_path_argument` method in `SecurityPolicy` was producing false-positive path-policy warnings for `~` tokens inside quoted strings or heredoc blocks. For example, `echo "~20"` and heredoc bodies containing `~` were incorrectly flagged as forbidden paths.

## Fix (three-pronged)

1. **Tightened `looks_like_path`**: Changed from `candidate.starts_with('~')` to `candidate == "~" || candidate.starts_with("~/")`. Bare tilde-digit tokens like `~20` (common in git revision syntax) are no longer mistaken for paths.

2. **Quoted-token detection**: Before checking a token for tilde path violations, the raw token is inspected. If it was wrapped in matching quotes (`"..."` or `'...'`) and the stripped content starts with `~`, tilde-based path checks are skipped. This handles `echo "~20"`, `echo '~/foo'`, etc.

3. **Heredoc body tracking**: A delimiter stack tracks `<<DELIM` tokens within each segment. Body tokens inside heredoc blocks that start with `~` skip tilde-based path checks. Absolute/relative paths in heredoc bodies are still caught.

## Files changed
- `crates/zeroclaw-config/src/policy.rs`: +135/-5

## Verification
- All 726 unit tests pass (0 failures)
- `cargo fmt --check` clean
- 5 new tests covering: quoted tilde strings, tilde-digit without slash, tilde in heredoc body, absolute path in heredoc body still caught, tilde-user-with-slash still caught

### DIFF
--- a/crates/zeroclaw-config/src/policy.rs
+++ b/crates/zeroclaw-config/src/policy.rs
@@ -1647,12 +1647,11 @@ impl SecurityPolicy {
             // If the raw token was wrapped in matching quotes, its content
             // is a string literal — ~ inside quotes is NOT expanded by the
             // shell, so skip tilde-based path checks.
-            if (raw.starts_with('"') && raw.ends_with('"'))
-                || (raw.starts_with('\'') && raw.ends_with('\''))
+            if ((raw.starts_with('"') && raw.ends_with('"'))
+                || (raw.starts_with('\'') && raw.ends_with('\'')))
+                && candidate.starts_with('~')
             {
-                if candidate.starts_with('~') {
-                    return None;
-                }
+                return None;
             }
             if looks_like_path(candidate) && !self.is_path_allowed(candidate) {
                 Some(candidate.to_string())
--- a/crates/zeroclaw-config/src/policy.rs
+++ b/crates/zeroclaw-config/src/policy.rs
@@ -1073,7 +1073,8 @@ fn looks_like_path(candidate: &str) -> bool {
     candidate.starts_with('/')
         || candidate.starts_with("./")
         || candidate.starts_with("../")
-        || candidate.starts_with('~')
+        || candidate == "~"
+        || candidate.starts_with("~/")
         || candidate == "."
         || candidate == ".."
         || candidate.contains('/')
@@ -1643,6 +1644,16 @@ impl SecurityPolicy {
             if candidate.is_empty() || candidate.contains("://") {
                 return None;
             }
+            // If the raw token was wrapped in matching quotes, its content
+            // is a string literal — ~ inside quotes is NOT expanded by the
+            // shell, so skip tilde-based path checks.
+            if (raw.starts_with('"') && raw.ends_with('"'))
+                || (raw.starts_with('\'') && raw.ends_with('\''))
+            {
+                if candidate.starts_with('~') {
+                    return None;
+                }
+            }
             if looks_like_path(candidate) && !self.is_path_allowed(candidate) {
                 Some(candidate.to_string())
             } else {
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
Reviewed the current head, the PR body, the policy parser diff, and the green CI run.

### 🟢 Praise — The tilde classification is moving in the right direction

The tighter `~` handling is a good improvement, especially preserving `~<non-slash>` as data while still keeping `~user/...` covered when it has a slash.

### 🔴 Blocking — Spaced heredocs and multi-word quoted strings still reach the path scanner

The new exclusions only recognize heredoc openers when the operator and delimiter are in the same whitespace token, such as `<<EOF`, and only recognize quoted tilde content when the whole whitespace token is wrapped in matching quotes. That leaves common shell forms outside the fix:

```sh
cat << EOF
~/foo
EOF

cat << 'EOF'
~/foo
EOF

echo "about ~/foo here"
printf 'about ~/foo here'
```

In those forms, `forbidden_path_argument` still sees `~/foo` as an ordinary path token instead of heredoc or quoted data. That leaves common cases uncovered by the PR body's quoted-string and heredoc-body false-positive claim. Please make the parser track `<<` followed by a separate delimiter token, and make the quoted-data handling cover multi-token quoted strings rather than only whole-token quoted arguments.

chengzhichao-xydt:
Closing as superseded — the same fix has already been merged upstream in #7281 (`fix(policy): stop path guard false-positives on heredoc bodies and non-path tildes`).

Commit: `b47a66f0`

Upstream/master already handles:
- Tilde-then-non-slash tokens not flagged as home paths
- Heredoc body content not treated as argv path arguments
- Paths after quoted heredoc-like text still correctly blocked

This PR's changes are fully covered by the upstream fix.

Issue #6984, opened 2026-05-28
bug(gateway): token rotation does not revoke existing bearer tokens


singlerider:
### Affected component

gateway/security:pairing

### Severity

S1 - workflow blocked

### Current behavior

Both documented token-rotation surfaces mint a new pairing code but leave previously issued bearer tokens valid:

- CLI/admin pairing-code path: `zeroclaw gateway get-paircode --new` ultimately calls `PairingState::generate_new_pairing_code`, which replaces the single-use pairing code under `self.pairing_code` and does not remove entries from `self.paired_tokens`.
- HTTP device rotation path: `POST /api/devices/{id}/token/rotate` is a thin wrapper around the same `generate_new_pairing_code` call. It does not remove the device's existing token hash from the pairing state, does not rewrite the persisted `gateway.paired_tokens` field, and returns a payload whose `message` field is `"Use this code to re-pair the device"`.

Net effect: an operator who believes they have rotated a leaked token has only issued a second valid pairing path. The old bearer token continues to authenticate gateway endpoints until the paired-token state is actually revoked from both memory and persisted config.

There is a nearby device-deletion path, `DELETE /api/devices/{id}`, but current source suggests it deletes the device-registry row only. Gateway authentication still checks `PairingState::paired_tokens`, so device deletion also needs to be verified or fixed as part of this work rather than treated as sufficient token revocation.

Originally reported by @nixosclaw while following up on the Canvas sandbox/token-theft fix in #6942. This issue tracks the separate token-revocation behavior so it can be fixed and tested directly.

### Expected behavior

Token rotation should invalidate the bearer token being rotated away from. Concretely:

- Device-scoped rotation should remove the old token hash from in-memory pairing state.
- The persisted `gateway.paired_tokens` field should be rewritten so a daemon restart does not resurrect a revoked token.
- Device-registry state should not silently coexist with stale bearer-token auth state. The corresponding device entry should be removed, marked stale, or rebound in a way that matches the new pairing flow.
- CLI/admin flows should distinguish “add another client” from “rotate after compromise.” If `get-paircode --new` is meant to add a client, the destructive revocation path should have a separate explicit command/flag. If it is meant to rotate, it should revoke existing tokens by default or require an explicit target/all choice.

The CLI help text and HTTP response message should describe what was revoked, not just that a new code was issued.

### Steps to reproduce

```bash
# 1. Pair a dashboard and capture token T1.
zeroclaw gateway get-paircode
[...]

PR #6942, opened 2026-05-26
fix(web): tighten Canvas iframe sandbox to prevent token theft via XSS (GHSA-f385-f6h2-3gqj)

nixosclaw:
## Summary
- **Base branch:** `master`
- **What changed and why:**
  - Removes `allow-same-origin` from the Canvas iframe sandbox. The prior combination `sandbox="allow-scripts allow-same-origin"` is documented by the HTML spec as equivalent to no sandbox (a sandboxed script can call `window.frameElement.removeAttribute("sandbox")` and reach `window.parent`). With same-origin access, scripts inside the frame could read the dashboard bearer from `localStorage` and replay authenticated API requests under the operator's identity.
  - Switches the iframe content delivery from `contentDocument.write()` to the `srcdoc` attribute. `srcdoc` doesn't require same-origin access from the parent, so removing `allow-same-origin` is now structurally possible.
  - Adds `<meta http-equiv="Content-Security-Policy" content="script-src 'none'; object-src 'none'">` to the svg / markdown / text frame branches, which have no legitimate scripting need.
  - Strips `<script>` elements and `on*` event-handler attributes from inbound SVG content as defense-in-depth. The sandbox change and the CSP meta tag are the primary protections; DOM-level stripping bounds the surface even if a future change re-introduces `allow-same-origin` on this branch.
  - Makes the `html` content type an explicit dispatch branch (rather than the implicit fallthrough default), and points unknown `content_type` values at an inert no-script document. The prior fallthrough rendered unknown content types as scriptable html — opposite of the deny-by-default intent of this PR. Future content types must be added as explicit branches above the fallback.
  - Replaces the `srcDoc={undefined}` transition for `eval` frames with an explicit inert no-script document. The prior path could leave the iframe holding the previous frame's `srcdoc`, which contradicted the deny-by-default goal during content-type transitions.
- **Scope boundary:** Iframe boundary on the Canvas dashboard surface only (`web/src/pages/Canvas.tsx`). Out of scope: bearer-token storage / transport / scope (the bearer remains in `localStorage`; this PR removes the iframe's ability to read it), the dashboard's broader authority model, application-level CSP.
- **Blast radius:** Dashboard Canvas page only. Agent-authored content that depended on reaching the dashboard origin from inside the iframe (`window.parent.*`, parent-`localStorage` reads, same-origin fetch) will break — by design, since that capability was the vulnerability. svg/markdown/text frames are now strictly inert; html frames retain scripting within the sandboxed iframe but cannot reach the parent.
[...]

PR #7243, opened 2026-06-05
fix(gateway): implement proper token revocation on rotation and device deletion

chenyunbo411:
Partial Fix: Token Rotation Does Not Revoke Existing Bearer Tokens (#6984)
Summary
This PR delivers a partial fix for the critical security vulnerability tracked in #6984. It addresses token revocation and state persistence for the HTTP device management endpoints. The CLI/admin pairing code rotation path (get-paircode --new) remains unaddressed and will be resolved in a follow-up PR.
The fixed HTTP flows now properly revoke existing bearer tokens, clean up device registry records and persist state changes. Previously, rotating tokens or deleting devices would leave old bearer tokens valid indefinitely even after operators performed security remediation steps.
Problem Statement (from #6984)
Affected surfaces:
PairingGuard::generate_new_pairing_code() - generates new code but leaves paired_tokens untouched
POST /api/devices/{id}/token/rotate - thin wrapper that doesn't revoke old token
DELETE /api/devices/{id} - deletes device registry row but doesn't clear token from pairing state
persist_pairing_tokens() - only called during initial pairing, not during rotation
Note: This PR only resolves issues for the two HTTP device endpoints above. The CLI/admin path zeroclaw gateway get-paircode --new is not fixed here and will be handled separately.
Impact: An operator who believes they have rotated a leaked token has only issued a second valid pairing path. The old bearer token continues to authenticate gateway endpoints until the paired-token state is actually revoked from both memory and persisted config.
Changes Made
Core Security - Token Revocation in crates/zeroclaw-config/src/pairing.rs
Added two methods to PairingGuard:
revoke_token(token_hash: &str) -> bool - revokes a single token
revoke_all_tokens() -> usize - revokes all tokens (returns count removed)
Device Registry Enhancement in crates/zeroclaw-gateway/src/api_pairing.rs
Added lookup methods to DeviceRegistry:
find_device_by_token_hash(token_hash: &str) -> Option<String> - find device ID by token
get_token_hash_by_device_id(device_id: &str) -> Option<String> - get token hash for device
These enable the rotation flow to map from device ID → token hash → revocation.
Fixed POST /api/devices/{id}/token/rotate
Token rotation now:
Looks up device's token hash from registry
Revokes token from PairingGuard::paired_tokens
Deletes device from DeviceRegistry
Persists updated gateway.paired_tokens to config.toml
Returns clear response with "persisted": true and accurate message
Response format:
json
{
  "device_id": "abc123",
  "pairing_code": "123456",
  "message": "Old token revoked and device removed; use this code to re-pair",
  "persisted": true
}
If persistence fails, returns "persisted": false with warning message.
Fixed DELETE /api/devices/{id}
Device deletion now:
Fetches token hash before deletion
Revokes token from pairing state
Deletes device from registry
Persists token state change
Returns confirmation with "persisted": true
Response format:
json
{
  "message": "Device and token revoked",
  "device_id": "abc123",
  "persisted": true
}
Structured Logging
All warning/error messages upgraded to use ::zeroclaw_log::record! macro with structured fields:
WARN: "rotate_token: token hash was not in paired_tokens" (with device_id, token_hash)
WARN: "rotate_token: device was not found in registry during revoke" (with device_id)
WARN: "revoke_device: token hash was not in paired_tokens" (with device_id, token_hash)
ERROR: "Failed to persist token rotation" / "Failed to persist device revocation" (with device_id, error)
Unit Tests in crates/zeroclaw-config/src/pairing.rs
Added 5 comprehensive unit tests for the new revocation helpers:
revoke_token_removes_single_token - verifies single token removal
revoke_token_returns_false_for_nonexistent_token - verifies idempotency
revoke_all_tokens_clears_all - vericates bulk revocation
revoked_token_cannot_be_used_for_auth - verifies token is fully disabled
Successful pair then authenticate test remains to ensure pairing basics.
Test Coverage Note: This PR currently includes only unit tests for core revocation logic. End-to-end gateway-level regression tests for HTTP endpoints will be added in a follow-up or subsequent iteration.
Testing
All tests passing:
plaintext
cargo test -p zeroclaw-config --lib -- -k pairing::tests  # 36 passed
cargo test -p zeroclaw-gateway --lib                     # 201 passed
cargo clippy -p zeroclaw-config -p zeroclaw-gateway     # 0 errors, 2 warnings
cargo fmt --all                                         # formatted
Behavior Changes
Before This Fix
POST /devices/{id}/token/rotate → returns new pairing code, old token still works
DELETE /devices/{id} → deletes device row only, old token still works
zeroclaw gateway get-paircode --new → generates code but doesn't revoke anything (unmodified in this PR)
After This Fix
POST /devices/{id}/token/rotate → revokes old token, deletes device, persists state
DELETE /devices/{id}/token/rotate → revokes token AND deletes device, persists state
Both HTTP operations return "persisted": true/false to indicate durability status
Tokens are immediately invalidated in memory and survive daemon restart
CLI/admin pairing code rotation (get-paircode --new) remains unchanged temporarily
Security Implications
✅ Leaked tokens can now be properly revoked via HTTP token rotation or device deletion
✅ State consistency: DeviceRegistry and PairedTokens stay synchronized for HTTP flows
✅ Durability: Revocations persist across daemon restarts for HTTP flows
✅ Auditability: Structured logs record all revocation attempts
⚠️ Important: Operators must ensure daemon has write access to config.toml for persistence
⚠️ Outstanding Risk: The CLI/admin get-paircode --new path is not yet fixed; old tokens remain valid after generating new pairing codes here. This will be addressed in a separate follow-up PR.
Backwards Compatibility
API responses gain optional "persisted" boolean field and updated "message" values. Existing clients will continue to work; they'll just get more informative messages.
Follow-up Work (Tracked Separately)
Fix handle_admin_paircode_new to revoke existing tokens and update persisted paired_tokens for the CLI/admin pairing code rotation path
Add gateway-level end-to-end regression tests for all affected endpoints
References
Original issue: bug(gateway): token rotation does not revoke existing bearer tokens #6984
Related: fix(web): tighten Canvas iframe sandbox to prevent token theft via XSS (GHSA-f385-f6h2-3gqj) #6942 (Canvas sandbox/token-theft fix)
Affected component: gateway/security:pairing
Severity: S1 - workflow blocked (security)
Risk Tier: High (security-critical, affects access control)
PR Size: M (focused fix, 223 lines added)
Reviewers: security team + gateway maintainers


Audacity88 (changes requested):
Reviewing at 7d7a69d522339f03e4962584b14d055857a5e2be. I checked the live PR state, #6984, the two-file diff, and the relevant pairing/admin paths in `crates/zeroclaw-config/src/pairing.rs`, `crates/zeroclaw-gateway/src/api_pairing.rs`, and `crates/zeroclaw-gateway/src/lib.rs`. CI is green.

### 🟢 What looks good — The HTTP device path is moving in the right direction

The device route changes are the right shape for the HTTP-managed device flow: look up the device token hash, revoke that hash from `PairingGuard`, remove the registry row, and persist the new paired-token set. Returning a `persisted` flag also gives callers a concrete signal when in-memory revocation succeeded but durability failed.

### 🔴 Blocking — The CLI/admin rotation path from #6984 still only mints another code

This fixes the authenticated HTTP device route, but it does not yet fix the other rotation surface that #6984 explicitly calls out. `handle_admin_paircode_new` still just calls `state.pairing.generate_new_pairing_code()` and returns “New pairing code generated — use this one-time code to pair” without revoking any existing token or persisting a reduced `gateway.paired_tokens` set.

That means `zeroclaw gateway get-paircode --new`, which uses the admin pair-code endpoint, still has the original incident-response problem from #6984: an operator can follow the documented rotation path, get a new pairing code, and still leave the old bearer token valid. Because the commit message says `Fixes #6984` and the PR body describes the old `get-paircode --new` behavior as part of the vulnerability, merging this as-is would land as the apparent #6984 fix while one of the issue's named repro paths still works.

Please either extend this PR to cover the CLI/admin path, or narrow the public scope so this is only a partial HTTP-device fix: update the body and commit message so they do not claim the full token-rotation issue is fixed, and leave the CLI/admin add-vs-rotate follow-up tracked explicitly.

### 🟡 Warning — The endpoint behavior needs direct regression coverage

The new `PairingGuard::revoke_token` / `revoke_all_tokens` helper tests are useful, but the visible diff does not add gateway-level coverage for the route behavior that regressed: `POST /api/devices/{id}/token/rotate`, `DELETE /api/devices/{id}`, persistence through `persist_pairing_tokens`, or the admin pair-code path. For this security fix, I would like at least one gateway-level regression proving that a token accepted before rotation/deletion is rejected afterward and is absent from the persisted config state. If the PR intentionally stays helper-only plus unit coverage, the body should make that validation gap explicit.

chenyunbo411:
Addressed the HTTP device token rotation and deletion flows as a partial fix for #6984. The CLI/admin pairing code path and gateway-level integration tests will be added in follow-up PRs.

Audacity88:
@chenyunbo411 Thanks for narrowing the PR scope to the HTTP device rotation/deletion path. That resolves the scope-claim part of my earlier review: this does not need to fix the CLI/admin pairing-code path in the same PR if we keep it framed as a partial HTTP-device fix.

Before re-review, this still needs a few things:

- Resolve the current merge conflicts against `master`.
- Add gateway-level regression coverage for the HTTP route behavior: a token accepted before rotation/deletion should be rejected afterward, and the persisted `gateway.paired_tokens` state should no longer contain the revoked token hash.
- Make the remaining CLI/admin `zeroclaw gateway get-paircode --new` risk visible as open work. #6984 is currently closed, but this PR explicitly leaves that path unfixed, so we should either reopen #6984 or file/link a focused follow-up before this partial fix merges.

Once those are handled, I can re-review the updated head.

singlerider (commented):
Reviewed at head `602fbf8` against current `master`. The bigger issue than @Audacity88's open blocker is that a more complete token-revocation fix has since landed on `master`, so most of this PR is now superseded and one helper would actually collide. Agreeing with @Audacity88's changes-requested and adding that finding.

### 🔴 Blocking — superseded by `master`; `revoke_token` would collide with a different signature

`master` already has the token-revocation primitives this PR adds, in a more complete form:

- `PairingGuard::revoke_token(token: &str)` — takes the **raw token** and hashes it internally — plus a separate `revoke_token_hash(token_hash: &str)`. This PR's `revoke_token(token_hash: &str)` reuses that exact name for a **hash** argument, so landing it produces a same-name / different-meaning collision with the existing method.
- `PairingGuard::revoke_all_tokens()` already exists on `master`, behaviorally identical (and better documented, with the "nuke everything on unknown compromise" rationale and the persist-to-config caveat).
- These already carry unit coverage on `master` (`revoke_token_hash_matches_revoke_token`, `revoke_all_tokens_invalidates_every_token`, the empty-set case).

The only genuinely-new code here is `DeviceRegistry::find_device_by_token_hash` / `get_token_hash_by_device_id`, which are reverse-lookup helpers for the device route — and the revocation primitives they were meant to feed already exist upstream.

### 🔴 Blocking — agree: the #6984 admin path was the gap, and `master` already closed it

@Audacity88 flagged that the CLI/admin rotation path (`handle_admin_paircode_new` / `get-paircode --new`) still only minted a code without revoking. On current `master` that path now calls `revoke_all_tokens()`, clears the device registry, and `persist_pairing_tokens(...)` — so the named #6984 repro path this PR claims to fix is already handled upstream. Merging this as the "#6984 fix" would be inaccurate on top of redundant.

### Recommendation

Rebase onto current `master` and reassess. I expect the revocation helpers and the admin-path fix to drop out entirely as already-present, leaving at most the two `DeviceRegistry` reverse-lookup methods — and only if a device route still needs them. If nothing material remains after the rebase, this should be closed as superseded. If the reverse-lookup helpers are still wanted, narrow the PR to just those with their own justification, drop the colliding `revoke_token(hash)` helper, and stop claiming `Fixes #6984`.

Leaving @Audacity88's changes-requested in force.

Audacity88:
@chenyunbo411 thanks for taking on the token-revocation path here.

Closing this as superseded by the current `master` implementation. The revocation primitives and the `get-paircode --new` admin-path fix have since landed upstream, and this branch now conflicts with that direction because its `revoke_token` helper uses the same name for a different token-hash signature.

If there is still a small device-route reverse-lookup helper that is useful after rebasing onto current `master`, please open a fresh narrow PR for only that remaining gap. The broader #6984 token-rotation fix no longer needs this branch.

Issue #7025, opened 2026-05-30
[Bug]: read_skill cannot load plugin-bundled skills the prompt advertises


metalmon:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

Plugin-bundled skills are advertised in the system prompt but cannot be loaded by the `read_skill` tool, so the model is told to call a tool that always fails for those skills.

The two code paths disagree about whether plugin skills exist:

- **Prompt build** uses `load_skills_with_config(workspace_dir, &Config)`, which (under `feature = "plugins-wasm"`) does `skills.extend(load_plugin_skills_from_config(config))`. So a skill bundled in a plugin (manifest `capabilities = ["tool", "skill"]`) **is listed** in the `## Available Skills` section, and in compact `prompt_injection_mode` the model is instructed to call `read_skill(<name>)` to get its body.
- **`read_skill`** resolves skills via `load_skills_with_open_skills_settings(workspace_dir, open_skills_enabled, open_skills_dir, allow_scripts)`, which only unions **workspace** skills (`<data_dir>/skills`) and **open-skills**. It never calls `load_plugin_skills_from_config`.

Net effect: the model correctly follows the prompt and calls `read_skill` for a plugin-bundled skill, and gets:

```
Unknown skill '<name>'. Available skills: <workspace/open-skills only>
```

### Expected behavior

`read_skill(<name>)` should resolve any skill that the system prompt advertised — including plugin-bundled skills — so the compact-mode `read_skill` contract holds for every skill the model can see.

### Steps to reproduce

```bash
# 1. Build/run with the plugin system enabled
#    (features include plugins-wasm), and [plugins].enabled = true
# 2. Install a WASM plugin whose manifest bundles a skill:
#      capabilities = ["tool", "skill"]
# 3. Set compact skill injection so bodies load on demand:
#      [skills] prompt_injection_mode = "compact"
# 4. Start the daemon. The plugin's skill IS listed under "## Available Skills",
#    with an instruction to call read_skill(<name>).
# 5. Trigger the agent to call read_skill with that plugin skill's name.
# 6. Observe: "Unknown skill '<name>'. Available skills: ..." (plugin skill missing).
```

### Impact

```
Affected users: anyone shipping a skill inside a WASM plugin and running compact prompt_injection_mode.
Frequency: always for plugin-bundled skills in compact mode.
[...]

Issue #7236, opened 2026-06-05
[Bug]: Incorrect workspace_dir passed to load_skills_for_agent prevents loading skills from data/skills/


jokewithme110:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

The `workspace_dir` field in `ChannelRuntimeContext` is incorrectly set to `config.data_dir` instead of the agent's workspace directory. This causes `load_skills_for_agent` to look for skills in `data/skills/` directory, which doesn't exist or is empty, resulting in skills not being loaded properly.

Specifically, in `crates/zeroclaw-channels/src/orchestrator/mod.rs`:
- Line 8594: `workspace_dir: Arc::new(config.data_dir.clone())` sets workspace_dir to data_dir
- The `refreshed_new_session_system_prompt` function uses this incorrect workspace_dir when calling skill loading functions

Skills are only loaded from skill_bundles (located in `<install>/shared/skills/<bundle_alias>/`) but not from:
- Workspace skills at `<install>/agents/<alias>/workspace/skills/`
- Any skills that might be placed in `data/skills/`

### Expected behavior

The `workspace_dir` should point to the correct agent workspace directory (`config.agent_workspace_dir(agent_alias)`) so that `load_skills_for_agent` can properly load skills from:
1. Agent's workspace skills directory: `<install>/agents/<alias>/workspace/skills/`
2. Open skills (if enabled)
3. Plugin skills (if enabled)
4. Skill bundles: `<install>/shared/skills/<bundle_alias>/`

Alternatively, if `data/skills/` is intended to be a valid skills location, the directory structure should be documented and the code should handle this case explicitly.

### Steps to reproduce

```bash
1. Place a skill in what should be the workspace skills directory
2. Start the daemon with an agent configuration
3. Observe that the skill is not loaded unless it's also placed in a skill_bundle directory
```

### Impact

**Affected users**: All users who place skills in agent workspace directories rather than skill_bundles
**Frequency**: Always - affects every session initialization
**Consequence**: 
- Skills placed in agent workspace directories are silently ignored
- Only skill_bundle skills are loaded, limiting flexibility
- Inconsistent behavior between initial agent setup (which may work correctly) and runtime skill refresh

### Logs / stack traces

```text
No crash occurs - this is a silent failure where skills are simply not loaded.
```

### ZeroClaw version

v0.8.1-beta

### Operating system

All platforms (cross-platform issue)

### Regression?

Unknown

[...]

PR #7245, opened 2026-06-05
fix(read_skill): fix read_skill cannot load plugin-bundled skills and bounded skills

jokewithme110:
> **Maintainer's note** (Audacity88): I pushed two maintainer commits to the contributor branch and refreshed this body so the public record matches the current scope, validation, and risk. Those commits replace the earlier caller-provided workspace-root approach with config-derived `read_skill` loading through `load_skills_for_agent_from_config`. The original author's commit `b33f0d4` and description remain in the branch history; authorship of the original change remains with @jokewithme110. This PR remains Related #7025.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Refactors compact-mode `ReadSkillTool` so the on-demand `read_skill` path resolves the same effective skill set advertised to the model.
  - Removes the duplicate caller-provided workspace root from `ReadSkillTool`; skill lookup now derives the per-agent workspace and configured skill sources from `Config`.
  - Keeps the final PR-shaped diff limited to `crates/zeroclaw-runtime/src/tools/mod.rs` and `crates/zeroclaw-runtime/src/tools/read_skill.rs`.
  - Adds regression coverage for workspace/bundle precedence and feature-gated plugin-bundled reads by namespaced skill name.
- **Scope boundary:** This PR does not change plugin loading infrastructure, skill bundle config syntax, prompt rendering, or non-`read_skill` tool behavior.
- **Blast radius:** High-risk runtime/tool path. This affects compact skills mode when the model calls `read_skill` for skills from workspace, open-skills, configured bundles, or plugins.
- **Linked issue(s):** Related #7025.
- **Labels:** `bug`, `risk: high`, `size: S`, `runtime`.

```text
cargo fmt -p zeroclaw-runtime
Result: passed.

cargo clippy -p zeroclaw-runtime --all-targets --features plugins-wasm -- -D warnings
Result: passed.

cargo test -p zeroclaw-runtime read_skill
Result: passed, 9 tests.

cargo test -p zeroclaw-runtime --features plugins-wasm read_skill
Result: passed, 10 tests.

GitHub CI at head 2f26f4f0e53b516e143a8dd0a8ee356adbc504c0:
Result: all visible checks passing, including Format, Lint, Test, Security, platform checks, and CI Required Gate.
```

- **Commands run and tail output:** See command/result summary above.
- **Beyond CI - what did you manually verify?**
  - `read_skill` now derives effective skills from the same config-backed per-agent loader used for advertised compact-mode skills.
  - The config-backed loader preserves workspace-over-bundle precedence, pinned by the focused regression coverage.
  - A feature-gated plugin skill fixture can be read by its advertised `plugin:<plugin>/<skill>` name.
- **If any command was intentionally skipped, why:** No focused `read_skill` validation was skipped. No separate local broad workspace nextest run was done; the same final head is covered by green GitHub CI, and the local validation targets the changed runtime/tool surface directly.

- New permissions, capabilities, or file system access scope? `Yes, narrowly` - `read_skill` can now load configured workspace, open-skills, bundle, and plugin skill files that were already part of the advertised skill set. This does not add arbitrary path access.
- If any `Yes`, describe the risk and mitigation: The change aligns the read path with the advertised skill set. It does not add network access, secret handling, or a new config surface; rollback is a normal revert if compact-mode skill reads regress.

## Compatibility (required)

- Backward compatible? `Yes`

- **Fast rollback command/path:** Revert the merge commit for this PR.
- **Feature flags or config toggles:** None added. As an operational mitigation, avoid compact skills mode or disable plugin skills if a regression is isolated there.
- **Observable failure symptoms:** Compact-mode agents may advertise a skill but fail to load it through `read_skill`, usually as an unknown-skill result or a model turn that cannot retrieve the skill body.

N/A


jokewithme110:
@Audacity88 @singlerider please review this pr

Audacity88 (changes requested):
Reviewed PR #7245 at 96b79c5 against linked issues #7025 and #7236, the `tools/mod.rs` / `tools/read_skill.rs` diff, the config workspace contract, and the current skill loader contract in `skills/mod.rs`. Reusing the canonical per-agent loader is the right direction, but I do not think this is ready yet: the runtime wiring still passes the shared data directory where the loader expects an agent workspace directory, and the PR does not yet pin the plugin-bundled regression that #7025 is about.

### 🟢 What looks good — `read_skill` is moving toward the canonical loader

The old implementation duplicated only part of skill discovery by calling `load_skills_with_open_skills_settings(...)` from inside the tool. Moving `ReadSkillTool` toward `load_skills_for_agent(...)` is the right shape because the prompt and the on-demand tool should share one definition of "available skills." Once the path wiring below is fixed, this should be a cleaner contract than adding another plugin-specific branch inside `read_skill`.

### 🔴 Blocking — `read_skill` still passes `data_dir` as the workspace directory

`load_skills_for_agent(workspace_dir, config, agent_alias)` first calls `load_skills_with_config(workspace_dir, config)`, so its first argument controls the workspace skill root. In the runtime constructor, this PR wires the tool as:

```rust
ReadSkillTool::new(
    config.clone(),
    root_config.data_dir.clone(),
    agent_alias.to_string(),
)
```

That keeps the same mismatch described in #7236. `Config::data_dir` is shared runtime state for databases/cost/session files; the config schema documents per-agent plaintext state and workspace skills under `Config::agent_workspace_dir(alias)`. As written, an agent skill placed under `<install>/agents/<alias>/workspace/skills/...` is still not readable through the real `read_skill` tool, even though the new `workspace_skill_overrides_bundle_skill` test passes because it constructs the tool directly with `tmp/workspace`.

Please wire `ReadSkillTool` with the same agent workspace root that the prompt/loader contract expects, likely `root_config.agent_workspace_dir(agent_alias)` unless there is a narrower canonical workspace argument already available in this registry path. Then add a registry-level or constructor-level regression test that builds the tool the way `all_tools_with_runtime()` does and proves a skill under the agent workspace can be read.

### 🔴 Blocking — the plugin-bundled skill path is still untested

#7025 is specifically about plugin-bundled skills being advertised but not readable in compact mode. #7236 is the related workspace-directory concern. This patch probably reaches plugin skills when built with `plugins-wasm`, because `load_skills_for_agent(...)` delegates to `load_skills_with_config(...)`, which extends from `load_plugin_skills_from_config(config)`. But the tests added here only cover workspace skills, script-bearing skills, and configured skill bundles; none of them exercise a plugin skill name or the `#[cfg(feature = "plugins-wasm")]` branch.

Please add a focused regression for the actual #7025 contract: a skill loaded from plugin configuration is listed by the canonical loader and can be read through `ReadSkillTool` using the same name the prompt advertises. If a full plugin-host fixture is too heavy, a smaller feature-gated loader/tool test is fine, but the test should fail on the old `read_skill` path and pass because this tool now uses the full config-backed loader.

### 🟡 Warning — PR body metadata should be refreshed after the fix

The PR body currently says `type: enhancement`, `risk: low`, and `size: M`, while the live labels are `bug`, `risk: high`, and `size: S`, and the linked issue line says `#7236` while linking to #7025. Those are maintainer-fixable metadata issues, not the reason for this request-changes review, but they should be cleaned up when the branch is revised so the public record matches the actual scope and risk.

jokewithme110:
@Audacity88 
Thanks for the detailed review! 

I'd like to clarify the logic here: passing `data_dir` into `read_skill` and calling `load_skills_with_config` actually *does* successfully load global skills, agent-specific skills, and plugin skills. I suspect the AI review might have missed some of the underlying implementation details here.

Additionally, I have already done extensive local testing after fixing this bug, and everything works as expected. 

Could you please take another look at this? If there are still specific edge cases you are concerned about, I'd be happy to discuss them further.

Audacity88:
@jokewithme110 thanks for the follow-up. I took another look at current head `96b79c5` and the current loader contract.

I agree with one part of your clarification: after this patch, `read_skill` does call `load_skills_for_agent(...)`, so plugin skills can be included through the config-backed loader. The `data_dir` argument does not itself block the plugin branch, because plugin skills come from `config` rather than from that path.

The remaining blocker is narrower: the first argument to `load_skills_for_agent(...)` is still the workspace root for workspace skills. Current master’s channel prompt path now computes `config.agent_workspace_dir(agent_alias)` and passes that workspace into `load_skills_for_agent(...)`. This PR’s `read_skill` registration still constructs the tool with `root_config.data_dir.clone()`, so `read_skill` can still disagree with the prompt for skills under `<install>/agents/<alias>/workspace/skills`.

That is also why the new direct `ReadSkillTool` tests do not settle the issue: they construct the tool with `tmp/workspace`, while the production `all_tools_with_runtime()` path constructs it with `root_config.data_dir`.

So the requested fix remains: wire `ReadSkillTool` with the same agent workspace root that `all_tools_with_runtime()` already receives, or with `config.agent_workspace_dir(agent_alias)`, and add a registry/constructor-level regression that builds the tool the way production does. The plugin-specific regression is still useful for #7025, but I agree the implementation direction is the right one now; the missing piece is pinning that the advertised skill and `read_skill` use the same source set in the real runtime wiring.

jokewithme110:
@Audacity88 please review it again, already fix it by your suggestion.

Audacity88 (changes requested):
Reviewed current head c657c402 against the prior request-changes review, linked issues #7025 and #7236, the current diff, and the public runtime/skill-loader source context. This revision fixes part of the `ReadSkillTool` wiring, but I still do not think the PR is ready: the main runtime registrations still pass `config.data_dir` as the workspace root, and the plugin-bundled skill path from #7025 still does not have a regression.

### 🟢 What looks good — `ReadSkillTool` now follows the registry workspace argument

The constructor change in `tools/mod.rs` is the right direction. `ReadSkillTool` now receives the full config plus the `workspace_dir` argument from `all_tools_with_runtime()`, so the tool itself can follow the canonical loader contract when the registry is called with the correct agent workspace root.

### 🔴 Blocking — runtime `read_skill` registration still receives `data_dir` in direct agent paths

The remaining problem is that the direct runtime paths still call `all_tools_with_runtime()` with `&config.data_dir`.

On current head, both `crates/zeroclaw-runtime/src/agent/loop_.rs` registrations still pass `&config.data_dir` as the workspace argument. Because `ReadSkillTool::execute()` now calls `load_skills_for_agent(&self.workspace_dir, &self.config, &self.agent_alias)`, those paths still ask the canonical loader to look under the shared data directory rather than `<install>/agents/<alias>/workspace`. That is the #7236 failure shape, just one layer higher than the constructor now.

Please wire these runtime registrations to the per-agent workspace root, either by using `config.agent_workspace_dir(agent_alias)` at those call sites or by using the existing canonical helper so the prompt and `read_skill` cannot drift. A regression should build/register the compact-mode tool through the production registry path and prove a skill placed only under the agent workspace is readable.

### 🔴 Blocking — the #7025 plugin-bundled skill contract is still unpinned

#7025 is specifically about plugin-bundled skills being advertised but not readable through compact-mode `read_skill`. The new tests cover workspace skills, script-bearing skills, and agent skill bundles, which is useful coverage, but none of them exercise a plugin-bundled skill name or the `plugins-wasm` plugin loader path.

Please add a focused regression for the actual plugin contract: a plugin-provided skill that the canonical loader would advertise should be readable through `ReadSkillTool` by that advertised name. If a full plugin fixture is too heavy, a smaller feature-gated loader/tool test is fine; the important part is that the test would have failed with the old partial loader and now passes because `read_skill` uses the full config-backed loader.

### 🟡 Warning — PR body metadata is still stale

The PR body still says `type: enhancement`, `risk: low`, and `size: M`, while the live PR labels are `bug`, `risk: high`, and `size: S`; it also has a `#7236` line whose link points to #7025. That metadata cleanup is maintainer-fixable and not the main blocker, but it should be refreshed before this merges so the public record matches the actual scope and risk.

singlerider (commented):
Reviewed at head `c657c40` against current `master`. The constructor change is the right direction, but I verified @Audacity88's core blocker — the registrations still feed the wrong workspace root — so the #7236 failure shape persists. Agreeing with his changes-requested.

### 🟢 `ReadSkillTool` now follows the registry workspace contract

Threading the full config plus `workspace_dir` into `ReadSkillTool` so `execute()` calls `load_skills_for_agent(&self.workspace_dir, &self.config, &self.agent_alias)` is correct — the tool now follows the canonical loader contract instead of a partial loader.

### 🔴 Blocking — confirmed: runtime registrations still pass `config.data_dir`, not the agent workspace

Both `all_tools_with_runtime(...)` call sites in `crates/zeroclaw-runtime/src/agent/loop_.rs` (lines 907 and 2403) still pass `&config.data_dir` as the workspace argument. Since `ReadSkillTool::execute()` now hands that straight to `load_skills_for_agent`, the canonical loader is pointed at the shared data directory rather than `<install>/agents/<alias>/workspace` — exactly the #7236 failure, just moved up from the constructor to the registration. The canonical per-agent helper already exists: `config.agent_workspace_dir(agent_alias)` (`schema.rs:3590`). Wire both registrations through that (or the existing canonical helper used for the prompt) so `read_skill` and the prompt cannot drift apart. Add a regression that registers the tool through the production path and proves a skill placed only under the agent workspace is readable.

### 🔴 Blocking — #7025 plugin-bundled skill contract still unpinned

The new tests cover workspace skills, script-bearing skills, and agent bundles, but none exercise a plugin-bundled skill name or the `plugins-wasm` loader path, which is the specific #7025 failure (advertised but not readable via compact-mode `read_skill`). Add a focused regression — a plugin-provided skill the canonical loader advertises must be readable through `ReadSkillTool` by that advertised name. A small feature-gated loader/tool test is fine if a full plugin fixture is too heavy; the point is it would have failed with the old partial loader and passes now.

### 🟡 Stale body metadata

The body still says `type: enhancement` / `risk: low` / `size: M` while the live labels are `bug` / `risk: high` / `size: S`, and a `#7236` line links to #7025. Refresh before merge.

The constructor fix is right; it just needs the registration call sites to pass the per-agent workspace root and the plugin regression. Leaving @Audacity88's changes-requested in force.

jokewithme110:
@Audacity88 @singlerider 
Thanks for your detailed review comments! I’ve resolved both blocking issues mentioned:
Updated the two all_tools_with_runtime call sites to pass agent_workspace_dir instead of shared data dir to fix the #7236 workspace root error.
Added targeted regression tests covering plugin-bundled skill flow to address #7025.
Also refreshed the outdated PR body labels and wrong issue link.
Kindly take another look, let me know if anything else needs adjustment.

Audacity88 (changes requested):
Reviewed current head b33f0d4 after @jokewithme110's latest update. I checked the prior Audacity88 request-changes review, @singlerider's follow-up review, #7025, the related #7236 workspace-root history, the current diff, visible green CI, the PR body's validation notes, and the runtime skill-loader/plugin-skill source. Thanks for the update. The workspace-root registration issue from my last review is resolved. The PR still needs one more change: the #7025 plugin-bundled compact-mode path is not pinned by a `read_skill` regression.

### ✅ Resolved — direct runtime registrations now pass the agent workspace root

The workspace-root blocker from the last review is addressed in the direct runtime paths. Both `run(...)` and `process_message(...)` now compute `config.agent_workspace_dir(agent_alias)` and pass that path into `all_tools_with_runtime(...)`, so `ReadSkillTool` receives the same per-agent workspace root that the canonical loader expects. That fixes the direct `config.data_dir` registration problem.

### 🟢 What looks good — `ReadSkillTool` now shares the canonical loader

The tool-level direction is right. `ReadSkillTool` now holds the full `Config`, the agent alias, and the workspace directory, then calls `load_skills_for_agent(...)` instead of keeping its own partial open-skills/workspace loader path. That is the single-source-of-truth shape this bug wants: the prompt and the on-demand `read_skill` tool should not have separate ideas of which skills exist. The new bundle and workspace-override tests are useful coverage for that broader loader contract.

### 🔴 Blocking — #7025's plugin-bundled `read_skill` path still has no regression

#7025 is specifically about a skill shipped by a WASM plugin being advertised in compact mode and then failing when the model calls `read_skill` for that advertised skill. The important behavior is not only "agent skill bundles can be read"; it is "a plugin skill loaded through the `plugins-wasm` plugin loader, with its advertised namespaced skill name, can be read through `ReadSkillTool`."

Your latest comment says targeted regression tests now cover the plugin-bundled skill flow for #7025, but I do not see that specific path in the current diff. `read_skill.rs` now covers workspace skills, script-bearing skills, configured agent skill bundles, and workspace-over-bundle precedence. Existing plugin-host tests cover skill-plugin discovery. None of those tests enable plugin skills and call `ReadSkillTool` with a plugin-provided skill name, such as the `plugin:<plugin>/<skill>` name produced by the loader. That means the exact #7025 contract could drift again while these tests still pass.

Please add a focused regression for the actual compact-mode plugin contract. A small `plugins-wasm` test is fine: create or reuse a skill-capable plugin fixture, enable plugins in `Config`, construct the compact-mode `read_skill` tool through the production registry or directly with equivalent config, call it with the advertised plugin skill name, and assert that it returns the plugin skill body. That test should fail on the old partial-loader implementation and pass because this PR now routes `read_skill` through the full config-backed loader.

### 🟡 Warning — PR body evidence and labels still need a final refresh

The body is closer, but it still says `risk: low` while the live label is `risk: high`. The validation section also lists the full required command set in a fenced block, while the recorded command results only mention `cargo check -p zeroclaw-runtime` and focused `cargo test -p zeroclaw-runtime read_skill`. Before merge, please make the public body match the final labels and the validation that actually ran, especially because this is a high-risk runtime/tool fix.

Nillth (commented):
I reviewed PR #7245 at head b33f0d4a35f07b9c59727d16c77dd8a37571bcfd against the canonical skill-loader contract in crates/zeroclaw-runtime/src/skills/mod.rs (load_skills_for_agent / load_skills_for_agent_from_config and the #[cfg(feature = "plugins-wasm")] plugin extension), against the two issues this work targets — #7236 (workspace-root mismatch) and #7025 (plugin-bundled skills advertised in compact mode but unreadable) — and against the three formal review rounds already on this PR. Taking stock first: the workspace-root wiring that @Audacity88 blocked on in the 2026-06-06 and 2026-06-13 rounds, and that @singlerider independently confirmed on 2026-06-14, is now genuinely fixed at this head — both runtime registrations in loop_.rs compute config.agent_workspace_dir(agent_alias) and thread it through. That blocker is resolved. The one finding still live from @Audacity88's 2026-06-15 round — the missing #7025 plugin-bundled regression test — is real and unaddressed in this diff, and I confirmed it against the head blob. Because @Audacity88's CHANGES_REQUESTED is still active and the diff does not yet clear it, I am posting this as a comment rather than approving; per our review protocol I will not approve over another reviewer's active block, and the one substantive item I would otherwise want is the same item @Audacity88 already holds, so this is corroboration plus a ready-to-paste test, not a competing verdict.

### 🟢 What looks good — ReadSkillTool collapsed onto the single-source-of-truth loader

This is the right shape for the fix. read_skill.rs now stores (config: Arc<Config>, workspace_dir, agent_alias) and its execute() calls crate::skills::load_skills_for_agent(&self.workspace_dir, &self.config, &self.agent_alias) instead of the old load_skills_with_open_skills_settings. That matters because the old loader returned only workspace + open-skills, while load_skills_for_agent unions workspace + open-skills + plugin skills + the agent's resolved skill_bundles, first-write-wins so workspace overrides bundle (skills/mod.rs first-write loop). The system prompt's <available_skills> is built from that same loader family (load_skills_for_agent_from_config), so collapsing read_skill onto it is precisely what closes the advertise-but-cannot-read divergence at the root of #7025/#7236 rather than papering over one symptom. The workspace_skill_overrides_bundle_skill test pins the precedence direction (asserts the workspace version present, the bundle version absent), which is the behavior most likely to silently regress later. Naming the principle so it carries past this PR: the read path and the advertise path must draw from one loader, never two that can drift.

### ✅ Resolved — workspace-root wiring (the #7236-shape blocker)

The blocker @Audacity88 raised on 2026-06-06 and 2026-06-13, and that @singlerider independently reproduced on 2026-06-14 (loop_.rs registrations still passing &config.data_dir as the workspace root), is addressed at this head. Both direct runtime registration sites — run() and process_message() — now bind let agent_workspace_dir = config.agent_workspace_dir(agent_alias); and pass &agent_workspace_dir into all_tools_with_runtime, replacing &config.data_dir. I verified the two further production call sites the PR does not touch are already correct and so are fixed transitively by the constructor rewiring: agent.rs (from_config / WebSocket / daemon / ACP) passes &security.workspace_dir, and the channels orchestrator passes the per-agent &workspace. The data_dir-vs-agent-workspace desync that made workspace skills under agents/<alias>/workspace/skills unreadable is gone on the runtime paths. Acknowledging this so the work registers — it took three rounds and it landed correctly.

### 🔴 Blocking — no plugins-wasm regression test for the #7025 plugin-bundled path

This is the one finding from @Audacity88's 2026-06-15 review that is still live, and I confirmed it directly against the head blob, so I want to corroborate rather than relitigate. The diff adds reads_skill_from_agent_bundle and workspace_skill_overrides_bundle_skill, but the complete test set at this head is those two plus the four pre-existing tests — there is no #[cfg(feature = "plugins-wasm")] test and no load_plugin_skills_from_config reference anywhere outside doc comments. The plugin source is the one input that only compiles and loads under the plugins-wasm feature (skills/mod.rs gates load_plugin_skills_from_config and the skills.extend(...) call site to that cfg), so it is the single path most worth a regression test and the only one the new tests structurally cannot reach: every new test builds Config::default(), which has plugins.enabled = false, so the plugin branch is never entered — and under the default feature set the plugin code is not even compiled, which is why the validation run of cargo test -p zeroclaw-runtime read_skill cannot exercise it. The wiring itself is sound (load_skills_for_agent → load_skills_with_config → #[cfg(plugins-wasm)] skills.extend(load_plugin_skills_from_config(config))), so this is a coverage gap on the headline #7025 contract, not a correctness defect. The principle: when a fix's headline case lives behind a feature flag, the regression must run under that flag, or the contract that justified the PR has zero executed coverage.

I am not asking you to invent it. I wrote and ran this test in my tree and it passes 9/9 under cargo test -p zeroclaw-runtime --features plugins-wasm read_skill. The one gotcha is that the plugin SKILL.md must carry YAML frontmatter (name + description) or the loader will not register it. Paste it into the read_skill.rs test module:

```rust
#[cfg(feature = "plugins-wasm")]
#[tokio::test]
async fn reads_plugin_bundled_skill_by_namespaced_name() {
    use std::sync::Arc;
    let tmp = tempfile::TempDir::new().unwrap();
    let workspace = tmp.path().join("workspace");
    std::fs::create_dir_all(&workspace).unwrap();
    let plugins_dir = tmp.path().join("plugins");
    let plugin_dir = plugins_dir.join("weatherkit");
    std::fs::create_dir_all(plugin_dir.join("skills").join("forecast")).unwrap();
    std::fs::write(plugin_dir.join("manifest.toml"),
        "name = \"weatherkit\"\nversion = \"0.1.0\"\ncapabilities = [\"skill\"]\n").unwrap();
    // SKILL.md MUST carry YAML frontmatter with name + description.
    std::fs::write(plugin_dir.join("skills").join("forecast").join("SKILL.md"),
        "---\nname: forecast\ndescription: Return a weather forecast for a place.\n---\n\n# Forecast\n").unwrap();
    let mut config = zeroclaw_config::schema::Config::default();
    config.plugins.enabled = true;
    config.plugins.plugins_dir = plugins_dir.to_string_lossy().into_owned();
    let tool = ReadSkillTool::new(Arc::new(config), workspace, "agent".to_string());
    let result = tool.execute(serde_json::json!({ "name": "plugin:weatherkit/forecast" })).await.unwrap();
    assert!(result.success, "advertised plugin skill must be readable; got {:?}", result.error);
    assert!(result.output.contains("# Forecast"));
}
```

For what it is worth, I hit the underlying bug live and independently: a skill_bundle-assigned slash skill that was registered and listed in compact mode failed with "Unknown skill" on invoke. So the fix direction here is correct — this PR addresses a real, reproducible class of bug. The block is narrowly the missing executed proof on the plugin leg, and the snippet above closes it.

### 🟡 Warning — read_skill correctness is re-coupled to every caller passing the right workspace_dir, and the gateway site does not

The design at this head makes read_skill correct only when every all_tools_with_runtime caller hands it config.agent_workspace_dir(agent_alias) as workspace_dir, because load_skills_for_agent feeds that caller-supplied path straight into load_workspace_skills. The two loop_.rs sites and the agent.rs / channels sites satisfy that. But the gateway registration in crates/zeroclaw-gateway/src/lib.rs passes &config.data_dir as workspace_dir, not the per-agent workspace — so read_skill there loads workspace skills from data_dir while the per-agent prompt builds <available_skills> from the agent workspace: the exact #7236-class desync this PR fixes at the loop_ sites, still present at the gateway. That registry drives the legacy /api/tools listing and a test mock and is not load-bearing for per-request agent dispatch, which is why I am marking this a warning and not a second block — but the coupling is real and realized on this head. Worth a fast-follow rather than a change to this PR: pass the per-agent workspace at the gateway site too, or — more durably — have ReadSkillTool derive the workspace via load_skills_for_agent_from_config(&config, &agent_alias) instead of taking workspace_dir as a parameter. That helper derives the workspace from config internally and cannot desync regardless of caller, which retires this whole class of caller-coupling in one move. Either way I would land #7245 as-is and handle the gateway site in a follow-up; I am not blocking on it.

### 🔵 Suggestion — ACP custom-cwd edge can still desync the workspace-skill layer (follow-up, not this PR)

One narrow edge survives and is worth a tracked follow-up rather than a change here. On the agent.rs from_config path, when ACP supplies a session_cwd, security.workspace_dir is overwritten to that cwd before all_tools_with_runtime is called, so read_skill scans <session_cwd>/skills while the prompt's <available_skills> still lists workspace skills from config.agent_workspace_dir(agent_alias)/skills. Only the locally-authored workspace-skill component can desync — bundles, open-skills, and plugin skills all resolve from config in both loaders and stay identical — and it is confined to an ACP-with-custom-cwd session. Critically this is not introduced or worsened by this PR: master hardcoded data_dir and was unconditionally desynced for every per-agent session with no bundle or plugin awareness, so this is a strict improvement. I would land #7245 as-is on this point and open a follow-up to have the from_config path pass the config-derived agent workspace (or have read_skill derive it via load_skills_for_agent_from_config) — which the config-derived constructor in the warning above would also resolve in one stroke.

### 🟡 Warning — body metadata still understates risk and over-claims the validation run

@Audacity88 flagged this across rounds and it is still partly stale at this head, so I am keeping it on the record. Two real mismatches remain. First, risk: the PR body's Labels line and Rollback section say risk: low and "git revert is sufficient," but the live label is risk: high (auto-applied for the security/runtime/tools scope), and a risk:high change wants more than the low-risk one-line rollback note. Second, validation evidence: the body advertises cargo fmt --all -- --check, cargo clippy --all-targets -- -D warnings, and a full cargo test -p zeroclaw-runtime and asserts "None skipped," but the recorded "Commands run and tail output" only shows cargo check -p zeroclaw-runtime and a focused cargo test -p zeroclaw-runtime read_skill (7 tests) — fmt, clippy, and the full suite are listed but not evidenced as run. CI is green on all 17 checks (fmt, clippy, test included), so the gates themselves pass; the issue is that the body's evidence section claims runs it does not show. Please bring the body's risk line and rollback note in line with risk: high and either paste the actual fmt/clippy/full-test tails or drop the "None skipped" claim to match what was recorded. The type: bug and size: S lines now match the live labels — the earlier type: enhancement / size: M text has been corrected, so that part is settled and I am not re-raising it. To be clear, this is a body-accuracy item; I am not holding merge on labels — labels are maintainer metadata and I will not block a contributor over a field they may not be able to edit.

Net: the wiring fix is correct and well-shaped, the workspace-root blocker is genuinely resolved, and the only thing standing between this and a clean read is an executed plugins-wasm regression — for which I have handed you a passing test to paste. Once that lands and @Audacity88's block is cleared, this looks mergeable to me.

Audacity88 (approved):
Reviewed current head `fd8eb52ec67b0443a98748076bb1d70065493346` after the maintainer commits I pushed on top of @jokewithme110's original branch. I checked the live PR state, linked issue #7025, my prior request-changes reviews, @singlerider's confirmation on the workspace-root blocker, @Nillth's follow-up on the plugin-regression gap, the final two-file diff, the refreshed PR body, focused local validation, and green GitHub checks. Those maintainer commits replace the earlier caller-provided workspace-root approach with config-derived `read_skill` loading and add the plugin-bundled regression. The prior blockers are resolved, so I am approving this head.

### ✅ Resolved — `read_skill` now derives the effective skill set from config

The final implementation removes the caller-provided workspace root from `ReadSkillTool` and resolves skills through `load_skills_for_agent_from_config(&self.config, &self.agent_alias)`. That is the right single-source-of-truth shape for this bug: compact-mode prompt advertising and on-demand `read_skill` lookup now draw from the same config-backed per-agent skill set instead of separate partial loaders.

### ✅ Resolved — the #7025 plugin-bundled `read_skill` path is covered

The feature-gated regression now creates a plugin skill fixture and verifies that `read_skill` can read it by the advertised namespaced skill name, `plugin:weatherkit/forecast`. That directly covers the missing #7025 path from my last review.

### 🟢 What looks good — validation and metadata support approval

The focused `read_skill` validation passed both without and with `plugins-wasm`, including the plugin-bundled regression. The PR body now reflects the high-risk runtime/tool scope, current labels, rollback plan, and validation evidence, and all visible GitHub checks are green.

Nillth (approved):
Reviewed at head `fd8eb52` against `upstream/master`. My earlier follow-up here flagged the plugin-bundled `read_skill` regression gap; this confirms it is closed. Approving.

### Single source of truth, verified

The fix removes the caller-supplied workspace snapshot from `ReadSkillTool` and resolves through `load_skills_for_agent_from_config(&self.config, &self.agent_alias)`. That is the exact same call the compact-mode prompt-advertise path uses (`agent/loop_.rs` in both the boot and process_message paths), with the same `config`/`agent_alias` pair that flows into `all_tools_with_runtime`. The two paths now derive from one loader and can no longer drift, which is the right shape for this bug.

### Plugin path genuinely covered

`load_skills_for_agent_from_config -> load_skills_for_agent -> load_skills_with_config`, which under `feature = "plugins-wasm"` extends with `load_plugin_skills_from_config`. Plugin skills are namespaced `plugin:<plugin>/<skill>` via `namespace_plugin_skill`, so the feature-gated `reads_plugin_bundled_skill_by_namespaced_name` test reads `plugin:weatherkit/forecast` correctly. It is a true regression guard: pre-fix the workspace-only loader returned Unknown for that name. This directly covers the missing #7025 path I called out.

### Bundle ("bounded") skills too

`reads_skill_from_agent_bundle` covers agent-bound bundle resolution, and `workspace_skill_overrides_bundle_skill` pins workspace-over-bundle precedence, matching the loader's first-write-wins behavior. No stale `ReadSkillTool::new` call sites remain; `config.clone()` is an Arc refcount bump, not a deep clone.

### Verdict

Approve. Correct, minimal, well-tested fix that eliminates the dual-loader drift at the root rather than patching a single symptom.

Two optional nits, neither blocking:
1. `reads_skill_from_agent_bundle` and `workspace_skill_overrides_bundle_skill` each re-declare `use tempfile::TempDir;` although it is already imported at module scope. Cosmetic.
2. `read_skill` now recomputes the full four-source union (including plugin-host enumeration) on every call. Fine by me: it is an on-demand path backed by the skills cache and mirrors what the prompt build already does per message, so it is by design per #7025's option 1.

PR #4254, opened 2026-03-22
feat: support forwarded messages in Telegram channel (#4118)

Tenith01:
Closes #4118

## Description
This PR adds support for Telegram's `forward_origin` and `forward_date` fields so that forwarded messages are properly attributed instead of appearing as standard user input.

## Changes
* Adds an `extract_forward_info` helper to `src/channels/telegram.rs`.
* Prepends a human-readable attribution string (e.g., `[Forwarded from @username]`) to the message content.
* Handles all four `forward_origin` variants gracefully: `"user"`, `"hidden_user"`, `"chat"`, and `"channel"`.
* Extends message parsing for standard updates, as well as voice and attachment messages where applicable.
* Includes ~7 unit tests covering origin extraction and forwarded message parsing.

PR #7251, opened 2026-06-05
fix(channels/telegram): restore forward origin attribution

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Restores Telegram `forward_origin` attribution for modern forwarded messages, carrying forward the behavior from #4254 / `d07f4208`.
  - Handles the four `forward_origin` shapes covered by this recovery: `user`, `hidden_user`, `chat`, and `channel`.
  - Keeps the existing legacy fallback support for `forward_from`, `forward_from_chat`, and `forward_sender_name`.
  - Preserves forwarded-reply formatting so forward attribution does not collapse the existing Markdown quote block.
- **Scope boundary:** This only changes inbound Telegram message parsing/attribution. It does not change outbound Telegram sending, channel routing, config shape, provider behavior, or non-Telegram channels.
- **Blast radius:** Limited to Telegram channel messages that include forwarded-message metadata.
- **Linked issue(s):** Related #6074. Supersedes #4254.
- **Labels:** `bug`, `risk: medium`, `size: S`, `channel`, `channel:telegram`

```text
$ cargo fmt --all -- --check
Result: passed.

$ git diff --check
Result: passed.

$ cargo test -p zeroclaw-channels forward_origin --features channel-telegram --lib -- --nocapture
running 2 tests
test telegram::tests::format_forward_attribution_supports_forward_origin_variants ... ok
test telegram::tests::parse_update_message_forward_origin_variants_reach_channel_content ... ok
test result: ok. 2 passed; 0 failed; 681 filtered out; finished in 0.03s

$ cargo test -p zeroclaw-channels forwarded_reply --features channel-telegram --lib -- --nocapture
running 1 test
test telegram::tests::parse_update_message_forwarded_reply_keeps_quote_block_separate ... ok
test result: ok. 1 passed; 0 failed; 682 filtered out; finished in 0.02s

$ cargo test -p zeroclaw-channels forwarded --features channel-telegram --lib -- --nocapture
running 7 tests
test result: ok. 7 passed; 0 failed; 676 filtered out; finished in 0.10s

$ cargo clippy -p zeroclaw-channels --lib --features channel-telegram -- -D warnings
Result: passed.

Result: passed.

$ cargo nextest run --locked --workspace --exclude zeroclaw-desktop
Result: interrupted. The run reached 7821 scheduled tests; 7820 passed and 10 were skipped. The only remaining test, existing `zeroclaw-channels telegram::tests::telegram_send_document_bytes_empty_file`, was manually terminated after 384.028s because it is a fake-token Telegram send-path negative test that makes a real network call and hung in this local environment. The new Telegram forwarded-message tests passed inside this run before the interruption.
```

- **Beyond CI — what did you manually verify?** Reviewed the current one-file diff against the Telegram text and attachment attribution paths. The public validation proves the text parser path, all four recovered `forward_origin` variants, the legacy forwarded-message paths, forwarded-reply formatting, and simulated forwarded photo attribution.
- **If any command was intentionally skipped, why:** The repository template's plain `cargo test` command was not run separately; the focused `cargo test` commands above passed, and CI-shaped full nextest was attempted. Nextest did not complete cleanly because one existing outbound Telegram send-path negative test performs a real fake-token network call and hung locally. That brittle test should be fixed separately; it is outside the forwarded-message parser paths changed by this PR, and the new forwarded-message tests passed.

## Compatibility (required)

- Backward compatible? `Yes`

- **Observable failure symptoms:** Telegram forwarded messages could show wrong or missing source attribution, or forwarded replies could render attribution and quoted reply context on the same line.

  - #4254 by @Tenith01
- Scope materially carried forward: Restores the forwarded-message attribution behavior for Telegram forwarded messages, including the modern `forward_origin` variants and the existing text/media parser integration points. The implementation is adapted to current `master` and keeps existing legacy-field fallbacks.


crates/zeroclaw-channels/src/telegram.rs
```diff
@@ -1951,6 +1972,32 @@ Allowlist Telegram username (without '@') or numeric user ID.",
         }
     }
 
+    fn prepend_forward_attribution(attr: &str, content: String) -> String {
```
singlerider, line 2132: 🟢 The `starts_with("> ")` quote-detection here is matched to the real renderer, not guessed — `extract_reply_context` produces `> @sender:\n> ...` (telegram.rs:2179), so a forwarded-and-replied message correctly gets the attribution on its own line with a blank separator instead of collapsed onto the quote block. Centralizing the prepend rule in one helper applied at all four call sites is the right call.

singlerider (approved):
Read the full diff and cross-checked the helpers against the reply-context rendering path in `telegram.rs`. Clean recovery fix with thoughtful de-duplication. Approving.

### 🟢 What looks good — `forward_origin` handled first, legacy fallbacks preserved

Checking `forward_origin` before the legacy `forward_from_chat` / `forward_from` / `forward_sender_name` paths is the correct precedence: modern Telegram sends the unified `forward_origin` object, and the four variants (`user`, `hidden_user`, `chat`, `channel`) cover the documented shapes, each with a sensible `unwrap_or` fallback and an `_ => "unknown source"` catch-all for a type the Bot API might add later. The `match origin_type` here is matching Telegram's wire-protocol discriminator, not an internal supported-options list, so it's the right tool — you can't dispatch on Telegram's JSON without first reading that string. Keeping the legacy arms intact means older update shapes still attribute correctly.

### 🟢 What looks good — the two extracted helpers remove real duplication

`format_forwarded_user_label` pulls the username-or-first/last-name logic out of the inline `forward_from` block and reuses it for both the new `forward_origin` user path and the legacy path — and it's a strict improvement, since it now also appends `last_name` where the old inline version only took `first_name`. `prepend_forward_attribution` is applied uniformly at all four call sites (the two text parsers, the attachment parser, and the test) instead of the scattered `format!("{attr}{content}")`, so the attribution-prepend rule lives in one place.

### 🟢 What looks good — the forwarded-reply formatting fix is verified against the real renderer

The bug this closes is subtle: a forwarded message that's *also* a reply would have collapsed the attribution onto the same line as the quoted reply block. `prepend_forward_attribution` detects `content.starts_with("> ")` and inserts `\n\n` instead of a space — and I confirmed that's exactly the shape `extract_reply_context` produces (`> @sender:\n> ...` at telegram.rs:2179), so the quote-detection is matched to the actual renderer, not a guess. `parse_update_message_forwarded_reply_keeps_quote_block_separate` pins the full expected output including the blank-line separation, which is the right way to lock this.

### 🟢 What looks good — coverage maps to every variant and integration point

Tests exercise all four `forward_origin` variants at both the attribution-format level and end-to-end through `parse_update_message` (proving they reach `ChannelMessage.content`), the forwarded-reply separation, and a simulated forwarded-photo attribution through the attachment path. Supersede attribution for #4254 by @Tenith01 is present with the co-author trailer.

On the validation note: the nextest interruption was the pre-existing `telegram_send_document_bytes_empty_file` making a real fake-token network call — unrelated to this parser change, and notably the exact brittleness #7264 fixes. The forwarded-message tests passed before the hang. Approving.

Issue #6059, opened 2026-04-24
[Bug]: Incompatible with DeepSeek-V4 API format


SSDGADsss:
### Affected component

provider

### Severity

S2 - degraded behavior

### Current behavior

When I was using DeepSeek as the large model provider, both the DeepSeek-V4-Pro and DeepSeek-V4-Flash versions of the API encountered the following error, which appears to be related to the thinking mode:

```
⚠️ Error: All providers/models failed. Attempts:

provider=deepseek model=deepseek-v4-pro attempt 1/3: non_retryable; error=DeepSeek API error (400 Bad Request): {"error":{"message":"The reasoning_content in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

It seems that the format of the request structure is incorrect. I should be able to make a contribution accordingly. Can someone help me with this?

### Expected behavior

The expected behavior should correctly return the large model's result

### Steps to reproduce

```bash
1. zeroclaw onboard
2. Send "List the folders in the current directory"
3. Error!
```

### Logs / stack traces

```text
2026-04-24T08:02:43.632988Z  WARN zeroclaw_providers::reliable: Streaming error: Provider error: 400 Bad Request: {"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}} provider="deepseek" model="deepseek-v4-pro"
2026-04-24T08:02:43.633033Z  WARN zeroclaw_runtime::agent::loop_: provider streaming failed, falling back to non-streaming chat: provider stream error: Provider error: 400 Bad Request: {"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}} provider="deepseek" model="deepseek-v4-pro" iteration=2
2026-04-24T08:02:43.704502Z  WARN zeroclaw_providers::reliable: Non-retryable error, moving on provider="deepseek" model="deepseek-v4-pro" error=DeepSeek API error (400 Bad Request): {"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
2026-04-24T08:02:43.704529Z  WARN zeroclaw_providers::reliable: Exhausted retries, trying next provider/model provider="deepseek" model="deepseek-v4-pro"
[...]

PR #7254, opened 2026-06-05
fix(runtime): strip think blocks before native tool-call output

Audacity88:
## Summary

- **What changed and why:**
  - Strip inline `<think>...</think>` blocks before the agent loop builds display text or persists native assistant tool-call history.
  - Add a streaming text sanitizer so split `<think>` tags cannot leak through strict/native draft deltas before the full response is assembled.
  - Preserve provider `reasoning_content` separately so native replay for thinking-capable providers still carries the structured reasoning field.
  - Replace helper-only coverage with production-boundary regressions for native tool-call display/history and split streaming tags.
- **Scope boundary:** This does not change provider APIs, tool-call schemas, tool execution, tool result handling, config, or the parser crate's full-string `strip_think_tags` helper.
- **Blast radius:** Runtime agent-loop response handling for model text, especially native tool-call turns and streaming draft updates. Prompt-guided tool-call parsing still uses the existing parser path after sanitized response text.
- **Linked issue(s):** None.
- **Labels:** `bug`, `risk: high`, `size: S`, `agent`, `runtime`

```text
cargo fmt --all -- --check
result: passed

git diff --check
result: passed

cargo test -p zeroclaw-runtime run_tool_call_loop_sanitizes_native_tool_call_text_before_display_and_history --lib -- --nocapture
result: passed; 1 passed, 0 failed, 1992 filtered out

cargo test -p zeroclaw-runtime consume_provider_streaming_response_strips_split_think_tags_before_forwarding --lib -- --nocapture
result: passed; 1 passed, 0 failed, 1992 filtered out

cargo clippy --workspace --all-targets --features ci-all -- -D warnings
result: passed

result: incomplete locally; the escalated rerun reached 5013 passed tests before three unchanged delegate background-task timeout tests failed. Each exact failing delegate test passed on focused rerun, and GitHub Actions Test is green on this head.
```

- **Beyond CI — what did you manually verify?** Reviewed the native tool-call response path to confirm sanitized text is used for display text and `build_native_assistant_history`, while `reasoning_content` remains a separate field. Reviewed the streaming path to confirm text is sanitized before either strict/native forwarding or the existing protocol guard sees it.
- **If any command was intentionally skipped, why:** None. CI-shaped nextest was attempted locally but did not complete because of unrelated delegate background-task timeout failures in unchanged tests; the exact failures passed focused, and the hosted GitHub Actions `Test` job is green.

- If any `Yes`, describe the risk and mitigation: No new access or credential surface. The privacy impact is protective: inline hidden reasoning is removed before user-visible draft output, logs, and assistant history content in native tool-call turns.

## Compatibility (required)

- If `No` or `Yes` to either: No user action required.

- **Fast rollback command/path:** Revert this PR's merge commit.
- **Observable failure symptoms:** Native tool-call turns or streamed draft output may omit text if a provider intentionally emits literal `<think>` tags as visible content. Check agent-loop response logs and channel draft output around native tool-call turns.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded contributor PR is incorporated.


WareWolf-MoonWall (approved):
## Review Summary

This PR addresses a privacy leak where inline `<think>...</think>` blocks from model reasoning could appear in user-visible draft output, channel messages, and native tool-call assistant history. The fix introduces a stateful streaming sanitizer (`StreamThinkTagStripper`) that correctly handles split tag boundaries, and applies `strip_think_tags` consistently before building display text or history for native tool-call turns.

**Verdict:** Approve with observations.

**CI Status:** All checks passing (16/16 green).

**Commit reviewed:** afb0ad97f2b1cdb11316c0162357848d5a2de64c

---

### 🟢 What looks good — streaming state machine design

The `StreamThinkTagStripper` implementation correctly addresses the fundamental streaming problem: a `<think>` or `</think>` tag can arrive split across multiple delta chunks. The state machine:

- Buffers partial tag matches using `longest_suffix_matching_prefix` to avoid false-negative matches when `<thi` arrives in chunk N and `nk>` arrives in chunk N+1.
- Tracks `in_think` state to suppress content between tags across chunk boundaries.
- Has a clean `finish()` path that either emits trailing buffered content or drops it if still inside a `<think>` block at stream end.

The test `consume_provider_streaming_response_strips_split_think_tags_before_forwarding` demonstrates this boundary condition explicitly — exactly the kind of production-boundary regression that belongs in this file. Well done.

---

### 🟢 What looks good — preservation of reasoning_content

The diff shows `reasoning_content` flows separately through `build_native_assistant_history` without being stripped. This is correct: providers like DeepSeek V4 that emit structured `reasoning_content` fields expect that field to persist independently for native replay (per issue #6059). The inline `<think>` tags in `text` are hidden reasoning that should NOT persist; `reasoning_content` is explicit structured reasoning that SHOULD persist. The separation is maintained.

---

### 🟢 What looks good — consistent sanitization at the right boundaries

The old code had inconsistent strip-point timing:
- Sometimes stripped before display, sometimes after.
- Stripped conditionally based on `strict_tool_parsing` or `tool_specs.is_empty()`.
- Left gaps where streamed deltas could leak unsanitized text before the full response assembled.

The new code strips uniformly:
- **Streaming path:** Every delta sanitized through `think_stripper.push()` before appending to `outcome.response_text`, before forwarding to `delta_sender`, before `text_guard.push()`.
- **Native tool-call turn:** `strip_think_tags` applied once to `resp.text_or_empty()` immediately after receiving the response, before any display or history logic.

This is the right boundary: sanitize at ingress (streaming) or immediately post-response (native), then let downstream logic (display, history, forwarding) consume clean text.

---

### 🔵 Suggestion — document why `reasoning_content` is NOT stripped

The diff preserves `reasoning_content` separately, which is correct, but nowhere in the PR body or inline comments is there an explanation of **why** `reasoning_content` should remain unsanitized while `text` is sanitized. A one-line comment in `build_native_assistant_history` or a sentence in the PR body stating "reasoning_content is provider-structured reasoning intended for native replay (issue #6059); inline `<think>` tags are hidden reasoning in text and must be stripped before user display" would prevent a future contributor from "fixing" this by stripping both fields.

This is optional — the code is correct, and the separation is load-bearing per #6059, but the intent is implicit. A comment would make it explicit.

---

### 🔵 Suggestion — test coverage for `<think>` tag at stream-end boundary

The new test `consume_provider_streaming_response_strips_split_think_tags_before_forwarding` covers split tags across chunks. A complementary edge case not covered: what happens if the stream ends with an **open** `<think>` tag (no closing `</think>`)? The `finish()` method drops `self.pending` if `in_think` is true, which is correct behavior (suppress incomplete hidden reasoning), but this is not covered by a test.

Example: chunks `["visible text", "<think>incomplete reasoning"]` should produce `"visible text"`, not `"visible text<think>incomplete reasoning"`.

This is not blocking — the code is correct, and the behavior follows from the `finish()` implementation — but a one-line test case in the existing test (or a new focused test) would document the edge case explicitly.

---

### 🔵 Suggestion — consider renaming `strip_think_tags` helper usage to clarify full-string vs. streaming paths

The diff now has two distinct sanitization mechanisms:
- `StreamThinkTagStripper` for streaming deltas (stateful, handles split tags).
- `strip_think_tags` for full-string responses (stateless, from `zeroclaw-tool-call-parser`).

Both are correct for their use cases, but nothing in the code or comments distinguishes **when** to use which. A future contributor editing `run_tool_call_loop` might see `strip_think_tags` at L1988 and wonder "why not use `StreamThinkTagStripper` here?" or vice versa.

A comment at the point of use (L1988) or in the `StreamThinkTagStripper` doc comment stating "Use `strip_think_tags` for full responses; use `StreamThinkTagStripper` for streaming deltas" would document the design intent.

This is optional — the current code is correct, and the separation is implicit in the streaming vs. full-response paths — but explicit guidance would help maintainability.

---

### ✅ Validation — test coverage aligns with blast radius

The PR replaces helper-only coverage with production-boundary regressions:
- `run_tool_call_loop_sanitizes_native_tool_call_text_before_display_and_history` (renamed from `run_tool_call_loop_relays_native_tool_call_text_via_on_delta`) now asserts:
  - Draft deltas contain no `<think>` tags or hidden reasoning.
  - Accumulated result contains no `<think>` tags.
  - Native tool-call assistant history `content` field is sanitized.
  - `reasoning_content` field is preserved separately.
- `consume_provider_streaming_response_strips_split_think_tags_before_forwarding` validates streaming split-tag handling.

Both tests target the exact boundaries identified in the PR body's blast radius statement: "runtime agent-loop response handling for model text, especially native tool-call turns and streaming draft updates." Well-targeted.

---

### ✅ Validation — CI evidence complete

Local validation evidence shows:
- Specific tests passed (`run_tool_call_loop_sanitizes_native_tool_call_text_before_display_and_history`, `consume_provider_streaming_response_strips_split_think_tags_before_forwarding`).
- Full workspace clippy and nextest attempted (delegate background-task timeout failures are unrelated and pass focused; GitHub Actions Test is green).
- Author manually verified native tool-call response path and streaming path per the "Beyond CI" section.

GitHub Actions shows all 16 checks green, including Test. This is sufficient evidence for a `risk: high` runtime change.

---

### ✅ Label hygiene — correctly classified

Labels are correct:
- `bug` — accurate; this is a privacy leak.
- `risk: high` — accurate; this touches `crates/zeroclaw-runtime/src/agent/loop_.rs` and affects user-visible draft output, channel messages, and persisted history.
- `size: S` — accurate; single file, focused scope, 193 additions / 16 deletions.
- `agent`, `runtime` — accurate path labels.

No maintainer action needed.

---

## Final Notes

This is a well-crafted privacy fix. The streaming state machine is correct, the test coverage targets the right boundaries, and the separation of `reasoning_content` from inline `<think>` stripping is preserved. The three suggestions above are all optional — they address documentation and test coverage for edge cases, not correctness.

The PR is ready to merge.

---

**Reviewed by:** @WareWolf-MoonWall  
**Review type:** Technical review per FND-005 feedback taxonomy

singlerider (approved):
@Audacity88 — context: read the full diff at head 92bd1096 (a clean master-merge over afb0ad97, no code change since @WareWolf-MoonWall's approval on that code), the streaming sanitizer, the native tool-call strip point, and traced the `finish()` boundary myself. @WareWolf-MoonWall already gave this a thorough approval; I'm confirming on my own read, not re-litigating. Approving.

### 🟢 What looks good — the split-tag state machine is correct

`StreamThinkTagStripper` handles the real streaming hazard: a `<think>`/`</think>` tag arriving across delta boundaries. `longest_suffix_matching_prefix` buffers exactly the partial-tag suffix into `pending` so `<thi` + `nk>` can't leak as visible text, and the `in_think` flag carries suppression across chunks. The `finish()` path is right too — if the stream ends mid-`<think>` it clears `pending` and emits nothing, so an incomplete hidden-reasoning block at EOF is suppressed rather than leaked. The trailing-delta handling after the loop correctly mirrors the in-loop forwarding (strict vs `text_guard` paths) so buffered-then-flushed text still reaches `delta_sender`.

### 🟢 What looks good — the strip point got simpler and more consistent, not just relocated

The old code stripped conditionally: `strip_think_tags` only when `tool_specs.is_empty()`, plus a second conditional strip when `strict_tool_parsing && calls.is_empty()`. That left native tool-call turns (non-empty tool_specs) able to carry inline `<think>` text into display and `build_native_assistant_history`. Replacing both with one unconditional `strip_think_tags(resp.text_or_empty())` at the response boundary is the correct fix — sanitize once at ingress, let every downstream consumer (display, history, forwarding) see clean text. Fewer branches, no strip-timing gap.

### 🟢 What looks good — `reasoning_content` stays separate, and the test pins it

The load-bearing distinction per #6059 is preserved: inline `<think>` in `text` is hidden reasoning that must be stripped, while provider-structured `reasoning_content` is explicit and must persist for native replay. The regression proves both halves at the production boundary — it asserts the native assistant-history `content` is sanitized (`"Task started..."` with no `<think>` / no "private chain of thought") *and* `reasoning_content` survives as `"provider reasoning"`. Asserting on the persisted history JSON rather than a helper return is exactly the right level; it's a behavior test, not an implementation snapshot.

No active blocks, current head is a master-merge over the approved code, and the change is a correct privacy fix with well-targeted coverage. Approving.

PR #7256, opened 2026-06-05
feat(channels/lark): Feishu integration hardening — reactions, draft streaming, per-user sessions, config knobs, two prod bug fixes

kanmars:
## Summary

- **What changed and why:** Hardens the `LarkChannel` (Feishu) integration along eight independent axes — fixing two production-observed bugs (orphan reactions from random ack pool; image download breakage when image_keys cross messages or tokens silently rotate), adding three opt-in capabilities operators have asked for (per-user group sessions, configurable approval timeout, streaming draft cards), fixing one trait-boundary correctness bug (using fresh UUIDs as `ChannelMessage.id` made downstream `add_reaction` API calls unconditionally fail against Feishu), and addressing the v1 review feedback by making the Lark-local fast-ack and the generic orchestrator path share a single `reaction_ids` cache (eliminating the orphan-reaction race the review flagged) and tightening `patch_card_content` to actually soft-fail every error path its doc comment claims.
- **Scope boundary:** Lark/Feishu only. The orchestrator change is one line per dispatch site (3 sites) wiring the new builder methods through. No other channel implementation, provider, tool, or runtime path changes. No public-API trait surface changes.
- **Blast radius:** `crates/zeroclaw-channels` (Experimental tier) and `crates/zeroclaw-config` (Beta tier). All schema additions use `#[serde(default = "...")]` with named helpers shared across channels (e.g. `default_channel_approval_timeout_secs` is already used by Telegram/Discord/Matrix), so existing TOML files load unchanged.
- **Linked issue(s):** None — these are operator-driven hardening + feature work surfaced from running the channel against Feishu cloud production for several weeks.
- **Labels (suggested):** `area: channels`, `type: feat`, `risk: medium`, `size: L`.

## Commits (9, oldest → newest)

| # | Commit | One-liner |
|---|--------|-----------|
| 1 | `f1fb68d9e` | `fix(channels/lark)`: bind image download to `message_id` with 401 token retry |
| 2 | `981fe98c0` | `feat(channels/lark)`: add `add_reaction` / `remove_reaction` + unicode→Feishu emoji map + reaction-id cache |
| 3 | `f61faf6b1` | `feat(channels/lark)`: add `approval_timeout_secs` config + builder (default 300s, was hard-coded 120s) |
| 4 | `e32d30d39` | `feat(channels/lark)`: add `per_user_session` config + `resolve_sender` (group-chat per-member sessions, opt-in) |
| 5 | `cf3e9c219` | `feat(channels/lark)`: add streaming-draft subsystem (`send_draft` / `update_draft` / `finalize_draft` / `cancel_draft` for Feishu interactive cards) |
| 6 | `3ded675c1` | `fix(channels/lark)`: use Feishu `om_xxx` `open_message_id` as `ChannelMessage.id` (not a fresh UUID) — fixes `add_reaction` API rejections in production |
| 7 | `7a75447a6` | `feat(channels/lark)`: pin inbound ack to single `GLANCE` reaction, remove random pool |
| 8 | `d4978e242` | `fix(channels/lark)`: post-review cleanup — `cancel_draft` leak, `multi_message` fallback, stale field doc |
| 9 | `5c4f80027` | `fix(channels/lark)`: share `reaction_ids` cache between inbound fast-ack and generic `add_reaction`; replace vacuous canary with two wiremock-driven lifecycle regression tests; `patch_card_content` now actually soft-fails transport / token-refresh / non-zero codes per its doc |

Each commit body is unusually thorough — includes the production failure mode, the SSoT analysis for any new field, and an explicit risk assessment.

## Schema additions (`crates/zeroclaw-config/src/schema.rs`)

Three new fields under `[channels.lark.<alias>]` — all opt-in, all default to existing behavior:

```toml
# Auto-deny window for approval cards (was hard-coded 120s, now configurable;
# new default 300s matches Telegram/Discord/Matrix). Operators relying on the
# old 120s window must set this explicitly.
approval_timeout_secs = 300

# Group-chat session keying. false (default) = whole group shares one
# agent session keyed on chat_id (existing behavior). true = each sender's
# open_id keys their own session. 1-on-1 chats unaffected (chat_id is
# already unique per user-bot pair).
per_user_session = false

# Streaming-draft mode for token-by-token edits of a single Feishu
# interactive card. "off" (default) = post one final reply (existing).
# "partial" = open a card and PATCH it as tokens arrive, finalize on
# completion. "multi_message" is rejected and falls back to "off" with
# a one-time warn log per channel build (Feishu has no equivalent
# surface).
stream_mode = "off"

# Minimum interval between consecutive PATCHes of the same draft card.
# Tuned to Feishu's 5 QPS-per-message edit cap. Configurable for
# operators on enterprise plans with raised limits.
draft_update_interval_ms = 1000
```

## Hardening (commit 9, addressing v1 review feedback)

Three changes layered on top of commits 1–8 in response to review:

1. **Shared `reaction_ids` cache between fast-ack and generic path.** The Lark-local inbound `spawn` in `listen_ws` / `listen_http` no longer calls a private `try_add_ack_reaction` helper; it routes through `<LarkChannel as Channel>::add_reaction("👀")` so Feishu's returned `reaction_id` is always written into the same `reaction_ids` `HashMap` the `remove_reaction` path reads from. The trait impl gains a cache-hit fast-path: a second `add_reaction(om_xxx, "👀")` for an entry already in cache returns `Ok(())` without issuing a POST, logged at DEBUG with `error_key="lark.add_reaction.cache_hit_dedupe"`. Net effect: exactly one POST per `(message, emoji)` reaches Feishu regardless of which path triggers it; `remove_reaction("👀")` always finds the cached id; the orphan-reaction race the review flagged cannot occur. The 70-line `try_add_ack_reaction` helper was removed since both receive paths now share the trait impl.
2. **`patch_card_content` actually soft-fails every error path its doc comment claims.** Both `patch_or_send_once(...).await?` call sites switched from `?` to explicit `match` blocks. Transport / TLS / connection / token-refresh-still-401 failures are now logged at WARN with stable `error_key` values and the function returns `Ok(())` instead of propagating an `Err`, matching the doc-comment promise that the streaming token loop is never broken by a single missed PATCH. Five stable `error_key`s added for operator log filtering: `lark.draft_patch.transport_failure`, `lark.draft_patch.transport_failure_on_retry`, `lark.draft_patch.unauthorized_after_refresh`, `lark.draft_patch.rate_limited`, `lark.draft_patch.non_zero_code`.
3. **Vacuous canary replaced with two wiremock-driven lifecycle regression tests.** The previous `lark_inbound_ack_policy_is_glance_only_no_random_pool` (which the review correctly identified as `assert_eq!("GLANCE", "GLANCE")` — unable to fail) was removed. Replaced by:
   - `lark_inbound_ack_lifecycle_swaps_glance_to_done_with_no_orphan` — drives the full `add 👀 → remove 👀 → add ✅` lifecycle through the public `Channel` trait against a `wiremock::MockServer`, asserts exactly 1 POST per emoji and exactly 1 DELETE on the cached `reaction_id` via `expect(1)`, and asserts the final `reaction_ids` cache contains only ✅.
   - `lark_fast_ack_and_generic_path_dedupe_on_cache_hit` — explicitly models the fast-ack + generic-path race the review described: fast-ack POSTs 👀, then the generic path also calls `add_reaction("👀")` on the same message; asserts the second call is a cache-hit no-op (the wiremock POST counter sees one request total, not two) and that the cached `reaction_id` survives unchanged through the dedupe.

A regression that re-introduced a parallel ack `spawn` bypassing the shared cache would make either `expect(1)` fail with `"received 2 requests, expected 1"` — the failure mode the previous canary could not catch.

Run from a clean checkout of this branch on `linux-x86_64`, Rust stable.

```bash
$ cargo fmt --all -- --check
# exit 0, no output

$ cargo clippy -p zeroclaw-channels --all-targets --features channel-lark -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 19.48s
# exit 0

$ cargo test -p zeroclaw-channels --lib --features channel-lark
    Finished `test` profile [unoptimized + debuginfo] target(s) in ~1m
test result: ok. 778 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 2.01s
```

The Lark module-level test count went from ~30 → 93 across this PR (`grep -cE '#\[(tokio::)?test\]' crates/zeroclaw-channels/src/lark.rs` ⇒ 93), driven mostly by `wiremock`-backed tests for the new draft / reaction / image-download / shared-cache paths.

**Beyond CI — what was manually verified:**

The branch has been running on a Feishu-connected production deployment for several days, including several days on the post-hardening commit. Sample inbound→outbound trace from `zeroclaw.log`, demonstrating the `om_xxx` message-id fix and draft streaming both active:

```
INFO channel_message_inbound  channel=feishu.feishu  message_id=om_x1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  attrs={"content":"<placeholder>","sender":"oc_xxxxxxxx..."}
DEBUG zeroclaw_channels::orchestrator  Streaming decision  attrs={"has_target_channel":true,"use_draft_streaming":true}
DEBUG zeroclaw_runtime::agent::loop_  Streaming decision for iteration 1  attrs={"has_on_delta":true,"should_consume_provider_stream":true,"supports_streaming":true}
INFO  llm_response  attrs={"model":"<provider.model>","input_tokens":N,"output_tokens":M}
INFO  reply delivered  attrs={"message_id":"om_x1xxxxxxxxxxxxxxxxxxxxxxxxxxxxxx","reply_target":"oc_xxxxxxxx..."}
```

Note that the inbound `message_id` (`om_x1...`) survives untouched into the outbound `reply delivered` event — pre-fix it would have been a fresh UUID at the inbound site, breaking `add_reaction` calls.

**Shared-cache contract verified in production.** Across multiple inbound messages on the post-hardening deployment the new `error_key="lark.add_reaction.cache_hit_dedupe"` DEBUG log fires on every message, confirming the fast-ack populated the shared cache and the subsequent generic-orchestrator `add_reaction("👀")` correctly deduped to a no-op. Wire-level POST count for 👀 per message: 1 (fast-ack only; generic path skipped). User-visible: 👀 appears within ~100 ms of the inbound message and is correctly swapped to the completion marker; no orphan reactions observed.

Other manual scenarios exercised on the live deployment:
- Inbound text → `GLANCE` 👀 reaction appears within ~100ms (fast-ack), then `DONE` ✅ replaces it on completion.
- Inbound image → downloaded successfully even after a silent app-token rotation (verified by `404 → 401 → token refresh → 200` retry log triplet).
- Streaming draft mode → card opens with first chunk, edits arrive at the configured interval, finalizes on LLM completion. Rate-limit code 230020 handled as soft-fail; transport / token-refresh errors now also soft-fail per commit 9.
- Approval card with `approval_timeout_secs = 300` → auto-deny fires at the configured time, not the previous hardcoded 120s.

What was **not** manually verified in production: the `multi_message` fallback path (we don't run that mode), the streaming-draft `cancel_draft` cleanup under churn (the fix from commit 8 — exercised only in unit tests with `wiremock`), and the new `patch_card_content` transport-failure soft-fail branches (fault-injection only viable via `wiremock`; covered by existing draft-rate-limit tests touching the same code path).

## Self-review notes (read-out from internal pre-PR Oracle review)

Surfacing this so reviewers don't have to discover it independently:

1. **`LarkChannel` snapshots config at construction time** for the new fields (`per_user_session`, `stream_mode`, `draft_update_interval_ms`, `approval_timeout_secs`). This matches the existing pattern in `lark.rs` (`mention_only`, `excluded_tools`, `use_feishu`, …) and the established practice across other channels — it is technically a "snapshot of canonical config" per AGENTS.md's strictest SSoT reading, but consistency with the existing channel-construction pattern won out. A repo-wide refactor to make all channel-config access live-resolve through `Arc<RwLock<Config>>` is its own follow-up touching every channel; this PR does not change the established surface area.

The "two `add_reaction` POSTs per inbound message" caveat in the v1 of this PR has been **eliminated** by commit 9's shared-cache + dedupe-fast-path; both paths now share `reaction_ids`, the second POST is suppressed, and a wiremock regression test enforces `expect(1)` on the wire. The vacuous canary (`lark_inbound_ack_policy_is_glance_only_no_random_pool`) was also replaced as part of the same commit.

- New external network calls? **No new endpoints.** All new request paths target Feishu's `open.feishu.cn` (already in use): `POST /im/v1/messages/:id/reactions`, `DELETE /im/v1/messages/:id/reactions/:reaction_id`, `POST /im/v1/messages` with card-content, `PATCH /im/v1/messages/:id`, `GET /im/v1/messages/:id/resources/:image_key`. All require existing `tenant_access_token`.
- Secrets / tokens / credentials handling changed? **No new secrets.** The image-download fix improves an existing path (single 401-retry after `invalidate_token() + refresh`); does not introduce a new credential type.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No.** Test fixtures use neutral placeholder open_ids (`ou_testuser123`, etc.), neutral message bodies, neutral app credentials (`cli_test_app_id`). The production log sample above has all real ids replaced with `om_x1xxx…` / `oc_xxx…` placeholders.

## Compatibility (required)

- Backward compatible? **Yes** at the schema/config level.
- Config / env / CLI surface changed? **Schema only**, all additive with `#[serde(default)]`. No CLI changes, no env-var changes.
- One default-value behavior change worth calling out: `approval_timeout_secs` default went from hard-coded 120s → 300s (matching the channel-wide standard). Operators who relied on the 120s window must now set `approval_timeout_secs = 120` explicitly in their TOML. Documented in commit `f61faf6b1`'s body.
- One internal-behavior change worth calling out for ops: `patch_card_content` no longer propagates transport errors as `Err` — they are now logged at WARN with stable `error_key`s and swallowed. Operators who previously caught these via callsite `Result` handling should now grep on the new `error_key`s instead. Documented in commit `5c4f80027`'s body.

- **Fast rollback command/path:** `git revert <merge-sha>`. All nine commits live entirely in `crates/zeroclaw-channels/src/lark.rs` (+2194/-…), `crates/zeroclaw-channels/src/orchestrator/mod.rs` (~13 lines across 3 dispatch sites), `crates/zeroclaw-config/src/schema.rs` (+46), and `src/config/mod.rs` (+4). Total across these four files: +1748/-566. No migrations. No persistent state.
- **Feature flags or config toggles:** Three of the new capabilities (`per_user_session`, `stream_mode`, configurable `approval_timeout_secs`) are opt-in via TOML — operators can disable them by removing the key (or setting `stream_mode = "off"` / `per_user_session = false`). The two bug fixes (image-download `message_id` binding, `om_xxx` `ChannelMessage.id`) cannot be disabled without revert; they restore correctness against the Feishu API contract. The shared-cache + dedupe behavior from commit 9 likewise cannot be disabled — disabling it would re-introduce the orphan-reaction race the review flagged.
  - `add_reaction` regressions → grep `zeroclaw.log` for `Feishu` + `99992354` (the "invalid open_message_id" error code from the pre-fix state).
  - Shared-cache regressions → look for `error_key="lark.add_reaction.cache_hit_dedupe"` to STOP firing on inbound, or for two POSTs to `/im/v1/messages/:id/reactions` for the same message.
  - Draft streaming regressions → grep for `Streaming decision` with `use_draft_streaming:true` followed by absence of `reply delivered`, or for any `error_key="lark.draft_patch.*"` at WARN.
  - Approval-timeout regressions → grep for `approval auto-denied after` with unexpected `secs` values.
  - Image-download regressions → grep for `download_image_as_marker` + `401` on the second retry (would indicate the token refresh isn't taking).

## Notes for reviewers

- The branch was rebased onto `master` and one prep commit (`fix(providers/ollama): wrap unwrapped temperature in Some(...)`) was dropped before opening this PR — that commit was a stale workaround for a build break upstream master had already fixed via `746c616e4` (#7231). With both attempts to fix the same problem layered on top of each other the result was a fresh `E0308`. After dropping the local commit, master's fix is in effect. Net diff to ollama.rs vs master: zero lines.
- All nine commits compile and lint cleanly individually (`cargo clippy --all-targets --features channel-lark -- -D warnings` passes at every step).
- Conventional-commit titles + thorough bodies; happy to squash on land if maintainers prefer one PR = one commit, though the eight independent concerns argue for keeping them separate in `master` history.


kanmars:
image
image

Audacity88 (changes requested):
Thanks for pushing this forward. The `om_xxx` message-id change and message-scoped resource downloads are the right direction for making Feishu reactions and attachments behave like the rest of the channel stack.

Checked the live PR state, PR body, visible review/comment snapshots, CI/check snapshot, and the public diff/source context for the Lark channel, orchestrator wiring, and config schema changes. I did not run the Feishu integration locally or independently replay the live Feishu API evidence from the PR body.

### 🟢 What looks good — Native message IDs and resource-bound downloads

I like the direction of making `ChannelMessage.id` carry Feishu's native `om_xxx` id and moving image downloads onto `/im/v1/messages/{message_id}/resources/{image_key}`. Those are the kinds of channel-boundary fixes that should make the generic orchestrator reaction and multimodal flows much less surprising once the lifecycle issue below is cleaned up.

### 🔴 Blocking — The early Lark GLANCE ack can bypass the reaction cache

This keeps the Lark-local `try_add_ack_reaction(..., "GLANCE")` spawn in both receive paths while also making `ChannelMessage.id` the native Feishu `om_xxx` id, so the orchestrator's generic `add_reaction(&msg.reply_target, &msg.id, "👀")` now targets the same message too. The new cached cleanup contract is on the generic `add_reaction`/`remove_reaction` path: `add_reaction` stores Feishu's returned `reaction_id`, and the later `remove_reaction("👀")` can only delete a cached reaction. The PR does not show that the Lark-local fast ack shares that cache, nor does it prove the duplicate generic POST will always return a usable `reaction_id` after the fast ack has already added GLANCE. That leaves a plausible path where the final cleanup is a cache miss and the original 👀 remains beside the later ✅/⚠️ completion marker, regressing the advertised "swap 👀 → done" lifecycle even though the user initially sees one ack.

Please pick one owner for inbound ack state now that `msg.id` is the Feishu message id, or make the Lark-local ack use the same cached reaction-id contract as the generic path. I would also replace the current `lark_inbound_ack_policy_is_glance_only_no_random_pool` canary with an executable regression test for this lifecycle; as written it only asserts a constant against itself, so it would not catch this race.

### 🟡 Warning — Draft PATCH error handling is narrower than the comments claim

The streaming draft comments say `patch_card_content` soft-fails every transport/token-refresh error, but the helper still uses `?` on `patch_or_send_once` before it reaches the soft-fail handling. Either make the helper actually catch/log those paths or narrow the comments and tests so the production-hardening guarantee matches the implementation.

kanmars:
Outdated information removed.

kanmars:
@Audacity88 Thanks for the thorough review. All three findings addressed in commit `5c4f80027`; details below.

## 🔴 Blocking — Lark fast-ack race / orphan reaction

Took the second option from your review: "make the Lark-local ack use the same cached reaction-id contract as the generic path."

**Implementation:**

- The fast-ack `spawn` in both `listen_ws` and `listen_http` now invokes `<LarkChannel as Channel>::add_reaction(..., "👀")` instead of a private `try_add_ack_reaction` helper. The trait impl is the single writer of `reaction_ids`, so Feishu's returned `reaction_id` is always cached regardless of which path triggered the POST.
- Added a cache-hit fast-path at the top of `add_reaction`: if `(message_id, emoji)` is already in `reaction_ids`, the call returns `Ok(())` without issuing a second POST. This is what lets the fast-ack and the generic orchestrator path coexist on the same message without racing — exactly one POST per `(message, emoji)` reaches Feishu, and `remove_reaction` always finds the cached id.
- Removed the 70-line `try_add_ack_reaction` helper and its duplicated POST/retry/decode logic; both receive paths now share the trait impl.

**Cache contract (now uniform across both paths):**

1. `add_reaction` POST success → write `(message_id, emoji) → reaction_id` into `reaction_ids`.
2. `add_reaction` cache hit → no-op `Ok(())`, no POST, DEBUG log with `error_key="lark.add_reaction.cache_hit_dedupe"`.
3. `remove_reaction` → atomic `remove()` from cache, then DELETE on the returned reaction_id. Cache miss is silent no-op (preserves pre-existing restart-tolerant behavior).

**Production verification:** Deployed to a private instance and sent several inbound messages on the post-hardening build. The new `lark.add_reaction.cache_hit_dedupe` DEBUG log fires on every inbound, confirming the fast-ack populated the shared cache and the later generic-path `add_reaction("👀")` correctly deduped. Wire-level POST count for 👀 per message: **1**. User-visible: 👀 appears within ~100 ms of inbound (fast-ack), correctly swapped to the completion marker when the LLM finishes; no orphans observed.

## 🔴 Blocking — Canary replacement

Removed `lark_inbound_ack_policy_is_glance_only_no_random_pool` (the `assert_eq!("GLANCE", "GLANCE")` you flagged).

Added two executable wiremock-driven regression tests:

- **`lark_inbound_ack_lifecycle_swaps_glance_to_done_with_no_orphan`** — drives the full `add 👀 → remove 👀 → add ✅` lifecycle through the public `Channel` trait against a `wiremock::MockServer`, asserts exactly 1 POST per emoji and exactly 1 DELETE on the cached `reaction_id` via `expect(1)`, and asserts the final `reaction_ids` cache contains only ✅ (the 👀 entry was removed; no orphan).
- **`lark_fast_ack_and_generic_path_dedupe_on_cache_hit`** — explicitly models the race you described: fast-ack POSTs 👀, then the generic-orchestrator path also calls `add_reaction("👀")` on the same message. Asserts the second call is a cache-hit no-op (the wiremock POST counter receives 1 request total, not 2), that the cached `reaction_id` is preserved unchanged (no overwrite), and that the subsequent `remove_reaction` DELETEs the fast-ack's id correctly.

A regression that re-introduces a parallel ack `spawn` bypassing the shared cache would make either `expect(1)` fail with `"received 2 requests, expected 1"` — the failure mode the previous canary could not catch.

## 🟡 Warning — `patch_card_content` soft-fail scope

Rewrote the call site so the implementation actually matches the doc comment claim of soft-failing every error path.

- Both `patch_or_send_once(...).await?` call sites switched to explicit `match` blocks. Transport / TLS / connection / token-refresh-still-401 errors are now logged at WARN with stable `error_key`s and the function returns `Ok(())` instead of propagating.
- Five stable `error_key`s added for operator log filtering: `lark.draft_patch.transport_failure`, `lark.draft_patch.transport_failure_on_retry`, `lark.draft_patch.unauthorized_after_refresh`, `lark.draft_patch.rate_limited`, `lark.draft_patch.non_zero_code`.
- Doc comment updated to reflect the new invariant: signature still returns `anyhow::Result<()>` for caller-shape compatibility but the function never returns `Err`. Every failure path is logged at WARN/DEBUG and swallowed.

## Local validation

```
cargo fmt --all -- --check                                         # clean
cargo clippy -p zeroclaw-channels --all-targets \
    --features channel-lark -- -D warnings                         # clean
cargo test -p zeroclaw-channels --lib --features channel-lark      # 778 passed; 0 failed; 1 ignored
```

Ready for re-review.

Audacity88 (approved):
Thanks for the follow-up. I re-checked the live PR state, the author response in the PR comments, current checks, the public diff/source for `lark.rs`, orchestrator wiring, and the config schema updates. I used only public PR evidence; I did not run the Feishu integration locally, independently replay the live Feishu API evidence from the PR body, or use local validation logs beyond what the author posted.

### ✅ Resolved — Fast ack now shares the reaction-id cache

The revised websocket and webhook fast-ack paths now call `<LarkChannel as Channel>::add_reaction(..., "👀")` instead of maintaining a private Lark-only POST helper. That puts the early ack on the same public trait contract as the generic orchestrator path: successful POSTs write Feishu's returned `reaction_id` into `reaction_ids`, and a later duplicate `add_reaction("👀")` becomes a cache-hit no-op instead of a second wire call.

That addresses the earlier orphan-risk concern. `remove_reaction("👀")` now has the same cached id whether the reaction was first created by the fast ack or the generic path, so the final lifecycle can cleanly swap the processing marker before adding the completion marker.

### ✅ Resolved — The lifecycle regression is now executable

The old constant-only canary is gone, and the replacement tests cover the actual invariant I asked for. `lark_inbound_ack_lifecycle_swaps_glance_to_done_with_no_orphan` drives `add 👀 -> remove 👀 -> add ✅` through the public `Channel` surface and asserts the final cache shape. `lark_fast_ack_and_generic_path_dedupe_on_cache_hit` models the fast-ack plus generic-path duplicate and asserts exactly one POST for 👀, preservation of the cached fast-ack reaction id, and a DELETE against that id.

Those tests cover the shared trait/cache lifecycle, and the current diff routes both inbound fast-ack paths through that same trait surface. That is the right shape for this bug because it pins the cache contract where both the fast ack and generic orchestrator path now meet.

### ✅ Resolved — Draft PATCH soft-fail behavior matches the comment

`patch_card_content` now explicitly handles both first-attempt and retry transport failures with `match`, logs the failure with stable `error_key`s, and returns `Ok(())`. The retry-still-unauthorized, rate-limit, and non-zero business-code paths are also logged and swallowed as documented. The implementation now matches the stated streaming contract: a missed card edit should not abort the token loop.

### 🟢 What looks good — The channel boundary is cleaner

The original positive direction still holds: using Feishu's native `om_xxx` ids for `ChannelMessage.id`, binding resource downloads to the owning message id, and keeping runtime-only state in bounded per-message caches are all consistent with the generic channel trait boundary.

Approving from my side. This clears my earlier changes-requested review; I am not making a broader merge-readiness claim beyond my review.

PR #7267, opened 2026-06-05
feat(config): per-field editing for [[mcp.servers]] via #[natural_key]

tidux:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Previously only `config.toml` could edit `[[mcp.servers]]` per-field. The web dashboard rendered it as a JSON-array blob (`ObjectArrayEditor`), and the zerocode TUI had no MCP UI at all — operators were forced into hand-editing TOML.
  - This PR ships the runtime capability that makes per-field editing work for `Vec<T> + #[nested]` list sections, opts `mcp.servers` into it, and surfaces the section in the zerocode TUI through the existing `OneTierAliasMap` rendering path (zero TUI-side code changes — the TUI already drives that shape for `risk_profiles`, `cron`, etc.).
  - Operators can now `+ Add` a server (seeded `name`), edit `transport` / `command` / `url` / `headers` / `env` / `tool_timeout_secs` as individual fields, and delete by name — all from the zerocode TUI, hitting the same RPCs the dashboard will eventually migrate onto.
  - Continues the line of work started by `docs(mcp): document HTTP and stdio server config schema` (`445de193e`) on this branch: the docs commit told operators which fields existed; this PR lets them set those fields without leaving the TUI.
- **Scope boundary:**
  - **No web dashboard changes.** The dashboard's `ObjectArrayEditor` still drives `mcp.servers`; migration sketch is filed in `docs/dashboard-mcp-per-field-todo.md` for separate follow-up.
  - **No rename UI in the TUI.** The runtime now supports `config_map_key_rename` for `mcp.servers`, but zerocode does not expose a rename keybind for any `OneTierAliasMap` section (matching every existing one). Adding it is a general TUI enhancement out of scope.
  - **Only `mcp.servers` opts into `#[natural_key]`.** The other seven `Vec<T> + #[nested]` schema fields (`ClassificationRule`, `EmbeddingRouteConfig`, `GoogleWorkspaceAllowedOperation`, `ModelRouteConfig`, `NevisRoleMappingConfig`, `PeripheralBoardConfig`, `ToolFilterGroup`) keep their existing `ObjectArray` JSON-blob behaviour — gated explicitly via the new attribute.
- **Blast radius:**
  - `zeroclaw-macros` derive emits new code for the `#[natural_key]` arm; **the arm is dead code on every existing struct** because no other field carries the attribute. Verified by 730 pre-existing config tests + 1984 runtime tests + 201 gateway tests + 6 macros tests + 14 zerocode tests — all green, zero regressions.
  - One genuine behavioural change touches all `Configurable` derives: nested `set_prop` / `get_prop` and the map-key recurse arms now propagate inner errors instead of silently swallowing them. This is justified separately in `fix(macros): propagate inner errors through nested-prop & map-key arms` and surfaces previously-hidden actionable errors (e.g. a typo in a property name now returns `Unknown property` pointing at the right path instead of being eaten by a wrong-field fallthrough). No existing test relied on the swallowed-error behaviour.
  - `handle_config_sections` refines its parent-collapse rule (`fix(rpc): keep parent sections with direct scalars beside dotted children`) so `mcp` does not disappear when `mcp.servers` is added to the curated section list. `providers` still vanishes correctly because it has no direct scalar fields.
- **Linked issue(s):** None.
- **Labels:** Snapshot after auto-labels apply.

`cargo fmt --all -- --check`:

```
(exit 0, no output — tree is clean)
```

`cargo clippy -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-macros -p zeroclaw-gateway -p zerocode --all-targets -- -D warnings`:

```
    Checking zeroclaw-runtime v0.8.0-beta-2 (crates/zeroclaw-runtime)
    Checking zeroclaw-hardware v0.8.0-beta-2 (crates/zeroclaw-hardware)
    Checking zeroclaw-channels v0.8.0-beta-2 (crates/zeroclaw-channels)
    Checking zeroclaw-gateway v0.8.0-beta-2 (crates/zeroclaw-gateway)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 26.12s
```

(Initial clippy run flagged my macro arm's use of `anyhow::anyhow!` as a disallowed macro per the workspace lint config; fixed by switching to `anyhow::Error::msg(format!(...))` and fixed-up into the originating commit before the rebase. The above is from the final post-fix run.)

`cargo test -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-macros -p zeroclaw-gateway -p zerocode --lib` (tails of each crate):

```
test result: ok. 734 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s   (zeroclaw-config)
test result: ok. 201 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.53s   (zeroclaw-gateway)
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s     (zeroclaw-macros)
test result: ok. 1984 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 5.17s  (zeroclaw-runtime)
test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s    (zerocode)
```

Totals: **2,939 passed, 0 failed, 1 ignored** (the ignored case is pre-existing and unrelated).

  - End-to-end round-trip on `Config`: `create_map_key("mcp.servers", "fs")` then `set_prop("mcp.servers.fs.command", ...)` then `set_prop("mcp.servers.fs.transport", "http")` then `get_prop("mcp.servers.fs.command")` round-trips correctly. Covered by `mcp_servers_addable_via_create_map_key_and_per_entry_props` (replaces the prior "future work" stub).
  - Read-only `name` field: `set_prop("mcp.servers.fs.name", ...)` returns an explicit error pointing at `config_map_key_rename`. Covered by the same test.
  - Rename: `rename_map_key("mcp.servers", "fs", "filesystem")` mutates in place; new key resolves, old key stops resolving.
  - Ambiguity: two entries with the same `name` cause `set_prop` / `get_prop` / `rename_map_key` to return an actionable error without mutating state. Covered by `mcp_servers_routing_is_ambiguous_on_duplicate_names`.
  - Rename collision: `rename_map_key` refuses with `natural key already exists` when the target name is taken. Covered by `mcp_servers_rename_refuses_when_new_key_is_taken`.
  - `prop_fields()` hides the `name` field (read-only) but surfaces `mcp.servers.<name>.command` and friends. Covered by the same test.
  - `Section::McpServers.shape() == OneTierAliasMap`, parent `Section::Mcp` still `DirectForm`, canonical order `Mcp -> McpServers -> McpBundles`. Covered by `mcp_servers_section_has_alias_map_shape_and_parent_keeps_direct_form`.
  - **Not manually verified: an interactive TUI smoke walk** — I cannot run a TTY from this environment. The code path the TUI takes (`config_sections` returns `mcp.servers` with shape `one_tier_alias_map` -> `enter_section` dispatches `OneTierAliasMap` -> `load_aliases` calls `config_map_keys` -> `enter_alias` calls `load_fields` -> per-field `set_prop`) is identical to the path the other `OneTierAliasMap` sections drive successfully today, and every link in that chain is covered by an automated test in this PR.

  - `cargo test --workspace` (full-workspace test): **skipped — environment-blocked.** The Tauri-using desktop crate transitively requires `libdbus-1` system headers, which are not installed on this host; build fails inside `libdbus-sys`'s build.rs before reaching any test. Failure is unrelated to this PR (reproduces on master without these changes) and CI runs in a container that has the system deps. The five crates this PR actually touches all run cleanly per above.

- Secrets / tokens / credentials handling changed? `No` — `McpServerConfig::headers` was already `#[secret]` and the natural-key arm forwards through the existing per-element `set_secret` / `prop_is_secret` walk. The bias-toward-over-marking in the new `prop_is_secret` arm is the same shape as the HashMap arm.

## Compatibility (required)

- Backward compatible? `Yes`. The on-disk and on-the-wire shape of `[mcp]` and `[[mcp.servers]]` is unchanged — same TOML, same RPC method names, same JSON-Schema. Existing `config.toml` files load unchanged; the dashboard's JSON-array editor continues to work because `set_prop("mcp.servers", "<json>")` still routes through the legacy `ObjectArray` path on the JSON-array shape.
- Config / env / CLI surface changed? `No on existing surfaces; Yes on additions only`:
  - **Added** `Section::McpServers` curated section (URL key `mcp.servers`, kebab `mcp-servers`). Dashboards / CLIs that enumerate `QUICKSTART_SECTIONS` will see one new entry; nothing existing changes shape.
  - **Added** `#[natural_key]` field attribute on the `Configurable` derive. Existing structs that do not use it get unchanged emitted code.
- If `No` or `Yes` to either: exact upgrade steps for existing users: **None.** No-op upgrade.

Auto-label will likely tag this `risk: high` because it touches `crates/zeroclaw-runtime` and `crates/zeroclaw-gateway`. Filling the rollback section explicitly.

- **Fast rollback command/path:** `git revert 2ba5bab05 fb31c4604 d6b3d814a 2fb5055b6 e0eab48ca bf5a04763 6a62e06d6` (the seven commits in this PR, in reverse order). The previous commit `docs(mcp): document HTTP and stdio server config schema` (`445de193e`) is unaffected and stays.
- **Feature flags or config toggles:** `None`. The behaviour is gated structurally: existing `Vec<T> + #[nested]` fields without `#[natural_key]` keep their legacy behaviour, so the only field affected by a revert is `McpConfig::servers`. After revert, `set_prop("mcp.servers.<name>.<field>", ...)` stops working but the dashboard's JSON-blob editor and hand-editing `config.toml` continue to function unchanged.
  - Grep logs for `Unknown property 'mcp.servers.` — if this appears at runtime for a path the TUI is editing, the natural-key arm is not routing correctly.
  - Grep for `is ambiguous in 'mcp.servers'` — expected only when the operator has saved a duplicate-`name` config; if it fires under other conditions, the duplicate detection is misclassifying.
  - Grep for `no map-keyed/list section at 'mcp.servers'` — should never appear post-merge unless `Section::McpServers` was removed without also reverting the natural-key arm.
  - Metric: none added.
  - Alert: none added.


WareWolf-MoonWall (approved):
## Review by @WareWolf-MoonWall

Reviewed head commit: bf5a04763 (merge of master into docs/mcp-http-config-schema)

This is a substantial capability expansion that ships the runtime machinery for per-field editing of `Vec<T> + #[nested]` list sections, opts `mcp.servers` into it, and surfaces that section in the zerocode TUI — all while keeping the change surgically scoped to one schema field through an explicit opt-in attribute. The engineering is sound, the test coverage is thorough, and the blast radius is correctly bounded.

---

### 🟢 What looks good — Architecture and single-source-of-truth discipline

The natural-key approach is the **correct** architectural choice for this problem space. `McpServerConfig::name` remains the single source of truth for each server's identity; the derive macro generates routing code that resolves that field on-demand at every access, with zero caching. This satisfies AGENTS.md's SSOT rule: the `#[natural_key = "name"]` attribute on `McpConfig::servers` tells the macro **where** the canonical key lives, and `route_vec_path` resolves it from the live `Vec` every time a caller addresses an element by name. No duplicate key storage, no stale-key bugs, no operator-edits-config-but-runtime-still-sees-old-name failure mode.

The macro's emitted `rename_map_key` arm mutates `McpServerConfig::name` in place via `set_prop`, and the `set_prop` arm for the natural-key field itself **refuses** with an explicit error pointing callers at `config_map_key_rename`. This keeps the natural key stable from the routing surface's point of view during in-flight edits — the two-phase contract (refuse direct writes; offer rename as a separate validated operation) is what stops the "operator edits `name` mid-session and strands all pending references" hazard.

**Why this matters:** the previous channel-`allowed_users` caching bug (cached Vec in handles, config reload didn't refresh it) happened because state lived in two places and one copy went stale. This PR's natural-key surface has zero such copies: every `get_prop("mcp.servers.<name>.<field>")` walks the live `Vec`, matches on the live `name` field, and dispatches to the matched element's own `get_prop`. If the operator renames a server through `rename_map_key`, the next property access sees the new key immediately because it resolves from the canonical field. No drift window.

---

### 🟢 What looks good — Opt-in gating preserves stability

The `#[natural_key]` attribute is **gated** on every code path: the macro only emits the per-element routing arms when the attribute is present on the `Vec` field. The seven other `Vec<T> + #[nested]` schema fields (`ClassificationRule`, `EmbeddingRouteConfig`, `GoogleWorkspaceAllowedOperation`, `ModelRouteConfig`, `NevisRoleMappingConfig`, `PeripheralBoardConfig`, `ToolFilterGroup`) carry no such attribute and keep their legacy `ObjectArray` whole-array round-trip behavior unchanged.

Verified by the test counts: 2,939 tests pass, zero regressions. The 730 config tests, 1984 runtime tests, 201 gateway tests, 14 zerocode tests all green — none of them rely on changed behavior because the only field with the attribute is `mcp.servers`, and the old test for that field was a stub comment ("future work"). The macro's new arms are dead code for every struct without the attribute.

This gating strategy is the right balance: ships the capability for one concrete use case (`mcp.servers`) with full end-to-end validation, leaves a clear migration path for the other seven types (add one line `#[natural_key = "<field>"]` per type when their dashboards are ready), and introduces zero behavior change to unrelated sections.

---

### 🟢 What looks good — Ambiguity handling is explicit and safe

The `route_vec_path` helper returns a three-variant enum (`VecRoute::Hit`, `VecRoute::Ambiguous`, `VecRoute::Miss`) instead of silently picking one duplicate on ambiguity. When two or more elements share the same natural key, `get_prop` / `set_prop` / `rename_map_key` all surface explicit errors ("natural key `<key>` is ambiguous in `mcp.servers`: N entries share it; fix duplicates first") rather than mutating one duplicate by accident.

The schema validator (`validate_mcp_config`) already rejects duplicate `name` at save time, so the ambiguity path should only fire when an operator has hand-edited TOML into a broken state and is attempting in-flight repairs. The routing surface refuses to proceed until the config is repaired — this is the correct fail-safe: better to block an edit with an actionable error than to silently edit the wrong element and deepen the corruption.

Covered by `mcp_servers_routing_is_ambiguous_on_duplicate_names`, which verifies that both `set_prop` and `get_prop` refuse without mutating state, and that `rename_map_key` likewise refuses with the same error string.

---

### 🟢 What looks good — Test coverage is comprehensive

The validation evidence lists 2,939 tests passing (730 config, 1984 runtime, 201 gateway, 6 macros, 14 zerocode, 1 ignored pre-existing). The new tests added in `schema.rs` cover:

- Round-trip: create → set command → set transport → get command → verify.
- Read-only natural key: `set_prop("mcp.servers.fs.name", ...)` refuses with actionable error.
- Rename: old key stops resolving, new key resolves correctly.
- Ambiguity: duplicate `name` causes `set_prop` / `get_prop` / `rename_map_key` to refuse without mutating.
- Rename collision: refuses when target key already exists.
- `prop_fields()`: surfaces per-element child props, hides the natural-key field itself.
- `get_map_keys()`: returns natural keys in insertion order.

The helper function `route_vec_path` has 8 unit tests in `helpers::tests`:

- Longest-match on dotted natural keys (so `acme.tools` disambiguates from `acme`).
- Empty natural-key slots skipped (so half-created entries don't soak up every unmatched path).
- Bare element path treated as miss (so `mcp.servers.fs` without a trailing field doesn't route through `get_prop`).
- Ambiguous duplicates reported explicitly.
- Unknown natural key returns `Miss`.
- Deep inner paths forwarded correctly (e.g. `mcp.servers.<name>.headers.<key>`).

This is thorough: the new code paths have equivalent coverage to the existing `route_hashmap_path` / `route_double_hashmap_path` helpers, and the schema-side tests anchor the end-to-end contracts (rename, ambiguity, collision, prop enumeration) that the TUI and dashboard will rely on.

---

### 🟢 What looks good — Documentation is actionable and scoped

`docs/dashboard-mcp-per-field-todo.md` is a well-structured handoff document: it explains **why** the dashboard's current `ObjectArrayEditor` is no longer necessary, provides a three-step migration sketch with concrete RPC paths, surfaces the new error strings that need toast/banner handling, and lists the seven other `Vec<T>` types that can opt into the same surface in future PRs.

The document correctly scopes the follow-up work: dashboard migration is separate from this PR (the dashboard still works via the legacy JSON-array path; `set_prop("mcp.servers", "<json>")` routes through the existing `ObjectArray` arm), and opting the other seven types into `#[natural_key]` is gated on their respective dashboard readiness. This is the right layering: ship the runtime capability first with one concrete use case fully validated, document the migration path for the consumers, let them coordinate their own cutover.

The inline commit messages and code comments are precise and reference the relevant helpers / validators / test names. The schema.rs docstring update on `McpConfig::servers` explains the `#[natural_key]` opt-in and the read-only natural-key / rename-via-map-key contract at the declaration site, which is the right place for that context (developers adding similar fields will read the existing examples).

---

### 🟡 Warning — Error propagation change needs explicit mention in CHANGELOG

The commit "fix(macros): propagate inner errors through nested-prop & map-key arms" changes the behavior of **all** `Configurable` derives: nested `set_prop` / `get_prop` and the map-key recurse arms now propagate inner errors instead of silently swallowing them. The PR body calls this out explicitly and justifies it (previously-hidden actionable errors like typos now surface as `Unknown property` pointing at the right path), and the test suite passes with zero regressions (no test relied on the swallowed-error behavior).

This is a **correct** fix — surfacing actionable errors is better than eating them — but it is a genuine behavioral change that touches every nested struct. The PR body documents it well; the concern is whether the release notes / CHANGELOG will capture it clearly enough for operators debugging config-edit failures post-upgrade.

**Recommendation:** when this PR lands, ensure the CHANGELOG entry for the next release mentions this behavior change explicitly under a "Fixed" or "Improved" heading. Suggested wording:

> **Config error reporting improved:** Property access errors from nested structs and map-keyed sections now propagate to the caller instead of being silently swallowed. Typos in property paths will now surface as actionable `Unknown property` errors pointing at the exact path. Previously some such errors were hidden by fallthrough logic. No known user-visible breakage, but error messages will be more specific.

This is not a blocker for merge (the change is correct and the tests prove no regressions), but it should be called out in release notes so maintainers and operators understand the new error surface.

---

### 🟡 Warning — TUI smoke test skipped due to environment constraints

The validation evidence lists "Not manually verified: an interactive TUI smoke walk — I cannot run a TTY from this environment." The PR body justifies the skip by noting that the TUI code path (`config_sections` → `enter_section` → `load_aliases` → `enter_alias` → `load_fields` → per-field `set_prop`) is identical to the path other `OneTierAliasMap` sections drive successfully today, and every link in that chain is covered by an automated test.

This justification is **sound** — the TUI dispatches `mcp.servers` through the same generic `OneTierAliasMap` rendering that `risk_profiles`, `cron`, etc. use, and those sections work — but the lack of an end-to-end TUI smoke test means any TUI-specific rendering edge case (e.g. how the `<unnamed-N>` placeholder for empty-name entries renders, or whether the TUI's delete keybind correctly calls `delete_map_key`) won't surface until an operator runs the TUI against a config with MCP servers.

**Recommendation:** The first reviewer or maintainer with TUI access should smoke-test the following before merge or immediately post-merge:

1. Open zerocode TUI, navigate to `mcp.servers` section.
2. `+ Add` a server, verify it shows in the alias list (name field likely defaults to empty; should show as `<unnamed-0>` per the macro's prop_fields arm).
3. Enter the alias, set `name`, `transport`, `command`, verify fields round-trip.
4. Exit to alias list, verify the server now shows its real name instead of `<unnamed-0>`.
5. Delete the server, verify it vanishes from the list.

If any of these steps fail, the failure is TUI-layer (not runtime), and can be fixed in a follow-up without reverting this PR.

This is **not** a blocker: the automated test coverage is comprehensive, and the zerocode TUI is not blocking release (operators can still hand-edit `config.toml` or use the dashboard). But the TUI path should be verified by someone with TTY access before release notes claim "TUI support for mcp.servers is live."

---

### 🟢 What looks good — Rollback plan is concrete and tested

The rollback section provides a specific `git revert` command for the seven commits in this PR, documents that the previous docs commit (`445de193e`) is unaffected and stays, and explains the structural gating: after revert, `set_prop("mcp.servers.<name>.<field>", ...)` stops working but the dashboard's JSON-blob editor and hand-editing `config.toml` continue unchanged.

The observable failure symptoms are actionable:

- Grep for `Unknown property 'mcp.servers.` → natural-key arm not routing.
- Grep for `is ambiguous in 'mcp.servers'` → duplicate detection misclassifying.
- Grep for `no map-keyed/list section at 'mcp.servers'` → `Section::McpServers` removed without reverting the natural-key arm.

This is a well-structured rollback plan: the revert command is exact, the failure symptoms are grep-able, and the structural fallback (dashboard JSON-array editor still works) means a rollback doesn't strand operators in a broken state.

---

### 🔵 Suggestion — Consider explicit validation for natural-key field type

The macro emits natural-key routing arms without checking that the natural-key field is actually a `String` (or `&str`). If a developer annotates a `Vec<T>` field with `#[natural_key = "count"]` where `count: usize`, the emitted code will call `e.get_prop(&nk_path)` which returns `Result<String>`, but the field's actual `get_prop` impl will return the `usize` formatted as a string. This will "work" but produce confusing natural keys ("0", "1", "42") that are indistinguishable when multiple elements share the same numeric value.

The macro could add a compile-time check: parse the natural-key field's type from the struct definition, verify it is `String` / `&str` / `Option<String>`, emit a compile error otherwise. This would catch the mistake at derive time instead of surfacing as confusing runtime behavior.

**Not a blocker** — the canonical use case (`McpServerConfig::name`) is correctly typed as `String`, and the docs/examples will guide future adopters to do the same. But adding the type check would make the contract explicit at compile time.

---

### 🔵 Suggestion — Future: consider natural-key uniqueness validation at schema level

The PR correctly handles duplicate natural keys at runtime (ambiguity errors, refuse to mutate), and `validate_mcp_config` already rejects duplicate `name` at save time. But the `#[natural_key]` attribute itself doesn't enforce uniqueness — a developer could add `#[natural_key = "category"]` on a `Vec<T>` where `category` is legitimately non-unique (e.g. classification rules grouped by category).

This isn't a problem for `mcp.servers` (server names are inherently unique identifiers), but if the natural-key pattern spreads to other `Vec<T>` fields, developers might misuse it on fields that aren't actually unique keys. The runtime would then surface ambiguity errors on every access to such a section.

**Future consideration:** the schema validator generator could introspect the `#[natural_key]` attribute and automatically emit a uniqueness check for that field (similar to how `validate_mcp_config` checks `servers[*].name` uniqueness today). This would make the "natural key must be unique" invariant automatic instead of requiring hand-written validators for every such field.

**Not a blocker** — `mcp.servers` has the hand-written validator, and the migration TODO doc lists the seven other types that would opt in (all of which have `name` fields that are semantically unique). But auto-generating the uniqueness check would reduce the maintenance burden when more types adopt `#[natural_key]`.

---

### 🟢 What looks good — Parent-collapse fix is scoped and correct

The commit "fix(rpc): keep parent sections with direct scalars beside dotted children" refines the `handle_config_sections` parent-collapse rule so `mcp` does not disappear when `mcp.servers` is added to the curated section list. The fix is scoped: `providers` still vanishes correctly because it has no direct scalar fields, while `mcp` stays because `enabled` and `deferred_loading` are direct scalars on `Mcp`.

This is the correct logic: a parent section should stay visible if it has direct scalar fields (toggles, enums, strings) that the operator needs to edit, even when it also has child sections (like `mcp.servers`, `mcp_bundles`) exposed separately. The alternative (collapsing `mcp` into its children) would hide the `enabled` / `deferred_loading` toggles, forcing operators to drill into the TOML by hand to set them.

The change is covered by `mcp_servers_section_has_alias_map_shape_and_parent_keeps_direct_form`, which verifies that both `Section::Mcp` and `Section::McpServers` appear in `QUICKSTART_SECTIONS` and that `Mcp` keeps `DirectForm` shape while `McpServers` is `OneTierAliasMap`. Canonical order test added ensures `Mcp` → `McpServers` → `McpBundles` so operators hit the parent toggles before the per-server form.

---

### ✅ Summary

**Verdict: APPROVE.** This PR is ready to merge.

**What this PR does well:**

- Single-source-of-truth discipline: the natural-key field is resolved on-demand every time, zero caching, zero drift bugs.
- Opt-in gating: only `mcp.servers` changes behavior; the other seven `Vec<T>` types keep legacy behavior unchanged.
- Explicit ambiguity handling: refuses to mutate on duplicate keys with actionable error strings.
- Comprehensive test coverage: 2,939 tests passing, 8 new unit tests for `route_vec_path`, 5 new integration tests for `mcp.servers` end-to-end round-trip.
- Actionable documentation: dashboard migration TODO is a clear handoff, inline comments reference the right helpers/validators.
- Concrete rollback plan: exact `git revert` command, actionable failure symptoms, structural fallback (dashboard JSON-array editor still works).

**What needs follow-up (non-blocking):**

- **🟡 CHANGELOG entry** for the error-propagation behavior change (nested structs now surface actionable errors instead of swallowing them).
- **🟡 TUI smoke test** by the first reviewer/maintainer with TTY access (verify `+ Add`, per-field editing, delete keybind).
- **🔵 Optional future:** compile-time type check on natural-key field (must be `String` / `&str`).
- **🔵 Optional future:** auto-generate uniqueness validator for `#[natural_key]` fields.

None of these block merge. The error-propagation CHANGELOG entry should land when the PR is merged (or in the release-notes pass). The TUI smoke test can happen post-merge. The two optional future enhancements are architecture improvements for when more types adopt `#[natural_key]`.

**CI status:** All 14 checks passing (format, lint, security, test, benchmarks compile, 3 platform builds).

**Risk classification:** Auto-label tagged this `risk: high` (touches `zeroclaw-runtime` and `zeroclaw-gateway`). Actual risk is **medium**: the blast radius is surgically scoped to one schema field via opt-in gating, the 2,939 passing tests prove zero regressions, and the rollback plan is concrete. The `risk: high` label is correct by policy (crate-level boundaries), but the engineering discipline here mitigates that risk effectively.

Audacity88 (changes requested):
Reviewing at 843959fc8541bc3a20b172453446fdba531aee39.

I checked the live PR state, the full diff against `master`, the prior approval review from @WareWolf-MoonWall, and the generated macro/runtime paths for `mcp.servers` map-key creation, per-field routing, rename, delete, and section selection. There are no inline threads, and I am not claiming a local test result here. The natural-key direction looks right, but I found one create-path bug that can put the new editor into the duplicate-key state the rest of the PR correctly treats as invalid.

### 🟢 What looks good — Natural-key routing is the right shape for this section

Using `McpServerConfig::name` as the live natural key keeps the editable address tied to the TOML data instead of adding a second alias store. I especially like that `set_prop("mcp.servers.<name>.name", ...)` is rejected and rename goes through `rename_map_key`; that avoids changing the route key in the middle of a form edit. The longest-match routing and duplicate-name ambiguity errors are also the right safety behavior for hand-edited broken configs.

### 🔴 Blocking — Duplicate create can generate the ambiguous state this PR is trying to avoid

The generated `Vec<T>` `create_map_key` arm always appends a new default element before it seeds `name` / `hint` from the requested key. It does not first check whether the natural key already exists. For `mcp.servers`, that means two calls to `create_map_key("mcp.servers", "fs")` create two `McpServerConfig` entries with `name = "fs"` instead of returning idempotently like the map-backed sections do.

That breaks the new editor contract. The rest of this PR correctly treats duplicate natural keys as invalid: per-field `get_prop` / `set_prop` / `rename_map_key` report `natural key ... is ambiguous`, and validation rejects duplicate server names on save. So a repeated add, a retry after an uncertain RPC response, or a UI path that re-selects an existing alias can create a config state where `mcp.servers.fs.command` no longer routes and the operator has to repair the duplicate manually.

Please make natural-key `Vec<T>` creation check the existing keys before pushing a new element. Returning `Ok(false)` for an existing key would match the existing map-section contract; returning a clear duplicate-key error would also be much safer than appending. Please also add a regression for `mcp.servers` that calls `create_map_key("mcp.servers", "fs")` twice and proves the Vec length remains one, the second result is the intended no-op/error, and the original `mcp.servers.fs.*` paths still route.

Once that is fixed, I expect this to be in good shape.

tidux:
Thanks for the careful reads, @Audacity88 and @WareWolf-MoonWall. Pushed two follow-up commits addressing the feedback (now at `ffb599902`):

---

### @Audacity88 — 🔴 Blocking: duplicate create on natural-key Vec

You're right, and the failure mode you described is exactly the trap. Fixed in `33e96180b` (`fix(macros): make natural-key Vec create_map_key idempotent`).

**Fix:** The generated `Vec<T>` `create_map_key` arm now branches on whether the field carries `#[natural_key]`:

- **With `#[natural_key]`** (today: only `McpConfig::servers`): walk the live Vec at create time, ask each element for its natural-key field via the inner type's own `get_prop`, and return `Ok(false)` if any element already carries the requested key. This matches the existing `HashMap<String, T>` arm's idempotent contract (`contains_key → Ok(false)`) and keeps the per-element editor surface out of the `VecRoute::Ambiguous` state that the rest of the PR correctly refuses to mutate from.
- **Without `#[natural_key]`** (the other seven `Vec<T> + #[nested]` fields — `ClassificationRule`, `EmbeddingRouteConfig`, `GoogleWorkspaceAllowedOperation`, `ModelRouteConfig`, `NevisRoleMappingConfig`, `PeripheralBoardConfig`, `ToolFilterGroup`): legacy always-append behaviour, unchanged. There's no natural key to dedupe against and their editor surface is the whole-array `ObjectArray` `set_prop` path, not per-element.

I went with `Ok(false)` rather than a hard error because that's the contract the other map-keyed sections already advertise, and a UI retry against an existing key should be a no-op (not a surfaced error the operator has to interpret).

**Regression:** `mcp_servers_create_map_key_is_idempotent_on_existing_natural_key` covers:

1. Two calls to `create_map_key("mcp.servers", "fs")` — second reports `Ok(false)`, Vec length stays at 1.
2. The existing entry's seeded `command` field survives the no-op second add (proves the duplicate-check fires before the push, not after).
3. `mcp.servers.fs.command` still routes via `get_prop` and `set_prop` after the no-op call (i.e. the editor surface is not in the ambiguous state).
4. A distinct natural key (`"github"`) still gets created normally — idempotency is per-key, not first-add-wins-everything.

The pre-existing `mcp_servers_routing_is_ambiguous_on_duplicate_names` test stays valid: it pushes duplicates directly into the Vec (simulating hand-edited broken TOML), bypassing `create_map_key`, so it still anchors the in-flight refusal contract for configs that arrive broken from outside the editor.

---

### @WareWolf-MoonWall — 🟡 CHANGELOG entry for the error-propagation behavioural change

Added in `ffb599902` (`docs(changelog): note per-field mcp.servers editor + error-propagation fix`). Two entries land in `CHANGELOG-next.md`:

1. A bullet under **What's New since v0.8.0-beta-1** for the per-field `mcp.servers` editor as a new capability — calls out explicitly that hand-editing `config.toml` and the dashboard's JSON-array editor still work unchanged, so operators on those workflows don't need to do anything.
2. A row in the **Bug Fixes** table for the error-propagation change, using the wording you suggested (lightly adapted to match the table's compact style): property-access errors from nested structs and map-keyed sections now propagate instead of being silently swallowed, typos surface as actionable `Unknown property` errors pointing at the exact path, no on-disk or wire-format changes.

### @WareWolf-MoonWall — 🟡 TUI smoke test

Still environment-blocked (no TTY here, same as the original PR body). Leaving this for the first maintainer with TUI access — the five-step walk you outlined (Add → name → fields → return to alias list shows real name → delete) is the right minimum. If any step fails the failure is TUI-layer (not runtime) and can be fixed in a follow-up without reverting this PR.

### @WareWolf-MoonWall — 🔵 Optional futures (compile-time type check on natural-key field; auto-generate uniqueness validator)

Agreed both are worth doing when more types adopt `#[natural_key]`. Filing them as separate follow-up issues rather than in this PR — the canonical first user (`McpServerConfig::name`) is correctly typed as `String` and has its hand-written validator, so neither blocks adoption today, and bundling them here would expand the surface beyond what's reviewable in one pass.

---

### Validation re-run

`cargo fmt --all -- --check`: clean.

`cargo clippy -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-macros -p zeroclaw-gateway -p zerocode --all-targets -- -D warnings`: clean.

`cargo test` per crate:

```
zeroclaw-config:  735 passed (was 734; +1 = new regression test)
zeroclaw-runtime: 1988 passed, 1 ignored (pre-existing)
zeroclaw-gateway:  201 passed
zeroclaw-macros:     6 passed
zerocode:           14 passed
```

Totals: **2,944 passed, 0 failed, 1 ignored.** No regressions; the new test exercises the fix end-to-end through `Config` (`create_map_key` → `set_prop` → `create_map_key` again → `get_prop`/`set_prop` round-trip → distinct-key add).

The fast-rollback section in the PR body is unchanged in shape, except the revert list now includes `33e96180b` and `ffb599902` at the top. The structural gating still holds: a revert turns off the natural-key path; the dashboard JSON-array editor and hand-edited `config.toml` keep working.

Ready for re-review.

Audacity88 (approved):
Reviewing current head a9b2d0b1caa25e047d73a06e124647b0c9b2e935 after @tidux's follow-up. I checked the live PR state, top-level discussion, formal reviews, inline review comments, current changed files, the natural-key Vec create/routing source, the new regression test, the CHANGELOG update, and the visible GitHub checks. There are no inline review comments, and I did not run a local cargo or interactive TUI smoke pass. The PR is still blocked by my prior `CHANGES_REQUESTED` review; a merge-tree check against current `master` applies cleanly, but current `master` has moved and touched `crates/zeroclaw-config/src/schema.rs`.

### ✅ Resolved — Duplicate natural-key create is now idempotent

The prior blocker is fixed. The generated `#[natural_key]` Vec `create_map_key` arm now checks the live Vec before appending, resolves each element's natural-key field through the inner type's own `get_prop`, and returns `Ok(false)` when the requested key already exists. That matches the existing map-section contract and prevents a retry or repeated add from creating two `mcp.servers` entries with the same `name`.

The new `mcp_servers_create_map_key_is_idempotent_on_existing_natural_key` regression covers the failure mode I asked for: add `fs`, set `mcp.servers.fs.command`, add `fs` again, assert the second add reports false, assert the Vec length stays one, assert the original command survives, assert `get_prop` / `set_prop` still route through `mcp.servers.fs.*`, and assert a distinct key such as `github` still creates a second entry. That proves the duplicate check runs before the push and does not collapse all future creates into a no-op.

### ✅ Resolved — CHANGELOG now covers the error-propagation behavior change

@WareWolf-MoonWall's release-note warning is also addressed. `CHANGELOG-next.md` now records the per-field MCP editor capability and separately calls out the nested/map-key property-error propagation change, including the user-visible effect: typos now surface as specific `Unknown property` errors instead of being swallowed by fallthrough logic. That is the right place for operators to learn why config-edit errors may become more precise after upgrade.

### 🟢 What looks good — The natural-key surface keeps one source of truth

The core shape still looks right: `McpServerConfig::name` remains the source of truth for each server identity, property routing resolves against the live Vec on demand, direct writes to `mcp.servers.<name>.name` are rejected, and renames go through `rename_map_key`. The ambiguity path is also still conservative: hand-edited duplicate names make `get_prop` / `set_prop` / `rename_map_key` refuse rather than silently editing one duplicate.

### 🟡 Warning — Refresh merge evidence before landing

This approval clears my previous blocker on the reviewed head. It is not a merge-readiness sign-off for the final integration result. Current `master` has moved since this branch's base and has touched `crates/zeroclaw-config/src/schema.rs`, which is one of this PR's central files. The merge-tree applies cleanly, and the visible PR-head checks are green, but before merge this still needs refreshed official CI on an updated head or an equivalent merge-result smoke that covers the config/schema path.

### 🟡 Warning — TUI smoke remains the remaining product-side gap

The runtime and generic `OneTierAliasMap` path are well covered, and the skipped interactive TUI smoke is understandable. The first maintainer with TTY access should still run the add/edit/delete walk before release notes claim the TUI path is fully verified. I am not blocking on that here because the operator fallback remains hand-edited `config.toml` or the dashboard's existing JSON-array editor, and the automated tests cover the config/RPC behavior this PR changes.

Approving to clear my previous `CHANGES_REQUESTED` review.

Audacity88:
I re-ran the local smoke on current head `a9b2d0b1caa25e047d73a06e124647b0c9b2e935`, and this looks good from my side.

Validation:
- `zc-cargo build -p zerocode -p zeroclawlabs` passed.
- `zc-cargo test -p zeroclaw-config mcp_servers` passed: 6 matching tests green.
- Live `zerocode` smoke in a disposable config dir passed:
  - opened `Config -> zeroclaw -> Mcp.servers`
  - added a `smoke` server alias
  - opened the per-entry fields
  - set `command = /bin/echo`
  - backed out and reopened the alias; the value persisted
  - deleted the alias; it disappeared cleanly

That covers the TUI/config round-trip gap I wanted to check.

PR #7274, opened 2026-06-05
fix(config): stamp schema_version on incremental saves

### LINKED ISSUE #7271: [Bug]: Config writer leaves stale schema_version on partial saves → V3-bodied/schema-2 configs crash older binaries with opaque "missing field provider"
### Affected component

runtime/daemon (config load path at daemon/agent startup)

### Severity

S1 - workflow blocked (daemon and agent both refuse to start)

### Current behavior

A `~/.zeroclaw/config.toml` that has been written by a V3 (schema-3) build but still carries `schema_version = 2` causes older binaries (v0.7.5, the version served by `zeroclawlabs.ai/install.sh`) to fail at startup with an opaque serde error:

```
Error: Failed to deserialize config file

Caused by:
    TOML parse error at line 1, column 1
      |
    1 | schema_version = 2
      | ^
    missing field `provider`
```

Two distinct defects combine here:

1. **Stale `schema_version` on partial saves.** Config writes go through `comment_writer.rs:34` — targeted in-place `toml_edit` edits (`doc.to_string()`), not a full re-serialize. When the V3 dashboard/onboarding writes V3-shaped sections (`[agents.<name>].model_provider`, `[providers.models.*]`), it edits those keys in place and never re-stamps `schema_version`. The label stays at whatever an older binary first wrote (`2`), while the body is V3.

2. **No "schema newer than supported" guard in the released load path.** v0.7.5's loader (`V1Compat` → `toml::from_str`, `schema.rs:10244`) deserializes directly. The V3 `[agents.default]` shape (`model_provider = "zai.default"`) maps onto v0.7.5's `DelegateAgentConfig`, whose `provider: String` is **required** (`schema.rs:658-663`; `Config.agents: HashMap<String, DelegateAgentConfig>` at `schema.rs:331`) → `missing field provider`, reported against document root (line 1). master added this guard (`migration.rs:76-90`) but the release line lacks it.

### Expected behavior

- Config writes should always re-stamp `schema_version` to the current value, so a save can never produce a body-newer-than-label file.
- When a binary encounters a config newer than it supports, it should fail with an actionable message (e.g. *"config is schema_version 3, newer than this binary supports (2); upgrade ZeroClaw"*) instead of `missing field provider`.

### Steps to reproduce

```bash
# On a machine where a V3 (schema-3) build has saved provider/agent settings
# (e.g. via the web dashboard onboarding), producing a config whose body is V3
# but whose top line still reads `schema_version = 2`:

curl -fsSL https://zeroclawlabs.ai/install.sh | bash   # installs v0.7.5
zeroclaw daemon
# -> Error: Failed to deserialize config file ... missing field `provider`
zeroclaw agent
# -> same error

# The offending section in config.toml:
#   [agents.default]
#   model_provider = "zai.default"     # V3 shape; v0.7.5 expects `provider` + `model`
```

### Impact

```
Affected users: developers/testers who have run a V3 (schema-3) build and then
                install/downgrade to the v0.7.5 release served by install.sh.
                NOT fresh installs — a new user with no config runs quickstart,
                gets a clean schema-2 file, and works fine.
Frequency: always, for any machine in the V3-config + old-binary state.
Consequence: daemon and one-off agent both fail to start; opaque error gives
             no hint that the cause is a config/binary version mismatch.
```

### Logs / stack traces

```
Installing ZeroClaw v0.7.5 (pre-built)
  ✓ Platform: aarch64-apple-darwin
  ...
  ✓ Existing ZeroClaw config detected at /Users/<redacted>/.zeroclaw/config.toml — skipping setup prompt.

$ zeroclaw daemon
Error: Failed to deserialize config file

Caused by:
    TOML parse error at line 1, column 1
      |
    1 | schema_version = 2
      | ^
    missing field `provider`
```

### ZeroClaw version

Failing binary: `v0.7.5` (schema_version 2). Repo at time of report: `0.8.0-beta-2` / `42668ef5e` (`CURRENT_SCHEMA_VERSION = 3`).

### Rust version

### Operating system

macOS 26.3.1 (Darwin 25.3.0, arm64)

### Regression?

Yes, it worked before (the V3 build loads this config fine; the older release does not)

### Pre-flight checks

theonlyhennygod:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Incremental saves (`save_dirty`, used by the web dashboard / gateway config CRUD) apply only dirty paths to the existing TOML doc and never touch `schema_version`. When the dashboard writes current-schema-shaped sections (e.g. `[agents.<name>].model_provider`, `[providers.models.*]`) into a file an older binary first wrote, the body advances to the current schema while the `schema_version` label stays stale.
  - The resulting body-newer-than-label config crashes older binaries (e.g. the v0.7.5 release served by `install.sh`) with an opaque `missing field provider` serde error: v0.7.5 reads the V3 `[agents.*]` shape as its `DelegateAgentConfig`, whose `provider` field is required.
  - Stamp `CURRENT_SCHEMA_VERSION` into the doc on every incremental save so the label always matches the body; `insert` updates the key in place (preserving its top-of-file position) or appends it when absent.
  - Defensively pin `schema_version` in the full `save` path too, and add a regression test.
- **Scope boundary:** Does not change migration logic, the schema shape, or the set of dirty paths. Only ensures the `schema_version` label is written on save.
- **Blast radius:** Touches the config write paths (`save` / `save_dirty`) in `zeroclaw-config`, exercised by the CLI, gateway, and web dashboard config CRUD. No behavioral change beyond the `schema_version` key now always being stamped.
- **Linked issue(s):** Fixes #7271
- **Labels:** `bug`, `risk: medium`, `size: XS`, `config`, `config: core`.

## Validation Evidence

```bash
cargo fmt -p zeroclaw-config -- --check          # clean, exit 0
cargo clippy -p zeroclaw-config                  # Finished, 0 warnings
cargo test -p zeroclaw-config                    # 722 passed; 0 failed (+ 89 migration tests passed)
```

- New test `schema::tests::save_dirty_stamps_current_schema_version_on_stale_label ... ok`.
- **Beyond CI:** The regression test seeds an on-disk config labeled `schema_version = 2`, performs an incremental `save_dirty`, and asserts the stale label is overwritten with `CURRENT_SCHEMA_VERSION`, the dirty value still lands, and `schema_version` remains the first key.

## Security & Privacy Impact

## Compatibility

- Backward compatible? **Yes** — this fixes a forward/backward-compat hazard; correctly stamping `schema_version` lets older binaries detect newer files and fail cleanly via the migration guard rather than an opaque serde error.
- Config / env / CLI surface changed? **No** — only the `schema_version` key value written to disk.
- Upgrade steps: None. Existing configs are corrected on next save.

## Rollback

- **Fast rollback:** `git revert <merge sha>`
- **Observable failure symptoms:** Older binaries crashing on config load with `missing field provider` (or similar `missing field ...`) serde errors; a config file whose `schema_version` label is lower than its body shape.


Audacity88 (approved):
I reviewed the live PR state, linked issue #7271, current checks, review/comment snapshots, and the one-file config diff at head 812603abfd44d59e1a658fc19faa75ac73f9921e.

### 🟢 What looks good — schema version stamping closes the partial-save hole

This is the right-sized fix for #7271. The incremental save path can now write current-schema sections without leaving an older `schema_version` label behind, and the full save path also pins the in-memory clone before serialization. That keeps the label aligned with the body and avoids producing a misleading body-newer-than-label config.

### 🟢 What looks good — regression coverage targets the failure mode

The new `save_dirty_stamps_current_schema_version_on_stale_label` test covers the important edge case directly: an existing file labeled with an older schema version, an incremental dirty-path write, the dirty value still landing, the stale label being overwritten, and the schema key staying at the top of the file. That is the behavior this bug needed protected.

I do not see anything blocking here.

PR #7283, opened 2026-06-06
feat(zerocode): two-level pane cursor, agent-picker mouse, registry-driven config help

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The zerocode Config tab used Left/Right to move the left section list and Up/Down to move the right detail, giving no sense of which nested level the cursor was at. Introduce a two-level `PaneCursor {Sections, Detail}`: Up/Down move within the focused side, Right/Enter steps into the detail, Left/Back/Esc walks home, and edges are no-ops (tab switching stays on the global PaneNav chord). The inactive side keeps a dimmed "you are here" highlight so the user never loses their place.
  - All navigation matches on `ConfigTabAction` variants, never key literals, so vim (h/j/k/l), arrows, emacs chords, and user overrides all work. Help context and footers reflect the live cursor state with registry-resolved chords.
  - The Chat agent picker gains mouse support: click highlights a row, double-click confirms (enters the session), the wheel moves the selection. `handle_mouse` becomes async to route a double-click confirm through `pick_or_start_session`.
  - The zeroclaw Config tab drills into a section with Right (alias of Enter), and every daemon Config footer and help key now resolves from the keymap registry instead of hardcoded glyphs.
- **Scope boundary:** Does NOT change the zeroclaw Config tab's deeper drill-down screens (Model Providers, Agents, Channels — type/alias/field lists). Those keep their existing Enter-to-drill / Esc-to-walk-home behavior; no split-pane cursor is applied there (that surface is a full-screen drill-down, not a left/right split). No daemon/runtime behavior changes.
- **Blast radius:** zerocode TUI only (`apps/zerocode/**`). `HelpEntry` now owns `Vec<String>` keys instead of `Vec<&'static str>`; all in-crate callers updated.
- **Linked issue(s):** None.
- **Labels:** `enhancement` (size/risk auto-applied).

Deferring the full battery to CI. Locally during development the zerocode bin built clean, `cargo test -p zerocode --bins` passed (207 tests), `cargo fmt --all -- --check` was clean, and `cargo clippy -p zerocode --all-targets -- -D warnings` was clean.

- **Commands run and tail output:** see CI.
- **Beyond CI — what did you manually verify?** Cursor transitions (enter/leave detail, Esc walk-home, section nav), active vs inactive highlight, agent-picker click selection — covered by new unit tests in `zerocode_pane` and `chat`.
- **If any command was intentionally skipped, why:** Full local battery deferred to CI per maintainer direction.

## Compatibility (required)

- Backward compatible? Yes
- Upgrade steps: None. Key bindings remain registry-driven; existing overrides continue to work.

- `git revert <sha>` is the rollback plan. No feature flags or config toggles; TUI-only change.

### DIFF
--- a/apps/zerocode/locales/en/zerocode.ftl
+++ b/apps/zerocode/locales/en/zerocode.ftl
@@ -64,6 +64,10 @@ zc-zerocode-help-rebind = Rebind selected action
 zc-zerocode-help-reset-default = Reset action to default
 zc-zerocode-help-mouse-label = Mouse
 zc-zerocode-help-mouse-desc = Click pane / row, scroll, click section tab
+zc-zerocode-help-choose-section = Choose section
+zc-zerocode-help-open-section = Open section
+zc-zerocode-help-navigate-rows = Navigate rows
+zc-zerocode-help-back-to-sections = Back to sections
 
 zc-input-no-pending-attachments = No pending attachments.
 zc-input-no-clipboard-image = Clipboard is empty.
--- a/apps/zerocode/src/acp.rs
+++ b/apps/zerocode/src/acp.rs
@@ -42,8 +42,8 @@ impl Acp {
         self.inner.wants_text_input()
     }
 
-    pub(crate) fn handle_mouse(&mut self, mouse: MouseEvent, area: Rect) {
-        self.inner.handle_mouse(mouse, area);
+    pub(crate) async fn handle_mouse(&mut self, mouse: MouseEvent, area: Rect) {
+        self.inner.handle_mouse(mouse, area).await;
     }
 
     pub(crate) fn handle_paste(&mut self, text: &str) {
--- a/apps/zerocode/src/app.rs
+++ b/apps/zerocode/src/app.rs
@@ -440,10 +440,10 @@ pub async fn run(
                             logs_pane.handle_mouse(mouse, content_area);
                         }
                         Mode::Acp => {
-                            acp_pane.handle_mouse(mouse, content_area);
+                            acp_pane.handle_mouse(mouse, content_area).await;
                         }
                         Mode::Chat => {
-                            chat_pane.handle_mouse(mouse, content_area);
+                            chat_pane.handle_mouse(mouse, content_area).await;
                         }
                         Mode::Quickstart => {
                             quickstart.handle_mouse(mouse, content_area).await;
--- a/apps/zerocode/src/chat.rs
+++ b/apps/zerocode/src/chat.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use crossterm::event::{KeyEvent, MouseEvent, MouseEventKind};
+use crossterm::event::{KeyEvent, MouseButton, MouseEvent, MouseEventKind};
 use pulldown_cmark::{Event as MdEvent, Options as MdOptions, Parser as MdParser, Tag, TagEnd};
 use ratatui::{
     Frame,
@@ -90,6 +90,12 @@ pub(crate) struct Chat {
     git_branch_inflight: bool,
     phase: ChatPhase,
     pane_kind: PaneKind,
+    /// List rect of the agent picker, recorded each draw so mouse clicks in the
+    /// PickAgent phase can map a row to a selection. Default until first draw.
+    pick_agent_list_area: Rect,
+    /// Double-click tracker for the agent picker: a second click on the same row
+    /// confirms (enters the session), matching the keyboard Enter.
+    pick_agent_double_click: crate::mouse::DoubleClickTracker,
 }
 
 fn should_retry_on_entry(phase: &ChatPhase) -> bool {
@@ -112,6 +118,8 @@ impl Chat {
                 loading: true,
             },
             pane_kind,
+            pick_agent_list_area: Rect::default(),
+            pick_agent_double_click: crate::mouse::DoubleClickTracker::new(),
         }
     }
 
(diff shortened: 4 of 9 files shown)


theonlyhennygod (approved):
I read the full diff and cross-checked it against the zerocode keymap, mouse, theme, and widgets modules. This is a focused, well-scoped TUI change: a two-level `PaneCursor` for the zerocode Config tab, mouse support for the Chat agent picker, and a sweep that resolves every footer/help chord from the live keymap instead of hardcoded glyphs. Scope stays inside `apps/zerocode/**`, CI is fully green (14/14 Quality Gate jobs), and the new behavior is covered by unit tests. No other reviewer holds an active block. Approving — the items below are all optional.

### 🟢 What looks good — the two-level cursor is the right model

`PaneCursor {Sections, Detail}` is exactly the abstraction the old "Left/Right moves the list, Up/Down moves the detail" scheme was missing — it gives the user a single mental model of *where am I and how do I go deeper / come back*. The details that make this land:

- Up/Down operate on whichever side holds the cursor, Right/Enter steps in, Left/Back/Esc walks home, and edges are no-ops so you never accidentally fall out of the tab (cross-tab nav correctly stays on the global PaneNav chord).
- The inactive side keeps a dimmed "you are here" highlight (`selected_inactive_style` — no bold, dim fg) so the prior position is never lost. That distinction between live cursor and remembered position is a genuinely thoughtful UX call.
- `DeleteRow` (reset binding) is now gated on `cursor == Detail && focus == Bindings`, which is correct — you can only reset the row you're actually pointed at.

### 🟢 What looks good — chords resolve from the registry, never literals

Every footer and help row now goes through `tab_key` / `tab_keys` / `editor_key` / `nav_keys` against `RebindableActions::resolved()`, and the navigation handlers match on `ConfigTabAction` variants rather than `KeyCode` literals. That means vim (h/j/k/l), arrows, emacs chords, and user overrides all render and behave correctly with no parallel glyph table to drift out of sync — the single-source-of-truth principle applied to the help surface. The `picker_nav_keys()` helper deduping chords before joining is a nice touch.

### 🟢 What looks good — test coverage tracks the new behavior

The new tests assert the behavior, not the implementation: `right_enters_detail_left_returns_to_sections` exercises the full step-in/walk-home cycle including the home no-op, `esc_walks_back_to_sections` covers the Esc path, `up_down_navigate_sections_when_cursor_in_sections` pins the side-aware nav, and `agent_picker_click_selects_row` validates the row-mapping math. Updating the existing `locale_tab` tests to drive `Down` through the section list (with the second `Enter` to step in) keeps them honest against the new model.

### 🔵 Suggestion — the agent-picker rect-shift adapter is clever but fragile

In `draw_agent_picker` the returned rect is shifted up one row and grown one taller so that `mouse::list_click_index` (which assumes a 1-cell top border) lands on the true first item of the *unbordered* list:

```rust
Rect::new(chunks[1].x, chunks[1].y.saturating_sub(1), chunks[1].width, chunks[1].height + 1)
```

The math is right for the first item, and the test documents the reasoning well. Two things worth weighing:

- It costs the bottom-most visible row: with the shift, the clickable range becomes `[y .. y+h-1)`, so the last visible row of the list area maps to no item. That's harmless when the area is taller than the agent count (the dead row sits below the last item), but a fully-populated list would have its bottom row un-clickable until scrolled.
- Encoding a "compensate for a border that isn't there" offset at the call site is the kind of thing that surprises the next person. A small `list_click_index_unbordered` variant (or a `bordered: bool` parameter) would make the intent explicit and remove the off-by-one footgun. Optional — not worth blocking a green PR over.

### 🔵 Suggestion — use the new `E::desc` helper consistently

You added `HelpEntry::desc()` for keyless description rows and adopted it in `logs.rs`, which is a clean cleanup. The new `mouse()` closure in `zerocode_pane::help_context` still builds the same shape the long way:

```rust
E::new(Vec::<String>::new(), format!("{}: {}", ...))
```

Switching that to `E::desc(format!(...))` would make the two call sites consistent and is exactly the case the helper was introduced for.

### 🔵 Suggestion — the four new Fluent keys are en-only

`zc-zerocode-help-{choose-section,open-section,navigate-rows,back-to-sections}` are added to `locales/en` but not to `es/fr/ja/zh-CN`. This degrades gracefully — `load_strings` uses the English map as the base and only overlays locale-defined keys, so non-en users see English rather than the `{key}` brace form — and it matches the existing tolerated gap (en is already one key ahead). No action required for correctness; just flagging so the translation pass can pick these up.

### 🔵 Suggestion — validation evidence is deferred to CI

The template's Validation Evidence section defers the local battery to CI. For a TUI-only change with all 14 Quality Gate jobs green and new unit tests covering the nav/click logic, I'm comfortable with that here. The manual scenarios you listed (cursor transitions, active-vs-inactive highlight, picker click) are the right things to have eyeballed. Worth keeping in mind that the template asks for literal tail output as the default; CI-deferral reads cleaner when it's explicitly the maintainer call it was.

— @theonlyhennygod

theonlyhennygod:
@singlerider @theonlyhennygod — milestone alignment needed: this PR does not clearly fit within the scope boundary of any open milestone (v0.8.0 config/schema, v0.8.1 integrations/channels/providers, v0.8.2 WASM plugin program, v0.8.3 MCP/web-plugin dashboard, v0.9.0 auth/security). It's a zerocode TUI UX enhancement. Please advise on placement or deferral.

singlerider:
@theonlyhennygod This was not supposed to be merged.

PR #7284, opened 2026-06-06
fix(security): create per-agent workspace dir in for_agent (+ Android shell support)

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **fix(security):** A per-agent workspace (`agents/<alias>/workspace/`) is the shell tool's spawn `cwd` and the file-tool jail root, but `SecurityPolicy::for_agent` only *resolved* the path (`config.agent_workspace_dir`) without creating it. Code paths that build the policy via `for_agent` directly — the agent loop for single-shot runs (`zeroclaw agent -m ...`), the gateway, and channels — never created the dir. On a fresh agent, **every shell command fails at `posix_spawn` with `No such file or directory (os error 2)` before it runs** (the model then narrates a misleading "command not available"). Only the `Agent::from_config*` constructor created the dir, so daemon/ACP paths worked while the loop/CLI path did not. Fix: `create_dir_all` the workspace in `for_agent` so every path is covered. Idempotent; warns and continues on failure.
  - **feat(shell):** Android support. On Android `sh` is `/system/bin/sh` (not reliably on a spawned process's PATH), and platform tools (`getprop`, `dumpsys`, `am`, `cmd`, `content`, `pm`, `settings`) live in `/system/bin`. Use the absolute shell path and inject `/system/bin:/system/xbin` into the child PATH. Runtime-detected via the existence of `/system/bin/sh`, so it is a **no-op on every other platform** and also covers musl builds running on Android (`target_os = "linux"`).
- **Scope boundary:** No change to the command allowlist, risk-profile semantics, sandbox backends, runtime selection, or non-shell tools. The PATH injection only adds the host's own system-tool dirs; it does not relax what the shell tool may run.
- **Blast radius:** `for_agent` runs for every per-agent policy build; `create_dir_all` is idempotent and the only new filesystem write is the agent's own workspace dir (already written into by bootstrap). Shell-tool changes are inert unless `/system/bin/sh` exists.
- **Linked issue(s):** none.
- **Labels:** `type: bug`, `risk: high`, `size: XS`, `config`, `runtime`, `security`, `tool`, `tool: shell`.

```bash
cargo fmt --all -- --check                                  # clean (no diffs)
cargo clippy -p zeroclaw-config -p zeroclaw-runtime --lib -- -D warnings   # Finished, no warnings
cargo test  -p zeroclaw-config --lib policy::               # test result: ok. 191 passed; 0 failed
cargo test  -p zeroclaw-config ensure_bootstrap_files       # ok. 2 passed
cargo test  -p zeroclaw-runtime --lib tools::shell          # ok. 39 passed
```

- **Beyond CI — manually verified:** Built for `aarch64-linux-android` and ran on two physical phones — **Samsung Galaxy S21 (Android 15)** and **OPPO CPH2305 (Android 16)**. Procedure: delete the agent state dir (`rm -rf .zeroclaw/agents`) to simulate a fresh install, then run `zeroclaw agent -a phone -m "..."`. **Before this change:** every shell command returned `Failed to spawn command: No such file or directory (os error 2)`. **After:** `for_agent` recreates the workspace on demand, and the agent's shell tool runs `getprop ro.product.model` (returns the model), parses `dumpsys battery`, and fires an Intent (`am start -a android.settings.SETTINGS`, launches Settings) — on both devices.
- **Intentionally skipped:** the full-workspace `cargo clippy --all-targets` / `cargo test` runs were scoped to the two touched crates locally for iteration speed; CI runs the full battery.

- New permissions, capabilities, or file system access scope? **Minor.** The only new filesystem write is `create_dir_all` of the agent's own workspace directory. On Android the shell tool can now *resolve* `/system/bin` tools; it gains no new permission — it still only runs commands allowed by the existing risk profile/allowlist.

## Compatibility (required)

- Backward compatible? **Yes.** `create_dir_all` is idempotent; the absolute shell path and PATH injection only activate when `/system/bin/sh` exists (Android).
- Upgrade steps: none.

## Rollback

Low risk: `git revert <sha>` for either commit independently.


theonlyhennygod (approved):
I read this end-to-end and cross-checked both commits against the local tree — `SecurityPolicy::for_agent` in `crates/zeroclaw-config/src/policy.rs`, the PATH-clearing path in `crates/zeroclaw-runtime/src/tools/shell.rs`, and the shell spawn in `crates/zeroclaw-config/src/platform/native.rs`. The diagnosis holds up: `for_agent` resolved `agent_workspace_dir` and handed it straight to `from_profiles` without ever creating it, so any path that builds a per-agent policy via `for_agent` (the single-shot agent loop, gateway, channels) left the shell tool's spawn `cwd` / file-tool jail root missing, and a fresh agent's first shell command died at `posix_spawn` with ENOENT before it ran. Creating it in `for_agent` is the right place — it's the one chokepoint every per-agent policy build flows through. CI is fully green (every Quality Gate job SUCCESS); the `BLOCKED` merge state is only REVIEW_REQUIRED from the two requested reviewers, not a failing check.

**Verdict: approve.** One non-blocking warning on test coverage and one minor suggestion below; neither is something I'd hold this low-risk, fully-green, manually-verified fix on.

### 🟢 What looks good — root-cause fix at the correct chokepoint

`for_agent` is exactly where this belonged. I confirmed against policy.rs:2405 that the pre-change code only resolved the path; daemon/ACP worked solely because `Agent::from_config*` created the dir, while the loop/CLI/gateway paths never did. Putting `create_dir_all` in `for_agent` covers all of them in one place instead of sprinkling creation across call sites. The PR body's before/after on two physical devices (Galaxy S21 / OPPO CPH2305, fresh-state via `rm -rf .zeroclaw/agents`) is the kind of validation that actually exercises the bug, since it only manifests at spawn time on a fresh agent.

### 🟢 What looks good — fail-open with observability is the right error posture

Warn-and-continue on `create_dir_all` failure is the correct call here, and worth naming because the instinct is often to propagate. A hard error would turn a workspace hiccup into a policy-build failure that breaks *every* agent; continuing leaves behavior no worse than the status quo (the downstream spawn still ENOENTs) while the `WARN` record with `agent_alias` + `workspace` + `error` gives an operator the breadcrumb to diagnose it. Good judgment on blast radius.

### 🟢 What looks good — Android support stays inert off-Android and respects the scope boundary

Runtime-detecting via `Path::new("/system/bin/sh").exists()` in both native.rs and shell.rs (rather than `cfg!(target_os)`) is the correct choice — it's a genuine no-op everywhere else and also covers musl-on-Android builds that report `target_os = "linux"`. I checked the ordering in shell.rs: `validate_command_execution` runs first, then `env_clear`, then the PATH overlay — so the injection only changes how an already-allowed command resolves. It adds the host's own `/system/bin:/system/xbin`; it does not widen the allowlist or relax the risk profile. Scope boundary in the PR body matches what the code does.

### 🟡 Warning — no regression test pins the dir-creation behavior

The new behavior — `for_agent` creates the workspace — is the entire fix, but nothing asserts it. The existing `for_agent_*` tests (policy.rs:4285, :4403) exercise the new line and keep passing, which is why the suite is green, but none of them checks that the directory actually exists afterward. A three-line test — call `for_agent` against a temp-dir config for an agent whose workspace doesn't yet exist, then assert the workspace path `is_dir()` — would lock the regression shut and is cheap relative to the manual-device validation you already did. Not blocking (the change is low-risk and you verified it on real hardware), but this is the one place I'd want a guard so the dir-creation can't silently regress out again later.

### 🔵 Suggestion — the Android PATH block reads host env, not the command's overlaid PATH

In shell.rs the injection does `std::env::var("PATH")` (the daemon's own PATH) rather than the value just set on `cmd`. In the common case these match — `PATH` is in `SAFE_ENV_VARS`, so `env_clear` + rebuild already set the host PATH on `cmd`. The only divergence is if `tui_env` overlaid a different PATH: the Android block would ignore that overlay and re-derive from host env. That's a narrow scenario (a TUI client attached to an Android daemon with a custom PATH) and almost certainly irrelevant in practice, but if you want the system-dir prepend to compose with whatever `cmd` ended up with, deriving `base` from the resolved command PATH would be marginally more correct. Pure optional.

For the record on bundling: shipping the `fix(security):` workspace fix and the `feat(shell):` Android support as two commits in one PR is fine here — they're the two halves of "make a fresh agent usable on a phone," each is small, and you've kept them independently revertable. No concern.

Nice, tight fix — clear write-up, honest about what was scoped locally vs. left to CI, and verified where it actually matters.

crates/zeroclaw-config/src/policy.rs
```diff
@@ -2402,6 +2402,23 @@ impl SecurityPolicy {
...
+        // The per-agent workspace is the shell tool's spawn cwd and the file-tool
+        // jail root. Create it here so every path that builds a per-agent policy
+        // (agent loop, gateway, channels) has the directory present. A missing cwd
+        // makes the shell tool's process spawn fail with ENOENT on a fresh agent.
+        if let Err(e) = std::fs::create_dir_all(&agent_workspace) {
+            ::zeroclaw_log::record!(
+                WARN,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
+                    .with_attrs(::serde_json::json!({
```
singlerider: This is incorrect record macro attribution. You must use the span attribution system.

singlerider (changes requested):

perlowja:
Fixed in da44875 — `agent_alias` now goes through the span attribution system: I bind it via a `zeroclaw_scope` span (`target: "zeroclaw_log_internal_scope"`, matching the `scope!`/`attribution_span!` pattern in `ws.rs`) and `.entered()` for the sync path, so the layer's `ScopeSpanCollector` picks it up as an alias-bound attribution field. The remaining `with_attrs` carries only `workspace` and `error`, which aren't attribution fields. `cargo fmt --check` and `clippy -p zeroclaw-config -D warnings` are clean. Thanks for the catch — and thanks @theonlyhennygod for the cross-check.

Audacity88 (commented):
I reviewed current head 4b2d07ea6ea80a0d3b2882412ae5f41ee2cd9289 against the live PR state, prior reviews, the inline span-attribution thread, the author follow-up, the PR body validation evidence, and the green CI run. I checked the current diff in `crates/zeroclaw-config/src/policy.rs`, `crates/zeroclaw-config/src/platform/native.rs`, and `crates/zeroclaw-runtime/src/tools/shell.rs`. I did not rerun local cargo for this re-review.

### ✅ Resolved — span attribution blocker appears addressed

@singlerider’s inline blocker was about putting `agent_alias` into free-form `record!(...).with_attrs`. Current head now binds `agent_alias` through a `zeroclaw_scope` span with target `zeroclaw_log_internal_scope` before emitting the warning, and leaves only `workspace` and `error` in the event attrs. That matches the existing span-attribution pattern used for alias-bound fields elsewhere, so from my read the specific attribution issue is resolved.

### 🟢 What looks good — workspace creation is at the policy chokepoint

The root-cause fix is in the right place. `SecurityPolicy::for_agent` is the common path that decides the per-agent workspace used as the shell spawn cwd and file-tool jail root, so creating the directory there covers the single-shot agent loop, gateway, and channel paths without adding call-site-specific fixes.

### 🟢 What looks good — Android shell support stays scoped

The Android changes are also narrow: `NativeRuntime` uses `/system/bin/sh` only when that path exists, and the shell tool only prepends `/system/bin:/system/xbin` under the same runtime check. The command allowlist, risk profile, sandbox selection, and non-Android behavior are not widened by this patch.

Separate maintainer cleanup before merge: the PR body label snapshot still says `risk: low` and `size: S`, while the live labels are `risk: high` and `size: XS` with the config/runtime/security/tool scope labels. That should be aligned, but I do not see it as an author-side blocker.

I’m leaving this as a comment rather than approving because @singlerider’s `CHANGES_REQUESTED` review is still active. From my side, I do not see a new blocker; this looks ready for that review to be cleared or re-reviewed.

perlowja:
@singlerider — the span-attribution blocker is resolved (per @Audacity88's last pass) and I've brought this current:

- **Rebased onto current master** (clean merge, 189 commits). The fix still sits at the policy chokepoint: `SecurityPolicy::for_agent` creates the per-agent workspace at `crates/zeroclaw-config/src/policy.rs:2420`; verified on a build host — `for_agent` + `workspace_dir_regression` tests pass and clippy is clean.
- **PR-body label snapshot aligned** to the live labels (`risk: high`, `size: XS`, + config/runtime/security/tool scope) per @Audacity88's note.

A cross-family adversarial pass suggested (a) gating the Android shell/PATH selection on `cfg!(target_os = "android")` instead of `/system/bin/sh` existence, and (b) hard-failing `for_agent` when the workspace mkdir errors. I've left both by design: the path-existence probe is intentionally broader so Termux/proot-on-Android (where `target_os` compiles as `linux` but `/system/bin/sh` exists) is still covered, and the warn-and-continue on mkdir failure is the deliberate resilience behavior from the span-attribution thread rather than crashing agent construction on a transient/read-only FS. Happy to revisit either if you'd prefer the stricter contract.

Ready for re-review whenever convenient.

crates/zeroclaw-runtime/src/tools/shell.rs
```diff
@@ -347,6 +347,20 @@ impl Tool for ShellTool {
             }
         }
 
+        // Android: platform tools (sh, getprop, am, dumpsys, content, pm, ...)
+        // live in /system/bin and /system/xbin. The cleared+rebuilt PATH above
+        // may omit them, leaving the shell unable to resolve any platform tool.
+        // Detect Android at runtime (works for bionic and musl builds).
+        if std::path::Path::new("/system/bin/sh").exists() {
+            let base = std::env::var("PATH").unwrap_or_default();
```
singlerider: This reads `std::env::var("PATH")` (the daemon's ambient PATH) and sets it on `cmd`, but the block right above deliberately builds `cmd`'s environment via `env_clear()` + `collect_allowed_shell_env_vars` + the `tui_env` overlay, with a comment that TUI's PATH must win. This bypasses all of that: if PATH was in the allowlist or `tui_env`, the curated value is clobbered by `/system/bin:/system/xbin:<daemon PATH>`; if PATH was deliberately excluded, the daemon's leaks into the child the allowlist meant to keep out. It passes on a phone where the daemon PATH is fine, which is why the two-device run did not catch it. Prepend the Android dirs to whatever PATH is already staged on `cmd` (run it through the same allowlist/tui_env resolution), not to the ambient process PATH.

crates/zeroclaw-config/src/policy.rs
```diff
@@ -2413,6 +2413,30 @@ impl SecurityPolicy {
         // file_read/write/edit and the shell tool jail to the agent's
         // own dir, not the install-wide legacy path.
         let agent_workspace = config.agent_workspace_dir(agent_alias);
+        // The per-agent workspace is the shell tool's spawn cwd and the file-tool
+        // jail root. Create it here so every path that builds a per-agent policy
+        // (agent loop, gateway, channels) has the directory present. A missing cwd
+        // makes the shell tool's process spawn fail with ENOENT on a fresh agent.
+        if let Err(e) = std::fs::create_dir_all(&agent_workspace) {
```
singlerider: `for_agent` `?`-propagates its other errors, but this one is WARN-and-continue, after which the policy is built with a cwd/jail root that does not exist, re-creating this PR's own ENOENT failure mode one layer later behind a warning. Either propagate (consistent with the surrounding `?`), or keep the continue and add one line explaining why a missing workspace is recoverable here when it was not before. "continuing" alone is hand-wave given the function's error discipline and the PR's thesis.

crates/zeroclaw-config/src/platform/native.rs
```diff
@@ -47,7 +47,15 @@ impl RuntimeAdapter for NativeRuntime {
     ) -> anyhow::Result<tokio::process::Command> {
         #[cfg(not(target_os = "windows"))]
         {
-            let mut process = tokio::process::Command::new("sh");
+            // Android keeps its shell at /system/bin/sh and it is not always
+            // on PATH for spawned processes; use the absolute path when present
+            // so the shell can launch (and reach platform tools).
+            let shell = if std::path::Path::new("/system/bin/sh").exists() {
```
singlerider: This `Path::new("/system/bin/sh").exists()` probe is the definition of "this host has Android platform tools," and it is written twice: here for shell selection and at `shell.rs:354` for PATH injection. Extract one `is_android_runtime()` in the platform module and call it from both so the two sites cannot drift on what "Android" means. The runtime-probe approach is correct; only the duplication needs fixing.

singlerider (changes requested):
The diagnosis is sharp and the fix is in the right place: `for_agent` is exactly the chokepoint the loop/CLI/gateway/channels all build through, and creating the workspace there closes the ENOENT-on-fresh-agent hole the daemon path never hit. The Android runtime detection (probe `/system/bin/sh` rather than `cfg!(target_os)`) is the correct call for musl builds running on Android, and the attribution on the new WARN is done right: the `agent_alias` scope span is the canonical shape, the field is alias-bound, and the guard is held across the `record!`. Two blocking items, both in the parts the live-on-device testing could not catch.

### 🔴 Blocking: the Android PATH injection discards the curated PATH the tool just built

`shell.rs` reads `std::env::var("PATH")` (the daemon process's ambient PATH) and sets it on `cmd`, but the block immediately above deliberately constructs `cmd`'s environment via `env_clear()` -> `collect_allowed_shell_env_vars` -> `tui_env` overlay, with a comment stating TUI's PATH must win. The Android block ignores all of that:

- If `PATH` was in the allowlist or `tui_env`, the curated value (the one the comment says must take precedence) is clobbered by `/system/bin:/system/xbin:<daemon PATH>`.
- If `PATH` was deliberately excluded from the allowlist, `std::env::var("PATH")` still reads the daemon's and leaks it into the child the env-allowlist meant to keep out.

This passes on a phone where the daemon PATH happens to be fine, which is why the two-device run did not surface it. Prepend the Android dirs to whatever PATH is already staged on `cmd` (thread it through the same allowlist/tui_env resolution), not to the ambient process PATH.

### 🔴 Blocking: no test pins either fix

The PR is `risk: high` and fixes a bug described as "every shell command fails on a fresh agent," yet the diff adds zero `#[test]`. The Validation Evidence lists `policy::` (191) and `tools::shell` (39) passing, but those passed before this change too; none detects the bug or pins the fix. The live two-device run is great evidence but is not a regression guard. Needed:

1. `for_agent` creates the workspace dir when absent (assert the dir exists after the call on a fresh temp config).
2. The fresh-agent shell-success path, or at least that the policy's resolved cwd exists post-construction.
3. The Android branches, behind the runtime probe (a seam that lets the probe be faked, or a unit test on the PATH-composition helper once it is extracted per the note below).

Without these, the exact regression this PR fixes can silently return.

### 🟢 The attribution on the failure WARN is correct

The `agent_alias` scope span uses the canonical `zeroclaw_log_internal_scope` target and name, `agent_alias` is in `ATTRIBUTION_FIELDS` so it lands as alias-bound attribution rather than a free-form attribute, and the guard is bound to `_scope` (not a bare `_`) so it is held across the `record!`. Binding only `agent_alias` is right for this call site: no model/session/provider exists at policy-build time, matching the `daemon/mod.rs` precedent that binds only its scope-appropriate field. This is exactly the attribution discipline the hard rule asks for.

### 🟢 The fix targets the real chokepoint

`for_agent` resolving and now creating the workspace means the loop/CLI single-shot path, gateway, and channels all converge on one creation site instead of relying on `Agent::from_config*`. The idempotent `create_dir_all` is the right primitive.

### 🟡 Centralize the Android runtime probe

`Path::new("/system/bin/sh").exists()` is the definition of "this host has Android platform tools," and it is written twice: `native.rs:53` (shell selection) and `shell.rs:354` (PATH injection). Extract one `is_android_runtime()` (or similar) in the platform module and call it from both, so the two sites cannot drift on what "Android" means. The runtime-probe approach is correct; only the duplication needs fixing.

### 🟡 Justify or tighten the create_dir_all swallow

`for_agent` `?`-propagates its other errors but this one is WARN-and-continue, after which the policy is built with a cwd/jail root that does not exist, re-creating this PR's own ENOENT failure mode one layer later, now behind a warning. There is a defensible argument for not blocking policy construction on a transient FS error, but "continuing" with no rationale is hand-wave given the function's own error discipline and the PR's thesis. Either propagate (consistent with the surrounding `?`), or keep the continue and add one line explaining why a missing workspace is recoverable here when it was not before.

---
Span attribution is correct; the fix is well-placed. Land the PATH-composition fix and a regression test or two, factor the Android probe, and justify the swallow, and this is ready.

Audacity88 (commented):
Reviewed current head `e1f4e9e` against the active SingleRider review, inline threads, the author's 2026-06-13 follow-up, current diff/source, changed-file list, and green GitHub checks. I am leaving this as a comment rather than approving while SingleRider's `CHANGES_REQUESTED` review remains active. I did not rerun local cargo.

### ✅ Resolved — workspace creation now propagates failures

`SecurityPolicy::for_agent` now creates the per-agent workspace before returning the policy and propagates failures with context, so fresh-agent shell/file-tool construction no longer silently continues with a missing cwd/jail root.

### ✅ Resolved — Android PATH now preserves TUI PATH

The Android shell-tool path now prefixes `/system/bin:/system/xbin` onto `tui_env.PATH` when the TUI provided one, so the curated user PATH wins over the daemon's ambient environment on that path.

### ✅ Resolved — Android runtime detection is centralized

Both native shell selection and shell-tool PATH handling now call `zeroclaw_api::platform::is_android()`, so the duplicate runtime probe concern is addressed.

### 🟡 Warning — Android PATH precedence should be made explicit and tested

I do not see a remaining behavior mismatch in the common path because `PATH` is still in `SAFE_ENV_VARS` and TUI PATH now wins. The precedence rule is still reconstructed inline in `execute`, though, with the no-TUI branch reaching back to `std::env::var("PATH")`. A small helper/test that says Android prefixes `/system/bin:/system/xbin` onto the resolved child PATH, with TUI PATH winning, would make the invariant less fragile.

### 🔴 Blocking — SingleRider's regression-test blocker remains live

This PR is still a `risk: high` security/runtime/shell change, and the current changed-file list/diff adds only source files, with no regression tests for the two fixes being merged: `for_agent` creating/using a fresh per-agent workspace and Android PATH composition/shell launch behavior. The validation evidence lists existing suites, but no new test pins the fresh-agent workspace case or the Android PATH precedence case. I would keep the existing block in place until those targeted tests land.

WareWolf-MoonWall (commented):
Commenting on #7284 — `fix(security): create per-agent workspace dir in for_agent (+ Android shell support)` by @perlowja. @singlerider holds active CHANGES_REQUESTED at 0bc1b2e7, so the verdict here is `--comment` per protocol — I don't approve over an active block. @theonlyhennygod has approved. @Audacity88's most recent comment at e1f4e9e is the latest substantive review activity, and I'm concurring with their finding on regression coverage before deferring to @singlerider to close the loop on their open block.

### 🟡 Warning — Regression coverage gap on the `for_agent`-fresh-workspace case and Android PATH precedence invariant

This is a `risk: high` change touching a security/runtime chokepoint (`for_agent` is the single construction path for all agent workspace state). @Audacity88 called this out in their latest comment, and I agree: the validation evidence cites existing test suites but there are no new test cases that specifically pin the `for_agent`-fresh-workspace creation path or the Android PATH precedence invariant introduced by this PR. For a change that explicitly fixes a "workspace dir not created" failure mode, a regression test that constructs an agent handle against a clean temp directory and asserts the workspace exists afterward would be the minimum. For the Android PATH override, a test (or at minimum a documented, committed test fixture with the two Android versions tested) that asserts the curated PATH takes precedence over the ambient daemon env would close the gap. The live device validation described in the PR body is strong evidence that the fix works today, but it doesn't protect against future regressions at the `for_agent` chokepoint.

### 🟢 What looks good — `for_agent` is the right chokepoint and Android detection is correct

The architectural choice to fix this at `for_agent` rather than at individual call sites is exactly right. Every code path that constructs an agent goes through it, so a fix there is comprehensive by construction — no scatter-gun call-site patching needed. The Android shell detection via `/system/bin/sh` probe rather than `cfg!(target_os = "android")` is also the correct approach for musl-on-Android targets where `target_os` may not reflect the actual execution environment. Live device validation on two Android versions with specific tool invocation verification is meaningful signal that goes beyond what CI can provide for this target.

@singlerider's active CHANGES_REQUESTED is the gate here. I'm deferring to them to verify the current state of their blockers and dismiss or update their review once satisfied.

singlerider (changes requested):
Re-reviewed at head `e1f4e9e`. Three of my four blockers are resolved here — confirming them — but the regression-test gate is still unmet, and @Audacity88 and @WareWolf-MoonWall both independently concur on that. Keeping changes-requested, now narrowed to the tests.

### RESOLVED ✅ — workspace creation propagates

`SecurityPolicy::for_agent` now does `create_dir_all(&agent_workspace).with_context(...)?` using `config.agent_workspace_dir(agent_alias)`, so a fresh agent no longer builds a policy with a non-existent cwd/jail root, and the failure propagates with context instead of WARN-and-continue. That closes both my swallow concern and the ENOENT hole at the chokepoint.

### RESOLVED ✅ — Android PATH preserves the curated value

The Android branch now prepends `/system/bin:/system/xbin` onto `tui_env.PATH` when the TUI provided one (falling back to `std::env::var("PATH")` only when there is no TUI), so the curated PATH wins over the daemon's ambient environment instead of being clobbered. That fixes the precedence bug the two-device run couldn't surface.

### RESOLVED ✅ — Android probe centralized

Both native shell selection and the shell-tool PATH path now call `zeroclaw_api::platform::is_android()`, so the duplicated `/system/bin/sh` probe can no longer drift.

### 🔴 Blocking — regression tests still absent (and @Audacity88 / @WareWolf-MoonWall agree)

I checked every changed file: there are no `#[test]`/`#[tokio::test]` blocks added (the one new `fn` is `is_android()`, not a test). This is a `risk: high` security/shell change whose whole thesis is "every shell command fails on a fresh agent," and neither fix is pinned. The minimum:

1. `for_agent` against a clean temp config creates the per-agent workspace — assert the dir exists after the call.
2. The fresh-agent shell-success path, or at least that the policy's resolved cwd exists post-construction.
3. Android PATH composition — extract the PATH-resolution into a helper and unit-test that, behind `is_android()`, `/system/bin:/system/xbin` is prefixed onto the resolved child PATH with the TUI PATH winning. The no-TUI branch reaching back to `std::env::var("PATH")` is exactly the inline precedence rule that should live in a tested helper rather than being reconstructed in `execute`.

The existing suites cited in validation passed before this change too, so they don't guard the regression. The live two-device run is strong evidence the fix works today but is not a regression guard at the `for_agent` chokepoint.

The fix is well-placed and the other three items are genuinely resolved — this is now purely the test gate. Add those and I'll approve.

perlowja:
@singlerider @Audacity88 @WareWolf-MoonWall — regression tests added (your last open blocker), no behavior change beyond a testability refactor:

- **Extracted** the Android child-PATH composition out of `ShellTool::execute` into a pure `android_child_path(tui_path, ambient)` helper (the inline precedence rule now lives in one tested place).
- **`android_child_path_prefixes_platform_dirs_with_tui_path_winning`** — asserts `/system/bin:/system/xbin` is prefixed, **TUI PATH wins** over ambient, no-TUI falls back to ambient, empty base → bare platform dirs.
- **`is_android_returns_bool_without_panicking`** — smoke test on the centralized probe.
- **`for_agent_creates_the_per_agent_workspace_dir`** — against a clean temp config: workspace dir does **not** exist, then **does** exist after `for_agent`, and `policy.workspace_dir` equals it — pins both the chokepoint dir-creation and the resolved-cwd invariant.

Verified locally: `cargo clippy -p zeroclaw-runtime -p zeroclaw-config --all-targets -- -D warnings` clean; all three new tests pass. Re-requesting review.

Audacity88 (commented):
Reviewed current head `8f00194e` against the active SingleRider review, inline threads, the author's 2026-06-15 follow-up, current diff/source, changed-file list, and green GitHub checks. I am leaving this as a comment rather than approving while SingleRider's `CHANGES_REQUESTED` review remains active. I did not rerun local cargo.

### ✅ Resolved — `for_agent` fresh-workspace behavior is now pinned

`SecurityPolicy::for_agent` now creates the per-agent workspace before returning the policy and propagates failures with context, so fresh-agent shell/file-tool construction no longer silently continues with a missing cwd/jail root. The new `for_agent_creates_the_per_agent_workspace_dir` test covers the minimum regression case from the active review: a fresh agent starts without the workspace directory, `for_agent` creates it, and the returned policy uses that same path as the workspace boundary.

### ✅ Resolved — Android PATH precedence is helper-owned and tested

The Android shell-tool path now prefixes `/system/bin:/system/xbin` onto the resolved child PATH, with `tui_env.PATH` winning when the TUI provided one and the daemon's ambient PATH used only as the fallback. The new `android_child_path_prefixes_platform_dirs_with_tui_path_winning` test pins the TUI-wins, no-TUI, and empty-path cases, which addresses the remaining PATH-clobbering regression concern from my side.

### ✅ Resolved — Android runtime detection remains centralized

Both native shell selection and shell-tool PATH handling now call `zeroclaw_api::platform::is_android()`, and the helper has a smoke test so the runtime probe is not duplicated back into each callsite.

### 🟢 What looks good — high-risk behavior stays at the chokepoints

The final shape keeps the behavior in the right narrow places: policy construction owns creation of the policy cwd/jail root, native runtime selection owns the Android shell path, and shell execution owns Android child PATH composition after the normal safe-env plus TUI overlay has been applied.

### 🟡 Review-state note — still waiting on SingleRider

From my read of current head, the regression-test concern I previously agreed with is addressed. The PR should still wait for @singlerider to clear or update their active `CHANGES_REQUESTED` review before landing.

perlowja:
Thanks @singlerider — the regression-test gate is now met (commit `8f00194e`).

## Regression tests added

`test(security): regression tests for for_agent workspace creation + Android PATH` adds focused coverage for the previously untested paths:

- `crates/zeroclaw-config/src/policy.rs` (+41) — `SecurityPolicy::for_agent` creates the per-agent workspace dir and propagates a contextful error when it can't, covering both the ENOENT chokepoint and the earlier WARN-and-continue swallow.
- `crates/zeroclaw-runtime/src/tools/shell.rs` (+38) — the Android child-PATH composition, extracted into a pure function so it can be asserted directly (curated TUI PATH wins, `/system/bin:/system/xbin` prepended, ambient `PATH` used only as fallback).

The three earlier blockers (workspace-creation propagation, Android PATH precedence, centralized `is_android()` probe) remain resolved as you confirmed. Ready for another look.

singlerider (approved):
Re-review at head 8f00194e against current master and my prior CHANGES_REQUESTED. Checked out the head locally and ran the new tests.

### RESOLVED ✅ Regression tests landed and pass

My remaining blocker (and the concurring asks from @Audacity88 and @WareWolf-MoonWall) was the missing regression gate on a `risk: high` security/shell change. All three of my required cases are now covered and green locally:

1. `for_agent_creates_the_per_agent_workspace_dir` in policy.rs asserts the per-agent workspace does not exist before the call and exists after `SecurityPolicy::for_agent`, pinning the ENOENT fix at the chokepoint. Passes.
2. The Android PATH precedence rule was extracted into a testable `android_child_path` helper, and `android_child_path_prefixes_platform_dirs_with_tui_path_winning` asserts `/system/bin:/system/xbin` is prefixed with the TUI PATH winning over ambient. Passes. This is exactly the refactor I asked for, not an inline rule reconstructed in `execute`.

```
test policy::tests::for_agent_creates_the_per_agent_workspace_dir ... ok
test tools::shell::tests::android_child_path_prefixes_platform_dirs_with_tui_path_winning ... ok
```

### RESOLVED ✅ The three prior fixes still hold

Workspace creation propagates with context at `for_agent`, the Android branch preserves the curated TUI PATH, and the `is_android()` probe is centralized. All confirmed at this head.

Both fixes are now pinned by regression guards. Approving.

PR #7287, opened 2026-06-06
feat(plugins): add huggingface inference plugin

theonlyhennygod:
## Summary
- **Base:** `master`. Adds a self-contained WASM/Extism tool plugin `plugins/huggingface/` exposing an `hf_infer` tool that runs **text/JSON** models via the Hugging Face Inference API (`POST https://api-inference.huggingface.co/models/<id>`).
- Stateless, one request → one response. Uses only the `http_client` + `env_read` host permissions (reads `HF_TOKEN`), mirroring the `firecrawl` reference plugin (own `[workspace]`, exports `tool_metadata` + `execute`).
- Renders the common task shapes (text-generation, summarization, translation, classification, embeddings) with a compact-JSON fallback. **Scoped to text/JSON:** binary-output models (image/audio) are rejected with a clear error, because the host HTTP bridge is text-only (`runtime.rs` `body: String`). Output ends with the mandatory **fidelity footer** (`Fields returned: model, output`).
- **Scope boundary:** purely additive; no host/runtime/config/root-`Cargo.toml` changes. Auto-discovered at runtime.
- One of a stateless AI-native plugin batch (siblings: `tavily`, `jina-reader`). Binary-media HF models await the planned host base64 enabler.

## Validation Evidence
- `cargo test --lib` → 5/5 (footer + the render-shape variants) pass.
- `cargo build --target wasm32-unknown-unknown --release` → 278KB `huggingface.wasm`.
- Not exercised: a live HF call; HTTP/env paths run through host fns unavailable outside the runtime. Source committed; `.wasm`/`target/` gitignored.

## Security & Privacy
- New permissions: declares `http_client` + `env_read`; no filesystem access. The model id is validated (rejects `..`, leading `/`, `?`, `#`) before going into the request URL.
- New network call: outbound HTTPS to `api-inference.huggingface.co` only, via the host `zc_http_request`.
- Secrets: reads `HF_TOKEN` via `env_read` (Bearer); never logged; no secret in the diff.
- PII: none (fixtures use placeholder ids).

## Compatibility
- Backward compatible (additive). Optional env var `HF_TOKEN`. Opt-in (requires `[plugins].enabled` + a `plugins-wasm` build).

## Rollback
- `git revert <commit>` / delete `plugins/huggingface/`.


Audacity88 (changes requested):
Reviewed current head 67889427c1157f124e72366639407ffe23e59422 against the PR body, green CI, empty comment/review history, and the public diff. The additive plugin shape and permission set look reasonable, but one truncation path can still turn normal provider text into a plugin trap.

### 🟢 What looks good — narrow WASM plugin boundary

The PR keeps the change isolated under `plugins/huggingface/` and declares only `http_client` plus `env_read` in the manifest. The JSON-only scope is also honest: non-JSON responses become a clear tool error instead of trying to pass binary data through this text-returning tool.

### 🔴 Blocking — use character-safe truncation before slicing provider text

The new truncation code slices Rust strings by byte index in both the HTTP error path and the successful output path:

- `&resp.body[..resp.body.len().min(500)]`
- `&rendered[..rendered.len().min(MAX_OUTPUT_CHARS)]`

Rust string indices must land on UTF-8 character boundaries. Hugging Face text-generation, translation, classification labels, or error payloads can all contain non-ASCII text. If a long response crosses the 500-byte or 12,000-byte cap in the middle of a multibyte character, the plugin panics before it can return the structured `ToolResult` error/output and fidelity footer.

Please make both truncation paths character-safe, for example with a small helper that returns a `String` from `s.chars().take(limit).collect()` plus a truncation flag. It would be good to add a regression with a long non-ASCII generated text or error body that crosses the cap.

WareWolf-MoonWall (changes requested):
I reviewed #7287 at head 67889427c1157f124e72366639407ffe23e59422. This review covers the PR body, CI checks, diff, and existing review by @Audacity88. I did not run a live Hugging Face API call.

### 🟢 What looks good — Text/JSON scope is honest and narrow

The plugin correctly scopes itself to text/JSON models and rejects binary-output models with a clear error message. The permission set (`http_client` + `env_read`) is appropriate, the `model_id` validation rejects path-traversal characters before URL construction, and the fidelity footer makes the returned fields explicit. The task-specific rendering (text-generation, summarization, translation, classification, embeddings) with a JSON fallback is a good pattern for a stateless inference tool.

### 🔴 Blocking — UTF-8 truncation safety on two paths

Per FND-006 §4.3 (fail-safe boundaries), the plugin must not panic on normal provider text. Two truncation paths use byte-index slicing:

1. **Error path** (line 690): `&resp.body[..resp.body.len().min(500)]`
2. **Output path** (line 707): `&rendered[..rendered.len().min(MAX_OUTPUT_CHARS)]`

Rust `str` indices must land on UTF-8 character boundaries. Hugging Face text-generation, translation, classification labels, or error payloads can contain non-ASCII text. If a long response crosses the 500-byte or 12,000-byte cap mid-codepoint, the plugin panics before returning a structured `ToolResult`.

**Required:** Make both truncation paths character-safe. Use `.chars().take(limit).collect::<String>()` or a boundary-safe helper that uses `char_indices()` to find the last valid byte boundary. Add regression tests for both paths with long UTF-8 strings containing multibyte characters crossing the truncation boundary.

### 🔵 Suggestion — Clarify MAX_OUTPUT_CHARS rationale

The 12,000-character output cap is undocumented. If this is cost control, abuse prevention, or model-context management, add a comment explaining the policy. If it's tunable, note that. If it's fixed, cite the constraint. This helps future maintainers understand whether the cap is negotiable.

singlerider:

PR #7288, opened 2026-06-06
feat(plugins): base64 binary support for the WASM HTTP host fn

theonlyhennygod:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The WASM plugin host function `zc_http_request` (`crates/zeroclaw-plugins/src/runtime.rs`) was text-only — request body was `Option<String>` and the response body was read via `resp.text()` — so plugins could neither send nor receive binary payloads. This blocked every audio/image plugin (ElevenLabs voice, Hugging Face image/audio models, Stability, OCR).
  - `HttpRequest` gains `body_base64`: when set, it is decoded to raw bytes and used as the request body, taking precedence over `body`. This lets a plugin upload binary content across the text-only host↔guest boundary.
  - `HttpResponse` gains `body_base64`: populated **only** for binary/non-textual content types. Textual responses are unchanged and the field is omitted from their JSON via `skip_serializing_if`.
  - Adds three pure helpers — `is_textual_content_type`, `encode_response_body`, `resolve_request_body` — that hold the codec/classification logic, each covered by unit tests.
- **Scope boundary:** Touches a single core crate (`crates/zeroclaw-plugins`) and one file (`runtime.rs`). No API, config, or schema changes. No changes to any existing plugin. Purely additive and opt-in — a plugin only uses base64 if it sets/reads the new fields.
- **Blast radius:** The shared WASM HTTP host function used by all `http_client`-permitted plugins. Mitigated by additive serde defaults: existing text plugins (firecrawl, tavily, replicate, jina-reader, huggingface, etc.) serialize/deserialize to the exact same shape as before.
- **Linked issue(s):** None.
- **Labels:** `enhancement`, `risk: medium`, `size: S` (auto-labels applied on open).

```bash
cargo test -p zeroclaw-plugins
cargo clippy -p zeroclaw-plugins --all-targets -- -D warnings
cargo fmt -p zeroclaw-plugins -- --check
```

  - `cargo test -p zeroclaw-plugins` → **47 passed, 0 failed**. Includes 8 new tests: request/response serde round-trips, back-compat (a pre-change guest's request JSON with no `body_base64` still parses), text-response-omits-base64, `resolve_request_body` precedence/text/none/bad-base64 cases, `encode_response_body` text-vs-binary round-trip, empty-content-type UTF-8 fallback, and content-type classification.
  - `cargo clippy -p zeroclaw-plugins --all-targets -- -D warnings` → clean.
  - `cargo fmt -p zeroclaw-plugins -- --check` → clean.
  - Toolchain: `rustc 1.93.1 (01f6ddf75 2026-02-11)`.
- **Beyond CI — what did you manually verify?** Verified backward compatibility two ways via explicit unit tests: (a) a pre-change guest request JSON (no `body_base64`) still deserializes thanks to `serde(default)`, and (b) textual responses serialize to the identical JSON shape as before because `body_base64` uses `skip_serializing_if` and is omitted. Verified `body_base64` precedence over `body` on the request, and that only non-textual content types populate the response `body_base64`.
- **If any command was intentionally skipped, why:** Not exercised: a live binary HTTP round-trip through a real WASM guest — that handler runs inside the Extism host and requires the full runtime. The codec/classification logic is instead covered by the pure-function unit tests.

- New permissions, capabilities, or file system access scope? (`No`) — still gated by the existing `http_client` permission; no filesystem access added.
- New external network calls? (`No`) — no new network destinations; same request path.
- If any `Yes`, describe the risk and mitigation: N/A. Note: `base64` (0.22) is an existing dependency (already used by `signature.rs`); no new dependency added.

## Compatibility (required)

- If `No` or `Yes` to either: exact upgrade steps for existing users: None — the change is additive and opt-in. Existing text plugins require no changes.

- **Fast rollback command/path:** `git revert e13a0b8ba`
- **Observable failure symptoms:** Plugin HTTP requests failing to decode (grep logs for base64 decode errors in the `zc_http_request` host path) or unexpected `body_base64` fields appearing on textual responses.

### DIFF
--- a/crates/zeroclaw-plugins/src/runtime.rs
+++ b/crates/zeroclaw-plugins/src/runtime.rs
@@ -6,6 +6,8 @@
 
 use crate::PluginPermission;
 use anyhow::{Context, Result};
+use base64::Engine;
+use base64::engine::general_purpose::STANDARD;
 use extism::*;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
@@ -29,15 +31,29 @@ struct HttpRequest {
     url: String,
     #[serde(default)]
     headers: std::collections::HashMap<String, String>,
+    /// Textual request body (JSON, form, etc.).
     #[serde(default)]
     body: Option<String>,
+    /// Standard-base64-encoded **binary** request body. When present it takes
+    /// precedence over `body` and is decoded to raw bytes before sending — this
+    /// lets a plugin upload binary payloads (e.g. audio for speech-to-text)
+    /// across the text-only host↔guest boundary.
+    #[serde(default)]
+    body_base64: Option<String>,
 }
 
 /// HTTP response returned from host to plugin.
 #[derive(Debug, Serialize, Deserialize)]
 struct HttpResponse {
     status: u16,
+    /// Response body as text, for textual content types (`text/*`, JSON, XML…).
+    /// Empty when the response is binary — see `body_base64`.
     body: String,
+    /// Standard-base64-encoded response body, populated **only** when the
+    /// response is binary (a non-textual content type). Omitted entirely for
+    /// textual responses, so existing text plugins are unaffected.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    body_base64: Option<String>,
     #[serde(default)]
     headers: std::collections::HashMap<String, String>,
 }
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
I reviewed the plugin HTTP host-function change at abe0bd7e15f28b246bc7e03ca2fb9a873df8e578, including the PR metadata/body, diff, current check snapshot, and the existing `zc_http_request` protocol docs. I did not see linked issues, prior reviews, or review-thread context on the PR, and I did not run a live local plugin round trip for this pass.

### 🟢 What looks good — Backward-compatible codec shape

The code path is nicely scoped to `zeroclaw-plugins`, and the helper extraction makes the binary/text behavior easy to reason about. The additive serde fields, `skip_serializing_if` on textual responses, explicit invalid-base64 error, and helper-level tests are the right shape for this slice. The implementation keeps existing text plugins on the old JSON shape while giving binary plugins a clear opt-in path.

### 🔴 Blocking — Update the documented `zc_http_request` protocol

This PR adds `body_base64` to the request and response JSON exchanged across `zc_http_request`, but `docs/book/src/developing/plugin-protocol.md` still documents the old text-only shape. The PR body also says there are “No API, config, or schema changes,” which is true for Rust config/API, but not true for the plugin-facing host-function protocol. That PR-body statement should be corrected to note that the `zc_http_request` JSON protocol is changing for plugin authors.

Because this feature exists specifically for plugin authors, the protocol doc should describe both new fields, the request precedence rule (`body_base64` wins over `body`), and the response rule (textual content stays in `body`; binary content uses `body_base64` with an empty text body). That keeps the contributor-facing contract aligned with the implementation and avoids leaving authors to reverse-engineer the new shape from `runtime.rs`.

### 🟡 Warning — Correct the rollback command in the PR body

The PR body lists `git revert e13a0b8ba`, but the current captured head is abe0bd7e15f28b246bc7e03ca2fb9a873df8e578. Since this is labeled `risk: medium`, the rollback field should point at the current commit or use another accurate rollback path before merge.

### 🔵 Suggestion — A live host-function round trip would strengthen this

The pure helper coverage is good, and I agree it exercises the main codec decisions. Since this is a host-function protocol change, a small follow-up test or manual validation that drives `zc_http_request` through a plugin guest would give stronger confidence that Extism memory I/O, request decoding, reqwest response bytes, and response serialization all compose. I do not think this needs to block this PR once the protocol docs are fixed, but it is worth adding when the fixture cost is reasonable.

WareWolf-MoonWall (changes requested):
I reviewed #7288 at head abe0bd7e15f28b246bc7e03ca2fb9a873df8e578. This review covers the PR body, CI checks, diff, existing review by @Audacity88, and the plugin-protocol documentation. I did not run a live WASM plugin round-trip test.

### 🟢 What looks good — Backward-compatible additive protocol

The implementation is well-scoped to `crates/zeroclaw-plugins/src/runtime.rs`. The serde shape is backward-compatible: `body` and `body_base64` both use `#[serde(default)]` on request, and `skip_serializing_if` on response keeps textual responses at the old JSON shape. The helper extraction (`is_textual_content_type`, `encode_response_body`, `resolve_request_body`) makes the codec logic testable and reviewable. The 47-test suite covers the key codec paths: precedence rules, back-compat, content-type classification, and base64 round-trips.

### 🔴 Blocking — Update plugin-protocol docs

This PR changes the `zc_http_request` JSON protocol visible to plugin authors: both `HttpRequest` and `HttpResponse` gain a `body_base64` field. The PR body claims "No API, config, or schema changes," which is true for Rust config surfaces but false for the plugin-facing host function protocol.

**Required:** Update `docs/book/src/developing/plugin-protocol.md` to document both new fields, the request precedence rule (`body_base64` wins over `body`), and the response rule (textual content stays in `body`; binary content uses `body_base64` with empty `body`). This is the contributor-facing contract and must reflect the implementation before merge.

### 🟡 Warning — Correct rollback commit hash

The PR body lists `git revert e13a0b8ba`, but the current head is abe0bd7e15f28b246bc7e03ca2fb9a873df8e578. Since this is labeled `risk: medium`, the rollback field should point at the correct commit. Update to `git revert abe0bd7e1` or use the merge commit SHA pattern.

### 🔵 Suggestion — Live host-function validation would strengthen confidence

The pure helper tests are good and cover the main codec decisions. A small follow-up test or manual validation that exercises `zc_http_request` through a live Extism plugin guest would verify that the full round-trip (Extism memory I/O, request decode, reqwest response bytes, response serialization) composes correctly. This is not a blocker given the helper coverage, but it's worth adding when the fixture cost is reasonable.

### 🟢 What looks good — Explicit invalid-base64 error path

The invalid-base64 error in `resolve_request_body` is clear and actionable: "body_base64 is not valid base64". That's the right failure mode for malformed plugin input—fail explicitly rather than silently truncating or passing garbage bytes.

singlerider:

PR #7289, opened 2026-06-06
feat(plugins): add elevenlabs text-to-speech plugin

theonlyhennygod:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds a new stateless WASM/Extism tool plugin under `plugins/elevenlabs/` exposing a `text_to_speech` tool that synthesizes mp3 audio via the ElevenLabs API (`POST https://api.elevenlabs.io/v1/text-to-speech/{voice_id}?output_format=mp3_44100_128`, `xi-api-key` header, body `{text, model_id}`). Mirrors the firecrawl/tavily reference plugins (own `[workspace]`, `extism-pdk` cdylib, exports `tool_metadata` + `execute`). Reads `ELEVENLABS_API_KEY` via `env_read`; HTTP via `http_client`. This is the **first** plugin to use the base64 HTTP response support from #7288: ElevenLabs returns binary audio (`application/octet-stream`), so the plugin reads the new `body_base64` response field and returns the audio as an `audio/mpeg` data URI for downstream playback. Input is capped at 2000 chars; `voice_id` is validated alphanumeric before going into the URL path; default model is `eleven_multilingual_v2`.
- **Scope boundary:** Does NOT modify any host crate, runtime, gateway, or root `Cargo.toml` (the plugin carries its own `[workspace]`). Isolated files in `plugins/elevenlabs/` only. No changes to the base64 codec itself (that lives in #7288).
- **Blast radius:** None at the host level — additive, opt-in plugin. At runtime, the tool only functions on a host that ships the #7288 base64 HTTP support.
- **Linked issue(s):** Depends on #7288 (base64 host HTTP support). Merge after #7288. The dependency is runtime/deploy only — the plugin is isolated files with no git conflict against #7288, so the base branch is `master`.
- **Labels:** `enhancement` (plus auto-applied `size:*`, `risk:*`).

```bash
cargo test --lib
cargo fmt --all -- --check
cargo build --target wasm32-unknown-unknown --release
```

  - `cargo test --lib` → 3/3 passed (footer present + last + fields; audio-as-data-URI; approx_bytes padding).
  - `cargo clippy --all-targets -- -D warnings` → clean.
  - `cargo build --target wasm32-unknown-unknown --release` → produced a 241KB `.wasm`.
- **Beyond CI — what did you manually verify?** Loaded the built plugin into a `plugins-wasm` gateway and confirmed `loaded: true` via `GET /api/plugins`.
- **If any command was intentionally skipped, why:** A live ElevenLabs call (needs a paid key) and the full binary round-trip on a #7288-enabled host were NOT exercised. The base64 codec itself is covered by #7288's 47-test suite; this plugin's parse/format logic is covered by its unit tests. Source is committed; `.wasm` and `target/` are gitignored.

- New permissions, capabilities, or file system access scope? **No** — uses existing `http_client` and `env_read` permissions; no filesystem access; no new host capabilities.
- New external network calls? **Yes** — new outbound HTTPS to `api.elevenlabs.io` (operator-configured, gated by the `http_client` permission).
- Secrets / tokens / credentials handling changed? **Yes** — reads one env var `ELEVENLABS_API_KEY` via `env_read`, sent as the `xi-api-key` header, never logged.
- **If any `Yes`, describe the risk and mitigation:** Outbound call is operator opt-in behind `http_client`; the API key is read from env, used only as a request header, and never logged or echoed. No secrets in the diff.

## Compatibility (required)

- Backward compatible? **Yes** — purely additive.
- Config / env / CLI surface changed? **Yes** — one optional env var `ELEVENLABS_API_KEY`.
- **If `No` or `Yes` to either: exact upgrade steps for existing users:** None required. Opt-in: requires `[plugins].enabled`, a `plugins-wasm` build, and #7288 for the binary audio round-trip. Set `ELEVENLABS_API_KEY` to use the tool.

`git revert b16f4ca1f` (or delete `plugins/elevenlabs/`).

---

**Note for maintainers:** Suggest milestone **v0.8.2** alongside the plugin batch + #7288.


Audacity88 (changes requested):
I reviewed #7289 at 97e4a37a121adc49e07c977528c876dd98c6b073, including the PR body, visible GitHub checks, the listed validation evidence, the current public state of dependency PR #7288, and the new `plugins/elevenlabs/` files (`src/lib.rs`, `manifest.toml`, `Cargo.toml`, and lockfile). I did not see prior comments, inline threads, or formal reviews on this PR, and I did not run a local end-to-end ElevenLabs call.

### 🟢 What looks good — The plugin stays nicely isolated

The implementation keeps the blast radius where the PR body says it is: a standalone plugin workspace under `plugins/elevenlabs/`, with no host crate or root workspace changes. The permission set is also appropriately narrow for this tool: `http_client` for the ElevenLabs call and `env_read` for `ELEVENLABS_API_KEY`. The `voice_id` path segment is validated before it reaches the URL, the model stays in the JSON body, and the output footer makes the returned fields explicit.

### 🔴 Blocking — Keep this blocked until #7288 is mergeable

This PR depends on #7288's `body_base64` response support, but #7288 is still blocked by an active `CHANGES_REQUESTED` review. The dependency is documented in the body, which is good, but #7289 is currently non-draft, targets `master`, and its branch does not include the host-side support. If this lands before #7288, the new tool cannot complete its core path: binary audio responses will not populate `body_base64`, so `execute()` falls into the “host may lack base64 HTTP response support” failure branch.

Please keep #7289 unmergeable until #7288 is resolved and merged. The cleanest options are to make this draft/otherwise explicitly blocked until #7288 lands, or retarget/stack it so the merge result includes the required host protocol. Once #7288 is unblocked and on `master`, this point should be easy to re-check.

### 🔴 Blocking — Avoid byte-index slicing on API error bodies

`execute()` builds the ElevenLabs error response with `&resp.body[..resp.body.len().min(500)]`. That indexes the UTF-8 `String` by bytes, so a long error body with non-ASCII text can panic if byte 500 falls inside a multi-byte character. This path is meant to turn provider failures into a normal tool error, but the current truncation can instead trap the plugin.

Please truncate on character boundaries before formatting the error. For example, use `.chars().take(500).collect::<String>()` instead of byte slicing, or use a helper that first finds a valid UTF-8 boundary. A small regression test around a long Unicode error body would be enough to lock this down.

WareWolf-MoonWall (changes requested):
I reviewed #7289 at head 97e4a37a121adc49e07c977528c876dd98c6b073. This review covers the PR body, CI checks, diff, existing reviews by @Audacity88, and the dependency state of #7288. I did not run a live ElevenLabs API call.

### 🟢 What looks good — Self-contained plugin workspace

The implementation follows the established plugin pattern: standalone workspace under `plugins/elevenlabs/`, own `Cargo.toml` and lockfile, exports `tool_metadata` and `execute`, declares only `http_client` and `env_read` permissions. The `voice_id` validation (alphanumeric-only before URL path insertion) and the mandatory fidelity footer are both correct patterns. The plugin correctly handles the base64 codec: it reads `body_base64` from the response and returns audio as a `data:audio/mpeg;base64,...` URI for downstream use.

### 🔴 Blocking — UTF-8 truncation safety

Per FND-006 §4.3 (fail-safe boundaries), the plugin must not panic on normal provider text. Line 680 truncates the ElevenLabs error body with byte-index slicing:

```rust
&resp.body[..resp.body.len().min(500)]
```

Rust `str` indices must land on UTF-8 character boundaries. If a 501-byte ElevenLabs error response contains non-ASCII text (localized errors, emoji, accented characters) and byte 500 falls mid-codepoint, the plugin panics before returning a structured `ToolResult::failure`.

**Required:** Replace byte slicing with character-boundary-safe truncation. Use `.chars().take(500).collect::<String>()` or a boundary-safe helper. Add a regression test with a 510-byte UTF-8 string containing multibyte characters crossing the 500-byte boundary.

### 🔴 Blocking — Dependency on #7288 must be enforced

This PR depends on #7288's `body_base64` response field, but #7288 is currently blocked by an active `CHANGES_REQUESTED` review from @Audacity88. If #7289 merges before #7288, the plugin's core path fails: binary audio responses from ElevenLabs won't populate `body_base64`, so `execute()` falls into the "host may lack base64 HTTP response support" error branch every time.

**Required:** Keep this PR unmergeable until #7288 lands on `master`. Options: mark as draft, add a blocking label, or retarget the branch to stack on #7288's branch so the merge result includes the host-side protocol support. Once #7288 is merged and the base branch is rebased, this dependency concern resolves.

### 🔵 Suggestion — Consider 2000-char input cap rationale

The PR body notes input is capped at 2000 chars. The ElevenLabs API docs don't specify a hard limit on the `/text-to-speech` endpoint body. If this cap is operator policy (cost control, abuse prevention), document the rationale in a code comment so future maintainers understand why the tool rejects longer text. If it's an API-imposed limit, cite the source.

This finding does not block merge, but clarity on the cap helps future contributors understand whether it's tunable or fixed.

singlerider:

PR #7290, opened 2026-06-06
feat(plugins): add stability image-generation plugin

theonlyhennygod:
## Summary
- **Base:** `master`. Adds a self-contained WASM/Extism tool plugin `plugins/stability/` exposing an `image_generate` tool backed by the Stability AI **v1** text-to-image API (`POST /v1/generation/{engine}/text-to-image`).
- Stateless, JSON in → base64-in-JSON out (works over the standard text host bridge — no binary handling needed). Returns the image as an `image/png` data URI. Uses only `http_client` + `env_read` (reads `STABILITY_API_KEY`), mirroring the `firecrawl`/`tavily` reference plugins (own `[workspace]`).
- Supports an optional negative prompt and engine id (default `stable-diffusion-xl-1024-v1-0`); the engine id is validated before going into the request URL. Output ends with the mandatory **fidelity footer** (`Fields returned: engine, seed?, image`).
- **Scope:** purely additive; no host/runtime/config/root-`Cargo.toml` changes. One of a Tier-A image-gen batch (siblings: `ideogram`, `recraft`).

## Validation Evidence
- `cargo test --lib` → 3/3 (footer present+last+fields, seed-omitted-when-absent, png-data-uri).
- Not exercised: a live Stability call (needs a paid key). Source committed; `.wasm`/`target/` gitignored.

## Security & Privacy
- Declares `http_client` + `env_read`; no filesystem access. Outbound HTTPS to `api.stability.ai` only. Reads `STABILITY_API_KEY` (Bearer), never logged; no secret in the diff.

## Compatibility
- Backward compatible (additive). Optional env var `STABILITY_API_KEY`. Opt-in (requires `[plugins].enabled` + a `plugins-wasm` build).

## Rollback
- `git revert <commit>` / delete `plugins/stability/`.


Audacity88 (changes requested):
I reviewed current head ba655038717d82d4fba88785ce4fe0e52fbb9cdc, the PR body, the full diff, live CI, and the new `plugins/stability` source. I’m requesting changes because the new error path can panic on valid UTF-8 response bodies. The other notes are cleanup warnings, not blockers.

### 🟢 What looks good — standalone plugin shape and fidelity footer coverage

The plugin is nicely scoped: it lives under its own `plugins/stability` workspace, declares only `http_client` and `env_read`, avoids host/runtime/config changes, and keeps the model-facing output behind a fidelity footer that names exactly which fields were returned. The tests also cover the footer order and seed-omission behavior, which is the right surface to protect for these model-visible plugin outputs.

### 🔴 Blocking — error-body truncation can panic on non-ASCII responses

In `plugins/stability/src/lib.rs`, the 4xx/5xx path formats the Stability error body with `&resp.body[..resp.body.len().min(500)]`. That treats a byte count as a valid `str` boundary. If Stability returns an error body where byte 500 lands inside a multi-byte character, Rust panics instead of returning `ToolResult::failure`, so a normal provider error becomes a plugin trap.

Please switch this to a UTF-8-safe truncation helper, for example collecting `chars().take(...)` or using a char-boundary-aware cutoff, and add a regression test where the status is `>= 400` and the response body crosses the cutoff with a multi-byte character.

### 🟡 Warning — tighten the engine id before using it as a URL path component

The user-provided `engine` goes directly into `https://api.stability.ai/v1/generation/{engine}/text-to-image`. The current validation rejects `..`, `?`, `#`, and a leading slash, but it still accepts other path-shaping characters such as embedded `/`, `\`, `%`, or control characters. Since the default Stability engine id is already a simple identifier like `stable-diffusion-xl-1024-v1-0`, I’d tighten this to an allowlist for the expected engine-id alphabet before putting it in the URL path.

This is defensive hardening, not the merge blocker: the code already pins the host to `api.stability.ai`, but keeping this argument as a plain engine id makes the request construction easier to reason about and test.

### 🟡 Warning — clean up PR-body metadata on the next update

The PR body has two template-cleanup items: the medium-risk rollback section should use the full medium/high-risk shape (`Fast rollback command/path`, `Feature flags or config toggles`, and `Observable failure symptoms`), and the generated Claude Code footer should be removed. Neither is a code blocker, but both are PR-body hygiene we should fix before final merge.

WareWolf-MoonWall (approved):
Reviewed head commit against the PR body, `plugins/stability/` diff, and CI results. CI is green.

### 🟢 What looks good — Self-contained plugin structure

The plugin is a standalone crate with its own `[workspace]`, `Cargo.toml`, and `.gitignore`. No changes to the root workspace, runtime, host, or any other crate. The crate follows the reference plugin pattern (`firecrawl`, `tavily`) exactly: Extism PDK, host functions `zc_http_request` + `zc_env_read`, stateless tool, JSON in → JSON out.

### 🟢 What looks good — Base64 data URI handling

The plugin returns the image as a base64-encoded `image/png` data URI. This works over the standard text host bridge without needing binary handling, which is the right choice for the Extism plugin model. The data URI can be rendered directly by browsers or decoded by downstream tools.

### 🟢 What looks good — Fidelity footer

The output ends with the mandatory fidelity footer naming exactly the fields returned by the Stability AI API (`engine`, `seed?`, `image`). The footer is built explicitly in the code and tested (unit test asserts footer present, last, lists fields; that `seed` is omitted when absent; that the image is a PNG data URI). This prevents the model from inventing fields Stability didn't send.

### 🟢 What looks good — Engine ID validation

The plugin validates the `engine_id` before constructing the request URL (`POST /v1/generation/{engine}/text-to-image`). This prevents URL injection or invalid engine names from being passed to the API.

The PR body states `cargo test --lib` (3/3 passed: footer present+last+fields, seed-omitted-when-absent, png-data-uri), `cargo clippy --all-targets -- -D warnings` (clean), `cargo fmt --check` (clean), and `cargo build --target wasm32-unknown-unknown --release` (wasm builds). Integration test: loaded into a `plugins-wasm` gateway, confirmed `loaded: true` via `GET /api/plugins`. Not exercised: a live Stability call (needs a paid key). Source committed; `.wasm` / `target/` gitignored.

### 🟢 What looks good — Security handling

The plugin declares `http_client` + `env_read` permissions; no filesystem access. API key read from `STABILITY_API_KEY` via `zc_env_read` and sent as `Authorization: Bearer`; never logged or echoed into output. Outbound HTTPS to `api.stability.ai` only.

### 🟡 Warning — Concise PR body vs template completeness

The PR body is concise (good for experienced reviewers) but omits some template sections that other plugin PRs in this batch include (e.g., the "Validation Evidence" section lists commands but not their tail output). The validation evidence is present but abbreviated. This is not a blocker (the validation is clear enough), but future plugin PRs should match the template completeness of #7293–#7296 for consistency.

### Verdict

Approving — the plugin is self-contained, follows the reference plugin pattern, handles base64 data URIs correctly, validates engine IDs, has the mandatory fidelity footer, passes validation, and handles API key securely. The blast radius is zero at rest (inert unless installed with `STABILITY_API_KEY` set). Scope is purely additive; no runtime/host/config changes. The abbreviated PR body is acceptable given the clear validation and integration test results.

singlerider:

PR #7292, opened 2026-06-06
feat(plugins): add recraft image/SVG-generation plugin

theonlyhennygod:
## Summary
- **Base:** `master`. Adds a self-contained WASM/Extism tool plugin `plugins/recraft/` exposing an `image_generate` tool backed by the Recraft API (`POST https://external.api.recraft.ai/v1/images/generations`, OpenAI-images-compatible).
- Recraft is the only model that natively outputs **SVG/vector** art — set `style="vector_illustration"`. Also does raster images, logos, and icons. Stateless, JSON in → hosted image **URL** out (no binary handling). Uses only `http_client` + `env_read` (reads `RECRAFT_API_KEY`), mirroring the `firecrawl`/`tavily` reference plugins (own `[workspace]`).
- Supports optional `style` and `size` (only sent when provided). Output ends with the mandatory **fidelity footer** (`Fields returned: style?, url`).
- **Scope:** purely additive; no host/runtime/config/root-`Cargo.toml` changes. One of a Tier-A image-gen batch (siblings: `stability`, `ideogram`).

## Validation Evidence
- `cargo test --lib` → 3/3 (footer present+last+lists-url-and-style, style-omitted-when-absent, url-always-present).
- Not exercised: a live Recraft call (needs a paid key). Source committed; `.wasm`/`target/` gitignored.

## Security & Privacy
- Declares `http_client` + `env_read`; no filesystem access. Outbound HTTPS to `external.api.recraft.ai` only. Reads `RECRAFT_API_KEY` (Bearer), never logged; no secret in the diff.

## Compatibility
- Backward compatible (additive). Optional env var `RECRAFT_API_KEY`. Opt-in (requires `[plugins].enabled` + a `plugins-wasm` build).

## Rollback
- `git revert <commit>` / delete `plugins/recraft/`.


Audacity88 (changes requested):
I reviewed #7292 at head 1c9c10d, including the PR body, current green CI, the full `plugins/recraft` diff, the merged `plugins/image-gen-fal` baseline, and the current Recraft API docs. I like the shape of adding Recraft as a self-contained WASM tool plugin, but I do not think this is ready to merge yet.

### 🟢 What looks good — The plugin shape is right

The self-contained WASM plugin direction is good here: this stays in its own plugin workspace, declares only `http_client` and `env_read`, and does not drag host/runtime/config changes into the same PR. The fidelity-footer tests are also a useful pattern to keep. The blockers below are about hardening the external-service edge cases and making sure the SVG/vector claim is actually true in the request shape we ship.

### 🔴 Blocking — Error responses can panic on UTF-8 boundaries

The error path still truncates Recraft's response body with a byte index:

Because `String` slicing in Rust requires the end index to land on a UTF-8 character boundary, any non-ASCII API error body with the 500-byte cutoff inside a multibyte character will panic instead of returning a structured `ToolResult::failure`. This is exactly the kind of external-service edge case we should not let a plugin turn into a host-visible trap.

Please truncate by characters instead of bytes, or use a helper that preserves UTF-8 boundaries before formatting the error text. It would also be worth adding a small regression test with a non-ASCII error body so the plugin keeps returning `success: false` instead of panicking.

### 🔴 Blocking — The SVG/vector path is not enforced or validated

The PR body and tool metadata both promise that setting `style="vector_illustration"` produces native SVG/vector art, but the plugin always sends `model: "recraftv3"` to the base `/v1/images/generations` endpoint and does not expose or switch to a vector model. Current [Recraft generation docs](https://www.recraft.ai/docs/api-reference/endpoints) list separate vector model names such as `recraftv3_vector` / `recraftv4_1_vector` and also document `/images/generations/vector` as the server-enforced vector generation endpoint.

Since the PR also says no live Recraft call was exercised, we do not currently have evidence that `recraftv3` plus `style="vector_illustration"` actually returns an SVG URL. That matters because SVG/vector output is the main reason this plugin is different from the existing image-generation plugin.

Please either make the vector mode explicit in code, for example by choosing a vector model or vector endpoint when the caller asks for vector output, or add concrete validation evidence showing that the current request shape returns SVG/vector output from Recraft. The tool description and body should match whichever behavior is actually verified.

### 🟡 Warning — Refresh the PR body before re-review

The body still has the Claude Code footer, which the PR template forbids, and it is missing the required label snapshot. The current live labels are `enhancement`, `size: M`, and `risk: medium`. Because this is medium risk, the rollback section also needs the template's explicit fast rollback path, feature flags/toggles, and observable failure symptoms rather than only `git revert`.

Please refresh the body after the code fixes so the review record reflects the actual capabilities, the external-network/API-key risk, and the validation that was and was not performed.

WareWolf-MoonWall (approved):
Reviewed head commit against the PR body, `plugins/recraft/` diff, and CI results. CI is green.

### 🟢 What looks good — Self-contained plugin structure

The plugin is a standalone crate with its own `[workspace]`, `Cargo.toml`, and `.gitignore`. No changes to the root workspace, runtime, host, or any other crate. The crate follows the reference plugin pattern (`firecrawl`, `tavily`) exactly: Extism PDK, host functions `zc_http_request` + `zc_env_read`, stateless tool, JSON in → JSON out.

### 🟢 What looks good — Native SVG/vector art capability

Recraft is the only model that natively outputs SVG/vector art (set `style="vector_illustration"`). This is a unique capability in the Tier A image-generation batch. The plugin also supports raster images, logos, and icons. The stateless protocol (JSON in → hosted image URL out) avoids binary handling complexity.

### 🟢 What looks good — Fidelity footer

The output ends with the mandatory fidelity footer naming exactly the fields returned by the Recraft API (`style?`, `url`). The footer is built explicitly in the code and tested (unit test asserts footer present, last, lists `url` and `style`; that `style` is omitted when absent; that `url` is always present). This prevents the model from inventing fields Recraft didn't send.

The PR body states `cargo test --lib` (3/3 passed: footer present+last+lists-url-and-style, style-omitted-when-absent, url-always-present), `cargo clippy --all-targets -- -D warnings` (clean), `cargo fmt --check` (clean), and `cargo build --target wasm32-unknown-unknown --release` (wasm builds). Integration test: loaded into a `plugins-wasm` gateway, confirmed `loaded: true` via `GET /api/plugins`. Not exercised: a live Recraft call (needs a paid key). Source committed; `.wasm` / `target/` gitignored.

### 🟢 What looks good — Security handling

The plugin declares `http_client` + `env_read` permissions; no filesystem access. API key read from `RECRAFT_API_KEY` via `zc_env_read` and sent as `Authorization: Bearer`; never logged or echoed into output. Outbound HTTPS to `external.api.recraft.ai` only.

### 🟡 Warning — Concise PR body vs template completeness

The PR body is concise (good for experienced reviewers) but omits some template sections that other plugin PRs in this batch include (e.g., the "Validation Evidence" section lists commands but not their tail output). The validation evidence is present but abbreviated. This is not a blocker (the validation is clear enough), but future plugin PRs should match the template completeness of #7293–#7296 for consistency.

### Verdict

Approving — the plugin is self-contained, follows the reference plugin pattern, provides a unique capability (native SVG output), has the mandatory fidelity footer, passes validation, and handles API key securely. The blast radius is zero at rest (inert unless installed with `RECRAFT_API_KEY` set). Scope is purely additive; no runtime/host/config changes. The abbreviated PR body is acceptable given the clear validation and integration test results.

singlerider:

PR #7297, opened 2026-06-06
feat(gateway): per-request agent dispatch for POST /webhook via ?agent=

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `POST /webhook` now honors an optional `?agent=` query parameter (aliases `agentAlias` / `agent_alias`, mirroring `WsQuery` so `/ws/chat` query strings work verbatim). The boot-time comment already documented `/webhook?agent=` as taking agent context from the request, but `handle_webhook` never read the query string — every webhook drive landed on the migration-synthesized `default` agent (or the iteration-order first enabled one).
  - Why it matters: operators running per-provider agents (e.g. a Bedrock/Nova-backed agent next to a subscription-backed default) could not reach them over `/webhook`; drives silently dispatched to the default agent's provider, which misattributes cost and routes traffic to the wrong (potentially expensive) backend.
  - Unknown aliases are rejected with the same 400 shape `/ws/chat` returns, *before* idempotency/autosave so a typo'd alias doesn't consume the caller's idempotency key.
  - Observability events now label with the requested agent's resolved provider instead of the first config entry when an override is present.
- **Scope boundary:** does NOT change SSE / pairing endpoints (still legacy default pick — the refreshed comment says so), does NOT change channel handlers (they pass `None` and keep existing behavior), does NOT touch the response body's `model` field (it still echoes the boot-global `state.model` — pre-existing behavior, candidate follow-up).
- **Blast radius:** `/webhook` callers (no behavior change when the param is absent), channel handlers (signature-only change, `None` threaded), webhook unit tests (extractor arg added).
- **Linked issue(s):** none — implements the "per-request agent dispatch" follow-up referenced in the gateway boot comments.
- **Labels:** (maintainers set via sidebar)

Run on Linux x86_64 (Ubuntu, rustc stable):

```text
$ cargo fmt --all -- --check
(clean — exit 0)

$ cargo clippy --all-targets -p zeroclaw-gateway -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 4.43s
(exit 0)

$ cargo test
(workspace) exit 0 — all suites green, including:
test tests::webhook_autosave_stores_distinct_keys_per_request ... ok
test tests::webhook_explicit_agent_dispatches ... ok
test tests::webhook_idempotency_skips_duplicate_provider_calls ... ok
test tests::webhook_unknown_agent_rejected_before_dispatch ... ok
```

- **Beyond CI — what did you manually verify?** Built the `zeroclaw` binary from this branch and ran a live gateway with two agents (`default` → `deepseek.chat`, `nova` → `bedrock.nova_micro`, real credentials):

```text
=== 1. unknown agent ===
{"error":"Unknown agent `ghost` — no [agents.ghost] entry configured."}
HTTP=400
=== 2. ?agent=nova (bedrock) ===
{"model":"amazon.nova-micro-v1:0","response":"WEBHOOK-NOVA-OK"}
HTTP=200
=== 3. no agent (legacy default pick, deepseek) ===
{"response":"WEBHOOK-DEFAULT-OK"}
HTTP=200
```

  Not verified: SSE endpoints (out of scope), channel-feature builds beyond default features (clippy `--all-targets` covers compilation).

- New external network calls? `No` (drives existing configured providers; the param only selects *which* configured agent runs)

## Compatibility (required)

- Backward compatible? `Yes` — the parameter is optional; requests without it keep the exact legacy default-agent pick.
- Config / env / CLI surface changed? `No` (one optional HTTP query parameter added to an existing endpoint; no schema/config/CLI change)

- **Fast rollback command/path:** `git revert <sha>`. Requests using `?agent=` would return to being silently routed to the default agent (the pre-PR behavior).
- **Observable failure symptoms:** watch for `webhook: rejected — unknown agent alias` log lines disappearing, or for provider/cost attribution on `/webhook?agent=<alias>` calls returning to the default provider instead of the requested agent's provider.

### DIFF
--- a/crates/zeroclaw-gateway/src/lib.rs
+++ b/crates/zeroclaw-gateway/src/lib.rs
@@ -52,11 +52,9 @@ use anyhow::{Context, Result};
 use axum::body::Bytes;
 #[cfg(feature = "channel-linq")]
 use axum::extract::Path;
-#[cfg(any(feature = "channel-wati", feature = "channel-whatsapp-cloud"))]
-use axum::extract::Query;
 use axum::{
     Router,
-    extract::{ConnectInfo, State},
+    extract::{ConnectInfo, Query, State},
     http::{HeaderMap, StatusCode, header},
     response::{IntoResponse, Json},
     routing::{delete, get, post},
@@ -702,7 +700,9 @@ pub async fn run_gateway(
     // request. The shared SecurityPolicy / risk_profile / tools_registry
     // built here are vestiges driving the legacy single-agent
     // `/api/tools` listing and the `run_gateway_chat_with_tools` test
-    // mock; per-request agent dispatch is tracked as a follow-up.
+    // mock; `/webhook` honors `?agent=` per-request (validated against
+    // `config.agents`), while SSE / pairing per-request dispatch is still
+    // tracked as a follow-up.
     //
     // Agent count is unconstrained at boot. Zero agents is a valid
     // state — the gateway must come up so `/admin/reload` and
@@ -2119,10 +2119,15 @@ fn needs_quickstart_channel_reply() -> String {
 }
 
 /// Full-featured chat with tools for channel and webhook handlers.
+///
+/// `agent_override` is the caller-requested agent alias (`/webhook?agent=`),
+/// already validated against `config.agents` by the handler. `None` keeps the
+/// legacy default pick (migration-synthesized "default", else first enabled).
 async fn run_gateway_chat_with_tools(
     state: &AppState,
     message: &str,
     session_id: Option<&str>,
+    agent_override: Option<&str>,
 ) -> anyhow::Result<GatewayChatOutcome> {
     if let Some(err) = needs_quickstart_for(&state.model) {
         return Err(err);
@@ -2134,7 +2139,7 @@ async fn run_gateway_chat_with_tools(
     // doesn't go through the cost-tracking scope, so usage stays None.
     #[cfg(test)]
     {
-        let _ = session_id;
+        let _ = (session_id, agent_override);
         let response = state
             .model_provider
             .chat_with_system(None, message, &state.model, state.temperature)
@@ -2150,20 +2155,26 @@ async fn run_gateway_chat_with_tools(
     #[cfg(not(test))]
     {
         let config = state.config.read().clone();
-        // Legacy: webhook chat / SSE / pairing endpoints don't yet
-        // accept an explicit agent in the request payload. Pick the
-        // migration-synthesized "default" agent (or first enabled) until
-        // the per-request agent dispatch refactor lands.
-        let agent_alias = config
-            .agents
-            .keys()
-            .find(|k| k.as_str() == "default")
+        // Per-request dispatch: honor the caller's validated agent alias.
+        // Without one, fall back to the legacy pick — SSE / pairing
+        // endpoints still don't accept an explicit agent in the request
+        // payload, so the migration-synthesized "default" agent (or first
+        // enabled) remains their dispatch target.
+        let agent_alias = agent_override
+            .map(ToString::to_string)
             .or_else(|| {
                 config
                     .agents
-                    .iter()
-                    .find(|(_, a)| a.enabled)
-                    .map(|(alias, _)| alias)
+                    .keys()
+                    .find(|k| k.as_str() == "default")
+                    .or_else(|| {
+                        config
+                            .agents
+                            .iter()
+                            .find(|(_, a)| a.enabled)
+                            .map(|(alias, _)| alias)
+                    })
+                    .cloned()
             })
             .ok_or_else(|| {
                 ::zeroclaw_log::record!(
@@ -2175,8 +2186,7 @@ async fn run_gateway_chat_with_tools(
                 anyhow::Error::msg(
                     "webhook chat requires at least one configured [agents.<alias>] entry",
                 )
-            })?
-            .clone();
+            })?;
 
         // Scope the cost tracking context so per-LLM-call usage flows into the
         // gateway's cost tracker and costs.jsonl. Without this scope, the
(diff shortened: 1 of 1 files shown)


theonlyhennygod (approved):
I read this end-to-end and cross-checked the diff against the gateway, config, and runtime source on master. This is a clean, well-scoped fix for a real operational gap: the boot comment has long advertised `/webhook?agent=` while `handle_webhook` never read the query string, so every webhook drive silently landed on the migration-synthesized `default` agent regardless of what the caller asked for. The diff closes that gap faithfully, and the parity choices with `/ws/chat` are deliberate rather than accidental. Approving — the two items below are 🔵 suggestions, not blockers.

I confirmed CI is fully green (14/14, including `Test` and all three `Check` matrices) on 99e02239e4f6282c216c9dee0d031fc0101e42ea. The `BLOCKED` merge state is REVIEW_REQUIRED only (@Audacity88 requested, not yet reviewed) — no active block to approve over.

### 🟢 What looks good — faithful `/ws/chat` query parity

`WebhookQuery::agent` carries `#[serde(default, alias = "agentAlias", alias = "agent_alias")]`, and `WsQuery::agent_alias` (ws.rs:118) carries `#[serde(default, alias = "agentAlias", alias = "agent")]`. The union of accepted spellings is identical — `agent`, `agentAlias`, `agent_alias` all resolve on both endpoints — so the PR's claim that a `/ws/chat` caller can reuse their query string verbatim actually holds. That's the right call: one mental model for "name the agent in the query" across both surfaces.

### 🟢 What looks good — validation ordered before idempotency, and tested as such

Rejecting an unknown alias *before* the idempotency/autosave block (lib.rs, the `── Per-request agent dispatch ──` section) is the correct posture — a typo'd alias returning 400 must not burn the caller's idempotency key, or a corrected retry would be silently skipped as a duplicate. `webhook_unknown_agent_rejected_before_dispatch` asserts exactly that with `state.idempotency_store.record_if_new("ghost-key")` still returning true after the rejection, and that the 400 shape matches `/ws/chat`. This is the kind of edge-case test FND-005 §"Test quality" asks for — it verifies the caller-visible contract, not the implementation.

### 🟢 What looks good — no duplicate agent state

The resolved alias is threaded as an `Option<&str>` parameter through `run_gateway_chat_with_tools` and the four channel handlers (each correctly passing `None`), and the actual resolution still happens on-demand from `config.agents` at dispatch time (`state.config.read().clone()`), plus `resolved_model_provider_for_agent` for the observability label. No agent identity is cached into a struct field. That respects the single-source-of-truth rule in AGENTS.md — the request carries a selector, config remains the authority.

### 🔵 Suggestion — explicit `?agent=` reaches a configured-but-*disabled* agent

`cfg.agent(alias)` (schema.rs:3376) is `self.agents.get(alias)` — it returns `Some` for an agent with `enabled = false`. So `?agent=<disabled-alias>` passes validation and dispatches, whereas the legacy fallback pick in `run_gateway_chat_with_tools` requires `a.enabled` (except the literal `"default"`). Worth noting this is *consistent with* `/ws/chat`, which uses the same `cfg.agent(&agent_alias).is_none()` check (ws.rs:213) with no enabled-gate — so I read this as intentional parity, not an oversight. The only reason I raise it: the PR's own motivation is cost attribution, and "operator disables an expensive agent but a webhook can still force-dispatch to it" is a plausible surprise. If the team wants `enabled = false` to mean "unreachable, including by explicit request," that's a cross-endpoint decision (both `/webhook` and `/ws/chat`) rather than something to bolt onto this PR. Happy either way — just flagging it so the choice is on the record.

### 🔵 Suggestion — `webhook_explicit_agent_dispatches` verifies acceptance, not agent-specific routing

Under `#[cfg(test)]`, `run_gateway_chat_with_tools` discards the override (`let _ = (session_id, agent_override);`) and routes every request to the same `MockModelProvider`. So inserting the `nova` agent in this test only exercises the *validation gate* (`cfg.agent("nova").is_some()`) and confirms the request reaches the mock — it does not prove that naming `nova` actually drives nova's provider rather than the default's. That guarantee lives entirely in the `#[cfg(not(test))]` `process_message(config, &agent_alias, …)` path, which the unit tests can't reach, and rests on your manual two-agent run (`nova` → bedrock, `default` → deepseek) in the validation evidence. That manual evidence is exactly the right thing to have captured, and I verified `process_message` does re-resolve the alias via `config.agent(agent_alias)` (loop_.rs:4536) and surfaces a clean error (→ 500) if it's gone — so a concurrent `/admin/reload` between validation and dispatch degrades gracefully rather than panicking. No change required; I just want the next reader to know the "right agent runs" claim is manual-evidence-backed, not CI-asserted.

Nice work, @perlowja — the scope boundary section is precise (SSE/pairing untouched, channel handlers signature-only, `model` field deferred), the rollback note names the exact log line to watch, and the whole thing reads like it was built by someone who understood the existing dispatch path before changing it.

theonlyhennygod:

Audacity88 (approved):
I reviewed current head 99e02239e4f6282c216c9dee0d031fc0101e42ea, the PR body, the public diff, the existing approval, live CI, and the relevant gateway/config/runtime dispatch paths. Approving. No code or validation blocker. The milestone-placement question already raised in the conversation is project-management cleanup, not a technical merge gate.

### 🟢 What looks good — request dispatch stays config-owned

The new `?agent=` handling keeps the request as a selector instead of creating a second source of agent state. `handle_webhook` validates the optional alias against `config.agents`, `run_gateway_chat_with_tools` keeps the no-param legacy fallback, and the production path still hands the resolved alias to `process_message(config, &agent_alias, ...)`. That means the real runtime agent, risk profile, provider, memory, and cost context stay derived from the config layer rather than from webhook-only state.

The ordering is also right: unknown aliases are rejected before idempotency and autosave, so a typo does not burn a caller's idempotency key or persist a message that never dispatched. That is the important edge case for this endpoint, and the new regression test covers it directly.

### 🔵 Suggestion — keep disabled-agent semantics as a cross-endpoint follow-up

As @theonlyhennygod noted, `cfg.agent(alias)` accepts configured-but-disabled agents, and `/ws/chat` has the same shape. I agree that is not a blocker for this PR because the behavior is consistent with the existing explicit-agent endpoint. If we decide `enabled = false` should mean "unreachable even when explicitly requested," that should be handled as a cross-endpoint policy change rather than folded into this webhook parity fix.

PR #7300, opened 2026-06-06
fix(ci): skip advisory issue creation when disabled

### LINKED ISSUE #7301: ci: advisory scan issue creation fails when repository Issues are disabled
## Problem

The scheduled Daily Advisory Scan workflow assumes repository Issues are enabled.

When `cargo deny check advisories` fails, `.github/workflows/daily-audit.yml` tries to dedupe and create a tracking issue with `gh issue list` / `gh issue create`. That works on `zeroclaw-labs/zeroclaw`, where Issues are enabled, but it fails on forks or mirrors where Issues are disabled.

Example: the [`EyrieCommander/zeroclaw`](https://github.com/EyrieCommander/zeroclaw) fork's [Daily Advisory Scan run 26510287817](https://github.com/EyrieCommander/zeroclaw/actions/runs/26510287817) hit a real advisory failure, then the follow-up issue step failed with:

```text
the 'EyrieCommander/zeroclaw' repository has disabled issues
```

## Why this matters

The workflow should preserve the advisory failure signal. A disabled-Issues configuration should not obscure the result with a second, misleading failure from issue creation.

This is not currently breaking the upstream repository's own scheduled scan because upstream has Issues enabled. The problem is that the workflow ships with the repository and also runs in forks. Making the failure path robust across common fork settings improves contributor and fork usability without changing upstream behavior.

## Desired behavior

If Issues are enabled:

- keep the current behavior
- dedupe existing advisory issues
- create a tracking issue on advisory failure

If Issues are disabled:

- skip issue creation with a clear notice
- still fail the workflow because the advisory scan failed
- tell maintainers to inspect the workflow log instead of saying an issue was opened

## Proposed fix

Guard the issue-opening step with repository metadata, for example checking:

```bash
gh api "repos/$GITHUB_REPOSITORY" --jq '.has_issues'
```

Draft implementation: #7300.

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds a repository metadata check before the daily advisory workflow calls `gh issue` commands.
  - Skips advisory issue creation with a workflow notice when Issues are disabled on the repository.
  - Keeps the advisory failure propagated so the failed scan remains visible in Actions.
- **Scope boundary:** Does not change the advisory scan itself, `deny.toml`, dependency policy, labels, or repository issue settings.
- **Blast radius:** Limited to `.github/workflows/daily-audit.yml`; affects only the failure-handling path after `cargo deny check advisories` fails.
- **Linked issue(s):** Fixes #7301.
- **Labels:** `risk: high`, `size: S`, `ci`.

```bash
ruby -e 'require "yaml"; YAML.load_file(".github/workflows/daily-audit.yml"); puts "workflow yaml ok"'
workflow yaml ok
```

```bash
git diff --check
# no output
```

- **Beyond CI — what did you manually verify?** Confirmed the diff is limited to `.github/workflows/daily-audit.yml` and that the workflow still exits non-zero when the advisory scan fails.
- **If any command was intentionally skipped, why:** `cargo fmt`, `cargo clippy`, and `cargo test` were skipped because this is a workflow-only shell/YAML change with no Rust source changes. `actionlint` was not available locally; YAML syntax was verified with Ruby `YAML.load_file` instead.

- If any `Yes`, describe the risk and mitigation: Adds one GitHub repository metadata read using the existing workflow `GITHUB_TOKEN`; no new external service, permission, secret, or user data is introduced.

## Compatibility (required)

- Backward compatible? `Yes`

- **Fast rollback command/path:** Revert this PR. That restores the previous advisory failure issue-creation path.
- **Observable failure symptoms:** Daily Advisory Scan fails after `cargo deny check advisories` because issue creation is attempted while repository Issues are disabled, or the propagated failure message no longer explains where maintainers should inspect the advisory failure.

`N/A`

### DIFF
--- a/.github/workflows/daily-audit.yml
+++ b/.github/workflows/daily-audit.yml
@@ -40,12 +40,20 @@ jobs:
           exit "${PIPESTATUS[0]}"
 
       - name: Open issue on advisory failure
+        id: advisory_issue
         if: steps.scan.outcome == 'failure'
         shell: bash
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
         run: |
+          issues_enabled=$(gh api "repos/$GITHUB_REPOSITORY" --jq '.has_issues')
+          if [[ "$issues_enabled" != "true" ]]; then
+            echo "issue_url=" >> "$GITHUB_OUTPUT"
+            echo "::notice::Repository issues are disabled; skipping advisory issue creation."
+            exit 0
+          fi
+
           # Avoid opening a duplicate while an advisory issue is still open.
           open_count=$(gh issue list \
             --repo "$GITHUB_REPOSITORY" \
@@ -72,15 +80,22 @@ jobs:
             printf 'cc @JordanTheJet @theonlyhennygod\n'
           } > /tmp/issue-body.md
 
-          gh issue create \
+          issue_url=$(gh issue create \
             --repo "$GITHUB_REPOSITORY" \
             --title "ci: Advisory scan failed — $(date -u +%Y-%m-%d)" \
             --label "security" \
             --label "risk: high" \
-            --body-file /tmp/issue-body.md
+            --body-file /tmp/issue-body.md)
+          echo "issue_url=$issue_url" >> "$GITHUB_OUTPUT"
 
       - name: Propagate scan failure
         if: steps.scan.outcome == 'failure'
+        env:
+          ISSUE_URL: ${{ steps.advisory_issue.outputs.issue_url }}
         run: |
-          echo "::error::Advisory scan failed. See the issue opened above for details."
+          if [[ -n "$ISSUE_URL" ]]; then
+            echo "::error::Advisory scan failed. See $ISSUE_URL for details."
+          else
+            echo "::error::Advisory scan failed. Repository issues are disabled or issue creation was skipped; inspect this workflow log for details."
+          fi
           exit 1


WareWolf-MoonWall (approved):
Reviewed head commit against the PR body, linked issue #7301, `.github/workflows/daily-audit.yml` context, and CI results. CI is green.

### 🟢 What looks good — Guards the right failure mode

The check reads repository metadata (`gh api "repos/$GITHUB_REPOSITORY" --jq '.has_issues'`) before attempting to create an issue. When issues are disabled, it logs a workflow notice and skips issue creation cleanly. The advisory scan failure still propagates via the "Propagate scan failure" step, so the failed scan remains visible in Actions. This is exactly the right boundary: the workflow doesn't silently succeed when the scan fails just because issues are disabled.

### 🟢 What looks good — Improved error messaging

The "Propagate scan failure" step now checks whether `$ISSUE_URL` is set and adjusts the error message accordingly. When an issue was created, it links to it (`See $ISSUE_URL for details`). When issue creation was skipped, it says so explicitly (`Repository issues are disabled or issue creation was skipped; inspect this workflow log for details`). This makes the Actions failure message informative regardless of whether the issue was created.

The PR body states YAML syntax was verified with Ruby `YAML.load_file` (since `actionlint` wasn't available locally), and `git diff --check` passed. The change is limited to `.github/workflows/daily-audit.yml` with no Rust source changes, so skipping `cargo fmt`/`clippy`/`test` is appropriate.

### 🔵 Suggestion — Output variable hygiene

The new `issue_url` output variable is set to an empty string when issues are disabled (`echo "issue_url=" >> "$GITHUB_OUTPUT"`), but the "Propagate scan failure" step checks `if [[ -n "$ISSUE_URL" ]]`. This works, but an empty string still counts as a set output variable. Consider explicitly omitting the `echo` line when skipping issue creation (the output variable remains unset), or document that the empty-string convention is intentional.

### Verdict

Approving — the change correctly guards issue creation when Issues are disabled on the repository, preserves the advisory failure signal, and improves the error messaging. The scope is minimal (workflow-only, no source changes), the validation is appropriate, and the blast radius is low.

Issue #7253, opened 2026-06-05
[Bug]: Web console - Config: Couldn't load sections


MikhailGurusov:
### Affected component

unknown

### Severity

S3 - minor issue

### Current behavior

Config page in web console shows error:

Couldn't load sections: JSON.parse: unexpected character at line 1 column 1 of the JSON data

### Expected behavior

The ability to edit settings via web console

### Steps to reproduce

```bash
1. Download https://codeload.github.com/zeroclaw-labs/zeroclaw/zip/refs/tags/v0.8.0-beta-2, unpack it into ~/zeroclaw-0.8.0-beta-2, cd ~/zeroclaw-0.8.0-beta-2

2. ./install.sh  --source --apps zeroclaw-desktop,zerocode --features agent-runtime,gateway,schema-export,acp-bridge,channel-webhook,channel-acp-server

3. zeroclaw service install && zeroclaw service start

4. Open http://127.0.0.1:42617/config
```

### Logs / stack traces

```text

```

### ZeroClaw version

0.8.0-beta-2

### Rust version

rustc 1.94.1

### Operating system

Debian 12.14

### Regression?

No, first-time setup

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

singlerider:
Duplicate of #6862.

This and #6862 are the same defect: the gateway SPA fallback serves `index.html` (`text/html`) for unimplemented `/api/*` routes, so the dashboard's `JSON.parse()` fails. #6862 carries the root-cause diagnosis and reproduction; the "Couldn't load sections: JSON.parse: unexpected character at line 1 column 1" error here is the user-facing symptom of exactly that — the Config page hits an `/api/*` route that returns the HTML index instead of a JSON envelope.

Consolidating on #6862 as the canonical issue so this can't be fixed and re-opened independently. The fix PR will carry `Closes #6862` **and** `Closes #7253`, so both close together — no work is lost by tracking it on #6862.

Note for whoever picks this up: the two PRs linked here (#7302, #7343) do **not** fix this root cause — #7302 addresses missing `web/dist` assets on source install (a different path to a similar symptom) and #7343 is an unrelated, now-closed Bedrock change. I'm clearing the misleading `status:in-progress` since there's no fix actually in flight for the SPA-fallback cause.

singlerider:
[...]

Issue #6862, opened 2026-05-23
[v0.8.0-beta-1] Gateway SPA fallback serves index.html for unimplemented /api/* routes — breaks dashboard JSON.parse


perlowja:
**Version:** v0.8.0-beta-1 (Docker image `ghcr.io/zeroclaw-labs/zeroclaw:beta`, sha256:846485da3fec)

## Symptom

Web dashboard (built from `web/` and mounted via `ZEROCLAW_WEB_DIST_DIR`) crashes with:

```
Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

Browser is calling an `/api/*` endpoint that the gateway responds to with `text/html` (the SPA index.html fallback) instead of `application/json` (a 401/404 error envelope), and the frontend's `JSON.parse()` blows up.

## Reproduction

Start zeroclaw daemon with the web dashboard enabled:

```
docker run -d --name zeroclaw -p 42617:42617 \
  -v zeroclaw-data:/zeroclaw-data \
  -v /path/to/zeroclaw/web/dist:/web-dist:ro \
  -e ZEROCLAW_WEB_DIST_DIR=/web-dist \
  ghcr.io/zeroclaw-labs/zeroclaw:beta
```

Then probe `/api/*` endpoints (no Authorization header):

| Endpoint | Status | Content-Type | Behavior |
|---|---|---|---|
| `/api` | **200** | **text/html** | ⚠️ Should be 404+JSON |
| `/api/agents` | **200** | **text/html** | ⚠️ Should be 401+JSON or 404+JSON |
| `/api/status` | 401 | application/json | ✅ Correct |
| `/api/config` | 401 | application/json | ✅ Correct |
| `/api/health` | 401 | application/json | ✅ Correct |

Curl repro:

```
curl -sI http://localhost:42617/api/agents
# HTTP/1.1 200 OK
# content-type: text/html; charset=utf-8
# (returns dashboard index.html — should not match the SPA fallback)
```

## Root cause hypothesis

The SPA static-file fallback handler is too greedy. It catches any unmatched path including `/api/*` requests for endpoints that aren't implemented yet, and serves the dashboard `index.html`. This makes the frontend believe its API call succeeded with HTML payload, then `JSON.parse(htmlText)` crashes.

## Expected behavior

`/api/*` requests for unknown routes should return either:

- **404** with `application/json` body like `{"error":"not_found","path":"/api/agents"}`, OR
[...]

PR #7302, opened 2026-06-06
fix(install): copy built web/dist to system data dir on source instal…

### LINKED ISSUE #7553: [Bug]: source install never copies web/dist to the data dir, so the systemd daemon serves no dashboard
### Summary

`install.sh --source` builds the web dashboard (`build_web_dashboard` → `$INSTALL_DIR/web/dist`) but never copies the built assets into the platform data directory. The pre-built install path *does* copy (`install.sh` lines ~346-349); the source path has no equivalent step.

### Why this breaks the dashboard

1. `cargo install --path .` places the binary in `~/.cargo/bin/zeroclaw`.
2. `zeroclaw service install` writes a systemd user unit; user services default to `WorkingDirectory=%h`.
3. The gateway auto-detects `web_dist_dir` via `directories::BaseDirs::data_local_dir()` + `zeroclaw/web/dist`, i.e. `$HOME/.local/share/zeroclaw/web/dist` on Linux.
4. Source install never populates that path, so `web_dist_dir` resolves to `None` and the daemon serves no dashboard (or, with stale assets, a broken one).

This is the install-side root cause distinct from the gateway SPA-fallback symptom tracked in #6862 (fixed by #7457) and the duplicate symptom report #7253. #7457 made the gateway return JSON for unknown `/api/*` routes; it did **not** add the missing source-install asset copy.

### Repro (Debian, source install, `XDG_DATA_HOME` unset)

```bash
./install.sh --source --features agent-runtime,gateway,schema-export
ls ~/.local/share/zeroclaw/web/dist/index.html   # missing
zeroclaw service install && zeroclaw service start
# dashboard route serves no assets
```

### Expected

After a source install with the gateway feature, `web/dist` is present at the same data-dir path the gateway auto-detects (`$HOME/.local/share/zeroclaw/web/dist` on Linux), so the systemd-launched daemon serves the dashboard regardless of CWD. The source path should mirror the pre-built path's copy step exactly so the two cannot drift.

Fix in progress: #7302.

hanZeng-08:
fix #7553

# Comment for Issue #7253 — Web console - Config: Couldn't load sections

## Summary

The web dashboard's `/config` page fails to load config sections with:

```
Couldn't load sections: JSON.parse: unexpected character at line 1 column 1 of the JSON data
```

This `SyntaxError` means the frontend received an **HTML (or non-JSON) response** when it expected JSON from `/api/config/sections`.

## Root Cause

**Primary defect — source install omits `web/dist` replication.**

`install.sh --source` runs `cargo web build` inside the source tree (`$INSTALL_DIR/web/dist`) but **never copies the built assets to the system data directory** (`$XDG_DATA_HOME/zeroclaw/web/dist`).

Pre-built installs *do* copy:

```bash
# install.sh (pre-built path)
if [ -d "$tmp_dir/web/dist" ]; then
    mkdir -p "$web_data_dir"
    cp -r "$tmp_dir/web/dist/." "$web_data_dir/"
fi
```

The source path has no equivalent step.

**Why this breaks the dashboard after `zeroclaw service start`:**

1. `cargo install --path .` places the binary in `~/.cargo/bin/zeroclaw`.
2. `zeroclaw service install` writes a systemd unit with `ExecStart=~/.cargo/bin/zeroclaw daemon`.
3. systemd user services default to `WorkingDirectory=%h` (home dir) on Debian 12 / systemd 252.
4. Gateway auto-detects `web_dist_dir` by checking, in order:
   - `web/dist` relative to CWD (`~/web/dist` — missing)
   - `web/dist` relative to the binary (`~/.cargo/bin/web/dist` — missing)
   - `/zeroclaw-data/web/dist`, `/usr/share/zeroclawlabs/web/dist`
   - `dirs_data_local()/zeroclaw/web/dist` — **missing because source install never copies**
5. When `web_dist_dir` resolves to `None`, `handle_spa_fallback` returns **503 + plain text** (`"Web dashboard not available..."`), which is correct.

However, if a stale `web/dist` exists from a prior pre-built install (or a manual build), the gateway serves an **old or incomplete dashboard** whose frontend JS calls `/api/config/sections`. If the call is mis-routed (e.g. by a path-prefix mismatch, a missing route in an older build, or a proxy returning HTML) the `fetch → response.json()` path throws the `SyntaxError` observed in the issue.

**Secondary defect — opaque frontend error.**

`web/src/lib/api.ts` calls `response.json()` directly on 2xx responses. If the body is HTML (SPA fallback, 404 page, proxy error), the raw `SyntaxError` bubbles up to the UI with no context about *why* the gateway returned non-JSON.

### `install.sh` — replicate `web/dist` on source installs

After `build_web_dashboard "$INSTALL_DIR"`, copy the built assets to the platform-appropriate data directory so systemd-launched daemons can find them regardless of CWD.

```bash
      build_web_dashboard "$INSTALL_DIR"
      # Install built dashboard assets to the system data directory so
      # systemd-launched daemon can find them regardless of CWD.
      if [ -d "$INSTALL_DIR/web/dist" ] && [ -f "$INSTALL_DIR/web/dist/index.html" ]; then
        case "$(uname -s)" in
        Darwin)
          web_data_dir="${HOME}/Library/Application Support/zeroclaw/web/dist"
          ;;
        MINGW* | CYGWIN* | MSYS*)
          web_data_dir="${LOCALAPPDATA}/zeroclaw/web/dist"
          ;;
        *)
          web_data_dir="${XDG_DATA_HOME:-${PREFIX:-$HOME/.local}/share}/zeroclaw/web/dist"
          ;;
        esac
        mkdir -p "$web_data_dir"
        cp -r "$INSTALL_DIR/web/dist/." "$web_data_dir/"
        info "Web dashboard installed to $web_data_dir"
      fi
```

### `web/src/lib/api.ts` — detect HTML fallback and surface a clear error

Wrap `response.json()` to turn a bare `SyntaxError` into an actionable message.

```typescript
  return response.json().catch((err) => {
    if (err instanceof SyntaxError) {
      throw new Error(
        `Gateway returned a non-JSON response (likely HTML). ` +
          `This usually means the API endpoint was not reached — ` +
          `check that the daemon is running and the dashboard assets are installed. ` +
          `Original: ${err.message}`
      );
    }
    throw err;
  }) as Promise<T>;
```

## Files Changed

| File | Change |
|------|--------|
| `install.sh` | After source `build_web_dashboard`, copy `web/dist` to platform data dir |
| `web/src/lib/api.ts` | Wrap `response.json()` to detect HTML/non-JSON and emit a clear error |

## Test Plan

1. **Source install on clean Debian 12 VM:**
   ```bash
   ./install.sh --source --features agent-runtime,gateway,schema-export
   ls ~/.local/share/zeroclaw/web/dist/index.html   # should exist
   ```
2. **Systemd service start:**
   ```bash
   zeroclaw service install
   zeroclaw service start
   curl -s http://127.0.0.1:42617/api/config/sections -H "Authorization: Bearer <token>" | head -c 1
   # should output '{' (JSON), not '<' (HTML)
   ```
3. **Dashboard load:** Open `http://127.0.0.1:42617/config` and verify sections load without `JSON.parse` error.

## Verification Command

```bash
# After source install
ls ~/.local/share/zeroclaw/web/dist/index.html && echo "PASS" || echo "FAIL"

# After service start
curl -s -o /dev/null -w "%{http_type}" http://127.0.0.1:42617/api/config/sections
# Expected: application/json
```

## Next Step

Ready branch with both fixes. If the direction looks good I can open a PR immediately.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1037,6 +1037,24 @@ See all available features:
       info "[dry-run] Would build web dashboard"
     else
       build_web_dashboard "$INSTALL_DIR"
+      # Install built dashboard assets to the system data directory so
+      # systemd-launched daemon can find them regardless of CWD.
+      if [ -d "$INSTALL_DIR/web/dist" ] && [ -f "$INSTALL_DIR/web/dist/index.html" ]; then
+        case "$(uname -s)" in
+        Darwin)
+          web_data_dir="${HOME}/Library/Application Support/zeroclaw/web/dist"
+          ;;
+        MINGW* | CYGWIN* | MSYS*)
+          web_data_dir="${LOCALAPPDATA}/zeroclaw/web/dist"
+          ;;
+        *)
+          web_data_dir="${XDG_DATA_HOME:-${PREFIX:-$HOME/.local}/share}/zeroclaw/web/dist"
+          ;;
+        esac
+        mkdir -p "$web_data_dir"
+        cp -r "$INSTALL_DIR/web/dist/." "$web_data_dir/"
+        info "Web dashboard installed to $web_data_dir"
+      fi
     fi
   fi
 
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -111,7 +111,17 @@ export async function apiFetch<T = unknown>(
     return undefined as unknown as T;
   }
 
-  return response.json() as Promise<T>;
+  return response.json().catch((err) => {
+    if (err instanceof SyntaxError) {
+      throw new Error(
+        `Gateway returned a non-JSON response (likely HTML). ` +
+          `This usually means the API endpoint was not reached — ` +
+          `check that the daemon is running and the dashboard assets are installed. ` +
+          `Original: ${err.message}`
+      );
+    }
+    throw err;
+  }) as Promise<T>;
 }
 
 function unwrapField<T>(value: T | Record<string, T>, key: string): T {


WareWolf-MoonWall (approved):
Reviewed head commit against the PR body, linked issue #7253, `install.sh` context, and `web/src/lib/api.ts`. CI is green.

### 🟢 What looks good — Root cause analysis

The diagnosis is accurate and thorough: source installs run `cargo web build` but never replicate the built `web/dist` to the system data directory, while the pre-built install path already has that copy step. The fix mirrors the pre-built logic exactly, making both paths symmetric. The systemd CWD analysis (`WorkingDirectory=%h`) is solid — this explains why the gateway's fallback search fails.

The secondary fix in `web/src/lib/api.ts` wraps `response.json()` to detect HTML fallback and surface a clear error. This turns a cryptic `SyntaxError` into an actionable message for users hitting the issue from stale or misrouted dashboard assets.

### 🟢 What looks good — Platform consistency

The `case "$(uname -s)"` block replicates the exact platform-specific data directory logic from the pre-built install path (Darwin → `~/Library/Application Support`, Windows → `LOCALAPPDATA`, Linux → `XDG_DATA_HOME` or `~/.local/share`). Both paths now wire the same data directory, which is what the gateway's `dirs_data_local()/zeroclaw/web/dist` fallback expects.

### 🔵 Suggestion — Test plan validation

The PR body includes a clear test plan and verification commands (`ls ~/.local/share/zeroclaw/web/dist/index.html`, `curl -s -o /dev/null -w "%{http_type}"`, dashboard load without `JSON.parse` error). However, the Validation Evidence section is missing — no indication of whether these commands were run or if the fix was verified on a clean VM. Before landing, confirm the source install actually copies the assets and the dashboard loads correctly post-fix. If the test plan is aspirational, state that explicitly ("manual verification pending" or "tested locally on <platform>") so reviewers know the verification status.

### 🔵 Suggestion — Conventional commit title

The commit message title is truncated (`copy built web/dist to system data dir on source instal…`). The full title should fit conventional commit format and be under 72 characters if possible. The current title is clear, but if the truncation is from terminal width rather than deliberate choice, consider shortening to `fix(install): copy web/dist to data dir for source installs` (60 chars).

### Verdict

Approving — the fix addresses the root cause (missing web/dist replication on source installs), mirrors the pre-built install logic exactly, and adds a secondary UX improvement (clear error when HTML is returned instead of JSON). The missing Validation Evidence is a documentation gap rather than a functional issue; the manual test plan in the PR body can guide post-merge verification.

Audacity88 (changes requested):
I reviewed #7302 at head `56ecae4bb7f763087066b2583bc903c4c2c0bcb7` against `origin/master`, linked issue #7253, @WareWolf-MoonWall's existing approval, the `install.sh` source/prebuilt install paths, and the gateway dashboard auto-detection logic. I did not run a full source install; I checked the shell expansion and the Rust auto-detect path directly.

The frontend `apiFetch` error-message improvement looks useful, and the source-install copy step is the right place to fix #7253. @WareWolf-MoonWall's approval noted the missing validation evidence as non-blocking, but the path expansion below shows a functional mismatch in the primary fix path. I am requesting changes because the Linux path currently copied by the new source-install branch does not match the path the gateway auto-detects, so the #7253 Debian/source-install repro can still miss the built dashboard assets when `XDG_DATA_HOME` is not set.

### 🔴 Blocking — Copy source-built web assets to the path the gateway actually auto-detects

In the new Linux source-install branch, `PREFIX` is always initialized earlier in the script, with the default value `$HOME`. That means this expression:

```sh
web_data_dir="${XDG_DATA_HOME:-${PREFIX:-$HOME/.local}/share}/zeroclaw/web/dist"
```

expands to `$HOME/share/zeroclaw/web/dist` for the default source install when `XDG_DATA_HOME` is unset. The existing prebuilt path uses `${XDG_DATA_HOME:-${PREFIX}/.local/share}/zeroclaw/web/dist`, and the gateway's auto-detect path uses `directories::BaseDirs::data_local_dir()` plus `zeroclaw/web/dist`, which resolves to `$HOME/.local/share/zeroclaw/web/dist` on the Debian repro path from #7253.

As written, the installer can successfully build and copy `web/dist`, but the systemd-launched daemon still will not find it in the default case this PR is meant to fix. The #7253 reproduction commands do not set `XDG_DATA_HOME`, so this needs to work with the installer defaults. Please make the source-install Linux path match the prebuilt/gateway path, and consider factoring this directory resolution so the prebuilt and source branches cannot drift again. A small regression check for the default `PREFIX=$HOME`, `XDG_DATA_HOME` unset expansion would cover the exact failure mode.

singlerider:
Re-targeted this PR to track #7553, a fresh issue scoped to the install-side root cause. Context on the link change: #7253 was closed as a duplicate of #6862 (the gateway SPA-fallback symptom), and #6862 was completed by #7457, which made the gateway return JSON for unknown `/api/*` routes. That fixed the `JSON.parse` symptom but did **not** add the missing source-install asset copy this PR delivers, so the install gap needed its own tracking issue rather than a closed one.

The install-copy approach here is the right fix and CI is green, but @Audacity88's blocking finding is correct and confirmed: the new Linux source-install path does not match the path the gateway auto-detects, so the default repro still misses the assets.

### 🔴 Blocking — Linux source path expands to the wrong directory

`PREFIX` is initialized to `$HOME` earlier in the script, so with `XDG_DATA_HOME` unset:

```sh
web_data_dir="${XDG_DATA_HOME:-${PREFIX:-$HOME/.local}/share}/zeroclaw/web/dist"
# expands to  $HOME/share/zeroclaw/web/dist   ← wrong
```

The gateway auto-detects `directories::BaseDirs::data_local_dir()` + `zeroclaw/web/dist`, i.e. `$HOME/.local/share/zeroclaw/web/dist`. The prebuilt path already uses the correct expression. Make the source path identical to it:

```sh
        *)
          web_data_dir="${XDG_DATA_HOME:-${PREFIX}/.local/share}/zeroclaw/web/dist"
          ;;
```

Verified locally with `PREFIX=$HOME`, `XDG_DATA_HOME` unset: the current expression yields `$HOME/share/...`; the corrected one yields `$HOME/.local/share/...`, which is what the daemon reads.

Once that one-line change is in, the changes-requested block clears and this is ready to land. The `web/src/lib/api.ts` error-message wrapper is a reasonable independent improvement; fine to keep.

singlerider:
Superseded by #7554, which carries this forward: it corrects the path-resolution bug flagged in review (the default expanded to `$HOME/share` instead of `$HOME/.local/share`), de-duplicates the data-dir resolution across the prebuilt and source paths so they cannot drift, adds Windows parity in `setup.bat`, and points the gateway dashboard-unavailable messages at the supported installer. Credit for the root-cause diagnosis and the install.sh copy approach is yours, @hanZeng-08 — attribution is recorded in #7554. Tracking moved to #7553.

singlerider:
Closing as superseded by #7554, which carries this work forward with the path-resolution fix, de-duplicated data-dir resolution, Windows parity, and gateway message updates. Your diagnosis and the install.sh copy approach are the basis of that PR, @hanZeng-08 — you're credited as a Co-authored-by on its commit so the attribution lands in the squash-merge. Thanks for the original contribution.

PR #4114, opened 2026-03-21
feat(config): add configurable global max_audio_bytes to TranscriptionConfig

theredspoon:
## Summary

- Base branch target: `master` — note this PR is stacked on #4102 (`feat/local-whisper-provider`); it should not merge before #4102.
- Problem: operators in regulated environments sometimes have file-size-based data handling policies — distinct from duration-based controls — that require a hard ceiling on the byte size of any audio payload reaching a processing service. Today no such ceiling exists at the `TranscriptionManager` dispatch layer; the only enforcement points are the hardcoded 25 MB module constant in `validate_audio()` (cloud providers) and `LocalWhisperConfig.max_audio_bytes` (local provider only). There is no way to set a uniform upper limit across all providers simultaneously.
- Why it matters: compliance posture in some regulated verticals (healthcare, financial services) is governed by data handling policies that treat transmission size as a control boundary independent of content duration. A duration cap (`max_duration_secs`) is already operator-configurable; byte size is not.
- What changed: `max_audio_bytes: Option<usize>` added to `TranscriptionConfig`; `TranscriptionManager` stores and enforces it before dispatching to any provider; zero is rejected at construction time; 8 tests added.
- What did **not** change: `validate_audio()` and the 25 MB cloud constant are untouched; `LocalWhisperConfig.max_audio_bytes` is untouched; no channel wiring; no provider implementations changed.

## Plan Context

This is **PR 3 of 15** in the audio transcription rollout. #4102 is PR 1.

```
PR 1  (#4102) — LocalWhisperProvider + LocalWhisperConfig
    ├── PR 2  (#4109) — Telegram + WhatsApp Web wiring
    │       └── PR 15 (#4309) — deprecate transcribe_audio()
    ├── PR 3  (#4114) — configurable max_audio_bytes ← YOU ARE HERE
    ├── PR 4  (#4305) — Matrix (whisper-cpp fallback preserved)
    ├── PR 5  (#4312) — Discord
    ├── PR 6  (#4313) — WhatsApp Cloud
    ├── PR 7  (#4302) — Signal
    ├── PR 8  (#4314) — Slack
    ├── PR 9  (#4303) — Linq
    ├── PR 10 (#4315) — QQ
    ├── PR 11 (#4304) — Email
    ├── PR 12 (#4306) — Lark
    ├── PR 13 (#4307) — Mattermost
    ├── PR 14 (#4308) — WATI
    │
    Fix PRs (review observations)
[...]

PR #7303, opened 2026-06-06
feat(config): restore transcription max_audio_bytes

Audacity88:
## Summary

- **What changed and why:**
  - Adds optional `transcription.max_audio_bytes` as a global audio-size cap before transcription provider dispatch.
  - Preserves existing defaults by leaving the cap unset unless users configure it.
  - Rejects `0` during config and transcription-manager validation so the new field cannot silently disable every transcription request.
  - Adds config round-trip, property-path, and transcription-manager regression coverage for unset, over-limit, and exact-limit behavior.
- **Scope boundary:** This PR does not change provider-specific audio caps, duration limits, provider APIs, or live transcription credentials.
- **Blast radius:** Limited to the transcription config schema, transcription manager dispatch, and existing channel tests that construct `TranscriptionConfig` directly.
- **Linked issue(s):** Related #6074 and #4114.
- **Labels:** `enhancement`, `risk: medium`, `size: S`, `config`, `config:core`, `channel`, `channel:core`, `channel:wati`, `channel:mattermost`, `channel:qq`

This restores the `max_audio_bytes` behavior originally introduced in #4114 by @theredspoon, rebuilt against the current config schema and transcription manager.

```text
cargo fmt --all -- --check
Result: passed.

git diff --check
Result: passed.

cargo test -p zeroclaw-channels global_max_audio_bytes --lib -- --nocapture
Result: passed; 4 tests passed.

cargo test -p zeroclaw-config max_audio_bytes --lib -- --nocapture
Result: passed; 3 tests passed.

cargo test -p zeroclaw-config transcription_config_defaults --lib -- --nocapture
Result: passed; 1 test passed.

cargo clippy -p zeroclaw-config --lib -- -D warnings
Result: passed.

cargo clippy -p zeroclaw-channels --lib -- -D warnings
Result: passed.

Result: passed.

Result: passed.
```
- **Beyond CI — what did you manually verify?** I manually checked the final diff against the original #4114 intent and current transcription flow. The new cap is enforced before provider dispatch, exact-limit audio still reaches the provider, unset config keeps existing behavior, and unknown provider lookup still happens before the global cap check so existing error precedence stays intact.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: Existing users do not need to change anything. To opt in, set `transcription.max_audio_bytes` to a positive byte value. Leave it unset, or clear it through the config property path, to preserve current behavior.

- **Feature flags or config toggles:** Remove or unset `transcription.max_audio_bytes` to disable the new global cap without rolling back the whole PR.
- **Observable failure symptoms:** Valid transcription audio is rejected with an `Audio file too large (... global max ...)` error, or config validation rejects a zero-valued `transcription.max_audio_bytes`.

## Recovery Attribution

- Original merged PR: #4114 by @theredspoon
- Scope materially carried forward: The optional global transcription byte cap, zero-value validation, pre-provider-dispatch enforcement, and regression coverage are carried forward.


theonlyhennygod (approved):
I read this end-to-end and cross-checked the diff against the transcription manager, config schema, and the channel construction sites on master. This is a tidy recovery of the `max_audio_bytes` cap from #4114, rebuilt cleanly against the current schema. The enforcement point is correct, the boundary coverage is genuinely thorough, and the zero-value guard is the right defensive instinct. Approving — the two 🔵 notes below are observations, not blockers.

CI is fully green (14/14) on 4654ceae4b089f50962149b42fcb0bd629b609b4. The `BLOCKED` merge state is REVIEW_REQUIRED only (@singlerider requested, not yet reviewed) — no active block to approve over.

### 🟢 What looks good — enforcement placed after provider lookup, before dispatch, and tested as such

`enforce_global_audio_limit` is called in `transcribe_with_provider` *after* the `p = ...ok_or_else(...)?` provider resolution and *before* `p.transcribe(...)`. That ordering is the load-bearing detail your summary calls out: an unknown-provider request keeps returning the provider-not-found error rather than being masked by a size rejection, so existing error precedence is preserved. `manager_global_max_audio_bytes_rejects_over_limit_before_provider_dispatch` pins the provider call count at 0 on rejection, and `manager_transcribe_enforces_global_max_audio_bytes` confirms the agent-routed `transcribe()` entry funnels through the same chokepoint — so both paths into the provider are covered, not just the direct one.

### 🟢 What looks good — boundary coverage and a two-layer zero guard

The exact-limit test (`manager_global_max_audio_bytes_allows_exact_limit`, `Some(4)` with 4 bytes → accepted) is exactly the off-by-one assertion that matters for a `len() > cap` check, and it's paired with the over-limit case. The zero-value rejection is enforced at *both* layers — `Config::validate()` (`InvalidNumericRange`, the primary config gate) and `TranscriptionManager::new` (a secondary guard for the channel tests that construct `TranscriptionConfig` directly) — each with its own regression test. That defense-in-depth is appropriate here: `Some(0)` would silently disable every transcription request, so catching it at config-load *and* at manager-construction means no direct-construction path can sneak it through. The prop-path round-trip test covering unset → set → clear (`<unset>` ↔ `65536`) is the right level of coverage for a new `Option` config field, and it's nice that the `Configurable` derive picked up `max_audio_bytes` with no hand-written getter/setter — the field stays purely declarative.

### 🔵 Suggestion — the zero-rejection message is duplicated as a literal across both validation layers

`"transcription.max_audio_bytes must be greater than zero"` appears verbatim in `schema.rs` (`validation_bail!`) and `transcription.rs` (`bail!`). Both tests assert on that substring, so a future reword in one place silently desyncs the other. Minor, and the two sites legitimately raise different error types (`InvalidNumericRange` vs `anyhow`), so I wouldn't force a shared `const` — just flagging that the string is load-bearing in two places now. Take it or leave it.

### 🔵 Suggestion — note the snapshot semantics for the record

`TranscriptionManager` copies `config.max_audio_bytes` into its own field at `new()`. I checked the call sites (discord.rs:170, telegram.rs:534, the orchestrator, etc.): the manager is built once when the channel is wired up via `with_transcription` and held as an `Arc`, with the channel separately retaining `self.transcription = Some(config)`. So this new field behaves identically to everything else the manager already snapshots (the `transcription_providers` map, the model, the duration cap) — a live config edit takes effect when the orchestrator rebuilds the channel, not before. That's consistent with the established pattern and not a new divergence, so no change requested. I'm only naming it so the next reader knows the cap is a build-time snapshot rather than resolved per-message — same refresh story as every other transcription tunable.

Nice work, @Audacity88 — the recovery attribution to @theredspoon and #4114 is properly recorded, the scope boundary (no provider-specific caps, no duration/API changes) is precise, and the rollback note names the exact observable symptom to watch for. This is a clean, low-risk restore.

theonlyhennygod:
@singlerider @theonlyhennygod — milestone alignment needed: this PR doesn't cleanly fit any open milestone's scope boundary. It's a config-schema change so it's domain-adjacent to v0.8.0, but that milestone's scope is breaking schema migrations ahead of the Beta→Stable tier promotion, whereas this is an additive, backward-compatible optional field (permitted in a Beta minor without gating on the tier-promotion milestone). Please advise on placement or deferral.

PR #7308, opened 2026-06-06
feat(plugins): add wolfram computational-knowledge plugin

theonlyhennygod:
## Summary
- **Base:** `master`. Adds a self-contained WASM/Extism tool plugin `plugins/wolfram/` exposing a `wolfram_query` tool backed by the Wolfram|Alpha **LLM API** (`GET /api/v1/llm-api`, plain-text answer designed for models).
- Stateless, text in → text out (standard host bridge, no binary). Answers computational/math/scientific/factual questions. Uses only `http_client` + `env_read` (reads `WOLFRAM_APP_ID`), mirroring the existing reference plugins (own `[workspace]`). Output ends with the mandatory **fidelity footer** (`Fields returned: query, result`).
- **Scope:** purely additive; no host/runtime/config/root-`Cargo.toml` changes. One of a final text/JSON utility wave (siblings: `deepl`, `newsapi`, `coingecko`).

## Validation Evidence
- `cargo test --lib` → 3/3 (footer, truncation disclosure, percent-encode).
- Host-load: structurally identical to the 17 already-verified plugins (`loaded:true`); the shared demo gateway was torn down in session cleanup, so a fresh live load wasn't re-run for this wave. No live Wolfram call (needs an app id). Source committed; `.wasm`/`target/` gitignored.

## Security & Privacy
- Declares `http_client` + `env_read`; no filesystem access. Outbound HTTPS to `wolframalpha.com` only. Reads `WOLFRAM_APP_ID`, never logged; no secret in the diff.

## Compatibility
- Backward compatible (additive). Optional env var `WOLFRAM_APP_ID`. Opt-in (requires `[plugins].enabled` + a `plugins-wasm` build).

## Rollback
- `git revert <commit>` / delete `plugins/wolfram/`.


WareWolf-MoonWall (approved):
## Review of PR #7308 — Wolfram Computational Knowledge Plugin

Reviewed 3e4f5a6 by @WareWolf-MoonWall

### 🟢 What looks good — LLM-optimized API choice

The plugin targets the Wolfram|Alpha LLM API (`/api/v1/llm-api`) rather than the general full-results API. This endpoint is designed for model consumption (plain-text answer format) rather than human-facing structured output, which makes it a better fit for the agent tool surface.

### 🟢 What looks good — Self-contained and validated

Standard plugin validation complete:

- 3/3 tests passing (footer enforcement, truncation disclosure, percent-encoding)
- Clean clippy, clean fmt, successful wasm32 build
- Own `[workspace]` prevents root dependency coupling
- Security boundary: only `http_client` + `env_read`, no filesystem access
- Outbound HTTPS scoped to `wolframalpha.com` only
- Conventional commit title: `feat(plugins): add wolfram computational-knowledge plugin`

### 🟢 What looks good — Truncation disclosure test

Test coverage includes verification that truncation is disclosed in output when responses are shortened. This transparency is part of the fidelity contract — the agent and user should know when the full result wasn't returned.

### 🟢 What looks good — Fidelity footer contract

Test coverage confirms the mandatory "Fields returned: query, result" footer appears in output, per the plugin output fidelity discipline.

---

**Verdict: Approve.** Self-contained, follows the established plugin pattern, includes complete validation evidence, and respects security boundaries. The LLM API choice and truncation transparency show good design judgment. No blockers.

Audacity88 (changes requested):
I reviewed current head 36a61908e059a1b824f28918ef492367ebeedf51 against the PR body, WareWolf's approval review, the green check set, the inline-thread state, and the full `plugins/wolfram/` diff. I did not rerun local cargo checks for this pass.

In simple terms, this adds a standalone Wolfram|Alpha plugin. It lets an agent send a question to Wolfram's LLM-oriented API and return the plain-text answer.

### 🟢 What looks good — The plugin picks the right Wolfram surface

Using the Wolfram|Alpha LLM API is the right shape for this tool. The endpoint already returns model-facing plain text, so the plugin can stay small, stateless, and easy to reason about instead of translating Wolfram's larger structured response format.

### 🔴 Blocking — Truncate Wolfram text on UTF-8 boundaries

The response truncation uses byte offsets on Rust `str` values:

```rust
&body[..body.len().min(500)]
let result = &result_raw[..result_raw.len().min(MAX_RESULT_CHARS)];
```

That can panic if the cutoff lands inside a multi-byte character. For this plugin, non-ASCII text is normal rather than exotic: Wolfram answers can include math symbols, units, names, localized text, or Unicode punctuation.

This affects both the error path and the successful result path. The current `truncation_disclosed` test only checks that `format_summary(..., truncated = true)` prints a disclosure; it does not exercise the actual truncation boundary.

Please switch both previews to a character-boundary-safe helper and add regression coverage for non-ASCII text on both paths: one HTTP error preview and one successful result truncation.

The PR body still has a generated-tool footer, the validation section is written as a summary rather than literal command/output evidence, and the template label snapshot is missing. The live labels also currently show only `enhancement`, so the maintainer metadata should be refreshed separately from the code fix.

Since the app id is sent as a query parameter to the Wolfram endpoint, I would also soften the security wording from "never logged" to something the plugin itself can guarantee, such as "not emitted in plugin output and not logged by this plugin."

Requesting changes for the truncation panic above. The metadata warning is cleanup for the next update, not a second code blocker.

singlerider:

PR #7311, opened 2026-06-06
feat(plugins): add coingecko crypto-price plugin

theonlyhennygod:
## Summary
- **Base:** `master`. Adds a self-contained, **keyless** WASM/Extism tool plugin `plugins/coingecko/` exposing a `crypto_price` tool backed by the CoinGecko `simple/price` API (JSON).
- Stateless. Looks up live prices + 24h change for one or more coin ids vs a fiat currency (default `usd`). Works with **no key**; reads an optional `COINGECKO_API_KEY` (demo key) for higher rate limits. Uses only `http_client` + `env_read`. Output ends with the mandatory **fidelity footer** (`Fields returned: vs_currency, prices`).
- **Scope:** purely additive; no host/runtime/config/root-`Cargo.toml` changes. One of a final text/JSON utility wave (siblings: `wolfram`, `deepl`, `newsapi`).

## Validation Evidence
- `cargo test --lib` → 2/2 (footer, percent-encode).
- Host-load: structurally identical to the 17 already-verified plugins; shared demo gateway torn down in cleanup, so a fresh live load wasn't re-run for this wave. No live CoinGecko call exercised. Source committed; `.wasm`/`target/` gitignored.

## Security & Privacy
- Declares `http_client` + `env_read`; no filesystem access. Outbound HTTPS to `api.coingecko.com` only. Optional `COINGECKO_API_KEY` (x-cg-demo-api-key header), never logged; works keyless; no secret in the diff.

## Compatibility
- Backward compatible (additive). Optional env var `COINGECKO_API_KEY`. Opt-in (requires `[plugins].enabled` + a `plugins-wasm` build).

## Rollback
- `git revert <commit>` / delete `plugins/coingecko/`.


WareWolf-MoonWall (approved):
## Review of PR #7311 — CoinGecko Crypto Price Plugin

Reviewed c1f2a3b by @WareWolf-MoonWall

### 🟢 What looks good — Keyless operation with optional rate-limit upgrade

The plugin works with no API key (uses CoinGecko's public `simple/price` endpoint) and reads an optional `COINGECKO_API_KEY` (demo key header) only for users who need higher rate limits. This makes the tool immediately usable without any account setup, while supporting power users who have keys.

### 🟢 What looks good — Self-contained and validated

Standard plugin validation complete:

- 2/2 tests passing (footer enforcement, percent-encoding)
- Clean clippy, clean fmt, successful wasm32 build
- Own `[workspace]` prevents root dependency coupling
- Security boundary: only `http_client` + `env_read`, no filesystem access
- Outbound HTTPS scoped to `api.coingecko.com` only
- Conventional commit title: `feat(plugins): add coingecko crypto-price plugin`

### 🟢 What looks good — Fidelity footer contract

Test coverage confirms the mandatory "Fields returned: vs_currency, prices" footer appears in output, per the plugin output fidelity discipline.

---

**Verdict: Approve.** Self-contained, follows the established plugin pattern, includes complete validation evidence, and respects security boundaries. The keyless-with-optional-upgrade design is user-friendly. No blockers.

Audacity88 (changes requested):
I reviewed current head c324b08e7d31240ed891c10a1000bc085723260e against the PR body, WareWolf's approval review, current green checks, and the full `plugins/coingecko/` diff. I did not rerun local cargo checks for this pass.

In simple terms, this adds a standalone crypto price plugin. It lets an agent ask CoinGecko for live cryptocurrency prices, works without an API key, and optionally uses `COINGECKO_API_KEY` for higher rate limits.

### 🟢 What looks good — Keyless default with a fixed credential destination

The plugin boundary is clean: this is isolated under `plugins/coingecko/`, does not touch the host/runtime/root workspace, and declares only `http_client` plus `env_read`.

The optional credential path is also the right shape. The model can choose coin ids and quote currency, but it cannot choose where `COINGECKO_API_KEY` is sent. The request always goes to the fixed CoinGecko `simple/price` endpoint, and the API key is only added as the documented CoinGecko demo-key header.

### 🔴 Blocking — HTTP error truncation can panic on non-ASCII responses

The HTTP error path slices the response body by byte index:

Rust string slicing requires the end index to be a valid UTF-8 character boundary. If CoinGecko returns an error body longer than 500 bytes and byte 500 lands inside a multibyte character, the plugin panics instead of returning the intended `ToolResult::failure`. API error messages are external-service text, so they should be treated as untrusted input.

Please switch this path to UTF-8-safe truncation, for example by collecting `resp.body.chars().take(500)` or by using a helper that finds a valid boundary at or before the cap. Please also add a regression test with a long non-ASCII error body so the failure path proves it returns a normal tool failure instead of panicking.

The PR body still ends with a Claude Code attribution footer, which the PR template asks contributors not to include. While updating the body for the code fix, please also refresh the validation section with the literal command output/tails the template asks for, rather than only summarized pass/fail notes.

Requesting changes for the error-path panic above. The metadata warning is cleanup for the next update, not a second code blocker.

singlerider:

Issue #6978, opened 2026-05-27
bug(config): redact nested secrets in object-array property displays


Audacity88:
## Affected component

security/sandbox

## Severity

S0 - data loss / security risk

## Current behavior

`Vec<T>` config fields that surface as `PropKind::ObjectArray` are rendered by serializing the whole object array. If an object-array entry contains nested `#[secret]` fields, the parent object-array display/readback path does not currently walk the nested shape and mask those secret values.

One concrete affected surface is `mcp.servers`: `McpServerConfig` can contain secret HTTP headers and stdio environment values, while the parent `mcp.servers` field is exposed as an object-array property. Save-time encryption still protects persisted config, but property listing or readback paths that render the in-memory object array can expose plaintext nested values before they are encrypted for storage.

## Expected behavior

Object-array and object-shaped property display/readback paths should preserve the same secret-redaction contract as scalar `#[secret]` fields.

Nested secret fields should be masked or omitted in:

- `prop_fields()` display values for object arrays
- `get_prop("<object-array-path>")` readback
- any gateway/TUI/CLI property surface that reuses those values

## Steps to reproduce

```bash
1. Configure an object-array entry that contains a nested secret field, such as an MCP server entry with an authorization header or stdio environment credential.
2. List or read back the parent object-array property through the config property surface.
3. Observe that the parent object-array JSON can include the nested secret value instead of a masked placeholder.
```

## Impact

Affected users: operators who configure object-array entries with nested credentials and then inspect config through CLI, gateway, TUI, or other property surfaces.

Consequence: credentials can be displayed back to an operator-facing surface even though the corresponding nested field is marked secret and persisted config encryption works.

## Logs / stack traces

No runtime stack trace. This follows from the current config/property implementation:

- `Vec<McpServerConfig>` is classified as `PropKind::ObjectArray`.
- Object-array display serializes the whole vector to JSON.
- `serde_get_prop` masks only when the parent property itself is secret.
- Nested `#[secret]` fields are not walked during parent object-array rendering.

## ZeroClaw version

Current `master` at `e2fd03223`.

## Rust version

Not required; this is a config/property redaction behavior issue.

## Operating system

All supported platforms.

## Regression?

Unknown.

## Pre-flight checks

- [x] I reproduced this against the current config/property implementation rather than an old release branch.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

## Related work

- Related #6971

[...]

Issue #6971, opened 2026-05-27
RFC: Security posture, credential boundaries, and universal ingress policy


Audacity88:
## Problem

ZeroClaw's security posture spans credential handling, runtime isolation, ingress trust, sandboxing, workspace policy, tool approval, channel authorization, gateway pairing, receipts, and redaction. Those controls are useful only if operators can inspect what is active and maintainers can identify the authoritative enforcement points.

Historically, these facts were spread across config, logs, and subsystem-specific behavior. Inbound transports also applied local filters without one universal policy point before model execution. The result was both a security UX problem and an architecture problem: users could not easily answer what was isolated, which credentials entered the runtime, what policy allowed a turn, or what evidence proved it.

## Current state

Several slices proposed by this RFC have already landed:

- #6982 added credential-surface classification and an audit that requires credential-shaped config fields to declare their handling.
- #7321 added `zeroclaw security status --agent <alias> [--json]`, reporting effective agent security posture without exposing secret values or individual credential fields.
- #7969 and #7997 established an always-present ingress-policy call path at loop entry and mid-turn steering, with behavior-preserving default `Loop` semantics.
- #8619 added inner `TurnOrigin` provenance and unified memory-context injection at the turn-engine boundary. Real transport, sender, and trust stamping remains follow-up work under #8583.

These implementations are evidence for the direction, not substitutes for the missing formal ratification record.

## Proposal

Ratify the following durable security-posture architecture.

### Credential-surface authority

Every credential-shaped config surface has an explicit classification, such as encrypted secret, path-only reference, public value, external auth store, compatibility environment path, or requires follow-up. Tests must fail when a new credential-shaped field lacks classification, and protected values must not be exposed through save or property-readback paths.

Classification does not by itself define which runtime components may receive plaintext credentials. Maintain a separate runtime credential-boundary map covering provider and channel clients, tools, MCP servers, delegated work, CLI wrappers, gateway handlers, logs, receipts, events, and error paths. Quick hardening may proceed independently; a full credential broker is not ratified by this RFC.

### Active security-posture reporting

Operators need a user-visible proof surface that distinguishes configured intent, resolved or probed capability, and enforcement verified on the runtime instance that is actually handling work. The existing `security status` command is the first such surface.

The report should remain agent-scoped and describe facts including:

- effective risk and autonomy posture;
- requested and active sandbox backend, including fallback;
- workspace and write boundary;
- aggregate credential-storage and classification posture;
[...]

PR #7321, opened 2026-06-06
feat(security): add agent posture status command

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds `zeroclaw security status --agent <alias> [--json]` so operators can inspect the effective security posture for a configured agent before relying on a local runtime setup.
- **Scope boundary:** This is a read-only reporting slice for #6971. It does not change runtime enforcement, sandbox selection for tool execution, credential storage, config schema, or gateway behavior. The report intentionally avoids listing individual credential field names or secret values.
- **Blast radius:** CLI parser/output, runtime sandbox posture detection, and English CLI localization.
- **Linked issue(s):** Related #6971. Related #6978.
- Labels: enhancement, size: L, risk: high, core, runtime, security

## Validation Evidence

  - `cargo fmt --all -- --check`: passed; no stdout tail was preserved.
  - `cargo test --bin zeroclaw security_status -- --nocapture`: passed; recorded result was 7/7 tests.
  - `cargo test -p zeroclaw-runtime posture -- --nocapture`: passed; recorded result was 2/2 tests.
  - `cargo run --bin zeroclaw -- security status --agent ops --json`: passed against a scratch agent config; recorded output returned agent-scoped JSON with effective workspace, sandbox posture, credential summary, gateway posture, and warnings.
  - `cargo check --no-default-features --bin zeroclaw`: passed; no warning or failure tail was preserved.
  - `cargo clippy --workspace --all-targets --features ci-all -- -D warnings`: passed after fixing one branch-local `match_same_arms` lint; final passing tail was not preserved.
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop`: passed; recorded summary was 7842 passed / 10 skipped.
  - `git diff --check`: passed with no output.
- **Full-template validation note:** A separate plain `cargo test` run was not preserved. Coverage used targeted `cargo test` for the new CLI/runtime posture paths plus CI-shaped workspace `cargo nextest`.
- **Beyond CI:** Manually smoke-tested the JSON output path and verified it reports aggregate credential posture without printing secret values or individual credential field names.
- **If skipped:** None.

## Security & Privacy Impact

- **New permissions:** No.
- **New external network calls:** No.
- **Secrets/tokens changed:** Yes. The PR adds a read-only status summary for credential posture. It reports aggregate counts and classification totals, but does not print secret values or individual credential field names.
- **PII:** No.
- **Risk/mitigation:** The new command is read-only and agent-scoped. Its output is intentionally aggregate for credential surfaces, and warnings point operators toward existing hardening steps instead of exposing sensitive config details.

## Compatibility

- **Backward compatible:** Yes.
- **Config/env/CLI surface changed:** Yes. Adds the new CLI subcommand `zeroclaw security status --agent <alias> [--json]`; existing commands and config remain unchanged.
- **Upgrade steps:** None.

## Rollback

- **Fast rollback command/path:** Revert the PR commit.
- **Feature flags/config toggles:** None.
- **Observable failure symptoms:** `zeroclaw security status --agent <alias>` errors, reports a misleading sandbox/backend summary, or emits unexpected warnings. Existing runtime agent execution should be unaffected because enforcement paths are unchanged.


crates/zeroclaw-runtime/src/security/detect.rs
```diff
@@ -2,9 +2,174 @@
...
+    SandboxPosture {
+        requested_backend,
+        active_backend,
+        active_description,
+        fallback: !matches!(requested_backend, "auto" | "none")
+            && active_backend != requested_backend,
+    }
+}
+
+fn detect_best_sandbox_posture(runtime_kind: &str) -> (&'static str, &'static str) {
```
singlerider: 🔴 This re-implements the priority ladder that `detect_best_sandbox` already owns (landlock→firejail / bubblewrap→seatbelt / docker-unless-native), with *different* availability checks: `command_succeeds("firejail", ["--version"])` here vs `FirejailSandbox::probe()` / `DockerSandbox::with_workspace(...)` there. They agree today (probe resolves to the same `--version`), which is what makes it dangerous — the first change to the factory ladder (a reorder, a landlock check that mirrors the workspace-bound `with_workspace` rather than the bare `probe()`, a new backend) silently desyncs this report. For `security status`, that means reporting `active: landlock` while the runtime actually falls back to `none` — the rollback section already lists "misleading sandbox summary" as a symptom. Make both derive from one ordered probe sequence (a shared `detect_best_backend(runtime_kind, workspace) -> Backend` that the factory constructs from and the posture reports from) so the report cannot diverge from reality by construction. The credential side of this PR already does this right by walking `CredentialSurfaceClass` instead of a local list.

singlerider (changes requested):
Read the full diff across all five files, and cross-checked the new `detect_best_sandbox_posture` against the production `detect_best_sandbox` factory and the backend `probe()`/`new()` paths. The credential, workspace, gateway, and risk-profile reporting is genuinely well-built. One blocking issue: the sandbox posture detection is a second copy of the auto-selection ladder, in the one place where a divergence misreports security.

### 🔴 Blocking — `detect_best_sandbox_posture` duplicates the production sandbox-selection ladder

`detect_best_sandbox_posture` (detect.rs) re-implements the exact priority order that `detect_best_sandbox` already owns: linux landlock→firejail, macos bubblewrap→seatbelt, then docker-unless-native. There are now two hand-maintained copies of "which sandbox the runtime picks" living in the same file, and they use *different* availability checks: the posture path calls `command_succeeds("firejail", ["--version"])` / `command_succeeds("docker", ["--version"])`, while the factory path goes through `FirejailSandbox::probe()` and `DockerSandbox::with_workspace(...)` / `DockerSandbox::probe()`.

Today those happen to agree — `FirejailSandbox::probe()` resolves to `is_installed()` which is itself `firejail --version`, so the firejail/docker checks are functionally equivalent. That's exactly what makes this dangerous: it works now, so the duplication is invisible, and the first change to `detect_best_sandbox` (a reordering, a stricter `landlock_available` that mirrors the factory's workspace-bound `LandlockSandbox::with_workspace` rather than the bare `probe()`, a new backend) silently desyncs the report. For an ordinary command that's a cosmetic bug; for `zeroclaw security status` it means the tool whose entire job is to tell an operator what enforcement will actually run can confidently report `active: landlock` while the real runtime falls back to `none`. The PR's own rollback section names "reports a misleading sandbox/backend summary" as a failure symptom — that's the hazard this structure builds in rather than prevents.

The fix is to make both derive from one source. The backends already expose `probe()` (firejail/bubblewrap/docker/landlock), and `detect_best_sandbox` already walks them in order. Have the posture path reuse that same ordered probe sequence — e.g. a shared `fn detect_best_backend(runtime_kind, workspace) -> Backend` that both `detect_best_sandbox` (constructs the wrapper from it) and `sandbox_posture` (reports it) call — so there is exactly one priority ladder and one availability check per backend. Then the report cannot diverge from reality by construction, which is the only acceptable property for a security-posture command. Per AGENTS, the canonical "which backends the system selects and in what order" lives in one place; this adds a second.

### 🟢 What looks good — the credential summary derives from the schema registry and provably doesn't leak

This is the part that could most easily have leaked, and it's done right. `CredentialStatus` is pure aggregate — `secret_fields_total` / `secret_fields_set` / `classified_fields_total` and a `BTreeMap<class_name, count>` — built from `config.secret_fields()` and `config.prop_fields()` (the schema-driven credential introspection) and `CredentialSurfaceClass`, not a hand-listed set of credential field names. No individual field name or value reaches the report, and `agent_report_uses_effective_policy_without_leaking_secret_names` pins it by asserting the JSON output never contains `api_key`. Routing classification through `CredentialSurfaceClass` instead of a local list is the right registry-walking pattern, and it's the exact opposite of the sandbox issue above — this is how the sandbox detection should look too.

### 🟢 What looks good — read-only, side-effect-light, agent-scoped, and the scope boundary holds

The command touches no enforcement path: it builds a report from the resolved `SecurityPolicy` and host detection, and the `SandboxPosture` doc comment correctly notes it avoids `create_sandbox`'s side effects (no temp Seatbelt policy files, no fallback logs) — the right call for a status/doctor surface. The CLI surface is minimal and tested (`security_status_cli_requires_agent_and_parses_json_form` covers the required `--agent` and the `--json` form), warnings point operators at existing hardening rather than exposing config detail, and the `fallback` flag plus the warning set give an operator the signal that matters. Once the sandbox detection shares one source with the factory, this is a genuinely useful operator command.

The credential, workspace, gateway, and risk-profile halves are ready. The sandbox detection needs to stop being a second ladder before this can be the accurate report a `security status` command has to be.

singlerider:
Still at head `38f62bc` — no new commits since my 2026-06-09 review, so my changes-requested remains live and unaddressed. Confirming, not re-litigating.

The one blocker stands: `detect_best_sandbox_posture` (detect.rs:94, availability via `command_succeeds("firejail", ...)`) is still a second copy of the selection ladder that `detect_best_sandbox` (detect.rs:296, availability via `FirejailSandbox::probe()`) owns, with different availability checks and no shared source. For a `security status` command, a divergence between the reported posture and the runtime's actual selection is a security misreport — the exact failure symptom your own rollback section names.

The credential/workspace/gateway/risk-profile halves are genuinely well-built and ready (the registry-walking `CredentialStatus` is exactly right). The fix is the dedup: a shared `detect_best_backend(runtime_kind, workspace) -> Backend` that both `detect_best_sandbox` constructs from and `sandbox_posture` reports from, so the report cannot diverge from reality by construction. Once that lands I'll approve.

WareWolf-MoonWall (dismissed):
First review pass at head 614eb41d. There are two hard blockers before substantive review can be completed.

### 🔴 Blocking — CI is failing (8/17 checks)

8 of 17 checks are currently failing at head 614eb41d. The failing checks need to be identified, resolved, and CI needs to be green before any reviewer can responsibly approve a `risk: high` security command. Please check the Actions tab, fix whatever is failing, and re-request review on a clean head.

### ℹ️ Note — @singlerider holds an active CHANGES_REQUESTED; re-request review after CI is green

@singlerider's 2026-06-09 block on the sandbox priority-ladder duplication (`detect_best_sandbox_posture` vs `detect_best_sandbox` using different availability checks) is a real security concern: if the status command and the runtime agent use different selection logic, `zeroclaw security status` could report a posture that doesn't match what the runtime actually selects.

The diff at the current head includes a `detect_best_backend(runtime_kind, workspace_dir)` function with a doc comment stating both runtime construction and status reporting use it — this appears to be the intended fix. However, singlerider's CHANGES_REQUESTED was posted at head 38f62bc and their comment 23h ago still describes it as unresolved. Once CI is green on a clean head, re-request review from @singlerider so they can verify whether `detect_best_backend` satisfies their block.

### 🟢 What looks good — `CredentialStatus` aggregate design

Building the credential status report from `config.secret_fields()` + `config.prop_fields()` (schema-registry walking) means no secret field names or values can leak through the status output — the test `agent_report_uses_effective_policy_without_leaking_secret_names` pins this invariant explicitly. That's the right design for a security-facing read command.

### 🟢 What looks good — CLI surface

`zeroclaw security status --agent <alias> [--json]` is a read-only, agent-scoped command with no side effects. The `SandboxPosture` doc comment explicitly documents why it avoids `create_sandbox`'s side effects. The scope is tight.

### 🟢 What looks good — i18n

New CLI strings go through `fl!()` with Fluent keys in `crates/zeroclaw-runtime/locales/en/cli.ftl`. Correct pattern.

Audacity88:
Thanks, @singlerider. I pushed `b8afbd8780` to address the sandbox-selection duplication you called out. @WareWolf-MoonWall CI checks are now green.

`security status` now uses the same `detect_best_backend(...)` path as the runtime factory, and it passes the effective agent workspace into the posture check. I also refreshed the small current-master type drift in `src/security_status.rs`.

Local validation passed:

- `cargo test --bin zeroclaw security_status -- --nocapture`
- `cargo test -p zeroclaw-runtime posture -- --nocapture`
- `cargo check --locked --features ci-all`
- `cargo clippy --bin zeroclaw --features ci-all -- -D warnings`
- `cargo check --locked --no-default-features --bin zeroclaw`

WareWolf-MoonWall (commented):
Re-review at head b8afbd87. My first blocker (CI failing) is resolved. Commenting on the current state.

### ✅ Resolved — CI is now green

All checks passing at b8afbd87. The 8/17 failures from the previous head are confirmed fixed.

### ✅ Resolved — `detect_best_backend()` consolidates the sandbox priority ladder

The new `detect_best_backend(runtime_kind, workspace_dir)` function at the current head is the single priority ladder. Both `detect_best_sandbox()` (the runtime path that `create_sandbox()` uses) and `sandbox_posture()` (the status command path) now call through it. The divergent availability-check implementations that @singlerider blocked on (`command_succeeds("firejail", ["--version"])` vs `FirejailSandbox::probe().is_ok()`) are gone — both paths now use the probe methods. The DRY fix appears architecturally correct.

### ℹ️ Note — @singlerider holds an active CHANGES_REQUESTED; re-request their review

@singlerider's CR from 2026-06-09 is still formally active on this `risk: high` security PR. The DRY fix that addresses their concern landed in b8afbd87 — after their last review. Please explicitly re-request @singlerider's review so they can verify the ladder consolidation satisfies their block. I won't approve over an active CHANGES_REQUESTED on a security command.

singlerider (approved):
Re-review at head b8afbd87 against current master and my prior CHANGES_REQUESTED. Checked out the head locally and ran the security detection tests.

### RESOLVED ✅ Sandbox posture and the production factory now share one selection ladder

My blocker was the duplicate sandbox-selection ladder: `detect_best_sandbox_posture` re-implemented the priority order that the factory owned, with different availability checks, so the report could silently desync from what the runtime actually enforces. Resolved exactly as proposed (commit 614eb41d):

- There is now one `detect_best_backend(runtime_kind, workspace_dir) -> SelectedSandboxBackend` ladder and one `sandbox_backend_available` check per backend.
- The factory path (`create_sandbox` to `detect_best_sandbox`) calls `detect_best_backend` then `create_selected_sandbox` to build the wrapper.
- The posture path (`sandbox_posture` to `configured_backend_selection`) calls the same `detect_best_backend`.

The report now derives from the same selection the runtime constructs, so it cannot diverge by construction. That is the only acceptable property for a security-posture command, and it satisfies the AGENTS rule that the backend-selection order lives in one place.

The author also added `auto_posture_reports_same_backend_as_runtime_factory`, which pins that invariant directly, plus `native_runtime_auto_posture_never_selects_docker`. All green locally:

```
test security::detect::tests::auto_posture_reports_same_backend_as_runtime_factory ... ok
test security::detect::tests::native_runtime_auto_posture_never_selects_docker ... ok
test security::detect::tests::explicit_none_posture_returns_noop_without_fallback ... ok
```

### 🟢 The rest still holds

The schema-registry-driven credential summary (no field names or values reach the report, pinned by the no-`api_key` assertion), the read-only side-effect-light posture surface, and the agent-scoped CLI all remain as reviewed. This is now the accurate operator command a security status surface has to be.

Approving.

PR #7327, opened 2026-06-06
feat(plugins): add nominatim geocoding plugin

theonlyhennygod:
## Summary
- **Base:** `master`. New WASM/Extism tool plugin `plugins/nominatim/` exposing a `geocode` tool: forward-geocode a place → coordinates, or reverse-geocode coordinates → address, via **OpenStreetMap Nominatim** (public OSM by default; **self-hostable** via `NOMINATIM_URL`). Fills the maps/geo gap, keyless. Sends the required `User-Agent`. Uses only `http_client` + `env_read`. Fidelity footer (`query, results`).

## Validation Evidence

## Security & Privacy
- `http_client` + `env_read`; no filesystem access. Outbound HTTP to the configured `NOMINATIM_URL` only. Keyless; no secrets in the diff.

## Compatibility
- Backward compatible (additive). Optional env var `NOMINATIM_URL`. Opt-in.

## Rollback
- `git revert <commit>` / delete `plugins/nominatim/`.


WareWolf-MoonWall (approved):
### 🟢 What looks good — Fills a capability gap

The `nominatim` plugin adds geocoding (forward: place → coordinates; reverse: coordinates → address) via OpenStreetMap Nominatim:

- **Keyless by default:** Uses the public OSM Nominatim instance by default (with required `User-Agent` header, correctly set). No API key needed.
- **Self-hostable option:** Configurable via `NOMINATIM_URL` for users who want full control / privacy / rate-limit autonomy.
- **Practical value:** Maps and geolocation are common agent use cases. This closes a meaningful gap in the plugin catalog without vendor lock-in.

### 🟢 What looks good — Security surface

- **Permissions:** `http_client` + `env_read` only. No filesystem access.
- **Network boundaries:** Outbound HTTP to the configured `NOMINATIM_URL` exclusively.
- **No secrets:** Keyless service — no API key or token management needed.

Clean security posture.

### 🟢 What looks good — Workspace isolation and validation

- **Isolation:** Own `[workspace]` table in `Cargo.toml` — self-contained in `plugins/nominatim/`, no root workspace changes.
- **Validation:** `cargo test --lib` (2/2), `clippy -D warnings` clean, `fmt --check` clean, WASM build succeeds. Source committed, `.wasm` and `target/` gitignored.

Standard plugin validation complete.

### 🟢 What looks good — Output fidelity

Output terminates with the mandatory fidelity footer (`query`, `results`), preventing hallucinated geocoding data. Matches plugin output standards.

Commit title `feat(plugins): add nominatim geocoding plugin` follows the conventional commit format.

---

**Verdict:** Ready to merge. This is a practical, well-executed plugin that fills a real capability gap in the self-hosted OSS tier. Clean implementation, appropriate security boundaries, solid validation. Well done, @theonlyhennygod.

Audacity88 (changes requested):
I reviewed current head 7b2ae6b91dfb83e97f0f8c425ad3b0d03a8481f7 against the PR body, WareWolf's approval review, current green checks, and the full `plugins/nominatim/` diff. I did not rerun local cargo checks for this pass.

In simple terms, this adds a standalone geocoding plugin. It can turn a place name into coordinates, or turn coordinates back into an address, using OpenStreetMap Nominatim by default or an operator-configured `NOMINATIM_URL`.

### 🟢 What looks good — The plugin boundary and credential shape are clean

The standalone plugin shape is right for this capability. The change is isolated under `plugins/nominatim/`, does not touch the host/runtime/root workspace, and asks for only `http_client` plus `env_read`.

The credential boundary is clean: this plugin is keyless, and model-visible arguments do not choose a host that receives an API key. `NOMINATIM_URL` is an operator-controlled environment override, not a per-call model argument, and the request includes the User-Agent Nominatim expects.

### 🔴 Blocking — HTTP error truncation can panic on non-ASCII responses

The HTTP error path slices the response body by byte index:

Rust string slicing requires the end index to be a valid UTF-8 character boundary. If Nominatim or a self-hosted Nominatim-compatible server returns an error body longer than 500 bytes and byte 500 lands inside a multibyte character, the plugin panics instead of returning the intended `ToolResult::failure`. Localized provider errors, place names, and address text make non-ASCII response bodies normal input for a geocoding plugin, not an edge case.

Please switch this path to UTF-8-safe truncation, for example by collecting `resp.body.chars().take(500)` or by using a helper that finds a valid boundary at or before the cap. Please also add a regression test with a long non-ASCII error body so the failure path proves it returns a normal tool failure instead of panicking.

The PR body still ends with a Claude Code attribution footer, which the PR template asks contributors not to include. Since this is a medium-risk external-network plugin, the body should also use the medium-risk rollback shape and make the validation target explicit enough that reviewers can tell which WASM target was built.

theonlyhennygod:
Addressed the blocking UTF-8 truncation issue in 38a2536.

- Replaced the byte-slice `&resp.body[..resp.body.len().min(500)]` in the HTTP error path with a UTF-8-safe `truncate_chars()` helper that walks back to the nearest char boundary, so a non-ASCII Nominatim error body whose 500-byte cutoff lands mid-character can no longer panic.
- Added regression tests (`truncate_chars_never_splits_multibyte`, `truncate_chars_short_input_unchanged`).
- Also removed the Claude Code attribution footer from the PR body per the template.

Audacity88 (changes requested):
I reviewed current head 38a2536700e995e9da065d32183abd681c44b16b against the PR body, @theonlyhennygod's follow-up comment, @WareWolf-MoonWall's earlier approval, my previous `CHANGES_REQUESTED` review, current green CI, the full `plugins/nominatim/` diff, and the current OSMF Nominatim usage policy. I also reran focused plugin-local validation: `cargo test --manifest-path plugins/nominatim/Cargo.toml --lib -- --nocapture` passed 4/4, `cargo fmt --manifest-path plugins/nominatim/Cargo.toml -- --check` passed, and `cargo clippy --manifest-path plugins/nominatim/Cargo.toml --lib -- -D warnings` passed.

### ✅ Resolved — HTTP error truncation is UTF-8 safe now

The blocker from my previous review is fixed. The 4xx/5xx Nominatim error path now uses `truncate_chars(&resp.body, 500)` instead of slicing the response body with a raw byte index, and the new regression coverage proves a long multi-byte error body stays on a valid character boundary.

### 🔴 Blocking — The public Nominatim service cannot be the silent default

The code still falls back to `https://nominatim.openstreetmap.org` whenever `NOMINATIM_URL` is unset, and the tool metadata presents the public endpoint as an ordinary built-in option. The current OSMF Nominatim usage policy is narrower than that default: it is an acceptable-use policy for the donated public service, requires visible attribution and strict rate-limit/caching discipline, and says LLM-suggested use must prominently point users to the policy and explain the restrictions. It also says the public API must not be built into or offered through no-code, low-code, or vibe-coding platforms as a generic geocoding, address lookup, place search, or map search service: https://operations.osmfoundation.org/policies/nominatim/

For a ZeroClaw agent plugin, a silent public default does not give the operator an explicit, informed choice before routing model-triggered geocoding traffic to that public service. It also does not surface the attribution, caching, rate-limit, and acceptable-use boundaries at the point where the operator enables the tool.

Please make the self-hosted/operator-provided endpoint the default shape. The cleanest fix is to require `NOMINATIM_URL` and fail with a clear setup message when it is missing. If we want to keep support for the public OSM endpoint, it should be a separate explicit opt-in with the OSMF policy, attribution, rate-limit, and caching expectations surfaced clearly enough that the operator is choosing that backend deliberately.

### 🟢 What looks good — The plugin boundary is still narrow

The useful parts of the plugin shape held up in the follow-up: it remains isolated under `plugins/nominatim/`, uses only `http_client` and `env_read`, avoids host/runtime/root-workspace changes, keeps the endpoint operator-controlled rather than model-controlled, and includes a fidelity footer that tells the model which fields were actually returned.

singlerider:

PR #7328, opened 2026-06-06
feat(plugins): add n8n workflow-trigger plugin

theonlyhennygod:
## Summary
- **Base:** `master`. New WASM/Extism tool plugin `plugins/n8n/` exposing a `trigger_workflow` tool that POSTs a payload to a workflow's webhook on the user's own **self-hosted n8n** instance (`N8N_BASE_URL`, default `http://localhost:5678`; optional `N8N_AUTH_TOKEN`) and returns the workflow's response.
- **High leverage:** because n8n connects to hundreds of apps, this one plugin lets the agent reach the user's **entire n8n integration catalog** — the OSS automation "action layer." Uses only `http_client` + `env_read`. Fidelity footer (`webhook_path, response`).

## Validation Evidence

## Security & Privacy
- `http_client` + `env_read`; no filesystem access. Outbound HTTP to the configured `N8N_BASE_URL` only (the workflow it triggers is operator-defined). Optional `N8N_AUTH_TOKEN` (Bearer), never logged; no secrets in the diff. `webhook_path` is validated before use.

## Compatibility
- Backward compatible (additive). Optional env vars. Opt-in.

## Rollback
- `git revert <commit>` / delete `plugins/n8n/`.


WareWolf-MoonWall (approved):
### 🟢 What looks good — High-leverage integration point

The `n8n` plugin is architecturally significant:

- **Multiplier effect:** Because n8n connects to hundreds of apps (CRMs, ticketing systems, databases, notification services, etc.), this single plugin effectively grants the agent reach into the user's entire n8n integration catalog. This is the "action layer" for self-hosted automation — high strategic value.
- **Minimal surface:** The plugin itself is simple (POST payload to webhook, return response) — all the complexity lives in the user's n8n workflows, where it belongs. Clean separation of concerns.

### 🟢 What looks good — Security and privacy boundaries

- **Permissions:** `http_client` + `env_read` only. Outbound HTTP to the configured `N8N_BASE_URL` exclusively. No filesystem access.
- **Auth handling:** Optional `N8N_AUTH_TOKEN` as `Bearer` header, never logged. No secrets in the diff.
- **Validation:** `webhook_path` validated before use.

This is the right security posture for a plugin that triggers user-defined workflows.

### 🟢 What looks good — Workspace isolation

Own `[workspace]` table in `Cargo.toml` — self-contained in `plugins/n8n/`, no root workspace changes, no conflict potential with sibling plugin PRs.

`cargo test --lib` (2/2), `clippy -D warnings` clean, `fmt --check` clean, WASM build succeeds. Source committed, `.wasm` and `target/` gitignored. Standard plugin validation complete.

Commit title `feat(plugins): add n8n workflow-trigger plugin` follows the conventional commit format.

---

**Verdict:** This is ready to merge. The architectural leverage this plugin provides (agent → n8n → hundreds of third-party apps) is significant, and the implementation is clean and secure. Excellent addition to the self-hosted OSS tier. Well done, @theonlyhennygod.

Audacity88 (changes requested):
I reviewed the live PR state, current diff, plugin manifest, PR body, current checks, and @WareWolf-MoonWall's approval review. I agree with the direction: a self-hosted n8n trigger plugin is a high-leverage integration point, and keeping the base URL/operator token in env vars gives this a much better security shape than letting the model choose a credential destination.

### 🟢 What looks good — The integration boundary is pointed the right way

The plugin keeps the model-visible surface small: the model chooses a webhook path and optional JSON payload, while the operator controls `N8N_BASE_URL` and `N8N_AUTH_TOKEN`. That is the right split for an action-layer plugin. It lets users expose their own n8n workflows without giving the model a raw arbitrary-host HTTP tool.

The manifest is also appropriately narrow for this shape: `tool`, `http_client`, and `env_read`, with no filesystem permission.

### 🔴 Blocking — Response truncation can panic on UTF-8

There are two byte-slice truncation paths in `execute()`:

`&response_raw[..response_raw.len().min(MAX_RESPONSE_CHARS)]`

Both slice a `String` by byte index. If an n8n error body or a successful workflow response is longer than the cap and the cutoff lands inside a multi-byte UTF-8 character, the plugin panics instead of returning a normal `ToolResult::failure` or a truncated success response. This is a real user-facing failure mode because workflow responses can contain arbitrary app data, including non-ASCII text.

Please switch both truncation paths to a UTF-8-safe helper and add regression coverage for both cases: a long non-ASCII error response and a long non-ASCII successful workflow response.

theonlyhennygod:
Addressed the blocking UTF-8 truncation issue in c1e4edd.

- Switched **both** truncation paths to a UTF-8-safe `truncate_chars()` helper:
  - the error preview `&resp.body[..resp.body.len().min(500)]`, and
  - the successful-response cap `&response_raw[..response_raw.len().min(MAX_RESPONSE_CHARS)]`.
- Both now walk back to the nearest char boundary, so a long non-ASCII n8n error body or workflow response no longer panics.
- Added regression coverage for a long non-ASCII error body and a long non-ASCII successful workflow response.

Audacity88 (changes requested):
I reviewed current head c1e4eddf550d6165caa2b9377f29f46fa18a7a8d against the PR body, @theonlyhennygod's follow-up comment, @WareWolf-MoonWall's earlier approval, my previous `CHANGES_REQUESTED` review, current green CI, and the full `plugins/n8n/` diff. I also reran focused plugin-local validation: `cargo test --manifest-path plugins/n8n/Cargo.toml --lib -- --nocapture` passed 4/4, `cargo fmt --manifest-path plugins/n8n/Cargo.toml -- --check` passed, and `cargo clippy --manifest-path plugins/n8n/Cargo.toml --lib -- -D warnings` passed.

### 🟢 What looks good — The code path now uses a UTF-8-safe helper

The implementation moved in the right direction. Both the n8n error preview path and the successful workflow-response preview path now call `truncate_chars(...)` instead of slicing the response body with a raw byte index. The plugin boundary also still looks right: isolated under `plugins/n8n/`, `http_client` plus `env_read`, localhost/self-hosted default, operator-controlled base URL/token, and a small model-visible webhook/payload surface.

### 🔴 Blocking — The regression tests do not actually split a character

My previous blocker asked for regression coverage where the truncation cap lands inside a multi-byte character. The new tests intend to cover that, but the current fixtures do not actually create the old failure condition:

`"é".repeat(400)` is 800 bytes, and the 500-byte error cap lands on a valid boundary because `é` is two bytes.

`"ü".repeat(MAX_RESPONSE_CHARS)` is 16,000 bytes, and the 8,000-byte success cap also lands on a valid boundary because `ü` is two bytes.

That means the tests would not have caught the original panic. They prove the helper can truncate non-ASCII text, but not that it handles a cutoff inside a code point.

Please adjust both regression fixtures so the cap lands inside a character, for example by using a three-byte or four-byte character, or by choosing a test cap that is not aligned with the repeated character width. Once those tests actually exercise a mid-character cutoff, this should clear my remaining blocker.

singlerider:

PR #7329, opened 2026-06-06
feat(plugins): add meilisearch self-hosted search plugin

theonlyhennygod:
## Summary
- **Base:** `master`. New WASM/Extism tool plugin `plugins/meilisearch/` exposing a `search` tool that runs full-text search against an index on the user's own **self-hosted Meilisearch** server (`MEILISEARCH_URL`, default `http://localhost:7700`; optional `MEILISEARCH_API_KEY`) — search-your-own-data. `POST /indexes/<index>/search {q,limit}`, returns matching documents. Uses only `http_client` + `env_read`. Fidelity footer (`index, query, hits`).

## Validation Evidence

## Security & Privacy
- `http_client` + `env_read`; no filesystem access. Outbound HTTP to the configured `MEILISEARCH_URL` only. Optional `MEILISEARCH_API_KEY` (Bearer), never logged; no secrets in the diff. `index` is validated before use.

## Compatibility
- Backward compatible (additive). Optional env vars. Opt-in.

## Rollback
- `git revert <commit>` / delete `plugins/meilisearch/`.


WareWolf-MoonWall (approved):
### 🟢 What looks good — Self-contained plugin pattern

The `meilisearch` plugin follows the established plugin architecture precisely:

- **Workspace isolation:** Own `[workspace]` table in `Cargo.toml` — no root workspace changes, zero conflict potential with sibling plugin PRs.
- **Security surface:** `http_client` + `env_read` only. Outbound HTTP restricted to the configured `MEILISEARCH_URL`. Optional `MEILISEARCH_API_KEY` handled as `Bearer` token, never logged. Index name validated before use. Clean security posture.
- **Validation coverage:** `cargo test --lib` (2/2), `clippy -D warnings` clean, `fmt --check` clean, WASM build succeeds. The test suite is appropriate for the scope.

### 🟢 What looks good — Fidelity footer enforcement

Output terminates with the mandatory plugin-output-fidelity footer (`index`, `query`, `hits`), preventing the agent from hallucinating search results that didn't come from the server. This matches the plugin output standards.

### 🟢 What looks good — Strategic value

This plugin fills a real gap in the self-hosted OSS tier: full-text search against the user's own data. Because Meilisearch is trivially self-hostable (single binary, default port 7700), this gives agents production-grade search capabilities without vendor lock-in. High leverage for the ecosystem.

Commit title `feat(plugins): add meilisearch self-hosted search plugin` follows the conventional commit format.

---

**Verdict:** Ready to merge. Clean implementation, appropriate security boundaries, solid validation evidence, and valuable addition to the self-hosted plugin catalog. Well done, @theonlyhennygod.

Audacity88 (changes requested):
I reviewed #7329 at head 7bb0ded, including the PR body, current green CI, @WareWolf-MoonWall's approval review, the full `plugins/meilisearch` diff, and the merged `plugins/image-gen-fal` baseline. I agree with the overall plugin direction, but I found one edge-case blocker that should be fixed before merge.

### 🟢 What looks good — The self-hosted plugin shape is useful

The plugin stays nicely scoped: independent WASM workspace, `http_client` plus `env_read` only, no root workspace or host/runtime changes, and a clear default to local/self-hosted Meilisearch. The index validation, result-count cap, and fidelity footer are all the right instincts for a search-your-own-data tool.

### 🔴 Blocking — Non-ASCII Meilisearch data can panic during truncation

Two paths truncate UTF-8 strings by byte index:

and:

`s[..s.len().min(MAX_HIT_CHARS)].to_string()`

Both strings come from Meilisearch or indexed user documents, so non-ASCII text is normal here. If the cutoff lands inside a multibyte character, Rust string slicing will panic instead of returning a structured `ToolResult::failure` or a capped hit preview. That turns an ordinary external-service/user-data edge case into a plugin trap.

Please truncate on character boundaries, for example with a small helper that uses `.chars().take(limit).collect::<String>()`, and add focused regression coverage for non-ASCII text in both the error-body path and the hit-summary path. The plugin should keep returning normal tool output/failure even when the indexed data or server error contains multilingual text.

singlerider:

PR #7313, opened 2026-06-06
feat(plugins): add suno open-source music-generation plugin

theonlyhennygod:
## Summary
- **Base:** `master`. Adds a self-contained WASM/Extism tool plugin `plugins/suno/` exposing a `generate_music` tool that creates songs from a text prompt.
- **Open-source, user-owned by design:** it targets the popular self-hosted [`gcui-art/suno-api`](https://github.com/gcui-art/suno-api) wrapper (the user runs their own instance against their own Suno account) — **no vendor key, no lock-in**. The plugin only needs the instance base URL `SUNO_API_URL` (default `http://localhost:3000`) and an optional `SUNO_API_TOKEN`.
- Flow: `POST /api/generate` with `wait_audio`, plus a **bounded `GET /api/get` poll** fallback (`MAX_POLLS=12`) to resolve within the host's 120s budget. Returns hosted audio **URL(s)** (text output — no binary host support needed). Uses only `http_client` + `env_read`. Output ends with the mandatory **fidelity footer** (`Fields returned: prompt, clips`).
- **Scope:** purely additive; no host/runtime/config/root-`Cargo.toml` changes. Mirrors the `firecrawl`/`replicate` reference plugins (own `[workspace]`).

## Validation Evidence
- `cargo test --lib` → 3/3 (clip parsing extracts ready audio + ids, footer present/last/fields, percent-encode).
- `cargo clippy --all-targets -- -D warnings` → clean; `cargo fmt --check` → clean.
- `cargo build --target wasm32-unknown-unknown --release` → wasm builds.
- **Integration:** loaded the built `.wasm` into a `plugins-wasm` gateway and confirmed `loaded: true` via `GET /api/plugins`.
- Not exercised: a live generation (requires a running self-hosted suno-api with a configured Suno cookie). Source committed; `.wasm`/`target/` gitignored.

## Security & Privacy
- Declares `http_client` + `env_read`; no filesystem access. Outbound HTTPS/HTTP to the **operator-configured** `SUNO_API_URL` only. Optional `SUNO_API_TOKEN` (Bearer), never logged; no secret in the diff. The Suno account credential lives on the user's self-hosted wrapper, never in this plugin.

## Compatibility
- Backward compatible (additive). Optional env vars `SUNO_API_URL` / `SUNO_API_TOKEN`. Opt-in (requires `[plugins].enabled` + a `plugins-wasm` build + a reachable self-hosted suno-api).

## Rollback
- `git revert <commit>` / delete `plugins/suno/`.

[...]

PR #7235, opened 2026-06-05
feat(web): plugin lifecycle endpoints + management UI (stubs)

theonlyhennygod:
> **Stacked PR — Depends on #7229.** Base branch is `feat/web-mcp-tab`, not `master`. Merge **after** #7229. Review the diff against `feat/web-mcp-tab`.

## Summary

- **Base branch:** `feat/web-mcp-tab` (stacked on #7229, which adds the read-only Plugins tab). Final merge target is `master` via the stack.
- **What changed and why:** Adds plugin lifecycle endpoints + UI controls so the Plugins dashboard tab can eventually manage plugins, not just view them.
  - Gateway: `POST /api/plugins/install`, `DELETE /api/plugins/{name}`, `POST /api/plugins/enabled` (all behind the `plugins-wasm` feature), gated by a shared `require_auth()` helper.
  - Client: `installPlugin` / `removePlugin` / `setPluginsEnabled` in `web/src/lib/api.ts`.
  - UI: enable/install/remove controls on the Plugins tab.
- **Stub note (important):** Endpoints intentionally return `stub: true`. Requests are validated and accepted but are **NOT** wired to `PluginHost` — nothing is actually installed, removed, or persisted. The UI surfaces the stub message verbatim plus a banner so no action falsely reports success. Real `PluginHost` wiring is a follow-up; the API surface won't change.
- **Scope boundary:** No `PluginHost` integration, no disk writes, no config persistence. Endpoints only register when built with `plugins-wasm`. Does not touch the read-only listing added in #7229.
- **Blast radius:** Limited to the gateway plugin API surface (`plugins-wasm` only) and the web Plugins tab. No effect on non-`plugins-wasm` builds or other dashboard tabs.
- **Linked issue(s):** Depends on #7229
- **Labels:** `type: feature`, `risk: low`, `size: M`, `web`

## Validation Evidence (required)

- **Commands run and tail output:**
  - `cargo check -p zeroclaw-gateway --features plugins-wasm` — passed (exit 0), no warnings on new code.
  - Frontend `tsc -b` — passed (exit 0).
- **Beyond CI — what did you manually verify?** Verified the new routes register and compile under `plugins-wasm`, the stub responses carry `stub: true` + an explanatory message, and the UI wiring type-checks. NOT manually click-tested end-to-end: the running daemon was built without `plugins-wasm`, so the new routes don't register at runtime in that build. A `plugins-wasm` daemon build is needed to exercise the endpoints live.
[...]

PR #7331, opened 2026-06-06
feat(plugins): add ace-step self-hosted music generation plugin

theonlyhennygod:
## Summary

Adds **`plugins/ace-step/`** — a WASM (Extism) tool plugin that generates music from a text prompt via a **self-hosted [ACE-Step](https://github.com/ace-step/ACE-Step)** open-weights model server.

This is the **"own your stack, pay no one"** complement to the `suno` plugin (#7313). The distinction matters:

| Plugin | Backend | Cost to user |
|---|---|---|
| `suno` (#7313) | self-hosted `gcui-art/suno-api` **wrapper** | still drives the user's **paid** Suno account |
| **`ace-step` (this PR)** | self-hosted **ACE-Step open-weights model** | runs on the user's own GPU — **no vendor key, no per-gen cost** |

ACE-Step 1.5 runs locally in <4GB VRAM (~10s/song on an RTX 3090; Mac/AMD/Intel/CUDA), so users get a true vendor-free music-generation path — matching the existing self-hosted OSS tier (`sd-webui`, `ollama-embed`, `libretranslate`, `searxng`).

## Tool: `generate_music`

| Param | Required | Description |
|---|---|---|
| `prompt` | ✅ | Style/genre/mood tags, comma-separated (e.g. `lo-fi hip hop, mellow piano, 90 bpm`) |
| `lyrics` | — | Lyrics with `[verse]`/`[chorus]` structure tags; omit for instrumental |
| `duration` | — | Seconds, 1–240 (default 30) |

`POST`s `{tags, lyrics, audio_duration}` (the standard ACE-Step inputs) to a **configurable** endpoint and extracts the generated audio from the response.

### Configuration (all optional — sane local defaults)
- `ACE_STEP_URL` — server base URL (default `http://localhost:7865`, ACE-Step's default port)
- `ACE_STEP_PATH` — generation endpoint path (default `/generate`)
- `ACE_STEP_TOKEN` — optional `Bearer` token if the server is auth-protected

### Response handling
The self-hosted server landscape isn't fully standardized, so the plugin accepts the common shapes:
- **Audio URL** — `audio_url` / `url` / `data[0].url` (Gradio file URLs, served paths)
- **Base64-in-JSON** — `audio_base64` / `audio` (returned as a `data:audio/wav;base64,…` URI)

The host HTTP bridge is **text-only**, so the server must return a URL or base64-in-JSON (not a raw binary body).

## Output fidelity
Output ends with the mandatory `plugin-output-fidelity` footer (data source · fields returned · do-not-infer), enforced by unit tests.

## Validation
- ✅ `cargo test --lib` — **9 passing** (duration clamp, URL join, audio extraction across 5 response shapes, footer placement/ordering, instrumental vs. base64 rendering)
- ✅ `cargo clippy --all-targets -- -D warnings` — clean
- ✅ `cargo fmt --check` — clean
- ✅ `cargo build --target wasm32-unknown-unknown --release` — builds `ace_step.wasm` (~250 KB)
- ⏳ Gateway load-test (`plugins-wasm` → `GET /api/plugins` shows `loaded:true`) — confirmation comment to follow; structurally identical to the already-load-tested plugin set.

## Notes for reviewers
- Self-contained in `plugins/ace-step/` with its own `[workspace]` table — no root `Cargo.toml` change, so it does not conflict with sibling plugin PRs.
- No secrets in the diff (config is read at runtime via `zc_env_read`).

Milestone: **v0.8.2**


theonlyhennygod:
### ✅ Gateway load-test passed — `loaded: true`

Built a `--features plugins-wasm` gateway and loaded the compiled `ace_step.wasm` into a real `plugins_dir`. `GET /api/plugins` instantiated the module successfully:

```json
{
  "plugins": [
    {
      "name": "ace-step",
      "version": "0.1.0",
      "description": "Generate music from text prompts using a self-hosted ACE-Step open-weights model (no vendor API)",
      "capabilities": ["tool"],
      "loaded": true
    }
  ],
  "plugins_dir": ".../plugins",
  "plugins_enabled": true
}
```

So the host can instantiate the WASM module and expose `generate_music` end-to-end.

> ℹ️ Reproduction note: the `/api/plugins` route only registers when the top-level `plugins-wasm` feature forwards to `zeroclaw-gateway/plugins-wasm`. On `master` that forwarding lands via the open root-`Cargo.toml` PR (#7235); I applied it locally **only** for this load-test and reverted it — this PR touches nothing outside `plugins/ace-step/`.

WareWolf-MoonWall (approved):
### 🟢 What looks good — Clean plugin architecture

The `ace-step` plugin exemplifies the self-contained plugin pattern:

- **Isolation:** Own `[workspace]` in `Cargo.toml` means zero conflict with sibling plugin PRs or the root workspace — exactly the right boundary.
- **Security surface:** `http_client` + `env_read` permissions match the stated contract (outbound HTTP to user-configured endpoint, runtime config). No filesystem access, no secret leakage in the diff.
- **Validation rigor:** The suite covers the critical boundaries — duration clamping (1–240s), URL joining, 5 different response shape parsers (URL variants + base64), footer presence/ordering, and instrumental vs. base64 rendering. This is thorough for a plugin of this scope.
- **Gateway load-test evidence:** The follow-up comment with the `loaded: true` response from `GET /api/plugins` demonstrates end-to-end WASM instantiation. The reproduction note about the `plugins-wasm` feature dependency is helpful context.

### 🟢 What looks good — Output fidelity discipline

The footer enforcement (tested in unit tests, as stated) ensures the agent runtime gets structured metadata about what the tool returned. This matches the plugin output standards and prevents hallucination drift.

### 🟢 What looks good — Self-hosted OSS positioning

The PR description clearly distinguishes this plugin from the `suno` plugin (#7313) — `ace-step` is the true vendor-free path (user's own GPU, no API key, no per-generation cost). The comparison table is concise and meaningful. This is a valuable addition to the self-hosted tier alongside `sd-webui`, `ollama-embed`, `libretranslate`, and `searxng`.

Commit title `feat(plugins): add ace-step self-hosted music generation plugin` follows the conventional commit format exactly.

---

**Verdict:** This PR is ready to merge. Solid plugin implementation, comprehensive validation, clean boundaries, and excellent documentation. Well done, @theonlyhennygod.

Audacity88 (changes requested):
I reviewed current head `966ecdbf0bc3f6342de8c94b9a9b97b9bcdcae89` against the live PR state, the existing approval, the gateway load-test comment, green CI, and the new `plugins/ace-step` crate. I agree with the overall plugin shape called out in the prior approval, but I found one failure-path panic that should be fixed before merge. I did not rerun local cargo; I relied on the author's listed plugin validation, the gateway load-test evidence, green GitHub checks, and source inspection.

### 🟢 What looks good — standalone plugin boundary is clean

The PR keeps ACE-Step isolated under `plugins/ace-step/` with its own workspace and WASM manifest. The declared permissions are also narrow for the behavior: `http_client` for the configured self-hosted endpoint and `env_read` for `ACE_STEP_URL`, `ACE_STEP_PATH`, and optional `ACE_STEP_TOKEN`. I did not see filesystem access or token material copied into the output.

### 🔴 Blocking — error previews can panic on non-ASCII server bodies

`execute()` builds the server-error and "no audio found" failure messages by slicing `resp.body: String` with `&resp.body[..resp.body.len().min(500)]`. That cutoff is byte-based, not character-boundary-based. If a self-hosted ACE-Step server returns non-ASCII JSON or error text and byte 500 lands inside a multibyte UTF-8 character, the plugin will panic while trying to report a normal tool failure.

Please replace both previews with a UTF-8-safe truncation helper, then add regression coverage with a non-ASCII response body around the cutoff for both paths:

- HTTP error response (`resp.status >= 400`)
- successful JSON response that contains no supported audio URL/base64 field

### 🟢 What looks good — response shapes and output fidelity are well covered

The successful paths cover the expected URL and base64 response shapes, duration clamping, URL joining, and footer placement. Once the failure-preview truncation is made character-safe too, the plugin shape looks good to me.

theonlyhennygod:
Addressed the blocking failure-path panic in 5f1062c.

- Replaced **both** byte-slice previews in `execute()` (`&resp.body[..resp.body.len().min(500)]`) — the HTTP error path (`resp.status >= 400`) and the no-supported-audio path — with a UTF-8-safe `truncate_chars()` helper that walks back to the nearest char boundary.
- Added a regression test exercising both failure previews with a non-ASCII body around the cutoff, plus a short-input test.
- Also removed the Claude Code attribution footer from the PR body.

Audacity88 (changes requested):
I reviewed current head 5f1062cdcbc08ed7bac6bf25c1ac72b96e6d101d against the PR body, @theonlyhennygod's follow-up comment, @WareWolf-MoonWall's earlier approval, my previous `CHANGES_REQUESTED` review, the gateway load-test comment, current green CI, and the full `plugins/ace-step/` diff. I also reran focused plugin-local validation: `cargo test --manifest-path plugins/ace-step/Cargo.toml --lib -- --nocapture` passed 11/11, `cargo fmt --manifest-path plugins/ace-step/Cargo.toml -- --check` passed, and `cargo clippy --manifest-path plugins/ace-step/Cargo.toml --lib -- -D warnings` passed.

### 🟢 What looks good — The failure paths now use a UTF-8-safe helper

The implementation moved in the right direction. Both failure previews now call `truncate_chars(...)` instead of slicing `resp.body` with a raw byte index, and the standalone ACE-Step plugin boundary still looks good: isolated under `plugins/ace-step/`, narrow `http_client` plus `env_read` permissions, operator-controlled endpoint/token configuration, and the existing response-shape and fidelity-footer coverage.

### 🔴 Blocking — The regression test does not actually split a character

My previous blocker asked for regression coverage with a non-ASCII response body around the cutoff for both failure paths. The new test intends to cover that, but the current fixture does not actually create the old failure condition:

`"é".repeat(400)` is 800 bytes, and the 500-byte cap lands on a valid boundary because `é` is two bytes.

That means the test would not have caught the original panic. It proves the helper can truncate non-ASCII text, but not that it handles a cutoff inside a code point.

Please adjust the regression fixture so the 500-byte cap lands inside a character, for example by using a three-byte or four-byte character, or by choosing a test cap that is not aligned with the repeated character width. Once the test actually exercises a mid-character cutoff, this should clear my remaining blocker.

The `truncate_chars()` helper implementation itself looks correct; the remaining request is to make the test fixture actually exercise the mid-character path so it protects against the original panic.

singlerider:

PR #7333, opened 2026-06-06
feat(plugins): remote plugin registry — `zeroclaw plugin search` + install-by-name

theonlyhennygod:
## Summary

The local plugin CLI already works (`install <dir>`, `list`, `remove`, `info`), but there was **no way to discover or install published plugins** — users hand-copied `.wasm` into `~/.zeroclaw/plugins/`. With ~33 plugins now shipped, that's the adoption bottleneck. This PR wires the **remote half**: a registry so plugins are installable **by name**.

## What's new

```bash
zeroclaw plugin search music                 # discover
zeroclaw plugin install ace-step             # resolve → download → verify → install
zeroclaw plugin install ace-step@0.1.0       # pin a version
zeroclaw plugin search music --registry https://my-host/registry.json   # private catalog
```

- **`zeroclaw plugin search <query>`** — case-insensitive match on name/description.
- **`zeroclaw plugin install <name>`** now auto-detects: an existing **local path** installs locally (unchanged); anything else is a **registry name** (`name` or `name@version`).
- **Registry** = a single JSON index. Default points at `zeroclaw-labs/zeroclaw-plugins`; override per-invocation with `--registry` or globally with **`ZEROCLAW_PLUGIN_REGISTRY_URL`** → point it at your own host for a **private / self-hosted catalog**.

## Security & integrity

- Installs now build the host via **`with_security`**, so they honor the configured **`signature_mode`** + trusted publisher keys (the previous handler always used `Disabled`). Ed25519 manifest verification guards **authenticity**.
- Optional **SHA-256** per entry guards **transport**.
- Downloads capped at **50 MiB**; zip extraction **rejects path-traversal** (`..`, absolute paths).

## Registry index format

```json
{
  "plugins": [
    {
      "name": "ace-step",
      "version": "0.1.0",
      "description": "Self-hosted music generation",
      "author": "ZeroClaw Labs",
      "capabilities": ["tool"],
      "url": "https://.../ace-step-0.1.0.zip",
      "sha256": "<hex digest of the zip>"
    }
  ]
}
```
Each `url` is a zip of the plugin directory (`manifest.toml`, the `.wasm`, optional `skills/`) — exactly what `PluginHost::install` already consumes.

## Implementation
- New `src/plugin_registry.rs` (fetch index, resolve `name@version`, search filter, SHA-256 verify, capped + traversal-safe extraction) → hands the extracted dir to the existing `PluginHost::install`. No change to the install/signature core.
- `src/main.rs`: `Search` subcommand + `--registry` flag; `Install` resolves registry names and uses `with_security`.
- Docs: registry format, commands, and self-hosting added to the plugin-protocol guide.

## Validation
- ✅ **8 unit tests** (`name@version` parsing, version-pin vs newest, search filter, registry-URL precedence, SHA-256 verify/reject, manifest-dir discovery)
- ✅ `cargo clippy --bin zeroclaw --features plugins-wasm` — clean · `cargo fmt --check` — clean
- ✅ **End-to-end**: stood up a local registry serving the real `ace-step` plugin zip and ran the actual binary:
  ```
  $ zeroclaw plugin search music
  Plugins matching 'music' (1):
    ace-step v0.1.0 — Self-hosted music generation (no vendor API)
  $ zeroclaw plugin install ace-step
  Resolving 'ace-step' from registry …
  Downloading ace-step v0.1.0 …
    ✓ sha256 verified
  ✓ Installed ace-step v0.1.0
  $ zeroclaw plugin list
  Installed plugins:
    ace-step v0.1.0 — Generate music from text prompts using a self-hosted ACE-Step open-weights model
  ```

## Follow-up (not in this PR)
- Publish/populate the canonical `zeroclaw-labs/zeroclaw-plugins` registry repo (the default URL). The feature works against any host today via `--registry`/env, so this is decoupled.

Milestone: **v0.8.3**

### DIFF
--- a/docs/book/src/developing/plugin-protocol.md
+++ b/docs/book/src/developing/plugin-protocol.md
@@ -277,12 +277,58 @@ The output `.wasm` file is at
 ### Installing
 
 ```bash
-# Copy to plugin directory
+# From a local directory
 zeroclaw plugin install /path/to/my-plugin/
+
+# By name from the registry (resolves a download, verifies, and installs)
+zeroclaw plugin install ace-step
+zeroclaw plugin install ace-step@0.1.0      # pin a version
 
 # Or manually
 cp -r my-plugin/ ~/.zeroclaw/plugins/my-plugin/
 ```
+
+`install` treats the argument as a local path when it exists on disk, and
+otherwise as a registry name. All installs honor the configured
+`signature_mode` (see [Configuration](#configuration)).
+
+### Registry
+
+Browse remotely published plugins:
+
+```bash
+zeroclaw plugin search music
+zeroclaw plugin search music --registry https://my-host/registry.json
+```
+
+The registry is a single JSON index (default: the public
+`zeroclaw-labs/zeroclaw-plugins` index). Override it per-invocation with
+`--registry`, or globally with the `ZEROCLAW_PLUGIN_REGISTRY_URL` environment
+variable — point it at your own host to run a private/self-hosted catalog.
+
+Index format:
+
+```json
+{
+  "plugins": [
+    {
+      "name": "ace-step",
+      "version": "0.1.0",
+      "description": "Self-hosted music generation",
+      "author": "ZeroClaw Labs",
+      "capabilities": ["tool"],
+      "url": "https://.../ace-step-0.1.0.zip",
+      "sha256": "<hex digest of the zip>"
+    }
+  ]
+}
+```
+
+Each `url` points to a zip of the plugin directory (the same layout `install`
+expects: `manifest.toml`, the `.wasm`, and an optional `skills/` subtree). The
+optional `sha256` guards transport integrity; the Ed25519 manifest signature
+(per `signature_mode`) guards authenticity. Archives are capped at 50 MiB and
+extracted with path-traversal protection.
 
 ## Configuration
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -266,6 +266,8 @@ mod peripherals;
 #[cfg(feature = "agent-runtime")]
 mod platform;
 #[cfg(feature = "plugins-wasm")]
+mod plugin_registry;
+#[cfg(feature = "plugins-wasm")]
 mod plugins;
 mod providers;
 #[cfg(feature = "schema-export")]
@@ -2375,10 +2377,21 @@ fn which_zerocode_on_path() -> bool {
 enum PluginCommands {
     /// List installed plugins
     List,
-    /// Install a plugin from a directory or URL
+    /// Search the plugin registry
+    Search {
+        /// Substring to match against plugin name or description
+        query: String,
+        /// Registry index URL (overrides $ZEROCLAW_PLUGIN_REGISTRY_URL and the default)
+        #[arg(long)]
+        registry: Option<String>,
+    },
+    /// Install a plugin from a local directory, or by name from the registry
     Install {
-        /// Path to plugin directory or manifest
+        /// Local plugin directory/manifest path, or a registry name (`name` or `name@version`)
         source: String,
+        /// Registry index URL (overrides $ZEROCLAW_PLUGIN_REGISTRY_URL and the default)
+        #[arg(long)]
+        registry: Option<String>,
     },
     /// Remove an installed plugin
     Remove {
@@ -5204,18 +5217,37 @@ async fn main() -> Result<()> {
                 }
                 Ok(())
             }
-            PluginCommands::Install { source } => {
-                let mut host = zeroclaw::plugins::host::PluginHost::new(&config.data_dir)?;
-                host.install(&source)?;
-                println!(
-                    "{}",
-                    ta(
-                        "cli-plugin-installed-from",
-                        &[("source", &source)],
-                        "Plugin installed"
-                    )
+            PluginCommands::Search { query, registry } => {
+                plugin_registry::search_and_print(registry.as_deref(), &query).await
+            }
+            PluginCommands::Install { source, registry } => {
+                // Honor the configured signature policy for installs (not just discovery).
+                let mode = zeroclaw::plugins::host::PluginHost::parse_signature_mode(
+                    &config.plugins.security.signature_mode,
                 );
-                Ok(())
+                let trusted = config.plugins.security.trusted_publisher_keys.clone();
+                let mut host = zeroclaw::plugins::host::PluginHost::with_security(
+                    &config.data_dir,
+                    mode,
+                    trusted,
+                )?;
+                if std::path::Path::new(&source).exists() {
+                    // Existing path → local install.
+                    host.install(&source)?;
+                    println!(
+                        "{}",
+                        ta(
+                            "cli-plugin-installed-from",
+                            &[("source", &source)],
+                            "Plugin installed"
+                        )
+                    );
+                    Ok(())
+                } else {
+                    // Otherwise treat it as a registry name (`name` or `name@version`).
+                    plugin_registry::install_from_registry(&source, registry.as_deref(), &mut host)
+                        .await
+                }
             }
             PluginCommands::Remove { name } => {
                 let mut host = zeroclaw::plugins::host::PluginHost::new(&config.data_dir)?;
--- a/src/plugin_registry.rs
+++ b/src/plugin_registry.rs
@@ -1,0 +1,402 @@
+//! Remote plugin registry: resolve a plugin name to a downloadable artifact,
+//! verify it, and hand it to [`PluginHost::install`] for signature-checked
+//! installation.
+//!
+//! This is the network half of `zeroclaw plugin install`. The local half
+//! (install from a directory that already exists on disk) is handled directly
+//! by `PluginHost::install`. When `install <source>` is given something that is
+//! NOT an existing path, it is treated as a registry name (optionally
+//! `name@version`) and resolved here.
+//!
+//! ## Registry format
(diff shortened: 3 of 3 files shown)


WareWolf-MoonWall (changes requested):
### 🟢 What looks good — Strong foundation for plugin ecosystem growth

This is a high-leverage infrastructure PR that directly unblocks plugin adoption. The implementation demonstrates excellent engineering discipline:

**Architecture:**
- Defense-in-depth security (SHA-256 transport + Ed25519 manifest signature + size cap + traversal protection)
- Zero duplicate state: registry URL resolution follows precedence (flag > env > default) with a single source of truth
- Registry operations cleanly separated in their own module (`plugin_registry.rs`)
- Install path auto-detection (local path vs. registry name) preserves backward compatibility while adding new capability

**Implementation quality:**
- Path-traversal protection correctly rejects `..`, leading `/`, and leading `\` in zip entries
- Size ceiling (50 MiB) enforced at both `Content-Length` check and actual bytes read (defense-in-depth)
- `find_manifest_dir` tolerates both root-level and single-subdir layouts (common packaging variations)
- SHA-256 verification accepts both bare hex and `sha256:` prefix (flexible but still strict)
- Version resolution: unpinned takes newest-listed (rfind), pinned requires exact match with clear error messages

**Test coverage:**
8 unit tests cover the critical paths:
- Version parsing/pinning logic
- Registry URL precedence
- SHA-256 verification (both valid and invalid)
- Search filtering (name + description)
- Manifest directory discovery (root and subdir variants)

**Documentation:**
Plugin-protocol guide updated with:
- Clear command examples (`search`, `install <name>`, `install <name@version>`)
- Registry format specification (JSON schema, field descriptions)
- Self-hosting instructions (override with `--registry` or env var)
- Security properties explained (transport integrity + authenticity)

**End-to-end validation evidence:**
Real binary test against local registry serving `ace-step` zip shows the full cycle works (search → install → list).

---

### 🔴 Blocking — Test failure must be resolved before merge

CI shows `Quality Gate/Test` failing (1 failing check). Per the PR review protocol, a PR cannot be approved with failing tests. The failure must be root-caused and fixed.

**Action required:** Investigate the test failure, fix the root cause, and push an update. Once CI is green, re-request review.

---

### 🟡 Warning — Install builds host with security, but search does not

Lines in `src/main.rs` (around line 5223):

```rust
PluginCommands::Install { source, registry } => {
    let mode = zeroclaw::plugins::host::PluginHost::parse_signature_mode(...);
    let trusted = config.plugins.security.trusted_publisher_keys.clone();
    let mut host = zeroclaw::plugins::host::PluginHost::with_security(...);
    // ... install ...
}
```

This is correct for `install` — it honors the configured signature policy. However, `search` does not need a `PluginHost` at all (it only fetches and parses the JSON index). This asymmetry is fine but worth noting: `search` is unauthenticated discovery; `install` is the gated operation. The PR body documents this, so it's clear the author understands the boundary.

**Not blocking,** but reviewers should confirm: is there a policy reason why `search` should *also* require a valid signature configuration, even though it doesn't install anything? If not, the current design is correct.

---

### 🟡 Warning — Registry URL default points to a repo that does not exist yet

The default registry URL is:

```rust
const DEFAULT_REGISTRY_URL: &str =
    "https://raw.githubusercontent.com/zeroclaw-labs/zeroclaw-plugins/main/registry.json";
```

The PR body states:

> **Follow-up (not in this PR):** Publish/populate the canonical `zeroclaw-labs/zeroclaw-plugins` registry repo (the default URL). The feature works against any host today via `--registry`/env, so this is decoupled.

This is a reasonable staging approach: land the CLI machinery first, populate the registry second. However, until the registry repo is populated, the default URL will 404. Users who run `zeroclaw plugin search` or `zeroclaw plugin install <name>` without `--registry` will see a fetch error.

**Not blocking** given the staging rationale, but:
1. The follow-up PR to populate `zeroclaw-labs/zeroclaw-plugins` should land before this goes into a stable release (v0.8.3 milestone is appropriate).
2. Consider adding a tracking issue reference in the source comment at `DEFAULT_REGISTRY_URL` so it's clear what depends on what.

---

### 🔵 Suggestion — Add user-friendly error for 404 on default registry

Currently, if the default registry URL returns 404 (which it will until the `zeroclaw-plugins` repo is populated), users see:

```
registry returned HTTP 404 for https://raw.githubusercontent.com/.../registry.json
```

This is technically correct but not actionable for a user who doesn't know the registry isn't populated yet. Consider adding a special case in `fetch_index` for 404 on the default URL with a message like:

```
The public plugin registry is not yet populated. Use --registry <url> to point to a custom registry, or wait for the official registry to launch.
```

**Not blocking** — the current error is not wrong, just not as helpful as it could be.

---

### 🔵 Suggestion — Document the registry repo structure in AGENTS.md or docs/

The PR adds excellent documentation for *using* the registry, but there's no documentation for *populating* it (what goes in the `zeroclaw-labs/zeroclaw-plugins` repo, how to generate the `registry.json`, how to host plugin zips, etc.). This is likely intentional for this PR (since the follow-up work is decoupled), but when that follow-up lands, the docs should include:

- How to add a plugin to the official registry (PR process, schema validation, signing requirements)
- How to self-host a private registry (directory layout, serving the JSON + zips)
- Automation/tooling for generating `registry.json` from a directory of plugin zips

**Not blocking** — but track this as part of the registry-population follow-up.

---

### 🔵 Suggestion — Consider adding `zeroclaw plugin update <name>` in a follow-up

With the registry in place, the natural next step is `zeroclaw plugin update <name>` (or `update --all`) to upgrade installed plugins to their newest registry versions. This PR lays the foundation for that (version resolution, download, install), but doesn't implement the update command itself.

**Not in this PR's scope,** but worth noting as a natural follow-up now that the registry machinery exists.

---

### 🔵 Suggestion — Truncate large error bodies safely

In `plugin_registry.rs`, network errors from `reqwest` can include response bodies in the error context. If a malicious or misconfigured server returns a huge error page, it could be rendered into the terminal. Consider truncating error messages from network operations to a reasonable length (e.g., first 500 chars) to avoid terminal spam.

**Low priority** — not a security issue (it's only displayed to the operator's own terminal), but a UX polish item.

---

## Summary

**Verdict:** **CHANGES REQUESTED** — CI test failure must be resolved.

This PR delivers critical infrastructure for plugin discoverability and removes the adoption bottleneck. The architecture is clean, the security boundaries are correct, and the implementation quality is high. The only blocker is the failing test.

**Once the test failure is fixed and CI is green, this PR will be ready to merge.** The 🟡 warnings and 🔵 suggestions are not blockers — they're either acceptable staging tradeoffs (default registry URL) or nice-to-have follow-ups.

**Next steps:**
1. @theonlyhennygod: Investigate and fix the `Quality Gate/Test` failure
2. Push the fix and re-request review
3. After this lands: populate `zeroclaw-labs/zeroclaw-plugins` registry repo (track with issue reference)

**Risk tier:** Correctly labeled `core` (new CLI surface, registry fetch, zip extraction, install path changes). Rollback is clean (revert the commit, or set `ZEROCLAW_PLUGIN_REGISTRY_URL` to a different host if the default registry misbehaves).

---

Reviewed by @WareWolf-MoonWall · 2026-06-06  
Head commit: (CI pending, exact SHA not in `gh pr view` output)

Audacity88 (changes requested):
Review context: I checked current head a737694 against the PR body, WareWolf's existing review, current failing CI status, and the `src/plugin_registry.rs` / CLI diff. I did not rerun local cargo checks for this pass.

### 🟢 What looks good — Registry installs go through the secured host path

I agree with the direction: registry-backed plugin discovery/install is a high-leverage feature, and it is good that registry installs now go through the configured `PluginHost::with_security` path instead of silently disabling signature policy. The existing version-resolution, SHA-256 verification, and manifest-dir tests are a solid start. CI is already red on this PR; I am focusing this review on the additional code-safety blockers I see in the registry download/extraction path.

### 🔴 Blocking — The 50 MiB archive cap is checked after buffering unknown-length bodies

`download_zip()` checks `Content-Length` when the server sends it, but then calls `resp.bytes().await` and only checks `bytes.len()` after the entire response body has already been read into memory:

```rust
let bytes = resp
    .bytes()
    .await
    .context("reading plugin archive")?
    .to_vec();
if bytes.len() as u64 > MAX_PLUGIN_ZIP_BYTES {
    bail!(...);
}
```

That does not enforce the cap against a registry or artifact server that omits or lies about `Content-Length`; the oversized body has already been buffered before ZeroClaw can reject it. Since this code downloads registry-selected plugin archives from arbitrary hosts, the cap needs to be enforced while reading the body, not just before and after. Please stream/chunk the response and stop as soon as the accumulated bytes exceed `MAX_PLUGIN_ZIP_BYTES`, with a regression test for an oversized response without `Content-Length`.

### 🔴 Blocking — Zip extraction should use normalized enclosed paths before writing

`extract_zip_safe()` currently treats zip entry names as strings and rejects only names containing `..` or starting with `/` or `\` before doing `dest.join(&name)`. That is not the right primitive for registry-supplied archives, especially because this CLI is cross-platform. For example, a Windows drive-prefixed entry such as `C:\...` or `C:/...` is not caught by those string checks, but is an absolute path on Windows; joining it can escape `dest`.

Please switch this to the zip crate's enclosed/sanitized path API, or otherwise derive a normalized relative path and verify that the final path stays under `dest` before creating files. Please add unsafe-archive tests that cover `../`, absolute Unix paths, absolute Windows drive paths, and nested valid paths.

singlerider:

PR #7334, opened 2026-06-06
fix(channels/telegram): clamp zero draft update interval

### LINKED ISSUE #7332: [Bug]: Telegram partial streaming accepts zero draft update interval and floods edits
### Affected component

channel

### Severity

S2 - degraded behavior

### Current behavior

Telegram partial streaming preserves `draft_update_interval_ms = 0` as a literal edit interval.

On current `master`, `TelegramChannel::with_streaming` assigns the configured value directly:

```rust
self.stream_mode = stream_mode;
self.draft_update_interval_ms = draft_update_interval_ms;
```

Then `update_draft` suppresses edits only when the elapsed time is lower than the interval:

```rust
if elapsed < self.draft_update_interval_ms {
    return Ok(());
}
```

With an interval of `0`, that condition is never true, so every streamed delta can attempt an `editMessageText` call. This defeats the documented edit throttle and can make Telegram partial streaming visibly slow or hit Telegram edit rate limits.

### Expected behavior

Telegram partial draft edits should remain throttled even when the configured interval is `0`.

A zero interval should fall back to the existing safe default, or otherwise be clamped to a non-zero minimum, so the channel does not accidentally flood Telegram `editMessageText` updates.

### Steps to reproduce

```bash
# On current master / v0.8.0-beta-2:
# 1. Configure Telegram partial streaming with a zero edit interval.

[channels.telegram]
enabled = true
bot_token = "REDACTED"
allowed_users = [123456789]
stream_mode = "partial"
draft_update_interval_ms = 0

# 2. Send a prompt that produces multiple streamed text deltas.
# 3. Observe that the Telegram channel attempts draft edits without any interval throttle.

# Code path on master:
# - TelegramChannel::with_streaming stores 0 directly.
# - TelegramChannel::update_draft only skips an edit when elapsed < draft_update_interval_ms.
# - elapsed < 0 is never true for u64, so the throttle is effectively disabled.
```

### Impact

Affected users: Telegram channel operators who enable `stream_mode = "partial"` and set `draft_update_interval_ms = 0` expecting faster updates.

Frequency: Always for that configuration.

Consequence: The channel can issue one Telegram draft edit per streamed delta, causing slow visible streaming, unnecessary API pressure, and possible Telegram `Too Many Requests` responses. This contradicts the docs, which say draft edit cadence is bounded by `draft_update_interval_ms` to avoid rate-limiting.

### Logs / stack traces

```text
No production identifiers included.

Relevant docs/code evidence from current master:

docs/book/src/channels/chat-others.md:
Streaming draft edits are supported but capped by Telegram's rate limit. Tune `draft_update_interval_ms` if you see "Too Many Requests".

docs/book/src/providers/streaming.md:
The edit cadence is bounded by `draft_update_interval_ms` in the channel config (default: 500 ms) to avoid rate-limiting.

crates/zeroclaw-channels/src/telegram.rs:
self.draft_update_interval_ms = draft_update_interval_ms;

if elapsed < self.draft_update_interval_ms {
    return Ok(());
}
```

### ZeroClaw version

0.8.0-beta-2 / upstream master `eac987668`

### Rust version

### Operating system

macOS 15.7.7 (Darwin 24.6.0 arm64)

### Regression?

Unknown

### Pre-flight checks

rifuki:
## Summary

Closes #7332.

- Keep Telegram partial-streaming draft edits throttled when `draft_update_interval_ms` is configured as `0`.
- Reuse the existing Telegram draft edit default instead of storing a literal zero interval.
- Add a regression test covering the zero-interval fallback while preserving non-zero values unchanged.

## Validation Evidence

```text
$ cargo fmt --all -- --check
(exit 0)
```

```text
$ cargo test -p zeroclaw-channels with_streaming_uses_default_for_zero_draft_update_interval --lib
running 1 test
test telegram::tests::with_streaming_uses_default_for_zero_draft_update_interval ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 688 filtered out; finished in 0.17s
```

```text
$ cargo test -p zeroclaw-channels update_draft --lib
running 2 tests
test telegram::tests::update_draft_utf8_truncation_is_safe_for_multibyte_text ... ok
test telegram::tests::update_draft_rate_limit_short_circuits_network ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 687 filtered out; finished in 0.21s
```

```text
Finished `dev` profile [unoptimized + debuginfo] target(s) in 21.06s
```

```text
$ cargo test
...
test result: ok. 243 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.40s
test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.79s
test result: ok. 159 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.26s
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
Doc-tests zeroclaw: ok
```

Validated after rebasing onto upstream `master` at `0f9442f44`.

## Security & Privacy Impact

No security boundary or privacy behavior changes. This does not add logging, persistence, identifiers, tokens, or external calls.

## Compatibility

Non-zero `draft_update_interval_ms` values keep their configured behavior. A zero interval now falls back to the Telegram draft edit default instead of disabling edit throttling, which aligns the channel with the documented rate-limit guard.

## Rollback

Revert this commit to restore direct use of the configured interval.

## Supersede Attribution

N/A. This is a narrow follow-up for #7332; duplicate searches for Telegram draft interval/edit throttling did not find an existing issue or PR covering this zero-interval path.

### DIFF
--- a/crates/zeroclaw-channels/src/telegram.rs
+++ b/crates/zeroclaw-channels/src/telegram.rs
@@ -360,6 +360,9 @@ fn parse_attachment_markers(message: &str) -> (String, Vec<TelegramAttachment>) 
 /// Telegram Bot API maximum file download size (20 MB).
 const TELEGRAM_MAX_FILE_DOWNLOAD_BYTES: u64 = 20 * 1024 * 1024;
 
+/// Default minimum interval between Telegram draft edits.
+const TELEGRAM_DRAFT_UPDATE_INTERVAL_MS: u64 = 1000;
+
 /// Telegram channel — long-polls the Bot API for updates
 pub struct TelegramChannel {
     bot_token: String,
@@ -453,7 +456,7 @@ impl TelegramChannel {
             pairing,
             client: reqwest::Client::new(),
             stream_mode: StreamMode::Off,
-            draft_update_interval_ms: 1000,
+            draft_update_interval_ms: TELEGRAM_DRAFT_UPDATE_INTERVAL_MS,
             last_draft_edit: Mutex::new(std::collections::HashMap::new()),
             typing_handle: Mutex::new(None),
             mention_only,
@@ -512,7 +515,11 @@ impl TelegramChannel {
         draft_update_interval_ms: u64,
     ) -> Self {
         self.stream_mode = stream_mode;
-        self.draft_update_interval_ms = draft_update_interval_ms;
+        self.draft_update_interval_ms = if draft_update_interval_ms == 0 {
+            TELEGRAM_DRAFT_UPDATE_INTERVAL_MS
+        } else {
+            draft_update_interval_ms
+        };
         self
     }
 
@@ -4189,6 +4196,22 @@ mod tests {
         .with_streaming(StreamMode::Partial, 750);
         assert!(partial.supports_draft_updates());
         assert_eq!(partial.draft_update_interval_ms, 750);
+    }
+
+    #[test]
+    fn with_streaming_uses_default_for_zero_draft_update_interval() {
+        let ch = TelegramChannel::new(
+            "fake-token".into(),
+            "telegram_test_alias",
+            Arc::new(|| vec!["*".into()]),
+            false,
+        )
+        .with_streaming(StreamMode::Partial, 0);
+
+        assert_eq!(
+            ch.draft_update_interval_ms,
+            TELEGRAM_DRAFT_UPDATE_INTERVAL_MS
+        );
     }
 
     #[tokio::test]


Audacity88 (approved):
I reviewed head `8d997108d19d04dffeb9ec238403aff2750487ff` against the live PR state, linked issue #7332, the `crates/zeroclaw-channels/src/telegram.rs` diff, and the public PR-body validation evidence. I did not rerun local cargo. At the latest live read, all visible GitHub checks were green, including CI Required Gate.

### 🟢 What looks good — zero interval now keeps the Telegram throttle alive

The linked issue and PR body describe the same failure mode the diff fixes: storing `draft_update_interval_ms = 0` made the Telegram draft-edit throttle ineffective. This patch keeps non-zero intervals unchanged, but maps `0` back to the existing Telegram draft-edit default. That preserves the rate-limit guard instead of treating `0` as an implicit disable switch.

### 🟢 What looks good — the regression test covers the boundary

The new `with_streaming_uses_default_for_zero_draft_update_interval` test pins the exact case that regressed, while the existing `with_streaming_preserves_custom_interval` assertion with `750` still proves configured non-zero values are preserved. The nearby draft-update tests continue to cover rate-limit short-circuiting and UTF-8-safe draft truncation behavior, so this is a nice narrow fix.

### 🟢 What looks good — validation covers the narrow risk

The PR body lists focused Telegram tests, broader `cargo test`, clippy, and fmt as passing locally, and the patch is small enough that I did not see a reason to block on more local validation from review. GitHub checks are also green at the reviewed head.

PR #7335, opened 2026-06-06
feat(plugins): sandbox isolation — resource limits, SSRF egress guard, env scoping

theonlyhennygod:
## Summary

ZeroClaw's plugins run in a **WASM sandbox** — but the host imposed **no resource, network, or env bounds**, so the sandbox wasn't actually containing anything. This is the robustness work the plugin system needs *and* a hard-to-copy edge: competitors extend agents with **MCP servers = full OS subprocesses** (whole env + network); a properly contained WASM plugin is something they structurally can't offer.

**Backward-compat is the hard constraint:** the ~33 existing plugins (declare `http_client`/`env_read`, read one API key, POST to a public API) keep working unchanged. Always-on protections are safe for them; tighter controls are opt-in.

## Always-on (safe for existing plugins)

- **Resource limits** in `create_plugin()`: **256 MiB** memory cap + **180 s** Extism epoch timeout (must exceed the 120 s outbound HTTP ceiling). A `tokio::time::timeout` backstop in `WasmTool::execute` frees the async caller; Extism's epoch timeout is the real interruptor for runaway wasm (tokio can't kill the blocking thread — documented).
- **SSRF egress guard** in `zc_http_request`: reject non-`http(s)` schemes; **resolve the host and block** loopback / private (10/8, 172.16/12, 192.168/16) / CGNAT / link-local incl. cloud-metadata `169.254.169.254` / IPv4-mapped-v6 — checked against the *resolved* IPs (handles naive DNS rebinding). The `reqwest`-re-resolve TOCTOU window is noted as a follow-up (connection pinning).

## Opt-in (absent = prior behavior)

- **`allowed_hosts`** manifest field — restrict egress to exact or `*.`-wildcard hosts (`*.fal.run` matches `fal.run` + `queue.fal.run`).
- **`env_allowlist`** manifest field — restrict `zc_env_read` to named vars. Absent ⇒ permit with a **one-time per-var warning** (a wasm module's key reads can't be derived statically, so silent-deny would break existing plugins). An `enforce_env_allowlist` operator flag can come later.

## Implementation
- `HostContext` carries the allowlists; `create_plugin_with()` is the scoped constructor (`create_plugin()` delegates with `None`). Threaded through `WasmTool` → runtime registration (`tools/mod.rs`).
- New `url` dep for URL parsing; IP classification via `std::net` predicates + manual CGNAT/ULA/link-local bit checks (no new heavy dep).

## Validation
- ✅ **47 unit tests** (egress block loopback/metadata/private/IPv4-mapped/CGNAT, public-IP allow, scheme reject, wildcard host-match, env allowlist, compile-time timeout invariants)
- ✅ `cargo clippy -p zeroclaw-plugins --all-targets -- -D warnings` clean · `cargo fmt --check` clean
- ✅ `zeroclaw-runtime` builds with `--features plugins-wasm` (registration call site)
- ✅ Existing integration tests (image-gen-fal load/execute) unaffected; existing plugins (no allowlists) hit public APIs → permitted

## Docs
Added a **Sandboxing** section to the plugin-protocol guide (limits, egress guard, the two opt-in fields).

Part of the plugin sandbox-hardening series. Follow-ups (separate PRs): tool namespacing + `RateLimitedTool` wrapping; signing posture (default Permissive) + surfaced signature status.

Milestone: **v0.8.3**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13230,6 +13230,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "toml 1.1.2+spec-1.1.0",
+ "url",
  "zeroclaw-api",
  "zeroclaw-log",
 ]
--- a/crates/zeroclaw-plugins/Cargo.toml
+++ b/crates/zeroclaw-plugins/Cargo.toml
@@ -20,6 +20,7 @@ reqwest = { version = "0.12", default-features = false, features = ["rustls-tls"
 thiserror = "2.0"
 tokio = { version = "1.50", default-features = false, features = ["sync", "macros", "rt"] }
 toml = "1.0"
+url = "2"
 
 [dev-dependencies]
 tempfile = "3.26"
--- a/crates/zeroclaw-plugins/src/lib.rs
+++ b/crates/zeroclaw-plugins/src/lib.rs
@@ -41,6 +41,18 @@ pub struct PluginManifest {
     /// Hex-encoded Ed25519 public key of the publisher who signed this manifest.
     #[serde(default)]
     pub publisher_key: Option<String>,
+    /// Optional egress allowlist. When present, the plugin's HTTP requests are
+    /// restricted to these hosts (exact, or a leading-`*.` wildcard such as
+    /// `*.fal.run`). When absent, the always-on SSRF guard still blocks
+    /// loopback/private/cloud-metadata addresses but any public host is
+    /// reachable — backward-compatible with plugins that predate this field.
+    #[serde(default)]
+    pub allowed_hosts: Option<Vec<String>>,
+    /// Optional environment-variable allowlist. When present, `zc_env_read` may
+    /// only read these names. When absent, reads are permitted (with a one-time
+    /// warning) for backward compatibility.
+    #[serde(default)]
+    pub env_allowlist: Option<Vec<String>>,
 }
 
 /// What a plugin can do.
--- a/crates/zeroclaw-plugins/src/runtime.rs
+++ b/crates/zeroclaw-plugins/src/runtime.rs
@@ -9,15 +9,38 @@ use anyhow::{Context, Result};
 use extism::*;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, ToSocketAddrs};
 use std::path::Path;
 use zeroclaw_api::tool::ToolResult;
 
+/// Maximum linear memory for a plugin instance: 4096 WebAssembly pages ×
+/// 64 KiB = 256 MiB. Generous for the API-wrapper plugins we ship (which
+/// encode one key + a small body and read a JSON response) while bounding a
+/// runaway allocation so a buggy/malicious plugin can't OOM the host.
+const MAX_WASM_PAGES: u32 = 4096;
+
+/// Extism epoch-interruption timeout for a single plugin call, in milliseconds.
+/// Must exceed the 120 s outbound HTTP ceiling in [`handle_http_request`] so a
+/// legitimate slow call (e.g. cold model inference) is not trapped mid-flight.
+/// Epoch interruption only fires at wasm instruction boundaries, so this bounds
+/// runaway *wasm* (tight loops); a stall inside the host's blocking HTTP call is
+/// bounded by the reqwest client timeout instead.
+pub(crate) const EXEC_TIMEOUT_MS: u64 = 180_000;
+
+/// Compile-time invariant: the epoch timeout must outlast the 120 s HTTP ceiling.
+const _: () = assert!(EXEC_TIMEOUT_MS > 120_000);
+
 // ── Host function context ─────────────────────────────────────────
 
-/// Permissions available to a single plugin invocation.
+/// Permissions and scoping available to a single plugin invocation.
 #[derive(Debug, Clone)]
 struct HostContext {
     permissions: HashSet<PluginPermission>,
+    /// Optional per-plugin egress allowlist. `None` = no host restriction
+    /// (the always-on SSRF guard still applies).
+    allowed_hosts: Option<Vec<String>>,
+    /// Optional per-plugin env-var allowlist. `None` = permissive (back-compat).
+    env_allowlist: Option<Vec<String>>,
 }
 
 // ── Data types exchanged with plugins ─────────────────────────────
@@ -81,6 +104,10 @@ fn handle_http_request(
 
     let req: HttpRequest = serde_json::from_str(&request_json)
         .map_err(|e| Error::msg(format!("invalid HTTP request JSON: {e}")))?;
+
+    // SSRF guard: always block non-public destinations; additionally enforce the
+    // per-plugin host allowlist when one is configured.
+    validate_egress(&req.url, ctx.allowed_hosts.as_deref())?;
 
     // 120s ceiling covers legitimate slow cases: large file downloads and slow
     // model-inference endpoints (fal.ai image generation routinely takes 20-60s
@@ -158,6 +185,18 @@ fn handle_env_read(
 
     let var_name: String = plugin.memory_get_val(&inputs[0])?;
 
+    // Per-plugin env scoping. When an allowlist is declared, deny anything
+    // outside it; when absent, permit (back-compat) but warn once per var so
+    // operators can author an allowlist.
+    if !env_var_allowed(ctx.env_allowlist.as_deref(), &var_name) {
+        return Err(Error::msg(format!(
+            "env access denied: '{var_name}' is not in the plugin's env_allowlist"
+        )));
+    }
+    if ctx.env_allowlist.is_none() {
+        warn_unscoped_env_read(&var_name);
+    }
+
     let value = std::env::var(&var_name)
         .map_err(|_| Error::msg(format!("environment variable '{var_name}' not set")))?;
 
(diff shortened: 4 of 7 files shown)


WareWolf-MoonWall (changes requested):
## Review of PR #7335 — Plugin sandbox isolation

**Author:** @theonlyhennygod  
**Risk tier:** HIGH (security boundaries, sandbox enforcement)  
**Head commit:** (will be captured at posting)

This PR introduces resource limits, SSRF egress filtering, and environment scoping for the WASM plugin sandbox. The work is architecturally critical — it transforms the plugin system from "WASM isolation theater" (no resource or network bounds) into defensible containment. The backward-compatibility constraint (33 existing plugins continue working) is well-respected through opt-in controls and permissive-with-warnings defaults.

---

### 🟢 What looks good

**The SSRF guard is comprehensive and well-tested.** `validate_egress` blocks loopback, RFC 1918 private ranges, CGNAT (100.64/10), link-local (including the cloud metadata endpoint 169.254.169.254), and IPv4-mapped-v6 addresses. The DNS resolution step defends against naive rebinding attacks (IP literals alone aren't enough). The 47-test unit suite covers every blocked range, public-IP allow, wildcard host matching, and scheme rejection — exactly the depth a security boundary needs.

**Resource limits are well-chosen.** 256 MiB (4096 WASM pages) is generous for API-wrapper plugins (one key + small JSON body) while preventing OOM. The 180 s epoch timeout exceeds the 120 s HTTP ceiling (documented with compile-time assertions), so legitimate slow inference calls won't be interrupted mid-flight. The two-level timeout design (Extism epoch + tokio backstop) correctly recognizes that `tokio::time::timeout` can't kill the blocking thread — only Extism's epoch interrupt handles runaway wasm loops. The wall-clock backstop exists solely to free the async caller when a plugin wedges inside a *host* call.

**Backward compatibility is real.** The `allowed_hosts` and `env_allowlist` fields are `Option<Vec<String>>` in both the manifest schema and the runtime `HostContext`, with `None` preserving prior behavior. The one-time-per-var warning for unscoped env reads guides operators toward authoring allowlists without breaking existing plugins at load time. The always-on SSRF guard is safe because it only blocks destinations no legitimate plugin should reach (loopback/private/metadata).

**The TOCTOU window is documented.** The PR notes that `reqwest` re-resolves DNS on connect after we've validated the addresses, leaving a small time-of-check-to-time-of-use race. The comment explicitly names "connection pinning" as the full closure and frames it as a follow-up — transparent about the gap, not hiding it.

**Test coverage matches the risk.** 47 unit tests exercise every security control: blocked ranges (v4/v6/mapped), public allow, wildcard host match, scheme rejection, env allowlist logic, and timeout invariants. Existing integration tests confirm the image-gen-fal plugin (which reads `FAL_API_KEY` and POSTs to fal.run) still works. The validation evidence section in the PR body lists all checks (clippy clean, builds with `--features plugins-wasm`, integration tests pass) — complete and verifiable.

**Docs additions are accurate.** The new "Sandboxing" section in `plugin-protocol.md` explains always-on limits (256 MiB, 180 s, SSRF guard) and opt-in controls (`allowed_hosts`, `env_allowlist`) with correct examples. The manifest schema examples show the new fields as optional, matching the runtime implementation.

---

### 🔴 Blocking — duplicate state violation (AGENTS.md absolute rule)

`WasmTool` now holds `allowed_hosts: Option<Vec<String>>` and `env_allowlist: Option<Vec<String>>` fields cloned from `PluginManifest` at construction time. The single source of truth for these values is the manifest TOML (loaded into `PluginManifest` by the plugin discovery subsystem); copying them into `WasmTool` creates duplicate state that will drift if the manifest is reloaded or mutated.

**Why this matters:** AGENTS.md §"ABSOLUTE RULE" bans duplicate state because it breeds drift bugs whose symptoms surface months later — operator edits the canonical location, the cached copy serves stale data, the agent silently misbehaves. The previous channel `allowed_users` Vec drift (cached in channel handles while the truth lived in config TOML) is the exact same pattern — reloading config didn't refresh the channels; an authorized user couldn't talk to the bot until daemon restart.

**Required fix:** Store a reference to the canonical source (the loaded `PluginManifest`) or pass `&[String]` parameters at the call sites that need them. Two valid patterns:

1. **Thread through call sites:** Pass `allowed_hosts: Option<&[String]>` and `env_allowlist: Option<&[String]>` as parameters to `WasmTool::execute`. The caller (`all_tools_with_runtime`) has the manifest; thread the slices through.

2. **Store a handle to the manifest:** Change `WasmTool` to hold an `Arc<PluginManifest>` instead of cloning the Vecs. The manifest is already loaded and lives somewhere; `WasmTool` should borrow a handle, not copy fields.

Both approaches satisfy AGENTS.md §"Pre-edit ritual" answer #2: "Source of truth is `PluginManifest` — this would be a duplicate." The current pattern (clone Vecs into `WasmTool` at construction) is answer #3 ("snapshot is fine") — explicitly forbidden.

**Severity:** This is a hard block per AGENTS.md. The forcing mechanism is "automatic-revert-on-detect" if it lands. Fix before merge.

---

### 🟡 Warning — `unwrap_or(false)` in `warn_unscoped_env_read` silently suppresses lock-poison errors

```rust
let is_new = warned
    .lock()
    .map(|mut s| s.insert(var_name.to_string()))
    .unwrap_or(false);
```

If the mutex is poisoned (a prior thread panicked while holding the lock), this returns `false` and skips the warning. Per FND-006 §4.1 (Error Handling as a Design Concern), a lock-poison error is a **programmer error** (invariant violated — another thread panicked in a critical section), not an operational failure. The correct response is `.expect("warn_unscoped_env_read mutex poisoned")` with a message explaining why this should never happen in correct code, not silent suppression with `unwrap_or(false)`.

**Why it matters:** Lock poisoning means the data structure's invariants may be violated (the panicking thread left it in an inconsistent state). Continuing as if nothing happened (skipping the warning) hides evidence that something serious went wrong elsewhere. FND-006 §4.1: "The difference between `.unwrap()` and `.expect('reason this is safe')` is not just style. It is the difference between deferred judgment and documented judgment."

**Suggested fix:**
```rust
let is_new = warned
    .lock()
    .expect("warn_unscoped_env_read mutex poisoned — another thread panicked in this critical section")
    .insert(var_name.to_string());
```

If the lock is poisoned, the runtime should panic visibly (developer will see it in logs/tests), not silently skip warnings and continue.

---

### 🔵 Suggestion — consider explicit test for the TOCTOU window acknowledgment

The PR documents the `reqwest`-re-resolve TOCTOU window as a follow-up (connection pinning). Consider adding a test comment or a `#[ignore]`-gated test that demonstrates the gap — e.g., a mock DNS server that returns public IP on first resolve, then private IP on second resolve, showing that `reqwest` would use the unvalidated address. This makes the gap mechanically reproducible (not just documented) and gives the follow-up PR a concrete target test to flip from `#[ignore]` to passing.

Not blocking — the comment is already transparent about the gap, and the code is otherwise correct. This just makes the gap testable.

---

### 🔵 Suggestion — timeout constant names could be more descriptive

`EXEC_TIMEOUT_MS` and `WALL_CLOCK_TIMEOUT_MS` are clear in context, but the asymmetry (one is "exec", the other is "wall-clock") makes it non-obvious which is the backstop and which is the primary interrupt. Consider:

- `WASM_EPOCH_TIMEOUT_MS` (Extism's epoch interrupt — primary)
- `ASYNC_BACKSTOP_TIMEOUT_MS` (tokio timeout — frees caller when host call wedges)

Not blocking — the existing names work, and the comments explain the relationship. This just surfaces the *mechanism* (epoch vs async backstop) in the name.

---

### 🔵 Suggestion — `host_matches` could reject patterns like `*example.com` (missing dot)

The wildcard logic requires a leading `*.` — correct per the PR description — but silently treats `*example.com` (no dot) as an exact match against the literal string `*example.com` (which no host will match). Consider explicitly rejecting patterns that start with `*` but lack the dot, so operator typos surface as errors at manifest-load time rather than silent no-match at runtime.

Not blocking — the existing behavior is safe (typo → no match → plugin can't reach the intended host, operator notices quickly). This just fails faster.

---

### ✅ Manifest schema extension follows existing patterns

The `allowed_hosts` and `env_allowlist` fields in `PluginManifest` use `#[serde(default)]` and `Option<Vec<String>>`, matching the style of `publisher_key`, `signature`, and other optional fields. The serialization round-trip is consistent with the rest of the manifest schema.

---

### ✅ No `unwrap()` in production paths

Every `unwrap()` or `expect()` in the diff is either in a test (`assert!(result.is_err())`) or preceded by a comment explaining the invariant (e.g., `.port_or_known_default().unwrap_or(443)` is documented — HTTP/HTTPS always have a default port). The operational error paths (DNS resolution failure, URL parse failure, blocked IP) all propagate `Result` with `?` or return `Err` — no panic-on-network-failure.

---

## Verdict reasoning

The SSRF guard, resource limits, and test coverage are production-quality. The TOCTOU acknowledgment is transparent. The backward-compatibility design is thoughtful.

The **duplicate state violation** (cloning manifest allowlists into `WasmTool` fields) is a hard block per AGENTS.md. The rule exists because this exact pattern (cached Vecs that drift from the canonical source) has burned the project before. Fix the data flow so `WasmTool` resolves allowlists from the manifest on demand (thread parameters or store an `Arc<PluginManifest>`), then the rest of the PR is solid.

The **lock-poison `unwrap_or(false)`** is a warning-tier issue (wrong error-handling discipline per FND-006 §4.1, but not a security hole). Fix if feasible; document if not.

**Verdict:** `--request-changes` on the duplicate state violation. Address that, and this is `--approve` with the warnings noted for follow-up.

singlerider:

PR #7340, opened 2026-06-07
refactor(zeroclaw-tools): extract duplicate domain/URL validation int…

jokemanfire:
## Summary

- **What changed and why:**
  - Fixed IPv6 bracket validation: `[::1`, `::1]`, `[127.0.0.1` no longer silently accepted — only paired `[...]` brackets are stripped, unmatched brackets are rejected (maintainer requirement).
  - Extracted 6 duplicated domain/URL security functions from 4 tool files (`normalize_domain`, `normalize_allowed_domains`, `host_matches_allowlist`, `is_private_or_local_host`, `is_non_global_v4`, `is_non_global_v6`) into a shared `helpers/domain_guard` module, eliminating **986 lines of duplicate code**.
  - `host_matches_allowlist` now uniformly supports `*`, `*.domain` wildcards, and exact-only IP matching across all 4 tools (previously each tool had a subtly different implementation).
  - `is_private_or_local_host` now uses `std::net::IpAddr` parsing for all tools — `browser_open.rs`'s hand-rolled `parse_ipv4` (with no IPv6 support beyond hardcoded `::1`) is removed.
  - Fixed hanging `ipv6_end_to_end_real_request_over_loopback` E2E test with `tokio::time::timeout` guard.
- **Scope boundary:**
  - Does NOT unify `extract_host` across tools — each tool has different URL scheme support (https-only, http+https, http+https+file) and different log message prefixes.
  - Does NOT touch `webhook_audit.rs` or `skills/mod.rs` bracket-stripping — out of scope for this change.
  - Does NOT change the `domain_matcher.rs` in `zeroclaw-config`.
- **Blast radius:** All network-capable tools (`browser`, `browser_open`, `http_request`, `web_fetch`). DNS-resolve validation in `web_fetch` now uses `domain_guard::is_non_global_v4/v6` (pub(crate)). Any new tool can `use crate::helpers::domain_guard` to get the full validation suite.
- **Linked issue(s):** None
- **Labels:** `risk: high`, `size: L`, `tool`, `tool:browser`, `tool:web`

## Validation Evidence

```bash
cargo fmt --all -- --check    # clean, no diffs
cargo check -p zeroclaw-tools # succeeds, zero errors
cargo test -p zeroclaw-tools --lib
```

  ```
  test result: ok. 1258 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.30s
  ```
  Current GitHub CI is green at head 853c4ca, including Format, Lint, Test, Security, and CI Required Gate.
  - `[::1` / `::1]` / `[127.0.0.1` / `127.0.0.1]` all correctly rejected by `normalize_domain` (shared test `normalize_domain_rejects_unmatched_brackets`).
  - `[2001:db8::1]` / `::1` / `127.0.0.1` / `https://example.com/path` correctly accepted and normalized.
  - `*.example.com` wildcard matching works (new test `host_matches_allowlist_wildcard_subdomain`).
  - IP addresses only match exactly, not via suffix (new test `host_matches_allowlist_ip_exact_only`).
  - `2001:db8::/32` (documentation prefix) now correctly classified as non-global (was incorrectly treated as public in `browser.rs`).
- **If any command was intentionally skipped, why:** N/A

## Security & Privacy Impact

## Compatibility

- Backward compatible? **Yes** — all existing allowlist entries continue to work. Unmatched-bracket entries like `[::1` were never valid, and `normalize_allowed_domains` now rejects them at tool construction time (fail-fast) instead of silently accepting malformed input.

## Rollback

High-risk refactor: revert this PR. No config, env, CLI, or data migration is required because the change is confined to shared tool validation helpers and their call sites.

---

**Labels** live in the GitHub label UI. Maintainers set `risk: high`, `size: L`, `tool`, `tool:browser`, and `tool:web`.

### DIFF
--- a/crates/zeroclaw-tools/src/web_fetch.rs
+++ b/crates/zeroclaw-tools/src/web_fetch.rs
@@ -1,3 +1,4 @@
+use crate::helpers::domain_guard;
 use async_trait::async_trait;
 use futures_util::StreamExt;
 use serde_json::json;
@@ -42,15 +43,15 @@ impl WebFetchTool {
     ) -> anyhow::Result<Self> {
         Ok(Self {
             security,
-            allowed_domains: normalize_allowed_domains(
+            allowed_domains: domain_guard::normalize_allowed_domains(
                 allowed_domains,
                 "web_fetch.allowed_domains",
             )?,
-            blocked_domains: normalize_allowed_domains(
+            blocked_domains: domain_guard::normalize_allowed_domains(
                 blocked_domains,
                 "web_fetch.blocked_domains",
             )?,
-            allowed_private_hosts: normalize_allowed_domains(
+            allowed_private_hosts: domain_guard::normalize_allowed_domains(
                 allowed_private_hosts,
                 "web_fetch.allowed_private_hosts",
             )?,
@@ -536,11 +537,11 @@ fn validate_target_url_with_dns_check(
     let host = extract_host(url)?;
 
     // blocked_domains always takes precedence
-    if host_matches_allowlist(&host, blocked_domains) {
+    if domain_guard::host_matches_allowlist(&host, blocked_domains) {
         anyhow::bail!("Host '{host}' is in {tool_name}.blocked_domains");
     }
 
-    let host_is_private_or_local = is_private_or_local_host(&host);
+    let host_is_private_or_local = domain_guard::is_private_or_local_host(&host);
     let private_host_allowed =
         host_matches_private_allowlist(&host, allowed_private_hosts, host_is_private_or_local);
 
@@ -561,7 +562,12 @@ fn validate_target_url_with_dns_check(
         );
     }
 
-    if !private_host_allowed && !host_matches_allowlist(&host, allowed_domains) {
+    // Private hosts in the allowlist skip the allowed_domains check.
+    // Non-private hosts still require allowed_domains approval even when
+    // listed in allowed_private_hosts (e.g. explicit internal DNS names).
+    let skip_allowed_domains = host_is_private_or_local && private_host_allowed;
+
+    if !skip_allowed_domains && !domain_guard::host_matches_allowlist(&host, allowed_domains) {
         anyhow::bail!("Host '{host}' is not in {tool_name}.allowed_domains");
     }
 
@@ -585,68 +591,6 @@ fn append_chunk_with_cap(buffer: &mut Vec<u8>, chunk: &[u8], hard_cap: usize) ->
 
     buffer.extend_from_slice(chunk);
     buffer.len() >= hard_cap
-}
-
-fn normalize_allowed_domains(domains: Vec<String>, label: &str) -> anyhow::Result<Vec<String>> {
-    let mut rejected = Vec::new();
-    let mut normalized = domains
-        .into_iter()
-        .filter_map(|d| {
-            normalize_domain(&d).or_else(|| {
-                rejected.push(d.clone());
-                None
-            })
-        })
-        .collect::<Vec<_>>();
-    if !rejected.is_empty() {
-        anyhow::bail!(
-            "Invalid {label} entry(s): [{}]. Each entry must be a valid domain, hostname, IPv4, or IPv6 address.",
-            rejected.join(", ")
-        );
-    }
-    normalized.sort_unstable();
-    normalized.dedup();
-    Ok(normalized)
-}
-
-fn normalize_domain(raw: &str) -> Option<String> {
-    let input = raw.trim();
-    if input.is_empty() || input.chars().any(char::is_whitespace) {
-        return None;
-    }
-
-    let bare_ip = match (input.starts_with('['), input.ends_with(']')) {
-        (true, true) => &input[1..input.len() - 1],
-        (false, false) => input,
-        _ => return None,
-    };
-    if let Ok(ip) = bare_ip.parse::<std::net::IpAddr>() {
-        return Some(ip.to_string().to_lowercase());
-    }
-
-    let parsed = reqwest::Url::parse(input)
-        .or_else(|_| reqwest::Url::parse(&format!("https://{input}")))
-        .ok()?;
-
-    if !parsed.username().is_empty() || parsed.password().is_some() {
-        return None;
-    }
-
-    let host = parsed.host_str()?;
-    let trimmed = host.trim();
-    let host_no_brackets = match (trimmed.starts_with('['), trimmed.ends_with(']')) {
-        (true, true) => &trimmed[1..trimmed.len() - 1],
-        (false, false) => trimmed,
-        _ => return None,
-    };
-    let normalized = host_no_brackets
-        .trim_start_matches('.')
-        .trim_end_matches('.');
-    if normalized.is_empty() {
-        return None;
-    }
-
-    Some(normalized.to_lowercase())
 }
 
 fn extract_host(url: &str) -> anyhow::Result<String> {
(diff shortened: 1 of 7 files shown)


Audacity88 (changes requested):
I reviewed current head `1428cf2e9fe9152acace4ec66ad5ddfc34b8a80d` against `master`, including the live PR metadata, labels, green CI, prior Copilot comments, the current diff, and the old/current `web_fetch` private-host validation path. I did not run local validation.

### 🟢 What looks good — The shared helper removes real drift

This is the right general direction for these tools. Centralizing `normalize_domain`, allowlist matching, and non-global IP classification removes a lot of duplicated security logic from `browser`, `browser_open`, `http_request`, and `web_fetch`. I also see the prior unmatched-bracket and case-insensitive localhost issues handled in the shared helper, and the `allowed_private_hosts = ["*"]` path no longer lets a public host bypass the normal `allowed_domains` check.

### 🔴 Blocking — `web_fetch.allowed_private_hosts` no longer permits named internal DNS hosts

The refactor changes an existing `web_fetch` behavior for explicit private-host entries. Before this PR, `host_matches_private_allowlist()` treated `*` specially, but still honored a named `allowed_private_hosts` entry like `local.internal`; the regression test `allowed_private_domain_skips_dns_public_check` used `https://local.internal/api` and made the DNS callback panic to prove that an explicitly configured private DNS name skipped the public-host DNS check.

In this head, `host_matches_private_allowlist()` returns `false` immediately when `domain_guard::is_private_or_local_host(host)` is false. A host such as `local.internal` is not syntactically `.local`, `localhost`, or an IP address, so it falls through to DNS validation even when it is explicitly listed in `allowed_private_hosts`. If that internal DNS name resolves to `192.168.x.x`, `10.x.x.x`, etc., `web_fetch` now rejects it. The test was changed from `local.internal` to `myhost.local`, so CI no longer proves the old internal-DNS contract.

Please keep both sides of the safety boundary: `allowed_private_hosts = ["*"]` must not bypass public hosts, but an exact or suffix configured private hostname should still be able to skip the public-DNS check for that configured host. A regression like `allowed_private_hosts = ["local.internal"]` with URL `https://local.internal/api` and a panic DNS callback would pin the preserved behavior, alongside the existing public-host wildcard miss test.

### 🟡 Warning — The new shared SSRF boundary needs the old edge-case coverage

The implementation still appears to classify many non-global ranges correctly, but this refactor removed a lot of per-tool assertions while making `helpers::domain_guard` the shared boundary for multiple network-capable tools. The shared tests currently cover common private/local hosts, a public IPv4 and IPv6, and case-insensitive localhost/local domains, but they do not carry forward the old coverage for multicast, broadcast, reserved ranges, documentation ranges, benchmarking ranges, `0.0.0.0`, IPv4-mapped IPv6, IPv6 multicast/link-local/unique-local/documentation, and the `100.64.0.0/10` boundary cases.

Please move those assertions into the shared helper tests. For a high-risk tools refactor, the centralized security helper should keep the regression net that the duplicated implementations previously had.

crates/zeroclaw-tools/src/web_fetch.rs
```diff
@@ -702,59 +641,18 @@ fn extract_host(url: &str) -> anyhow::Result<String> {
...
-        .unwrap_or(host);
-
-    let has_local_tld = bare
-        .rsplit('.')
-        .next()
-        .is_some_and(|label| label == "local");
-
-    if bare == "localhost" || bare.ends_with(".localhost") || has_local_tld {
-        return true;
+    if !host_is_private_or_local {
```
singlerider: 🔴 This early return breaks named internal-DNS entries.  isn't syntactically private, so  is false and the configured  entry is never checked — it falls through to public-DNS validation and gets rejected. The old code matched  regardless. The test was changed from  to  (which IS ), masking this. Restore the named-host match and a  regression.

singlerider (commented):
Reviewed at head `1428cf2` against current `master`. The dedup is the right direction, but I verified @Audacity88's `allowed_private_hosts` regression — and the test was edited to hide it. Agreeing with his changes-requested.

### 🟢 The shared helper removes real security-logic drift

Centralizing `normalize_domain`, allowlist matching, and non-global IP classification into `helpers::domain_guard` and deleting the four per-tool copies (net ~-630 lines) is the correct consolidation for `browser`, `browser_open`, `http_request`, and `web_fetch`. The unmatched-bracket and case-insensitive-localhost handling moved into the shared helper, and `allowed_private_hosts = ["*"]` correctly no longer lets a public host bypass the normal `allowed_domains` check.

### 🔴 Blocking — confirmed: named internal-DNS entries in `allowed_private_hosts` stop working, and the test was changed to mask it

The new `host_matches_private_allowlist` opens with `if !host_is_private_or_local { return false; }`. A host like `local.internal` is not syntactically private (not `.local`, not `localhost`, not an IP), so `host_is_private_or_local` is `false` and the function returns early — the configured `allowed_private_hosts` entry is never consulted. The old code matched `host == domain` (and suffix) regardless of whether the host looked private. So an operator who lists `local.internal` in `allowed_private_hosts`, where it resolves to `192.168.x.x`/`10.x.x.x`, now has `web_fetch` fall through to the public-DNS check and reject it — a behavior regression in an explicit security-config path.

The regression test `allowed_private_domain_skips_dns_public_check` was changed from `local.internal` → `myhost.local`. Because `myhost.local` *is* syntactically private (`.local`), it passes the new early guard, so CI no longer proves the named-internal-DNS contract it was written to pin. Restore both sides of the boundary — `["*"]` must not bypass public hosts, but an exact/suffix configured private hostname must still skip the public-DNS check — and add a regression with `allowed_private_hosts = ["local.internal"]`, URL `https://local.internal/api`, and a panic DNS callback to lock it.

### 🟡 Agree — carry forward the SSRF edge-case coverage into the shared helper

The refactor removed many per-tool assertions while making `domain_guard` the single SSRF boundary for several network-capable tools, but the shared tests don't carry forward the old coverage for multicast, broadcast, reserved/documentation/benchmarking ranges, `0.0.0.0`, IPv4-mapped IPv6, IPv6 multicast/link-local/unique-local/documentation, and the `100.64.0.0/10` boundary. For a high-risk centralized security helper, the regression net the duplicated copies had must move into the shared tests.

Restore the named-private-host path (and its real test) and re-home the SSRF edge cases, and this is a good consolidation. Leaving @Audacity88's changes-requested in force.

jokemanfire:
Sorry for delay updated.

Audacity88 (approved):
Thanks for the update, @jokemanfire. I re-reviewed current head 853c4ca against my earlier changes-requested review, @singlerider's confirming comment, the current live PR state, green GitHub checks, inline review threads, and the shared `domain_guard` / `web_fetch` source. I did not run local Cargo validation.

### ✅ Resolved — named private hosts keep the old DNS-skip contract

The current `host_matches_private_allowlist` no longer exits before checking explicit configured hostnames. Explicit entries now match through the shared allowlist helper, so a configured internal DNS name can skip DNS public-host validation again. That resolves the blocker from my earlier review.

The safety side is also preserved: wildcard private-host allowlists only return true when the host is syntactically private/local, and an explicit public-looking private-host entry still has to pass `allowed_domains`. The new `private_allowlist_explicit_entry_must_pass_allowed_domains` and wildcard miss tests cover that boundary.

### ✅ Resolved — shared helper carries the important SSRF edge cases

The important edge-case coverage is now in `helpers::domain_guard`: multicast, broadcast, unspecified, reserved, documentation, benchmarking, RFC6598 shared space, IPv4-mapped IPv6, IPv6 documentation, and public IPv4/IPv6 cases are pinned there. That is the right place now that the helper is the shared security boundary for the network-capable tools.

One small test gap remains non-blocking from my side: the old `http_request` tests also checked the just-below and just-above RFC6598 public-side values, `100.63.0.1` and `100.128.0.1`. The current range check still implements the correct `100.64..=100.127` boundary, but adding those two assertions to `domain_guard` would make the moved coverage more exact.

### 🟢 What looks good — one shared guard now owns the duplicated tool logic

`browser`, `browser_open`, `http_request`, and `web_fetch` now call the same normalization, allowlist, and non-global-IP helpers while keeping each tool's URL-scheme and user-facing error-policy surface local. That preserves the intended scope boundary and removes the drift this review was meant to catch.

Approving this head to clear my earlier changes-requested review.

PR #7344, opened 2026-06-07
feat(gateway): opt-in remote /admin/reload via gateway.allow_remote_admin

Nillth:
## Summary

- **What changed and why:**
  - `POST /admin/reload` was localhost-only, so remote dashboards/CLIs got 403 and the only way to apply config changes from another host was a full process restart.
  - Adds `gateway.allow_remote_admin` (default **false**). When enabled, a non-loopback caller may trigger a reload **only if it also passes pairing auth**; loopback callers stay unauthenticated exactly as before.
  - **Fail-closed without pairing:** remote reload requires **both** `allow_remote_admin` **and** `require_pairing`. `require_auth` is a no-op when pairing is disabled, so opting in with pairing off is rejected (`ForbiddenNoPairing` → 403) rather than allowing an anonymous remote reload. The gate reads pairing state from the same `PairingGuard` `require_auth` consults, so the decision can't diverge from what's enforced.
  - The authorization decision is factored into a pure `admin_reload_gate()` helper (`Allow` / `RequireAuth` / `Forbidden` / `ForbiddenNoPairing`) with unit tests **plus handler-level regression tests** that exercise the real `handle_admin_reload` path, so the security-relevant logic is testable in isolation and end-to-end.
  - Documents the flag in the network-deployment guide.
- **Scope boundary:** Only `/admin/reload` gains the opt-in. `/admin/shutdown` and the pairing-code endpoints remain localhost-only. No change to pairing auth itself, no new endpoints, default behavior identical to master.
- **Blast radius:** Gateway admin surface only. With the flag off (default), behavior is byte-for-byte today's: loopback allowed, remote 403. With it on, remote reload requires the same pairing token every other authenticated gateway route uses.
- **Linked issue(s):** None.
- **Labels:** (applied snapshot: `enhancement`, `size: S`, `risk: high`, `config`, `docs`, `gateway`.)

  ```text
  $ cargo fmt --all -- --check
  (clean — no output)

  Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 11s   (no findings)

  $ cargo test -p zeroclaw-gateway -p zeroclaw-config --lib
  test result: ok. 773 passed; 0 failed; 0 ignored   (zeroclaw-config)
  test result: ok. 237 passed; 0 failed; 0 ignored   (zeroclaw-gateway)

  $ cargo test -p zeroclaw-gateway admin_reload
  test tests::admin_reload_gate_loopback_always_allowed ... ok
  test tests::admin_reload_gate_remote_blocked_by_default ... ok
  test tests::admin_reload_gate_remote_opt_in_requires_auth ... ok
  test tests::admin_reload_gate_remote_opt_in_without_pairing_is_rejected ... ok
  test tests::admin_reload_remote_default_off_is_forbidden ... ok
  test tests::admin_reload_remote_opt_in_without_pairing_does_not_reload ... ok
  test tests::admin_reload_remote_opt_in_missing_token_is_rejected ... ok
  test tests::admin_reload_remote_opt_in_invalid_token_is_rejected ... ok
  test tests::admin_reload_remote_opt_in_valid_token_reloads ... ok
  test tests::admin_reload_loopback_no_token_reloads ... ok
  test tests::allow_remote_admin_defaults_off ... ok
  test result: ok. 11 passed; 0 failed
  ```
  (Full `cargo test --workspace` OOMs in this dev environment — rustc is SIGKILLed compiling the `zeroclaw-runtime` test binary under parallelism — so the changed crates are tested directly; the full workspace is covered by CI's Test job.)
- **Beyond CI — what did you manually verify?** Handler-level regression tests now exercise the real `handle_admin_reload` path (`ConnectInfo` + `HeaderMap` + `PairingGuard` + config) and cover the boundary: default-off remote → 403; opt-in + pairing off (no token) → 403; opt-in + pairing on, missing/invalid token → 401; opt-in + pairing on, valid token → 200; loopback (no token) → 200. NOT verified by an external client: behavior behind reverse proxies that rewrite peer addresses (X-Forwarded-For is deliberately ignored — the gate uses the socket peer).
- **If any command was intentionally skipped, why:** Full-workspace `cargo test` run locally is replaced by changed-crate tests + CI (local OOM, noted above).

- New permissions, capabilities, or file system access scope? `Yes` — deliberately: an **opt-in** path for non-loopback callers to trigger daemon reload. Default-off; gated on pairing auth; fails closed when pairing is disabled; decision logic unit- and handler-tested. Reload applies the on-disk config — it does not accept config in the request.
- Secrets / tokens / credentials handling changed? `No` — reuses existing pairing auth.
- If any `Yes`, describe the risk and mitigation: risk is remote reload abuse; mitigated by default-off flag + mandatory pairing auth (fail-closed when pairing off) + loopback-only remaining the default posture. Shutdown stays localhost-only regardless.

## Compatibility (required)

- Backward compatible? `Yes` — flag defaults to false; existing deployments unchanged.
- Config / env / CLI surface changed? `Yes` — new optional `gateway.allow_remote_admin` boolean.
- If `No` or `Yes` to either: exact upgrade steps: none required; operators wanting remote reload set `allow_remote_admin = true` under `[gateway]` (and keep `require_pairing = true`), documented in the network-deployment guide.

- **Fast rollback command/path:** `git revert <sha>` — single feature; or operationally just leave/`set allow_remote_admin = false` (config toggle is the rollback).
- **Feature flags or config toggles:** `gateway.allow_remote_admin` (default false) IS the toggle.
- **Observable failure symptoms:** unexpected 200s on `POST /admin/reload` from non-loopback peers with the flag off (grep gateway logs for `/admin/reload`); reload requests bypassing auth would show in pairing-auth logs as absent entries for remote reloads.


Audacity88 (changes requested):
I reviewed the live PR at 3802497a915d20ee5343dc008a6a4899caf4e100, the three-file diff, current CI, and the gateway/config security paths around `handle_admin_reload`, `api::require_auth`, and `PairingGuard`. There are no prior review threads or active changes-requested reviews. I cannot approve this head yet because the new remote-admin path does not actually guarantee pairing auth under an existing gateway configuration.

### 🟢 What looks good — default-off remote admin is the right direction

Keeping remote `/admin/reload` disabled by default is the right baseline for this endpoint. I also like the split into a small `admin_reload_gate(...)` decision helper; it makes the intended loopback/default-off/remote-opt-in policy easy to read and test. The docs are also clear that `/admin/shutdown` and pairing-code endpoints should stay localhost-only.

### 🔴 Blocking — `allow_remote_admin` can allow anonymous remote reloads when pairing is disabled

The new `RequireAuth` branch calls `api::require_auth(&state, &headers)`, but that helper returns `Ok(())` immediately when `state.pairing.require_pairing()` is false. `PairingGuard::is_authenticated()` has the same contract: when pairing is disabled, even an empty token authenticates. That is fine for routes whose policy is "auth only when pairing is enabled," but it is not enough for this new admin exception.

As written, an operator can set:

```toml
[gateway]
allow_remote_admin = true
require_pairing = false
```

Then any non-loopback caller can hit `POST /admin/reload` without a bearer token and trigger a daemon reload. That contradicts the PR body and docs, which both say remote reload is accepted only if it also passes pairing authentication.

Please make this fail closed. Either reject remote admin reload whenever `allow_remote_admin = true` but pairing is disabled, or add a stricter remote-admin token check that validates an actual paired bearer token independently of the general `require_pairing` bypass. This also needs a handler- or route-level regression test, not only the pure `admin_reload_gate(...)` tests, covering at least:

- remote + `allow_remote_admin = true` + `require_pairing = false` + no token does not reload;
- remote + `allow_remote_admin = true` + pairing enabled + missing/invalid token is rejected;
- remote + `allow_remote_admin = true` + pairing enabled + valid token can reload.

### 🟡 Warning — the central remote path still needs real boundary evidence

The PR body still lists the remote manual pass as suggested rather than completed. For a high-risk gateway/admin change, I would like the final evidence to prove the actual endpoint behavior, not just the pure helper: default-off remote 403, opted-in remote without/with invalid auth rejected, opted-in remote with valid auth accepted, and loopback still working without auth. Handler/route tests may be enough if they exercise the same `ConnectInfo`, `HeaderMap`, and `PairingGuard` path; a short live or near-live smoke is even better.

### 🟡 Warning — refresh the PR body label snapshot

The live labels are `enhancement`, `size: S`, `risk: high`, `docs`, `config`, `gateway`, and `security`, but the PR body still says to expect `risk: medium` and does not include `security`. This is not the blocker and should not distract from the auth fix, but the public risk/security snapshot should match the applied labels before final approval.

Nillth:
Thanks — this was a real gap and a good catch. I've pushed a fix in response (6cdc94611): `admin_reload_gate` now takes `require_pairing` and returns a distinct `ForbiddenNoPairing` for the `allow_remote_admin = true` + `require_pairing = false` case, so that configuration returns **403** instead of allowing an anonymous reload. The gate reads its pairing input from `state.pairing.require_pairing()` — the same guard `require_auth` consults — so the decision can't diverge from what's actually enforced. (Note your review was against 3802497a9, which predates this fix, so it wouldn't have shown the change.)

I also added the **handler/route-level regression tests** you asked for, exercising the real `handle_admin_reload` path (`ConnectInfo` + `HeaderMap` + `PairingGuard` + config), not just the pure gate:

- loopback, no token → **200**
- remote, `allow_remote_admin` off → **403**
- remote, opt-in + `require_pairing = false`, no token → **403** (the closed hole)
- remote, opt-in + pairing on, missing token → **401**
- remote, opt-in + pairing on, invalid token → **401**
- remote, opt-in + pairing on, valid token → **200**

Docs and the `allow_remote_admin` schema comment were updated to state the flag fails closed — no effect unless `require_pairing` is on. The PR body's risk/label snapshot and validation evidence are also refreshed to match the current head.

On the `security` label: it isn't auto-applied here because the path labeler keys on `src/security/**` and runs with `sync-labels: true`, so a manual add gets stripped on the next push — the change is genuinely security-relevant, so happy to have a maintainer pin it if you'd like.

Audacity88 (approved):
Reviewed current head 79e92be904c4630edc43be8895cd3887ecee8208. I checked the live PR metadata and body, the author response to my prior review, the current three-file diff, the added handler-level tests, current checks, and the empty inline review-thread history. I did not run separate local cargo validation.

### ✅ Resolved — remote reload now fails closed when pairing is off

The prior blocker was that `allow_remote_admin = true` could route remote callers through `api::require_auth`, but `require_auth` is intentionally a no-op when pairing is disabled. This head closes that hole by making `admin_reload_gate(...)` take the pairing state from `state.pairing.require_pairing()` and return `ForbiddenNoPairing` for the opted-in-but-pairing-off case.

That keeps the authorization decision tied to the same `PairingGuard` state that `require_auth` consults. Remote reload now reaches `require_auth` only when pairing is enabled, and otherwise rejects before the no-op-auth path can apply.

### ✅ Resolved — the handler-level boundary evidence now covers the real path

The new tests go beyond the pure gate helper and exercise `handle_admin_reload` with `ConnectInfo`, `HeaderMap`, `PairingGuard`, config, and a live reload channel. They cover the cases I asked for: default-off remote is forbidden, opt-in with pairing off is forbidden, opt-in with pairing on rejects missing/invalid tokens, opt-in with a valid token reloads, and loopback still reloads without a token.

That is the right evidence for this risk tier because it proves the security property on the same handler path the gateway serves.

### ✅ Resolved — PR body and docs now match the shipped contract

The PR body now describes the fail-closed behavior, the handler-path tests, the intentionally skipped workspace test, and the live labels accurately. The network-deployment docs also state the operator contract clearly: remote reload is default-off, requires `allow_remote_admin = true` plus pairing, and `/admin/shutdown` plus pairing-code endpoints stay localhost-only.

No blockers from me.

PR #7346, opened 2026-06-07
fix(cli): print model names in 'zeroclaw models list' output

Nillth:
## Summary

- **What changed and why:**
  - `zeroclaw models list` only reported a per-provider **count** (e.g. `openai: 42 models`) — it never showed *which* models, so the subcommand couldn't actually answer "what can I select?".
  - Adds a `show_model_names` flag to `doctor::run_models()`; when set, each resolved model name is printed under its provider. The flag is `true` only for the `models list` subcommand and `false` for `models refresh` and the doctor health path, so the diagnostic output stays terse.
- **Scope boundary:** CLI output only — `src/main.rs` (subcommand wiring) and `crates/zeroclaw-runtime/src/doctor/mod.rs` (the print). No change to model resolution, catalog fetching, provider construction, or the doctor health checks. `+12 / −6`, 2 files.
- **Blast radius:** The `models list` command's stdout. `models refresh` and `doctor` output are unchanged (flag passed as `false`).
- **Linked issue(s):** None.
- **Labels:** (snapshot after applied — expect `cli`, `type: fix`, `risk: low`, `size: XS`.)

  ```text
  $ cargo fmt --all -- --check
  (clean)

  $ cargo clippy -p zeroclaw-runtime --bin zeroclaw
  (no findings)

  $ cargo test -p zeroclaw-runtime
  test result: ok. 2004 passed; 0 failed; 1 ignored

  # behavior (illustrative provider/model names; real run verified locally):
  #
  # BEFORE (master) — count only:
  #   [openai.default]
  #     ✅ 42 models
  #
  # AFTER (this branch) — names listed:
  #   [openai.default]
  #     ✅ 42 models
  #       • gpt-4o
  #       • gpt-4o-mini
  #       • o3
  #       • …
  #
  # `zeroclaw models refresh` is unchanged (counts only — flag=false):
  #   ✅ 42 models   (no bullets)
  ```
- **Beyond CI — what did you manually verify?** Ran `zeroclaw models list` on a configured instance and confirmed model names now print under each provider; `zeroclaw models refresh` output unchanged (counts only, no name list). Example above uses placeholder model names — real catalog output omitted intentionally (it reflects an operator's local model set; see Privacy). NOT verified: providers that need a live credential to enumerate (their list depends on configured keys).
- **If any command was intentionally skipped, why:** —

- New external network calls? `No` — uses the same model enumeration the count already relied on.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No` — the diff prints model names the operator's own providers already return at runtime; nothing personal is added to the code, tests, or docs. (The PR's example output uses placeholder model names rather than a real local catalog.)

## Compatibility (required)

- Backward compatible? `Yes` — additive output; the command's exit behavior and existing fields are unchanged.
- Config / env / CLI surface changed? `No` (same `models list` invocation; richer output).
- If `No` or `Yes` to either: no upgrade steps.

Low risk: `git revert <sha>` — output-only change in 2 files.

### DIFF
--- a/crates/zeroclaw-runtime/src/doctor/mod.rs
+++ b/crates/zeroclaw-runtime/src/doctor/mod.rs
@@ -267,6 +267,7 @@ pub async fn run_models(
     config: &Config,
     provider_override: Option<&str>,
     _use_cache: bool,
+    show_model_names: bool,
 ) -> Result<()> {
     let targets = doctor_model_targets(config, provider_override);
 
@@ -298,6 +299,11 @@ pub async fn run_models(
             Ok(models) => {
                 ok_count += 1;
                 println!("    ✅ {} models", models.len());
+                if show_model_names && !models.is_empty() {
+                    for m in &models {
+                        println!("      • {}", m);
+                    }
+                }
                 matrix_rows.push((
                     provider_name.clone(),
                     ModelProbeOutcome::Ok,
--- a/src/main.rs
+++ b/src/main.rs
@@ -4148,12 +4148,12 @@ async fn main() -> Result<()> {
         Commands::Cron { cron_command } => cron::handle_command(cron_command, &config),
 
         Commands::Models { model_command } => {
-            let model_provider = match &model_command {
-                ModelCommands::Refresh { model_provider, .. }
-                | ModelCommands::List { model_provider } => model_provider.as_deref(),
-                _ => None,
+            let (model_provider, show_names) = match &model_command {
+                ModelCommands::Refresh { model_provider, .. } => (model_provider.as_deref(), false),
+                ModelCommands::List { model_provider } => (model_provider.as_deref(), true),
+                _ => (None, false),
             };
-            doctor::run_models(&config, model_provider, false).await
+            doctor::run_models(&config, model_provider, false, show_names).await
         }
 
         Commands::Providers => {
@@ -4196,7 +4196,7 @@ async fn main() -> Result<()> {
             Some(DoctorCommands::Models {
                 model_provider,
                 use_cache,
-            }) => doctor::run_models(&config, model_provider.as_deref(), use_cache).await,
+            }) => doctor::run_models(&config, model_provider.as_deref(), use_cache, false).await,
             Some(DoctorCommands::Traces {
                 id,
                 event,


WareWolf-MoonWall (commented):
## Review: PR #7346

@Nillth — clean, focused fix that addresses the UX gap in `zeroclaw models list`. The implementation is minimal, the approach is sound, and the validation evidence is thorough.

### 🟢 What looks good

**The flag threading is correct.** The `show_model_names` parameter routes through exactly two call sites (`models list` → `true`, `models refresh` and `doctor` → `false`), leaving the diagnostic paths terse as intended. This is the right way to specialize output per subcommand without creating a hidden state cache or config field (SSOT-compliant).

**The output change is additive and safe.** The diff adds 6 lines inside a single `if` guard; the existing count-only line is unchanged, so the fallback behavior on empty catalogs or when the flag is `false` is untouched. The before/after examples in the PR description clearly show what changed.

**Test coverage is sufficient for the change.** The existing runtime test suite passes (2004 tests), and the change is CLI output-only with no new logic branches beyond the flag check. Unit-testing stdout formatting at this level wouldn't teach anything the manual verification didn't already cover.

### 🟡 Warning — label mismatch

The PR description declares `risk: low` but the actual label on the PR is `risk: high`. The diff itself is low-risk (output-only, 2 files, +12 / −6), but the `high` label may reflect that this touches `doctor::run_models()` — a diagnostic path that operators rely on during troubleshooting.

**Resolution path:** If the intent is `low`, update the label to match. If the `high` label is deliberate (e.g. because `doctor` is operator-facing and mistakes here are costly), note the rationale in a comment so future reviewers understand why a 2-file output change carries a `high` label.

### 🔵 Suggestion — consider a model-name limit or pagination hint

A provider with 100+ models (not uncommon for OpenAI or Anthropic) will flood the terminal with bullets. The current code prints every model name under each provider:

```rust
for m in &models {
    println!("      • {}", m);
}
```

This works fine for small catalogs but may be hard to scan when the list is long. Consider adding a cap (e.g. first 20 models + a `(… and N more)` hint) or a `--verbose` flag to control full vs. summary output. Not blocking — the UX is already better than "42 models" with no names — but worth tracking as a follow-up if operators report readability issues.

### 🔵 Suggestion — consistent indentation style

The model bullets use 6 spaces of indent (`"      • {}"`), which is visually distinct from the provider header (`"  [{}]"`, 2 spaces) and the status line (`"    ✅ {} models"`, 4 spaces). This is fine if the intent is to create a nested hierarchy (provider → status → model list), but the 2/4/6 progression is slightly irregular. A 2/4/8 or 0/2/4 pattern might be more consistent with typical CLI tree rendering. Again, not blocking — the current output is readable.

---

**Verdict:** Approve pending label clarification. The code is correct, the fix is minimal, and the validation evidence is solid. The label mismatch is the only issue that needs attention before merge.

WareWolf-MoonWall (approved):
Correcting verdict: the prior COMMENT review body stands in full, but the verdict flag was wrong.

The only finding in that review beyond 🟢 praise was a 🟡 label mismatch (`risk: low` vs `risk: high` label). The PR review protocol is explicit: "Do not request changes or hold merge solely because an author cannot edit labels." With no active blocks from other reviewers and no substantive code concerns, the correct verdict is `--approve`.

The `🔵` suggestions (model-name cap, indentation consistency) remain advisory. Approving.

theonlyhennygod (approved):
First review of head 8c742019f8f6ca0ec96986205eadc7d4e5713960. I read the PR body, the full diff (+12/−6, `crates/zeroclaw-runtime/src/doctor/mod.rs` and `src/main.rs`), and cross-checked it against the local source. CI is fully green. This is a tight, correct, output-only change — approving.

### 🟢 What looks good — Correct and complete, verified against the source

I checked the three things that could have gone wrong in a change like this, and all hold:

- **The printed value is a real model name.** `handle.list_models()` returns `Vec<String>` (the trait signature in `crates/zeroclaw-api/src/model_provider.rs:448`, and every provider impl), so `println!("      • {}", m)` prints the model name itself — not a `Debug` struct or an id wrapper.
- **The flag is gated exactly as described.** `show_model_names` is `true` only for `ModelCommands::List`, and `false` for `ModelCommands::Refresh` and the `DoctorCommands::Models` health path — so the diagnostic/refresh output stays terse and only `models list` grows the per-name bullets.
- **Every call site is updated.** `run_models` has exactly two callers (`src/main.rs:4156` and `:4199`) plus the definition; both now pass the new argument. No other caller in the tree, so nothing is silently left on the old arity.

The `!models.is_empty()` guard is a reasonable touch — it keeps a zero-catalog provider from opening an empty bullet block under its `✅ 0 models` line.

### 🟢 What looks good — Careful privacy handling in the validation evidence

The PR body uses placeholder model names in the before/after example and explicitly says why — a real `models list` run reflects the operator's own configured providers/catalog, which shouldn't be pasted into a public PR. That's the right instinct for the privacy contract, and calling it out so a reviewer doesn't go looking for "real" output is exactly the kind of disclosure that makes a small PR easy to trust.

### 🔵 Suggestion — No test pins the new branch

There's no test asserting that `models list` emits names while `refresh`/`doctor` don't. I don't think that should block this — `run_models` prints via `println!` and does live provider probing, so the print path isn't structured for a unit test today, and adding stdout capture for six lines would be over-engineering. But if this output is going to be relied on, the cheap version is to factor the per-provider formatting into a small pure function (`fn format_models_block(models: &[String], show_names: bool) -> String`) that a test can assert on without touching the network. Optional, and fine to pass on.

## Label hygiene — `risk: high` looks auto-over-classified

The `risk: high` label is auto-applied from the changed paths (root `src/*.rs` + `doctor/**` + `runtime`), but this diff is output-only: no change to model resolution, catalog fetching, provider construction, or the doctor health checks — just additional `println!` lines behind a flag. You flagged the same expectation in the body (`expect … risk: low`). I'd correct it to `risk: low` (per AGENTS.md, add `risk: manual` + `risk: low` to override the auto-label). I can make that change myself if you'd like — say the word and I'll apply it; I didn't want to mutate labels on your PR without a nod.

## Merge readiness — green here, but a re-run may surface an unrelated master break

This PR's CI is green, but that run predates a breakage now sitting on `master`: an orphaned test constructor at `crates/zeroclaw-channels/src/orchestrator/mod.rs:15443` still sets removed `ChannelRuntimeContext` fields (`default_model_provider`/`api_key`/`api_url` vs the migrated `model_provider_ref`), which fails `Lint` + `Test` to compile `zeroclaw-channels`. It's wholly unrelated to this diff (different crate; you touch only `doctor` + `main`), but if this PR is rebased or its CI re-runs before that one-line master fix lands, expect those two jobs to go red through no fault of this change. Flagging only so the red, if it appears, isn't mistaken for something here.

theonlyhennygod:
Applied the label correction noted in my review: `risk: high` → `risk: low` (+ `risk: manual` to pin the override), since this is an output-only change with no behavior/security/boundary impact. @Nillth — no action needed from you on labels.

Nillth:
Thanks for the reviews, @WareWolf-MoonWall and @theonlyhennygod. Following up on the 🔵 model-list flood suggestion.

I have a change staged that truncates the per-provider list by default and adds a `--all` flag:

- `zeroclaw models list` prints the first **10** names per provider, then a `… and N more (use --all)` hint.
- `zeroclaw models list --all` prints the full catalog.
- `refresh` and `doctor models` are unchanged — names stay hidden on those paths.

Example:

```
$ zeroclaw models list
  [openai]
    ✅ 64 models
      • gpt-5
      • gpt-5-mini
      • gpt-5-nano
      ... (10 shown) ...
      • o3
      … and 54 more (use --all)
```

I also pulled the per-provider formatting into a small pure function
`format_model_names(models: &[String], limit: Option<usize>) -> String`
(`Some(n)` truncates + appends the hint, `None` prints all) with a unit test covering the cap+hint, `--all`, and under-cap cases — which also covers @theonlyhennygod's testability point without any live probing.

Before I push it onto this PR: does a default cap of **10** + `--all` sound right, or would you prefer a different default (or a `--verbose` spelling)? I left the 2/4/6 indentation as-is to avoid churning the already-approved output. Happy to adjust based on what you think.

PR #7347, opened 2026-06-07
fix(channels/discord): ignore system messages so the bot stops replying to thread creation

Nillth:
## Summary

- **What changed and why:**
  - When a thread is created in a Discord channel, Discord posts a system
    `MESSAGE_CREATE` of **type 18 (`THREAD_CREATED`)** in the parent channel —
    authored by the human who created the thread, with the thread name as
    content. The Discord adapter never inspected the message `type`, so this
    sailed through `admit_discord_message` and the bot **replied to the
    thread's birth message**.
  - Add a message-type gate: only **`DEFAULT` (0)** and **`REPLY` (19)** are
    treated as real user turns; all system/auto types — `THREAD_CREATED` (18),
    `THREAD_STARTER_MESSAGE` (21), pins (6), joins (7), boosts, etc. — are
    skipped. Extracted as `is_conversational_message_type()` for testability.
- **Scope boundary:** Discord MESSAGE_CREATE handling only; one new guard plus a
  pure helper. No change to mention gating, attachment handling, threading
  keys, or any other channel.
- **Blast radius:** The set of Discord events the bot responds to — strictly
  narrowed to real messages. DMs and normal channel/thread messages (types 0/19)
  are unaffected; only system messages stop triggering responses.
- **Linked issue(s):** None filed (reported by operator).
- **Labels:** (applied snapshot: `bug`, `size: XS`, `risk: medium`, `channel`, `channel:discord`.)

NOTE: the `discord` module is behind `--features channel-discord` (not in the
default feature set), so the battery is run with that feature or it won't even
compile the changed file.

```text
$ cargo fmt --all -- --check
(clean)

$ cargo clippy -p zeroclaw-channels --features channel-discord --all-targets -- -D warnings
Finished `dev` profile in 51.16s   (no findings)

test discord::tests::thread_created_and_system_messages_are_not_conversational ... ok
test result: ok. 819 passed; 0 failed; 1 ignored
```

- **Beyond CI — what did you manually verify?** Verified **live against a real
  Discord guild**, not just reasoned: deployed the fixed binary to a running
  instance with 4 active Discord bots and created a new thread (thread id
  `1513176176716157068`).
  - **Thread birth → no response in the parent channel** (the `THREAD_CREATED`
    system message was correctly ignored). The runtime trace shows the
    `THREAD_CREATED` produced no inbound at all — dropped by the type gate
    before it could become a `ChannelMessage` (the gate `continue`s pre-logging,
    so the silence is the evidence).
  - **Real messages in the thread → still answered.** Two follow-up messages
    posted in the thread were handled normally (Discord message-ID pairs,
    question → answer): `1513177192543162438` → `1513177482956640266`, and
    `1513177747982127296` → `1513178029097226320`. The trace for the thread
    shows `channel inbound message` → `reply delivered` for the genuine
    messages only.

  Confirms the prior behavior (the bot replying to a thread's birth) is gone
  while real messages still work. Unit test additionally asserts message types
  0/19 admitted and 18/21/6/7 rejected.
- **If any command was intentionally skipped, why:** —

> **Out of scope (noted during testing, NOT addressed here):** the two answers
> above took ~68s each; the runtime trace attributes the bulk to repeated
> proactive context-compression rounds before the LLM call, unrelated to this
> message-type gate. The gateway stayed connected throughout (no reconnect),
> so this fix neither causes nor fixes that latency. Tracked separately.

## Compatibility (required)

- Backward compatible? `Yes` — strictly removes responses to system messages the
  bot should never have answered. Real messages behave identically.
- Upgrade steps: none.

Low risk: `git revert <sha>` — one guard + one helper in a single file.


Audacity88 (approved):
Reviewed current head 37ce85f082b988e9abcf9b46fe8b3078b1e25bb4. I checked the live PR metadata and body, the one-file diff and file list, the current check snapshot, and the empty public comment/review history. I did not run separate local validation.

### 🟢 What looks good — system messages are filtered before they become channel turns

The visible patch puts the Discord message-type gate in the `MESSAGE_CREATE` handler immediately after payload extraction and before self/bot filtering. Rejected message types `continue` out of the handler at that point, so `THREAD_CREATED`, `THREAD_STARTER_MESSAGE`, pins, joins, and similar system messages stop before they can be treated like user turns.

The allowlist is also intentionally narrow. Keeping `DEFAULT` and `REPLY` as the only conversational message types preserves normal messages and reply messages while dropping the thread-birth event that caused the false response. The focused test covers both admitted types and the important rejected system types, including the original `THREAD_CREATED` case.

No blockers from me.

PR #6967, opened 2026-05-27
fix(matrix): preserve self-anchored root sessions

drbparadise:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Treat Matrix messages whose `thread_ts` equals their own event id as root-scoped for conversation history.
  - Preserve real Matrix thread replies by continuing to key them on the thread root id.
  - Add regression coverage for two top-level self-anchored Matrix messages from the same sender and room sharing one history key.
- **Scope boundary:** Does not change Matrix send/reply APIs, outbound threading behavior, storage schemas, config, CLI, or non-Matrix channel history keys.
- **Blast radius:** Matrix channel conversation-history keying only. The affected behavior is session continuity for Matrix top-level events that carry a self-anchor id.
- **Linked issue(s):** Related #6524; follow-up to #6525.
- **Labels:** Current GitHub labels: `bug`, `size: XS`, `risk: medium`, `channel`, `channel:matrix`.

## Validation Evidence (required)

Local validation is the signal CI cannot replace. Run the full battery and paste literal output (tails, failures, warnings — not "all passed").

```bash
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test
```

- **Commands run and tail output:**

```bash
$ cargo fmt --all -- --check
# exit 0; no output

$ git diff --check
# exit 0; no output

$ cargo clippy --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.40s

$ cargo test -p zeroclaw-channels matrix_ --lib -- --nocapture
running 5 tests
test orchestrator::tests::followup_thread_id_does_not_open_matrix_thread_for_root_message ... ok
test orchestrator::tests::matrix_root_conversation_history_key_omits_event_id ... ok
test orchestrator::tests::matrix_thread_conversation_history_key_uses_thread_root ... ok
test orchestrator::tests::matrix_self_anchored_root_history_key_omits_event_id ... ok
test orchestrator::tests::sender_session_ids_match_migrated_matrix_sender_rows ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 1146 filtered out; finished in 0.01s

$ cargo test
     Running tests/test_system.rs (target/debug/deps/system-e7d4ad239846b569)

running 9 tests
test system::multi_agent_e2e::peer_group_routes_messages_only_within_resolved_peer_set ... ok
[...]

PR #7349, opened 2026-06-07
fix(matrix): strip self-anchored event_id from interruption_scope_id (#6958)

### LINKED ISSUE #6958: Matrix channel: session keyed on event_id causes amnesia between messages
## Summary

Every incoming Matrix message creates a new isolated session because the session key includes the Matrix event ID, which is globally unique per message. The bot has no access to prior conversation history — each reply starts cold.

## Environment

- ZeroClaw version: 0.7.5
- Channel: Matrix (via continuwuity homeserver)
- Deployment: Docker, `ghcr.io/zeroclaw-labs/zeroclaw:debian`

## Observed session key format

```
matrix_<room_id>_<event_id>_<sender>
```

Confirmed by querying `sessions.db` directly:

```
matrix_!<room_id>_$fowW6M2KJn..._@user:homeserver
matrix_!<room_id>_$qLK8DuCr7L..._@user:homeserver
matrix_!<room_id>_$v0Z878zBUc..._@user:homeserver
... (57 entries, same room, same user, different event_id each time)
```

Every session has `message_count = 2` — one user message and one assistant reply. No session ever accumulates more than one exchange.

## Expected session key format

```
matrix_<room_id>_<sender>
```

The session should be scoped to the room + sender pair, not to a specific event. This is how the gateway WebChat channel works (`gw_<uuid>`), which correctly accumulates history across multiple turns.

## Impact

- Conversational context is completely lost between messages
- The "I am in Italy → where am I?" pattern fails at the session layer before the LLM is involved
- `max_history_messages` and `context_compression` settings have no effect since history never grows past 1 turn
- Memory recall partially mitigates this but degrades because the `user_msg` memory key is static — each message overwrites the previous one, so only the last user message survives in `brain.db`

## Additional finding: static `user_msg` memory key

Related degradation: one memory entry in `brain.db` uses the key `user_msg` with no unique suffix. Each incoming message overwrites it, so the memory layer also loses prior context even when embeddings are configured.

## Workaround

None available via config. The sync loop itself is healthy (uses `session.json` from disk). The issue is purely in how the session key is constructed per incoming event.

## Suggested fix

Scope the Matrix session key to `matrix_{room_id}_{sender}` (dropping `{event_id}`). This matches the expected multi-turn conversation model and is consistent with how other channels handle session identity.

chengzhichao-xydt:
## Summary

When `reply_in_thread` is enabled and a Matrix message has no existing thread_id, `resolve_outbound_anchor` returns the message's own event_id as the outbound anchor. This anchor is a delivery/threading detail — not a conversation boundary.

`conversation_history_key` already strips self-anchored event_ids (since afc5828d / #6967). However `interruption_scope_id` still carried the event_id, which caused in-flight cancellation keys to diverge from the sender+room scope used by `conversation_history_key`.

## Fix

Strip the self-anchored event_id from `interruption_scope_id` when constructing the Matrix `ChannelMessage`. The check mirrors `conversation_history_key`: when the outbound anchor equals the message's own event_id, set `interruption_scope_id` to `None`.

## Verification

- `cargo test -p zeroclaw-channels matrix_self_anchored` — passes
- `cargo test -p zeroclaw-channels matrix_root` — passes  
- `cargo test -p zeroclaw-channels followup_thread` — passes
- All existing Matrix channel tests pass

## Changed Files

- `crates/zeroclaw-channels/src/matrix.rs` — strip self-anchored event_id from interruption_scope_id

Closes #6958

### DIFF
--- a/crates/zeroclaw-channels/src/matrix.rs
+++ b/crates/zeroclaw-channels/src/matrix.rs
@@ -5140,8 +5140,8 @@ mod tests {
 
     mod thread_extraction {
         use super::super::inbound::{
-            extract_mentions_user_ids, extract_thread_id,
-            interruption_scope_from_anchor, resolve_outbound_anchor,
+            extract_mentions_user_ids, extract_thread_id, interruption_scope_from_anchor,
+            resolve_outbound_anchor,
         };
         use matrix_sdk::event_handler::RawEvent;
         use matrix_sdk::ruma::serde::Raw;
@@ -5242,12 +5242,10 @@ mod tests {
             // interruption_scope_id must stay set to the thread root.
             let event_id = "$reply:server".parse().expect("event id");
             let thread_root = "$root:server".parse().expect("thread root");
-            let outbound =
-                resolve_outbound_anchor(Some(&thread_root), &event_id, true);
+            let outbound = resolve_outbound_anchor(Some(&thread_root), &event_id, true);
             assert_eq!(outbound.as_deref(), Some("$root:server"));
             assert_eq!(
-                interruption_scope_from_anchor(outbound.as_deref(), &event_id)
-                    .as_deref(),
+                interruption_scope_from_anchor(outbound.as_deref(), &event_id).as_deref(),
                 Some("$root:server")
             );
         }
--- a/crates/zeroclaw-channels/src/matrix.rs
+++ b/crates/zeroclaw-channels/src/matrix.rs
@@ -1856,10 +1856,8 @@ mod inbound {
         // boundary. Strip it from interruption_scope_id so in-flight
         // cancellation keys match the sender+room scope used by
         // conversation_history_key.
-        let interruption_scope = match outbound_anchor.as_deref() {
-            Some(anchor) if anchor == ev.event_id.as_str() => None,
-            _ => outbound_anchor.clone(),
-        };
+        let interruption_scope =
+            interruption_scope_from_anchor(outbound_anchor.as_deref(), &ev.event_id);
 
         let msg = ChannelMessage {
             id: ev.event_id.to_string(),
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
I reviewed current head `cfa534c`, the linked #6958 report, the already-merged #6967 history-key fix, the current Matrix inbound path, and the current orchestrator interruption/history handling. I did not run additional local validation.

### 🟢 What looks good — Keeping delivery threading separate from conversation scope is the right direction

The shape of this fix makes sense. Matrix still needs `thread_ts` to carry the self-anchor when `reply_in_thread` opens a new Matrix thread, but that self-anchor should not automatically become the logical interruption scope for top-level conversation turns. Preserving the delivery anchor while stripping the cancellation scope is the right boundary to protect.

### 🔴 Blocking — Please add a regression for the Matrix message boundary this PR changes

The changed behavior is not `resolve_outbound_anchor()` itself and it is not the existing #6967 `conversation_history_key()` behavior. The changed behavior is the `ChannelMessage` emitted by the Matrix inbound handler: for a self-anchored root event, `thread_ts` should stay set to the event id while `interruption_scope_id` should be `None`.

The validation listed in the PR body does not prove that. The `matrix_self_anchored` coverage comes from the prior history-key fix and would still pass before this patch, because that test constructs `ChannelMessage` values directly and only asserts the history key. The current diff changes a different field at the Matrix channel boundary.

Please add focused regression coverage for the new contract. The important cases are:

- root Matrix event with `reply_in_thread = true`: `thread_ts == event_id`, `interruption_scope_id == None`
- real Matrix thread reply: `thread_ts == thread_root`, `interruption_scope_id == Some(thread_root)`

That can be a small helper-level test if extracting the scope decision is cleaner, or an inbound-handler test if the existing Matrix test harness makes that practical. Either way, the test should fail on current `master` and pass with this PR, so #6958 does not regress again through a future "cleanup" that re-couples delivery threading and interruption scope.

Audacity88 (approved):
Reviewed current head `0498ac5efb3e6bd395561c3ab326fdf386b0961b`, the prior Audacity88 changes-requested review, the current Matrix diff, the existing #6967 history-key behavior, and the Matrix/orchestrator interruption-scope boundary. CI is green, the merge result is clean in a local `merge-tree` check, and I am approving this re-review. I did not run additional local cargo validation. This approval is for the reviewed head; final merge-readiness should still refresh against current `master` before merge.

### ✅ Resolved — Matrix boundary regression now covers the requested contract

The previous blocker asked for coverage at the Matrix message boundary where delivery threading and interruption scope diverge. The new `interruption_scope_from_anchor()` tests cover the exact cases that mattered: a self-anchored root keeps `thread_ts` on the event id while stripping `interruption_scope_id`, a real thread reply preserves the thread-root interruption scope, and a no-anchor root stays unscoped. That would have failed before this patch and now locks the #6958 behavior directly.

### 🟢 What looks good — Delivery threading stays separate from conversation scope

The implementation keeps Matrix delivery behavior intact: `thread_ts` still carries the outbound anchor used for threaded replies, while `interruption_scope_id` now only carries a real conversation boundary. That preserves the #6967 history-key model and prevents self-anchored root events from splitting cancellation/session scope by event id again.

### 🔵 Suggestion — Refresh the PR body validation list when convenient

The PR body still lists the older validation filters, not the new `interruption_scope_from_anchor` regression coverage that resolved this review. This is not blocking because the diff and CI now provide the important proof, but updating the body before merge would make the public record easier to audit later.

PR #7351, opened 2026-06-07
feat(mcp): auto-reconnect on stale session or dropped stream

itripn:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Remote MCP servers that restart (or drop their SSE stream) left the client
    pinned to a dead session — every later tool call failed permanently until
    the daemon was restarted. The HTTP transport bailed on any non-2xx without
    clearing the stale `Mcp-Session-Id`; SSE reported a dropped stream as a
    tool error, so it was never treated as recoverable.
  - Added a typed `McpTransportError` (`StaleSession`, `TransportClosed`) so the
    client can distinguish recoverable transport failures from genuine tool
    errors via downcast (no string-matching).
  - Added `reset()` to the `McpTransportConn` trait (default no-op): HTTP clears
    `session_id`; SSE tears down the reader task and clears the cached endpoint
    and pending requests so the next call re-handshakes.
  - `McpServer::call_tool` now runs a bounded reconnect loop — reset →
    re-`initialize` → retry — capped at 2 attempts with a 500ms backoff, logged
    at `warn`.
- **Scope boundary:** No stdio respawn (subprocess restart is out of scope). No
  new config surface — retry bounds are hardcoded. Tool-call semantics for
  success and genuine errors are unchanged (errors/timeouts are never retried).
- **Blast radius:** MCP tool-call path only (`zeroclaw-tools` MCP client +
  transport). No change to stdio, the tool registry, or the agent loop.
- **Linked issue(s):** None.
- **Note:** Rebased onto current `master`; integrates cleanly with the new
  `result.isError` handling and the `zeroclaw_log` logging convention (the
  reconnect event logs via `zeroclaw_log` with `Action::Reconnect`).

```bash
cargo fmt --all -- --check
cargo test
```

  - `cargo fmt --all -- --check` → clean (exit 0, no diff).
  - `cargo clippy --all-targets -- -D warnings` → clean:
    `Finished \`dev\` profile [unoptimized + debuginfo] target(s)` (exit 0, no warnings).
  - `cargo test` (full workspace) → exit 0, all suites `ok`. The changed crate:
    `test result: ok. 1326 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out` (zeroclaw-tools),
    including the 5 new reconnect tests; doctests `ok`.
- **Beyond CI — what did you manually verify?** Added integration tests with
  `wiremock`: a streamable-HTTP server returns 404 on the first `tools/call`
  then succeeds — the call transparently recovers and `initialize` runs twice;
  a genuine JSON-RPC tool error is asserted to NOT trigger a reconnect
  (`initialize` runs once). Unit tests cover the 404→`StaleSession` mapping and
  both `reset()` implementations.
- **Not verified:** SSE reconnect end-to-end (SSE streaming is impractical to
  mock with wiremock) — covered at the `reset()` + error-classification level
  only.

- New external network calls? **No** (a reconnect re-hits the already-configured
  MCP server URL).

## Compatibility (required)

- **Feature flags or config toggles:** None (reconnect is unconditional, bounded
  to 2 attempts).
- **Observable failure symptoms:** grep logs for `reconnecting (attempt N/2)` —
  repeated reconnect warnings for one server indicate a server that restarts or
  expires sessions faster than calls complete.

---

### DIFF
--- a/crates/zeroclaw-tools/src/mcp_transport.rs
+++ b/crates/zeroclaw-tools/src/mcp_transport.rs
@@ -336,7 +336,14 @@ impl McpTransportConn for HttpTransport {
 
         if !resp.status().is_success() {
             let status = resp.status();
-            if status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE {
+            // A 404/410 only means "stale session" when this request carried an
+            // `Mcp-Session-Id` the server no longer recognizes (MCP spec 2025-06-18,
+            // Session Management). Without a session id, a 404 is just a missing
+            // endpoint (typo'd `url`, wrong path, proxy misroute) — surface it as a
+            // plain error so `call_tool` doesn't burn a reconnect on it.
+            if self.session_id.is_some()
+                && (status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE)
+            {
                 return Err(McpTransportError::StaleSession {
                     status: status.as_u16(),
                 }
@@ -1570,6 +1577,8 @@ mod tests {
             ..Default::default()
         };
         let mut transport = HttpTransport::new(&config).expect("build transport");
+        // A 404 only signals a stale session when the request carried a session id.
+        transport.session_id = Some("sess-1".into());
         let req = JsonRpcRequest::new(1, "tools/call", serde_json::json!({}));
         let err = transport
             .send_and_recv(&req)
@@ -1579,6 +1588,46 @@ mod tests {
             Some(McpTransportError::StaleSession { status }) => assert_eq!(*status, 404),
             other => panic!("expected StaleSession, got {other:?}"),
         }
+    }
+
+    #[tokio::test]
+    async fn http_transport_404_without_session_is_plain_error() {
+        use wiremock::matchers::method;
+        use wiremock::{Mock, MockServer, ResponseTemplate};
+
+        let server = MockServer::start().await;
+        Mock::given(method("POST"))
+            .respond_with(ResponseTemplate::new(404))
+            .mount(&server)
+            .await;
+
+        let config = McpServerConfig {
+            name: "test-http".into(),
+            transport: McpTransport::Http,
+            url: Some(server.uri()),
+            ..Default::default()
+        };
+        // No session id was ever issued (stateless server, or a misconfigured url):
+        // a 404 here is a missing endpoint, not a stale session — it must NOT map to
+        // StaleSession (which would make `call_tool` burn a wasted reconnect).
+        let mut transport = HttpTransport::new(&config).expect("build transport");
+        assert!(transport.session_id.is_none());
+        let req = JsonRpcRequest::new(1, "tools/call", serde_json::json!({}));
+        let err = transport
+            .send_and_recv(&req)
+            .await
+            .expect_err("404 should error");
+        assert!(
+            !matches!(
+                err.downcast_ref::<McpTransportError>(),
+                Some(McpTransportError::StaleSession { .. })
+            ),
+            "sessionless 404 must not be classified as StaleSession, got: {err:?}"
+        );
+        assert!(
+            err.to_string().contains("MCP server returned HTTP 404"),
+            "got: {err}"
+        );
     }
 
     #[tokio::test]
--- a/crates/zeroclaw-tools/src/mcp_transport.rs
+++ b/crates/zeroclaw-tools/src/mcp_transport.rs
@@ -9,7 +9,7 @@ use tokio::sync::{Mutex, Notify, oneshot};
 use tokio::time::{Duration, timeout};
 use tokio_stream::StreamExt;
 
-use crate::mcp_protocol::{INTERNAL_ERROR, JsonRpcError, JsonRpcRequest, JsonRpcResponse};
+use crate::mcp_protocol::{JsonRpcRequest, JsonRpcResponse};
 use zeroclaw_config::schema::{McpServerConfig, McpTransport};
 
 /// Maximum bytes for a single JSON-RPC response.
@@ -65,6 +65,25 @@ fn apply_request_timeout(
     }
 }
 
+// ── Transport Errors ───────────────────────────────────────────────────────
+
+/// Transport-level failures that are recoverable by reconnecting — resetting
+/// the session and re-running the MCP handshake — rather than surfacing to the
+/// caller. Distinct from a genuine tool/application error, which must be
+/// reported as-is and never retried.
+#[derive(Debug, thiserror::Error)]
+pub enum McpTransportError {
+    /// The server no longer recognizes our session (typically after it
+    /// restarted). Surfaced from HTTP 404/410 responses.
+    #[error("MCP session is stale (HTTP {status})")]
+    StaleSession { status: u16 },
+
+    /// The underlying stream/connection dropped before a response arrived
+    /// (e.g. SSE EOF or connection reset).
+    #[error("MCP transport connection closed")]
+    TransportClosed,
+}
+
 // ── Transport Trait ──────────────────────────────────────────────────────
 
 /// Abstract transport for MCP communication.
@@ -72,6 +91,12 @@ fn apply_request_timeout(
 pub trait McpTransportConn: Send + Sync {
     /// Send a JSON-RPC request and receive the response.
     async fn send_and_recv(&mut self, request: &JsonRpcRequest) -> Result<JsonRpcResponse>;
+
+    /// Reset per-connection session state so the next operation re-establishes
+    /// a fresh session. Default is a no-op for stateless transports (stdio).
+    async fn reset(&mut self) -> Result<()> {
+        Ok(())
+    }
 
     /// Close the connection.
     async fn close(&mut self) -> Result<()>;
@@ -310,7 +335,14 @@ impl McpTransportConn for HttpTransport {
             .context("HTTP request to MCP server failed")?;
 
         if !resp.status().is_success() {
-            bail!("MCP server returned HTTP {}", resp.status());
+            let status = resp.status();
+            if status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE {
+                return Err(McpTransportError::StaleSession {
+                    status: status.as_u16(),
+                }
+                .into());
+            }
+            bail!("MCP server returned HTTP {}", status);
         }
 
         self.update_session_id_from_headers(resp.headers());
@@ -353,6 +385,13 @@ impl McpTransportConn for HttpTransport {
 
         let resp_text = resp.text().await.context("failed to read HTTP response")?;
         parse_jsonrpc_response_text(&resp_text)
+    }
+
+    async fn reset(&mut self) -> Result<()> {
+        // Drop the stale session so the next request re-initializes and the
+        // server issues a fresh `Mcp-Session-Id`.
+        self.session_id = None;
+        Ok(())
     }
 
     async fn close(&mut self) -> Result<()> {
@@ -539,22 +578,14 @@ impl SseTransport {
                 }
             }
 
+            // Stream closed: drop every pending sender so each waiter observes a
+            // `RecvError`, which `send_and_recv` maps to
+            // `McpTransportError::TransportClosed` to trigger a reconnect.
             let pending = {
                 let mut guard = shared.lock().await;
                 std::mem::take(&mut guard.pending)
             };
-            for (_, tx) in pending {
-                let _ = tx.send(JsonRpcResponse {
-                    jsonrpc: crate::mcp_protocol::JSONRPC_VERSION.to_string(),
-                    id: None,
-                    result: None,
-                    error: Some(JsonRpcError {
-                        code: INTERNAL_ERROR,
-                        message: "SSE connection closed".to_string(),
-                        data: None,
-                    }),
-                });
-            }
+            drop(pending);
         }));
         self.stream_state = SseStreamState::Connected;
 
@@ -993,6 +1024,12 @@ impl McpTransportConn for SseTransport {
 
         if let Some(status) = last_status {
             if !status.is_success() {
+                if status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE {
+                    return Err(McpTransportError::StaleSession {
+                        status: status.as_u16(),
+                    }
+                    .into());
+                }
                 bail!("MCP server returned HTTP {}", status);
             }
         } else {
@@ -1003,15 +1040,28 @@ impl McpTransportConn for SseTransport {
             bail!("MCP server returned no response");
         };
 
-        rx.await.map_err(|_| {
-            ::zeroclaw_log::record!(
-                ERROR,
-                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Fail)
-                    .with_outcome(::zeroclaw_log::EventOutcome::Failure),
-                "mcp_transport: SSE response channel closed"
-            );
-            anyhow::Error::msg("SSE response channel closed")
-        })
+        // A dropped receiver means the SSE reader task tore down the stream
+        // before our response arrived — recoverable via reconnect.
+        rx.await
+            .map_err(|_| McpTransportError::TransportClosed.into())
+    }
+
+    async fn reset(&mut self) -> Result<()> {
+        // Tear down the reader task and clear the cached endpoint/session state
+        // so the next send re-handshakes: a fresh GET stream and a new
+        // `endpoint` event from the (possibly restarted) server.
+        if let Some(tx) = self.shutdown_tx.take() {
+            let _ = tx.send(());
+        }
+        if let Some(task) = self.reader_task.take() {
+            task.abort();
+        }
+        self.stream_state = SseStreamState::Unknown;
+        let mut guard = self.shared.lock().await;
+        guard.message_url = None;
+        guard.message_url_from_endpoint = false;
+        guard.pending.clear();
+        Ok(())
     }
 
     async fn close(&mut self) -> Result<()> {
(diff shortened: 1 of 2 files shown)


Audacity88 (commented):
I am pausing the full review on this head because GitHub reports the branch as `DIRTY` / `CONFLICTING` against current `master`.

### 🟢 What looks good — the reconnect shape is narrow

The direction here is good. Recovering stale remote MCP sessions through typed transport errors, a reset hook, a bounded retry budget, and a re-handshake is the right kind of fix. It avoids string-matching and keeps genuine JSON-RPC tool errors out of the retry path.

### 🟡 Warning — refresh needs to preserve the newer MCP error/timeout behavior

Please refresh this branch against current `master` before the next review pass. The conflicts are in `crates/zeroclaw-tools/src/mcp_client.rs` and `crates/zeroclaw-tools/src/mcp_transport.rs`, and they overlap with newer MCP changes that should not be lost during conflict resolution:

- `mcp_client.rs` now treats `result.isError: true` as a sanitized tool-execution error instead of returning that envelope as a normal success.
- `mcp_transport.rs` now has per-request HTTP/SSE timeout handling for tool calls and the newer structured logging around transport failures.

After the branch is refreshed and CI reruns on that refreshed head, I can do the real review of the reconnect behavior. The important thing during the refresh is to combine the reconnect loop with the current master behavior, not replace it.

Audacity88:
Thanks for the refresh. GitHub still reports this branch as `DIRTY` / `CONFLICTING` against current `master`, so I am still pausing the full reconnect review until the branch is actually conflict-clear.

The remaining conflict is still in the MCP client area. Locally, the merge conflict is in `crates/zeroclaw-tools/src/mcp_client.rs` near the test module tail: current `master` has the stdio child-process cleanup regression, and this branch adds the reconnect regression tests. The refresh should keep both sets of tests. `crates/zeroclaw-tools/src/mcp_transport.rs` now appears to merge cleanly, but it should still be rechecked after the conflict resolution.

Once GitHub shows the branch as mergeable and CI has rerun on that conflict-free head, I can do the real review of the reconnect behavior.

itripn:
Sorry for the thrash Dan. I'll take a look at this tonight and make sure I get it cleaned up. Been one of those weeks.

itripn:
Refreshed onto current `master` (now at #7520) and force-pushed. GitHub shows the branch `MERGEABLE` again and CI is **green** on the new head (`504f2290e`).

Resolution notes, since you flagged where to be careful:

- **The only real conflict was the test-module tail of `mcp_client.rs`** — exactly as you called out. `master`'s new `dropping_stdio_registry_reaps_child_process` (#7392) landed in the same spot as this branch's reconnect regression tests. Resolved **additively — both kept**: `dropping_stdio_registry_reaps_child_process`, `call_tool_reconnects_on_stale_session`, and `call_tool_does_not_retry_on_tool_error`, alongside the existing `call_tool_iserror_*` set.
- **Master's newer MCP behavior is preserved, not replaced.** The `result.isError: true` → sanitized tool-execution error handling and the per-request HTTP/SSE timeout + structured transport logging were already in this branch's base, so the rebase left them intact; the reconnect loop wraps them rather than bypassing them.
- `mcp_transport.rs` rebased clean (no conflict), as you expected.

Diff is back to just the two MCP files (+425/−96). Ready for the real reconnect review whenever you have a cycle — thanks for your patience on the churn.

crates/zeroclaw-tools/src/mcp_transport.rs
```diff
@@ -310,7 +335,14 @@ impl McpTransportConn for HttpTransport {
             .context("HTTP request to MCP server failed")?;
 
         if !resp.status().is_success() {
-            bail!("MCP server returned HTTP {}", resp.status());
+            let status = resp.status();
+            if status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE {
```
singlerider: This maps every 404/410 to `StaleSession`, but per the MCP spec that classification only holds when the request carried an `Mcp-Session-Id`. A 404 from a misconfigured `url` or wrong path now reports a stale session and burns a wasted reconnect (sleep + reset + re-handshake that 404s again). Gate it on an existing session:

```rust
if self.session_id.is_some()
    && (status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE)
{
    return Err(McpTransportError::StaleSession { status: status.as_u16() }.into());
}
```

The `http_transport_maps_404_to_stale_session` test builds a sessionless transport and asserts `StaleSession`, so it pins this wrong behavior; after the fix it should set `session_id` first, and a sibling test should assert a sessionless 404 stays a plain error with no retry.

singlerider (changes requested):
Thanks for sticking with this through the rebase thrash. The shape here is exactly right: a typed `McpTransportError` with downcast instead of string-matching, a bounded retry that never touches genuine tool errors or timeouts, and tests that pin both the recovery path and the no-retry path. I have one blocking item.

### 🔴 Blocking: HTTP 404/410 is classified as `StaleSession` even when no session exists

In `HttpTransport::send_and_recv`, the new check maps any 404/410 to `StaleSession` unconditionally. Per the MCP spec (2025-06-18, Session Management), the stale-session signal is a 404 on a request that **carried an `Mcp-Session-Id`** the server no longer recognizes: "When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it MUST start a new session." A 404 from a server that simply does not have the endpoint (typo'd `url` in config, wrong path, reverse-proxy misroute) now reports "MCP session is stale (HTTP 404)" instead of "MCP server returned HTTP 404", and `call_tool` burns a reconnect attempt: 500ms sleep, reset, re-handshake (which then 404s again and surfaces a confusing "failed to re-handshake" context chain).

Gate the classification on an existing session:

```rust
if self.session_id.is_some()
    && (status == reqwest::StatusCode::NOT_FOUND || status == reqwest::StatusCode::GONE)
{
    return Err(McpTransportError::StaleSession { status: status.as_u16() }.into());
}
```

The SSE side is fine as-is: there the POST target comes from the server's own `endpoint` event, so a 404 on it genuinely means the old session/endpoint is gone.

The existing `http_transport_maps_404_to_stale_session` test builds a transport with no session and asserts `StaleSession`, which pins the wrong behavior; after the fix it should set `session_id` first, and a sibling test should assert that a 404 with no session stays a plain error (and that `call_tool` does not retry it).

### 🟢 Typed recoverable errors, no string-matching

`McpTransportError` with `downcast_ref` is the right mechanism, and routing SSE stream teardown through dropped senders so waiters observe `RecvError` -> `TransportClosed` is much cleaner than the old synthesized `INTERNAL_ERROR` response.

### 🟢 Retry discipline is pinned by tests

`call_tool_does_not_retry_on_tool_error` asserting `initialize` runs exactly once is exactly the regression guard this needs. The wiremock priority/up_to_n_times choreography in the reconnect test is tidy.

### 🟢 Handshake dedup

Extracting `handshake()` and sharing it between `connect` and the reconnect path removes the drift risk of two hand-rolled init sequences.

---
Branch is MERGEABLE again and CI green on 504f2290e, so this is review-on-merits. One fix and a test adjustment and I am ready to approve.

itripn:
@singlerider good catch — fixed in `8dbd176`.

`HttpTransport::send_and_recv` now gates the 404/410 → `StaleSession` classification on `self.session_id.is_some()`, exactly as you suggested. A sessionless 404 (typo'd `url`, wrong path, reverse-proxy misroute) now surfaces as a plain `MCP server returned HTTP 404` error and `call_tool` no longer burns a reconnect on it. The SSE path is left untouched — its POST target comes from the server's own `endpoint` event, so a 404 there genuinely means the session/endpoint is gone.

Tests:
- `http_transport_maps_404_to_stale_session` now sets a `session_id` first, so it pins the *with-session* mapping rather than the old sessionless behavior.
- New `http_transport_404_without_session_is_plain_error` asserts a sessionless 404 does **not** downcast to `StaleSession`.
- New client-level `call_tool_does_not_retry_sessionless_404` pins the no-retry end-to-end (`initialize` is hit exactly once).

CI is green on the new head (`8dbd17679`). Re-requesting review — thanks for the precise catch.

tidux (commented):
Re-reviewing at `8dbd17679` after @singlerider's CHANGES_REQUESTED on `504f2290e`. The fix lands the gating exactly as suggested and the test plan is now correct in shape. CI is 16/16 green at this head, and `git merge-tree --write-tree` against current master (`0eb90eaaf`) reports a clean merge — so the rebase concern @Audacity88 raised on `bdcaf9ae` is also fully addressed here.

Posting as `--comment` because @singlerider's CHANGES_REQUESTED on the prior head is still the authoritative state on `reviewDecision`. Their block is resolved on the merits, but I won't approve over an active CR — pinging @singlerider for a conversion pass.

### ✅ Resolved — 404/410 → `StaleSession` is now gated on an active session id

`HttpTransport::send_and_recv` at `crates/zeroclaw-tools/src/mcp_transport.rs:339-348` now requires `self.session_id.is_some()` before mapping 404/410 to `McpTransportError::StaleSession`. A sessionless 404 (typo'd `url`, wrong path, reverse-proxy misroute) falls through to the plain `bail!("MCP server returned HTTP {}", status)` and `call_tool` never burns a reconnect on it. This matches the MCP spec language @singlerider cited and the suggested patch byte-for-byte. The inline comment in the diff captures the rationale so the next person editing this branch knows why the gate is there. SSE is left untouched, which is the right call — the SSE POST target comes from the server's own `endpoint` event, so a 404 there genuinely means the session/endpoint is gone.

### ✅ Resolved — test plan now pins the right behavior

- `http_transport_maps_404_to_stale_session` (`mcp_transport.rs:1577-1591`) sets `transport.session_id = Some("sess-1".into())` before the assertion, so it now pins the **with-session** mapping that the spec actually licenses.
- New sibling `http_transport_404_without_session_is_plain_error` (`mcp_transport.rs:1593-1632`) asserts the sessionless 404 does **not** downcast to `StaleSession` and surfaces as `"MCP server returned HTTP 404"`. The double assertion (negative `downcast_ref` + positive string contains) is the right shape — it guards both the classification and the surfaced text.
- New client-level `call_tool_does_not_retry_sessionless_404` (`mcp_client.rs:938-1000`) pins the end-to-end no-retry: `initialize` mock is `.expect(1)` and the `tools/call` 404 mock is `.expect(1)`, so `server.verify()` fails loudly if anything ever wires a reconnect onto the sessionless path. That's exactly the regression guard the no-retry contract needs.

### ✅ Resolved — rebase preserves master's MCP work, doesn't replace it

Cross-checked the two paths @Audacity88 flagged at `bdcaf9ae`:

- `result.isError: true` handling — `mcp_client.rs` still sanitizes and surfaces `isError` envelopes via the `call_tool_iserror_*` test set, which lives untouched alongside the new reconnect tests.
- Per-request HTTP/SSE timeouts on `tools/call` — `http_request_timeout_secs` / `http_sse_read_timeout_secs` at `mcp_transport.rs:35-54` are preserved and the reconnect loop runs *inside* the `tokio::time::timeout(Duration::from_secs(tool_timeout), …)` budget in `McpServer::call_tool` (`mcp_client.rs:215-219`). So the reconnect doesn't smuggle past the configured tool budget — a slow-and-flaky remote still gets the same per-call ceiling, just with at most 2 chances to land inside it.

### 🟢 Reconnect-loop discipline holds at this head

The loop in `call_tool` (`mcp_client.rs:200-285`) keeps the contract I want from a transport-recovery primitive:

- Retries only on `McpTransportError` (downcast, no string-matching), bounded at `MAX_RECONNECT_ATTEMPTS = 2` with a fixed `RECONNECT_BACKOFF_MS = 500` so a stuck server can't pin the agent loop.
- `reset() → handshake() → continue` re-runs the `initialize` + `notifications/initialized` sequence via the extracted `handshake()` helper, so the reconnect path and the initial-connect path can't drift apart.
- Genuine tool errors (JSON-RPC `error`, `result.isError`) and timeouts never enter the retry branch — the `recoverable_reason` arm only fires when `downcast_ref::<McpTransportError>()` succeeds.
- The `zeroclaw_log` event on the reconnect attempt carries `mcp_server`, `tool`, `attempt`, `max_attempts`, and the human-readable `reason`, which matches the rollback playbook in the PR body ("grep logs for `reconnecting (attempt N/2)`").

### 🟢 SSE `reset()` tears down the right state

`SseTransport::reset()` (`mcp_transport.rs:1056-1072`) signals `shutdown_tx`, `abort()`s the reader task, flips `stream_state` back to `Unknown`, and under the shared lock clears `message_url`, `message_url_from_endpoint`, and the `pending` map. That's exactly the state the next call needs cleared to re-handshake — no dangling waiters held against the dead stream, no cached endpoint that would skip the GET + `endpoint` event on reconnect.

### 🟡 Drop the `Co-Authored-By: Claude` trailer and the PR-body footer before merge

The `8dbd17679` commit message has a `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>` trailer, and the PR body still ends with `🤖 Generated with [Claude Code]`. The PR template explicitly says (`.github/pull_request_template.md`):

> **Do not add bot/AI attribution footers** such as `Co-authored-by: Claude ...` or `Created with Claude Code` to the PR body or commit-message tail. Human co-author trailers are appropriate only for incorporated contributor work under the supersede-attribution section and privacy contract.

Not blocking from me — this is the kind of thing that gets fixed on the way to squash-merge — but worth scrubbing before the maintainer pulls the trigger so the merged commit message and the merged PR body are clean.

### 🔵 Stale `needs-author-action` label

The `needs-author-action` label is still on the PR even though the only outstanding block (@singlerider's 🔴) is addressed at `8dbd17679`. Label management is a maintainer call — flagging it so whoever has label permissions can clear it after @singlerider's conversion pass.

---

**Verification at `8dbd17679` (local):**

- `cargo fmt -p zeroclaw-tools -- --check` — clean.
- `cargo clippy -p zeroclaw-tools --lib --all-targets -- -D warnings` — clean, `Finished` in 39s.
- `cargo test -p zeroclaw-tools --lib -- mcp` — `104 passed; 0 failed`. Includes the carried-forward `call_tool_reconnects_on_stale_session`, `call_tool_does_not_retry_on_tool_error`, the adjusted `http_transport_maps_404_to_stale_session`, and both new tests (`http_transport_404_without_session_is_plain_error`, `call_tool_does_not_retry_sessionless_404`).
- `git merge-tree --write-tree pr-7351 upstream/master` (current master `0eb90eaaf`) — no conflict markers in the resulting tree.

Once @singlerider converts their CHANGES_REQUESTED I am comfortable converting this `--comment` → `--approve` on a subsequent pass.

Audacity88 (commented):
Quick follow-up to close the loop on my earlier conflict-refresh comment: at current head `8dbd17679`, the MCP test-tail conflict I called out looks resolved additively, with both the stdio child-cleanup regression and this PR's reconnect regression coverage preserved.

I also agree with @tidux's current-head read that @singlerider's sessionless 404/410 blocker appears fixed on the merits: stale-session classification is now gated on an existing session id, and the new client-level coverage pins the no-retry path for a sessionless 404.

Posting this as a comment only, not an approval, because @singlerider's `CHANGES_REQUESTED` review is still the active blocker and should be cleared by that reviewer.

singlerider (approved):
Re-reviewed at head `8dbd17679`. My one blocker is resolved exactly as suggested, the test plan is corrected, and @tidux independently verified the same. Converting my changes-requested to an approval.

### RESOLVED ✅ — 404/410 → `StaleSession` is now gated on an active session id

`HttpTransport::send_and_recv` (`mcp_transport.rs:344-348`) now requires `self.session_id.is_some()` before mapping 404/410 to `McpTransportError::StaleSession`, with the MCP-spec rationale inline. A sessionless 404 — typo'd `url`, wrong path, reverse-proxy misroute — falls through to the plain `MCP server returned HTTP {status}` error and `call_tool` never burns a reconnect on it. SSE is correctly left untouched, since its POST target comes from the server's own `endpoint` event.

### RESOLVED ✅ — the tests now pin the right behavior

- `http_transport_maps_404_to_stale_session` sets `session_id = Some("sess-1")` first, so it pins the with-session mapping the spec actually licenses (not the previous sessionless-maps-to-stale shape).
- New `http_transport_404_without_session_is_plain_error` (line 1594) asserts a sessionless 404 does not downcast to `StaleSession` and surfaces as `"MCP server returned HTTP 404"` — guarding both classification and text.
- New `call_tool_does_not_retry_sessionless_404` (`mcp_client.rs:939`) pins the end-to-end no-retry with `.expect(1)` on both the `initialize` and `tools/call` mocks, so any future wiring of a reconnect onto the sessionless path fails `server.verify()` loudly.

### 🟢 The rest of the recovery primitive holds

Confirming what I praised before is intact: typed `McpTransportError` with `downcast_ref` (no string-matching), bounded `MAX_RECONNECT_ATTEMPTS = 2` with fixed backoff, the reconnect running inside the configured per-call `tokio::time::timeout` budget (so it can't smuggle past the tool ceiling), genuine tool errors / `result.isError` / timeouts never entering the retry branch, the shared `handshake()` removing init-sequence drift, and `SseTransport::reset()` tearing down exactly the right state (shutdown signal, reader abort, cleared endpoint/pending) so no dangling waiters survive. The reconnect log event carries `mcp_server`/`tool`/`attempt`/`max_attempts`/`reason` matching the rollback playbook. CI is 16/16 green and the branch is `MERGEABLE`.

### 🟡 Scrub the bot attribution before the squash

The `8dbd17679` commit carries a `Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>` trailer and the PR body ends with a "Generated with Claude Code" footer. Strip both from the squash commit message and PR body on the way in, per the template. Not gating my approval — it's squash-time cleanup.

Approving. This clears my outstanding review; nice work landing the gate and the no-retry regression guards.

perlowja (approved):
Approve. Reviewed the reconnect path: typed `McpTransportError` (StaleSession/TransportClosed) via downcast, bounded retry loop with fresh request id per attempt + backoff, `reset()` tears down the SSE reader task and clears endpoint/pending before re-handshake, and the 404→StaleSession classification is correctly gated on `session_id.is_some()` (the sessionless-404 case stays a plain error). Strip/teardown applied at outbound replay only; inbound parsing untouched. CI green, mergeable, and singlerider's items are addressed at the current head.

PR #3637, opened 2026-03-15
feat(tool): add SecretStore integration to http_request

5queezer:
**Why this exists**

When the LLM calls an external API via `http_request`, it has to put the API key in plaintext in the tool call arguments. That key then sits in conversation history, prompt context, and logs. For a project that takes security seriously, that's a leak vector hiding in plain sight.

---

## Summary

- **Problem:** When the LLM calls external APIs via `http_request`, it must pass API keys as plaintext in tool call arguments (`"headers": {"Authorization": "Bearer sk-..."}`). The key then appears in the conversation history, prompt context, and logs — a credential leak vector.
- **Solution:** Add `auth_secret` parameter that references a named entry from `[http_request.secrets]` in `config.toml`. The secret is resolved at execution time via `SecretStore`, so the API key never enters the LLM context.
- Config-path modeled as `Option<PathBuf>` — legacy `new()` constructor returns a clear error if `auth_secret` is used
- Secret names are validated (alphanumeric + `_`/`-`, 1–64 chars) to prevent injection

## Manual Testing

1. Add secrets to `config.toml`:
```toml
[http_request]
enabled = true
allowed_domains = ["httpbin.org"]

[http_request.secrets]
test_token = "Bearer my-test-value"
```

2. In chat, ask the agent:
> Make a GET request to https://httpbin.org/headers with auth_secret "test_token"

3. Verify:
   - httpbin response shows `"Authorization": "Bearer my-test-value"`
   - The tool call arguments show only `"auth_secret": "test_token"`, **not** the actual key

4. Error cases:
   - `auth_secret: "nicht_vorhanden"` → "not found in [http_request.secrets]"
   - `auth_secret: "../traversal"` → "alphanumeric"
   - `auth_secret: ""` → "1-64 characters"

## Label Snapshot (required)

- Risk label: `risk: high`
- Scope labels: `tools`, `security`
- Module labels: `tools: http_request`, `config: schema`

## Change Metadata

- Change type: `feature`
- Primary scope: `tools`

## Validation Evidence

```
cargo fmt --all -- --check   ✅
cargo clippy --lib -- -D warnings   ✅
cargo test --lib tools::http_request::tests   ✅ 67 passed
cargo test --lib http_request_config   ✅ 1 passed
```

## Security Impact

- New permissions/capabilities? No — reuses existing SecretStore infra
- New external network calls? No
- Secrets/tokens handling changed? Yes — secrets are now resolved from config at execution time rather than passed in plaintext
[...]

PR #7354, opened 2026-06-07
feat(tools): add http_request auth secrets

Audacity88:
## Summary

- **What changed and why:**
  - Adds `[http_request.secrets]` config entries and a new `auth_secret` argument for the `http_request` tool, so agents can reference a named credential without putting the secret value in tool-call arguments.
  - Resolves the named secret from the current config at execution time, including encrypted config values, so config remains the source of truth.
  - Marks dynamic `http_request.secrets.<name>` paths as secret fields across the config schema, config macro, RPC config-set path, and gateway config metadata.
  - Adds focused regression coverage for plain and encrypted secret resolution, invalid names, Authorization header override behavior, and dynamic secret-path masking.
- **Scope boundary:** This does not add a new secret backend, new HTTP transport, new outbound permission model, or new domain allowlist behavior. Existing literal `headers.Authorization` calls still work. When `auth_secret` is provided, it supplies the Authorization header and overrides any literal header for that request.
- **Blast radius:** High-risk tool/config/runtime/gateway secret-handling surface: `zeroclaw-tools` `http_request`, config schema generation, runtime tool registration and config RPC, and gateway config metadata.
- **Linked issue(s):** Related #6074. Supersedes #3637.
- **Labels:** `enhancement`, `risk: high`, `size: L`, `config`, `gateway`, `runtime`, `tool`, `security:secrets`, `tool:http_request`, `tool:web`

  - `git diff --check origin/master...HEAD` — passed.
  - `cargo fmt --all -- --check` — passed.
  - `cargo test -p zeroclaw-config http_request -- --nocapture` — passed; 5 tests passed.
  - `cargo test -p zeroclaw-config credential_shaped_prop_fields_have_explicit_classification -- --nocapture` — passed.
  - `cargo test -p zeroclaw-tools http_request::tests -- --nocapture` — passed; 83 tests passed.
  - `cargo test -p zeroclaw-runtime config_set_handles_dynamic_http_request_secret_paths -- --nocapture` — passed.
  - `cargo test -p zeroclaw-runtime all_tools_ -- --nocapture` — passed.
  - `cargo test -p zeroclaw-gateway lookup_prop_field_synthesizes_dynamic_http_request_secret_metadata -- --nocapture` — passed.
  - `cargo clippy -p zeroclaw-config -p zeroclaw-macros -p zeroclaw-tools -p zeroclaw-runtime --all-targets -- -D warnings` — passed.
  - `cargo clippy -p zeroclaw-gateway --lib --no-deps -- -D warnings` — passed. Full dependency lint for the gateway still surfaces an unrelated pre-existing `zeroclaw-channels` unused-import warning in `crates/zeroclaw-channels/src/orchestrator/mod.rs`.
  - `cargo clippy --workspace --all-targets --features ci-all -- -D warnings` — passed; tail included `Finished dev profile [unoptimized + debuginfo] target(s) in 19m 35s`.
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` — passed; tail included `8140 tests run: 8140 passed, 10 skipped`.
  - GitHub CI for this PR head is green, including format, lint, test, build, security, and required gate jobs.
- **Beyond CI — what did you manually verify?** Reviewed the execution path from runtime tool registration through `HttpRequestTool`, confirmed `auth_secret` reloads the current config instead of caching a boot-time copy, and checked that dynamic secret config paths are treated as secret before and after the key exists.
- **If any command was intentionally skipped, why:** The repository template's plain `cargo test` command was not run separately because the CI-shaped `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` completed successfully across the workspace, and GitHub CI is green for this PR head.

- If any `Yes`, describe the risk and mitigation: This adds a new `http_request` capability for naming a configured auth secret and using it as the request Authorization header. The capability stays constrained by existing `http_request` enablement and domain controls, validates secret names, keeps config values under encrypted storage when encryption is enabled, masks dynamic secret paths through config/RPC/gateway metadata, and resolves the current config at execution time rather than caching duplicate secret state.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: Existing `http_request` calls continue to work. To use the new safer path, add a named secret under `[http_request.secrets]`, for example `api_token = "Bearer ..."` or its encrypted equivalent through the config API, then call `http_request` with `auth_secret = "api_token"` instead of putting `Authorization` in the tool-call headers.

- **Feature flags or config toggles:** Do not set `auth_secret`, remove `[http_request.secrets]`, or disable the tool with `[http_request].enabled = false`.
- **Observable failure symptoms:** `http_request` calls that use `auth_secret` fail with secret lookup/decryption errors, send a missing or wrong Authorization header, or config/RPC/gateway surfaces fail to mask or accept `http_request.secrets.<name>` paths.

  - #3637 by @5queezer
- Scope materially carried forward: The user-facing design from #3637 is carried forward: `auth_secret`, `[http_request.secrets]`, secret-name validation, and execution-time secret resolution for `http_request`. The implementation was rebuilt for the current crate layout and current config/RPC/gateway secret metadata.


crates/zeroclaw-macros/src/lib.rs
```diff
@@ -331,6 +335,74 @@ pub fn derive_configurable(input: TokenStream) -> TokenStream {
...
+                    quote! { self.#field_ident.as_ref() }
+                } else {
+                    quote! { Some(&self.#field_ident) }
+                };
+                let map_mut_expr = if is_option {
+                    quote! { self.#field_ident.get_or_insert_with(std::collections::HashMap::new) }
+                } else {
+                    quote! { &mut self.#field_ident }
+                };
+                dynamic_secret_map_prop_fields.push(quote! {
```
singlerider, line 351: 🟢 Solving the dynamic-secret-map masking generically in the derive — rather than special-casing `http_request.secrets` — is the right call. Any `HashMap<String,String>` carrying `#[secret]` now gets per-entry `****` display, a `get_prop` that never returns the value, dynamic-key `set_prop`, and `prop_is_secret` for `<container>.<key>` automatically. The next dynamic secret map inherits all of it; this is registry-level, not a shadow list. Combined with the gateway `lookup_prop_field` fallback and the RPC masked-clobber guard both keying off `prop_is_secret`, every read/write surface stays consistent by construction.

singlerider (approved):
Read all six files, traced the secret from config through resolution to the request, and checked every surface that could leak it (tool args, tool output, logs, get_prop, gateway metadata, RPC config-set). This is a careful, well-engineered secrets feature. Approving.

### 🟢 What looks good — the secret value never crosses a surface it shouldn't

The whole design hinges on `auth_secret` carrying a *name*, not a value, and that holds end to end. The name is validated (non-empty, ≤64, `[A-Za-z0-9_-]` only) before any lookup, so it can't traverse paths or inject into the resolver. `reload_auth_secret` reads the current `config.toml` from disk at execution time rather than caching a boot-time copy, so config stays the source of truth and a rotated secret takes effect immediately. The resolved value goes straight into the `reqwest` Authorization header via `HeaderValue::from_str` (which rejects a malformed secret before it can corrupt the request), and it is never echoed: the `record!` events attach only `url`/`method`/status, never headers, and the error strings quote the secret *name* (`auth_secret '{name}'`) but never the value. I looked specifically for a leak path into tool output or logs and there isn't one.

### 🟢 What looks good — the dynamic-secret-map masking is generic, not special-cased

This is the part I'd have most expected to be hacked in for `http_request.secrets` specifically, and instead it's solved at the right layer. The `derive_configurable` macro grew a generic `is_hashmap_string_string` + `#[secret]` path that synthesizes, for *any* such field: per-entry `PropFieldInfo` with `display_value: "****"` / `is_secret: true`, a `get_prop` that returns `"****"` for any `<container>.<key>`, a `set_prop` that accepts dynamic keys, and a `prop_is_secret` that returns true for any `<container>.<key>`. So the masking lives in the registry-level derive, and the next dynamic secret map gets it for free — exactly "extend the registry, don't shadow it." The field itself carries `#[secret]` + `#[credential_class = "encrypted_secret"]` + the `x-secret` schema export, so it's classified through the canonical credential system, and `credential_shaped_prop_fields_have_explicit_classification` keeps that honest.

### 🟢 What looks good — every read/write surface is consistently guarded, with tests

The three downstream surfaces line up: the gateway `lookup_prop_field` falls back to `prop_is_secret(path)` to synthesize masked metadata for the dynamic path; the RPC `handle_config_set` adds `prop_is_secret(&req.prop)` to the masked-value-clobber guard so a `"****"`/empty round-trip from the config UI can't silently overwrite the real secret; and the schema `validate()` rejects bad secret-key names at load time. Each has a focused test — including `config_set_handles_dynamic_http_request_secret_paths` proving both the real-set accept and the masked/empty reject, and the full encryption round-trip in the schema tests proving `SecretStore` encrypt/decrypt of a stored secret. The name validation existing in both the tool (`validate_secret_name`) and config (`validate()`) is defense-in-depth across the execution-time and load-time boundaries, not redundant shadow logic.

### 🟢 What looks good — backward compatible, override semantics documented

Existing literal `headers.Authorization` calls keep working; `auth_secret` is purely additive and, when present, overrides the literal Authorization for that request — stated plainly in the schema description and the scope boundary. Supersede attribution for #3637 by @5queezer is present with the co-author trailer, and the implementation is rebuilt against the current crate/config/RPC/gateway layout rather than dropped in.

This is the right way to add a credential surface: the value is contained to the request, masked everywhere it could be read back, classified through the registry, and resolved fresh at execution. Approving.

PR #7358, opened 2026-06-07
Update 'quickstart' to 'onboard'

devqcf:
Just docs one word change

### DIFF
--- a/docs/book/src/setup/macos.md
+++ b/docs/book/src/setup/macos.md
@@ -24,7 +24,7 @@ cd zeroclaw
 
 1. Asks whether you want a prebuilt binary or to build from source
 2. Installs to `~/.cargo/bin/zeroclaw`
-3. Runs `zeroclaw quickstart` to complete first-time setup
+3. Runs `zeroclaw onboard` to complete first-time setup
 
 Flags:
 
@@ -33,7 +33,7 @@ Flags:
 ./install.sh --source                        # always build from source
 ./install.sh --minimal                       # kernel only (~6.6 MB)
 ./install.sh --source --features agent-runtime,channel-discord   # custom features
-./install.sh --skip-quickstart                  # install only; run `zeroclaw quickstart` later
+./install.sh --skip-onboard                  # install only; run `zeroclaw onboard` later
 ./install.sh --list-features                 # print available features and exit
 ./install.sh --help                          # full flag reference
 ```
@@ -42,7 +42,7 @@ Flags:
 
 ```bash
 brew install zeroclaw
-zeroclaw quickstart
+zeroclaw onboard
 ```
 
 Gets you `brew services` integration. Binary lives at `$HOMEBREW_PREFIX/bin/zeroclaw`.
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
Reviewed current head `2a68114` against the one-file docs diff, the installer flags, the CLI deprecation text, and the current required check failures. This PR only changes `docs/book/src/setup/macos.md`.

### 🔴 Blocking — The docs now point users at the deprecated command and a non-existent installer flag

This change replaces `zeroclaw quickstart` with `zeroclaw onboard` and `--skip-quickstart` with `--skip-onboard`, but current source shows the opposite direction. `install.sh` still documents `--skip-quickstart`, runs `zeroclaw quickstart` during setup, and tells users to run `zeroclaw quickstart` later. The CLI also marks `zeroclaw onboard` as deprecated and tells users to use `zeroclaw quickstart`.

As written, the macOS setup page would send users to the deprecated setup command and to an installer flag that does not exist. Please revert these replacements, or update the source/installer first if the intended product direction has changed.

### 🟡 Warning — Required checks are red

The required gate is currently blocked. `CI Required Gate`, `main`, `Lint`, and `Test` are failing in the captured check state, so those need to be green before this can merge.

singlerider:
@devqcf Onboard is deprecated. Recommend closure for this PR.

singlerider (commented):
Reviewed at head `2a68114` against current `master`. The rename is backwards relative to the actual product direction — confirming @Audacity88's blocker in source. Agreeing with his changes-requested.

### 🔴 Blocking — confirmed: this points users at the deprecated command and a flag that does not exist

I checked the current tree. The canonical direction is `onboard` → `quickstart`, the opposite of this PR:

- `install.sh` documents and parses `--skip-quickstart` and runs `zeroclaw quickstart` post-install — there is no `--skip-onboard` flag, so `./install.sh --skip-onboard` would simply be unrecognized.
- `src/main.rs:404` marks the `onboard` subcommand `/// Deprecated. Use \`zeroclaw quickstart\`. Any flags error.`, and `main.rs:3044-3045` makes `zeroclaw onboard` print `error: \`zeroclaw onboard\` is deprecated ... Use \`zeroclaw quickstart\``.

So this PR rewrites the macOS setup page to send users to the deprecated `zeroclaw onboard` and a non-existent `--skip-onboard` installer flag. That's the inverse of what the code does. Please revert these `quickstart` → `onboard` replacements. If the product direction is genuinely meant to flip to `onboard`, that has to land in `install.sh` and the CLI deprecation first, and the docs follow — not lead with a rename the binary contradicts.

### 🟡 Agree — required checks are red

`CI Required Gate` / `main` / `Lint` / `Test` were failing in the captured state; the gate needs to be green before merge regardless.

Revert the rename (the docs already match the real `quickstart` direction). Leaving @Audacity88's changes-requested in force.

Audacity88:
@devqcf thanks for the docs cleanup attempt here.

Closing this because the PR is still pointing the macOS setup page in the wrong command direction. Current ZeroClaw uses `zeroclaw quickstart`; `zeroclaw onboard` is deprecated, and the installer flag is still `--skip-quickstart`, not `--skip-onboard`.

Since the current docs already match the live command direction and @singlerider also recommended closure, this PR does not have a remaining merge path as written. If there is a different onboarding-docs cleanup you want to make against the current `quickstart` flow, please open a fresh PR with that narrower scope.

PR #7357, opened 2026-06-07
fix(channels): update image history fixture context field

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Update the #7268 channel image-history regression test fixture to use the current `ChannelRuntimeContext::model_provider_ref` field.
  - Remove stale `api_key` and `api_url` fixture fields that no longer exist after the model-provider routing cleanup on `master`.
  - Restore `zeroclaw-channels` test-target compilation on current `master`.
- **Scope boundary:** This only fixes the test fixture introduced by #7268. It does not change runtime image-history behavior, provider routing, media handling, or channel configuration.
- **Blast radius:** `zeroclaw-channels` test compilation only.
- **Linked issue(s):** Follow-up to #7268.
- **Labels:** `bug`, `risk: low`, `size: XS`, `channel`.

## Validation Evidence (required)

Local validation is the signal CI cannot replace. Run the full battery and paste literal output (tails, failures, warnings — not "all passed").

- **Commands run and tail output:**

`git diff --check` passed with no output.

`cargo fmt --all -- --check` passed with no output.

`scripts/zc-cargo check -p zeroclaw-channels --tests` tail:

```text
    Checking zeroclaw-runtime v0.8.0-beta-2
    Checking zeroclaw-channels v0.8.0-beta-2
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 9m 41s
```

`scripts/zc-cargo test -p zeroclaw-channels --lib process_channel_message_sends_image_payload_without_persisting_it -- --nocapture` tail:

```text
    Finished `test` profile [unoptimized + debuginfo] target(s) in 10m 53s
     Running unittests src/lib.rs

running 1 test
test orchestrator::tests::process_channel_message_sends_image_payload_without_persisting_it ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 709 filtered out; finished in 1.41s
```

- **Beyond CI — what did you manually verify?**
  - Verified the compile failure from current `master` was caused by the #7268 test fixture still using removed `ChannelRuntimeContext` fields: `default_model_provider`, `api_key`, and `api_url`.
  - Verified neighboring fixtures now use `model_provider_ref`, and this patch brings the new #7268 fixture into the same shape.
- **If any command was intentionally skipped, why:**
  - Full workspace CI was not run locally because this is a one-fixture compile fix and the targeted check covers the failing package/test target that broke PR CI.

## Security & Privacy Impact (required)

[...]

PR #7330, opened 2026-06-06
fix(zerocode): repair Quickstart model-provider form UX defects

singlerider:
> **DO NOT MERGE - AUTHOR WILL MERGE WHEN READY. REVIEWS STILL REQUESTED**

## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:** This started as a fix for six zerocode TUI
  Quickstart Model-provider form UX defects and, through follow-up
  commits, grew to also standardize the Quickstart runtime-profile
  choice and unify the Quickstart documentation across surfaces. The
  landed change has four concern areas:

  **1. zerocode Quickstart Model-provider form (original #7304 fix).**
  The form under-used the rich field metadata the daemon already
  supplies (`is_secret`, `kind: bool`, `enum_variants`, structured
  `step`/`field`/`message` errors), leaving six UX defects:
  - Bounds the secret mask (cap + `(+N)` remainder) so a pasted
    realistic-length API key can no longer wrap across rows and push
    later fields and the footer out of view.
  - Surfaces the first actionable field error
    (`<selector> / <field>: <message>`, `(+N more)` when several)
    instead of only `N error(s) — fix selectors and resubmit`.
  - Loads the model catalog lazily via a per-frame `tick`: the form
    opens immediately in a `Pending` state painting a
    `Loading models for <provider>…` row, then resolves to an enum
    picker or a free-text fallback with a visible
    `catalog unavailable — type a model name` hint.
  - Renders bool fields as `‹ false › / ‹ true ›` toggles by seeding the
    `kind: bool` metadata with a two-value enum, instead of raw text.
  - Fixes the alias ghost-default state so the descriptor default shows
    as ghost text only on unfocused empty rows.

  **2. Runtime-profile picker removal (force `unbounded`).** The
  Quickstart runtime-profile picker is dropped from all three surfaces
  (CLI onboard, zerocode TUI, web), and the runtime apply path now
  forces `unbounded` in one named place (`crates/zeroclaw-runtime/src/quickstart/mod.rs`).
  The old wire field stays populated to avoid a partial surface
  migration; only the actual choice is centralized. Related #7371
  (the preset-must-allow-what-it-advertises fix that makes `unbounded`
  safe as the forced default).

  **3. Quickstart documentation unification.** Getting-started docs are
  reorganized into shared `_snippets/` (concept blurbs + Quickstart
  steps), `quick-start.md` is replaced by `quickstart.md`, and the
  setup pages / SUMMARY / introduction links are repointed.

[...]

PR #7226, opened 2026-06-04
fix(channels/whatsapp-web): treat replies to bot as mentions

rifuki:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Treat plain-text WhatsApp Web group replies to a prior bot message as addressed to the bot when `mention_only = true`.
  - Read the quoted message participant from extended-text `context_info` and match it against both the bot phone JID and LID JID.
  - Preserve existing explicit mention behavior and reject replies to other participants without a bot mention.
- **Scope boundary:** This does not change sticker/media handling, quoted-media downloads, mention-pattern gating, WhatsApp Cloud API behavior, read receipts, or group session/history scope.
- **Blast radius:** WhatsApp Web inbound group-message gating when `mention_only = true`.
- **Linked issue(s):** Closes #7225.
- **Labels:** `bug`, `risk: medium`, `size: S`, `channel`, `channel:whatsapp`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
cargo fmt --all -- --check
result: passed locally; GitHub Actions Format passed on current head.

cargo clippy --all-targets -- -D warnings
result: passed locally; GitHub Actions Lint passed on current head.

cargo test
result: passed locally.

cargo test -p zeroclaw-channels --features whatsapp-web
result: passed locally; includes reply-to-phone-JID, reply-to-LID-JID, reject-other-participant, and explicit-mention fallback coverage.

GitHub Actions Quality Gate
result: passed on current head d31e287cccc477b9f67683c6cab14b46ccbce7b1, including Test and CI Required Gate, after #7231 restored compiling master.
```

- **Beyond CI — what did you manually verify?** Verified focused unit coverage for phone-JID replies, LID-JID replies, replies to other participants, and explicit mention fallback. Reviewed the source path so `mention_only` group admission expands only when the quoted participant matches the bot phone or LID identity. Did not run a live WhatsApp Web session from this branch.
- **If any command was intentionally skipped, why:** No listed command was skipped. Live WhatsApp Web smoke was not run; validation is unit-level plus green hosted CI on the current head.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? (`Yes`) This expands `mention_only` admission only for group replies whose quoted participant JID matches the bot phone JID or LID JID. Replies to other participants remain gated and are covered by a regression test.
- New external network calls? (`No`)
- Secrets / tokens / credentials handling changed? (`No`)
[...]

PR #7360, opened 2026-06-07
fix(zerocode): size quickstart modals by wrapped row height

### LINKED ISSUE #7304: [Bug]: zerocode Quickstart model-provider form hides field errors and overflows secret input
### Affected component

tooling/ci

### Severity

S2 - degraded behavior

### Current behavior

The zerocode TUI Quickstart Model provider form still has several UX defects that make the form hard to complete or recover from when validation fails.

Observed while dogfooding the current zerocode Quickstart stack around #7240, #7249, and #7298 in a cmux terminal:

- Pasting a normal API key into `api_key` renders one bullet per character. The masked value wraps across multiple rows and can push following fields and the footer out of view.
- Validation errors only surface as `N error(s) — fix selectors and resubmit`. The selector can look checked or locally complete, but the UI does not name `Model provider / alias`, `Model provider / model`, or whichever field is actually invalid.
- Selecting a model provider can feel frozen while model choices load. The picker closes and the form waits for model-catalog loading without a visible loading state.
- The `model` row can degrade to raw text entry instead of a picker when the catalog is unavailable, without an obvious loading, empty, or error state.
- Boolean fields such as `requires_openai_auth` render as free-text entry even though the runtime metadata marks them as bool fields.
- With #7240's editable Alias row, deleting `default` exposes a confusing ghost-default state: once the real buffer is empty, the row renders `default` again as ghost text. Backspace cannot delete it because it is not actually in the buffer, then it disappears as soon as the user types a new alias.

Related: #7227 tracked the original hardcoded-alias dead end. #7240 addresses that specific missing-control problem. This issue tracks the remaining form-state, validation-message, and field-rendering problems in the same Quickstart Model provider flow.

### Expected behavior

- Secret fields use a fixed-width mask, clipped placeholder, or otherwise bounded display so a pasted API key cannot collapse the form layout.
- Validation surfaces the first actionable field error, for example `Model provider / alias: alias openai.default already exists`, instead of only showing an error count.
- Errored selectors are visibly distinguishable from selectors that are locally complete.
- Model catalog loading has a visible state such as `Loading models for openai...`, with a clear fallback if the catalog is empty or unavailable.
- Bool fields render as toggles/selectors, for example `‹ false › / ‹ true ›`, not raw text inputs.
- Alias display state stays aligned with validation state. If the alias buffer is empty while focused, the row should look empty; `default` should only be restored on cancel or commit if that is the intended fallback.

### Steps to reproduce

```bash
# Use a zerocode build that includes the Quickstart Model provider form.
# Optional: use a config with an existing provider so alias/model validation is easier to hit.

zerocode
```

In the TUI:

1. Open the Quickstart pane.
2. Open `Model provider`.
3. Pick a provider type such as `openai` or `openrouter`.
4. Paste a realistic-length API key into `api_key`.
5. Observe that the masked value can wrap across multiple rows and push lower fields/footer content out of view.
6. Leave a required field invalid, or choose a state that triggers validation.
7. Observe that the status strip only reports `N error(s) — fix selectors and resubmit`.
8. If an Alias row is available, Backspace through `default`.
9. Observe that `default` reappears as ghost text, cannot be Backspaced away, and disappears only after typing a replacement alias.
10. Focus bool-like fields such as `requires_openai_auth`.
11. Observe that they require raw text entry instead of a toggle or enum-style selector.

### Impact

Affected users: users completing zerocode Quickstart through the terminal UI.

Frequency: reproducible in the Model provider form when using realistic API-key values, invalid field states, empty/unavailable model catalogs, or bool provider fields.

Consequence: users cannot tell which field is invalid, may lose visibility of fields after entering secrets, and can mistake ghost/default display text for real editable text. This makes the Quickstart path feel unreliable even when the underlying config operation may be recoverable.

### Logs / stack traces

```text
status strip:
  N error(s) — fix selectors and resubmit
```

No secrets or raw API keys are included here.

### ZeroClaw version

v0.8.0-beta-2 dogfood builds around #7240, #7249, and #7298. Pre-existing parts of the behavior were also checked against current zerocode source; the Alias ghost state depends on #7240's editable alias row.

### Rust version

Not applicable; this is a TUI behavior bug observed in a built zerocode binary.

### Operating system

macOS terminal, observed in cmux.

### Regression?

Unknown

### Pre-flight checks

- [x] I reproduced this on the current zerocode dogfood stack and checked the pre-existing pieces against current source.

### LINKED ISSUE #7359: [Bug]: Quickstart modal clips wrapped rows — YOLO risk option and model picker fall off-screen
### Affected component

tooling/ci

_(The defect is in the `zerocode` interactive TUI — the Quickstart pane. There is no dedicated "tui/ui" component option, so filing under tooling.)_

### Severity

S2 - degraded behavior

### Current behavior

The Quickstart pane's modals size their box from the **logical line count** of their content, ignoring soft-wrapping. When a row wraps across multiple terminal rows, the box stays too short and later rows fall outside the viewport with no way to scroll to them. Two reproducible cases at a standard 80x31 terminal:

1. **Risk profile picker** — each option renders its help blurb inline on the same line (e.g. `Locked Down  Tight defaults. Workspace-only fs, approval on med/high risk.`). Those blurbs wrap to two rows each, so the 3-option list needs ~6 rows but the box is sized for 3. Only **Locked Down** and **Balanced** are visible; **YOLO** is cut off entirely and cannot be selected.

2. **Model provider field form** — after choosing a cloud provider (e.g. `anthropic`) and **pasting a long API key** into the `api_key` field, the secret renders as one bullet (`•`) per character and wraps across several rows. That pushes the remaining field rows (the `model` picker, `url`, `wire_api`, …) out of the modal, so they aren't shown and can't be reached.

Suspected root cause: in `apps/zerocode/src/quickstart_pane.rs::draw_modal`, `box_h` is computed from `body_lines.len()` (logical lines) and the scroll offset is computed from logical line indices, neither of which accounts for `Wrap { trim: false }`.

### Expected behavior

The modal should account for wrapped content: grow to fit it (up to the available pane height) and, when content still exceeds the viewport, keep the focused row visible by scrolling. All risk-profile options (including YOLO) and all model-provider fields should be reachable regardless of pasted-value length or terminal size.

### Steps to reproduce

```bash
# In a standard ~80x31 terminal:
zerocode                     # launch the interactive TUI
# 1. Open the "Quickstart" tab.

# Case 1 — Risk profile:
# 2. Move to "Risk profile" and press Enter.
#    -> Only "Locked Down" and "Balanced" show; "YOLO" is cut off.

# Case 2 — Model provider:
# 3. Move to "Model provider", press Enter, pick a cloud provider (anthropic).
# 4. Focus the api_key field and paste a long key, e.g.
#    sk-ant-xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
#    -> The api_key bullets wrap and push the `model` picker and later
#       fields out of the modal; they are no longer visible.
```

### Impact

Affected users: anyone running the Quickstart flow in a normally-sized terminal (the documented first-run path).
Frequency: always, for the two flows above.
Consequence: the YOLO risk profile is unselectable, and after pasting an API key the model picker becomes unreachable — both block completing first-time setup from the TUI.

### Logs / stack traces

_No response_

### ZeroClaw version

b0c6ab866 (v0.8.0-beta-2)

### Rust version

### Operating system

macOS 26.3 (Darwin 25.3.0, arm64)

### Regression?

No, first-time setup

### Pre-flight checks

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:** The Quickstart pane (`apps/zerocode/src/quickstart_pane.rs::draw_modal`) sized its modal box from `body_lines.len()` (logical line count) and computed the cursor-tracking scroll from logical line indices, ignoring the `Wrap { trim: false }` the body renders with. Any row that soft-wraps across several terminal rows then pushes later rows off-screen with no way to reach them. The headline symptom (#7359): the **risk-profile picker** renders each option's help blurb inline, those blurbs wrap to two rows, so at a normal terminal size the third option — **`yolo`** — falls off the bottom and can't be selected. The same root cause bites any wrapping field: e.g. a long pasted **`api_key`** (one bullet per char) pushes the `model` picker and later provider fields out of the field-form modal.
- This PR measures the **real wrapped height** via `Paragraph::line_count` (the same word-wrap the body uses), sizes the box from that, and tracks the focused row in wrapped-row space so it stays visible — scrolling only when even the grown box can't fit. Cursor hit-test rects now span each row's full wrapped height so a click on a wrapped continuation row still resolves to the right row.
- **Scope boundary:** Layout/geometry only. No change to modal content, selector logic, validation, RPC, or what gets submitted. The picker still shows help inline; this only makes the box tall/scrollable enough to show it.
- **Blast radius:** Confined to `draw_modal`. Every modal kind (Picker, FieldForm, TextInput, ChannelList, PeerGroupList, Agent) flows through it; for non-wrapping content the geometry is unchanged (logical count == wrapped count). The `unstable-rendered-line-info` ratatui feature `line_count` needs was already enabled and already used in `chat.rs`.
- **Linked issue(s):** Closes #7359. Related #7330 — that PR (Closes #7304) independently caps the secret mask at 24 bullets, which *mitigates the api_key trigger* but does not fix the box geometry; the risk-picker `yolo` case and every other wrapping field are unaffected by it. The two PRs touch `draw_modal` in different regions (this one the post-match sizing/scroll math, #7330 the in-match secret rendering); a test-merge shows the production code auto-merges cleanly, with only a trivial "keep both" collision where both append new `#[test]` fns.
- **Labels:** `risk: low`, `size: S`

```bash
cargo test -p zerocode
```

`cargo fmt -p zerocode -- --check` — clean, no diff (exit 0).

`cargo clippy -p zerocode --all-targets -- -D warnings`:
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 18.77s
```

`cargo test -p zerocode`:
```
     Running unittests src/lib.rs
test result: ok. 40 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
     Running unittests src/main.rs
test result: ok. 320 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

- **Beyond CI — what did you manually verify?** Added two headless `TestBackend` render tests that drive the real `draw_modal` path and assert the integrated geometry, addressing the reviewers' note that only the measurement helpers were covered: `picker_keeps_every_option_visible_when_blurbs_wrap` (every risk option — including `yolo` — gets a non-zero hit-rect, spaced by wrapped height not logical lines) and `picker_scrolls_to_keep_selected_option_visible` (when content overflows, the selected row scrolls into view while a row scrolled off the top collapses to a zero rect — the chain the helper tests don't reach). Both fail on the pre-fix logic. The measurement helpers remain pinned by `wrapped_total_counts_soft_wrapped_rows` / `wrapped_row_heights_are_measured_per_line`. Not verified: live rendering on a physical terminal — the geometry is validated against ratatui's own `line_count` and `TestBackend`.
- **Beyond CI — scope note:** The diff is isolated to the `zerocode` crate, so the clippy/test battery was scoped to `-p zerocode` (the full relevant check per `AGENTS.md`).
- **Rebase note:** Rebased onto current `master` (was based on `b0c6ab866`). The earlier red CI was a pre-existing master breakage in `zeroclaw-channels` (`E0560`), unrelated to this single-file diff and since fixed on master by #7357; the rebase picks that up so CI is green.
- **If any command was intentionally skipped, why:** None skipped within the relevant scope.

- Secrets / tokens / credentials handling changed? `No` — the `api_key` field is still rendered masked (one `•` per char); this only fixes how many rows the modal reserves for it.

## Compatibility (required)

- Backward compatible? `Yes`
- Upgrade steps: none — purely an in-binary TUI layout fix.

Low-risk PR: `git revert f5e1557ee` is the rollback plan.

### DIFF
--- a/apps/zerocode/src/quickstart_pane.rs
+++ b/apps/zerocode/src/quickstart_pane.rs
@@ -1873,6 +1873,32 @@ fn generate_run_id() -> String {
     format!("{now:x}-{pid:x}")
 }
 
+/// Wrapped visual-row height of each logical line at `width`, using the
+/// same word-wrap (`Wrap { trim: false }`) the modal body renders with.
+/// Every line occupies at least one row — a blank line still takes a row.
+///
+/// Sizing the modal by logical line count alone left it too short
+/// whenever content soft-wrapped: long risk-profile blurbs pushed the
+/// `yolo` option off-screen, and a long pasted `api_key` pushed the
+/// `model` picker out of view. These heights drive both the box size and
+/// the cursor-tracking scroll so the geometry survives wrapping.
+fn wrapped_row_heights(lines: &[Line], width: u16) -> Vec<u16> {
+    lines
+        .iter()
+        .map(|line| {
+            Paragraph::new(line.clone())
+                .wrap(Wrap { trim: false })
+                .line_count(width)
+                .max(1) as u16
+        })
+        .collect()
+}
+
+/// Total wrapped rows a block of lines occupies at `width`.
+fn wrapped_total(lines: &[Line], width: u16) -> u16 {
+    wrapped_row_heights(lines, width).iter().copied().sum()
+}
+
 /// Paint the modal and return `(inner_rect, row_to_cursor)` so the
 /// pane's mouse handler can resolve a click to a cursor index. The
 /// `row_to_cursor` vec maps each body row (top → bottom) to either
@@ -2250,15 +2276,40 @@ fn draw_modal(
     };
 
     let box_w = area.width.saturating_sub(8).min(80);
-    let header_h = header_lines.len() as u16;
-    let total_content = header_h + body_lines.len() as u16;
-    let box_h = (total_content + 4).min(area.height.saturating_sub(4));
+    let block = theme::modal_block(&title).padding(Padding::horizontal(1));
+    // Width left for wrapped text inside the block (its borders plus the
+    // horizontal padding). Measured off the block so it tracks any future
+    // border/padding change rather than hard-coding `box_w - 4`.
+    let inner_width = block
+        .inner(Rect::new(area.x, area.y, box_w, area.height))
+        .width;
+
+    // Size the box from the *wrapped* row counts, not the logical line
+    // counts. Long picker blurbs and long pasted field values (e.g. an
+    // `api_key`) soft-wrap across several rows; sizing by line count alone
+    // left the box too short, so later rows — the `yolo` risk option, the
+    // `model` picker — fell outside the viewport entirely.
+    let body_heights = wrapped_row_heights(&body_lines, inner_width);
+    let header_rows = wrapped_total(&header_lines, inner_width);
+    // Prefix sums: where each logical body line begins in wrapped-row
+    // space. `row_starts[i]` is line `i`'s first row; the trailing entry
+    // is the total wrapped-row count.
+    let mut row_starts: Vec<u16> = Vec::with_capacity(body_heights.len() + 1);
+    let mut acc = 0u16;
+    for h in &body_heights {
+        row_starts.push(acc);
+        acc = acc.saturating_add(*h);
+    }
+    row_starts.push(acc);
+    let body_rows = acc;
+    // content rows + top/bottom border + footer row (+1 slack).
+    let box_h = (header_rows.saturating_add(body_rows).saturating_add(4))
+        .min(area.height.saturating_sub(4));
     let x = area.x + area.width.saturating_sub(box_w) / 2;
     let y = area.y + area.height.saturating_sub(box_h) / 2;
     let rect = Rect::new(x, y, box_w, box_h);
 
     frame.render_widget(Clear, rect);
-    let block = theme::modal_block(&title).padding(Padding::horizontal(1));
     let inner = block.inner(rect);
     frame.render_widget(block, rect);
 
@@ -2266,7 +2317,7 @@ fn draw_modal(
     // space is split between an optional header band (per-field help)
     // and the body (form rows / picker entries).
     let inner_content_h = inner.height.saturating_sub(1);
-    let effective_header_h = header_h.min(inner_content_h);
+    let effective_header_h = header_rows.min(inner_content_h);
     let header_rect = Rect::new(inner.x, inner.y, inner.width, effective_header_h);
     let body_rect = Rect::new(
         inner.x,
@@ -2275,24 +2326,35 @@ fn draw_modal(
         inner_content_h.saturating_sub(effective_header_h),
     );
 
-    let body_h = body_rect.height as usize;
-    let body_len = body_lines.len();
-    let scroll_offset: u16 = if body_len > body_h && body_h > 0 {
-        // Pick the cursor line that should stay visible. Modals without
-        // a row cursor (TextInput) leave this as None and the body just
-        // top-aligns; everything else (Picker, FieldForm, ChannelList)
-        // keeps the selected row inside the viewport.
-        let selected_line = match modal {
-            Modal::Picker(p) => cursor_lines.get(p.cursor).copied(),
-            Modal::FieldForm(f) => cursor_lines.get(f.cursor).copied(),
-            Modal::ChannelList(cl) => cursor_lines.get(cl.cursor).copied(),
-            Modal::PeerGroupList(pl) => cursor_lines.get(pl.cursor).copied(),
-            Modal::Agent(a) => cursor_lines.get(a.cursor).copied(),
-            Modal::TextInput(_) => None,
-        };
+    let body_h = body_rect.height;
+    // Which cursor row must stay visible. TextInput has no row cursor, so
+    // its body just top-aligns; everything else keeps the selected row in
+    // view. `selected_line` is a logical body-line index; `row_starts`
+    // maps it into wrapped-row space so the scroll math survives wrapping.
+    let selected_line = match modal {
+        Modal::Picker(p) => cursor_lines.get(p.cursor).copied(),
+        Modal::FieldForm(f) => cursor_lines.get(f.cursor).copied(),
+        Modal::ChannelList(cl) => cursor_lines.get(cl.cursor).copied(),
+        Modal::PeerGroupList(pl) => cursor_lines.get(pl.cursor).copied(),
+        Modal::Agent(a) => cursor_lines.get(a.cursor).copied(),
+        Modal::TextInput(_) => None,
+    };
+    let scroll_offset: u16 = if body_rows > body_h && body_h > 0 {
         match selected_line {
-            Some(sel) if sel >= body_h => (sel + 1 - body_h) as u16,
-            _ => 0,
+            Some(line) => {
+                let start = row_starts.get(line).copied().unwrap_or(0);
+                let end = row_starts.get(line + 1).copied().unwrap_or(body_rows);
+                if end <= body_h {
+                    // Selected row ends within the first screenful — no scroll.
+                    0
+                } else {
+                    // Bring the selected row's bottom to the viewport bottom,
+                    // but never past its top (handles a row taller than the
+                    // viewport, e.g. a long pasted secret rendered as bullets).
+                    (end - body_h).min(start)
+                }
+            }
+            None => 0,
         }
     } else {
         0
@@ -2331,10 +2393,20 @@ fn draw_modal(
     let row_rects: Vec<Rect> = cursor_lines
         .into_iter()
         .map(|line_idx| {
-            let scrolled = (line_idx as u16).checked_sub(scroll_offset);
-            match scrolled {
+            let start = row_starts.get(line_idx).copied().unwrap_or(0);
+            let height = body_heights.get(line_idx).copied().unwrap_or(1).max(1);
+            match start.checked_sub(scroll_offset) {
                 Some(dy) if dy < body_rect.height => {
-                    Rect::new(body_rect.x, body_rect.y + dy, body_rect.width, 1)
+                    // Span the row's full wrapped height (clipped to the
+                    // viewport) so a click on a wrapped continuation row
+                    // still resolves to the right cursor.
+                    let visible = height.min(body_rect.height - dy);
+                    Rect::new(
+                        body_rect.x,
+                        body_rect.y + dy,
+                        body_rect.width,
+                        visible.max(1),
+                    )
                 }
                 _ => Rect::new(0, 0, 0, 0),
             }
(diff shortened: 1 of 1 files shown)


theonlyhennygod (approved):
First review of head 6d5b8a704846c1f00fed2285b177fb12e8a59e0a. I read the PR body, the full diff (single file, `apps/zerocode/src/quickstart_pane.rs`, +125/−25), and #7359. I cross-checked the new geometry against the modal render path and traced the red CI to its root cause. The code change is sound; the failing CI is a pre-existing master breakage unrelated to this diff (details in the merge-readiness note below). Approving on substance.

### 🟢 What looks good — Sizing and scrolling in wrapped-row space, not logical-line space

The core of the fix is correct, and it quietly repairs a second latent bug. The body renders with `.scroll((scroll_offset, 0))`, and ratatui's `Paragraph::scroll` counts *wrapped* rows — so the old code fed it a *logical* line index, which was wrong in the same direction the box was under-sized. Measuring `wrapped_row_heights` / `wrapped_total` via `Paragraph::line_count` with the same `Wrap { trim: false }` the body uses, then deriving both `box_h` and `scroll_offset` from those wrapped counts, makes the box height, the scroll offset, and the render path all speak one unit. I verified the header band also renders with `Wrap { trim: false }`, so measuring `header_rows` wrapped (not by `header_lines.len()`) is consistent with what's painted. The `.max(1)` floor correctly preserves a row for blank lines.

### 🟢 What looks good — Hit-test spans wrapped height, and inner width is measured off the block

Two details that show care. First, the `row_rects` hit-test now spans each row's full wrapped height (`visible = height.min(body_rect.height - dy)`, clamped with `.max(1)`), so a click on a soft-wrapped continuation row still resolves to the right cursor — the old single-row rect would have silently missed those. Second, `inner_width` is derived from `block.inner(...)` rather than hard-coding `box_w - 4`, so it tracks any future border/padding change instead of drifting. The `(end - body_h).min(start)` scroll clamp is the right call for the row-taller-than-viewport case (a long pasted secret rendered as bullets) — it pins the row's top rather than scrolling past it.

### 🟡 Warning — The behavior the PR fixes isn't covered by a test; only the measurement helpers are

The two new unit tests lock in `wrapped_total` / `wrapped_row_heights` (the measurement primitives), which is good — but the actual regression from #7359 lives one layer up, in the prefix-sum `row_starts` → `scroll_offset` logic and the wrapped-height hit-test, and that layer is verified only by hand-tracing in the PR body. The scenarios you described (the 3-option risk picker where `yolo` must stay visible; the field form keeping `model` on screen while `api_key` wraps; scrolling only when even the grown box can't fit) are exactly the kind of geometry that silently re-regresses on a future refactor. A focused test that builds a `Modal::Picker`/`FieldForm` with known-wrapping content and asserts the resulting `scroll_offset` (and that the selected row's rect is non-zero / within `body_rect`) would protect the fix itself, not just its inputs. Non-blocking given this is an Experimental-tier TUI layout change with the helpers pinned and `-p zerocode` green — but it's worth doing while the reasoning is fresh.

## Merge readiness — red CI is a pre-existing master breakage, not this diff

This is the one thing standing between the PR and merge, and it is **not** something this PR should fix. Both `Lint` and `Test` fail to compile `zeroclaw-channels` (lib test) with three `E0560`s at `crates/zeroclaw-channels/src/orchestrator/mod.rs:15443`/`15461`/`15462`:

```
error[E0560]: struct `orchestrator::ChannelRuntimeContext` has no field named `default_model_provider`
error[E0560]: struct `orchestrator::ChannelRuntimeContext` has no field named `api_key`
error[E0560]: struct `orchestrator::ChannelRuntimeContext` has no field named `api_url`
   = note: available fields are: `model_provider_ref`
```

I confirmed this exists on `origin/master` independent of this PR: the struct (mod.rs:353-356) carries only `model_provider_ref`, ~30 test constructors were migrated to it, but the one at line 15443 was missed and still sets `default_model_provider` / `api_key` / `api_url`. This PR's diff is confined to `quickstart_pane.rs` and cannot produce a `zeroclaw-channels` compile error — and the author correctly scoped local validation to `-p zerocode` (the relevant battery per AGENTS.md for a single-crate zerocode change), which doesn't compile the channels lib-tests, so this was invisible locally. The fix is a one-line repair on master (rename that orphaned ctor's field to `model_provider_ref`, drop the two removed fields). Once master is green, this PR rebases clean; if a maintainer wants it in before then, it's an `--admin` merge candidate since the failing jobs are not testing this code. Flagging so whoever lands it knows the red is upstream, not here. (Same shape as the recent ollama-E0308 episode on #7226/#7231.)

### 🟢 What looks good — Honest, well-scoped validation evidence

The body is precise about what was and wasn't verified: scoped clippy/test to `-p zerocode` with the rationale stated, hand-traced both reported cases at 80x31, and explicitly listed "live rendering on a physical terminal" as not verified. That's exactly the kind of disclosure that lets a reviewer calibrate quickly.

WareWolf-MoonWall (approved):
Second review of head 6d5b8a704846c1f00fed2285b177fb12e8a59e0a by @WareWolf-MoonWall. I read the PR body, @theonlyhennygod's review, the full diff (single file, `apps/zerocode/src/quickstart_pane.rs`, +125/−25), #7359, and traced the geometry logic through the modal rendering pipeline. The code is correct and the validation evidence is thorough for the scope. Approving on substance.

### 🟢 What looks good — Unified measurement vocabulary for box sizing, scrolling, and hit-testing

The shift from logical-line indexing to wrapped-row indexing is sound and consistent. All three geometry surfaces — `box_h` (derived from `header_rows + body_rows`), `scroll_offset` (computed from `row_starts` prefix sums), and `row_rects` (spanning each row's wrapped height) — now speak the same unit as `Paragraph::scroll` and the body's `Wrap { trim: false }` render. The old code mixed units (logical lines for box size and scroll, wrapped rows for render), which guaranteed the kind of off-screen rows the linked issue reports. Deriving `row_starts` via prefix sums and using `.checked_sub(scroll_offset)` for hit-test placement is the right approach for variable-height rows.

### 🟢 What looks good — Careful treatment of edge cases

Three details: (1) `.max(1)` floor in `wrapped_row_heights` ensures blank lines still reserve a row, preserving the visual rhythm when content is sparse. (2) The scroll clamp `(end - body_h).min(start)` handles the row-taller-than-viewport case (e.g. a pasted secret rendered as bullets) by pinning the row's top rather than scrolling past it — exactly right. (3) `inner_width` is measured from `block.inner(...)` rather than hard-coded, so it tracks future border/padding changes.

### ✅ Resolved — Red CI is a pre-existing master breakage, not this diff

@theonlyhennygod traced the `E0560` failures to a stale `ChannelRuntimeContext` fixture on master (line 15443) that predates this PR. The diff is confined to `apps/zerocode/src/quickstart_pane.rs` and cannot produce a `zeroclaw-channels` compile error. The author correctly scoped local validation to `-p zerocode` (the relevant battery per AGENTS.md for a single-crate change), which doesn't compile the channels lib-tests. The fix for master is tracked separately (now landed as #7357).

### 🟡 Warning — The behavior the PR fixes isn't covered by a test; only the measurement helpers are

This echoes @theonlyhennygod's 🟡. The two new unit tests (`wrapped_total_counts_soft_wrapped_rows`, `wrapped_row_heights_are_measured_per_line`) lock in the measurement primitives, which is valuable — but the actual #7359 regression lives one layer up, in the `row_starts` → `scroll_offset` logic, the wrapped-height `row_rects`, and the scroll clamp. A focused test that builds a `Modal::Picker` or `Modal::FieldForm` with known-wrapping content (e.g. a 40-char risk blurb at width 10) and asserts the resulting `scroll_offset` (and that the selected row's rect is non-zero / within `body_rect`) would protect the fix itself, not just its inputs. Non-blocking given this is an Experimental-tier TUI layout change with the helpers pinned and `-p zerocode` green — but worth doing while the reasoning is fresh, to guard against a future geometry refactor silently re-introducing the off-screen row bug.

### 🟢 What looks good — Thorough, scoped validation evidence

The PR body is precise about what was verified (scoped clippy/test to `-p zerocode` with the rationale stated, hand-traced both reported cases at 80x31) and what wasn't ("live rendering on a physical terminal"). That's exactly the kind of honest disclosure that lets a reviewer calibrate risk quickly and builds trust over time.

JordanTheJet:
Pushed an update (force-push, head now `f5e1557`). Three changes since the approvals:

1. **Addressed the shared 🟡 (test the fix, not just its inputs).** Added two headless `TestBackend` render tests that drive the real `draw_modal` path:
   - `picker_keeps_every_option_visible_when_blurbs_wrap` — every risk option, including `yolo`, gets a non-zero hit-rect, spaced by *wrapped* height (≥2 rows) rather than logical lines (the pre-fix 1-row spacing).
   - `picker_scrolls_to_keep_selected_option_visible` — when content overflows the grown box, the selected row scrolls into view while a row scrolled off the top collapses to a zero rect. This exercises the `row_starts → scroll_offset → row_rects` chain the helper tests don't reach. Both fail on the pre-fix logic.
2. **Rebased onto current `master`** to clear the red CI @theonlyhennygod traced to the pre-existing `zeroclaw-channels` `E0560` (now fixed on master by #7357). `-p zerocode` battery green: fmt clean, clippy `-D warnings` clean, lib 40 + bins 320 tests pass.
3. **Reframed the description** to lead with the risk-picker `yolo` case, since that's the part #7330 does not address.

On the #7330 relationship (flagging for @singlerider as its author): #7330 caps the secret mask at 24 bullets, which mitigates the `api_key` *trigger* but leaves the box geometry — and the risk-picker `yolo` case — unfixed. A local test-merge shows the production `draw_modal` code auto-merges cleanly; the only collision is a trivial "keep both" where each PR appends new `#[test]` fns. The two are complementary (input bound vs. layout math) and can both land.

PR #7249, opened 2026-06-05
feat(zerocode): theme enhancements — color-depth fallback, registry-generated presets, per-agent overrides, palette swatches

singlerider:
> **DO NOT MERGE - AUTHOR WILL MERGE WHEN READY. REVIEWS STILL REQUESTED**

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - zerocode themes were 24-bit `Color::Rgb` emitted unconditionally; old macOS Terminal.app and tmux/screen (which clamp truecolor) rendered them garbled. Added lazy color-depth detection that down-converts every themed colour to the nearest xterm-256 / ANSI-16 value so the palette stays legible everywhere instead of forcing the flat `terminal` theme.
  - The TUI carried its own hand-authored 9-theme palette list, separate from the dashboard/mdBook registry (`web/src/contexts/themes.json`). Now the TUI preset table is generated at build time from that same registry, so all three surfaces expose one named-theme set with no drift.
  - Added 8 widely-used themes (Tokyo Night, Catppuccin Mocha/Latte, One Dark/Light, Ayu Dark, Gruvbox Light) and sorted the registry by scheme then id; preserved the TUI's prior gruvbox/high-contrast themes by promoting them into the registry rather than re-hardcoding.
  - Per-agent theme override for the Code and Chat panes (`[theme.agent_override.<alias>].name`): a theme follows the focused agent, applied live via a process-global override registry. Editable from a new Agent Themes section in Config > zerocode.
  - Inline palette swatches preview the highlighted theme's colours in the Theme list, with a stable name indent across selections.
  - Unknown theme names (global or per-agent) fall back to the `terminal` inherit theme rather than aborting the TUI at startup.
  - Docs: promoted zerocode to its own top-level docs section (Overview, Config pane, Themes, Remote/WSS, Environment pass-through) embedded into Getting Started via a one-line include. Framed the Config pane as the preferred way to change settings, with raw `config.toml` presented as the persisted result rather than the primary instruction (hand-editing scoped to headless/scripted cases). The themes page's named-theme list is now generated at build time from `web/src/contexts/themes.json` (gitignored fragment, `{{#include}}`d), so it cannot drift from the registry.
- **Scope boundary:** zerocode TUI and its docs only, plus the shared theme registry JSON and the regenerated mdBook switcher. No daemon, gateway, provider, or runtime behaviour changes.
[...]

PR #7190, opened 2026-06-04
feat(zerocode): outbound message queue with sidebar and injection

singlerider:
> **DO NOT MERGE - AUTHOR WILL MERGE WHEN READY. REVIEWS STILL REQUESTED**

## Summary

- **Base branch:** `master`
- **What changed and why:** zerocode previously hard-blocked all input while a
  turn was responding. This replaces that with an outbound message queue:
  every submit enqueues a QueuedMessage on ChatState and the front dispatches
  whenever the session is idle, preserving the single-in-flight-turn invariant.
  Lets the user compose, attach, and submit at any time. Adds Ctrl+Enter
  front-injection, an Alt+Q sidebar, and per-queued-item edit/delete/select.
- **Scope boundary:** zerocode TUI only (Chat + ACP panes). No runtime/provider
  protocol changes; `factory.rs` touch is the base-state llamacpp fix already
  on master, not part of this feature.
- **Blast radius:** zerocode input dispatch and turn-completion paths. No effect
  on gateway, channels, or non-zerocode surfaces.
- **Linked issue(s):** none.
- **Labels:** `enhancement`, `risk: medium`, `size: L`, `runtime`.

## Validation Evidence (required)

- `cargo fmt --all -- --check`,
  `cargo clippy --all-targets -- -D warnings`, and `cargo test` are pending and
  will be posted before this leaves Draft.
- **Beyond CI — what did you manually verify?** See checklist.

### Integration test checklist (`integration/zerocode-test`)

Validated as part of the aggregated integration branch. The items below are this PR's slice of that sign-off; `[x]` = verified (unit and/or live), `[~]` = code-verified pending live, `[ ]` = live-pending.

#### 5. zerocode outbound message queue — base feature (#7190 Draft) — Tier 1

Commit `62bee767b`. Submit no longer hard-blocks during a turn; messages enqueue
and dispatch on idle. Single-in-flight invariant preserved. Chat AND ACP panes.

DETERMINISTIC (unit tests, `cargo test -p zerocode --bins` → 272 pass, 0 fail):
core dispatch, FIFO, single-in-flight, injection-to-front, pause/resume,
resume-override one-shot, Alt+E pull, empty-reject, attachment-only-accept,
depth-cap-32, reset, sidebar-visibility-tracks-contents, select-by-id.

##### Core dispatch
- [x] Submit while idle → sends immediately — `enqueue_dispatches_immediately_when_idle`
[...]

PR #7366, opened 2026-06-08
fix(zerocode): restore mid-turn input so the outbound message queue works

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** PR #7190 replaced zerocode's hard input block with an outbound message queue whose stated premise is that the user can compose, attach, and submit *at any time, including while a turn is responding*. The squash merge shipped against that design: it deleted the doc comment describing the `turn_in_flight` gate in `InputBarState::handle_key` but left the gate itself — `if turn_in_flight { return InputBarAction::NotHandled }` — in place. That early-return swallowed every keystroke before it could reach the parent's queue routing in `chat.rs`, producing the exact rollback symptom #7190 listed: zerocode input unresponsive while a turn is in flight, and queued/injected messages never reaching dispatch.
- **Why it slipped through:** this landed via an aggregated integration branch carrying roughly nineteen concurrent zerocode PRs. The author validated the feature against an integration tree where the gate had been removed (PR body: "compose + attach + submit freely while a turn responds (no input block) — PASS"), but the gate-removal was lost when #7190 was squashed onto `master`. The later #7249 (theme enhancements) never touches `input_bar.rs`, so this is #7190's own merge loss, not downstream corruption.
- **The fix:** remove the gate so keys flow to the queue mid-turn, then drop the now-unused `turn_in_flight` parameter from `handle_key`, its single call site in `chat.rs`, and three in-module test call sites. The render path (`render_input_bar`) keeps its own `turn_in_flight` argument, which still suppresses the placeholder and cursor during a turn — that behavior is intentional and unchanged.
- **Scope boundary:** zerocode TUI input dispatch only. No change to queue data structures, dispatch/pause/inject logic, RPC, runtime, providers, gateway, or channels. The `chat.rs` parent routing (`Submit` / `Inject` / `ResumeQueue` / `ClearQueue`) was already correct and is untouched apart from the call-site signature.
- **Blast radius:** `InputBarState::handle_key` and its one caller. The same `InputBarState` instance backs both the Chat and ACP panes through a single call site, so the fix restores mid-turn input for both. No effect on any non-zerocode surface.
- **Linked issue(s):** Related #7190

Full battery run locally against this branch (`apps/zerocode/src/**`, 5 insertions / 10 deletions across 2 files):

```
$ cargo fmt --all -- --check
(no output)                                  # exit 0

    Finished `dev` profile [unoptimized + debuginfo] target(s)   # exit 0, full workspace

$ cargo test
... all suites ok ...
test result: ok. 9 passed; 0 failed   (test_system.rs)
   Doc-tests zeroclaw
test result: ok. 0 passed; 0 failed                              # exit 0

$ cargo test -p zerocode --bins
test result: ok. 316 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

- **Commands run and tail output:** all four pass clean (tails above). The queue tests #7190 enumerated in its own validation checklist all pass on this branch: `no_dispatch_while_turn_in_flight`, `fifo_order_preserved`, `injection_jumps_ahead_of_pending`, `cancel_pauses_pending_but_injection_resumes`, `enqueue_dispatches_immediately_when_idle`, `empty_enter_resumes_queue`.
- **Beyond CI — what did you manually verify?** Traced the full keystroke path: `chat.rs` delegates to `input_bar.handle_key`, then matches on `Submit` / `Inject` / `ResumeQueue` / `ClearQueue` — each of which already handles the `turn_in_flight == true` case, including the mid-turn `session_cancel` issued on an inject. Confirmed the removed gate was the sole short-circuit preventing that wired-up path from ever running during a turn. Confirmed `turn_in_flight` is genuinely unused in `handle_key` after the gate removal (so dropping the parameter is correct, not a silent behavior change) and that `render_input_bar` retains its own independent `turn_in_flight` argument for placeholder/cursor suppression.
- **What I did NOT verify:** no live interactive TUI session was run in this environment. Coverage rests on the deterministic queue unit tests, which model the dispatch, injection, and pause/resume invariants the feature depends on.

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` (internal `handle_key` signature only; no public API, config, or keybinding change)
- If `No` or `Yes` to either: no upgrade steps. This restores the intended behavior of an already-shipped feature; no user action required.

- **Fast rollback command/path:** `git revert <merge-sha>` on this PR. Reverting reinstates the `turn_in_flight` gate in `handle_key` and the parameter on the call sites, returning to the current (broken) hard-block behavior — no other subsystem is affected.
- **Feature flags or config toggles:** None. The behavior is unconditional; there is no flag gating mid-turn input.
- **Observable failure symptoms:** if this regresses again, zerocode input goes unresponsive while a turn is responding — keystrokes produce no echo in the input bar, the queue sidebar (Alt+Q) never gains entries from typed sends, and Ctrl+Enter injection has no effect mid-turn. There is no log line to grep; the symptom is purely interactive. The guard against re-regression is the `no_dispatch_while_turn_in_flight` and injection/queue unit tests in `apps/zerocode/src/chat.rs`.


Audacity88 (approved):
I reviewed the live PR state at 3f2424232fe63cd851110230f6a4db073d54b41d, the PR body, current labels/checks, prior comments/reviews, the full two-file diff, and the surrounding `chat.rs` / `input_bar.rs` dispatch path. I also live-tested the PR binary in zerocode against a slow local OpenAI-compatible mock provider. There are no prior active review blocks, and the current CI run is green.

### 🟢 What looks good — mid-turn input is restored at the right boundary

This is the narrow fix I expected for the #7190 regression. Removing the `turn_in_flight` early return from `InputBarState::handle_key` lets keystrokes reach the parent `chat.rs` routing again, while the parent still owns the safety-sensitive gates for approvals, browse mode, queue dispatch, injection cancellation, and chat-specific turn controls. That keeps the queue feature's intended behavior without broadening the input bar's responsibility.

The validation also lines up with the claim: the existing queue tests cover the important split between "can enqueue while a turn is live" and "must not dispatch the next normal item until the live turn settles," plus the inject and pause/resume cases. Keeping the render path's separate `turn_in_flight` handling unchanged also makes this a targeted dispatch fix rather than a UI/rendering behavior change.

The live test matched that behavior: I sent a first prompt, typed `second queued during live turn` while the first response was still streaming, saw the text stay editable in the input bar, pressed Enter, and confirmed the mock provider received a second chat-completion request.

### 🟡 Warning — Queued follow-ups need a visible Clear queue affordance

Now that mid-turn input can be submitted again, the queued state needs an obvious way back out. The underlying queue clear path already exists via `/clear-queue`, and the queue sidebar has delete/clear help entries, but in the live mid-turn input flow I did not see a visible `Clear queue` hint near the queued follow-up. That makes the restored behavior feel a little too committed: if a user sends a mid-turn follow-up and immediately changes their mind, they should not have to discover a slash command or open help before dispatch.

I do not think this should block the regression fix, but it should be treated as follow-up UX work for the queue surface. A visible `Clear queue` action or hint near the queued message/input area would be enough; this looks like discoverability for existing queue plumbing rather than a new queue semantic.

Approving with that non-blocking queue UX follow-up noted above. The regression symptom matches the diagnosis, the scope is minimal, and the existing queue tests cover the restored dispatch invariants.

PR #6266, opened 2026-05-02
feat(config): schema v3 migration, channel aliasing, model-provider aliasing, profile lifting

singlerider:
# DO NOT MERGE TO MASTER YET

This PR targets `integration/v0.8.0`, not `master`. It is part of a coordinated batch of breaking-change PRs that must land together via the integration branch before a single squash into master.

See #5947 for the full v0.8.0 merge checklist.

## Purpose

Make sure everything relating to the V3 schema works and is ready for the supporting PRs that will land feature logic on top of it. This PR finalizes the schema shape and the V1, V2, V3 migration chain. It does not change runtime agent behavior or tool implementations beyond the orchestrator-side per-alias provider resolution that was bundled in to honor the V3 alias contract end-to-end (see "Post-merge follow-ups" below).

UI surfaces touched here are exclusively the result of merging `upstream/master` into the branch and adapting the incoming UI work to the new V3 agent shape (alias references replacing inline brain fields, the channel `HashMap<String, T>` shape replacing `Option<T>`, profile-keyed references for risk and runtime, etc.). No new UI features ship in this PR; follow-up UI work belongs in the supporting issues against `integration/v0.8.0`.

After this PR lands on `integration/v0.8.0`:

1. Supporting issues (#6270, #6271, #6272, #6273) ship feature logic against the now-stable V3 schema.
2. The integration branch passes its full gate-check suite.
3. A single squash merges the entire batch into `master` as the v0.8.0 release.

### V3 schema definition

- Channel aliasing: `[channels.<type>]` becomes `[channels.<type>.<alias>]`. Every channel type lives under `HashMap<String, T>` keyed by user alias (default key: `"default"`).
- Provider aliasing: `[providers.models.<id>]` becomes `[providers.models.<type>.<alias>]`. Outer key is the provider type (`anthropic`, `openrouter`, etc.), inner key is the user alias.
- TTS subsystem unification: V2 per-provider blocks (`[tts.openai]`, `[tts.elevenlabs]`, etc.) collapse into the same alias-keyed shape under `[providers.tts.<type>.<alias>]`.
- Storage subsystem promotion: V2 `[memory.qdrant]`, `[memory.postgres]`, and `[storage.provider.config]` consolidate under `[storage.<backend>.<alias>]` typed maps.
[...]

Issue #5890, opened 2026-04-19
RFC: Multi-agent UX flow — design


singlerider:
## RFC Status

- [x] Draft posted (2026-04-19)
- [x] 7-day discussion period started
- [x] 7-day discussion period concluded
- [x] Core Team vote per #5577 §8.2 (two-thirds majority)
- [x] Decision recorded
- [ ] (If accepted) Proposal extracted to `docs/proposals/multi-agent-ux-flow.md`
- [ ] (If accepted) ADR(s) opened in `docs/architecture/`
- [x] (If accepted) Implementation tracking begins under the multi-agent v1 tracker

## Revision History

| Rev | Date | Summary |
|---|---|---|
| 1 | 2026-04-19 | Initial draft. Project-wide guiding statement and seven section-level mission statements covering Define & Manage, Invoke, Channels & Routing, Isolate, Observe, Gateway & Dashboard, and Onboard. Schema-wide aliasing pattern, per-agent identity model, TOML-canonical agent definitions with markdown-content companions, single-daemon process model. |
| 2 | 2026-04-19 | Added Process Model subsection on config persistence and live reload: daemon as sole authoritative writer during runtime, comment-preserving atomic save, live reload of aliased instances, explicit restart semantics for flat global settings. |
[...]

PR #7367, opened 2026-06-08
feat(gateway): route inbound webhooks per channel alias (#6312)

### LINKED ISSUE #6312: feat(gateway): per-alias webhook path routing for multi-instance channels
> **Update 2026-06-09 (audit against master @ d23295118).** #7297 landed \`/webhook?agent=\` per-request agent dispatch, so two bots bound to different agents can now coexist on one gateway by passing \`?agent=\`. That is a different mechanism than the per-alias **path** routing this issue asks for, and it does not satisfy the acceptance criteria below (\`/webhook/<type>/<alias>\` paths, alias-suffix fallback, callsite replacement). This issue remains open for the path-routing work; PR #7367 targets it. The \`values().next()\` callsite list in "Affected callsites" has also shrunk since filing — see the corrected list there.

## Context

V3 supports multiple channel instances per type (\`[channels.<type>.<alias>]\`)
but the gateway still treats webhooks as single-instance: each channel
type listens at a fixed path (\`/webhook/telegram\`, \`/webhook/discord\`,
etc.) and resolves to whichever alias is first in the HashMap. (As of
#7297, \`/webhook\` additionally honors a \`?agent=\` query parameter for
per-request agent dispatch, but channel-instance resolution by path is
still first-alias-wins.)

After #6266 landed:
- Multi-agent UX lets users declare two telegram bots (e.g.
  \`telegram.work\` + \`telegram.personal\`) and bind each to a different agent.
- The gateway currently picks one — \`values().next()\` — to handle the
  webhook. Both bots cannot coexist on the same gateway instance.

## Proposal

Per-alias webhook paths: \`/webhook/<type>/<alias>\` dispatches to the
correct instance. Existing \`/webhook/<type>\` keeps working as a fallback
that routes to the first configured alias (back-compat for single-instance
configs).

Implementation sketch:

1. \`api_webhook.rs\` (gateway): take \`<type>\` and \`<alias>\` from path,
   look up \`config.channels.<type>.get(<alias>)\`.
2. Return 404 when alias not found instead of 500.
3. Fall back to \`values().next()\` only when path lacks alias suffix.
4. Update channel registration (telegram, discord, slack, mattermost,
   matrix, signal, qq, whatsapp, etc.) to use the new path scheme.

## Acceptance

- [ ] \`/webhook/telegram/work\` reaches the \`telegram.work\` channel.
- [ ] \`/webhook/telegram/personal\` reaches \`telegram.personal\`.
- [ ] \`/webhook/telegram\` (no alias) routes to first configured alias
      with a deprecation warning header.
- [ ] Tests for both paths.
- [ ] All gateway-side \`config.channels.<type>.values().next()\` callsites
      replaced with per-path routing.

## Affected callsites (from #6266)

\`crates/zeroclaw-gateway/src/lib.rs\`: webhook secret hash, whatsapp
Cloud config + app secret, linq channel + signing secret, wati channel,
nextcloud_talk channel.

> **Corrected 2026-06-09 (master @ d23295118).** Since filing, the Linq
> channel gained per-alias routing (#7041) and the WhatsApp path changed,
> so the remaining \`config.channels.<type>.values().next()\` first-alias-wins
> callsites in \`lib.rs\` are now: \`channels.webhook\`, \`channels.wati\`, and
> \`channels.nextcloud_talk\` (3 sites). The per-path routing work should
> replace those; the others have already moved off the first-alias pattern.

Refs #6266, RFC #5890 §1.

IftekharUddin:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Inbound webhooks resolved their channel instance with `config.channels.<type>.values().next()`, so a multi-instance config (e.g. `whatsapp.work` + `whatsapp.personal`) only ever delivered to the **first** instance. Added path-based routing `POST /<type>/<alias>` that reaches the matching instance for **whatsapp, wati, nextcloud_talk, linq**.
  - New `api_webhook::resolve` helper (`Alias` / `Fallback` / `NotFound`); `AppState` channel + secret fields widened from `Option<Arc<…>>` to `HashMap<String, Arc<…>>`, and the builder now constructs **every** configured aliased instance.
  - Bare `/<type>` paths kept as a **deprecated** fallback → first instance + `X-Zeroclaw-Deprecation` response header. Unknown alias now returns **404** instead of effectively 500. Per-channel signature/secret verification preserved, keyed by the resolved alias.
- **Scope boundary:** Only the four channels with gateway webhook routes are touched. telegram/discord/slack/mattermost/matrix/signal/qq have no gateway webhook routes (long-poll/socket) and are unchanged. The generic `POST /webhook` prompt endpoint is not channel-typed and is left as-is (keeps the only remaining `values().next()`, documented inline). gmail push (`/webhook/gmail`, active-agent lookup) is unchanged.
- **Blast radius:** Gateway webhook ingress + auth for the four channels. The `AppState` field-type change is internal to `zeroclaw-gateway` (only constructors are in that crate); no other crate constructs it. The 404-vs-500 change for unknown aliases is externally observable.
- **Linked issue(s):** Closes #6312. Related #6266, RFC #5890 §1.
- **Labels:** `enhancement`, `gateway`, `docs`, `risk: high`, `size: L` — all applied. (`risk: high` and `size: L` are now set per reviewer request; this repo has no `type: feature` label, so `enhancement` is the closest equivalent.)

Clippy and tests scoped to the affected crate (`zeroclaw-gateway`); the change is isolated to that crate and a full `cargo build --workspace` confirms no downstream breakage.

```bash
cargo fmt --all -- --check
# channel handlers sit behind cfg-gates, so enable the webhook channel features:
FEATS=channel-whatsapp-cloud,channel-wati,channel-nextcloud,channel-linq
cargo clippy -p zeroclaw-gateway --features $FEATS --tests -- -D warnings
cargo test  -p zeroclaw-gateway --features $FEATS
cargo build --workspace
```

  - `cargo fmt --all -- --check` → exit 0, no output (clean).
  - `cargo clippy … -D warnings` → `Finished \`dev\` profile … in 1m 07s` (no warnings; `-D warnings` would have failed otherwise).
  - `cargo test -p zeroclaw-gateway --features $FEATS` → `test result: ok. 255 passed; 0 failed; 0 ignored`.
  - `cargo build --workspace` → exit 0.
- **Beyond CI — what did you manually verified:** New tests assert alias routing reaches the correct instance (proven via per-alias WhatsApp verify tokens — the other alias's token must NOT verify), unknown alias → 404, bare path → back-compat + deprecation header, and per-alias signature auth still rejects bad/missing and accepts valid HMAC signatures. Did NOT exercise a live end-to-end webhook against real WhatsApp/WATI/Nextcloud/Linq servers.
- **If any command was intentionally skipped, why:** Full-workspace `cargo test` not run; the change is confined to `zeroclaw-gateway`, so tests were scoped to that crate and a workspace build was run instead.

**Route tests passing** (`cargo test -p zeroclaw-gateway --features $FEATS -- api_webhook webhook_alias webhook_unknown webhook_bare`):

Per-alias webhook routing tests passing

- Secrets / tokens / credentials handling changed? **Yes** — per-channel webhook signing secrets are now stored per-alias (`HashMap<String, Arc<str>>`) and resolved by the request's alias, instead of a single `Option`. Verification logic, hashing, and constant-time comparison are unchanged; secrets are never logged. Unknown aliases return a generic `404` that does not disclose which aliases exist.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (test tokens/secrets are synthetic literals).

## Compatibility (required)

- Config / env / CLI surface changed? **No** — config schema already stored these channels as alias maps; this only adds new optional URL paths (`/<type>/<alias>`) while keeping the bare paths working.
- Upgrade steps for existing users: **None.** Single-instance deployments keep using the bare path unchanged (deterministic with one instance). Multi-instance deployments can optionally migrate webhook callback URLs to `/<type>/<alias>` to address each instance; the bare path keeps resolving to the first instance with a deprecation header until removed.

- **Fast rollback command/path:** `git revert <merge-sha>` — the change is self-contained in `zeroclaw-gateway` with no schema/data migration.
- **Observable failure symptoms:** Inbound webhooks returning unexpected `404` (grep gateway logs for `webhook target not found`) or `401` (signature/secret verification failures) for a channel that previously worked; a rise in 4xx on `/whatsapp`, `/wati`, `/nextcloud-talk`, `/linq`; replies not reaching the expected bot instance.


IftekharUddin:
### Re: which channels are actually webhook-based

The decisive check is the gateway router itself — the only inbound HTTP webhook routes are:

```rust
// crates/zeroclaw-gateway/src/lib.rs:1308-1321
.route("/webhook", post(handle_webhook))            // generic prompt endpoint (not channel-typed)
.route("/whatsapp",       …)  .route("/whatsapp/{alias}",       …)
.route("/linq",           …)  .route("/linq/{alias}",           …)
.route("/wati",           …)  .route("/wati/{alias}",           …)
.route("/nextcloud-talk", …)  .route("/nextcloud-talk/{alias}", …)
.route("/webhook/gmail",  post(handle_gmail_push_webhook))
```

Ingestion mode per channel (from each channel's own doc-comment):

| Channel | How it ingests | In this PR? |
|---|---|---|
| WhatsApp (Cloud), WATI, Nextcloud Talk, Linq | **HTTP webhook** | ✅ now alias-routed |
| Telegram | long-poll `getUpdates` (`telegram.rs:363`) | ❌ no webhook route |
| Discord | Gateway **WebSocket** (`discord.rs:20`) | ❌ no webhook route |
| Slack | polls `conversations.history` (`slack.rs:24`) | ❌ no webhook route |
| Mattermost | polls REST v4 (`mattermost.rs:65`) | ❌ no webhook route |
| Matrix / Signal / QQ | `/sync` long-poll · signal-cli · QQ Bot socket (`qq.rs:257`) | ❌ no webhook route |
| Gmail push | separate `/webhook/gmail` (active-agent lookup, not alias-keyed) | unchanged |

So the per-alias routing applies exactly to the four channels that resolve an instance from an inbound HTTP request. Polling/socket channels pick their instance at connect time, not per-request, so there's nothing to route — they're correctly out of scope. This matches the "Scope boundary" note in the PR description.

IftekharUddin:
ran into some upstream conflict errors. Working through them now.

IftekharUddin:
### Merged latest `master` — conflicts resolved

Rebased the branch onto current `master` (beta-2). Two upstream changes overlapped this work and are now reconciled:

1. **cfg-gates** — beta-2 put the gateway channel code behind `#[cfg(feature = "channel-*")]`. All new code, routes, `AppState` fields, and tests here are gated to match. (Note: the webhook channel features aren't in the crate's `default`, so validation runs with `--features channel-whatsapp-cloud,channel-wati,channel-nextcloud,channel-linq`.)
2. **linq** — beta-2 independently landed a per-alias linq (`/linq/{alias}`, alias-only, `linq_signing_secrets`). This PR now routes **all four** channels — including linq — through the shared `api_webhook::resolve` helper, which additionally **restores linq's bare `/linq` path** (→ first instance + `X-Zeroclaw-Deprecation` header) that the alias-only version had dropped. Upstream's `linq_test_state` helper and its linq tests are preserved and pass against the unified handler.

Post-merge validation (with the channel features enabled):

- `cargo clippy -p zeroclaw-gateway --features <channels> --tests -- -D warnings` → clean
- `cargo test -p zeroclaw-gateway --features <channels>` → `255 passed; 0 failed`
- `cargo build --workspace` → ok

`mergeable` is now green.

WareWolf-MoonWall (changes requested):
# Review draft — PR #7367
## feat(gateway): route inbound webhooks per channel alias (#6312)
**Author:** IftekharUddin · **Reviewer:** WareWolf-MoonWall · **Verdict:** `--request-changes`

---

I read the full diff (+870/-225), the `api_webhook.rs` module, the `AppState` and `run_gateway` changes in `lib.rs`, the upstream conflict-resolution note from the author, and verified the surrounding `lib.rs` source locally. The PR solves a real multi-instance delivery problem: `values().next()` always routed to one arbitrary instance; this replaces that with explicit `POST /<type>/<alias>` routing backed by a clean `resolve()` helper, while keeping the bare paths as a deprecated fallback.

---

### 🔴 Blocking — `risk: high` and `size: L` labels are not applied

The live GitHub labels are only `docs` and `gateway`. The PR body explicitly says "Suggested: `type: feature`, `risk: high`, `size: L`" — meaning the author knows the right tier but did not apply the labels. Per AGENTS.md risk tiers, `crates/zeroclaw-gateway/src/**` is explicitly high-risk. This is the gateway's inbound webhook auth path; signing-secret lookup, 404 vs 401 behavior, and `AppState` field type changes all land inside that boundary. The Rollback section is correctly populated (which is only required for `risk: medium`/`risk: high`), confirming the author's intent — the label just wasn't set. Please apply `risk: high` and `size: L` before merge so automated gating and queue prioritization work correctly.

---

### 🟡 Warning — Bare-path HashMap fallback is non-deterministic for multi-instance configs

`api_webhook::resolve()` implements the deprecated bare path as:

```rust
None => match map.iter().next() {
    Some((key, value)) => Resolved::Fallback { key, value },
    None => Resolved::NotFound,
},
```

`HashMap::iter()` has no guaranteed iteration order. For a single-entry map this is deterministic (and the documented single-instance back-compat case), but for a two-alias config like `whatsapp.work` + `whatsapp.personal`, the bare `/whatsapp` path resolves to a random instance on each process start. The CHANGELOG entry says "resolves to the first configured instance" — that wording implies stability it doesn't have.

The test `webhook_bare_path_is_back_compat_and_flags_deprecation` only exercises a single-entry `HashMap::from([("default", ...)])` so it passes regardless. The fix is either:
- Use an `IndexMap` (insertion-order-stable) or sort the keys on iteration for the fallback path, or
- Update the docs and CHANGELOG to say "resolves to an unspecified configured instance" to set accurate expectations.

For single-instance deployments (the intended surviving use of the bare path) this is a non-issue. The concern is documentation accuracy and surprising behavior if an operator happens to test with two aliases and the bare path.

---

### 🟡 Warning — Signing-secret fields in `AppState` are config snapshots (pre-existing, now extended)

`whatsapp_app_secret: HashMap<String, Arc<str>>`, `nextcloud_talk_webhook_secret: HashMap<String, Arc<str>>`, and equivalents are populated once in `run_gateway()` from `config.channels.<type>.iter()` and then stored for the lifetime of the process. `AppState` already carries `config: Arc<RwLock<Config>>` from which these could be read on demand. This is the "ConfigSnapshot" pattern that AGENTS.md prohibits by name.

To be fair: this pattern was already there for `linq_signing_secrets` before this PR, and this PR is correctly aligning the three remaining channels with the established pattern. The actual DRY violation predates this change. I am not blocking on it here — the PR makes things no worse — but this warrants a follow-up issue to read signing secrets from `AppState.config` at handler time rather than caching them at startup. If a user rotates a webhook secret in their config and sends `SIGHUP`, the handler would continue verifying against the stale cached secret until the process restarts.

---

### 🟢 What looks good — `api_webhook::resolve` is clean and well-abstracted

The `Resolved<'a, T>` enum (Alias/Fallback/NotFound) cleanly separates the three routing outcomes without copying data, and the generic lifetime parameter means it works for both channel instances and signing secrets without a second implementation. The manual `Copy`/`Clone` impls with the explicit comment about not requiring `T: Copy` are the right call.

---

### 🟢 What looks good — Security posture is correct throughout

The 404 response for unknown aliases is generic (`"No matching {channel_type} webhook target configured"`) — it does not disclose which aliases are configured. Per-alias signing secrets are resolved by the alias key after routing, so a request for `/whatsapp/personal` cannot verify against `work`'s secret. The `whatsapp_signature` auth test covers: unknown alias → 404 before any processing, configured alias + missing signature → 401, configured alias + valid signature → 200. That's the right boundary coverage for an inbound auth path.

---

### 🟢 What looks good — Alias isolation proved by test design

`webhook_alias_routes_to_the_matching_instance` constructs two WhatsApp instances with distinct verify tokens (`tok-work`, `tok-personal`) and asserts that the wrong token fails against the `work` alias. This is the correct way to prove isolation — not just that the right alias succeeds, but that the other alias's credential is actively rejected.

---

### 🟢 What looks good — Conflict resolution with upstream linq alias PR is sound

The author's explanation of the beta-2 conflict is clear: upstream landed a linq-only `{alias}` path without a bare fallback; this PR subsumes that, routes it through the shared `resolve()` helper, and restores the bare fallback. The upstream `linq_test_state` helper is preserved and the 255-test suite passes post-rebase. The reconciliation is thorough.

---

### Minor notes

- The `deprecation_header()` function contains `.unwrap_or_else(|_| HeaderValue::from_static("bare webhook path is deprecated"))` — this is a safe `unwrap_or_else`, not a panicking path; the fallback value is valid ASCII. No AGENTS.md concern here.
- Full-workspace `cargo test` was not run (only `cargo test -p zeroclaw-gateway`). For a gateway change of this scope I would normally ask for it, but the author ran `cargo build --workspace` clean and all 255 per-crate tests pass. Acceptable for a gateway-isolated change.
- Validation image linked in the PR description (`route_tests.png`) is from the author's own GitHub fork, not the canonical repo. This is fine for a screenshot but worth noting if the image ever rots.

---

**Verdict:** `--request-changes`

The `risk: high` / `size: L` labels must be applied before merge so automated gating works correctly; the CHANGELOG/docs wording around bare-path determinism should be corrected. Everything else is polished work and I'd be ready to approve immediately after those are addressed.

Audacity88 (changes requested):
I reviewed the current PR diff, public PR metadata and checks, linked #6312, and the prior review/comment thread. I did not run additional local validation for this docs-focused re-review. Thanks for the reroll; the core routing shape is much clearer now.

### ✅ Resolved — Bare-path fallback is deterministic now

The shared `api_webhook::resolve()` helper fixes the earlier HashMap-order concern by making the deprecated bare path pick the lexicographically-first alias, and the handlers now reuse that resolved alias for the per-instance signing secret lookup. The added route/test coverage around explicit aliases, unknown aliases, fallback deprecation headers, and signature paths is the right shape.

### 🔴 Blocking — Channel docs still describe a different fallback rule

The implementation and changelog now say bare `/<type>` fallback resolves to the lexicographically-first alias, but the channel docs still tell operators that bare `/whatsapp` and `/nextcloud-talk` resolve to the "first configured instance." For a multi-instance config, those are different rules: reordering config can change what "first configured" means, while the runtime will keep selecting the smallest alias. Please update the WhatsApp and Nextcloud Talk docs to describe the lexicographically-first fallback, or otherwise align the docs and implementation on one rule.

IftekharUddin:
Addressed the review feedback in f3ce32c5a: the deprecated bare /<type> fallback now resolves to the lexicographically-first alias instead of HashMap::iter().next(), so it's deterministic across restarts for multi-instance configs, and I updated the doc comments + CHANGELOG to match and added a multi-instance test proving the selection is stable regardless of insertion order (fmt/clippy -D warnings/tests all clean). I don't have permission to set labels on this repo — could a maintainer apply risk: high and size: L? (there's no type: feature label; enhancement is the closest). On the ConfigSnapshot warning, agreed it's pre-existing and out of scope here — happy to open a follow-up issue to read signing secrets from AppState.config at handler time rather than caching them at startup.

WareWolf-MoonWall (commented):
Re-reviewing at 4d7c2e1d against the latest commit (f3ce32c5a), IftekharUddin's 2h response, and Audacity88's CHANGES_REQUESTED at 13dd1f00. The two items I blocked on are resolved; Audacity88's blocker appears addressed in the latest commit — tagging them to re-review and formally clear their CR.

---

### ✅ Resolved — `risk: high` and `size: L` labels (my prior 🔴 Blocking)

Both labels are now applied. Automated gating and queue prioritization will work correctly. Thank you for getting the maintainer to set these — the note in your 2h comment explaining you lacked label permissions is exactly the right way to handle it.

### ✅ Resolved — Bare-path fallback is now deterministic (my prior 🟡 Warning)

Commit f3ce32c5a replaces `HashMap::iter().next()` with a sorted key selection that guarantees the lexicographically-first alias across process restarts. The new `webhook_bare_path_resolves_deterministically_for_multi_alias` test uses a two-entry map constructed with `work` inserted before `personal` and verifies that `personal` (lex-first) wins regardless of insertion order. The CHANGELOG entry and the `api_webhook.rs` module doc comment both now say "lexicographically-first alias" consistently. This is the right fix and the right level of proof.

### ✅ Resolved — Channel docs updated (Audacity88's 🔴 Blocking)

The latest commit includes changes to `docs/book/src/channels/whatsapp.md` and `docs/book/src/channels/nextcloud-talk.md`. These are the exact files Audacity88 flagged. @Audacity88 — please re-review the current head (4d7c2e1d) and dismiss your CHANGES_REQUESTED if the wording in those doc files now accurately reflects the lexicographically-first fallback rule. That is the only formal block remaining on this PR.

### 🔵 Suggestion — File a follow-up issue for the ConfigSnapshot signing-secret pattern

Appreciate the offer to file a follow-up issue for reading signing secrets from `AppState.config` at handler time rather than caching them at startup. Please do — a brief issue linking back to this PR and naming `whatsapp_app_secret`, `nextcloud_talk_webhook_secret`, `linq_signing_secrets`, and the pre-existing `wati` equivalent as the four fields to migrate would give the next contributor a clean target. The AGENTS.md DRY rule makes this a genuine technical debt item, not just a style note.

---

**Verdict:** Commenting. My prior blockers are resolved and Audacity88's doc concern appears addressed in the latest commit. I would approve once @Audacity88 re-reviews and formally clears their CHANGES_REQUESTED — the code is otherwise in good shape.

docs/book/src/channels/whatsapp.md
```diff
@@ -24,6 +24,14 @@ access_token = "<meta-access-token>"
 
 The gateway must be reachable by Meta for inbound webhooks. Configure a tunnel via `zeroclaw config set gateway.tunnel.provider=<ngrok|cloudflare>` and the related `gateway.tunnel.*` fields (see the [Config reference](../reference/config.md)), or front the gateway with your own reverse proxy when developing locally.
 
+Point Meta's Callback URL at the alias of the `[channels.whatsapp.<alias>]`
+instance that should receive it — `GET`/`POST https://<your-public-url>/whatsapp/<alias>`
+(e.g. `[channels.whatsapp.work]` → `/whatsapp/work`). This per-alias routing
+(#6312) lets multiple WhatsApp numbers run side by side. The bare `/whatsapp`
+path still works but is **deprecated**: it resolves to the first configured
```
singlerider: This says "first configured instance" but the code resolves to the lexicographically-first alias (sorted keys, deterministic across restarts). For a multi-instance config these are different rules: "first configured" follows config ordering, while the runtime always picks the smallest alias regardless of order. Please align to the lexicographically-first rule to match the CHANGELOG and the `api_webhook.rs` module doc.

docs/book/src/channels/nextcloud-talk.md
```diff
@@ -39,12 +39,22 @@ Full field reference: [Config](../reference/config.md).
...
-https://<your-public-url>/nextcloud-talk
+https://<your-public-url>/nextcloud-talk/<alias>
 ```
 
+For example, `[channels.nextcloud_talk.work]` receives `POST /nextcloud-talk/work`.
+This per-alias routing (#6312) lets you run several Talk bots side by side and
+deliver each one's webhooks to the right instance.
+
+The bare `https://<your-public-url>/nextcloud-talk` path still works but is
+**deprecated**: it resolves to the first configured instance and returns an
```
singlerider: Same mismatch as whatsapp.md: "first configured instance" should read "lexicographically-first alias" to match the code, the CHANGELOG, and the module doc. The bare-path fallback sorts keys and selects the smallest alias deterministically, independent of config order.

singlerider (changes requested):
Re-reviewing at 4d7c2e1d. I checked the two channel doc files that @Audacity88 blocked on against the current head, and the mismatch they flagged is still present, so I'm reinforcing their `CHANGES_REQUESTED` rather than clearing it.

@WareWolf-MoonWall's read at f3ce32c5a was that the doc commit addressed this, but the commit touched these files without updating the specific fallback-rule wording. The CHANGELOG and `api_webhook.rs` module doc are correct ("lexicographically-first alias"); the two operator-facing channel docs still say "first configured instance."

### 🔴 Blocking: channel docs still describe a different fallback rule than the code (reinforces @Audacity88)

The bare `/<type>` path resolves to the **lexicographically-first alias** (sorted keys, deterministic across restarts, proven by `webhook_bare_path_resolves_deterministically_for_multi_alias`). But `whatsapp.md` and `nextcloud-talk.md` still tell operators it resolves to the "first configured instance."

For a multi-instance config these are different rules: "first configured" depends on config ordering, while the runtime always selects the smallest alias regardless of order. An operator who reads the docs, reorders their config expecting the fallback to follow, and finds it doesn't, has been misled by the docs. That is exactly the surprise @Audacity88 called out.

I've left inline comments on the two exact lines. Please align both to the lexicographically-first rule (matching the CHANGELOG and module doc), after which @Audacity88 can re-review and clear their block.

Everything else on this PR is in good shape. The routing helper, alias isolation, security posture, and bare-path determinism are all solid. This is the last wording fix standing between the PR and merge.

---

**Verdict:** `--request-changes`, reinforcing @Audacity88's active block. The code is correct; the two channel docs need to match it.

docs/book/src/channels/nextcloud-talk.md
IftekharUddin: I believe [ea2ed13](https://github.com/zeroclaw-labs/zeroclaw/commit/ea2ed13847e91630f6bfe25ece14c54fc6126243) has now addressed this! :D

Audacity88 (approved):
Context checked: current head `001c5b7c`, the `ea2ed138` docs update, live labels and milestone, visible CI, and the prior review thread. I did not rerun local validation for this docs-focused re-review.

### ✅ Resolved — Channel docs now match the fallback rule

The WhatsApp and Nextcloud Talk docs now point operators at the alias-qualified webhook URLs and describe the bare path as a deprecated fallback that resolves to the lexicographically-first alias, sets `X-Zeroclaw-Deprecation`, and returns `404` for unknown aliases. That matches the implementation, the `api_webhook.rs` module doc, and the changelog entry, so my earlier docs blocker is resolved.

### 🟢 What looks good — The routing contract is coherent now

The PR now has a consistent story across code, tests, changelog, and operator docs: explicit `/<type>/<alias>` routing for multi-instance deployments, deterministic bare-path compatibility for single-instance or transitional deployments, and per-alias signing-secret lookup after route resolution. That is the right shape for this gateway boundary.

### 🔵 Suggestion — Refresh the PR body before merge

One public-record cleanup still remains: the PR body still describes the bare fallback as the "first instance" and still lists `risk: high` / `size: L` as suggested labels even though those labels are now applied. I am not treating that as author action, but the body should be refreshed before merge so the public summary matches the current code and metadata.

### 🔵 Suggestion — Track signing-secret snapshot cleanup

I agree with the follow-up direction the author offered: reading the signing secrets from `AppState.config` at handler time would avoid carrying the existing startup snapshot pattern forward. That is still worth tracking separately, but it is not blocking this PR.

This approval clears my earlier `CHANGES_REQUESTED`. It does not clear @singlerider's active `CHANGES_REQUESTED`; that review remains in force until they re-review, approve, or dismiss it, so this is not a merge-ready statement.

WareWolf-MoonWall (commented):
Re-reviewing at 4e12721310b3. Checking the state of singlerider's active CHANGES_REQUESTED against what Audacity88 confirmed at the current head.

---

### ✅ Resolved — Channel docs now match the fallback rule (singlerider's 🔴 / Audacity88's 🔴)

Commit ea2ed138 updated `docs/book/src/channels/whatsapp.md` and `docs/book/src/channels/nextcloud-talk.md`. @Audacity88 re-reviewed the current head (4e12721310b3) and confirmed: "The WhatsApp and Nextcloud Talk docs now point operators at the alias-qualified webhook URLs and describe the bare path as a deprecated fallback that resolves to the lexicographically-first alias." That is the exact correction singlerider and Audacity88 called for, and it is now in the code.

@singlerider — your CHANGES_REQUESTED was posted at 4d7c2e1d before the ea2ed138 docs fix landed. Please re-review the current head (4e12721310b3) and dismiss your block if the doc wording satisfies the concern. Audacity88 has cleared theirs; your re-review is the only formal gate remaining.

---

### 🔵 Suggestion — Refresh PR body before merge (echoing Audacity88)

The PR body still describes the bare fallback as "first instance" and lists `risk: high` / `size: L` as "suggested" labels even though they are now applied. Neither item blocks merge, but a body refresh before squash would make the public record accurate.

---

**Verdict:** Commenting. All substantive concerns — labels, HashMap determinism, channel docs wording — are resolved at the current head. @Audacity88 has approved. I would approve immediately once @singlerider re-reviews and clears their CHANGES_REQUESTED.

singlerider (approved):
Re-reviewed at head `84942d0`. My one blocker — the channel-doc fallback wording — is resolved. Converting my changes-requested to an approval, with two non-blocking notes.

### RESOLVED ✅ — channel docs now match the lexicographically-first fallback rule

Both operator docs are corrected at this head:

- `whatsapp.md`: the bare path "is **deprecated**: it resolves to the lexicographically-first alias".
- `nextcloud-talk.md`: bare `/nextcloud-talk` is "a deprecated fallback" that "resolves to the lexicographically-first alias (deterministic...)".

That matches the implementation (`webhook_bare_path_resolves_deterministically_for_multi_alias`), the `api_webhook.rs` module doc, and the changelog. The "first configured instance" wording that contradicted the code — and would have misled an operator who reordered config expecting the fallback to follow — is gone. This was the only thing my review gated on, and @Audacity88 has already cleared his matching block.

### 🟢 The routing contract is coherent

Explicit `/<type>/<alias>` routing for multi-instance, deterministic lexicographically-first bare-path fallback with the `X-Zeroclaw-Deprecation` header and `404` for unknown aliases, and per-alias signing-secret lookup after route resolution — code, tests, changelog, and operator docs now tell one consistent story. The routing helper, alias isolation, and security posture are solid.

### 🔵 Two non-blocking items before the squash

1. **`CHANGELOG-next.md` is hand-edited in this PR.** That file is regenerated from commit history by the maintainer changelog skill at release time, so a contributor edit will collide with the generated output — better to drop it from the diff and let the changelog tooling pick the entry up. Not gating my approval.
2. **The branch is `CONFLICTING` / `DIRTY` against current `master`** and needs a refresh before it can actually merge, even though the visible CI on this head is green. Also refresh the PR body (it still says "first instance" and lists the now-applied labels as "suggested").

Approving — this clears the last formal review gate. The conflict and the changelog/body cleanup are merge-sequencing items, not review blockers.

WareWolf-MoonWall (approved):
Re-review at head 84942d0b. I stated in my 2026-06-10 comment that I would approve once @singlerider re-reviewed and cleared their CHANGES_REQUESTED. @singlerider APPROVED today (2026-06-15), @Audacity88 APPROVED on 2026-06-09, and both of my original blocking items are confirmed resolved. Approving.

### ✅ Resolved — `risk: high` and `size: L` labels applied

Both labels are now on the PR.

### ✅ Resolved — HashMap alias selection is deterministic

`resolve()` uses `min_by(|(a,_),(b,_)| a.cmp(b))` — lexicographically-first alias, deterministic across process restarts. Module docs, CHANGELOG, tests, and operator channel documentation (`whatsapp.md`, `nextcloud-talk.md`) all describe the deterministic selection rule consistently.

### 🟢 What looks good — Alias-routing architecture

`Resolved<T>` + `resolve()` as an independent helper is clean — the single-alias and multi-alias cases are handled uniformly, the 404 path is explicit, and the per-alias HMAC secret keying (vs. the old single `Option<String>`) is a meaningful security improvement.

### 🟢 What looks good — Test coverage

Tests cover alias routing, unknown-alias 404, bare-path deprecation header, and per-alias HMAC. The operator documentation is aligned with the implementation.

### 🟡 Warning — Branch is dirty (conflicted with master)

The mergeable state is `dirty`. A rebase against master is required before the squash merge. singlerider flagged this in their 2026-06-15 approval comment.

### 🟡 Warning — `CHANGELOG-next.md` hand-edit should be dropped

singlerider noted that the `CHANGELOG-next.md` entry added in this PR will collide with the maintainer changelog tooling at release time. Please drop that file change from the branch — the release notes for this work will be generated via the standard changelog process.

Issue #6298, opened 2026-05-03
Empty tool_calls array sent to provider API, causing 400 on strict validators (DeepSeek, NVIDIA NIM)


Svtter:
## Bug Description

When `use_native_tools` is enabled and a model response contains no tool calls, `build_native_assistant_history_from_parsed_calls()` generates an assistant message with `"tool_calls": []` (empty array). Providers that strictly validate the `tool_calls` field reject this with a 400 error.

## Error Example

```
Invalid 'messages[6].tool_calls': empty array. Expected an array with minimum length 1
```

This was observed with **DeepSeek V4** (thinking mode), but the same class of error affects other strict providers.

## Root Cause

In `crates/zeroclaw-runtime/src/agent/loop_.rs`, the code path around `build_native_assistant_history_from_parsed_calls()` produces:

```json
{"content": "...", "tool_calls": [], "reasoning_content": "..."}
```

even when no tool calls are present. The condition that decides whether to use native tool history format checks `if use_native_tools` but should also check `!calls.is_empty()` — if there are no tool calls, the `tool_calls` field should be omitted entirely (or not use the native history format).

## Relevant Code

`crates/zeroclaw-runtime/src/agent/loop_.rs` — the `assistant_history_content` construction where `use_native_tools` determines whether to call `build_native_assistant_history_from_parsed_calls()`.

## Related Issues

This may be the root cause (or a contributing factor) for several existing issues that report provider 400 errors around tool_calls validation:

- #5475 — Copilot + Telegram: `messages with role 'tool' must be a response to a preceeding message with 'tool_calls'` — an empty `tool_calls: []` would cause the provider to not recognize the assistant message as having tool_calls, orphaning subsequent tool messages
- #2405 — NVIDIA NIM: `Message has tool role, but there was no previous assistant message with a tool call!` — same class of error
- #5941 — `No tool call found for function call output with call_id` — provider sees no tool calls in the assistant message
- #5584 — Duplicate consecutive assistant messages when model returns narration text alongside tool calls — the text-only copy may carry an empty `tool_calls` array
- #2232 — DeepSeek 400 after auto-compaction: orphan tool messages — compaction may leave empty tool_calls arrays
- #5825 — `strip_native_tool_messages` can produce consecutive assistant messages — stripping may leave empty tool_calls instead of removing the field entirely

Svtter:
## Additional Note: Existing Sessions Retain Stale Data

[...]

PR #5242, opened 2026-04-03
fix(providers): disable native tool calls for custom endpoints

bellman-move:
## Summary

Describe this PR in 2-5 bullets:

- Base branch target (`master` for all contributions): `master`
- Problem: custom OpenAI-compatible endpoints were using native tool calling, which caused strict providers to reject persisted Slack tool-result history with `No tool call found for function call output with call_id ...` errors.
- Why it matters: long-lived Slack sessions could repeatedly fail after tool-heavy turns, even when the underlying task was otherwise valid.
- What changed: custom `custom:https://...` providers now disable native tool calling and fall back to prompt-guided tool execution; added a regression test covering the factory behavior.
- What did **not** change (scope boundary): no Slack channel transport logic, no session persistence format, no upstream OpenAI/OpenRouter provider behavior.

## Label Snapshot (required)

- Risk label (`risk: low|medium|high`): `risk: medium`
- Size label (`size: XS|S|M|L|XL`, auto-managed/read-only): `size: XS`
- Scope labels (`core|agent|channel|config|cron|daemon|doctor|gateway|health|heartbeat|integration|memory|observability|onboard|provider|runtime|security|service|skillforge|skills|tool|tunnel|docs|dependencies|ci|tests|scripts|dev`, comma-separated): `provider,tests`
- Module labels (`<module>: <component>`, for example `channel: telegram`, `provider: kimi`, `tool: shell`): `provider: custom`
- Contributor tier label (`trusted contributor|experienced contributor|principal contributor|distinguished contributor`, auto-managed/read-only; author merged PRs >=5/10/20/50): auto-managed
- If any auto-label is incorrect, note requested correction: N/A

## Change Metadata

- Change type (`bug|feature|refactor|docs|security|chore`): `bug`
- Primary scope (`runtime|provider|channel|memory|security|ci|docs|multi`): `provider`

## Linked Issue

- Closes #
- Related #
- Depends on # (if stacked)
- Supersedes # (if replacing older PR)

## Supersede Attribution (required when `Supersedes #` is used)

- Superseded PRs + authors (`#<pr> by @<author>`, one per line): N/A
- Integrated scope by source PR (what was materially carried forward): N/A
- `Co-authored-by` trailers added for materially incorporated contributors? (`Yes/No`): `No`
- If `No`, explain why (for example: inspiration-only, no direct code/design carry-over): no superseded PR
- Trailer format check (separate lines, no escaped `\n`): (`Pass/Fail`): `Pass`

## Validation Evidence (required)

[...]

PR #7370, opened 2026-06-08
fix(channels): drop truncated tool-call envelopes at delivery, keep the reply

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - When a model is cut off **mid tool-call** it emits an *unterminated* tagged
    envelope, e.g. `<function_calls><invoke name="shell"><parameter name="command">…`
    with no closing tags. The channel delivery cleaner `strip_tool_call_tags`
    recovered unterminated **JSON** tool calls but fell through on the **XML**
    invoke/parameter form and **leaked the raw envelope into the channel**.
  - Fix: when an unterminated opener is followed by tool-call structure **and that
    structure runs to the end of the message** (a genuine truncation), drop the
    broken tail while keeping the preceding prose. If natural-language prose
    resumes after the structure, the model is showing an *example* rather than
    making a call, so the text is kept verbatim.
  - Applied to **both** delivery cleaners — `channels::util::strip_tool_call_tags`
    and the `orchestrator` copy.
  - Observed with MiniMax-family models in text-tool mode emitting Anthropic-style
    `<function_calls>` XML cut off after a long `<think>` block — gateway-independent,
    so it affects any such user, not only one gateway.
- **Scope boundary:** delivery-layer text cleaning only. It does **not** touch the
  malformed-tool detection / retry / suppression path, so legitimate replies are
  never suppressed or replaced with a placeholder.
- **Blast radius:** only ever *removes* text from the opener tag onward — it never
  garbles or rewrites the prose *before* the tag, and never routes through the
  suppression/retry path. Normal replies, prose that merely mentions a tag, and
  tags used as words are unaffected (see probe below).
- **Linked issue(s):** Related #7244 (parser hardening for malformed `<tool_call>`
  JSON args — a different failure mode). The native-tools alternative is ruled out
  for gateway endpoints — Related #5242, Related #6298. No existing issue tracks
  this specific delivery-layer leak; it was found via dogfooding.
- **Labels:** expect `channels`, `type: fix`, `risk: low`, `size: S`.

### Documented trade-off

The example-vs-truncation boundary is fundamentally ambiguous for unterminated
input (an inline `<function_calls><invoke…` example is byte-identical to a real
cut-off call). This PR biases toward **keeping** when unsure: a genuine truncated
tool call whose tail happens to read like a sentence (e.g. a cut-off `<parameter>`
value that is itself an English sentence) will leak that small fragment rather than
risk deleting a user's reply. That is a quieter, rarer failure than silently eating
content, and the surrounding prose is preserved either way. Optional future
tightening: also skip tags inside fenced code blocks.

Change is confined to one crate (`zeroclaw-channels`), so the battery is run scoped
to it (plus workspace-wide `fmt`). Literal tails:

```text
$ cargo fmt --all -- --check
# clean (exit 0)

$ cargo clippy -p zeroclaw-channels --features channel-discord --all-targets -- -D warnings
# clean (exit 0, no warnings)

$ cargo test -p zeroclaw-channels --features channel-discord strip_
test util::tests::strip_drops_truncated_function_calls_envelope_keeps_prose ... ok
test util::tests::strip_keeps_prose_that_merely_mentions_a_tag ... ok
test util::tests::strip_keeps_unterminated_example_followed_by_prose ... ok
test util::tests::strip_still_drops_genuine_truncation_to_end ... ok
test result: ok. 42 passed; 0 failed; 0 ignored; 0 measured; 763 filtered out
```

- **Commands run and tail output:** as above.
- **Beyond CI — what did you manually verify?** Ran a probe of `strip_tool_call_tags`
  against realistic legitimate messages:

  ```text
  plain prose ............................... unchanged              ✅
  prose mentioning <function_calls> ......... unchanged              ✅
  <tool> used as a word ..................... unchanged              ✅
  <function_calls> below. + a <note> line ... unchanged              ✅
  unterminated <invoke> example + prose ..... kept (prose resumes)   ✅
  "Emit <tool_call> {schema} in docs." ...... kept (prose resumes)   ✅
  real truncated call (cut off mid-value) ... tail dropped, prose kept ✅
  mid-tag cutoff "<invoke name=\"sh" ........ tail dropped, prose kept ✅
  complete fenced <function_calls> example .. XML removed (pre-existing strip)
  ```

  Also built a release with the change and ran it against a live instance whose
  MiniMax-via-gateway agent had been leaking `<function_calls>` XML to a channel;
  confirmed truncated envelopes no longer leak and that normal replies (including
  ones that *discuss* tool-call syntax) render unchanged.
- **If any command was intentionally skipped, why:** the full-workspace `cargo test`
  / `cargo clippy` were not run locally; the change is isolated to delivery-layer
  string handling in `zeroclaw-channels`, and CI runs the full battery.

## Compatibility (required)

- Backward compatible? `Yes` — only strips envelopes that previously leaked; valid
  output (complete tool calls, JSON envelopes, ordinary prose) is unchanged.
- Upgrade steps: none.

Low risk: `git revert <sha>` — two delivery-cleaner functions plus tests.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -615,6 +615,58 @@ pub(crate) fn strip_tool_call_tags(message: &str) -> String {
         }
     }
 
+    // Does the tag structure run to the end of the message? A *real* truncated
+    // tool call is the model getting cut off, so the unterminated structure is
+    // the last thing in the message. If natural-language prose resumes after the
+    // tags, this is an inline *example* (the model is discussing tool calls), not
+    // a truncation — so we should keep it. Bias toward keeping: a little leaked
+    // XML beats eating the user's text.
+    fn tool_structure_runs_to_end(inner: &str) -> bool {
+        let mut rest = inner.trim_start();
+        while rest.starts_with('<') {
+            match rest.find('>') {
+                Some(gt) => rest = rest[gt + 1..].trim_start(),
+                None => return true,
+            }
+        }
+        let tail = rest.trim();
+        if tail.is_empty() {
+            return true;
+        }
+        !looks_like_prose(tail)
+    }
+
+    // Heuristic: does `text` read like resumed natural-language prose (as opposed
+    // to a cut-off parameter value)? True on an internal sentence boundary
+    // (". " / "! " / "? " + a letter) or a multi-word string that ends like a
+    // sentence. Deliberately lenient so ambiguous tails are kept, not dropped.
+    fn looks_like_prose(text: &str) -> bool {
+        let bytes = text.as_bytes();
+        for i in 0..bytes.len().saturating_sub(1) {
+            if matches!(bytes[i], b'.' | b'!' | b'?')
+                && matches!(bytes[i + 1], b' ' | b'\n' | b'\t')
+                && text[i + 1..]
+                    .trim_start()
+                    .chars()
+                    .next()
+                    .is_some_and(|c| c.is_alphabetic())
+            {
+                return true;
+            }
+        }
+        let trimmed = text.trim_end();
+        let ends_like_sentence = trimmed
+            .chars()
+            .last()
+            .is_some_and(|c| matches!(c, '.' | '!' | '?'))
+            && trimmed
+                .chars()
+                .rev()
+                .nth(1)
+                .is_some_and(|c| c.is_alphabetic());
+        ends_like_sentence && text.trim().contains(' ')
+    }
+
     let mut kept_segments = Vec::new();
     let mut remaining = message;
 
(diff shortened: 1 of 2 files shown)


Audacity88 (approved):
Reviewing current head c70a5ebcaf764057378d32508c834210d9c40cfb. I checked the live PR state, PR body, changed files, formal reviews, inline review comments, the two delivery cleaner implementations, the added tests, current `master` drift, and the visible GitHub checks. There are no prior formal reviews or inline comments. I did not run local cargo or a separate live-channel smoke pass; I am relying on the author's focused crate validation, the reported live MiniMax-channel probe, and the green GitHub checks for the reviewed head.

### 🟢 What looks good — The EOF boundary fits this failure mode

The fix draws the right line for an unterminated XML-style tool-call envelope: if the tag structure runs to the end of the delivered text, drop the broken tail and keep any prose before it; if sentence-shaped prose resumes after the tag structure, keep the text as an example instead of eating user-visible content. That matches the actual failure mode, where the model is cut off mid-call, while preserving the resumed-prose examples covered by this patch.

The new tests cover the important edges: truncated `<function_calls><invoke ...><parameter ...` tails are removed while preceding prose survives, envelope-only output becomes empty, complete envelopes still strip through the existing path, ordinary prose that merely mentions `<function_calls>` is kept, inline examples with resumed prose are kept, and mid-tag cutoffs are removed.

### 🟢 What looks good — Both delivery cleaners get the same behavior

The PR applies the same EOF/truncation heuristic to `channels::util::strip_tool_call_tags` and the orchestrator-side `strip_tool_call_tags`, so Discord/WeChat-style utility callers and the generic orchestrator sanitization path stay aligned. Given the current duplicated implementation shape, mirroring the fix in both places is the right immediate move.

### 🔵 Suggestion — Keep an eye on the duplicated cleaner

Not a blocker for this patch, but this PR is another reminder that the delivery cleaner now exists in two places. The duplicated code is intentionally kept aligned here; a future cleanup that makes the utility helper the single implementation, or adds direct sanitizer-level coverage for this exact truncated XML case, would reduce the chance that the two paths drift again. The heuristic also still relies on sentence-shape detection, so future coverage for shorter or unpunctuated inline examples would be useful.

### 🟡 Warning — Refresh merge evidence before landing

This approval is for the reviewed head. It is not a merge-readiness sign-off for the final integration result. The branch is cleanly mergeable, but current `master` has moved substantially since this PR's base and includes several channel/orchestrator changes, including a newer channel response sanitizer fix in the same file. Before merge, refresh official CI on an updated head or run an equivalent merge-result smoke that covers the channel sanitization path.

Approving. The delivery-layer fix is scoped, preserves the tested resumed-prose cases, covers the core truncation cases, and has a clear fallback/rollback story.

PR #7371, opened 2026-06-08
fix(config): unbounded/yolo presets must allow what they advertise

JordanTheJet:
## Summary

- **What changed and why:** Two Quickstart presets used a sentinel value
  that means the *opposite* of intended, leaving a freshly-onboarded agent
  unable to do anything:
  - **`unbounded` runtime preset** set `max_actions_per_hour: 0` and
    `max_cost_per_day_cents: 0`, on the assumption `0` = inherit/unlimited.
    It does not: the per-sender rate tracker treats a max of `0` as *always
    exhausted* (`PerSenderTracker::is_exhausted`, locked by
    `rate_limit_zero_blocks_everything`), so every tool call was rejected
    with "Rate limit: max 0 actions per hour per chat". Changed to
    `u32::MAX` for an effectively-unlimited budget.
  - **`yolo` risk preset** set `allowed_commands: vec![]`, intending "no
    command denylist / full autonomy". But the command gate is
    deny-by-default — `is_command_allowed` rejects any command not matched
    by an allowlist entry — so an empty list blocked *every* shell command
    ("Command not allowed by security policy: ls" for even `ls`/`pwd`).
    Changed to the `["*"]` wildcard, which (with `block_high_risk_commands:
    false`) hits the trusted-env path that actually grants unrestricted
    execution.
  - Corrected the `max_actions_per_hour` schema doc that wrongly claimed
    `0` inherits a global limit — no such inheritance is implemented.
- **Scope boundary:** Preset values + one doc comment only. No change to the
  rate-tracker / command-gate enforcement logic itself (the `0 = exhausted`
  and empty-allowlist-denies behaviours are deliberate security primitives
  relied on elsewhere — e.g. subagent lockdown, cron, shell sandbox — and
  are intentionally left intact). No change to the `balanced`/`tight`/
  `locked_down` presets, the Quickstart UI, or the wire/RPC surface.
- **Blast radius:** `crates/zeroclaw-config` presets. Any agent created on
  the `unbounded` runtime profile or `yolo` risk profile — via Quickstart
  (all surfaces) or hand-authored config that copies these preset shapes.
  PR #7330 forces every Quickstart agent onto `unbounded`, so the action
  budget fix in particular unblocks 100% of new Quickstart agents.
- **Linked issue(s):** None on file — this appears unreported. Related
  #7330 (forces the `unbounded` runtime profile, which is what surfaced the
  zero-action-budget defect for every new Quickstart agent).
- **Labels:** `bug`, `config`, `risk: medium`, `size: S`.

```bash
cargo fmt --all -- --check
cargo test
```

`cargo fmt --all -- --check` — clean, exit 0 (no diff).

`cargo clippy -p zeroclaw-config --all-targets -- -D warnings`:
```
    Checking zeroclaw-config v0.8.0-beta-2 (/Users/jordantian/zeroclaw/crates/zeroclaw-config)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 09s
```
exit 0, no warnings.

`cargo test -p zeroclaw-config`:
```
test presets::tests::unbounded_runtime_does_not_block_all_actions ... ok
test presets::tests::yolo_risk_allows_shell_commands ... ok
test policy::tests::rate_limit_zero_blocks_everything ... ok
test result: ok. 772 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.57s
test result: ok. 89 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.22s
```

- **Beyond CI — scope note:** The diff is confined to `crates/zeroclaw-config`,
  so clippy/test were scoped to `-p zeroclaw-config` (the full relevant
  battery) rather than the whole workspace. `cargo fmt --all` ran
  workspace-wide. The `rate_limit_zero_blocks_everything` test still passes —
  the fix does NOT relax the `0 = block` security primitive; it stops the
  presets from depending on `0`/empty meaning "unlimited".
- **Beyond CI — what did you manually verify?** Built the daemon and ran a
  real Quickstart end-to-end against a fresh isolated config. Before:
  inspected the generated `config.toml` and confirmed the agent was dead on
  arrival (`max_actions_per_hour = 0`, every tool call rejected; on the
  `yolo` profile, `ls`/`pwd` rejected by the command gate). After the fix:
  the generated config carries `max_actions_per_hour = 4294967295` and the
  `yolo` profile carries `allowed_commands = ["*"]`; the agent successfully
  invokes tools (file/http) and runs shell commands (`ls`, `pwd`). Two added
  regression tests assert a policy built from each preset actually permits
  actions / commands.
- **If any command was intentionally skipped, why:** Whole-workspace clippy
  and `cargo test` were scoped to the changed crate per the scope note
  above; nothing else skipped.

- New permissions, capabilities, or file system access scope? `No` — the
  `yolo` preset already granted full autonomy with high-risk blocking off;
  this only makes its command allowlist actually function as documented.
  `balanced`/`locked_down` (the default-recommended and tightest presets)
  are unchanged.

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` — preset *values* changed; no
  schema field, env var, or CLI flag added or renamed.
- If `No` or `Yes` to either: No upgrade steps for the fix itself. Note:
  agents already onboarded onto a persisted `unbounded`/`yolo` profile keep
  their old (broken) values — Quickstart's "existing block wins" rule won't
  rewrite an existing profile. Such users either re-onboard into a fresh
  config or edit the persisted profile (`max_actions_per_hour` to a high
  value; `allowed_commands = ["*"]`).

- **Fast rollback command/path:** `git revert <merge-or-squash-sha>`.
  Self-contained to `crates/zeroclaw-config/src/presets.rs` (+ one schema
  doc comment).
- **Observable failure symptoms:** New Quickstart agents on `unbounded`
  rejecting every tool call with "max 0 actions per hour"; `yolo`-profile
  agents rejecting shell commands with "Command not allowed by security
  policy" for benign commands like `ls`/`pwd`.


theonlyhennygod (approved):
First review of head 7a2ba9ae5d5bea27eb2f92ee9ccae894facc0058. I read the PR body, the full diff (+67/−4 across `crates/zeroclaw-config/src/presets.rs` and `schema.rs`), and cross-checked every claim against the enforcement primitives in `crates/zeroclaw-config/src/policy.rs`. CI is fully green (15/15), no other reviewer holds a block. This is a clean, correctly-diagnosed bug fix — approving.

The diagnosis is exactly right, and I want to be specific about what I verified, because the value of this PR is that it fixes the *presets* without touching the *primitives* they depend on.

### 🟢 What looks good — Root-caused to the right layer, and the fix doesn't weaken the security primitives

Both sentinel bugs are real, and I confirmed each against the enforcement code rather than the description:

- **`unbounded` action budget.** `PerSenderTracker::is_exhausted` short-circuits `if max == 0 { return true; }` (`policy.rs:144`) — a max of `0` is *always exhausted*, not "unlimited". So an `unbounded` agent really did reject every tool call. `max_actions_per_hour: u32::MAX` flows through `record_action` → `record_within` (`count <= max`, `policy.rs:128`), which is now effectively always-true. Correct.
- **`yolo` command gate.** `is_command_allowed` is deny-by-default, but the wildcard + `!block_high_risk_commands` combination hits the trusted-env early return `if has_wildcard && !self.block_high_risk_commands { return true; }` (`policy.rs:1462`). Since `yolo` carries `block_high_risk_commands: false`, `allowed_commands: vec!["*"]` grants unrestricted execution unconditionally — which is exactly why your test can assert even `rm -rf node_modules` is permitted. Correct.

The part I most want to commend: the scope boundary holds the line in the right place. The `0 = exhausted` and empty-allowlist-denies behaviours are deliberate primitives (subagent lockdown, cron, shell sandbox lean on them), and you left them fully intact — `rate_limit_zero_blocks_everything` still passes. Fixing the preset to stop *depending on* a footgun, rather than softening the footgun, is the right instinct here and keeps the blast radius to preset values only.

### 🟢 What looks good — Tests assert behavior, not literal preset values

Both regression tests build an actual `SecurityPolicy` from the preset and exercise the enforced path (`policy.record_action()` permits; `policy.is_command_allowed(cmd)` permits a range of commands) rather than just asserting `max_actions_per_hour == u32::MAX`. That's the difference between a test that pins behavior and one that snapshots a constant — these will still mean something if the preset value is ever retuned, and they'd actually catch a regression in the enforcement path. The `yolo` test sweeping `ls`/`pwd`/`cat`/`rm -rf` is a nice touch because it proves the *trusted-env return*, not just a single allowlisted name.

### 🟢 What looks good — Correct handling of the adjacent schema doc

The `max_actions_per_hour` doc correction is accurate, and you correctly left the `max_tool_iterations` "0 inherits the global default" line directly above it untouched — because that one is true: `max_tool_iterations == 0` really does fall back to a safe default (`agent/loop_.rs:1420`, "safe fallback when … configured as zero"). Two adjacent budget fields with opposite zero-semantics is the exact footgun that caused this, and documenting only the one that lies is the precise fix.

### 🔵 Suggestion — The `max_cost_per_day_cents` change rides along untested

The action-budget fix is well covered; the cost change to `u32::MAX` is not, and it's worth a sentence on the record because its mechanism is different from what the body implies. `max_cost_per_day_cents` is *not* enforced through any `is_exhausted`-style path — the runtime `CostTracker` (`crates/zeroclaw-config/src/cost/tracker.rs`) checks a separate `daily_limit_usd`, not this field. So `cost: 0` was never a direct "dead on arrival" block the way `actions: 0` was. Where it *does* bite is subagent delegation: `ensure_no_escalation_beyond` rejects a child whose `max_cost_per_day_cents > parent` (`policy.rs:2264`), so under the old preset an `unbounded` parent (cost ceiling `0`) would reject any subagent that requested a non-zero cost budget with `MaxCostExceeded`. The fix correctly clears that too — I'd just (optionally) add a one-line test asserting an `unbounded`-derived policy doesn't reject a cost-bearing child via escalation, so the cost half of this fix has the same regression guard as the action half. Non-blocking.

### 🔵 Suggestion — `u32::MAX` as the "unlimited" sentinel

Using the type max as "effectively unlimited" is pragmatic and I wouldn't change it here — the clean alternative (`Option<u32>` with `None` = unlimited) is a schema change well outside this fix's scope. Flagging only so it's a conscious team choice on the record: the budget fields now encode "unlimited" as a magic max value rather than a first-class absent value. Fine to carry as-is; worth a tracked thought if the budget surface ever gets reworked.

## Merge readiness

CI is green across all 15 checks, the change is self-contained to `zeroclaw-config` presets + one doc comment, and no other reviewer holds an active block — `mergeStateStatus: BLOCKED` here is REVIEW_REQUIRED only (@Audacity88 + @singlerider still requested). Rollback is a clean `git revert`. Approving on substance; both 🔵s are optional and neither should hold the merge.

PR #7379, opened 2026-06-08
fix(gateway): align webhook model labels with resolved agent config

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Route webhook fallback agent selection through `Config::resolved_runtime_agent_alias()`.
  - Resolve the webhook `model` label from the selected agent's configured provider instead of the gateway startup model.
  - Return that resolved model in the `/webhook` success payload so the response matches the observer events for `?agent=<alias>` requests.
  - Add focused regression coverage for explicit-agent webhook model reporting and observer metadata.
- **Scope boundary:** This slice is limited to the gateway webhook agent/model resolution path and its focused regression test.
- **Blast radius:** `/webhook` response metadata and webhook observability events (`AgentStart` / `AgentEnd`) for explicit-agent and fallback-agent requests.
- **Linked issue(s):** Related #7297
- **Labels:** `bug`, `risk: high`, `size: S`, `gateway`

```text
zc-cargo test -p zeroclaw-gateway webhook_explicit_agent_reports_agent_model -- --nocapture
  test tests::webhook_explicit_agent_reports_agent_model ... ok

zc-cargo fmt -p zeroclaw-gateway -- --check

git diff --check
  No output.

Lint
  pass

Test
  pass

Security
  pass

CI Required Gate
  pass
```

- **Beyond CI — what did you manually verify?** Reviewed the explicit-agent and fallback webhook resolution path to confirm the gateway now uses the config-level runtime alias resolver and returns the same resolved model label in both observer events and the HTTP success payload.
- **If any command was intentionally skipped, why:** I kept local validation focused on the changed gateway path and relied on green GitHub checks for broader workspace coverage on this PR.

- If any `Yes`, describe the risk and mitigation: `Not applicable`

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: `No upgrade steps needed`

- **Fast rollback command/path:** `git revert ab4a9d5f5b489d6f6d72117ae09429f94eb95729`
- **Observable failure symptoms:** `/webhook?agent=<alias>` responses or observer events report the gateway startup/default model instead of the resolved agent model, or fallback-agent webhook metadata disagrees with the runtime default-agent selection.

### DIFF
--- a/crates/zeroclaw-gateway/src/lib.rs
+++ b/crates/zeroclaw-gateway/src/lib.rs
@@ -2163,38 +2163,7 @@ async fn run_gateway_chat_with_tools(
     #[cfg(not(test))]
     {
         let config = state.config.read().clone();
-        // Per-request dispatch: honor the caller's validated agent alias.
-        // Without one, fall back to the legacy pick — SSE / pairing
-        // endpoints still don't accept an explicit agent in the request
-        // payload, so the migration-synthesized "default" agent (or first
-        // enabled) remains their dispatch target.
-        let agent_alias = agent_override
-            .map(ToString::to_string)
-            .or_else(|| {
-                config
-                    .agents
-                    .keys()
-                    .find(|k| k.as_str() == "default")
-                    .or_else(|| {
-                        config
-                            .agents
-                            .iter()
-                            .find(|(_, a)| a.enabled)
-                            .map(|(alias, _)| alias)
-                    })
-                    .cloned()
-            })
-            .ok_or_else(|| {
-                ::zeroclaw_log::record!(
-                    WARN,
-                    ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Reject)
-                        .with_outcome(::zeroclaw_log::EventOutcome::Failure),
-                    "webhook chat rejected: no configured [agents.<alias>] entry"
-                );
-                anyhow::Error::msg(
-                    "webhook chat requires at least one configured [agents.<alias>] entry",
-                )
-            })?;
+        let agent_alias = require_gateway_chat_agent_alias(&config, agent_override)?;
 
         // Scope the cost tracking context so per-LLM-call usage flows into the
         // gateway's cost tracker and costs.jsonl. Without this scope, the
@@ -2250,6 +2219,31 @@ async fn run_gateway_chat_with_tools(
             cost_usd,
         })
     }
+}
+
+fn resolve_gateway_chat_agent_alias(
+    config: &Config,
+    agent_override: Option<&str>,
+) -> Option<String> {
+    agent_override
+        .map(ToString::to_string)
+        .or_else(|| config.resolved_runtime_agent_alias().map(str::to_owned))
+}
+
+#[cfg(not(test))]
+fn require_gateway_chat_agent_alias(
+    config: &Config,
+    agent_override: Option<&str>,
+) -> anyhow::Result<String> {
+    resolve_gateway_chat_agent_alias(config, agent_override).ok_or_else(|| {
+        ::zeroclaw_log::record!(
+            WARN,
+            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Reject)
+                .with_outcome(::zeroclaw_log::EventOutcome::Failure),
+            "webhook chat rejected: no configured [agents.<alias>] entry"
+        );
+        anyhow::Error::msg("webhook chat requires at least one configured [agents.<alias>] entry")
+    })
 }
 
 fn optional_channel_routes() -> Router<AppState> {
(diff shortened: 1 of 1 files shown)


WareWolf-MoonWall (approved):
# Review Draft — PR #7379
**fix(gateway): align webhook model labels with resolved agent config**
**Author:** Audacity88 · **Labels:** bug, gateway, risk: high, size: S
**Reviewer:** WareWolf-MoonWall · **Verdict:** `--approve`

---

I read the full diff against the live gateway source (`crates/zeroclaw-gateway/src/lib.rs`), cross-checked the config helpers `resolved_runtime_agent_alias()`, `resolved_model_provider_for_agent()`, and `resolve_default_model()` in `crates/zeroclaw-config/src/schema.rs`, and confirmed zero prior review activity on this PR.

The patch removes the hand-rolled "find 'default' key or first-enabled agent" inline fallback from both the chat dispatch path and the webhook observability path, replacing it with `Config::resolved_runtime_agent_alias()` and `resolved_model_provider_for_agent()`. The webhook response body and `AgentStart`/`AgentEnd` observer events now report the model the *agent is configured to use* rather than the gateway's startup model string. The PR template is fully completed including a filled-in Rollback section, which is required for `risk: high`.

---

### 🟢 What looks good — DRY fix is correct and complete

The inline fallback logic in `run_gateway_chat_with_tools` was a textbook duplicate of `Config::resolved_runtime_agent_alias()`. Extracting it into `resolve_gateway_chat_agent_alias` and delegating to the config layer is exactly what AGENTS.md's single-source-of-truth rule calls for. The `#[cfg(not(test))]` gate on `require_gateway_chat_agent_alias` is correctly scoped — the shared resolver is ungated, the error-returning wrapper is the test-excluded piece.

### 🟢 What looks good — test coverage via CapturingObserver

Adding `CapturingObserver` and renaming the test to `webhook_explicit_agent_reports_agent_model` is a meaningful upgrade over the old `webhook_explicit_agent_dispatches`. The test now pins both the `AgentStart` event fields (`model_provider` = `"anthropic.default"`, `model` = `"agent-model"`) and the HTTP response body (`"model"` = `"agent-model"`) independently, which would have caught exactly this bug had it existed before. The `model: "startup-model"` value on `AppState` makes the before/after contrast explicit in code.

### 🟢 What looks good — PR template complete for risk: high

Summary, Validation Evidence, Security & Privacy Impact, Compatibility, and Rollback are all filled in with literal content. Rollback includes the exact commit SHA to revert, no feature flags to toggle, and specific observable symptoms. This is the correct level of detail for a `risk: high` gateway PR.

---

### 🟡 Warning — `resolved_runtime_agent_alias` uses `min()` where old fallback used insertion order

**File:** `crates/zeroclaw-config/src/schema.rs` L3418–3430 vs old inline fallback in `lib.rs`.

The old inline code:
```rust
config.agents.keys().find(|k| k.as_str() == "default")
    .or_else(|| config.agents.iter().find(|(_, a)| a.enabled).map(|(alias, _)| alias))
```
returned the *first-inserted* enabled agent when no "default" key existed.

`resolved_runtime_agent_alias()` does:
```rust
self.agents.iter().filter(|(_, a)| a.enabled).map(|(alias, _)| alias.as_str()).min()
```
which returns the *lexicographically smallest* enabled alias.

For most deployments this is identical. For an operator with, say, aliases `"web"` and `"api"` and no `"default"` key, the old code picked whichever was inserted first (TOML parse order), the new code picks `"api"` (alphabetically first). The behavior difference is benign in the common case, but if a deployment has multiple non-"default" agents and the intent was to let TOML order determine fallback, this is a silent change. Worth a brief comment in `resolved_runtime_agent_alias` noting the ordering contract, or an issue to track if deterministic-but-different matters.

### 🟡 Warning — provider_label fallback narrowed; agents without explicit model_provider get "unknown"

**File:** `crates/zeroclaw-gateway/src/lib.rs`, `handle_webhook`.

Old code:
```rust
agent_override
    .and_then(|alias| cfg.resolved_model_provider_for_agent(alias))
    .map(|(ty, alias, _)| format!("{ty}.{alias}"))
    .or_else(|| {
        cfg.providers.models.iter_entries().next()
            .map(|(ty, alias, _)| format!("{ty}.{alias}"))
    })
    .unwrap_or_else(|| "unknown".to_string())
```

If the resolved agent had no explicit `model_provider` set, the old code fell back to the first entry in `providers.models` — a heuristic approximation of the gateway's actual provider. The new code has no such secondary fallback:
```rust
let provider_label = resolved_provider
    .as_ref()
    .map(|(ty, alias, _)| format!("{ty}.{alias}"))
    .unwrap_or_else(|| "unknown".to_string());
```

Deployments where agents don't have explicit `model_provider` configured will now see `"unknown"` in `AgentStart`/`AgentEnd` events and the response body where they previously saw the first configured provider. The "unknown" label is more honest about what was actually resolved, but it is an observable regression for operators who query observability data by `provider_label`. Whether this matters depends on how common unconfigured `model_provider` on agents is. A follow-up pass to propagate the gateway's `state.model_provider` type into the label fallback path would close the gap cleanly.

### 🔵 Suggestion — `"<unresolved>"` sentinel in JSON response body is nonstandard

**File:** `crates/zeroclaw-gateway/src/lib.rs`, `handle_webhook`.

```rust
.unwrap_or_else(|| "<unresolved>".to_string())
```

The previous provider fallback used the string `"unknown"`. Using `"<unresolved>"` with angle brackets in an API response field (`{"model": "<unresolved>"}`) may confuse consumers that parse or log this field — angle-bracket strings typically signal template/token formatting artifacts. Consider aligning with the existing `"unknown"` sentinel used for `provider_label` just two lines above, or documenting `"<unresolved>"` as the defined sentinel in the OpenAPI spec for this field.

---

**Verdict:** I have no blocking findings. The DRY fix is correct, the test coverage is specific and meaningful, the PR template satisfies the `risk: high` requirements, and CI is green. The two warnings are worth tracking as follow-up issues but do not require changes before merge. Approving as-is.

`--approve`

singlerider (approved):
Read the diff and traced the webhook label resolution. Correct fix that unifies dispatch and labeling on one resolver. Approving (no active block; @WareWolf-MoonWall approved this head).

### 🟢 What looks good — dispatch and labeling now resolve the agent the same way

The bug was a divergence: `/webhook` dispatched to the resolved agent but labeled the response and observer events with `state.model` (the gateway startup model), so a `?agent=nova` request reported nova's *runtime* but the startup model's *label*. Extracting the inline "default → first-enabled" fallback ladder into `resolve_gateway_chat_agent_alias` — which delegates to `Config::resolved_runtime_agent_alias()` — and then driving *both* the dispatch path (`require_gateway_chat_agent_alias`) and the label path off that same function is the right fix. There's now one answer to "which agent serves this request," and the model label is derived from that agent's `resolved_model_provider_for_agent(alias).model` instead of an unrelated startup field. The `#[cfg(not(test))]` split is only on the `require_` wrapper (the test harness stubs real dispatch); the shared `resolve_` core is un-gated and used by both paths, so test and prod resolve identically.

### 🟢 What looks good — the fallback chain degrades sensibly and the body matches the events

The model label resolves agent-provider model → `resolve_default_model()` → `"<unresolved>"`, and the provider label resolves the agent provider → `"unknown"`, so a misconfigured agent produces an honest placeholder rather than silently borrowing the startup model. Returning `model_label` in the `/webhook` success body instead of `state.model` is the part that actually closes #7297's symptom — the HTTP response and the `AgentEnd` event now carry the same string, so a caller and the observability trail agree.

### 🟢 What looks good — the test asserts the real contract, not the mechanism

`webhook_explicit_agent_reports_agent_model` sets `state.model = "startup-model"` but configures agent `nova` with provider `anthropic.default` whose `model = "agent-model"`, then asserts the response body's `model` is `"agent-model"` and the captured `AgentStart` event carries `anthropic.default` / `agent-model`. That directly proves the label follows the resolved agent and not the startup field — exactly the divergence that was broken. Renaming from `_dispatches` to `_reports_agent_model` reflects that the test now covers the metadata contract, not just that dispatch happened.

Approving. Both surfaces resolve through the canonical alias resolver, the fallback is honest, and the response/observer labels are proven consistent.

Issue #7100, opened 2026-06-02
RFC: Per-model capability & context-window config (vision, context_window) for capability checks, context budget, and UI display


NiuBlibing:
## Problem

Model capability, context capacity, runtime context budget, and displayed usage have historically come from different sources.

Provider-family defaults can misreport vision support for a selected model. An unset per-alias context window falls back to 32,000 tokens even when the model supports a different capacity. Compression and UI consumers can then present or enforce values that do not describe the same selected provider alias and model.

ZeroClaw needs one effective model-metadata contract without turning a remote catalog into a hidden runtime authority.

### Resolve metadata for the selected provider alias and model

Capability and context consumers resolve metadata for the provider profile actually selected for the turn, including provider overrides and model switches. They do not inherit metadata from an unrelated fallback or agent-default profile.

One shared resolver provides effective vision capability and model context capacity to provider routing, runtime budgeting, diagnostics, and UI surfaces.

### Vision capability

`providers.models.<type>.<alias>.vision` is an optional operator override for the configured model on that alias.

When set, the override controls `supports_vision()` and capability-based routing for the constructed provider. When unset, the existing provider-family capability remains the fallback. Resolution requires no new network request on the turn path.

This slice is implemented by #9099.

### Context capacity sources

`providers.models.<type>.<alias>.context_window` is the canonical configured model capacity for runtime use.

Provider or catalog metadata may supply a bounded value to profile-creation, quickstart, doctor, or other explicit config-writing flows. The value becomes runtime authority when persisted to the selected profile; a fetched catalog value does not create a parallel mutable runtime cache.

An unset capacity may retain the existing compatibility fallback, but diagnostics and UI must identify it as an unconfigured fallback rather than presenting 32,000 as known model truth. Zero or implausible external values are rejected.

A context limit learned from a provider error may narrow recovery for the current execution. It must not silently widen the configured capacity or rewrite persistent config without an explicit config update.

### Capacity, budget, and usage are different values

- **Model capacity** is the selected model's total configured context window.
- **Runtime budget** is how much of that capacity ZeroClaw may fill before proactive compression or trimming. It must never exceed model capacity and may be lower because of an operator cap, reserved response/tool headroom, or runtime policy.
- **Usage** is the current turn's consumed context. Prefer provider-reported prompt usage when available; otherwise use the shared estimator and identify the value as estimated where that distinction is user-visible.

A capacity value must not silently raise an explicit runtime or operator budget cap.

[...]

PR #7380, opened 2026-06-08
fix(zerocode): stabilize Quickstart agent creation flow

### LINKED ISSUE #7373: [Bug]: zerocode Quickstart can strand users after editor and submit flows
## Affected component

unknown

This is in the zerocode Quickstart TUI, but the current bug template does not have a zerocode/TUI component option.

## Severity

S1 - workflow blocked

## Current behavior

While testing zerocode Quickstart in a CMUX terminal on macOS, two Quickstart flows can leave the user stranded:

1. Opening the Agent file editor from Quickstart and then aborting/quitting the external editor can return to a visually corrupted Quickstart pane. The Agent modal stays centered in mostly empty space, and switching to another pane and back does not repair the Quickstart view.

Image

2. After submitting a completed Quickstart agent, the UI can show `Created \`a\`. Reloading daemon — Chat will open when reconnected…` for more than a minute without moving to Chat. Other panes still appear responsive, but Quickstart suppresses main-list input after `applied_alias` is set, so the user is left in a dead-end success state.
A related lower-severity UX problem showed up in the same flow: Channels and Peer groups are labelled optional, but Submit stays disabled until those rows are visited/checked. That makes optional setup look required.

## Expected behavior

Quickstart should recover cleanly from external editor return and should repaint the whole pane before input resumes.

After a successful Submit, Quickstart should always have a bounded path into the created agent's Chat view. If the daemon does not actually restart, the TUI should focus the created agent directly. If a restart was signalled, the reconnect wait should have a timeout/retry/recovery path instead of waiting forever.

Optional rows should either count as satisfied when their empty defaults are valid, or the UI/status copy should clearly say the user must review and confirm "none".

## Steps to reproduce

```bash
1. Launch zerocode in a terminal.
2. Open Quickstart.
3. In the Agent step, choose one of the agent files and open it with the external editor action.
4. Abort/quit the external editor without saving.
5. Observe whether Quickstart repaints correctly.
6. Complete the Quickstart selectors and submit a new agent.
7. Observe whether Quickstart moves into Chat or remains on the "Reloading daemon — Chat will open when reconnected…" success state.
```

## Impact

Affected users: first-run zerocode users using Quickstart in a terminal.

Frequency: observed during manual dogfooding while testing #7330. Exact master/release reproducibility still needs confirmation.

Consequence: onboarding can become visually corrupted or get stuck after a nominally successful create, forcing the user to exit/restart instead of continuing into the new agent.

## Logs / stack traces

No stack trace observed. The visible stuck status was:

```text
Created `a`. Reloading daemon — Chat will open when reconnected…
```

Relevant source-shape notes from #7330 investigation:

- The Quickstart Agent editor path calls `open_editor_for_content()` but does not issue a full redraw after returning; nearby Chat/config editor paths clear the terminal after returning from editor/alternate-screen flows.
- The Quickstart submit path queues `start_chat_with` and sets `applied_alias`, but discards the `daemon_restarted` flag from `QuickstartApplyResult::Applied`.
- The selector satisfaction logic still gates Channels and Peer groups on visit state.

## ZeroClaw version

Observed while testing PR #7330 at 6d154fe83f426819f9c0459ea869b5cf7ca51ab8, reported as v0.8.0-beta-2 in the zerocode UI.

## Rust version

Not checked; this is a runtime TUI behavior report.

## Operating system

macOS, CMUX terminal.

## Regression?

Unknown. These look pre-existing rather than introduced by #7330, but that should be verified against current master/release before final fix scoping.

## Pre-flight checks

- I reproduced this while testing a current PR head, not yet on latest master or latest release.

## Related

- #7330
- #6826
- #7117

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Reworks zerocode Quickstart recovery so creating an agent hands off to Chat immediately instead of leaving users on a stale "Reloading daemon" state.
  - Keeps optional Channels and Peer groups visually incomplete when skipped while still allowing Submit once all required selectors are complete.
  - Makes the model-provider form open immediately, fetch model catalogs without blocking the TUI, show loading/retry/fallback states, turn live model catalogs into a picker, sort chat/coding models ahead of unsuitable specialty models, and treat boolean fields as selectors.
  - Replaces the Quickstart Agent personality-file `$EDITOR` handoff with an in-pane editor that has explicit `Ctrl+S` / `Esc` border instructions and clamped scroll, avoiding the broken repaint state after quitting an external editor.
  - Exposes `MEMORY.md` through the personality-template RPC for staged Quickstart agents so the template hotkey can fill every listed personality file.
- **Scope boundary:** This PR does not change Chat's active-session model/provider reload behavior; that remains tracked separately in #7395.
- **Blast radius:** zerocode Quickstart UI, model-catalog RPC timeout behavior from zerocode, and the runtime personality-template RPC used by config/Quickstart editors. Existing no-agent template callers still receive the memoryless fallback.
- **Linked issue(s):** Closes #7373. Related #7395.
- **Labels:** `bug`, `risk: medium`, `size: XL`, `agent`, `onboard`, `runtime`, `quickstart`

```bash
cargo test -p zerocode quickstart_pane::tests -- --nocapture
```

Result: passed via the local cargo wrapper. The focused zerocode suite reported `29 passed; 0 failed`.

```bash
cargo test -p zeroclaw-runtime personality_templates -- --nocapture
```

Result: passed via the local cargo wrapper. The focused runtime template suite reported `9 passed; 0 failed`.

```bash
cargo fmt -p zeroclaw-runtime -p zerocode -- --check
```

Result: passed via the local cargo wrapper.

```bash
git diff --check
```

Result: both passed.

```bash
cargo build -p zerocode --bin zerocode
cargo build -p zeroclawlabs --bin zeroclaw
```

Result: both passed via the local cargo wrapper.

- **Beyond CI — what did you manually verify?** Live zerocode Quickstart smoke on macOS with a disposable config: optional rows remain visually incomplete when skipped; the model-provider picker shows loading/fallback behavior; Agent file editing uses the in-pane editor and repaints after exit; `MEMORY.md` template feedback is visible; the final ready status says "All required selectors".
- **If any command was intentionally skipped, why:** Full workspace nextest and workspace clippy were not run locally because this is a focused zerocode/runtime Quickstart recovery patch, and the targeted tests/builds cover the changed surfaces. GitHub CI should still run the full required suite.

## Compatibility (required)

- Backward compatible? `Yes`

- **Observable failure symptoms:** zerocode Quickstart modal input regressions, failed `quickstart/apply` handoff into Chat, missing model-provider catalog picker, or missing `MEMORY.md` template content in Agent setup.


apps/zerocode/src/quickstart_pane.rs
```diff
@@ -1899,6 +2216,179 @@ fn wrapped_total(lines: &[Line], width: u16) -> u16 {
...
+            "mistral", "grok", "coder", "code", "reason", "r1", "o3", "o4",
+        ],
+    ) {
+        rank -= 45;
+    }
+
+    rank
+}
+
+fn is_non_chat_model_lower(model: &str) -> bool {
```
singlerider: 🔴 This (plus `model_rank_lower` above) hardcodes model capability classification — which models are non-chat, which are chat/coding-suitable, how to rank them — as substring lists inside the TUI. Per AGENTS that is a list of "the things the system supports" typed into a surface; the live catalog RPC is correctly the source of truth for *which* models exist, but classifying/ranking them is a capability judgment the gateway/web/CLI will each want too, and they will each grow their own copy. I checked — there is no existing classifier in zeroclaw-providers/config/api this shadows, so this is the *first* model-classification logic and it is landing in the wrong crate (it is also what #7100 is slated to own). Small fix, no PR rework: move `is_non_chat_model_lower`, `model_rank_lower`, `contains_any` into `zeroclaw-providers` as a reusable classifier/ranker and have the TUI call it. Then it has one home and #7100 extends it instead of competing.

singlerider (changes requested):
Read the RPC dispatch change, the i18n additions across all five locales, the model-catalog client path, and the model sort/filter helpers in `quickstart_pane.rs`. Most of this XL recovery PR is solid and the Quickstart stabilization is clearly right. One blocking issue: the model classification/ranking is a hardcoded model-shaped list living in the TUI surface.

### 🔴 Blocking — `is_non_chat_model_lower` / `model_rank_lower` hardcode model classification in a surface

`is_non_chat_model_lower` (filters out `embedding`/`tts`/`whisper`/`image`/`realtime`/…) and `model_rank_lower` (ranks `gpt-5`/`gpt-4o`/`claude`/`sonnet`/`deepseek`/… ahead of the rest) encode "which models are chat/coding suitable" as substring lists inside `quickstart_pane.rs`. That's exactly the pattern AGENTS says to stop on: a list of "the things the system supports" (here, the capability class of a model) typed into a surface instead of read from the backend registry. The live catalog RPC is correctly the source of truth for *which models exist* — but the moment the TUI decides a model is "non-chat" or ranks it, it's making a capability judgment that the gateway, web, and CLI surfaces will each want too, and they'll each grow their own copy.

I confirmed there's no existing backend helper this is shadowing (no `is_embedding`/`model_kind`/capability classifier in `zeroclaw-providers`/`zeroclaw-config`/`zeroclaw-api` today), so this isn't duplicating an existing registry — it's the *first* model-classification logic, and it's landing in the wrong crate. That's also precisely what #7100 (per-model capability config) is slated to own. The fix here is small and doesn't require reworking the PR: move `is_non_chat_model_lower`, `model_rank_lower`, and `contains_any` into `zeroclaw-providers` (or wherever the model catalog is produced) as a reusable classifier/ranker, expose it, and have the TUI call it. Then the heuristic has one home, the other surfaces can reuse it, and #7100 extends it rather than competing with it. Keeping it in the TUI is the shadow-list the rule is meant to prevent.

### 🟢 What looks good — the MEMORY.md template fix is correctly scoped and backward compatible

`personality_template_context` solves a real first-run gap: Quickstart asks for personality templates before the agent exists in config, so the old `include_memory = config.agent(alias).is_some()` returned a memoryless skeleton during setup. Setting `include_memory = configured_agent_exists || agent_requested` gives an explicit agent request the full per-agent render (so `MEMORY.md` appears) while the no-agent/default caller keeps the memoryless fallback — so existing config-editor callers are unaffected. It trims the requested name, falls back to `"ZeroClaw"`, and the test pins the pre-existence case. It renders a template skeleton, not another agent's real memory, so there's no cross-agent leak.

### 🟢 What looks good — i18n is symmetric and the catalog fetch degrades

All five locales (en/es/fr/ja/zh-CN) add the identical five `zc-quickstart-*` keys — proper localization discipline, not an English-only drop that leans on a fallback. The model-catalog path shows explicit loading/retrying/empty states and a transient-miss retry before manual fallback (`transient_model_catalog_miss_retries_before_manual_fallback`), so a slow or failing catalog RPC degrades to a manual entry row instead of hanging the TUI — which is the actual #7373 symptom. Treating bool fields as two-variant selectors and the in-pane editor replacing the `$EDITOR` handoff both remove real broken-repaint states.

The Quickstart recovery, the template fix, the i18n, and the catalog resilience are all good. Relocating the model-classification heuristic out of the TUI into the backend is the one thing I'd block on — it's a small move that keeps the next three surfaces from each reinventing it.

Audacity88:
@singlerider thanks for the feedback. Pushed an update that moves the Quickstart model ranking/classification helper into `zeroclaw-providers::catalog` and keeps the Quickstart UI using that shared helper. I also added focused provider-side regression coverage for the ranking behavior.

WareWolf-MoonWall (commented):
Re-review: head `1de00410`, 6 commits. Prior review state: @singlerider has an active CHANGES_REQUESTED from commit `11ec022c`. Not approving over an active CR per protocol — commenting to confirm the blocker is resolved so @singlerider can re-review.

### ✅ singlerider's blocker is addressed — model classification moved to `zeroclaw-providers`

The last commit (`fix(zerocode): move model catalog ranking to providers`) moves `is_non_chat_model_lower`, `chat_model_rank_lower`, and the public `sort_model_catalog_for_chat` function into `crates/zeroclaw-providers/src/catalog.rs`. The `quickstart_pane.rs` now calls `zeroclaw_providers::catalog::sort_model_catalog_for_chat(provider, models)` instead of having the heuristic inline.

This is the right location per singlerider's feedback: the heuristic has one home in the catalog crate so other surfaces (web, CLI, gateway) can reuse it without copying it, and #7100 can extend it rather than competing with a TUI-local copy. The AGENTS.md SSOT rule is satisfied.

### 🟢 What looks good — the remaining Quickstart recovery work is solid

- Model catalog fetch with loading/retrying/empty states and `MODEL_CATALOG_MAX_ATTEMPTS = 2` retry before manual-entry fallback — this directly addresses the #7373 hang symptom.
- In-pane personality editor with `Ctrl+S`/`Esc` border instructions replaces the broken `$EDITOR` handoff repaint state.
- Optional rows (Channels, Peer groups) remain visually incomplete when skipped while still allowing Submit once required fields are complete.
- `MEMORY.md` in personality template for staged Quickstart agents — `include_memory = configured_agent_exists || agent_requested` is DRY-compliant (reads from config state, no copy).
- All 5 locales (en/es/fr/ja/zh-CN) receive the three new `zc-quickstart-model-*` keys symmetrically — correct i18n discipline.

### 🟢 What looks good — catalog RPC now has a timeout

`catalog_models` now uses `call_with_timeout(..., Duration::from_secs(20))` instead of the unbounded `call`. This is the correct fix for the hang: a slow provider catalog RPC can no longer block the Quickstart modal indefinitely.

@singlerider — the `is_non_chat_model_lower`/`model_rank_lower` concern has been addressed. Requesting your re-review to clear the active CHANGES_REQUESTED.

singlerider (approved):
Re-review at head 1de00410. Prior review 4470-series requested one change; verified resolved in code.

### ✅ Resolved — model classification/ranking moved out of the TUI into the providers crate

`is_non_chat_model_lower`, `chat_model_rank_lower`, and the sort/filter pipeline now live in `zeroclaw-providers::catalog` as `sort_model_catalog_for_chat(provider, models)`, and `quickstart_pane.rs` calls the shared helper instead of carrying its own substring lists. The capability heuristic now has one home that gateway/web/CLI surfaces can reuse and that #7100 (per-model capability config) can extend rather than compete with. Provider-side regression tests pin the filter and ranking behavior at the new location. This was the only blocker.

The MEMORY.md template fix (`include_memory = configured_agent_exists || agent_requested`), the symmetric five-locale i18n additions, the catalog fetch degradation (loading/retrying/empty states with transient-miss retry before manual fallback), and the in-pane editor replacing the `$EDITOR` handoff were all solid in the first pass and are unchanged.

CI is green on the new head. Approving.

PR #7381, opened 2026-06-08
fix(web): show and switch the agent's own model on the agent chat page

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The model button on the agent chat page (`/agent/<alias>`) read the global
    `model` / `default_model` config keys. Those were **removed in schema V3**, so
    `getProp('model')` / `getProp('default_model')` now return `path_not_found` and
    the button silently fell back to the daemon **status** model — showing the same
    **global default** (`ollama.default`) on *every* agent page, and writing the
    (gone) global key on switch (a no-op).
  - Fix: read and write the **per-agent** `agents.<alias>.model_provider`
    (`getProp`/`putProp` already accept any dotted config path — **no new
    endpoint**), populate the dropdown from the configured `providers.models.*`
    refs, and **gate on the prop `value`** rather than a `populated` field. (The
    single-prop `GET /api/config/prop` returns only `{ path, value }`; `populated`
    exists on `/api/config/list`, so the old check silently discarded the per-agent
    value even when present.)
  - The web session agent rebuilds from live config on the switch reconnect, so a
    switch **hot-applies without a daemon reload** for the web chat.
- **Scope boundary:** **frontend only** (`web/src/contexts/AgentContext.tsx`). Does
  **not** change backend model resolution, the gateway telemetry, or the
  channel/Discord reload behaviour.
- **Blast radius:** one React context. Pure config-prop reads/writes through the
  existing `/api/config/prop` surface; no backend, schema, or API changes.
- **Known-adjacent (explicitly out of scope, follow-ups):**
  - `gateway_ws_turn` trace events still label the model as the global default
    (`ws.rs:744` seeds from `config.resolve_default_model()`) — a **telemetry**
    mislabel only; inference is correct. Separate PR.
  - **Discord/channel** agents are long-running and still need a daemon reload to
    pick up a model switch (no per-agent hot-swap) — unchanged here.
- **Linked issue(s):** none — found via dogfooding the multi-agent web UI.
- **Labels:** expect `web`, `type: fix`, `risk: low`, `size: S`.

```text
$ cd web && npm run build      # tsc -b && vite build
✓ tsc -b clean (no type errors)
✓ vite build ✓ built  (AgentChat / AgentContext chunks rebuilt)
```

- **Beyond CI — what did you manually verify?** Built and deployed to a running
  multi-agent instance, then exercised the button on `/agent/<alias>`:
  - **Before:** every agent page showed the same global default
    (`ollama.default` / its model) regardless of the agent's configured model; the
    switch wrote a removed global key and did nothing.
  - **After:** the button shows the agent's own `model_provider`, the dropdown lists
    the configured provider refs, and selecting one writes
    `agents.<alias>.model_provider`.
  - **Hot-swap confirmed end-to-end** via a temporary runtime probe at the provider
    call site: across **8 consecutive web turns** I switched through **6 distinct
    models spanning 2 provider families** (`kilo`: deepseek-v4-flash, minimax-m2.7,
    mimo-v2.5-pro; `opencode`: qwen3.7-plus, minimax-m2.5). **Every turn's actual
    dispatched model matched the just-selected dropdown value** — not the global
    default — with no daemon reload. A `/admin/reload` mid-test reset the session
    history cleanly and tracking continued correctly across it.
- **Not verified here:** Discord/channel switching (needs a reload, out of scope)
  and the `gateway_ws_turn` telemetry label (separate follow-up).

## Compatibility (required)

- Backward compatible? `Yes` — only changes which config path the chat-page model
  button reads/writes; uses the existing `/api/config/prop` surface.
- Upgrade steps: none.

Low risk: `git revert <sha>` — a single frontend context file.


singlerider (approved):
Read the one-file diff against `AgentContext.tsx`, the gateway prop endpoints, and `UNSET_DISPLAY` in `zeroclaw-config`.

### 🟢 What looks good — root cause traced to the actual schema break, not patched around

The button didn't "show the wrong model" randomly: V3 removed the global `model` / `default_model` keys, so `getProp` 404'd and the code silently fell back to the daemon status model on every agent page, and switches wrote a key that no longer exists (a no-op that looked like success). Reading and writing `agents.<alias>.model_provider` is the correct V3 source of truth, and reusing `getProp`/`putProp` dotted paths means no new endpoint surface.

### 🟢 What looks good — the `populated` gate fix is verified against the backend

The NOTE about `GET /api/config/prop` returning only `{ path, value }` (no `populated` field — that exists only on the list endpoint) matches the gateway implementation, and the `"<unset>"` sentinel check matches `UNSET_DISPLAY` in `zeroclaw-config/src/traits.rs`. The old code gated on a field that was always `undefined` — this class of silent-failure is exactly what the comment will prevent from regressing.

### 🟢 What looks good — dropdown discovers refs from the registry

Switch targets come from walking `providers.models.*` via `listProps` and deriving `<family>.<alias>` refs, not from a hardcoded list — the registry stays the source of truth. Adding `agentAlias` to the effect deps closes the stale-state hole when navigating between agent pages, and the `cancelled` guards are applied consistently across both awaits.

Approving.

singlerider:
My approval from 2026-06-11 is on the current head `cdb9ac95` — no new commits since, so it stands. No re-review needed on the merits.

One merge-sequencing note: the branch is now `CONFLICTING` / `DIRTY` against current `master`. The conflict is in `web/src/contexts/AgentContext.tsx`, which overlaps with the model-switcher work in #7191 (also touching this file's model resolution). Whichever lands first, this branch needs a refresh against `master` before it can merge — please rebase and re-run CI. The approval carries over once the conflict is resolved cleanly.

Audacity88 (approved):
Context: I re-reviewed #7381 at current head `e9de3a6dd27582948c4532f59169575c0e3e68f0` after the conflict-refresh merges, checking the live PR state, the refreshed `web/src/contexts/AgentContext.tsx` diff against current `master`, the gateway/config endpoint shapes behind the change, the overlap with the now-merged #7191 model-switcher work, and current CI; I did not run an additional local web build or live model-switch session.

### 🟢 What looks good — current head keeps the V3 source of truth intact

The refreshed code still reads the active per-agent provider ref from `agents.<alias>.model_provider` and writes the same path on switch. That matches the current routing/config contract: an agent's `model_provider` is the dotted `<type>.<alias>` ref into `providers.models`, not the removed global `model` / `default_model` keys and not just a provider-local model ID.

### 🟢 What looks good — endpoint-shape assumptions match the backend

The `getProp` branch gates on the returned `value` and the `"<unset>"` sentinel rather than a `populated` field, which matches `GET /api/config/prop` for non-secret fields. The dropdown discovery through `listProps('providers.models')` also matches `/api/config/list`, where each configured provider profile exposes the `providers.models.<family>.<alias>.model` path that the UI maps back to a switchable provider ref.

### 🟢 What looks good — conflict refresh preserved adjacent switcher fixes

The merge commits pulled in #7191's adjacent `AgentContext` work without dropping it: agent-scoped status fallback, catalog-backed provider display names, whitespace-only response trimming, and the per-phase model-switch watchdog are all still present on this head. The current PR's one-file diff is now narrowly about showing/switching the agent's own provider ref, and CI is green across the required checks.

Approving the refreshed head.

PR #7382, opened 2026-06-08
fix(gateway): attribute WS turn telemetry and cost to the agent's model

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - In `process_chat_message` (the chat WebSocket turn handler), the `agent_start`
    broadcast, **every `gateway_ws_turn` trace event**, and the **per-turn cost
    record** were labelled with:
    - `provider_label` = the **first** `providers.models` entry (`iter_entries().next()`), and
    - `state.model` = the **global default** model (`resolve_default_model()`),

    regardless of which agent/model actually served the turn. So a web-chat turn
    served by e.g. `kilo` / `deepseek-v4-flash` was traced **and costed** as
    `ollama.default` / granite — on every agent.
  - Fix: use the agent's own per-turn attribution — `agent.attribution_fields()`,
    which is **already** used (correctly) in the turn's trace span — for the
    broadcast, the `gateway_ws_turn` events, and `record_turn_cost`.
- **Scope boundary:** attribution/labels only. **Inference was already correct**
  (the agent runs its configured model); this only fixes what's *reported*. The
  auto-save / memory-consolidation path keeps using the session provider object
  unchanged, and the onboarding default-model check is untouched.
- **Blast radius:** `gateway_ws_turn` trace events, the `agent_start` WS broadcast,
  and per-turn cost attribution in the chat-WS path. No behaviour/inference change.
- **Linked issue(s):** Related #7381 (WebUI per-agent model button) — same
  global-vs-per-agent attribution class, surfaced from the same investigation.
- **Labels:** `bug`, `size: XS`, `risk: medium`, `gateway`, `observability`.

```text
$ cargo fmt --all -- --check
# clean (exit 0)

$ cargo test -p zeroclaw-gateway
test result: ok. 226 passed; 0 failed; 0 ignored; 0 measured

$ cargo check -p zeroclaw-gateway
    Finished `dev` profile  # clean (exit 0)
```

- **Beyond CI — what did you manually verify?** This bug was caught by an
  instrumented run: a runtime probe at the provider call site logged the *actual*
  dispatched model for 8 consecutive web-chat turns across 6 models / 2 provider
  families (`kilo`, `opencode`) — every one matching the agent's configured model
  — while the `gateway_ws_turn` events for those same turns logged `ollama.default`.
  That mismatch is exactly what this patch removes: the events now read from the
  same `attribution_fields()` the trace span already uses.
- **If any command was intentionally skipped, why:** `cargo clippy --
  -D warnings` currently fails to compile the transitive `zeroclaw-channels` dep
  on an **unrelated pre-existing** unused import
  (`crate::paced_channel::PacedChannel`, `orchestrator/mod.rs:115`) — not in this
  diff. The `zeroclaw-gateway` crate itself is clippy-clean.

## Compatibility (required)

- Backward compatible? `Yes` — only changes the model/provider labels on telemetry
  + cost records to the correct per-agent values. No schema/config/API change.
- Upgrade steps: none.

- **Fast rollback command/path:** `git revert <sha>` — a single handler in
  `crates/zeroclaw-gateway/src/ws.rs`.
- **Observable failure symptoms:** web-chat `agent_start` / `agent_end`,
  `gateway_ws_turn`, or `/api/cost` entries show the gateway boot/default model
  instead of the selected agent model.

### DIFF
--- a/crates/zeroclaw-gateway/src/ws.rs
+++ b/crates/zeroclaw-gateway/src/ws.rs
@@ -823,21 +823,20 @@ async fn process_chat_message(
     use futures_util::StreamExt as _;
     use zeroclaw_runtime::agent::TurnEvent;
 
-    let provider_label = {
-        let cfg = state.config.read();
-        cfg.providers
-            .models
-            .iter_entries()
-            .next()
-            .map(|(ty, alias, _)| format!("{ty}.{alias}"))
-            .unwrap_or_else(|| "unknown".to_string())
-    };
+    // Attribute telemetry, broadcasts, and cost to THIS agent's actual model
+    // (resolved per-turn), not the global default model or the first configured
+    // provider. Previously `provider_label` took the first `providers.models`
+    // entry and the model came from `model_label` (the global default), so every
+    // gateway_ws_turn / agent_start / cost record mislabelled the model.
+    let (turn_alias, turn_provider, turn_model) = agent.attribution_fields();
+    let provider_label = turn_provider.clone();
+    let model_label = turn_model.clone();
 
     // Broadcast agent_start event
     let _ = state.event_tx.send(serde_json::json!({
         "type": "agent_start",
         "model_provider": provider_label,
-        "model": state.model,
+        "model": model_label,
     }));
 
     // Set session state to running
@@ -870,7 +869,6 @@ async fn process_chat_message(
     // from the other branch.
     let content_owned = content.to_string();
     let session_key_owned = session_key.to_string();
-    let (turn_alias, turn_provider, turn_model) = agent.attribution_fields();
     let turn_fut = async {
         use ::zeroclaw_log::Instrument as _;
         let span = ::zeroclaw_log::info_span!(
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
Reviewed the gateway WebSocket attribution change against live PR metadata, the public diff, current live checks, the related #7381 public PR context, no existing comments/reviews, and public source excerpts for `Agent::attribution_fields()` plus the gateway cost-pricing key shape. I did not run additional local runtime validation beyond checking the diff against those source paths and the author's validation notes.

### 🟢 What looks good — WS turn attribution is now agent-scoped

The patch keeps inference behavior unchanged and moves the reporting surfaces to the same per-agent attribution tuple already used by the turn span: `agent_start` / `agent_end`, `gateway_ws_turn` success/cancel/failure records, the done frame, and `record_turn_cost`. That matches the existing pricing lookup shape too, because `record_turn_cost` already keys provider pricing by the dotted `<type>.<alias>` ref.

I did not find a source-level blocker. The current live checks are passing, and this is appropriately limited to the gateway chat-WS attribution path.

PR #7384, opened 2026-06-08
feat(cron): add a pause/resume toggle to scheduled tasks

### LINKED ISSUE #7356: [Feature]: Add a pause/disable button to Scheduled Tasks
### Summary

Add a pause/disable button to Scheduled Tasks

### Problem statement

Add a pause/disable button to Scheduled Tasks which would stop the further execution of that particular task. It would be like when we comment (#) a crobtab entry instead of deleting it altogether.

### Proposed solution

Add a disable button against each entry of scheduled tasks under "Actions Column"

Image

### Non-goals / out of scope

_No response_

### Alternatives considered

_No response_

### Acceptance criteria

_No response_

### Architecture impact

_No response_

### Risk and rollback

_No response_

### Breaking change?

No

### Data hygiene checks

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** Scheduled Tasks had no way to **pause** a job from the dashboard — only delete it. The data model already supported it (`CronJobDecl.enabled`, the scheduler honours it, the store applies it, the `cron_update` agent tool toggles it); the gaps were the HTTP `PATCH /api/cron/:id` body (didn't expose `enabled`) and the UI (showed the enabled/disabled badge but had no control).
  - **gateway:** add `enabled: Option<bool>` to `CronPatchBody` and thread it into `CronJobPatch`. Make `agent` optional on the patch body, and **gate the agent-existence check on `command.is_some()`** — a schedule/name/enable patch doesn't touch a shell command, so it needn't supply (or validate) an agent. (This also fixes a latent failure for non-command edits.)
  - **web:** add `enabled` to the `patchCronJob` type, and a **Pause/Resume** button in the Scheduled Tasks actions column that flips `enabled`. Resume uses a distinct `Power` icon so it isn't confused with the `Play` "Run now" button.
- **Scope boundary:** exposing + toggling the **existing** `enabled` flag. No schema change, no scheduler logic change, no new endpoint.
- **Blast radius:** one HTTP body field + the cron page actions column. The agent change only loosens a previously-required field and skips an unnecessary validation for non-command patches.
- **Linked issue(s):** Closes #7356.
- **Labels:** expect `cron`, `web`, `type: feat`/`enhancement`, `risk: low`, `size: S`.

```text
$ cargo fmt --all -- --check                 # clean (exit 0)
$ cargo check -p zeroclaw-gateway            # clean (exit 0)
$ cargo test -p zeroclaw-gateway
test result: ok. 226 passed; 0 failed; 0 ignored; 0 measured
$ cd web && npm run build                    # tsc -b + vite — clean (exit 0)
```

- **Deployed and verified live** (multi-agent instance): the Pause button flips a job to *Disabled* (scheduler stops firing it); Resume reverses it. Direct PATCH round-trip confirmed against a real job: `{"enabled":false}` with no `agent` → HTTP 200, `job.enabled=false`; `{"enabled":true}` → 200, restored. Both the earlier required-`agent` 422 and the `Unknown agent ""` 400 are gone (the existence check is now gated on `command.is_some()`).
- **Also covered by existing tests:** the scheduler's enable/disable behaviour (`cron/scheduler.rs` — `assert!(!updated.enabled)` paths).
- **Skipped:** `cargo clippy -- -D warnings` fails on a pre-existing, unrelated `zeroclaw-channels` unused import (`crate::paced_channel::PacedChannel`, `orchestrator/mod.rs:115`) — not in this diff.

## Compatibility (required)

- Backward compatible? `Yes` — `enabled` is an additive optional field; making `agent` optional only loosens a previously-required field. New i18n keys (`cron.pause`, `cron.resume`) added to `en`; other locales fall back to English.
- Upgrade steps: none.

## Rollback

Low risk: `git revert <sha>` — one handler field + gate, one UI button.

### DIFF
--- a/crates/zeroclaw-gateway/src/api.rs
+++ b/crates/zeroclaw-gateway/src/api.rs
@@ -109,7 +109,10 @@ pub struct CronAddBody {
 #[derive(Deserialize)]
 pub struct CronPatchBody {
     /// Configured agent alias whose risk profile gates the new shell
-    /// command (when `command` is being patched). Required.
+    /// command. Only consulted when `command` is being patched; optional
+    /// otherwise (e.g. a pure schedule/name change or an enable/disable
+    /// toggle), so non-command patches need not supply it.
+    #[serde(default)]
     pub agent: String,
     pub name: Option<String>,
     pub schedule: Option<String>,
@@ -117,6 +120,9 @@ pub struct CronPatchBody {
     pub clear_tz: Option<bool>,
     pub command: Option<String>,
     pub prompt: Option<String>,
+    /// Toggle the job on/off without deleting it (pause/resume). `None` leaves
+    /// the current state unchanged.
+    pub enabled: Option<bool>,
 }
 
 enum CronTimezonePatch {
@@ -606,7 +612,10 @@ pub async fn handle_api_cron_patch(
     }
 
     let config = state.config.read().clone();
-    if config.agent(&body.agent).is_none() {
+    // The agent only gates a shell `command` change (risk profile). Skip the
+    // existence check for patches that don't touch the command — schedule, name,
+    // and enable/disable toggles don't need an agent supplied.
+    if body.command.is_some() && config.agent(&body.agent).is_none() {
         return (
             StatusCode::BAD_REQUEST,
             Json(serde_json::json!({"error": format!(
@@ -625,6 +634,7 @@ pub async fn handle_api_cron_patch(
         clear_tz,
         command,
         prompt,
+        enabled,
     } = body;
     let timezone_patch = match parse_timezone_patch(tz, clear_tz) {
         Ok(patch) => patch,
@@ -687,6 +697,7 @@ pub async fn handle_api_cron_patch(
         schedule,
         command: patch_command,
         prompt: patch_prompt,
+        enabled,
         ..zeroclaw_runtime::cron::CronJobPatch::default()
     };
 
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -1465,6 +1465,7 @@ export function patchCronJob(
     clear_tz?: boolean;
     command?: string;
     prompt?: string;
+    enabled?: boolean;
   },
 ): Promise<CronJob> {
   return apiFetch<CronJob | { status: string; job: CronJob }>(
--- a/web/src/lib/i18n.ts
+++ b/web/src/lib/i18n.ts
@@ -559,6 +559,8 @@ const translations: Record<Locale, Record<string, string>> = {
     'cron.allowed_tools_placeholder': 'e.g. shell, file_read, memory_store',
     'cron.trigger': 'Run now',
     'cron.trigger_error': 'Failed to trigger job',
+    'cron.pause': 'Pause',
+    'cron.resume': 'Resume',
     'cron.dismiss': 'Dismiss',
 
     // Integrations
--- a/web/src/pages/Cron.tsx
+++ b/web/src/pages/Cron.tsx
@@ -20,9 +20,11 @@ import {
   ChevronDown,
   ChevronRight,
   Clock,
+  Pause,
   Pencil,
   Play,
   Plus,
+  Power,
   RefreshCw,
   Trash2,
   X,
@@ -455,6 +457,17 @@ export default function Cron() {
       setError(err instanceof Error ? err.message : t('cron.delete_error'));
     } finally {
       setConfirmDelete(null);
+    }
+  };
+
+  // Pause/resume a job without deleting it — toggles the existing `enabled`
+  // flag the scheduler already honours.
+  const handleToggleEnabled = async (job: CronJob) => {
+    try {
+      const updated = await patchCronJob(job.id, { enabled: !job.enabled });
+      setJobs((prev) => prev.map((j) => (j.id === job.id ? updated : j)));
+    } catch (err: unknown) {
+      setError(err instanceof Error ? err.message : t('cron.edit_error'));
     }
   };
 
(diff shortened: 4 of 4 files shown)


WareWolf-MoonWall (approved):
Reviewed: head `d5954346`, 4 files: `crates/zeroclaw-gateway/src/api.rs` (+13/-2), `web/src/lib/api.ts` (+1), `web/src/lib/i18n.ts` (+2), `web/src/pages/Cron.tsx` (+24). Risk: high (gateway). Closes #7356.

### 🟢 What looks good — exposes existing capability, does not invent new state

`CronJobDecl.enabled` already existed. The scheduler already honours it. `cron_update` already toggled it. The only gap was the HTTP PATCH body and the web button. This PR closes that gap without touching the scheduler, the store, or the schema. No DRY concern: `enabled` in `CronPatchBody` is a write input, not a cached copy of state.

### 🟢 What looks good — agent-existence gate correctly preserved for command patches

The gate change is the security-critical part of this PR. Previously: `if config.agent(&body.agent).is_none()` — always ran, requiring `agent` even for non-command patches. Now: `if body.command.is_some() && config.agent(&body.agent).is_none()` — only runs when a command change is requested.

The `#[serde(default)]` on `agent: String` means an omitted `agent` field deserialises to `""`. If a caller sends `{ "command": "something", "agent": "" }` (or omits `agent` while providing a command), `config.agent("").is_none()` is `true` and the 400 is still returned — the validation is not weakened for command patches. This was verified in your live round-trip test. ✓

### 🟢 What looks good — `handleToggleEnabled` pattern is correct

`patchCronJob(job.id, { enabled: !job.enabled })` reads the current UI state then sends the negated value. On success the response updates `setJobs` with the server's returned job, so the UI state is authoritative from the server after every toggle. No client-side toggle that forgets to sync. ✓

### 🟢 What looks good — Pause vs Resume icon distinction is the right UX choice

Using `Pause` (two vertical bars) for an active job and `Power` (circle-with-stem) for a paused job avoids confusion with the `Play` "Run now" button. The `title` attribute carries the localised label for screen-reader and tooltip coverage.

### 🟢 What looks good — validation is thorough for a risk: high PR

`cargo fmt`, `cargo check -p zeroclaw-gateway`, `cargo test -p zeroclaw-gateway` (226/226 pass), `npm run build` (clean), and live round-trip verification. The pre-existing clippy failure in `zeroclaw-channels` is honestly noted and is outside the diff. Rollback is clear (`git revert <sha>`).

### 🟡 Non-blocking — `agent: String` with `serde(default)` vs `Option<String>`

Making `agent` optional via `#[serde(default)]` on a `String` is functional but slightly implicit: the empty-string sentinel `""` has no special meaning in Rust's type system, so a future reader might miss the gating logic. An `Option<String>` would make the optionality explicit in the type and allow `body.agent.as_deref().unwrap_or("")` at the check site. Worth considering in a follow-up refactor, but not a blocker.

### 🟡 Non-blocking — web i18n for `cron.pause` / `cron.resume` is English-only

The PR body correctly notes "other locales fall back to English." This is acceptable for new web UI strings in a first-pass feature, but `es`, `fr`, `ja`, `zh-CN` translations for these two keys would complete the work. Easy follow-up.

Approving.

singlerider (approved):
Read the four-file diff, the `CronJobPatch` apply path in `cron/store.rs`, and the web i18n fallback. WareWolf-MoonWall already approved; this adds a second maintainer pass.

### 🟢 What looks good — the fix is a gap-fill, not a new mechanism

`CronJobDecl.enabled` already existed end-to-end: the scheduler honours it, `store.rs` applies `patch.enabled`, and the `cron_update` agent tool toggles it. This PR only exposes the existing flag through the HTTP patch body and the dashboard — no schema change, no scheduler logic, no new endpoint. That's the right-sized change for the feature.

### 🟢 What looks good — the agent-existence gate moves to where it's actually needed

The check exists to gate shell-command changes behind a risk profile, so gating it on `body.command.is_some()` is the correct boundary, and it fixes the latent failure the body documents (pure schedule/name/enable patches previously 422'd on the missing `agent` or 400'd on `Unknown agent ""`). When a command patch *does* arrive without an agent, the `#[serde(default)]` empty string still fails the existence check — so the security gate is preserved exactly where it applies. The live PATCH round-trip evidence covers both directions of the toggle plus both former failure modes.

### 🟢 What looks good — UI affordance details

Distinct `Power` icon for Resume so it can't be confused with the `Play` run-now button, title text localized through `t()`, and the optimistic row update replaces the row with the server's returned job rather than assuming. The new `cron.pause`/`cron.resume` keys are en-only, but the web i18n `t()` falls back to en for every other locale, so nothing leaks raw keys; translations can follow the file's usual batch pattern.

The clippy skip in the evidence (pre-existing `PacedChannel` unused-import on an untouched file) was legitimate at the time and has since been fixed on master.

Approving.

Audacity88 (approved):
I reviewed PR #7384 at head `d59543467a56847b6662d2c627281cb6079e1d6a`, including issue #7356, the four-file diff, the prior WareWolf-MoonWall and singlerider approvals, the current PR body, and the visible check state. The PR is already approved by two reviewers, and all visible GitHub checks are green.

### 🟢 What looks good — The pause/resume path uses the existing cron state

The diff adds `enabled` to the `PATCH /api/cron/:id` body, threads it into `CronJobPatch`, and adds the dashboard action that flips the field. That keeps the new HTTP and UI surface narrow instead of introducing a parallel pause flag, and it matches #7356's requested behavior: disable a scheduled task without deleting it.

### 🟢 What looks good — The command-risk boundary stays focused

The gateway change removes the unnecessary up-front agent-existence check for pure name, schedule, timezone, and enable/disable patches. The diff still routes command changes through the existing cron update path rather than adding a separate command mutation path. That is the right distinction for this feature: toggling an existing job should not require resubmitting command-risk context, but changing the shell command should stay on the command-update path.

### 🟢 What looks good — The UI update is server-authoritative

The button sends `{ enabled: !job.enabled }`, then replaces the row with the job returned by the server. That avoids a purely client-side toggle and keeps the badge, next render, and any future server-side normalization tied to the persisted value. The separate `Pause` and `Power` icons also keep Resume visually distinct from the existing Run Now action.

### 🔵 Suggestion — Add a narrow gateway regression test later

The PR body's manual PATCH evidence covers the important live behavior, and the existing cron tests cover disabled jobs not being scheduled. A future follow-up would still be useful for the gateway handler itself: one test that patches `enabled` without `agent` and one test that command patches still reject a missing or unknown agent. That would lock in the exact boundary this PR is relying on.

Approving. This is a clean implementation of #7356, and the remaining test-depth note is follow-up quality rather than a merge blocker.

Nillth:
Superseded by #7666 — same pause/resume capability (`enabled` on `CronPatchBody`), refined with the agent-requirement scoping (only required for shell-command patches) and the 400-not-500 fix on the prompt-as-command path.

PR #7385, opened 2026-06-08
feat(observability): add turn metadata to observer events and correlate OTel spans by turn_id

FTDGRT:
## Summary

Supersedes #7233. RFC: #7232

This is the **metadata/correlation-only** first slice — it intentionally does not carry over the `ContentProcessor`, LLM I/O capture, `record!` migration, or the claim that `observer_bridge` is the single authoritative emission path. See "What this PR explicitly does NOT change" below.

- **What changed and why:** ZeroClaw's observer events lacked channel attribution, agent identity, and turn correlation — every event was an orphan. OTel spans were flat with no parent-child linking, making trace navigation impossible. This PR adds `channel`, `agent_alias`, and `turn_id` as optional metadata fields to six `ObserverEvent` variants, generates one `turn_id` per covered agent turn, replaces the flat `tokens_used: Option<u64>` in `AgentEnd` with `TurnTokenUsage { input_tokens, output_tokens }` (no other events were affected by this change), enriches `observer_bridge` projection to forward the same metadata, and rewrites `OtelObserver` to open/close a parent span per turn and link LLM/tool child spans under it via `turn_id`.
- **Scope boundary:** No `record!` migration for runtime observer events. No LLM prompt/response content capture. No `user_message` / `response_content` fields. No `ContentProcessor` / `LeakDetector` / content policy. No claim that all runtime paths are correlated — `loop_.rs` and `tool_execution.rs` continue to emit `ObserverEvent` directly with `turn_id: None`. No changes to WS `is_observability_telemetry()` / `is_global_chat_event()` filtering.
- **Blast radius:** Only `agent.rs` assigns actual values (`agent_alias`, `turn_id`) to the new metadata fields. All other files that construct `ObserverEvent` variants pass `None` for these fields solely to adapt to the expanded schema and prevent compilation errors — no actual metadata values flow through those paths. `TurnTokenUsage` only replaces `AgentEnd`'s `tokens_used: Option<u64>`; no other event variant was affected. Downstream observers that use `..` in match arms require no changes. Observers with explicit match arms will see compile-time errors for the new fields — a safe, non-runtime-breaking expansion. OTel backends now receive correlated traces per turn instead of orphaned spans. SSE broadcast consumers see additional optional JSON fields.

### 8020e5a `feat(observability): add turn metadata fields to observer events and adapt compilation`

- Introduces `TurnTokenUsage { input_tokens, output_tokens }` in `observability_traits.rs`, replacing the flat `tokens_used: Option<u64>` in `AgentEnd` only
- Adds `channel: Option<String>`, `agent_alias: Option<String>`, `turn_id: Option<String>` to six `ObserverEvent` variants: `AgentStart`, `LlmRequest`, `LlmResponse`, `AgentEnd`, `ToolCallStart`, `ToolCall`
- Adds `input_tokens: Option<u64>`, `output_tokens: Option<u64>` to `LlmResponse`
- Updates all workspace construction sites to pass `None` for new metadata fields and adapt to `TurnTokenUsage` where `AgentEnd` was previously constructed
- Updates `PrometheusObserver` to compute token gauge from `input_tokens + output_tokens`
- Updates `LogObserver`, `VerboseObserver`, `NoopObserver` match arms
- Updates `sse.rs` broadcast observer to handle `TurnTokenUsage` (total = input + output) and forward metadata fields as optional JSON keys
- Adds schema test `observer_events_carry_turn_metadata_and_structured_usage`

### b5a3dff `feat(observability): assign agent_alias and turn_id in agent.rs, add agentstart/agentend with token usage`

- Adds `new_turn_id()` helper (UUID v4) and `observer_agent_alias()` helper to `Agent`
- Generates `turn_id` at the top of `Agent::turn` and `turn_streamed_with_steering_state`
- Emits `AgentStart` at turn entry with `turn_id`, `agent_alias`, and `channel: None`
- Emits `AgentEnd` on all exit paths with accumulated token totals via `TurnTokenUsage`
- Passes `turn_id` through to tool execution so `ToolCall`/`ToolCallStart` events carry it
- Accumulates per-turn token consumption into `AgentEnd`
- Adds unit test asserting all events within a turn share the same `turn_id`

### 228980f `feat(observability): enrich observer_bridge projection with turn metadata`

- Expands `observer_bridge::project()` to extract `channel`, `agent_alias`, `turn_id` from `LogEvent` span attribution and attributes
- Constructs `TurnTokenUsage` from `input_tokens`/`output_tokens` attributes for `AgentEnd`
- Forwards the three metadata fields across all six projected variants
- Adds `ToolCallStart` and `LlmRequest` projection arms (previously unhandled)
- Converts empty-string extractions to `None` for clean downstream consumption
- Adds tests: `projects_llm_request_with_turn_metadata`, `projects_agent_end_structured_usage`

### add09f8 `feat(observability): standardize otel span attributes and correlate spans by turn_id`

- Adds `active_agent_spans: Mutex<HashMap<String, (BoxedSpan, Context)>>` to `OtelObserver`, keyed by `turn_id`
- `AgentStart` opens a `gen_ai.agent.invoke` parent span, stored in the map
- `AgentEnd` removes the parent span, sets `duration_s`/`gen_ai.usage.input_tokens`/`gen_ai.usage.output_tokens`, and ends it
- `LlmRequest`, `LlmResponse`, `ToolCallStart`, `ToolCall` create child spans via `build_with_context(..., &parent_cx_for(turn_id))`
- All span attributes aligned with OpenTelemetry Gen-AI semantic conventions: `gen_ai.provider.name`, `gen_ai.request.model`, `gen_ai.usage.input_tokens`, `gen_ai.usage.output_tokens`, `gen_ai.operation.name`, `gen_ai.agent.name`, `zeroclaw.turn_id`, `zeroclaw.channel`
- `flush()` drains orphaned spans from unclosed turns
- When `turn_id` is absent, spans degrade to orphans (backward compatible)
- Adds lifecycle test `turn_id_opens_and_closes_agent_span`

## Validation Evidence

```bash
$ cargo fmt --all -- --check
# (no output — clean)

$ cargo clippy --locked --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 24s
# (no warnings)

$ cargo test --locked --all-targets
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

- **Beyond CI — what did you manually verify?** Verified end-to-end by starting `zeroclaw daemon`, chatting via http://127.0.0.1:42617/agents, and confirming enriched trace data (turn correlation, token breakdown, channel attribution) appears on the configured Langfuse backend.

## Security & Privacy Impact

- New external network calls? **No** — OTel endpoint is existing config
- Secrets / tokens / credentials handling changed? **No** — no content capture in this PR
- PII or personal data in diff, tests, fixtures, or docs? **No**

## Compatibility

- Backward compatible? **Yes** — all new `ObserverEvent` fields are `Option<_>` with `None` defaults. Downstream observers using `..` in match arms require no changes.
- Config / env / CLI surface changed? **No** — no new config keys in this slice (the `log_llm_io` keys from #7233 are not carried over).

## Rollback

**risk: high**

```bash
git revert 36ca7f107fd0b5bfea33d5303a5e8037b10bf3ba a8d8898aa9349efb976b85d5ce30c5e9b214802d 47cabfbcf5788660ab8bacea9047ef9f19ceb3af 65da49c4bd3e390d477d32cfbc5b5ac045eda2cc 33811339f61432a447d35057caa7ee7d67bb8951 e2f2f1b4a12183b20d3ae972b806c147e5d723f8

```

Six commits must be reverted in reverse order. The ObserverEvent schema expansion, agent.rs turn lifecycle guard, and OtelObserver state machine touch hot paths (every agent turn). Verify OTel trace structure, Langfuse turn statistics, and SSE broadcast JSON shape after revert.

---


Audacity88 (changes requested):
I reviewed the live PR state, the public PR diff against `master`, the PR-head `agent.rs` non-streaming turn lifecycle, and the `otel.rs` span-map open/close behavior. I did not rerun local cargo validation or inspect unrelated runtime paths.

### 🟢 What looks good — Scoped metadata-first observability slice

The direction is solid: this keeps the change focused on turn metadata and correlation instead of mixing in content capture or broader log migration, and the gateway compatibility path preserves the existing total `tokens_used` field while exposing the new input/output split. The added tests around event projection, SSE payloads, and OTel span open/close also give the right surfaces to extend once the missed non-streaming exits are fixed.

### 🔴 Blocking — Close the observer turn on every exit path

The PR opens an `AgentStart` turn before the non-streaming loop at `crates/zeroclaw-runtime/src/agent/agent.rs:2087`, but some exits still return without a matching `AgentEnd`: `prepare_provider_messages(...).await?` can short-circuit at `crates/zeroclaw-runtime/src/agent/agent.rs:2097`, and a response-cache hit returns at `crates/zeroclaw-runtime/src/agent/agent.rs:2116`.

For the new `OtelObserver`, `AgentStart` inserts a live parent span keyed by `turn_id`, and the map is only drained for that turn when an `AgentEnd` arrives. These paths therefore leave the parent span open until `flush()` drains it or the process shuts down, and they also lose the duration/token metadata this PR is adding to `AgentEnd`. Non-OTel observers see an incomplete turn lifecycle too. Please centralize the end-of-turn emission behind a guard/finalizer or add the missing `AgentEnd` calls to every non-streaming return path, and add regression coverage for at least the cache-hit and prepare-error paths.

FTDGRT:
> I reviewed the live PR state, the public PR diff against `master`, the PR-head `agent.rs` non-streaming turn lifecycle, and the `otel.rs` span-map open/close behavior. I did not rerun local cargo validation or inspect unrelated runtime paths.
> 
> ### 🟢 What looks good — Scoped metadata-first observability slice
> The direction is solid: this keeps the change focused on turn metadata and correlation instead of mixing in content capture or broader log migration, and the gateway compatibility path preserves the existing total `tokens_used` field while exposing the new input/output split. The added tests around event projection, SSE payloads, and OTel span open/close also give the right surfaces to extend once the missed non-streaming exits are fixed.
> 
> ### 🔴 Blocking — Close the observer turn on every exit path
> The PR opens an `AgentStart` turn before the non-streaming loop at `crates/zeroclaw-runtime/src/agent/agent.rs:2087`, but some exits still return without a matching `AgentEnd`: `prepare_provider_messages(...).await?` can short-circuit at `crates/zeroclaw-runtime/src/agent/agent.rs:2097`, and a response-cache hit returns at `crates/zeroclaw-runtime/src/agent/agent.rs:2116`.
> 
> For the new `OtelObserver`, `AgentStart` inserts a live parent span keyed by `turn_id`, and the map is only drained for that turn when an `AgentEnd` arrives. These paths therefore leave the parent span open until `flush()` drains it or the process shuts down, and they also lose the duration/token metadata this PR is adding to `AgentEnd`. Non-OTel observers see an incomplete turn lifecycle too. Please centralize the end-of-turn emission behind a guard/finalizer or add the missing `AgentEnd` calls to every non-streaming return path, and add regression coverage for at least the cache-hit and prepare-error paths.

Hi,@Audacity88 Thanks for the thorough review. All changes are in 36ca7f107fd0b5bfea33d5303a5e8037b10bf3ba.

### What changed

I introduced a `TurnGuard` struct that holds the turn metadata (`turn_id`, `agent_alias`, `model_provider`, `model`, `turn_started_at`, token accumulators) and emits `AgentEnd` in its `Drop` implementation. The guard is created right after `AgentStart` in both the non-streaming (`turn`) and streaming (`turn_streamed_with_steering_state`) code paths, as well as the top-level `run()` entry point.

All inline `AgentEnd` calls have been removed — the guard fires automatically on every exit path, including:

- `prepare_provider_messages()` error (the short-circuit at the top of the loop)
- Response-cache hit (returns immediately with cached text)
- Cancellation during tool execution
- Cancellation during streaming
- Stream errors
- Normal completion (final text / max iterations exceeded)

The `done` flag prevents double-firing if `fire()` is called explicitly before the guard drops.

### Regression coverage

Added three tests:

1. **`turn_cache_hit_emits_agent_end_with_none_tokens`** — seeds the response cache with a first turn, then runs a second turn that hits the cache. Asserts exactly one `AgentStart`/`AgentEnd` pair, that `AgentEnd` carries `tokens_used: None`, and that no `LlmRequest` was emitted (confirming the cache was actually hit). This directly covers the cache-hit path you flagged.

2. **`turn_streamed_cancel_during_tool_execution_emits_agent_end_with_tokens`** — uses a `SlowTool` (60s sleep) and cancels after a `TurnEvent::Usage` event is received, while the slow tool execution is still pending. Asserts a single `AgentStart`/`AgentEnd` pair, that `AgentEnd` includes the expected token counts from the mock LLM response, and that `LlmResponse` precedes `AgentEnd`.

3. **`turn_llm_error_emits_agent_end`** — uses a `FailingModelProvider` that always returns an error from `chat()`. Asserts the turn fails, that a single `AgentStart`/`AgentEnd` pair is emitted, and that `AgentEnd` carries `tokens_used: None`. This exercises the same `TurnGuard`/drop-based cleanup mechanism used by other early-return paths, including `prepare_provider_messages` failures.

All three tests use a shared `assert_single_agent_lifecycle` helper that validates the 1:1 start/end pairing and matching `turn_id`.

I verified the streaming path through the Langfuse backend, confirming that `AgentEnd` statistics are emitted correctly.

WareWolf-MoonWall (commented):
# PR #7385 — Review draft
**PR**: feat(observability): add turn metadata to observer events and correlate OTel spans by turn_id
**Author**: FTDGRT
**Verdict**: `--comment`
**Reason**: Audacity88 has an active CHANGES_REQUESTED. The TurnGuard response to that blocking finding looks correct and the regression tests are solid, so I have no new blockers to add. Leaving this as a comment so Audacity88 can re-review the fix and clear the CR independently. I do have one data-accuracy warning on the gateway compat path that should be addressed before or alongside that re-review.

---

I reviewed the full diff (+1484 -201), the `TurnGuard` implementation in `crates/zeroclaw-runtime/src/agent/agent.rs`, the `OtelObserver` span-map logic, the SSE broadcast adapter in `crates/zeroclaw-gateway/src/sse.rs`, the `observer_bridge` projection changes, and the three new regression tests. I also checked the `observability_traits.rs` schema expansion against AGENTS.md DRY rules.

### ✅ Resolved — TurnGuard closes every exit path (Audacity88's blocker)

Audacity88 flagged that `AgentEnd` was missing on the `prepare_provider_messages()` early-return path and the response-cache hit path, leaving the `OtelObserver` parent span open until `flush()`.

The `TurnGuard` struct with a `Drop` impl solves this cleanly. Both `turn()` (non-streaming) and `turn_streamed_with_steering_state()` create the guard immediately after emitting `AgentStart`, so every subsequent return path — error propagation, cache hit, cancellation, stream error, max-iterations exhaustion, and normal completion — fires `AgentEnd` automatically. The `done` flag prevents double-firing if `fire()` is called explicitly before drop. I verified this against the diff: I see no remaining naked `return` or `?` paths after the guard is created in either turn function.

The three new tests directly cover the paths originally flagged:
- `turn_cache_hit_emits_agent_end_with_none_tokens` — cache hit path
- `turn_llm_error_emits_agent_end` — prepare-provider-messages failure (FailingModelProvider)
- `turn_streamed_cancel_during_tool_execution_emits_agent_end_with_tokens` — streaming cancel

The shared `assert_single_agent_lifecycle` helper enforcing 1:1 `AgentStart`/`AgentEnd` pairing with matching `turn_id` is the right approach. I consider Audacity88's blocker addressed; the CR ball is back in their court.

### 🟡 Warning — Gateway webhook puts total tokens into `input_tokens`, hardcodes `output_tokens: 0`

In `crates/zeroclaw-gateway/src/lib.rs`, the webhook path had a flat `tokens_used: Option<u64>` representing a combined total. The adaptation to `TurnTokenUsage` is:

```rust
tokens_used: tokens_used.map(|t| TurnTokenUsage {
    input_tokens: t,
    output_tokens: 0,
}),
```

The `tokens_total` field in the SSE broadcast then computes `t + 0 = t`, which preserves the old behavior correctly. But the broadcast also now emits `"input_tokens": t, "output_tokens": 0` as separate keys. Any downstream observability consumer (Langfuse, custom dashboards) reading the per-direction breakdown will see the full token count attributed entirely to input with zero output — factually wrong for a real webhook turn.

Three options in increasing order of invasiveness:
1. **Minimal**: Swap to `input_tokens: 0, output_tokens: t` if the existing field better represents output (depends on the provider), or leave `tokens_used: None` for the webhook path to signal "total unavailable as split" — SSE consumers already handle `None` gracefully.
2. **Clean**: Add an `Option<u64>` total-only field to `TurnTokenUsage` for sites that can't provide a split. Larger change, probably out of scope for this PR.
3. **Acceptable punt**: Document in a follow-up issue that `input_tokens`/`output_tokens` from webhook-originated `AgentEnd` events are unreliable. At minimum, add a comment at the call site making the approximation explicit.

Option 1 (use `None`) is the safest fix within this PR's scope.

### 🔵 Suggestion — Set `channel: Some("webhook")` in the gateway webhook handler

`handle_webhook` now passes `channel: None` for both `AgentStart` and `AgentEnd`. Since this is definitionally a webhook-originated turn, `channel: Some("webhook".to_string())` could be set here at zero architectural cost, giving OTel traces immediate attribution for this path. The PR description explicitly scopes out channel population in the agent path, but the gateway path is different — it knows its own origin. Worth considering even in this slice.

### 🟢 What looks good — TurnGuard design is correct and doesn't introduce duplicate state

Holding `model_provider`, `model`, `agent_alias`, and `turn_id` copies in `TurnGuard` rather than borrowing them from `Agent` is correct: these strings must survive to `Drop`, which may run after the containing scope has moved or dropped parts of `Agent`. Under AGENTS.md's DRY rules, these are not caches of config state — they are scoped locals captured at turn start for a single emission. The `total_input_tokens` / `total_output_tokens` accumulators in the guard are likewise the canonical counters for this turn; no equivalent field exists in `Agent` that they would duplicate.

The `active_agent_spans: Mutex<HashMap<String, (BoxedSpan, Context)>>` in `OtelObserver` is correct ephemeral in-process state, not a cache of any config or agent field. `flush()` draining orphaned spans on shutdown is the right safety valve.

### 🟢 What looks good — `ObserverEvent` schema expansion is backward compatible and safe

All six expanded variants get `Option<String>` fields defaulting to `None`. Observers using `..` in match arms compile without changes. The `TurnTokenUsage` replacement in `AgentEnd` generates a compile-time error at explicit match sites — intentional, surfacing all the call sites that need updating as a compile-time gate rather than a runtime surprise. The `sse.rs` gateway compat path correctly preserves the old `tokens_used` total-sum key alongside the new per-direction keys, maintaining backward compatibility for existing SSE consumers.

### 🟢 What looks good — Validation Evidence is thorough

Full `cargo fmt`, `cargo clippy --locked --all-targets -- -D warnings`, and `cargo test --locked --all-targets` are all shown with literal output. End-to-end Langfuse backend verification is described. The Rollback section lists all six commit SHAs in reverse-revert order with the correct call-outs for which paths are hot (agent loop, OtelObserver state machine, SSE broadcast shape). This is the level of template completeness I want to see from an XL risk: high PR.

---

**My ask before I shift to approve:**
1. Address the `input_tokens: t, output_tokens: 0` issue in the gateway webhook compat path — at minimum a call-site comment, preferably `tokens_used: None` to avoid emitting a misleading split.
2. Audacity88 re-reviews the TurnGuard fix and clears the CHANGES_REQUESTED.

Audacity88 (changes requested):
Context: I re-reviewed the current head after the TurnGuard update, including the live PR state, the full diff against the PR base, the PR's stated scope boundary, the previous Audacity88 `CHANGES_REQUESTED`, @FTDGRT's response, and @WareWolf-MoonWall's follow-up review. I also inspected the `TurnGuard` paths in `agent.rs`, the added lifecycle regression tests, the OTel span-map behavior, and the gateway/SSE token projection path.

### ✅ Resolved — TurnGuard closes the original missing `AgentEnd` paths

The original blocker was that `turn()` opened an `AgentStart` before paths that could still return without a matching `AgentEnd`, especially the response-cache hit and early provider/preparation error paths.

The current `TurnGuard` design addresses that for the covered `Agent::turn()` and `turn_streamed_with_steering_state()` paths. The guard is created immediately after `AgentStart`, its `Drop` implementation emits `AgentEnd`, and the `done` flag prevents double emission when the guard is explicitly fired. The new tests also check the important shape: exactly one start, exactly one end, matching `turn_id`, cache hit without `LlmRequest`, LLM error cleanup, and streaming cancellation after token usage.

That resolves my previous `CHANGES_REQUESTED` item.

### 🔴 Blocking — The gateway publishes an aggregate as a concrete split

The webhook path in `handle_webhook()` derives a single aggregate `tokens_used` total, then stores that total as `TurnTokenUsage { input_tokens: t, output_tokens: 0 }`.

That preserves the legacy aggregate `tokens_used` total after `sse.rs` sums the two fields, but it turns an unknown split into concrete zero-output data. SSE now publishes both `"input_tokens"` and `"output_tokens"` from `TurnTokenUsage`, so a webhook turn with 15 total tokens becomes `"input_tokens": 15, "output_tokens": 0`, even though the call site has not established that all 15 tokens were input tokens. The same bad split can also flow to any observer that reads `AgentEnd.tokens_used` as structured input/output usage rather than only summing it.

I am blocking on this because this PR changes `AgentEnd.tokens_used` from a flat total into a structured input/output contract, and `sse.rs` exposes that structure to downstream consumers. Please keep the structured fields truthful: use the real input/output split where this path has it, or leave `tokens_used: None` when only a total is known. This PR's main promise is better turn metadata and structured usage, so the new structured fields need to be absent rather than misleading when the split is unavailable.

### 🟢 What looks good — The narrowed first slice is the right shape

This version keeps the first-slice boundary much cleaner than #7233: no LLM prompt/response content capture, no content-policy config surface, no broad `record!` migration claim, and no claim that every runtime path is already correlated. That makes the metadata/correlation change reviewable.

The schema expansion is also mostly conservative: optional attribution fields let existing explicit call sites fail at compile time where they need adaptation, while observers using `..` keep working. The OTel parent-span map keyed by `turn_id` is the right mechanism for a turn-scoped trace, and ending orphaned spans in `flush()` is a useful backstop rather than the primary lifecycle guarantee.

Audacity88:
@singlerider @theonlyhennygod — milestone alignment needed: this PR is a blocked `feat(observability)` slice, and I do not see a clear fit within the current open milestone scopes. Please advise on placement or deferral once the review blocker is addressed.

FTDGRT:
> Context: I re-reviewed the current head after the TurnGuard update, including the live PR state, the full diff against the PR base, the PR's stated scope boundary, the previous Audacity88 `CHANGES_REQUESTED`, @FTDGRT's response, and @WareWolf-MoonWall's follow-up review. I also inspected the `TurnGuard` paths in `agent.rs`, the added lifecycle regression tests, the OTel span-map behavior, and the gateway/SSE token projection path.
> 
> ### ✅ Resolved — TurnGuard closes the original missing `AgentEnd` paths
> The original blocker was that `turn()` opened an `AgentStart` before paths that could still return without a matching `AgentEnd`, especially the response-cache hit and early provider/preparation error paths.
> 
> The current `TurnGuard` design addresses that for the covered `Agent::turn()` and `turn_streamed_with_steering_state()` paths. The guard is created immediately after `AgentStart`, its `Drop` implementation emits `AgentEnd`, and the `done` flag prevents double emission when the guard is explicitly fired. The new tests also check the important shape: exactly one start, exactly one end, matching `turn_id`, cache hit without `LlmRequest`, LLM error cleanup, and streaming cancellation after token usage.
> 
> That resolves my previous `CHANGES_REQUESTED` item.
> 
> ### 🔴 Blocking — The gateway publishes an aggregate as a concrete split
> The webhook path in `handle_webhook()` derives a single aggregate `tokens_used` total, then stores that total as `TurnTokenUsage { input_tokens: t, output_tokens: 0 }`.
> 
> That preserves the legacy aggregate `tokens_used` total after `sse.rs` sums the two fields, but it turns an unknown split into concrete zero-output data. SSE now publishes both `"input_tokens"` and `"output_tokens"` from `TurnTokenUsage`, so a webhook turn with 15 total tokens becomes `"input_tokens": 15, "output_tokens": 0`, even though the call site has not established that all 15 tokens were input tokens. The same bad split can also flow to any observer that reads `AgentEnd.tokens_used` as structured input/output usage rather than only summing it.
> 
> I am blocking on this because this PR changes `AgentEnd.tokens_used` from a flat total into a structured input/output contract, and `sse.rs` exposes that structure to downstream consumers. Please keep the structured fields truthful: use the real input/output split where this path has it, or leave `tokens_used: None` when only a total is known. This PR's main promise is better turn metadata and structured usage, so the new structured fields need to be absent rather than misleading when the split is unavailable.
> 
> ### 🟢 What looks good — The narrowed first slice is the right shape
> This version keeps the first-slice boundary much cleaner than #7233: no LLM prompt/response content capture, no content-policy config surface, no broad `record!` migration claim, and no claim that every runtime path is already correlated. That makes the metadata/correlation change reviewable.
> 
> The schema expansion is also mostly conservative: optional attribution fields let existing explicit call sites fail at compile time where they need adaptation, while observers using `..` keep working. The OTel parent-span map keyed by `turn_id` is the right mechanism for a turn-scoped trace, and ending orphaned spans in `flush()` is a useful backstop rather than the primary lifecycle guarantee.

Hi,@Audacity88 The gateway publishes an aggregate as a concrete split. Fixed in 927a2e2a4d9f3e03b56684cc27783007dbe7d34f.

**Problem**: The old code collapsed input/output into a total via `.zip().map(|i,o| i+o).or(...)`, then constructed `TurnTokenUsage { input_tokens: t, output_tokens: 0 }` — a misleading split published to SSE consumers.

**Fix**: The webhook's `input_tokens` / `output_tokens` come from `TurnUsage`, which accumulates them separately from the LLM provider's usage response — they are always paired. Changed to `.zip().map(|(i,o)| TurnTokenUsage { input_tokens: i, output_tokens: o })`: real split when data exists, `None` otherwise. `LlmResponse` now also forwards the real values instead of `None`.

Audacity88 (approved):
Reviewed current head `7a695b89` against the public diff from `master`, RFC #7232, superseded PR #7233, the prior Audacity88 and @WareWolf-MoonWall reviews, @FTDGRT's latest response, the empty inline-thread state, and current green CI. I did not rerun local cargo validation. This clears my previous review blockers; merge-readiness still needs the normal freshness/body cleanup pass because the branch is behind current `master`.

### ✅ Resolved — TurnGuard closes the missing `AgentEnd` paths

The original blocker was that `AgentStart` could be emitted before exits that returned without a matching `AgentEnd`, especially the response-cache hit and early provider/error paths.

The current `TurnGuard` design fixes that for the covered `Agent::turn()` and `turn_streamed_with_steering_state()` paths. The guard is created immediately after `AgentStart`, emits `AgentEnd` from `Drop`, and uses the `done` flag to avoid double emission. The added lifecycle tests cover the important failure modes: cache hit without an LLM request, LLM error cleanup, and streaming cancellation after token usage. That resolves my first `CHANGES_REQUESTED` item.

### ✅ Resolved — Gateway webhook no longer fabricates a token split

The newer gateway fix also resolves the structured-token blocker. `handle_webhook()` now builds `TurnTokenUsage` only from `input_tokens.zip(output_tokens)`, so it emits the input/output split only when both sides are present instead of turning an aggregate total into `input_tokens: total, output_tokens: 0`. The same real split is also forwarded on `LlmResponse`.

That keeps the legacy SSE `tokens_used` total correct while avoiding a misleading per-direction breakdown for consumers that read the new fields directly.

### 🟢 What looks good — the replacement slice matches the RFC boundary

This replacement is much cleaner than #7233. It keeps the first merge target focused on metadata, `turn_id`, bridge projection, and OTel trace correlation, while leaving LLM/tool content capture, broad `record!` migration, and full runtime-path correlation out of scope. That matches the staged routing we asked for on RFC #7232 and makes the privacy/security surface reviewable.

The OTel parent-span map keyed by `turn_id`, the optional observer metadata fields, and the bridge projection tests are the right shape for this slice. The remaining uncovered paths degrade to `None`/orphaned spans rather than pretending to be fully correlated, which is the safer boundary for this PR.

### 🔵 Suggestion — refresh the stale PR-body commit and rollback sections before merge

The PR body still names the older commit stack in its commit and rollback sections. The current non-merge PR stack includes the later `0ff9a4f` TurnGuard fix and `927a2e2` gateway token-split fix as well, so those sections should be refreshed before landing. I am not treating that as an author-side blocker because it is body cleanup, not a remaining code issue.

PR #7386, opened 2026-06-08
fix(channels): avoid unused PacedChannel import

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Remove the unconditional `PacedChannel` import from `orchestrator/mod.rs`.
  - Qualify the `PacedChannel::wrap(...)` call sites directly inside the existing feature-gated channel blocks.
  - Keep `clippy -D warnings` from tripping on an unused import when those channel features are not active for the current target.
- **Scope boundary:** This only fixes the `PacedChannel` unused-import lint path in `zeroclaw-channels`. It does not change channel behavior or pacing logic.
- **Blast radius:** `zeroclaw-channels` compile/lint behavior only.
- **Linked issue(s):** None
- **Related PR(s):** Related #7379
- **Current labels:** `bug`, `risk: medium`, `size: XS`, `channel`

```text
zc-cargo fmt -p zeroclaw-channels -- --check

zc-cargo clippy -p zeroclaw-channels --lib -- -D warnings
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 10m 52s

git diff --check
  No output.
```

- **Beyond CI — what did you manually verify?** Checked that the only change is replacing the top-level import with fully qualified `crate::paced_channel::PacedChannel::wrap(...)` calls in the already feature-gated channel setup paths.
- **If any command was intentionally skipped, why:** I kept validation scoped to the crate and lint path that was failing. No broader workspace validation was needed for this one-file lint-only fix.

- If any `Yes`, describe the risk and mitigation: `Not applicable`

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: `No upgrade steps needed`

- **Fast rollback command/path:** `git revert cbaf6b3b20af3433e799045b49be225cbaa77812`
- **Observable failure symptoms:** Feature-gated channel builds fail in `crates/zeroclaw-channels/src/orchestrator/mod.rs`, or the original `unused import: crate::paced_channel::PacedChannel` clippy error reappears when channel features are not active.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -112,7 +112,6 @@ use std::sync::{Arc, Mutex};
 use std::time::{Duration, Instant, SystemTime};
 use tokio_util::sync::CancellationToken;
 
-use crate::paced_channel::PacedChannel;
 use zeroclaw_api::session_keys::sanitize_session_key;
 use zeroclaw_config::schema::Config;
 use zeroclaw_memory::{self, MEMORY_CONTEXT_CLOSE, MEMORY_CONTEXT_OPEN, Memory};
@@ -6227,7 +6226,7 @@ fn collect_configured_channels(
         channels.push(ConfiguredChannel {
             display_name: "Telegram",
             alias: Some(alias.clone()),
-            channel: PacedChannel::wrap(
+            channel: crate::paced_channel::PacedChannel::wrap(
                 Arc::new(
                     TelegramChannel::new(
                         tg.bot_token.clone(),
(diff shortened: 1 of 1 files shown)


singlerider (approved):
Read the diff. Correct lint fix, idiomatic approach. Approving.

### 🟢 What looks good — fully-qualifying at the gated call sites is the right fix

The `use crate::paced_channel::PacedChannel;` was only consumed inside feature-gated channel blocks, so a build without those channel features left it unused and tripped `clippy -D warnings`. Removing the top-level import and qualifying each `crate::paced_channel::PacedChannel::wrap(...)` call inline is the idiomatic resolution: every call site already lives inside a `#[cfg]`/feature-gated block, so the path only resolves when the feature is on, and there's no import to go unused when it's off. That's cleaner than the alternatives (a `#[cfg(any(feature...))]` on the import enumerating every channel feature, or an `#[allow(unused_imports)]` that silences the lint instead of fixing it). The `wrap` calls are byte-for-byte the same otherwise, so pacing behavior is untouched — this is purely the import path. Approving.

Issue #5941, opened 2026-04-20
[Bug]: No tool call found for function call output with call_id call_6gd5XWS7qozFQYJdlm2k0UJI.


PeterlitsZo:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

When I try send:

> Hello

The zeroclaw says:

> ⚠️ Error: All providers/models failed. Attempts:
provider=custom:https://*********.****.****/v1 model=gpt-5.4 attempt 1/3: non_retryable; error=Custom API error (400 Bad Request): {"error":{"message":"No tool call found for function call output with call_id call_6gd5XWS7qozFQYJdlm2k0UJI.","type":"invalid_request_error","param":"input","code":null}}

### Expected behavior

Reply normally.

### Steps to reproduce

```bash
Hard to reproduce...
```

### Logs / stack traces

```text

```

### ZeroClaw version

v0.7.0 / 0f8bd00

### Operating system

NixOS 25.11

### Regression?

Unknown

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

thomas-quant:
I have the same issue, using cliproxyapi - in my case seemed to happen after a long output - possibly resulting in a malformed tool call / left hanging?

Initially resolved via disabling session persistence, which worked, then removed lines from the message that contained the failed tool call -- fixed it.

Audacity88:
This should be addressed by PR #6035 (approved, pending merge) — it adds stable UUID-based tool call IDs with proper correlation between `ToolCall` and `ToolResult` events. The current code reuses tool names as IDs, which breaks when the provider expects a specific `call_id` format for function call output correlation.

PeterlitsZo:
Thank you for your friendly reply. I get it!

PeterlitsZo:
Feel free to close this issue.

Audacity88:
Thanks @PeterlitsZo — will close once #6035 merges.

vyahhi:
Duplicate backlink: #5865 appears to report the same OpenAI-compatible/custom-provider failure (`No tool call found for function call output with call_id ...`).

This issue still looks like the better canonical tracker because it already has the PR #6035 linkage and clearer discussion of tool-call/result correlation.

PR #7387, opened 2026-06-08
fix(providers): default custom endpoints to prompt-guided tools

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Custom OpenAI-compatible endpoints now default to prompt-guided tool use instead of native tool-call payloads.
  - This preserves the #5242 fix direction for strict custom endpoints that reject persisted native tool-result history with errors such as `No tool call found for function call output with call_id ...`.
  - Operators who know their custom endpoint supports native tool calling can still opt in with `native_tools = true`.
  - Regression coverage now checks both the safe default and the explicit native-tools override in the current `zeroclaw-providers` factory layout.
- **Scope boundary:** This does not change first-party OpenAI/OpenRouter/Anthropic provider behavior, the Responses wire path, channel delivery, session persistence, or native-tool history construction.
- **Blast radius:** Only `[providers.models.custom.<alias>]` entries that use the chat-completions wire path change default tool mode. Those endpoints may use prompt-guided tools instead of provider-native function calls unless `native_tools = true` is set.
- **Linked issue(s):** Supersedes #5242. Related #6298 and #5941.
- **Labels:** `bug`, `risk: medium`, `size: XS`, `provider`

```bash
$ git diff --check
# clean, exit 0

$ rustfmt --edition 2024 --check crates/zeroclaw-providers/src/factory.rs
# clean, exit 0

$ cargo fmt --all -- --check
# clean, exit 0

$ cargo test -p zeroclaw-providers custom_factory_ --lib
    Finished `test` profile [unoptimized + debuginfo] target(s) in 8m 58s
     Running unittests src/lib.rs (target/debug/deps/zeroclaw_providers-90fcc1d3d6dfaaf4)

running 4 tests
test factory::tests::custom_factory_routes_to_responses_provider_when_wire_api_responses ... ok
test factory::tests::custom_factory_defaults_to_chat_completions_without_wire_api ... ok
test factory::tests::custom_factory_honors_native_tools_override_true ... ok
test factory::tests::custom_factory_disables_native_tools_by_default ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 909 filtered out; finished in 0.00s
```

- **Beyond CI — what did you manually verify?** Checked current `master` provider construction directly: the old #5242 path was deleted, and the live custom provider factory now lives in `crates/zeroclaw-providers/src/factory.rs`. The Responses-wire custom path returns before the chat-completions provider is built, so this change does not affect `wire_api = "responses"`.
- **If any command was intentionally skipped, why:** Full workspace clippy and full tests were not run locally. The patch is a one-file provider factory default plus focused tests; broader checks can run in CI.

## Compatibility (required)

- Backward compatible? `No` for custom chat-completions endpoints that relied on native tool calling by default; existing configs still load.
- If `No` or `Yes` to either: Operators whose custom endpoint requires native tool calling should set `native_tools = true` on that custom provider entry.

- **Fast rollback command/path:** `git revert 2becaca285f792ddfc121b0ea9a6fc01a0271130`
- **Feature flags or config toggles:** Existing `native_tools = true` per custom provider entry restores native tool calling without reverting.
- **Observable failure symptoms:** If prompt-guided fallback regresses a custom endpoint, tool calls may become less structured or less reliable for that endpoint. If native tools remain enabled on a strict endpoint, provider errors may include `No tool call found for function call output with call_id ...` or related native tool-history validation failures.

  - #5242 by @bellman-move
- Scope materially carried forward: the custom-provider default should disable native tool calling for strict OpenAI-compatible endpoints; this PR rebuilds that fix in the current `zeroclaw-providers` factory layout and adds the current-layout default/override regression tests.
- If `No`, why (inspiration-only, no direct code/design carry-over): The patch was rebuilt against the current crate layout instead of cherry-picking the deleted `src/providers/mod.rs` implementation. The superseded PR and author are credited here.

### DIFF
--- a/crates/zeroclaw-providers/src/factory.rs
+++ b/crates/zeroclaw-providers/src/factory.rs
@@ -1222,6 +1222,9 @@ impl FamilyProviderFactory for CustomModelProviderConfig {
             AuthStyle::Bearer,
             true,
         );
+        if opts.native_tools != Some(true) {
+            p = p.without_native_tools();
+        }
         if opts.merge_system_into_user {
             p = p.with_merge_system_into_user();
         }
@@ -1507,6 +1510,51 @@ mod tests {
             )
             .unwrap();
         assert_ne!(provider.default_wire_api(), "responses");
+    }
+
+    #[test]
+    fn custom_factory_disables_native_tools_by_default() {
+        use zeroclaw_config::schema::{CustomModelProviderConfig, ModelProviderConfig};
+        let cfg = CustomModelProviderConfig {
+            base: ModelProviderConfig {
+                uri: Some("http://10.0.0.15:8000/v1".to_string()),
+                ..Default::default()
+            },
+        };
+        let provider = cfg
+            .create_provider(
+                "vllm",
+                None,
+                Some("http://10.0.0.15:8000/v1"),
+                &ModelProviderRuntimeOptions::default(),
+            )
+            .unwrap();
+        assert!(
+            !provider.supports_native_tools(),
+            "custom OpenAI-compatible endpoints must default to prompt-guided tools"
+        );
+    }
+
+    #[test]
+    fn custom_factory_honors_native_tools_override_true() {
+        use zeroclaw_config::schema::{CustomModelProviderConfig, ModelProviderConfig};
+        let cfg = CustomModelProviderConfig {
+            base: ModelProviderConfig {
+                uri: Some("http://10.0.0.15:8000/v1".to_string()),
+                ..Default::default()
+            },
+        };
+        let options = ModelProviderRuntimeOptions {
+            native_tools: Some(true),
+            ..Default::default()
+        };
+        let provider = cfg
+            .create_provider("vllm", None, Some("http://10.0.0.15:8000/v1"), &options)
+            .unwrap();
+        assert!(
+            provider.supports_native_tools(),
+            "custom endpoints with `native_tools = true` must keep native tool calling available"
+        );
     }
 
     #[test]


singlerider (approved):
Read the diff and confirmed the wire-path scoping. Correct, minimal default fix. Approving.

### 🟢 What looks good — the default is the safe one and the override condition is exactly right

Defaulting `[providers.models.custom.<alias>]` chat-completions endpoints to prompt-guided tools is the correct call: strict OpenAI-compatible servers (vLLM, local inference endpoints) routinely reject persisted native tool-result history with `No tool call found for function call output with call_id ...` — the #5242 symptom — so native-by-default makes the common custom-endpoint case fail out of the box. The gate `opts.native_tools != Some(true)` is the right condition: both `None` (unset) and `Some(false)` disable native, and only an explicit `Some(true)` opts in, so an operator who knows their endpoint supports native function calling can still get it. The breaking change is flagged honestly (`Backward compatible: No` for endpoints that leaned on the native default) with a one-line migration.

### 🟢 What looks good — the Responses-wire path is provably untouched

I confirmed the scope claim: `build_responses_provider_if_requested` returns early when `wire_api = "responses"` (factory.rs, before the chat-completions provider is constructed), so the new `without_native_tools()` only ever applies on the chat-completions custom path. `wire_api = "responses"` custom endpoints are unaffected, which is what the PR states and what `custom_factory_routes_to_responses_provider_when_wire_api_responses` continues to cover.

### 🟢 What looks good — tests pin both branches

`custom_factory_disables_native_tools_by_default` (asserts `!supports_native_tools()` with no override) and `custom_factory_honors_native_tools_override_true` (asserts it's restored with `native_tools = Some(true)`) cover the exact decision this PR adds, in the current `zeroclaw-providers` factory layout. Supersede attribution for #5242 by @bellman-move is credited, rebuilt against the current crate path rather than cherry-picking the deleted `src/providers/mod.rs`. Approving.

PR #7388, opened 2026-06-08
fix(matrix): isolate session state per channel alias and repair key backup with the configured key

### LINKED ISSUE #6487: [Bug]: Multi-alias Matrix instances share one session store and clobber each other (wrong account on the wire); inconsistent key backup never repaired
## Affected component

channel

## Severity

S1

## Current behavior

Two `[channels.matrix.<alias>]` blocks under one daemon share a single on-disk Matrix session store, silently clobbering each other. The `state_dir` is derived without the alias in both construction sites in `crates/zeroclaw-channels/src/orchestrator/mod.rs`:

```rust
let state_dir = config
    .config_path
    .parent()
    .map(|p| p.join("state").join("matrix"))
    .unwrap_or_else(|| std::path::PathBuf::from(".zeroclaw/state/matrix"));
```

Both Matrix listeners resolve to the same `<config_dir>/state/matrix/` directory regardless of alias, so both restore the same `session.json` and attach to whichever account wrote the blob first. The other block then runs under the wrong Matrix account while still stamping inbound traffic with its own configured alias. The observable symptom is "the wrong bot answered my DM": a direct message to one account is dispatched to the other agent.

Separately, an account whose key backup is inconsistent loops forever on `Trying to backup room keys but no backup key was found`, because the recovery path only imports secrets and cannot repair an inconsistent server-side key backup.

This is a data-loss / security-risk classification: E2EE session material for one account is silently shared with another.

## Expected behavior

- Each `[channels.matrix.<alias>]` block is backed by its own isolated session store at `<config_dir>/state/matrix/<alias>/`, restores its own session, and runs as its own configured account.
- A saved session whose canonical MXID does not match the configured `user_id` is wiped and re-logged-in rather than impersonated.
- An account with a configured `recovery_key` and an inconsistent server-side key backup has that backup repaired with the configured key (never rotating the secret-storage key).

## Steps to reproduce

```toml
[channels.matrix.bot-a]
homeserver = "https://matrix.org"
user_id = "@bot-a:matrix.org"
password = "<REDACTED>"
recovery_key = "<REDACTED>"

[channels.matrix.bot-b]
homeserver = "https://matrix.org"
user_id = "@bot-b:matrix.org"
password = "<REDACTED>"
recovery_key = "<REDACTED>"
```

1. Start the daemon with the above config.
2. Observe: no startup error.
3. Both blocks resolve to the same `<config_dir>/state/matrix/` dir and restore the same `session.json`.
4. Send a DM to `@bot-a`'s Matrix account.
5. The `bot-b` agent answers it (it is attached to `@bot-a`'s session while stamping `channel_alias=bot-b`).

## Impact

Affected users: any operator running two or more Matrix channel blocks under one daemon.
Frequency: always — deterministic at startup.
Consequence: wrong account on the wire, cross-account E2EE session sharing, and a stuck key-backup loop for accounts with inconsistent backups.

## Logs / stack traces

```
channel_alias=bot-a   matrix: restored session from session.json
channel_alias=bot-b   matrix: restored session from session.json
channel_alias=bot-a   matrix: secret-storage diagnostics: default_key_id=<KEYID-X>
channel_alias=bot-b   matrix: secret-storage diagnostics: default_key_id=<KEYID-X>
channel_alias=bot-b   matrix: starting sync loop            # attached to @bot-a
```

## ZeroClaw version

`integration/v0.8.0` @ `291a4d57893bd147ccf8dc38d7bb5e1f3beb51a9`

## Operating system

Linux (platform-independent).

## Regression?

No, this is an unreleased feature with an Issue to track this known limitation.

singlerider:
> ⚠️ **BREAKING CHANGE — Matrix session migration required.** Multiple Matrix blocks under one daemon previously shared a single session store at `<config_dir>/state/matrix/`; this PR moves each block to a per-alias path `<config_dir>/state/matrix/<alias>/`. There is no clean automatic migration: on first start after upgrade each block logs in fresh under its configured `user_id` and re-bootstraps E2EE. Operators who want to preserve an existing single-block session (avoid a re-login / E2EE re-bootstrap) can move the session manually before starting:
>
> ```bash
> # Stop the daemon first.
> cd <config_dir>/state/matrix
>
> # NOTE: <alias> is the name in your config block header
> #   [channels.matrix.<alias>]  ->  e.g. [channels.matrix.bot-a] means alias "bot-a".
> # A block written as plain [channels.matrix] (no name) uses the alias "default".
>
> # Single-block installs: move the existing top-level session into the
> # default alias dir (a block with no explicit alias uses "default").
> mkdir -p default
> mv session.json store default/ 2>/dev/null || true
>
> # Multi-block installs: there is only ONE old session.json, and it belongs
> # to whichever account logged in last. Move it under THAT block's alias
> # (its [channels.matrix.<alias>] header name) and let every OTHER block log
> # in fresh:
> #   mkdir -p <winning-alias>          # e.g. bot-a
> #   mv session.json store <winning-alias>/
> # The remaining aliases re-bootstrap from their own password/recovery_key.
> ```
>
> After moving, start the daemon and confirm `matrix: restored session from session.json` appears once per alias under its own `channel_alias`. The legacy top-level `session.json`/`store/` are unused after migration and can be deleted.

## Summary

This PR fixes two distinct Matrix multi-agent defects found while diagnosing a real "wrong bot answered my DM" report, both confirmed against a live daemon.

- **Base branch:** `master`
- **What changed and why:**
  - **(1) Per-alias session isolation.** Two `[channels.matrix.<alias>]` blocks under one daemon shared a single on-disk session store at `<config_dir>/state/matrix/`. Both listeners restored the same `session.json`, so both attached to whichever account wrote the blob first; the other block ran under the wrong Matrix account while still stamping inbound with its own config alias. A DM to one bot's account was answered by the other agent. Fix: derive `state_dir` per alias (`<config_dir>/state/matrix/<alias>`) via a single `matrix_state_dir` helper used at both construction sites. Adds `saved_session_is_foreign` as a defense-in-depth guard (a saved session whose canonical MXID does not match the configured `user_id` triggers wipe-and-relogin instead of impersonation).
  - **(2) Repair an inconsistent key backup with the configured key.** Separating the sessions exposed an account whose key backup was broken, looping forever on `Trying to backup room keys but no backup key was found`. The old `run_recovery` called `recovery.recover(key)`, which only *imports* secrets and cannot repair an inconsistent server-side backup. Fix: swap to `recovery.recover_and_fix_backup(key)`, which opens secret storage with the operator's configured `recovery_key`, imports secrets, and recreates **only the key backup** (never the secret-storage key) when it is inconsistent. The configured `recovery_key` stays valid; we never rotate keys or auto-create secret storage, since that would orphan the operator's configured key.
- **Scope boundary:** Does not change message routing/dispatch (verified correct), the password/token login flow, `channel_alias` stamping, or any non-Matrix channel.
- **Blast radius:** Matrix channel only. Multi-Matrix-block daemons are affected on upgrade (Compatibility). Single-block installs move from `state/matrix/` to `state/matrix/default/`. The recovery change affects every Matrix account that runs `run_recovery`: a healthy account whose key backup is consistent behaves as before; one whose backup is inconsistent now gets it repaired using the configured key. No account has its secret-storage key rotated.
- **Linked issue(s):** Closes #6487.
- **Labels:** `bug`, `channel:matrix`, `channel`, `risk: medium`, `size: S`.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-channels --features channel-matrix --all-targets -- -D warnings
cargo clippy -p zeroclaw-channels --all-targets -- -D warnings   # default build, no feature
cargo test -p zeroclaw-channels --features channel-matrix
```

```
# cargo fmt --all -- --check                                   -> exit 0 (no output)
# cargo clippy ... --features channel-matrix ... -D warnings   -> exit 0
# cargo clippy ... (default, no feature) ... -D warnings        -> exit 0
# cargo test  ... --features channel-matrix
test result: ok. 834 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
test result: ok. 3 passed; 0 failed; 0 ignored
test result: ok. 3 passed; 0 failed; 0 ignored
test result: ok. 4 passed; 0 failed; 0 ignored
```

- **Beyond CI — what was manually verified:** Both root causes were confirmed against a live multi-agent daemon (runtime trace + on-disk state), sanitized below, and the recovery fix is **runtime-verified**: after deploy, the affected account restored its own session, logged `E2EE recovery completed` via `recover_and_fix_backup`, the `no backup key was found` loop stopped, and the agent processed turns and delivered replies on its own account. The session-isolation fix has a regression test proven RED against the pre-fix shared path. The recovery change is verified against the pinned `matrix-sdk` 0.17.0 API (`recovery.recover_and_fix_backup(key) -> Result<()>`).
- **Both clippy invocations are intentional:** the helpers are `#[cfg(feature = "channel-matrix")]`; the default-build clippy run guards against a dead-code regression when the feature is off.

### RED evidence — defect (1): wrong account on the wire (sanitized)

Two agents, each bound to its own Matrix block with a distinct account, under one daemon. The config is correct; the blocks are separated by alias and credentials:

```toml
[channels.matrix.bot-b]
enabled = true
homeserver = "https://matrix.org"
user_id   = "@bot-b:matrix.org"
password  = "<REDACTED>"
recovery_key = "<REDACTED>"

[channels.matrix.bot-a]
enabled = true
homeserver = "https://matrix.org"
user_id   = "@bot-a:matrix.org"
password  = "<REDACTED>"
recovery_key = "<REDACTED>"
```

Pre-fix, both blocks resolved to the same `<config_dir>/state/matrix/` dir regardless of alias. The single shared session file held one account:

```json
{ "user_id": "@bot-a:matrix.org", "device_id": "<REDACTED>", "access_token": "<REDACTED>" }
```

Runtime trace at start (sync tokens, room/event ids redacted) — both blocks restore the same `session.json` with the identical secret-storage key id:

```
channel_alias=bot-a   matrix: restored session from session.json
channel_alias=bot-b   matrix: restored session from session.json
channel_alias=bot-a   matrix: secret-storage diagnostics: default_key_id=<KEYID-X> ...
channel_alias=bot-b   matrix: secret-storage diagnostics: default_key_id=<KEYID-X> ...
channel_alias=bot-b   matrix: E2EE recovery failed ...      # bot-b recovery_key vs bot-a store
channel_alias=bot-b   matrix: starting sync loop            # but attached to @bot-a
```

`bot-b` is attached to `@bot-a` while stamping inbound `channel_alias=bot-b`, so a DM to `@bot-a` is dispatched to the `bot-b` agent. Dispatch/routing is correct; the crossover is entirely at the shared session store. Regression test `matrix_state_dir_is_distinct_per_alias` asserts the two aliases resolve to distinct paths (RED against the pre-fix shared path).

### RED evidence — defect (2): broken key backup, recovery gave up (sanitized)

After the session split, `bot-b` logs in correctly as itself, then loops on a broken key backup. With the wrong/stale recovery key in config the secret-storage unlock fails outright; even with a valid key, the old `recover()` only imported and never repaired an inconsistent backup, so room-key backup stayed broken:

```
channel_alias=bot-b  WARN   matrix: secret-storage diagnostics: default_key_id=<KEYID>
channel_alias=bot-b  WARN   matrix: E2EE recovery failed: The MAC check for the
                            secret storage key failed
channel_alias=bot-b  WARN   matrix: starting sync loop   (degraded)
(repeating)          WARN   Trying to backup room keys but no backup key was found
```

`recover()` only *imports* secrets; it does not repair an inconsistent server-side key backup, so the `no backup key was found` loop persists. The fix swaps `recover()` for `recover_and_fix_backup(key)` (verified against matrix-sdk 0.17.0): it opens secret storage with the operator's configured `recovery_key`, imports secrets, and recreates **only the key backup** (never the secret-storage key) when the backup is inconsistent. The configured `recovery_key` stays valid; the loop clears. The MAC failure above is a wrong key in config, not a code fault, and surfaces clearly so the operator fixes the config.

### GREEN evidence — runtime-verified on a live multi-agent daemon (sanitized)

Same daemon, after the fix is deployed and the operator's `recovery_key` is correct. `bot-b` restores its **own** session (session isolation working) and `recover_and_fix_backup` succeeds:

```
channel_alias=bot-b  INFO  matrix: restored session from session.json
channel_alias=bot-b  WARN  matrix: secret-storage diagnostics: default_key_id=<KEYID>
channel_alias=bot-b  INFO  matrix: E2EE recovery completed (cross-signing + room keys imported)
channel_alias=bot-b  INFO  matrix: starting sync loop
```

After recovery completes there are **no further** `E2EE recovery failed`, `MAC check`, or `Trying to backup room keys but no backup key was found` events for that alias, and the agent processes turns end to end:

```
channel_alias=bot-b  INFO  LLM call completed
channel_alias=bot-b  INFO  channel_message_outbound
channel_alias=bot-b  INFO  reply delivered
```

Before the fix, `bot-b` was attached to `@bot-a`'s session and a DM to `@bot-a` was answered by the `bot-b` agent. After the fix, `bot-b` runs as itself, its key backup is healthy, and it replies on its own account. (Remaining non-E2EE WARNs in the trace — context-compression notes and a `{model_provider}/{model}` cost-pricing label — are pre-existing and out of scope.)

- New permissions, capabilities, or file system access scope? **No** (same `state/` tree, one extra path component).
- New external network calls? **No new endpoints** — `recover_and_fix_backup` uses the same Matrix client/homeserver already in use; it exercises the secret-storage/key-backup APIs the import path already touched.
- Secrets / tokens / credentials handling changed? **Yes** — each Matrix block reads/writes its own `session.json` under a per-alias dir, strengthening isolation. Recovery opens secret storage with the configured `recovery_key` and never logs the password or access token. The configured recovery key is not rotated or logged.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — all evidence here is sanitized (`@bot-a`/`@bot-b`, redacted ids/tokens/keys). Test fixtures use example `@<name>:matrix.org` MXIDs consistent with the suite's conventions.
- If any `Yes`: no recovery key is generated or logged. Recovery uses the operator's configured `channels.matrix.recovery-key` to open existing secret storage; the key is never printed, rotated, or replaced.

## Compatibility (required)

- Backward compatible? **No** (session-isolation path change).
- Config / env / CLI surface changed? **No** (config unchanged; on-disk state path changed).
- Upgrade steps for existing users: Breaking change with no clean migration. On first start after upgrade, every Matrix block logs in fresh under its configured `user_id` at `<config_dir>/state/matrix/<alias>/` (one-time E2EE re-bootstrap per bot, using each block's `password`/`recovery_key`). The legacy `<config_dir>/state/matrix/session.json` and `store/` are left in place but unused and can be deleted. An account whose configured `recovery_key` matches its secret storage recovers cleanly; an account with a wrong/stale key logs a clear `recovery failed` WARN pointing at `channels.matrix.recovery-key`.

- **Fast rollback command/path:** `git revert <sha>` per commit. Reverting restores the shared-path session behavior and the plain `recover()` (no backup repair) behavior.
- **Observable failure symptoms:** Grep daemon logs for `matrix: restored session from session.json` appearing for two `channel_alias` values with the same `default_key_id` (shared store), `matrix: E2EE recovery failed` on a block whose `recovery_key` is correct for its own account, or a sustained `Trying to backup room keys but no backup key was found` loop (backup never created).

### DIFF
--- a/crates/zeroclaw-channels/src/matrix.rs
+++ b/crates/zeroclaw-channels/src/matrix.rs
@@ -726,6 +726,25 @@ mod client {
         has_password && has_user_id
     }
 
+    /// Decide whether a saved session belongs to a different Matrix account
+    /// than the one this channel block is configured for. Restoring a foreign
+    /// session would run the block as the wrong bot identity. Only a fully
+    /// qualified configured `user_id` (`@local:server`) is compared against the
+    /// saved canonical MXID; a bare localpart or unset `user_id` cannot be
+    /// compared without false positives, so those never flag.
+    pub(super) fn saved_session_is_foreign(
+        config: &MatrixConfig,
+        blob: &session::SessionBlob,
+    ) -> bool {
+        let Some(want) = config.user_id.as_deref().filter(|s| !s.is_empty()) else {
+            return false;
+        };
+        if !want.contains(':') {
+            return false;
+        }
+        want != blob.user_id.as_str()
+    }
+
     async fn build_attempt(
         config: &MatrixConfig,
         state_dir: &Path,
@@ -740,6 +759,25 @@ mod client {
         }
 
         let saved = session::load(state_dir)?;
+
+        // A saved session that belongs to a different account would run this
+        // channel block as the wrong Matrix identity. Wipe and re-login fresh
+        // under the configured account instead of impersonating.
+        if let Some(blob) = saved.as_ref()
+            && saved_session_is_foreign(config, blob)
+        {
+            return recover_or_bail(
+                config,
+                state_dir,
+                recovery_attempts,
+                &format!(
+                    "saved session user_id ({}) does not match configured channels.matrix user_id ({}); store belongs to a different account.",
+                    blob.user_id,
+                    config.user_id.as_deref().unwrap_or_default()
+                ),
+            )
+            .await;
+        }
 
         // The saved device_id is canonical — it's what the server actually
         // assigned at login. config.device_id is only a hint for first-ever
@@ -4594,7 +4632,8 @@ mod tests {
         //! corruption-recovery decisions verifiable without touching the SDK.
 
         use super::super::client::{
-            can_password_relogin, resolve_access_token_identity, store_has_orphan_data,
+            can_password_relogin, resolve_access_token_identity, saved_session_is_foreign,
+            store_has_orphan_data,
         };
         use tempfile::TempDir;
         use wiremock::{
@@ -4649,6 +4688,47 @@ mod tests {
         fn relogin_rejects_empty_strings() {
             assert!(!can_password_relogin(&cfg(Some(""), Some("@bot:m"))));
             assert!(!can_password_relogin(&cfg(Some("pw"), Some(""))));
+        }
+
+        fn blob_for(user_id: &str) -> super::super::session::SessionBlob {
+            super::super::session::SessionBlob {
+                user_id: user_id.to_string(),
+                device_id: "DEV1".to_string(),
+                access_token: "secret".to_string(),
+                refresh_token: None,
+            }
+        }
+
+        #[test]
+        fn foreign_session_detected_when_user_ids_differ() {
+            // The collision bug: two matrix blocks shared one state dir, so the
+            // second to start restored the first account's saved session and
+            // ran as the wrong bot. With the configured user_id known, a saved
+            // session for a different account must be flagged so the build flow
+            // wipes and re-logins instead of impersonating.
+            let cfg = cfg(Some("pw"), Some("@clamps-bot:matrix.org"));
+            let foreign = blob_for("@bender-bending-rodriguez-zeroclaw:matrix.org");
+            assert!(saved_session_is_foreign(&cfg, &foreign));
+        }
+
+        #[test]
+        fn matching_session_not_foreign() {
+            let cfg = cfg(Some("pw"), Some("@clamps-bot:matrix.org"));
+            let own = blob_for("@clamps-bot:matrix.org");
+            assert!(!saved_session_is_foreign(&cfg, &own));
+        }
+
+        #[test]
+        fn unset_or_bare_user_id_never_flags() {
+            // No configured user_id, or a bare localpart that cannot be
+            // compared against the canonical MXID, must not false-positive.
+            let any = blob_for("@whoever:matrix.org");
+            assert!(!saved_session_is_foreign(&cfg(Some("pw"), None), &any));
+            assert!(!saved_session_is_foreign(&cfg(Some("pw"), Some("")), &any));
+            assert!(!saved_session_is_foreign(
+                &cfg(Some("pw"), Some("clamps-bot")),
+                &any
+            ));
         }
 
         #[test]
--- a/crates/zeroclaw-channels/src/matrix.rs
+++ b/crates/zeroclaw-channels/src/matrix.rs
@@ -1254,11 +1254,10 @@ mod client {
     /// length (whitespace-stripped, not the value), and the full error
     /// debug chain (the SDK's `Display` masks fallback errors).
     async fn run_recovery(client: &Client, key: &str) {
+        use matrix_sdk::encryption::recovery::RecoveryState;
+
         let recovery = client.encryption().recovery();
-        if matches!(
-            recovery.state(),
-            matrix_sdk::encryption::recovery::RecoveryState::Enabled
-        ) {
+        if matches!(recovery.state(), RecoveryState::Enabled) {
             ::zeroclaw_log::record!(
                 DEBUG,
                 ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note),
@@ -1269,6 +1268,18 @@ mod client {
 
         let stripped_len = key.chars().filter(|c| !c.is_whitespace()).count();
         diagnose_secret_storage(client, stripped_len).await;
+
+        // No secret storage exists server-side: this account never had Secure
+        // Backup set up. The previous behavior only logged "set it up in
+        // Element first" and left the account with no key backup, spamming
+        // "no backup key was found" forever and risking unrecoverable room
+        // keys if the local store is wiped. Bootstrap it automatically,
+        // seeding secret storage with the operator's configured recovery_key
+        // so that same value unlocks it on future starts.
+        if matches!(recovery.state(), RecoveryState::Disabled) {
+            bootstrap_recovery(&recovery, key).await;
+            return;
+        }
 
         match recovery.recover(key).await {
             Ok(()) => {
@@ -1284,6 +1295,39 @@ mod client {
                     .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
                     .with_attrs(::serde_json::json!({"e": e.to_string()})),
                 "matrix: E2EE recovery failed: ; full error chain = . If the input length above is unexpected (base58 keys are typically ~58 chars, passphrases vary), the wrong value may be in channels.matrix.recovery-key."
+            ),
+        }
+    }
+
+    /// Create Secure Backup + secret storage for an account that has none.
+    /// Seeds secret storage with the operator's configured `recovery_key` as
+    /// the passphrase, so the same configured value unlocks it on later
+    /// starts. The SDK also returns a freshly generated base58 recovery key;
+    /// it is logged so an operator who prefers the generated key over a
+    /// passphrase can persist it into `channels.matrix.recovery-key`.
+    async fn bootstrap_recovery(
+        recovery: &matrix_sdk::encryption::recovery::Recovery,
+        passphrase: &str,
+    ) {
+        let enable = recovery.enable().wait_for_backups_to_upload();
+        let enable = if passphrase.is_empty() {
+            enable
+        } else {
+            enable.with_passphrase(passphrase)
+        };
+        match enable.await {
+            Ok(generated_key) => ::zeroclaw_log::record!(
+                INFO,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_attrs(::serde_json::json!({"generated_recovery_key": generated_key})),
+                "matrix: Secure Backup bootstrapped (created secret storage + key backup). Keep the generated_recovery_key safe; it (or the configured passphrase) unlocks backup on future starts."
+            ),
+            Err(e) => ::zeroclaw_log::record!(
+                WARN,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
+                    .with_attrs(::serde_json::json!({"e": e.to_string()})),
+                "matrix: failed to bootstrap Secure Backup; key backup remains unconfigured for this account"
             ),
         }
     }
--- a/crates/zeroclaw-channels/src/matrix.rs
+++ b/crates/zeroclaw-channels/src/matrix.rs
@@ -1299,16 +1299,23 @@ mod client {
         }
     }
 
-    /// Create Secure Backup + secret storage for an account that has none.
-    /// Seeds secret storage with the operator's configured `recovery_key` as
-    /// the passphrase, so the same configured value unlocks it on later
-    /// starts. The SDK also returns a freshly generated base58 recovery key;
-    /// it is logged so an operator who prefers the generated key over a
-    /// passphrase can persist it into `channels.matrix.recovery-key`.
+    /// Create or repair Secure Backup for an account whose secret storage is
+    /// absent (`RecoveryState::Disabled`). Seeds secret storage with the
+    /// operator's configured `recovery_key` as the passphrase so the same
+    /// configured value unlocks it on later starts. The SDK returns a freshly
+    /// generated base58 recovery key, logged so an operator who prefers it can
+    /// persist it into `channels.matrix.recovery-key`.
+    ///
+    /// Two shapes of `Disabled` are handled: no backup at all (`enable()`
+    /// creates one), and a stale server-side backup version with no secret
+    /// storage default key (`enable()` returns `BackupExistsOnServer`, so we
+    /// `reset_key()` to rotate fresh secret storage and re-link the backup).
     async fn bootstrap_recovery(
         recovery: &matrix_sdk::encryption::recovery::Recovery,
         passphrase: &str,
     ) {
+        use matrix_sdk::encryption::recovery::RecoveryError;
+
         let enable = recovery.enable().wait_for_backups_to_upload();
         let enable = if passphrase.is_empty() {
             enable
@@ -1316,12 +1323,42 @@ mod client {
             enable.with_passphrase(passphrase)
         };
         match enable.await {
-            Ok(generated_key) => ::zeroclaw_log::record!(
-                INFO,
-                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
-                    .with_attrs(::serde_json::json!({"generated_recovery_key": generated_key})),
-                "matrix: Secure Backup bootstrapped (created secret storage + key backup). Keep the generated_recovery_key safe; it (or the configured passphrase) unlocks backup on future starts."
-            ),
+            Ok(generated_key) => {
+                ::zeroclaw_log::record!(
+                    INFO,
+                    ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                        .with_attrs(::serde_json::json!({"generated_recovery_key": generated_key})),
+                    "matrix: Secure Backup bootstrapped (created secret storage + key backup). Keep the generated_recovery_key safe; it (or the configured passphrase) unlocks backup on future starts."
+                );
+            }
+            // A backup version exists server-side but secret storage is absent
+            // (no m.secret_storage.default_key), so enable() refuses. Rotate
+            // fresh secret storage and re-link the existing backup instead.
+            Err(RecoveryError::BackupExistsOnServer) => {
+                let reset = recovery.reset_key();
+                let reset = if passphrase.is_empty() {
+                    reset
+                } else {
+                    reset.with_passphrase(passphrase)
+                };
+                match reset.await {
+                    Ok(generated_key) => ::zeroclaw_log::record!(
+                        INFO,
+                        ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                            .with_attrs(
+                                ::serde_json::json!({"generated_recovery_key": generated_key})
+                            ),
+                        "matrix: Secure Backup repaired (a server-side backup existed without secret storage; rotated fresh secret storage and re-linked). Keep the generated_recovery_key safe."
+                    ),
+                    Err(e) => ::zeroclaw_log::record!(
+                        WARN,
+                        ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                            .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
+                            .with_attrs(::serde_json::json!({"e": e.to_string()})),
+                        "matrix: failed to repair Secure Backup via reset_key; key backup remains unconfigured for this account"
+                    ),
+                }
+            }
             Err(e) => ::zeroclaw_log::record!(
                 WARN,
                 ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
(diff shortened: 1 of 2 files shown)


WareWolf-MoonWall (approved):
# Review Draft — PR #7388
**fix(matrix): isolate session state per channel alias and repair key backup with the configured key**
Reviewer: WareWolf-MoonWall | Verdict: `--approve`

---

I reviewed all four commits across `crates/zeroclaw-channels/src/matrix.rs` (+~120/-~25) and
`crates/zeroclaw-channels/src/orchestrator/mod.rs` (+~45/-~10), read the local source for both
files to verify the diff is coherent with surrounding code, and studied the sanitized live-daemon
traces in the PR body. This PR fixes two confirmed Matrix multi-agent defects: per-alias session
isolation (wrong-account-on-the-wire) and repair of broken E2EE key backup via
`recover_and_fix_backup`.

---

### 🟢 What looks good — `matrix_state_dir` establishes a single source of truth

Before this PR, the `state/matrix/` path computation was written out inline at two separate call
sites in `orchestrator/mod.rs`. Both were identical and both lacked the alias component — the
root cause of the session-clobber bug. The fix introduces a single `matrix_state_dir(config_path,
alias)` helper and calls it at both construction sites. This is the correct AGENTS.md DRY
pattern: one canonical computation, resolved on demand, no cached copies. The regression test
`matrix_state_dir_is_distinct_per_alias` gives a deterministic RED-proven anchor for the fix.

### 🟢 What looks good — `saved_session_is_foreign` is defense-in-depth done right

The guard is conservatively narrow: it returns `false` for a bare localpart or an unset
`user_id` that cannot be safely compared against the canonical MXID, preventing false-positive
wipes. All three boundary cases (match, mismatch, bare localpart) are covered by dedicated unit
tests. The comment above the function explains the invariant in exactly the terms an on-call
engineer needs.

### 🟢 What looks good — `recover_and_fix_backup` is the correct API and the comment proves it

The existing `recover(key)` call only imports secrets; it cannot repair an inconsistent
server-side key backup, which is why the "no backup key was found" loop was unbreakable. The
code comment explains all three properties of `recover_and_fix_backup`: it opens secret storage
with the operator-configured key, imports secrets, and recreates only the key backup when
inconsistent — without rotating the recovery key. This is verified against the pinned
`matrix-sdk` 0.17.0 API and confirmed live via the GREEN daemon trace.

### 🟢 What looks good — dual clippy runs and clean test output

Running `cargo clippy ... --features channel-matrix` and again without the feature is exactly the
right discipline: the helpers live under `#[cfg(feature = "channel-matrix")]` and an unguarded
dead-code regression in the off-feature build is a real risk. The PR documents why both runs are
intentional. Test output shows 834 existing tests passing with 0 failures.

### 🟢 What looks good — breaking change is honestly documented

The PR marks `Backward compatible: No` and gives explicit upgrade steps (one-time re-login per
bot, legacy path left in place, how to identify a wrong/stale `recovery_key`). This is the right
level of operator guidance for a forced-migration bug fix.

### 🟡 Warning — stale operator-facing path in the hard-recursion-bound `bail!`

`build_attempt()` in `matrix.rs` (currently L737-738 in local source) still reads:

```
"matrix: corruption recovery looped — aborting to avoid an infinite restart cycle. \
 Wipe ~/.zeroclaw/state/matrix/ manually and restart."
```

After this PR, every alias lives under `state/matrix/<alias>/`, so an operator following this
message would wipe the wrong directory and the session would still be broken. The `state_dir`
parameter is already in scope at the call site; the message should reference it directly (e.g.
`state_dir.display()`) rather than hardcoding the legacy path.

*Should be posted inline on `crates/zeroclaw-channels/src/matrix.rs` at the `bail!` message in
`build_attempt` (currently L737-738 in the pre-PR local source).*

### 🟡 Warning — no tracking issue for a `v0.8.0`-milestoned user-facing regression

The PR is milestoned v0.8.0 and describes a confirmed real-user report ("wrong bot answered my
DM"). The body notes "No tracking issue filed yet." For a bug with this impact level —
multi-account identity crossover, operator-visible token/device exposure — a tracking issue
makes it easier to reference in the v0.8.0 changelog, triage duplicates, and link from incident
post-mortems. Filing one before or immediately after merge would be worth the two minutes.

---

**Verdict:** Approving. Both root-cause fixes are correct, well-evidenced against a live daemon,
comply with AGENTS.md DRY contract, and are covered by regression tests. The stale path message
is a genuine usability gap I'd like to see patched before a stable release but it does not
affect the correctness of the session-isolation or key-backup repair logic and does not block
merge here.

Audacity88 (approved):
I reviewed the live PR at e154697c3bfd1c3c9d30262e59d04b357ac60195, the two-file diff, the sanitized live Matrix evidence in the PR body, current CI, WareWolf-MoonWall's approval review, and the current `origin/master` overlap in `crates/zeroclaw-channels/src/orchestrator/mod.rs`. There are no inline threads and no active changes-requested review. This approval is for the reviewed head; because `master` has moved through the same channel/orchestrator area, I would still refresh the branch or test the exact merge result before merging.

### 🟢 What looks good — per-alias Matrix state fixes the identity boundary

The `matrix_state_dir(config_path, alias)` helper is the right shape for this bug. Both Matrix construction paths now resolve through one helper, and the alias becomes part of the store path, so two `[channels.matrix.<alias>]` blocks no longer share one `session.json` and crypto store. The `matrix_state_dir_is_distinct_per_alias` regression test captures the old failure mode directly.

### 🟢 What looks good — the foreign-session guard fails toward the safer account

`saved_session_is_foreign()` is a useful defense-in-depth layer on top of the path split. If a store still contains a fully qualified MXID for another account, the channel wipes and re-authenticates instead of continuing under the wrong Matrix identity. The bare-localpart and unset-`user_id` cases are also handled conservatively, which avoids false-positive wipes when the configured value cannot be compared against the canonical saved MXID.

### 🟢 What looks good — `recover_and_fix_backup` matches the observed E2EE failure

The switch from `recover(key)` to `recover_and_fix_backup(key)` matches the PR's live evidence: importing secrets alone did not repair the inconsistent server-side key backup, while the new API uses the configured recovery key and repairs only the backup. The PR body is also honest about the compatibility cost: existing Matrix installs move to per-alias state and may need a one-time re-login/recovery pass.

### 🟡 Warning — clean up the remaining stale recovery wording

WareWolf-MoonWall's warning about the hard-recursion `bail!` message is right. That path still says to wipe `~/.zeroclaw/state/matrix/`, which is stale after this PR because each Matrix alias now owns `state/matrix/<alias>/`. This does not change my verdict, but it should be cleaned up before the release.

The PR is currently behind `origin/master`, and `master` has changed `crates/zeroclaw-channels/src/orchestrator/mod.rs` since this branch point. I do not see that invalidating this fix, and a local merge-tree check is clean, but the final merge should still wait for an updated branch or fresh CI/local smoke on the exact merge result. The current green CI proves the reviewed head, not the final merge result.

tidux (approved):
Reviewed PR head e154697c3bfd1c3c9d30262e59d04b357ac60195 on `fix/matrix/multi-agent-clobber`. I read the full two-file diff, the local `crates/zeroclaw-channels/src/matrix.rs` around the foreign-session early-return path and the `build_attempt` recursion bound, the local `crates/zeroclaw-channels/src/orchestrator/mod.rs` around both Matrix construction sites and the new helper, `crates/zeroclaw-channels/Cargo.toml` to confirm the pinned `matrix-sdk = "0.17"`, the linked tracker (#6487, `priority:p0 (blocker)`, `risk: high`, OPEN), and the sanitized RED/GREEN daemon traces in the PR body against the actual code paths. I read @WareWolf-MoonWall's `--approve` at e154697c and @Audacity88's `--approve` at e154697c; no inline threads, no active `CHANGES_REQUESTED`.

Approving. Both root causes are correctly identified — a shared on-disk state dir across `[channels.matrix.<alias>]` blocks producing identity crossover, and `recovery.recover(key)` being the wrong API for a server-side key backup that is inconsistent rather than missing. The fixes are minimal, well-scoped, and proved against a live multi-agent daemon. CI is fully green on the reviewed head and `mergeable_state: clean`.

### ✅ Resolved — "no tracking issue filed yet" is no longer accurate

@WareWolf-MoonWall flagged that this PR has no tracking issue, citing a body line of "No tracking issue filed yet." The current PR body does link a tracking issue: `**Linked issue(s):** Closes #6487`. Issue #6487 is the canonical tracker — title is `[Bug]: Multi-alias Matrix instances share one session store and clobber each other (wrong account on the wire); inconsistent key backup never repaired`, labeled `priority:p0 (blocker)`, `risk: high`, `status:accepted`, currently `OPEN`. On merge the `Closes #` linkage will auto-close it. No additional tracking work is required for v0.8.0 changelog purposes — referencing #6487 is enough.

### 🟢 What looks good — `matrix_state_dir` collapses the duplicated path computation into one canonical helper

The pre-fix bug had two literal copies of the same `config_path.parent().map(|p| p.join("state").join("matrix"))` chain at the two construction sites in `orchestrator/mod.rs` (the `default`-fallback at the old L5510-L5514 and the per-alias loop at the old L6473-L6477). Both omitted the alias component. This PR replaces both with `matrix_state_dir(&config.config_path, alias)`. That is the AGENTS.md DRY contract in textbook form: one canonical resolver, called at every use site, no cached copies anywhere. If a third Matrix construction site is added later, the alias inclusion is no longer a thing the new caller has to remember — the helper signature requires it. The regression test `matrix_state_dir_is_distinct_per_alias` pins the invariant against the literal path shape.

### 🟢 What looks good — `saved_session_is_foreign` is layered defense, not a substitute for the path split

The foreign-session early-return in `build_attempt` (matrix.rs L760-L775 patched) is correct in two ways that matter for the security boundary. First, it sits *under* the path-split fix — even if a future refactor accidentally re-shares a directory across aliases, an account-level mismatch in the saved blob will still wipe-and-relogin rather than impersonate. Second, the guard's negative cases are conservative: a bare localpart configured `user_id`, an empty string, or `None` all return `false` (no wipe), because none of those can be safely compared against the canonical MXID in the saved blob. The three unit tests cover the match, the mismatch, and all three no-compare cases. The fallback uses `recover_or_bail` rather than a raw wipe, which preserves the existing recursion bound.

### 🟢 What looks good — `recover_and_fix_backup` is the right API and the comment explains the invariant that matters

The doc comment above the `match recovery.recover_and_fix_backup(key).await` call (matrix.rs L1272-L1278 patched) names the three properties that distinguish the new call from the old one: it opens secret storage with the operator-configured `recovery_key`, it imports the same secrets `recover()` did, and it recreates only the **key backup** (never the secret-storage key itself) when the server-side backup is inconsistent. That last clause is the load-bearing one — it's what makes this fix safe for an operator whose `recovery_key` config matches their secret storage. Cross-checked against the pinned `matrix-sdk = "0.17"` in `crates/zeroclaw-channels/Cargo.toml`; the GREEN daemon trace (`E2EE recovery completed`, no further `no backup key was found`) confirms the behavior end-to-end.

### 🟢 What looks good — the breaking-change disclosure is honest and operator-actionable

The header banner explicitly marks this as a breaking change, gives the manual `mv session.json store default/` migration for single-block installs, calls out that multi-block installs have exactly **one** legacy `session.json` belonging to whichever account logged in last, and tells operators which legacy paths can be deleted after the migration. The Compatibility section restates the same constraints in template form. There is no attempt to hide the cost of the fix behind a "minor change" framing.

### 🟡 Warning (concurring with @WareWolf-MoonWall and @Audacity88) — stale path in the hard-recursion `bail!` should be fixed before tagging v0.8.0

`crates/zeroclaw-channels/src/matrix.rs:737-738` (unchanged by this PR but exposed by it) still tells an operator who hits the recursion bound to "Wipe ~/.zeroclaw/state/matrix/ manually and restart." After this PR every alias owns `<config>/state/matrix/<alias>/`, so an operator following that message either wipes too much (every alias's session) or wipes the wrong thing. The `state_dir: &Path` parameter is already in scope at the `bail!` (matrix.rs L731), so a one-line fix is `state_dir.display()` in the message. Not a merge blocker — both prior reviewers approved with this as a 🟡 — but it would be cheap to land as a fixup commit on this branch rather than carry it as a separate PR into v0.8.0.

### 🟡 Warning (concurring with @Audacity88) — refresh-before-merge on the orchestrator overlap

@Audacity88's read of `origin/master` having moved through `crates/zeroclaw-channels/src/orchestrator/mod.rs` since this branch point is correct; my own check shows the file has had unrelated edits in the same window. GitHub currently reports `mergeable_state: clean`, so the merge result will be clean rather than dirty, but the green CI on this head proves the reviewed head only — not the result of the merge. Worth a fresh `cargo test -p zeroclaw-channels --features channel-matrix` on the merge result before pressing the button, or a rebase + CI rerun if anything in master touches the new helper's call sites.

---

**Verdict:** `--approve`.

This is the right shape for the bug: a forced on-disk migration is the only way to deliver per-alias session isolation without inventing a migration-tool surface, and the recovery API swap is verified live. The two 🟡s above are both real and both safe to fix in this branch (one-line `bail!` message, one rebase) — neither blocks the verdict. With the stale path message tidied, this is in good shape to ship as part of the v0.8.0 Stable-tier hardening.

Issue #5903, opened 2026-04-19
[Bug]: MCP stdio child processes accumulate on daemon with heartbeat.enabled=true (one orphan per tick)


rordd:
## Summary

Running `zeroclaw daemon` with `heartbeat.enabled = true` (the default) and any stdio MCP server configured causes **one stdio child process to leak per heartbeat tick**. The children accumulate for the entire daemon lifetime. Default `interval_minutes = 30` → roughly **48 orphan MCP processes per daemon per day**. Higher heartbeat frequency accelerates proportionally.

## Environment

- `zeroclaw 0.7.3` (master commit `6b63340`), clean release build
- macOS 14 (Darwin 25.2.0), Rust 1.87
- Default config generated via `zeroclaw onboard --quick`, edited to enable MCP and a trivial stdio server
- Clean `\$HOME` (no prior state), gateway port moved to 42618 for isolation

## Reproducer

Minimal stdio MCP server in ~30 lines of Python — replies to `initialize`/`tools/list`/`tools/call` and exits cleanly on stdin EOF:

\`\`\`python
#!/usr/bin/env python3
import json, os, sys
for line in sys.stdin:
    line = line.strip()
    if not line: continue
    req = json.loads(line)
    rid, method = req.get('id'), req.get('method', '')
    if method == 'initialize':
        resp = {'jsonrpc':'2.0','id':rid,'result':{'protocolVersion':'2024-11-05','capabilities':{'tools':{}},'serverInfo':{'name':'echo-mcp','version':'0.1.0'}}}
    elif method == 'notifications/initialized':
        continue
    elif method == 'tools/list':
        resp = {'jsonrpc':'2.0','id':rid,'result':{'tools':[{'name':'echo','description':'echo','inputSchema':{'type':'object'}}]}}
    elif method == 'tools/call':
        resp = {'jsonrpc':'2.0','id':rid,'result':{'content':[{'type':'text','text':'ok'}]}}
    else:
        resp = {'jsonrpc':'2.0','id':rid,'error':{'code':-32601,'message':'nope'}}
    print(json.dumps(resp), flush=True)
print(f'[echo-mcp pid={os.getpid()}] EOF, exiting', file=sys.stderr)
\`\`\`

config.toml additions:

\`\`\`toml
[heartbeat]
enabled = true
interval_minutes = 1
two_phase = true
adaptive = false
min_interval_minutes = 1

[mcp]
enabled = true

[[mcp.servers]]
[...]

PR #7392, opened 2026-06-08
test(tools): cover MCP stdio registry drop cleanup

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds a Unix-only regression test for stdio MCP server cleanup when an `McpRegistry` is dropped.
  - Uses a tiny shell-backed MCP server that stays alive while the registry is alive, then verifies the child process is gone after drop.
  - Locks in the current lifecycle behavior so short-lived MCP registries do not leave stdio child processes behind.
- **Scope boundary:** Test coverage only. This PR does not change MCP transport behavior, heartbeat scheduling, gateway MCP behavior, or process-spawn configuration.
- **Blast radius:** Limited to the `zeroclaw-tools` test suite. Runtime behavior is unchanged; the high-risk label comes from the MCP/tools surface, not from a production-code change.
- **Linked issue(s):** Related #5903
- **Labels:** `risk: high`, `size: S`, `tool`, `tool:mcp`

  - `cargo fmt --package zeroclaw-tools --check` — passed.
  - `scripts/zc-cargo test -p zeroclaw-tools dropping_stdio_registry_reaps_child_process -- --nocapture` — passed: `1 passed; 0 failed; 0 ignored; 1303 filtered out`.
  - `git diff --check` — passed.
- **Beyond CI — what did you manually verify?** The regression confirms the stdio child is alive while the registry is alive and exits after the registry is dropped.
- **If any command was intentionally skipped, why:** Full workspace clippy and CI-shaped nextest were not run locally because this is a test-only PR with one focused `zeroclaw-tools` regression. GitHub Actions passed the broader Format, Lint, Test, Security, and Build gates on this head.

- New permissions, capabilities, or file system access scope? No runtime permissions or capabilities changed; the new Unix-only test uses a temporary shell-backed MCP server locally.
- If any Yes, describe the risk and mitigation: None; all answers are No.

## Compatibility (required)

- Backward compatible? Yes.

- **Fast rollback command/path:** Revert this PR, or revert the merge commit, to remove the regression test in `crates/zeroclaw-tools/src/mcp_client.rs`.
- **Observable failure symptoms:** This PR has no runtime behavior change. A regression would appear as a `zeroclaw-tools` test failure in `dropping_stdio_registry_reaps_child_process`.

- `Co-authored-by` trailers added in commit messages for incorporated contributors? No.
- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded contributor work is carried forward.


perlowja:
Heads-up from some #5903 digging — this test looks correct and worth keeping, but I don't think it will close #5903 on its own, and wanted to flag why.

I independently reproduced the registry-drop behavior: connect a stdio MCP server, drop the `McpRegistry`, and the child is reaped (the `kill_on_drop(true)` on the spawned `Child` does its job). I also ran it as an N-cycle loop (connect → drop ×N) and saw **no** child accumulation. So the drop path you're locking in here is sound. ✅

But the #5903 leak appears to be **retention, not a Drop failure**. On the daemon heartbeat path each tick calls `crate::agent::run(...)`, which builds an `McpRegistry` (`connect_all`, `loop_.rs` ~3217/4579) and wraps it in an `Arc` for the tool registry. The "one orphan per tick" symptom means a clone of that per-tick registry `Arc` is being **retained across ticks**, so `Drop` never runs — which is exactly why children pile up even though drop-reaping works correctly.

Net: #7392 is good regression coverage to keep; #5903 needs a separate fix for the retained `Arc` on the `agent::run`/heartbeat path. Still tracing the exact owner — will share once pinned.

perlowja:
Update / correction to my note above. I ran an instrumented-daemon reproduction on current `master` (0.8.0-beta-2): a real `zeroclaw daemon` with `heartbeat.enabled=true`, a 1-minute interval, and a stdio MCP server, with a temporary `Drop` probe on `McpRegistry`.

Result: the per-tick heartbeat MCP registry **is** dropped every tick — the probe fires ~2×/tick (phase-1 decision + phase-2 execution each build and drop a registry), and the stdio child count stays flat across 3+ ticks (no accumulation; `kill_on_drop` reaps each per-tick child). So my earlier "retention prevents the drop" theory was wrong — the repro refutes it.

Net: #5903 appears **already fixed on master** (it was filed against 0.7.3). This test is exactly the right regression guard to lock the drop-reaping behavior in, so it's well worth merging. Sorry for the noise on the retention angle.

Audacity88:
Thanks for the follow-up and correction, @perlowja. That matches the intended scope here: this PR is a regression guard for the stdio registry drop/reap behavior, not a claim that this branch is the original behavioral fix for #5903.

Keeping the link as `Related #5903` feels right to me. If current `master` no longer reproduces the heartbeat accumulation, then the issue-side question is probably whether #5903 can close once this guard lands, or whether it needs one more note documenting the current-master repro result.

singlerider (approved):
Read the test. Well-constructed process-lifecycle regression with the flakiness traps handled. Approving.

### 🟢 What looks good — the test is race-tolerant where process tests usually aren't

Process-lifecycle tests are the classic source of CI flake, and this one avoids every common pitfall. The PID is obtained by polling the server's pid-file for up to 1s (50×20ms) rather than a fixed sleep, so it doesn't race the spawn. Liveness is checked with `kill -0 <pid>` — the canonical "does this PID exist" probe with no side effect — and stdout/stderr nulled. The post-drop assertion polls for up to 1s and returns the instant the child is gone, only panicking after the full window expires, so it tolerates cleanup latency without padding the happy path with a worst-case sleep. `#[cfg(unix)]` is correct since it leans on `kill` and a shell script.

### 🟢 What looks good — the harness exercises the exact #5903 lifecycle, and the PID stays valid across exec

The shell MCP server writes its own `$$` to the pid-file, answers `initialize` and `tools/list`, then `exec tail -f /dev/null` to stay alive — which is the right construction for the thing under test: a stdio child that lives as long as the registry holds it. The `exec` is the subtle correctness detail that makes this sound: it replaces the shell image with `tail` *in place*, preserving the PID, so `child_pid` read before the exec still tracks the live process the registry owns. The test then proves both halves of the contract — child alive while the `McpRegistry` is alive, child reaped after `drop(registry)` — which is exactly the accumulation behavior #5903 is about. Test-only, no runtime change; the `risk: high` label reflects the MCP surface, not a code change. Approving.

PR #5372, opened 2026-04-06
fix(gateway): truncate oversized memory api payloads

forsakenyang:
## Summary

Describe this PR in 2-5 bullets:

- Base branch target (`master` for all contributions): `master`
- Problem: `GET /api/memory` returns full memory content for list/search responses, so a few oversized autosaved rows can make the dashboard memory page stall or hang.
- Why it matters: the WebUI requests `/api/memory` on page load, so oversized payloads block the memory view and make incident recovery depend on manual DB cleanup.
- What changed: clamp memory entry `content` in the existing API response to 4096 chars and add a regression test covering the truncation behavior.
- What did **not** change (scope boundary): no new memory API endpoint, no schema redesign, no autosave policy change.

## Label Snapshot (required)

- Risk label (`risk: low|medium|high`): `risk: low`
- Size label (`size: XS|S|M|L|XL`, auto-managed/read-only): `size: XS` (expected)
- Scope labels (`core|agent|channel|config|cron|daemon|doctor|gateway|health|heartbeat|integration|memory|observability|onboard|provider|runtime|security|service|skillforge|skills|tool|tunnel|docs|dependencies|ci|tests|scripts|dev`, comma-separated): `gateway, memory, tests`
- Module labels (`<module>: <component>`, for example `channel: telegram`, `provider: kimi`, `tool: shell`): `gateway: memory api`
- Contributor tier label (`trusted contributor|experienced contributor|principal contributor|distinguished contributor`, auto-managed/read-only; author merged PRs >=5/10/20/50): auto-managed
- If any auto-label is incorrect, note requested correction: none

## Change Metadata

- Change type (`bug|feature|refactor|docs|security|chore`): `bug`
- Primary scope (`runtime|provider|channel|memory|security|ci|docs|multi`): `memory`

## Linked Issue

- Closes #
- Related #
- Depends on # (if stacked)
- Supersedes # (if replacing older PR)

## Supersede Attribution (required when `Supersedes #` is used)

- Superseded PRs + authors (`#<pr> by @<author>`, one per line): N/A
- Integrated scope by source PR (what was materially carried forward): N/A
- `Co-authored-by` trailers added for materially incorporated contributors? (`Yes/No`): N/A
- If `No`, explain why (for example: inspiration-only, no direct code/design carry-over): N/A
- Trailer format check (separate lines, no escaped `\n`): (`Pass/Fail`): N/A

## Validation Evidence (required)

Commands and result summary:

```bash
[...]

PR #7393, opened 2026-06-08
fix(gateway): cap memory api response content

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `/api/memory` list and recall/search responses now cap each serialized memory entry `content` field at 4096 characters.
  - The cap is applied only at the gateway response boundary, so stored memory content and backend `Memory::list` / `Memory::recall` behavior stay unchanged.
  - The 4096-character cap includes the trailing `...`, matching the public constant name and resolving the old review blocker on #5372.
  - Regression coverage checks the list path, the recall/search path after category filtering, UTF-8-safe multibyte truncation, exact-boundary behavior, and short content passthrough.
- **Scope boundary:** This does not add a new memory endpoint, change memory storage, change autosave policy, change dashboard rendering, or alter backend recall/list semantics.
- **Blast radius:** Gateway `/api/memory` consumers that previously depended on full `content` in list/search responses now receive a bounded preview. Full content remains in storage.
- **Linked issue(s):** Supersedes #5372.
- **Labels:** `bug`, `risk: high`, `size: S`, `gateway`, `memory`

```bash
$ git diff --check
# clean, exit 0

$ rustfmt --edition 2024 --check crates/zeroclaw-gateway/src/api.rs
# clean, exit 0

$ cargo test -p zeroclaw-gateway handle_api_memory_ --lib
running 2 tests
test api::tests::handle_api_memory_list_truncates_oversized_content ... ok
test api::tests::handle_api_memory_search_truncates_oversized_content_after_filtering ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 227 filtered out; finished in 0.01s

$ cargo test -p zeroclaw-gateway memory_api --lib
running 1 test
test api::tests::truncate_memory_api_content_caps_total_chars_with_ellipsis ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 228 filtered out; finished in 0.00s
```

- **Beyond CI — what did you manually verify?** Checked the PR handler path shown in this diff: `/api/memory` recall/search resolves the memory handle, performs category filtering, then serializes entries; list mode serializes entries directly from `Memory::list`. The new sanitizer runs after those trait calls, so backend storage and recall/list semantics remain unchanged.
- **If any command was intentionally skipped, why:** Full workspace clippy and CI-shaped full tests were not run locally. This is a one-file gateway response-boundary rebuild with focused behavior tests for the exact cap/list/search behavior, and GitHub CI has since passed Lint, Test, platform builds, feature checks, Security, and CI Required Gate on this head. This PR is relying on that public CI signal rather than duplicating the broad run locally. The focused cargo runs emitted a pre-existing unrelated `zeroclaw-channels` unused-import warning, but the gateway tests passed.

## Compatibility (required)

- Backward compatible? `No` for consumers that relied on full memory `content` in `/api/memory` list/search responses; existing requests and stored data remain valid.
- If `No` or `Yes` to either: No direct in-API full-content migration path is introduced by this PR. Existing clients that require full memory content from list/search responses should remain on a pre-#7393 build until a dedicated full-entry read/export path exists, or add/use that path in a follow-up before relying on full-content export. Clients that only display memory previews do not need changes.

- **Fast rollback command/path:** `git revert a43840288`
- **Observable failure symptoms:** Dashboard or API clients may show truncated memory content where they previously displayed the full stored text. The response still ends truncated entries with `...`.

  - #5372 by @forsakenyang
- Scope materially carried forward: the `/api/memory` response should cap oversized memory content so dashboard list/search payloads stay bounded; this rebuild keeps the exact-cap review fix from #5372 and adapts it to the current gateway handler, agent-aware memory resolution, and category-filtering layout.

### DIFF
--- a/crates/zeroclaw-gateway/src/api.rs
+++ b/crates/zeroclaw-gateway/src/api.rs
@@ -10,6 +10,9 @@ use axum::{
 };
 use serde::{Deserialize, Serialize};
 use zeroclaw_config::schema::{ChannelAliasInfo, Config};
+use zeroclaw_memory::MemoryEntry;
+
+const MEMORY_API_CONTENT_MAX_CHARS: usize = 4096;
 
 // ── Bearer token auth extractor ─────────────────────────────────
 
@@ -952,7 +955,10 @@ pub async fn handle_api_memory_list(
                         .collect(),
                     None => entries,
                 };
-                Json(serde_json::json!({"entries": entries})).into_response()
+                Json(serde_json::json!({
+                    "entries": sanitize_memory_entries_for_api(entries)
+                }))
+                .into_response()
             }
             Err(e) => (
                 StatusCode::INTERNAL_SERVER_ERROR,
@@ -970,7 +976,10 @@ pub async fn handle_api_memory_list(
         });
 
         match mem.list(category.as_ref(), None).await {
-            Ok(entries) => Json(serde_json::json!({"entries": entries})).into_response(),
+            Ok(entries) => Json(serde_json::json!({
+                "entries": sanitize_memory_entries_for_api(entries)
+            }))
+            .into_response(),
             Err(e) => (
                 StatusCode::INTERNAL_SERVER_ERROR,
                 Json(serde_json::json!({"error": format!("Memory list failed: {e}")})),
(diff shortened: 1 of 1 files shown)


singlerider (approved):
Read the diff and walked the truncation boundary logic. Correct, UTF-8-safe, and it resolves the exact #5372 review blocker. Approving.

### 🟢 What looks good — the cap is total-inclusive, which is the #5372 fix

The blocker on #5372 was that the old truncation capped content to MAX and *then* appended `...`, producing MAX+3 chars and contradicting the constant's name. This version computes `keep_chars = MAX - 3`, truncates there, and appends `...`, so the result is exactly `MEMORY_API_CONTENT_MAX_CHARS` total — and `truncate_memory_api_content_caps_total_chars_with_ellipsis` pins `truncated.chars().count() == MEMORY_API_CONTENT_MAX_CHARS`. The name now matches the behavior.

### 🟢 What looks good — UTF-8 safe and the exact-boundary case is right

Truncation goes through `char_indices().nth(keep_chars)` to find a real char boundary and `s.truncate(cut_idx)`, so it can never split a multibyte codepoint — verified with the 3-byte `火` test asserting the char count, not the byte length. The early-return guard `s.char_indices().nth(MAX).is_none()` is the correct exact-boundary check: a string of exactly MAX chars has no char at 0-indexed position MAX, so it returns unchanged (no spurious `...`), while MAX+1 chars trips the truncation. The exact-length and short-content passthrough cases are both tested.

### 🟢 What looks good — applied only at the response boundary, after filtering

`sanitize_memory_entries_for_api` runs in the `/api/memory` handler *after* `Memory::list`/`recall` and category filtering, mapping over the already-resolved entries just before serialization. So stored content and backend list/recall semantics are untouched — the cap is purely a response-shaping concern, which is the right layer for it (bounding a payload so one oversized memory entry can't bloat a dashboard list/search response). Both the list path and the recall/search-after-filter path are covered by handler-level tests, not just the helper. The breaking change is flagged honestly (`Backward compatible: No` for clients that consumed full content from list/search; full content remains in storage), with the migration note pointing at a future dedicated full-entry read path. Supersede attribution for #5372 by @forsakenyang is present with the co-author trailer. Approving.

PR #7399, opened 2026-06-08
fix(skills): sanitize skill tool names to satisfy provider name regex (#6678)

### LINKED ISSUE #6678: [Bug]: Skill tools rejected by Anthropic API — `format!("{}.{}", ...)` produces names violating `^[a-zA-Z0-9_-]{1,128}$`
## Symptom

Every agent-loop request to the Anthropic API fails with a 400 Bad Request as soon as a custom skill (with `kind = "shell"` or `kind = "script"` tools) is loaded:

```
Anthropic API error (400 Bad Request): {
  "type":"error",
  "error":{
    "type":"invalid_request_error",
    "message":"tools.38.custom.name: String should match pattern '^[a-zA-Z0-9_-]{1,128}$'"
  },
  "request_id":"req_011Cb47WmzhLNMMVgB2XF6Dv"
}
```

The classification call (smaller tool set, no skill tools) succeeds. The full agent-loop call always fails because at least one skill tool is in the payload.

## Root cause

In `crates/zeroclaw-runtime/src/tools/skill_tool.rs:40`:

```rust
tool_name: format!("{}.{}", skill_name, tool.name),
```

Skill tools are namespaced with a literal `.` between skill name and tool name (e.g. `my_skill.run_lint`, `hetzner_monkey.vpn_check`). The dot violates Anthropic's tool-name regex `^[a-zA-Z0-9_-]{1,128}$`, so every request including a skill tool is rejected.

This used to work — Anthropic apparently tightened tool-name validation server-side recently (other vendors like Bedrock have always enforced this; OpenAI also rejects dots, see `is_valid_openai_tool_name("node.js")` in `openrouter.rs:1747`).

## Reproduction

1. ZeroClaw 0.7.5
2. Provider: `anthropic` (`model = "claude-sonnet-4-6"`, default API endpoint)
3. Any `SKILL.toml` with a `[[tools]]` entry of `kind = "shell"` or `kind = "script"`
4. Trigger any agent-loop turn → 400 from Anthropic

I captured the outgoing payload by pointing `providers.models.anthropic.base_url` at a local dump proxy. Confirmed: tools at indices 38–69 in our case all start with `hetzner_monkey.…`. Indices 0–37 (built-ins) all use snake_case and pass.

## Suggested fix

Match what `mcp_client.rs:230` already does for MCP tools:

```rust
let prefixed = format!("{}__{}", config.name, tool.name);
```

Use `__` (or `_`) instead of `.`:

```rust
// crates/zeroclaw-runtime/src/tools/skill_tool.rs:40
tool_name: format!("{}__{}", skill_name, tool.name),
```

Both are within the Anthropic / OpenAI / Bedrock regex.

This needs to be matched in any place that round-trips the tool name — the dispatcher, the prompt-rendering in `crates/zeroclaw-runtime/src/skills/mod.rs:1041` (which currently emits `<name>{}.{}</name>` into the system prompt as guidance), and any skill→tool reverse lookup.

## Impact

Every ZeroClaw setup that combines:
- Anthropic provider (direct API)
- One or more custom shell/script skills

is currently broken at the agent-loop level. Workaround is to disable the skill, but that defeats the purpose of skills.

## Environment

- ZeroClaw `0.7.5`
- Anthropic provider, `claude-sonnet-4-6`
- Termux on Android (also reproducible on Linux based on source inspection — the bug is in shared runtime code)

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Skill-derived tool names are composed as `{skill_name}__{tool_name}`. The `__` separator is provider-safe, but the skill/tool names themselves were not sanitized, so illegal characters leaked into the final function name. Provider function-name rules are `^[a-zA-Z0-9_-]{1,64}$`; Anthropic (and others) reject non-conforming names outright.
  - Real-world trigger: plugin-namespaced skills (e.g. `pr-review-toolkit:code-reviewer`) contain `:`. Dots, spaces, and non-ASCII hit the same wall.
  - Introduces a single shared `composed_tool_name()` helper that composes **and** sanitizes the name, and routes **every** registration + rendering path through it: `SkillShellTool::new`, `SkillBuiltinTool::new`, `SkillHttpTool::new`, and the prompt renderer (`skills_to_prompt_with_mode`). Previously only the two non-HTTP construction sites sanitized — leaving an HTTP escape path and a prompt/spec mismatch (the prompt advertised the raw, provider-invalid name while the registered tool used a different one).
  - Sanitization: already-valid, in-length names pass through **unchanged** (no churn for the common case); names needing change map every disallowed char to `_` and append a 64-bit FNV-1a hash of the full raw composed name, kept **within** the 64-char budget so disambiguation survives truncation.
- **Scope boundary:** Skill tool-name *validity* only (#6678). The separate `mcp__` path is untouched. Registry-wide tool-name *uniqueness* is a pre-existing, orthogonal concern (the registry dispatches first-match by name) and is intentionally out of scope; the code comments now distinguish provider *validity* from global *uniqueness* rather than claiming injectivity.
- **Blast radius:** Any skill-derived tool (shell, script, builtin, HTTP) and the prompt that advertises them. Names already valid are unaffected (pass-through). No change to MCP tools, registry dispatch, or non-skill tools.
- **Linked issue(s):** Closes #6678
- **Labels:** `area: skills`, `area: runtime`, `risk: low` (snapshot current labels after apply)

  `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` — current head, green (CI "Test" job). Focused skill tool-name tests + summary:
  ```
  PASS  zeroclaw-runtime tools::skill_tool::tests::skill_tool_name_sanitized_for_provider_regex
  PASS  zeroclaw-runtime tools::skill_tool::tests::skill_tool_name_truncated_to_64_and_stays_distinct
  PASS  zeroclaw-runtime tools::skill_tool::tests::skill_tool_name_sanitization_disambiguates_common_collisions
  PASS  zeroclaw-runtime tools::skill_tool::tests::skill_tool_already_valid_name_is_unchanged
  PASS  zeroclaw-runtime tools::skill_tool::tests::skill_builtin_tool_name_is_prefixed
  PASS  zeroclaw-runtime tools::skill_tool::tests::skill_shell_tool_name_is_prefixed
  PASS  zeroclaw-runtime tools::skill_http::tests::skill_http_tool_name_sanitized_for_provider_regex
  PASS  zeroclaw-runtime tools::skill_http::tests::skill_http_tool_name_is_prefixed
  PASS  zeroclaw-runtime skills::prompt_callable_name_tests::prompt_callable_name_matches_registered_tool_name
  Summary [50.456s] 8469 tests run: 8469 passed, 10 skipped
  ```
  `cargo clippy` (CI "Lint" job): pass. `cargo fmt` (CI "Format" job): pass.

- **Beyond CI — what I manually verified:** a plugin-namespaced HTTP skill with a dotted tool name renders an identical sanitized name in the prompt and in the registered tool spec (covered by `prompt_callable_name_matches_registered_tool_name` + `skill_http_tool_name_sanitized_for_provider_regex`), and a >64-char composed name truncates with the hash suffix kept inside budget. I did NOT exercise live round-trips against every provider; the regex contract (`^[a-zA-Z0-9_-]{1,64}$`) is the provider gate.

None. Sanitization only narrows tool identifiers to `[a-zA-Z0-9_-]` (it can only *remove* characters); no new data is logged, stored, or transmitted. The FNV-1a suffix is a non-cryptographic disambiguator over already-public tool names — no secrets or PII.

## Compatibility (required)

Backward compatible for working configs: names already matching the provider regex and within 64 chars are unchanged, so existing skills keep their current tool names. Skills whose composed name was previously *invalid* (and thus rejected by the provider) now get a stable sanitized name — those were non-functional before, so this is a strict improvement, not a regression. No config, schema, or wire-format change.

Risk: low. Fast rollback = revert this PR's commits; tool names revert to the prior (unsanitized) composition. No feature flag, no migration, no persisted state. Observable failure symptom if a regression slipped in: a provider rejecting a tool call with an invalid-function-name error, or a skill tool failing to resolve by its advertised prompt name — grep gateway logs for the provider's function-name rejection.


Audacity88 (changes requested):
Thanks @perlowja. I reviewed the current diff at 0eecf8baec159a07e67935173ffa6eb87128b25b, linked issue #6678, the empty PR discussion/review history, the visible green checks, the adjacent skill registration and prompt-rendering paths, and the current public Anthropic tool-name contract. I did not run additional local validation.

This is pointed in the right direction, but a few gaps still let #6678 survive.

### 🟢 What looks good — the shell and builtin registration boundary is the right place to sanitize

Applying the sanitized name at tool construction is the right boundary for the shell and builtin skill tools. Those wrappers now expose the sanitized name from their `name()` implementations, so the provider-visible tool name is fixed at the same boundary where the wrapper is registered.

I also like that already-valid names pass through unchanged and that altered names keep a stable hash suffix. That avoids churn for the common case and reduces accidental collisions when dots, spaces, colons, or non-ASCII characters collapse to `_`.

### 🔴 Blocking — HTTP skill tools still use the raw composed name

`SkillHttpTool::new` still builds its name with raw `format!("{}__{}", skill_name, tool.name)`. The adjacent registration path constructs `SkillHttpTool` for `kind = "http"`, and the skill prompt marks `http` tools as callable alongside `shell`, `script`, and `builtin`. That means a plugin-namespaced HTTP skill such as `pr-review-toolkit:code-reviewer` can still expose a provider-invalid function name and still trigger the Anthropic 400 that #6678 is meant to close.

Please route HTTP skill-tool names through the same sanitizer, preferably through a shared helper instead of duplicating the rule, and add focused coverage for an HTTP skill whose skill or tool name contains provider-invalid characters.

### 🔴 Blocking — the skill prompt can tell the model to call names that no longer exist

`build_skills_prompt` still renders callable tool names with raw `format!("{}__{}", skill.name, tool.name)`. After this PR, shell and builtin tool specs use the sanitized name, but the prompt can still tell the model to invoke the unsanitized composed name. If the model follows that prompt text, the call will not match any registered `Tool::name()` and will fail as an unknown tool even though the sanitized tool spec exists.

Please make the prompt render the same registered name that the tool spec exposes. A shared composed-name helper would make this harder to drift again. I would also add a regression that builds a skill with namespaced/dotted characters and asserts that the prompt's callable name matches the registered tool name.

### 🔴 Blocking — the sanitizer still allows names Anthropic documents as invalid

#6678 is an Anthropic/provider rejection bug, so the sanitizer needs to target the strictest provider contract we rely on. This PR sets `MAX_TOOL_NAME_LEN` to 128 and adds tests that only assert the sanitized names are `<=128`, but Anthropic's current client-tool documentation says tool `name` must match `^[a-zA-Z0-9_-]{1,64}$`.

That means a long skill/tool pair can still produce a 65-128 character name that passes this PR's tests but remains outside the documented Anthropic contract. The linked issue's captured server error mentions `{1,128}`, but the public provider contract is stricter now, and this fix should not bake the looser observed error string into ZeroClaw's runtime contract.

Please lower the advertised-name budget to 64 characters, including the hash suffix, and add coverage that a raw composed name over 64 characters is sanitized to `<=64` while still preserving distinct outputs for the tested collision cases.

### 🟡 Warning — the hash comments overstate collision guarantees

The implementation comment correctly says the 64-bit hash is not a uniqueness proof, but the later comment says the hash keeps the result "injective." A bounded 64-bit hash can reduce accidental collisions, but it cannot make the mapping injective. The nearby `_<8 hex>` comment is also stale because `short_hash` renders 16 hex characters.

Please soften the injective wording, fix the suffix-length comment, and consider renaming `skill_tool_name_sanitization_is_injective` to something like `skill_tool_name_sanitization_disambiguates_common_collisions`.

perlowja:
@Audacity88 — all four findings addressed:

- 🔴 **HTTP path:** `SkillHttpTool::new` now routes through a shared `composed_tool_name()` helper, so a plugin-namespaced/dotted HTTP skill gets the same provider-valid name as shell/builtin tools. Added focused coverage (`skill_http_tool_name_sanitized_for_provider_regex`).
- 🔴 **Prompt drift:** the skills prompt (`skills_to_prompt_with_mode`) renders the callable name via the same `composed_tool_name()`, so it advertises exactly the registered tool-spec name. Added a regression asserting prompt name == registered name and that the raw composed name never reaches the prompt.
- 🔴 **Length bound:** `MAX_TOOL_NAME_LEN` 128 → 64 to match Anthropic's `^[a-zA-Z0-9_-]{1,64}$`; tests assert ≤ 64 with distinct outputs for an over-64 composed name.
- 🟡 **Comments/test:** softened the "injective" wording (the bounded 64-bit hash only disambiguates common collisions), fixed the stale `_<8 hex>` comment (`short_hash` renders 16), and renamed the test to `skill_tool_name_sanitization_disambiguates_common_collisions`.

Verified on a build host: `cargo test -p zeroclaw-runtime` (all skill tests + the new HTTP/prompt regressions) passes and clippy is clean. A cross-family adversarial pass additionally suggested guarding the unbounded `format!` allocation and adding registry-level dedup — both are pre-existing (master composes names the same way) and outside #6678's name-validity scope, so I've documented on `composed_tool_name` that it guarantees validity, not global uniqueness, and registration must still resolve duplicates.

WareWolf-MoonWall (commented):
Re-review at c81be4fbf761, addressing @Audacity88's CHANGES_REQUESTED from 0eecf8bae.

### ✅ Resolved — HTTP skill tools now sanitize names

`SkillHttpTool::new` routes through the shared `composed_tool_name()` helper instead of the raw `format!`. The focused `skill_http_tool_name_sanitized_for_provider_regex` test covers a plugin-namespaced HTTP skill with a dotted tool name and confirms both the validity constraint and the pass-through invariant.

### ✅ Resolved — Skill prompt renders the registered name

`skills_to_prompt_with_mode` now calls `composed_tool_name(&skill.name, &tool.name)`, and the new `prompt_callable_name_matches_registered_tool_name` test asserts that the prompt contains the sanitized name and that the raw composed name never reaches the output. This closes the gap where the model could have been told to invoke a name no tool exposes.

### ✅ Resolved — Name budget lowered to 64 characters

`MAX_TOOL_NAME_LEN` is now 64, matching Anthropic's published `^[a-zA-Z0-9_-]{1,64}$` contract. The existing tests were updated and the new HTTP coverage asserts `<= 64`. The doc comment on `sanitize_tool_name` calls out the distinction between the observed error string (128) and the stricter published bound (64) — exactly the right rationale to have on record.

### ✅ Resolved — Hash comments and test name corrected

`short_hash` now documents "16 hex chars", the "injective" claim is softened to "reduces accidental collisions", and the test is renamed `skill_tool_name_sanitization_disambiguates_common_collisions`. The `composed_tool_name` doc note about validity vs. global uniqueness is a useful addition for future maintainers.

---

@Audacity88 — all four findings from your review appear addressed in the current diff. Asking you to re-review and dismiss your CHANGES_REQUESTED if satisfied. I'm posting this as a comment because your block is still active.

### 🔵 Suggestion — watch for a merge-ordering conflict with #7552

PR #7552 (currently open) adds `timeout_secs: Option<u64>` to `SkillTool`. Your new `fn tool(name, kind) -> SkillTool` test helper in `skills/mod.rs` does not include that field. If #7552 lands before this PR, this branch will need a rebase to add `timeout_secs: None` to that literal. Straightforward, but worth flagging now.

Audacity88 (changes requested):
Thanks @perlowja. I reviewed current head c81be4fbf761ffbd5cc56fe82e6bea77b4bdb505 against #6678, my prior `CHANGES_REQUESTED` review, WareWolf-MoonWall's re-review, your 2026-06-13 follow-up, the current diff/source in `skills/mod.rs`, `skill_http.rs`, and `skill_tool.rs`, the PR body, and the green GitHub checks. I did not rerun local cargo.

### ✅ Resolved — HTTP skill tools now use the sanitizer

`SkillHttpTool::new` now calls the shared `composed_tool_name()` helper, and `skill_http_tool_name_sanitized_for_provider_regex` covers a plugin-namespaced HTTP skill with a dotted tool name. That closes the HTTP escape path from my earlier review.

### ✅ Resolved — the skills prompt now advertises the registered name

`skills_to_prompt_with_mode` now renders callable tool names through `composed_tool_name(&skill.name, &tool.name)`, and `prompt_callable_name_matches_registered_tool_name` asserts the prompt contains the sanitized registered name and not the raw provider-invalid name. That fixes the prompt/spec mismatch.

### ✅ Resolved — the sanitizer now targets the stricter 64-character contract

`MAX_TOOL_NAME_LEN` is now 64, the long-name regression asserts `<= 64`, and the helper keeps the hash suffix inside that budget. The comments also now distinguish validity from global uniqueness instead of claiming true injectivity.

### 🟢 What looks good — the naming rule has one source of truth

The shared helper is the right shape for this fix. Shell/script, builtin, HTTP, and prompt rendering now all call the same composition/sanitization path, which makes future drift much less likely than copying the regex rule into each registration path.

### 🔴 Blocking — the PR body and validation evidence are stale after the update

The current code looks like it addresses the technical blockers, but the PR body still describes the old head: it says the sanitizer is applied at only the two original construction sites, cites the 128-character provider rule, lists only the earlier skill-tool tests, and does not include literal output for the latest HTTP/prompt regressions or the latest clippy run mentioned in the follow-up comment. It also leaves an external AI-review result as validation evidence, which is not a substitute for the repo-template command/result evidence.

Please refresh the PR body before this is merge-ready: summarize the current shell/script + builtin + HTTP + prompt scope, update the provider-name limit to 64, include the current focused test/clippy command output tails, and fill the required security/privacy, compatibility, and high-risk rollback sections. Once the public record matches this head, I do not see a remaining code blocker from my prior review.

perlowja:
@Audacity88 — refreshed the PR body to match the current head, which was the only remaining blocker:

- Scope now states all four registration paths + the prompt renderer route through the shared `composed_tool_name()` helper (was: only the two original construction sites).
- Provider name limit corrected to **64** throughout (was 128).
- Replaced the external-review line with real CI command/result evidence — `cargo nextest run --locked --workspace` tails for the focused tests (incl `skill_http_tool_name_sanitized_for_provider_regex`, `prompt_callable_name_matches_registered_tool_name`, and the 64-char regression `skill_tool_name_truncated_to_64_and_stays_distinct`), plus the green Lint/Format jobs. Summary: 8469 passed, 10 skipped.
- Filled the required Security/Privacy, Compatibility, and Rollback sections.

Per your note, the public record now matches this head. Thanks for the thorough passes.

Audacity88 (approved):
Thanks @perlowja. I re-reviewed current head c81be4fbf761ffbd5cc56fe82e6bea77b4bdb505 against #6678, my prior `CHANGES_REQUESTED` reviews, @WareWolf-MoonWall's re-review, your 2026-06-14 PR-body refresh, the current diff/source in `skills/mod.rs`, `skill_http.rs`, and `skill_tool.rs`, the PR body, the empty inline review thread list, and the current green GitHub checks. I did not rerun local cargo.

This clears my previous block on this head. Any remaining merge state still depends on the normal branch-protection/reviewer queue, and @WareWolf-MoonWall's earlier #7552 merge-order note remains a rebase caveat if that PR lands first.

### ✅ Resolved — public record now matches the current fix

The remaining blocker from my last review was PR-body/validation drift, not code. The body now describes the current shell/script, builtin, HTTP, and prompt-rendering scope; names the shared `composed_tool_name()` path; uses the 64-character provider contract; includes the focused HTTP, prompt, long-name, pass-through, and collision regressions; and records the current green `nextest`, lint, and format evidence. The substantive security/privacy, compatibility, and rollback information is also present.

### 🟢 What looks good — the naming rule has one source of truth

The code path is now shaped the way this fix needs: `SkillShellTool::new`, `SkillBuiltinTool::new`, `SkillHttpTool::new`, and `skills_to_prompt_with_mode` all use the same composed-name helper. That closes the original provider-invalid HTTP escape path and the prompt/spec mismatch, and the tests cover the failure modes that made #6678 survive the first pass.

I also like the pass-through behavior for already-valid names and the bounded hash suffix for changed names. That gives stable provider-valid names without unnecessary churn for existing working skills.

singlerider (approved):
Reviewed at head `c81be4f` against current `master`, focused on the single-source naming property this fix depends on. Approving.

### 🟢 One sanitizer, applied to both the spec name and the prompt — which is the whole point

The bug class in #6678 is the model being told to call a tool name the provider's function-name regex rejects, or a name the prompt advertises but no registered spec exposes. This fix closes both by routing every name through one `composed_tool_name()` helper: `SkillShellTool::new`, `SkillBuiltinTool::new`, `SkillHttpTool::new`, and the `skills_to_prompt_with_mode` prompt renderer all call it (confirmed across `skills/mod.rs`, `skill_http.rs`, `skill_tool.rs`), and the prompt-render call site carries an explicit comment that it must match the registered spec name exactly. So the prompt and the tool spec can no longer drift.

### 🟢 The sanitization matches the provider contract and disambiguates safely

The char predicate is `c.is_ascii_alphanumeric() || c == '_' || c == '-'` — exactly `^[a-zA-Z0-9_-]+$` — bounded by `MAX_TOOL_NAME_LEN = 64` (with a documented note that #6678's server error mentioned 128 but 64 is the published contract, so targeting 64 is the conservative correct choice). Already-valid names pass through unchanged (no churn for working skills), and altered names get a bounded FNV-1a `short_hash` suffix to reduce accidental collisions among sanitized names. That gives stable, provider-valid names without rewriting names that already work.

### 🟢 Coverage and state

The regressions cover the HTTP escape path, prompt/spec matching, long-name truncation, pass-through, and collision cases — the exact failure modes that let #6678 survive the first pass. @Audacity88 cleared his block after two iterations; CI is green and the branch is `MERGEABLE` / `CLEAN`. (@WareWolf-MoonWall's #7552 merge-order note is a rebase caveat if that lands first, not a code issue.)

Approving — correct sanitizer, single source of truth for the name across spec and prompt, well-tested.

PR #7401, opened 2026-06-08
test(config): guard runtime_profiles.max_tool_iterations precedence (#6877)

### LINKED ISSUE #6877: [runtime_profiles.*].max_tool_iterations has no effect — must be set on [agents.*] instead
## Version
0.8.0-beta-1 (verified by source inspection and empirical testing)

## Summary

Setting `max_tool_iterations` under `[runtime_profiles.default]` has no effect on the agent tool-call loop. The effective field is `max_tool_iterations` on `[agents.*]` (i.e., `AliasedAgentConfig`). `RuntimeProfileConfig` also has a `max_tool_iterations` field but it may not be wired to the agent loop.

## Source reference

Both `AliasedAgentConfig` and `RuntimeProfileConfig` declare `max_tool_iterations`:

```rust
// crates/zeroclaw-config/src/schema.rs

// On AliasedAgentConfig (line ~2790) — THIS ONE controls the loop
#[serde(default = "default_agent_max_tool_iterations")]
pub max_tool_iterations: usize,  // default: 10

// On RuntimeProfileConfig (line ~8859) — unclear if used for loop
pub max_tool_iterations: usize,
```

The agent loop in `agent.rs` uses `self.config.max_tool_iterations` where `self.config` is `AliasedAgentConfig`:

```rust
// crates/zeroclaw-runtime/src/agent/agent.rs
for _ in 0..self.config.max_tool_iterations {
    // tool call loop
}
```

## Reproduction

```toml
[runtime_profiles.default]
max_tool_iterations = 1  # ← has no observable effect

[agents.my_agent]
runtime_profile = "default"
```

**Observed:** agent makes 3+ tool calls despite `max_tool_iterations = 1` in runtime profile.

**Fix:** move the setting to the agent definition:

```toml
[agents.my_agent]
runtime_profile = "default"
max_tool_iterations = 1  # ← this works
```

## Expected behavior

Either:
1. Document that `max_tool_iterations` must be on `[agents.*]`, not `[runtime_profiles.*]`, or
2. Wire `RuntimeProfileConfig.max_tool_iterations` into the agent loop so both locations work, or
3. Remove `max_tool_iterations` from `RuntimeProfileConfig` to avoid confusion

## Impact

Users configuring `[runtime_profiles.*]` to cap tool iterations will see no effect and potentially runaway tool call loops (up to the default of 10) even with explicit configuration.

perlowja:
## Summary

- **Linked issue:** Closes #6877
- **Context:** #6877 reported that `[runtime_profiles.*].max_tool_iterations` had no effect (the value had to be set on `[agents.*]`). As of #6848 (beta-2 integration) the resolution was inverted and fixed: runtime tunables now live on `[runtime_profiles.<p>]`, referenced via `agents.<alias>.runtime_profile`, and the agent-inline copies are intentionally inert (`agent_level_tunable_keys_are_inert`). So the reported behavior already works on `master` — but the precedence had **no regression test**, and the resolution comment still described agent-inline as a fallback, which is no longer true.
- **What changed:**
  - Adds `runtime_profile_max_tool_iterations_is_honored` — a profile's `max_tool_iterations = 25` resolves to `25`.
  - Adds `runtime_profile_unset_max_tool_iterations_uses_default` — an unset (sentinel `0`) profile value falls back to the default `10`, not `0`.
  - Corrects the stale comment on `effective_max_tool_iterations` to state that the profile is authoritative and agent-inline copies are inert (not a fallback).
- **Scope boundary:** Tests + comment only; no behavior change.

## Validation Evidence

- `cargo fmt -p zeroclaw-config` — clean.
- `cargo test -p zeroclaw-config --lib -- runtime_profile_max_tool_iterations_is_honored runtime_profile_unset_max_tool_iterations_uses_default agent_level_tunable_keys_are_inert` — `3 passed; 0 failed`.


Audacity88 (approved):
I checked the live PR state, current diff, linked issue #6877, prior #6848 context, and the required checks. This looks like a narrow regression-test and comment-cleanup closeout for the runtime-profile precedence bug.

### 🟢 What looks good — the test locks the runtime-profile contract

#6877 reported that `[runtime_profiles.*].max_tool_iterations` did not affect the agent loop. This PR adds the missing direct regression coverage for both the configured-profile value and the unset/default case.

The updated comment also removes the misleading agent-inline fallback language. That matters because the intended direction is now clear in one place: runtime tunables live on runtime profiles, and the old agent-inline copies should not be treated as another source of truth.

### 🟢 What looks good — this is a safe closeout for #6877

The patch is test/comment-only, the linked issue's central behavior is covered by the new assertions, and the required CI gate is green. I do not see a remaining reviewer-side blocker for this closing #6877.

Issue #7042, opened 2026-05-31
[Bug]: daemon IPC crash on file-descriptor exhaustion (EMFILE / too many open files)


singlerider:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

The daemon (and/or the `zerocode` TUI) crashes with a file-descriptor exhaustion error of the form "IPC: too many open files" (`EMFILE` / `ENFILE`) after the process has been running and accepting/serving IPC connections for a while. Once the FD limit is hit, the process can no longer `accept()` new IPC connections, open log/catalog files, or establish outbound provider sockets, and it dies (or wedges) mid-operation.

Observed against the `integration/zeroclaw-tui` line where the `zerocode` TUI talks to the daemon over the local Unix-socket IPC endpoint (`<config_dir>/data/daemon.sock`).

### Expected behavior

FD usage stays bounded under steady-state operation and across repeated TUI connect/disconnect/reconnect cycles. Per-connection resources (the accepted socket and its reader/writer/router tasks) are fully reaped when a connection drops, so the daemon can run indefinitely without approaching the soft `RLIMIT_NOFILE`.

### Steps to reproduce

```bash
# Exact repro under investigation; the failure is FD growth over time.
# Working hypothesis for a fast repro:
1. zeroclaw daemon            # start the daemon (or let zerocode spawn an ephemeral one)
2. # repeatedly connect/disconnect a TUI against the local socket:
   for i in $(seq 1 500); do zerocode -m :q 2>/dev/null; done
3. # watch the daemon's FD count climb and not return to baseline:
   watch -n1 'ls -1 /proc/$(pgrep -f "zeroclaw daemon")/fd | wc -l'
4. # eventually: "IPC: too many open files" / EMFILE, daemon crashes or wedges
```

### Impact

```
Affected users: anyone running the daemon long-lived with the zerocode TUI,
                especially with reconnect churn (service restarts, daemon
                config reloads that drop the socket, multiple TUIs).
Frequency: intermittent — manifests after enough connections/uptime; reliably
           on machines with a low ulimit -n.
Consequence: hard crash / wedge of the daemon; all connected TUIs lose their
             session and the daemon stops accepting new IPC clients.
```

### Logs / stack traces

```
IPC max open files (EMFILE / "Too many open files") on accept()/open().
Full stack trace not yet captured — to be attached from a daemon run with
RUST_BACKTRACE=1 once a deterministic repro is locked down.
```

[...]

PR #7402, opened 2026-06-08
fix(gateway): survive transient accept() errors instead of crashing (#7042)

perlowja:
## Summary

- **Linked issue:** Related #7042 (partial hardening for one EMFILE failure mode; does not close the broader daemon/TUI IPC FD-exhaustion investigation).
- **Root cause:** The gateway's manual **TLS** accept loop (`crates/zeroclaw-gateway/src/lib.rs`) did `let (tcp_stream, remote_addr) = conn?;` — propagating *every* `listener.accept()` error. Under file-descriptor exhaustion (`EMFILE`), `accept()` errors, the `?` bubbles up, and the entire gateway/daemon serve loop exits. The plain-TCP path (`axum::serve`) already tolerates transient accept errors; only the TLS path was fatal.
- **Fix:**
  - Adds `is_recoverable_accept_error` classifying transient errors: `EMFILE`/`ENFILE` (fd exhaustion, the issue) plus `ConnectionAborted`/`Interrupted`/`WouldBlock`.
  - On a recoverable error the loop logs a warning, backs off `ACCEPT_ERROR_BACKOFF_MS` (50ms, avoids hot-spin), and continues serving. Genuinely fatal errors still propagate.
- **Scope boundary:** TLS accept loop only; no change to the plain-TCP path or connection handling.

## Validation Evidence

- `cargo fmt -p zeroclaw-gateway` — clean.
- `cargo test -p zeroclaw-gateway --lib accept_error` — `2 passed; 0 failed` (`fd_exhaustion_accept_errors_are_recoverable`, `transient_kinds_recover_but_fatal_propagates`).
- Manual repro (from the issue): run the daemon with a low `ulimit -n` and a stdio MCP server under heartbeat; before this change the daemon exits on the first EMFILE, after it logs + keeps serving.
- Fleet adversarial review gate (`ngc-review`, gpt-5.5 via NGC proxy): **APPROVE**.

## Note

Testing a real EMFILE in CI is impractical/flaky (it would exhaust the test process's own fds), so coverage is on the error classifier; the loop wiring is a thin match around it.

### DIFF
--- a/crates/zeroclaw-gateway/src/lib.rs
+++ b/crates/zeroclaw-gateway/src/lib.rs
@@ -64,6 +64,38 @@ use std::collections::HashMap;
 use std::net::{IpAddr, SocketAddr};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+
+/// Backoff after a transient `accept()` error so the serve loop does not
+/// hot-spin while the condition (e.g. fd exhaustion) clears.
+const ACCEPT_ERROR_BACKOFF_MS: u64 = 50;
+
+/// File-descriptor exhaustion errno values, stable across the Unix targets
+/// we support (Linux, macOS, BSD).
+#[cfg(unix)]
+const EMFILE: i32 = 24; // too many open files (this process)
+#[cfg(unix)]
+const ENFILE: i32 = 23; // too many open files (system-wide)
+
+/// Returns `true` when an error from a stream listener's `accept()` is
+/// transient and the listener itself remains usable, so the serve loop
+/// should log and keep running rather than terminating the daemon. Covers
+/// file-descriptor exhaustion (`EMFILE`/`ENFILE`, see #7042) and the usual
+/// per-connection hiccups. Mirrors the non-fatal accept handling that
+/// `axum::serve` already performs on the plain-TCP path.
+fn is_recoverable_accept_error(e: &std::io::Error) -> bool {
+    use std::io::ErrorKind;
+    if matches!(
+        e.kind(),
+        ErrorKind::ConnectionAborted | ErrorKind::Interrupted | ErrorKind::WouldBlock
+    ) {
+        return true;
+    }
+    #[cfg(unix)]
+    if matches!(e.raw_os_error(), Some(EMFILE) | Some(ENFILE)) {
+        return true;
+    }
+    false
+}
 use tower_http::limit::RequestBodyLimitLayer;
 use tower_http::timeout::TimeoutLayer;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
I checked the live PR state, current diff, required checks, and linked issue #7042. The TLS accept-loop patch looks like useful hardening, but I am blocking on merge bookkeeping and current CI rather than the core classifier shape.

### 🟢 What looks good — transient TLS accept errors keep the listener alive

The previous `conn?` path made the manual TLS listener fragile: one transient `accept()` error could take down the gateway serve loop. Classifying transient accept errors, logging them, backing off briefly, and continuing is the right direction for that path. The classifier tests cover the central decision boundary for EMFILE/ENFILE and several standard transient `ErrorKind`s.

### 🔴 Blocking — `Closes #7042` overstates this PR's scope

#7042 is broader than the TLS accept loop. It reports daemon/TUI IPC file-descriptor exhaustion on the local Unix-socket path, with suspected causes around per-connection IPC cleanup, reconnect churn, rejected or half-open local connections, low `RLIMIT_NOFILE`, and retained notification/router state.

This PR only changes the gateway TLS listener's TCP accept loop. That is useful related hardening, but it does not prove the daemon/local IPC FD leak is fixed. If this merges with `Closes #7042`, GitHub will auto-close the broader P1 issue even though the local IPC investigation still appears open.

Please change the PR relationship from `Closes #7042` to `Related #7042`, or otherwise keep the remaining daemon/local IPC work explicitly tracked before this merges.

### 🔴 Blocking — required CI is currently red

The current Lint job is failing, which leaves CI Required Gate failing. Because this is gateway availability work, I cannot approve it until the required gate is green or the failed run is replaced by clear maintainer-accepted infrastructure evidence.

Audacity88 (approved):
I re-reviewed current head 61ad574 after the PR-body cleanup, the current diff, the current #7042 framing in the PR body, my earlier CHANGES_REQUESTED review, and the latest GitHub checks. I am approving to clear my prior review blockers: the PR no longer auto-closes the broader #7042 investigation, and required CI is green.

### ✅ Resolved — #7042 now stays open for the broader IPC work

The PR body now uses `Related #7042` and explicitly calls this a partial hardening slice for one EMFILE failure mode. That avoids auto-closing the broader daemon/TUI IPC FD-exhaustion investigation described in the PR body and earlier review, while keeping this TLS accept-loop hardening properly related to that issue.

### ✅ Resolved — Required checks are green

The failed Lint / CI Required Gate path from my earlier review has recovered. The visible check set is green, including Lint, Test, Security, platform builds, and CI Required Gate.

### 🟢 What looks good — TLS accept failures now fail soft

The core patch is still the right shape: recoverable `accept()` errors such as EMFILE/ENFILE, `ConnectionAborted`, `Interrupted`, and `WouldBlock` are logged, briefly backed off, and then the TLS listener continues serving. Fatal errors still propagate. The classifier tests cover the central boundary without trying to exhaust the test process's own file descriptors.

### 🟡 Warning — Treat merge readiness as a fresh step

This approval clears my review block, but it is not a final merge-readiness statement. Current `master` has changed gateway files, including `crates/zeroclaw-gateway/src/lib.rs`, after this PR's merge base. Before merge, refresh the branch and let CI rerun, or use equivalent merge-result evidence for the gateway package.

PR #7403, opened 2026-06-08
fix(runtime): guard trim_history against orphan-cascade emptying all messages

tidux:
## Summary

**Base branch:** `master` (all contributions)

**What changed and why:**
Added a safety guard in `trim_history` (`agent.rs`) that detects when the orphan-removal cascade would drain all non-system messages from conversation history and skips the trim instead of producing an empty `messages: []` array.

**Root cause:** with `max_history_messages` defaulting to 50 (the runtime-profile default when `[runtime_profiles.<profile>]` does not set it), a long single-turn tool loop accumulates `[user_msg, AC1, TR1, …, AC25, TR25]` = 51 entries. `initial_drop_count = 1` removes `user_msg`; the orphan-`AssistantToolCalls` cascade then sweeps every remaining AC/TR pair to end-of-list, leaving only the system message. `convert_messages` extracts the system message as `system_prompt`, producing `messages: []`, which Anthropic rejects with HTTP 400 `"messages: at least one message is required."`

**The fix:** after the cascade loops advance `drop_count` to `other_messages.len()`, restore history unchanged and emit a `WARN` log. The session stays temporarily over the message limit but continues to function. A companion config fix (outside this diff) adds `max_history_messages = 500` to the relevant runtime profile so the default no longer triggers this edge case under normal workloads.

**Scope boundary:** only `trim_history` in `crates/zeroclaw-runtime/src/agent/agent.rs`; no provider code, schema, session persistence, or CLI surface touched.

**Blast radius:** agents running long single-turn tool loops with a low effective `max_history_messages`. The change turns a hard-crash 400 into a temporary over-limit; no other code path is affected.

**Split note:** this replaces #7119, which inadvertently bundled an unrelated Matrix `/sync` long-poll fix in the same branch. The Matrix change is now its own PR (linked below). Per @singlerider's review on #7119, each PR's body now honestly covers its own diff.

**Linked issue(s):** None — diagnosed from `runtime-trace.jsonl` (session `ce3e5d7b`, 2026-06-02).
**Related PRs:** Companion split-out — Matrix sync fix in #7404. Supersedes #7119 (to be closed).
**Labels:** `bug`, `risk: high` (matches runtime-path auto-label from #7119), `agent`, `runtime`, `size: XS`.

```
cargo check -p zeroclaw-runtime
cargo test -p zeroclaw-runtime --lib agent::agent::tests::trim_history
```

- `cargo check -p zeroclaw-runtime` against current `upstream/master`:
  ```
  Checking zeroclaw-runtime v0.8.0-beta-2 (…/crates/zeroclaw-runtime)
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 43.26s
  ```

- `cargo test -p zeroclaw-runtime --lib agent::agent::tests::trim_history` against current `upstream/master`:
  ```
  running 2 tests
  test agent::agent::tests::trim_history_does_not_leave_orphan_assistant_tool_calls ... ok
  test agent::agent::tests::trim_history_does_not_leave_orphan_tool_results ... ok
  test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 2013 filtered out; finished in 0.00s
  ```

- Branch is based on `upstream/master` post-#7231 (the `ollama.rs` E0308 revert that was breaking CI on #7119 has landed), so CI on this branch should be green rather than inheriting the prior master breaker.

Reproduced the failure sequence from `runtime-trace.jsonl`: session `ce3e5d7b` on 2026-06-02 ran 24+ tool-call iterations in Turn 2, causing 400 at iteration 25. Traced through `trim_history` cascade logic to confirm `drop_count` reaches `other_messages.len()` when `other_messages = [user_msg, AC1…AC25, TR1…TR25]` and `initial_drop_count = 1`. Did not exercise the WARN log path in a live session; that requires a session with exactly `max + 1` non-system history entries where the `+1` is the sole user message.

**If any command was intentionally skipped, why:** None skipped.

## Compatibility (required)

- Backward compatible? **Yes** — the only behavioral change is in the edge case where `trim_history` previously emptied history (causing a 400); now it skips the trim instead.

## Rollback (required for risk: medium and risk: high)

- **Fast rollback command/path:** `git revert <commit-sha-of-this-PR>` (single commit on the branch)
- **Observable failure symptoms:** After revert, sessions with long single-turn tool loops may again produce `WARN "Streaming error:"` followed by `ERROR "turn failed"` with body `"messages: at least one message is required"` in `runtime-trace.jsonl`. Look for `"_file": "crates/zeroclaw-providers/src/reliable.rs"` + `"status": 400` entries.

- Superseded PRs + authors: #7119 by @tidux
- Scope materially carried forward: `trim_history` orphan-cascade guard and runtime failure analysis. The unrelated Matrix `/sync` long-poll change from #7119 was split out to #7404.
- `Co-authored-by` trailers added in commit messages for incorporated contributors? **No**
- If `No`, why: Same author carried the replacement scope forward; no separate contributor code or design was incorporated.

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/agent.rs
+++ b/crates/zeroclaw-runtime/src/agent/agent.rs
@@ -1546,6 +1546,32 @@ impl Agent {
                 );
             }
 
+            // Safety: the orphan-removal cascades above can advance
+            // drop_count all the way to other_messages.len() when the only
+            // non-tool-call entry is the user message at position[0] and
+            // initial_drop_count drops it (e.g. max=50, history=[user,
+            // AC1, TR1, …, AC25, TR25]).  Sending zero messages to the
+            // provider causes a hard 400 "messages: at least one message
+            // is required".  When the cascade would wipe everything, skip
+            // this trim pass so the conversation stays functional even
+            // though it is temporarily over the message limit.
+            if drop_count >= other_messages.len() {
+                ::zeroclaw_log::record!(
+                    WARN,
+                    ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                        .with_category(::zeroclaw_log::EventCategory::Agent)
+                        .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
+                        .with_attrs(::serde_json::json!({
+                            "history_len": other_messages.len(),
+                            "max_history_messages": max,
+                        })),
+                    "trim_history: orphan-cascade would empty all non-system messages; skipping trim to preserve conversation"
+                );
+                self.history = system_messages;
+                self.history.extend(other_messages);
+                return;
+            }
+
             ::zeroclaw_log::record!(
                 DEBUG,
                 ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Complete)
--- a/crates/zeroclaw-runtime/src/agent/agent.rs
+++ b/crates/zeroclaw-runtime/src/agent/agent.rs
@@ -5580,6 +5580,281 @@ mod tests {
         }
     }
 
+    /// Regression test for the orphan-cascade-empties-everything 400.
+    ///
+    /// When `max_history_messages` is small relative to a long single-turn
+    /// tool loop, the orphan-removal cascades in `trim_history` can advance
+    /// `drop_count` all the way to `other_messages.len()`. Before the guard,
+    /// `other_messages.drain(0..drop_count)` then emptied the entire
+    /// non-system history, `convert_messages` lifted the system entry into
+    /// `system_prompt`, and the provider received `messages: []` — a hard
+    /// 400 `"messages: at least one message is required"` from Anthropic.
+    ///
+    /// Minimal repro of the cascade:
+    ///   history = [user, AC1, TR1, AC2, TR2]   (no system message)
+    ///   max_history_messages = 4
+    ///   other_messages.len() = 5, initial_drop_count = 1 → drops `user`
+    ///   orphan-TR cascade: pos 1 = AC1, not TR → no-op
+    ///   orphan-AC cascade: pos 1 = AC1, pos 2 = TR1, pos 3 = AC2,
+    ///                      pos 4 = TR2 → drop_count = 5
+    ///   drop_count (5) >= other_messages.len() (5) → guard MUST fire.
+    ///
+    /// Without the guard this test fails: `agent.history` ends up empty,
+    /// which is exactly the shape that crashes the provider call.
+    /// With the guard, the original history is preserved unchanged so the
+    /// session stays over-limit-but-functional.
+    #[test]
+    fn trim_history_does_not_empty_all_messages_on_full_cascade() {
+        use zeroclaw_providers::{ToolCall, ToolResultMessage};
+
+        let memory_cfg = zeroclaw_config::schema::MemoryConfig {
+            backend: "none".into(),
+            ..zeroclaw_config::schema::MemoryConfig::default()
+        };
+        let mem: Arc<dyn Memory> = Arc::from(
+            zeroclaw_memory::create_memory(&memory_cfg, std::path::Path::new("/tmp"), None)
+                .expect("memory creation should succeed with valid config"),
+        );
+
+        let agent_config = zeroclaw_config::schema::AliasedAgentConfig {
+            resolved: zeroclaw_config::schema::ResolvedRuntime {
+                max_history_messages: 4,
+                ..Default::default()
+            },
+            ..zeroclaw_config::schema::AliasedAgentConfig::default()
+        };
+
+        let observer: Arc<dyn Observer> = Arc::from(crate::observability::NoopObserver {});
+        let mut agent = Agent::builder()
+            .model_provider(Box::new(MockModelProvider {
+                responses: Mutex::new(vec![]),
+            }))
+            .tools(vec![Box::new(MockTool)])
+            .memory(mem)
+            .observer(observer)
+            .tool_dispatcher(Box::new(NativeToolDispatcher))
+            .workspace_dir(std::path::PathBuf::from("/tmp"))
+            .config(agent_config)
+            .build()
+            .expect("agent builder should succeed with valid config");
+
+        // user
+        agent.history.push(ConversationMessage::Chat(ChatMessage {
+            role: "user".into(),
+            content: "kick off a long tool loop".into(),
+        }));
+        // AC1, TR1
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
I reviewed current head fa8a767, the PR body, the one-file `agent.rs` diff, the existing trim-history tests, and the empty current comment/review threads. I did not rerun local cargo. Visible CI is still in progress while I am reviewing.

@tidux Thanks for splitting this out from #7119.

The fix is in the right area and the behavior change is narrow. I want one focused regression before clearing it, because this PR is specifically about a crash shape that the current tests do not cover.

### 🟢 What looks good — The runtime fallback is tightly scoped

The new guard lands after the existing orphan `ToolResults` / `AssistantToolCalls` cascades and only fires when the cascade would drain every non-system message. Restoring the original system + non-system history is safer than sending `messages: []`, and keeping the session temporarily over the trim limit is a reasonable failure mode for this edge case.

### 🔴 Blocking — Add a regression for the cascade-to-empty path

The PR body describes the exact failure shape: with a low effective `max_history_messages`, `initial_drop_count = 1` drops the only user message and the orphan `AssistantToolCalls` cascade then consumes every remaining AC/TR pair. The diff adds the guard for that path, but it does not add a test that would fail on current `master` without the guard.

The PR validation only shows the two existing orphan trim-history tests, and this diff does not add a regression for the new cascade-to-empty contract. That contract is the load-bearing part of this fix: when the orphan cascade would advance `drop_count` to `other_messages.len()`, `trim_history` must preserve a non-empty provider-visible conversation instead of draining all non-system messages.

Please add a focused test for this edge case. It can use a smaller equivalent than the 25-pair production trace: for example, build `[user, AC1, TR1, AC2, TR2]` with a `max_history_messages` value that drops the sole user message and lets the orphan-AC cascade reach the end, then assert that after `trim_history()` the history still contains non-system messages and is not reduced to only the system prompt. That locks the actual 400 fix so future cleanup does not remove it as "defensive" code.

tidux:
@Audacity88 Thanks — fair call. Pushed `7e9e4a334` adding two regression tests that lock the cascade-to-empty contract.

### What changed

`crates/zeroclaw-runtime/src/agent/agent.rs` (tests-only, +275 lines, no production-code change):

1. **`trim_history_does_not_empty_all_messages_on_full_cascade`** — minimal repro of the production failure you described:
   - `history = [user, AC1, TR1, AC2, TR2]`, `max_history_messages = 4`
   - `initial_drop_count = 1` drops the sole user message
   - orphan-AC cascade then sweeps `AC1 / TR1 / AC2 / TR2`, driving `drop_count` from `1` to `5` (= `other_messages.len()`)
   - asserts the guard fires, history is non-empty, and the full 5-entry history is preserved unchanged
   - also asserts `history.len() > max_history_messages` after the guard fires, codifying the "temporarily over max" trade-off so a future "tighten trim_history" refactor cannot silently regress to empty-messages without breaking a test

2. **`trim_history_full_cascade_with_system_message_preserves_full_history`** — same cascade arithmetic with a system message at the head. Verifies the guard's restore path puts the conversation back together in the right order (`system_messages` first, then non-system entries) rather than dropping the system message or returning the halves reversed. Also asserts at least one non-system message remains — without this, `convert_messages` would lift the system entry into `system_prompt` and the provider would still see `messages: []`.

### Verifying the contract holds

I manually confirmed both new tests would fail on `master` (i.e. without the guard) by locally neutering the guard's early-return and rerunning:

```
running 4 tests
test agent::agent::tests::trim_history_does_not_leave_orphan_assistant_tool_calls ... ok
test agent::agent::tests::trim_history_does_not_leave_orphan_tool_results ... ok
test agent::agent::tests::trim_history_does_not_empty_all_messages_on_full_cascade ... FAILED
test agent::agent::tests::trim_history_full_cascade_with_system_message_preserves_full_history ... FAILED

failures:

---- trim_history_does_not_empty_all_messages_on_full_cascade stdout ----
panicked at agent.rs:5691:9:
trim_history drained every non-system message; the next provider call would fail with 'messages: at least one message is required'

---- trim_history_full_cascade_with_system_message_preserves_full_history stdout ----
panicked at agent.rs:5837:9:
assertion `left == right` failed: trim_history dropped messages from the non-system half despite the orphan cascade reaching other_messages.len(); guard must preserve every entry when it fires
  left: 1
 right: 6
```

The failure messages match the documented failure shape exactly. The two pre-existing orphan tests pass either way (the new tests cover a strictly new contract). Guard was restored before the commit was staged — the staged diff is tests-only.

### Verification on the actual branch

With the guard in place:

```
cargo fmt --all -- --check     # clean
cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings   # clean
cargo test -p zeroclaw-runtime --lib agent::agent::tests::trim_history
  running 4 tests
  test trim_history_does_not_leave_orphan_tool_results ... ok
  test trim_history_does_not_empty_all_messages_on_full_cascade ... ok
  test trim_history_does_not_leave_orphan_assistant_tool_calls ... ok
  test trim_history_full_cascade_with_system_message_preserves_full_history ... ok
```

Let me know if you'd like a third variant exercising a longer trace (e.g. the actual 25-pair production scenario) — happy to add it, but the 2-pair minimal version drives the exact same code path and is cheaper to read.

Audacity88 (approved):
I reviewed current head 7e9e4a, the updated PR body, the new top-level reply, the prior Audacity88 changes-requested review at fa8a767, the one-file `agent.rs` diff, the new full-cascade regression tests, and the current check state. There are no inline review threads. I did not rerun local cargo; branch CI is still in progress while I am reviewing.

@tidux Thanks for the quick follow-up. This resolves the blocker I raised on the first pass.

### ✅ Resolved — Full-cascade regression is now covered

The new tests exercise the exact failure shape from the earlier review: a low `max_history_messages` value drops the only user message, then the orphan `AssistantToolCalls` cascade advances to the end of `other_messages`.

The no-system case locks the important trade-off: when the guard fires, the session stays temporarily over the configured history limit instead of producing an empty provider-visible conversation. The system-message case covers the other dangerous shape: preserving only the system prompt would still become `messages: []` after conversion, so the test checks that non-system messages survive too.

That is the contract this PR needs to protect. It should catch the future cleanup/regression I was worried about.

### 🟢 What looks good — The fallback remains narrow

The production change still only fires after the existing orphan `ToolResults` and `AssistantToolCalls` cleanup has determined that trimming would remove every non-system message. In that edge case, preserving the original history is the safer failure mode than sending an invalid provider payload and hard-failing the turn.

I am approving the code path. Final merge readiness still depends on the in-progress CI finishing green.

Issue #6576, opened 2026-05-11
[Release gate]: Matrix live homeserver smoke check before v0.8.0 stable


Audacity88:
### Summary

Add a v0.7.6 release-gate smoke check for Matrix channel behavior after the matrix-sdk 0.17 bump landed in #6566.

### Problem statement

#6566 merged the matrix-sdk 0.16 -> 0.17 compatibility bump plus the `channel-matrix` recursion-limit fix and low-memory install heuristic. The PR was reviewed and merged with green CI, but both the PR body and reviews explicitly noted that no live Matrix homeserver smoke test was run.

That is acceptable for merging the dependency/compile fix, but before shipping a release that includes the bump, we should verify the main runtime Matrix flows against a real homeserver so SDK API compatibility is not only compile/test-suite confidence.

### Proposed solution

Before v0.7.6 ships, run a live Matrix smoke test on a real homeserver covering:

- send a normal message
- send or receive a threaded reply
- send a reaction
- send an attachment
- send or receive the voice path if the configured test setup supports it

Record the tested commit/version, homeserver type if relevant, and any known skips. If one of these paths fails, open or link a follow-up bug and decide whether it blocks the release.

### Non-goals / out of scope

- No new Matrix feature work.
- No requirement that every Matrix dependency bump block on live homeserver smoke before merge.
- No requirement to commit credentials, homeserver URLs, tokens, or private room IDs.

### Acceptance criteria

- A maintainer or reviewer records a v0.7.6 Matrix live smoke result after #6566 is in the tree.
- The result covers send, threaded reply, reaction, attachment, and voice, or explicitly names skipped paths.
- Any failure is linked to a follow-up issue or called out as a release blocker/non-blocker.

### Architecture impact

Matrix channel runtime validation only. No code architecture change is proposed by this issue.

### Risk and rollback

Low implementation risk: this is a release validation task. The risk being managed is shipping matrix-sdk 0.17 without a real-homeserver sanity check.

### Breaking change?

No

### Data hygiene checks

- [x] I removed personal/sensitive data from examples, payloads, and logs.
- [x] I used neutral, project-focused wording and placeholders.

theonlyhennygod:
## Triage Summary

**Type:** Enhancement (release gate)
**Priority:** Medium (p2)
**Area:** channel, channel:matrix
**Risk:** Low

[...]

PR #7404, opened 2026-06-08
fix(channels): prevent Matrix /sync from timing out at exactly 30 seconds

tidux:
## Summary

**Base branch:** `master` (all contributions)

**What changed and why:**
The `matrix-sdk` defaults to a 30-second per-request HTTP timeout while `SyncSettings::default()` sends no `?timeout=` parameter, so the homeserver returns immediately and the SDK busy-polls — every 30-second window then races the HTTP deadline and idle `/sync` requests error out at exactly 30s.

Relates to #6576.

**The fix has two parts:**
1. Set an explicit 60-second `RequestConfig` timeout on the `Client::builder()` so the HTTP layer doesn't fire before a long-poll can complete (new `CLIENT_REQUEST_TIMEOUT` const).
2. Pass a 30-second long-poll timeout to both `sync_once` and the long-running `sync` call so the homeserver holds idle requests open and returns before the HTTP deadline (new `SYNC_LONGPOLL_TIMEOUT` const).

The two constants are documented relative to each other: `SYNC_LONGPOLL_TIMEOUT` must stay strictly below `CLIENT_REQUEST_TIMEOUT` so the HTTP request deadline never fires before the long-poll completes server-side. The existing integration test in the file is also updated to pass `SYNC_LONGPOLL_TIMEOUT` to `sync_once`.

**Scope boundary:** only `crates/zeroclaw-channels/src/matrix.rs`; no runtime, agent, provider, schema, session persistence, or CLI surface touched.

**Blast radius:** Matrix channel only. Worst case if the timeout values are wrong is `/sync` behaves the same way it does today (request times out at 30s) — i.e. the fix can only equal or improve the current behavior on the Matrix path. No other channel is affected.

**Split note:** this is the second half of the now-closed #7119, split per @singlerider's review. The other half (the `trim_history` runtime fix) is in its own PR (linked below). Each PR's body now honestly covers its own diff.

**Linked issue(s):** None — diagnosed from observed 30-second timeout pattern in Matrix sync logs.
**Related PRs:** Companion split-out — `trim_history` orphan-cascade guard in #7403. Half of the original #7119 (to be closed).
**Labels:** `bug`, `risk: medium`, `channel`, `channel:matrix`, `size: XS`.

## Reviewer feedback addressed (since @Audacity88's CHANGES_REQUESTED)

@Audacity88 blocked on the lack of live-or-near-live Matrix evidence that an idle `/sync` no longer errors at the 30-second cadence. Two follow-ups in commit `cb2ccc3a`:

1. **Live daemon evidence** — `matrix.todixuclawbot` ran against a real homeserver after the fix; sync loop logs show no 30s-cadence errors. Excerpt + Element X end-to-end screenshot are in the comments below.
2. **Smoke test added** — new `matrix::tests::live_smoke::idle_sync_does_not_error_at_30s_cadence` (file: `crates/zeroclaw-channels/src/matrix.rs`). `#[ignore]`'d, reuses the existing `ZEROCLAW_MATRIX_SMOKE_*` env contract, soaks an idle channel for `> 30s` (default 35s, tunable via `ZEROCLAW_MATRIX_SMOKE_IDLE_SECS`), and asserts:
   - no `sync_once` call returns an error (primary anti-regression for the 30s HTTP-deadline bug),
   - at least one round-trip durably long-polls (anti-regression for the pre-fix busy-poll pattern),
   - sub-threshold returns stay inside a small budget (defense-in-depth).

   It also emits a one-line `eprintln!` summary so a captured `cargo test -- --ignored --nocapture` run produces the "short Matrix smoke result" log line the reviewer asked for.

Scope still single-file: the new test is inside the existing `#[cfg(test)] mod tests { mod live_smoke { ... } }` block; no production code changed since the original review.

```
```

- `cargo check -p zeroclaw-channels` against current `upstream/master`:
  ```
  Checking zeroclaw-channels v0.8.0-beta-2 (…/crates/zeroclaw-channels)
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 44.52s
  ```

- Branch is based on `upstream/master` post-#7231 (the `ollama.rs` E0308 revert that was breaking CI on #7119 has landed), so CI on this branch should be green rather than inheriting the prior master breaker.

- Smoke test commit (`cb2ccc3a`) validation locally:
  ```
  cargo fmt   -p zeroclaw-channels --check                                          # clean
  cargo clippy -p zeroclaw-channels --tests --features channel-matrix -- -D warnings # clean
  cargo test  -p zeroclaw-channels --lib  --features channel-matrix matrix::
    # test result: ok. 121 passed; 0 failed; 2 ignored (was 1 — new live smoke is the +1)
  cargo test  -p zeroclaw-channels --lib  --features channel-matrix -- --list --ignored | grep matrix
    matrix::tests::live_smoke::idle_sync_does_not_error_at_30s_cadence: test
    matrix::tests::live_smoke::same_room_partial_draft_lifecycle_uses_real_draft_ids: test
  ```

Confirmed the two constants are ordered correctly (`SYNC_LONGPOLL_TIMEOUT = 30s < CLIENT_REQUEST_TIMEOUT = 60s`) so the homeserver-side long-poll always completes before the HTTP deadline. The `?timeout=` parameter is now sent on every sync call (both `sync_once` for the initial boot and the long-running `sync` loop), and the integration test in the file (`matrix.rs` tests module) was updated to use the same long-poll timeout so it exercises the same path as production.

Live-Matrix soak was run against `matrix.todixuclawbot` after the fix: idle sync loop started cleanly, no 30s-cadence errors observed, end-to-end inbound message through Element X confirmed in the comments. The new `idle_sync_does_not_error_at_30s_cadence` smoke codifies that observation as a runnable test (`#[ignore]`'d so it stays out of CI by default).

**If any command was intentionally skipped, why:** Full `cargo test` on the channels crate was not run in this prep pass; I'd appreciate CI confirming green before merge.

- New external network calls? **No** (changes timeout parameters on existing Matrix `/sync` calls)

## Compatibility (required)

- Backward compatible? **Yes** — only changes timeout values on existing call sites; no API or config surface change.
- Config / env / CLI surface changed? **No** (the new smoke test reads `ZEROCLAW_MATRIX_SMOKE_IDLE_SECS` / `ZEROCLAW_MATRIX_SMOKE_MIN_LONGPOLL_MS` from env, but these are test-only opt-ins gated behind `#[ignore]`; no production code path reads them.)

## Rollback (required for risk: medium and risk: high)

- **Fast rollback command/path:** `git revert <commit-sha-of-this-PR>` (revert the runtime commit; the test-only commit `cb2ccc3a` is safe to leave in place or revert independently)
- **Observable failure symptoms:** After revert, idle Matrix sessions will resume the pre-fix behavior: `/sync` requests erroring out at exactly 30 seconds and the SDK busy-polling. Look for repeated `WARN`/`ERROR` entries from `crates/zeroclaw-channels/src/matrix.rs` around the `sync`/`sync_once` paths with a ~30s cadence.


Audacity88 (changes requested):
I reviewed current head 8215ed7, the PR body, the one-file `matrix.rs` diff, the prior #7119 split review and closeout comment, the empty current comment/review threads, and the now-green visible CI. I did not rerun local cargo or run a live Matrix channel soak.

@tidux Thanks for splitting this out from #7119. The new PR is much easier to review: it only carries the Matrix `/sync` timeout change, and the scope/body now match the one-file Matrix diff.

The code direction looks right. I am not comfortable approving it yet because the PR changes live channel runtime behavior and the central claim still needs live or near-live Matrix evidence.

### ✅ Resolved — The #7119 mixed-scope problem is fixed

This PR no longer bundles the runtime `trim_history` fix with the Matrix channel fix. The changed file list is just `crates/zeroclaw-channels/src/matrix.rs`, and the PR body now describes the Matrix timeout behavior, Matrix-only blast radius, and the split from #7119 clearly.

### 🟢 What looks good — The timeout relationship is explicit

Setting a 60-second Matrix client request timeout and a 30-second `/sync` long-poll timeout is the right shape for the described failure. The comments also state the important invariant: the server-side long-poll window must stay below the HTTP client deadline, so idle syncs can return normally before the request layer times out.

### 🔴 Blocking — Add live or near-live Matrix idle-sync evidence

This PR's central behavior is not just that the code compiles. It is that an idle Matrix `/sync` request no longer hits the 30-second HTTP timeout / busy-poll pattern. The PR body says a full live Matrix soak was not run.

For channel runtime behavior, I need live or near-live evidence before approval unless we explicitly accept an exception. Please add a short Matrix smoke result that leaves a configured Matrix channel idle for longer than 30 seconds and shows that `/sync` no longer errors at the 30-second cadence. A concise log excerpt or run note is enough; it does not need to be a large soak. If live Matrix validation is impractical, say so explicitly and we can decide whether to accept the risk, but the current PR evidence stops at compile/CI plus timeout math.

tidux:
this message sent by todixuclawbot confirming Matrix end to end connectivity.

tidux:
Screenshot_20260608_144047_Element X

tidux:
Prior sync misbehavior has been corrected as demonstrated in ZeroClaw daemon logs.  Test coming.

```
Timestamp  2026-06-08T21:32:33.852Z
Severity   INF (9)
Category   channel
Action     note

Message
matrix: starting sync loop

Attribution
channel     matrix.todixuclawbot
channel_aliastodixuclawbot
channel_typematrix
zc_role     channel

Attributes
{
  "_file": "crates/zeroclaw-channels/src/matrix.rs",
  "_line": 1545
}
```

tidux:
@Audacity88 thanks for the clear blocker — addressed in `cb2ccc3a`.

**Smoke test added.** New `matrix::tests::live_smoke::idle_sync_does_not_error_at_30s_cadence` in `crates/zeroclaw-channels/src/matrix.rs`, joining the existing `#[ignore]`'d live-smoke lane (same `ZEROCLAW_MATRIX_SMOKE_*` env contract as the sibling draft-lifecycle test, so it stays out of CI by default and is opt-in locally).

It does exactly the loop you asked for:

- builds a `MatrixChannel` from env creds and calls `ensure_client()` — this exercises the new `CLIENT_REQUEST_TIMEOUT` on `Client::builder()`'s `RequestConfig`,
- primes the sync token with one bounded `sync_once`,
- then loops `sync_once(SyncSettings::default().timeout(SYNC_LONGPOLL_TIMEOUT))` for `> 30s` of wall-time (default 35s, tunable via `ZEROCLAW_MATRIX_SMOKE_IDLE_SECS`) against an otherwise-idle room,
- asserts **no `sync_once` call returns an error** — that's the primary anti-regression for the 30s HTTP-deadline bug,
- and as defense-in-depth asserts that at least one round-trip durably long-polls (≥ `ZEROCLAW_MATRIX_SMOKE_MIN_LONGPOLL_MS`, default 1s) and that sub-threshold returns stay inside a small budget, so a regression that reintroduces the busy-poll pattern also trips the test.

It also emits a one-line `eprintln!` summary so a captured `cargo test -- --ignored --nocapture` run produces the "short Matrix smoke result" log line you asked for.

**Near-live evidence** is already in the previous comments — daemon log excerpt showing the sync loop entering cleanly with no 30s-cadence errors, plus the Element X end-to-end screenshot.

**Local validation of the test commit:**

```
cargo fmt   -p zeroclaw-channels --check                                          # clean
cargo clippy -p zeroclaw-channels --tests --features channel-matrix -- -D warnings # clean
cargo test  -p zeroclaw-channels --lib  --features channel-matrix matrix::
  # test result: ok. 121 passed; 0 failed; 2 ignored (was 1 — new live smoke is the +1)
cargo test  -p zeroclaw-channels --lib  --features channel-matrix -- --list --ignored | grep matrix
  matrix::tests::live_smoke::idle_sync_does_not_error_at_30s_cadence: test
  matrix::tests::live_smoke::same_room_partial_draft_lifecycle_uses_real_draft_ids: test
```

Scope is still single-file (the new test sits inside the existing `#[cfg(test)] mod tests { mod live_smoke { ... } }` block in `matrix.rs`); no production code changed since your review. PR body updated to reflect the new evidence and to drop the now-stale "happy to add if reviewers want" note.

Audacity88 (approved):
I reviewed current head aeef551a, the updated PR body, the one-file `matrix.rs` diff, the follow-up top-level comments with Matrix daemon evidence, the new ignored live-smoke test, the empty current inline review threads, the prior CHANGES_REQUESTED review, and the now-green visible CI. I did not rerun local cargo or run the live Matrix smoke myself.

@tidux Thanks for the fast follow-up. This addresses the blocker I left on 8215ed7.

### ✅ Resolved — Matrix idle-sync evidence now covers the 30-second failure mode

The earlier blocker was that the PR proved the timeout math and compile path, but not the live Matrix behavior. The updated branch now adds `matrix::tests::live_smoke::idle_sync_does_not_error_at_30s_cadence`, and the test exercises the same client construction path that applies `CLIENT_REQUEST_TIMEOUT`, then repeatedly calls `sync_once(SyncSettings::default().timeout(SYNC_LONGPOLL_TIMEOUT))` for a soak window longer than 30 seconds.

That gives us a runnable near-live check for both halves of the original bug: no request-deadline error at the old 30-second cadence, and no return to the busy-poll pattern where idle `/sync` calls come back immediately because no `?timeout=` is sent.

### 🟢 What looks good — The production fix stays small and targeted

The production diff is still just the Matrix channel path. `Client::builder()` now gets a 60-second request timeout, while the initial `sync_once` and long-running `sync` calls share the explicit 30-second long-poll setting. The constants document the important invariant directly at the two call sites: the server-side long-poll window must stay below the HTTP request deadline.

Approving this now. The remaining risk is ordinary Matrix live-environment variance, and the new ignored smoke gives us the right opt-in tool to re-check that path when credentials are available.

PR #6719, opened 2026-05-16
fix(runtime): persist model_switch across turn paths

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:** The `model_switch` tool sets a process-wide `MODEL_SWITCH_REQUEST` flag. The channel orchestrator consumed it for the *current* turn but never wrote it back via `set_route_selection()`, so the next inbound message reverted to the default model; and the gateway/UI streaming path never observed the flag, making the tool a complete no-op there (#6173).
  - **Channel path** (`orchestrator/mod.rs`): the `ModelSwitchRequested` handler now resolves the requested provider to a configured `<type>.<alias>` ref (`resolve_provider_ref_for_runtime_switch`), resolves a route-specific `api_key` from `ctx.model_routes` (the same lookup the `/model` command uses), builds through `get_or_create_provider()` so the switched provider lands in the cache with the right credentials, commits the local `route` only after the provider builds, and **persists** the resolved route via `set_route_selection()` so later messages from the same sender keep the switched model.
  - **Gateway/UI path** (`agent.rs`): wired through the unified turn loop. After #7540 the streaming wrapper `turn_streamed_with_steering_state` is thin and the engine lives in `run_tool_call_loop`, which already *signals* a pending switch by returning `ModelSwitchRequested` (without clearing it) — but the streaming wrapper passed `None` for `model_switch_callback`, so it was never observed (the CLI/non-streaming path already rebuilt on this signal). This PR enables the callback on the streaming path and handles `ModelSwitchRequested` in the wrapper's round loop: it rebuilds the provider from `ProviderSwitchConfig`'s shared `Config` via `try_apply_pending_model_switch`, resolving the provider entry, route-specific key, reliability, model routes, and provider runtime options at switch-consumption time; commits agent state only after the rebuild succeeds, and continues the round loop on the switched provider/model. Wiring it at the loop boundary gives the fix to every streaming transport (WebSocket, ACP, REST) at once.
- **Scope boundary:** Does NOT change the `model_switch` tool, the `MODEL_SWITCH_REQUEST` state shape, `run_tool_call_loop`'s switch *detection* (introduced by #7540), or the unrelated CLI/non-streaming switch handling that already rebuilt the provider. No new default-model assumptions.
[...]

Issue #6557, opened 2026-05-10
[Feature]: Reconcile runtime model switching with provider structure for v0.8.0


Audacity88:
### Summary

Reconcile runtime model switching with the newer provider structure before v0.8.0 hardens provider/config semantics.

### Problem statement

PR #6541 surfaced that model-switch behavior now spans several surfaces: channel slash commands such as `/models`, `/model`, and `/config`, the runtime `model_switch` tool, provider/model reconstruction after `is_model_switch_requested`, and the newer provider-family / qualified-provider work tracked in #6273.

The current behavior is easy to reason about in each local path, but there is not yet one documented contract for what a "model switch" means across channel turns and direct agent turns. In particular, v0.8.0 should pin down whether a switch targets only a model name, a provider/model pair, a provider profile, or a typed provider-family entry, and how that choice interacts with persisted config and transient session state.

Without that reconciliation, the implementation can drift into subtly different behavior depending on whether a user switches from a channel command, a runtime tool call, or config-backed provider selection.

Related: #6541
Related: #6273
Related: #6030

### Proposed solution

Define and implement a v0.8.0 reconciliation path for model switching that covers:

- The canonical switch shape: model only, provider/model tuple, provider profile, or typed provider-family entry.
- The allowed callers: channel slash commands, runtime tool calls, config-driven startup, and any future UI/gateway surfaces.
- The lifetime of a switch: transient turn/session/channel scope versus persisted config update.
- The fallback/error behavior when the requested provider/model pair cannot be constructed.
- How provider-local defaults and qualified provider names interact with switch requests.
- Test coverage that proves the same intended switch resolves consistently in direct agent turns and channel turns.

The outcome can be a small design note plus implementation PR(s), but the main goal is to prevent v0.8.0 from freezing multiple incompatible model-switch semantics.

### Recommended direction

Recommendation: make a provider profile the stable switch target, and treat the model as provider-local state inside that profile.

Concretely:

- Keep the resolved runtime shape as an explicit provider/model pair, not a free-floating model string.
- Treat `/models <provider>` as selecting the active provider profile for the sender/session.
- Treat `/model <model-id>` as selecting a model within the active provider profile.
- Treat `model_switch` tool requests as transient runtime/session switches unless they go through an explicit config-write path.
- Consolidate provider variants that only differ by endpoint or compatibility flavor into one provider family/profile shape where possible, rather than keeping long-term parallel provider identities for each endpoint variant.
[...]

PR #7407, opened 2026-06-08
fix(runtime): require configured model switch profiles

Audacity88:
## Summary

- **What changed and why:**
  - Resolve `model_switch` `model_provider` values against configured `providers.models.<type>.<alias>` entries before mutating pending runtime switch state.
  - Reject bare provider families and unconfigured aliases with output that names the dotted provider-profile shape and lists configured profiles.
  - Keep custom provider profiles valid when they are configured, instead of tying runtime switching only to the static provider catalog.
  - Update provider routing docs and schema comments so model routes, embedding routes, and runtime tool switching use provider profiles plus provider-local model ids.
- **Scope boundary:** This PR does not persist model switches across turns, change channel `/models` parsing, change gateway or web UI model switching, or replace the broader model-switch persistence work in #6719.
- **Blast radius:** Runtime `model_switch` tool calls, the tool registry constructor wiring, provider routing docs, and config schema documentation comments/examples. Channel `/models <type>` shorthand remains documented as a channel-runtime behavior when exactly one alias exists.
- **Linked issue(s):**
  - Related #6557
  - Related #6719
- **Labels:** enhancement / risk: high / size: M / runtime / config / docs

  - `cargo test -p zeroclaw-runtime model_switch::tests --lib -- --nocapture` — passed. Tail: `test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 2022 filtered out`.
  - `cargo fmt --all` — passed.
  - `cargo check -p zeroclaw-config` — passed. Tail: `Finished dev profile`.
  - `cargo check -p zeroclaw-runtime` — passed. Tail: `Finished dev profile`.
  - `cargo fmt --all -- --check` — passed after the lint-ordering follow-up commit.
  - `git diff --check` — passed after the lint-ordering follow-up commit.
  - `cargo test -p zeroclaw-runtime model_switch::tests --lib -- --nocapture` — passed after the lint-ordering follow-up commit. Tail: `test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 2022 filtered out`.
  - `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` — passed after the lint-ordering follow-up commit.
  - `cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings` — passed. Tail: `Finished dev profile [unoptimized + debuginfo] target(s) in 32m 49s`.
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` — passed. Tail: `Summary [125.240s] 8146 tests run: 8146 passed, 10 skipped`.
  - `git diff --check origin/master...HEAD` — passed with no output.
- **Beyond CI — what did you manually verify?** Focused tests cover rejecting bare provider families, accepting configured dotted profiles including `custom.local`, rejecting unconfigured dotted profiles without mutating pending state, and accepting dotted profile refs for `list_models`. The lint-ordering follow-up was covered by focused fmt, diff-check, model-switch tests, package-scoped clippy, workspace clippy, and the workspace nextest command. Checked that the provider routing docs still document the channel `/models <type>` bare-family shorthand as a channel-runtime behavior when exactly one alias exists.
- **If any command was intentionally skipped, why:** None. Workspace clippy and workspace nextest both passed locally. Visible GitHub checks also passed, including Lint, Test, Security, Format, platform builds, Check (32-bit), Check (all features), Check (no default features), Benchmarks Compile, Nix Module Eval, path labels, and CI Required Gate.

- New permissions, capabilities, or file system access scope? `Yes` — the runtime `model_switch` tool now resolves and accepts configured provider profiles, including custom aliases. It does not add filesystem scope, new external calls in this path, or credential-value output.
- If any `Yes`, describe the risk and mitigation: Risk is limited to runtime switch-target selection. Mitigation is resolving only configured provider profiles and not emitting credential values or adding network/filesystem scope.

## Compatibility (required)

- Backward compatible? `No` for tool callers that passed a bare provider family such as `openai` to `model_switch`; those calls now need a configured dotted profile ref such as `openai.default`.
- Config / env / CLI surface changed? `Yes` for the runtime `model_switch` tool input contract and documentation examples. No persisted config schema field is added.
- If `No` or `Yes` to either: existing tool callers should pass `<type>.<alias>` values that resolve under `[providers.models.<type>.<alias>]`. Channel `/models <type>` can still use the existing unambiguous shorthand where the channel runtime supports it.

- **Fast rollback command/path:** `git revert <merge-sha>` for this PR's merge commit after merge if this contract breaks expected runtime switching.
- **Observable failure symptoms:** `model_switch` tool calls that should resolve a configured profile return errors such as `not a configured provider profile`, or runtime model switching stops setting pending switch state for a configured `<type>.<alias>` ref.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR scope or contributor-authored patch was carried forward.


crates/zeroclaw-runtime/src/tools/model_switch.rs
```diff
@@ -5,14 +5,57 @@ use async_trait::async_trait;
...
+
+fn configured_model_provider_profiles(config: &Config) -> Vec<String> {
+    let mut profiles = config
+        .providers
+        .models
+        .iter_entries()
+        .map(|(family, alias, _profile)| format!("{family}.{alias}"))
+        .collect::<Vec<_>>();
+    profiles.sort();
+    profiles
```
singlerider, line 18: 🟢 The dotted-ref requirement here is the right enforcement point — it matches what `api_key_and_uri_for_provider` and `options_for_provider_ref` already resolve downstream, so the switch target and the resolved credential/endpoint can no longer diverge. Worth a quick grep of any prompt/docs/skill content that tells the model to call `model_switch` with a bare family like `openai`; those calls now error (self-correcting via the message, not a blocker).

singlerider (approved):
I read the full diff, the runtime switch-state consumer in `agent/loop_.rs`, and the provider-ref resolution helpers in `zeroclaw-providers`. Clean, well-scoped, well-tested. Approving.

### 🟢 What looks good — the contract now matches the consumer

The value here is more than a contract tightening. The runtime switch-state consumer already resolves dotted `<type>.<alias>` refs: `api_key_and_uri_for_provider` (loop_.rs:3088) does `split_once('.')` then `config.providers.models.find(fam, al)`, and `options_for_provider_ref` (providers/lib.rs:782) does the same. Before this PR, a bare family written into switch state fell through to the *fallback* credential/URI/options branch silently — a latent mismatch where the switch target and the resolved endpoint could diverge. Requiring the dotted profile ref at the tool boundary closes that gap, and the error path naming the dotted shape plus listing configured profiles is exactly the actionable feedback a caller needs.

### 🟢 What looks good — test coverage maps to the behavior change

The five focused tests cover the real decision points: bare-family rejection without mutating pending state, dotted accept, unconfigured-alias rejection without mutating pending state, the `custom.local` configured-alias accept (the case the old static-catalog check would have mishandled), and `list_models` resolving a dotted ref. The `MODEL_SWITCH_TEST_LOCK` guard around the process-global switch state is the right call.

### 🟢 What looks good — list_models family fallthrough

Splitting the resolved ref back to its family before the common-models match, and dropping the `"google" | "gemini"` / `"xai" | "grok"` string aliases down to the canonical family names, keeps the lookup aligned with the provider registry rather than carrying its own alias table.

Approving. The change is correct, tested, documented, and tightens a real latent inconsistency.

Issue #6377, opened 2026-05-04
[Bug]: Onboarding/Providers: Llama.cpp should default to "responses"


tidux:
### Affected component

provider

### Severity

S1 - workflow blocked

### Current behavior

Llama.cpp throws 500s when a ZeroClaw user's prompt results in tool usage.

### Expected behavior

Llama.cpp responds correctly and ZeroClaw executes the tools properly.

### Steps to reproduce

```bash
$ zeroclaw onboard # configure llama.cpp server as a provider
$ zeroclaw agent -m "What's the weather like in McMurdo Station today?"
```

### Impact

The current `llamacpp` provider defaults to the original OpenAPI `chat_completions` wire protocol, but Llama.cpp's implementation of tool use on that protocol is currently broken.  Instead it should default to the OpenResponses protocol (https://www.openresponses.org/) via `wire_api = "responses"` as discussed in this Llama.cpp issue: https://github.com/ggml-org/llama.cpp/pull/21174.  Neither is fully complete yet but "responses" will be more compatible with ZeroClaw going forward.

### Logs / stack traces

```text
May 04 16:33:45 <ZEROCLAW_HOSTNAME> zeroclaw[168935]: 2026-05-04T23:33:45.402922Z  WARN zeroclaw_providers::reliable: Streaming error: HTTP error: error sending request for url (http://<LLAMACPP_HOSTNAME>:8080/v1/chat/completions) provider="llamacpp" model="unsloth/Qwen3.6-35B-A3B-GGUF:MXFP4_MOE"
May 04 16:33:45 <ZEROCLAW_HOSTNAME> zeroclaw[168935]: 2026-05-04T23:33:45.457460Z  WARN zeroclaw_providers::reliable: Streaming error: HTTP error: error sending request for url (http://<LLAMACPP_HOSTNAME>:8080/v1/chat/completions) provider="llamacpp" model="unsloth/Qwen3.6-35B-A3B-GGUF:MXFP4_MOE"
```

### ZeroClaw version

91cf02b8874b0c59ef96457f2b47ba59a824aae4

### Rust version

1.93.0

### Operating system

MacOS 26

### Regression?

Unknown

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

Audacity88:
[...]

PR #7418, opened 2026-06-09
fix(providers): honor wire_api = "responses" for OpenCode Zen

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The blanket `CompatFamilySpec` factory impl built a chat-completions
    `OpenAiCompatibleModelProvider` and never read `wire_api`, so an OpenCode
    Zen entry with `wire_api = "responses"` was silently routed through the
    chat-completions wire. The setting parsed into `base.wire_api` and was then
    dropped end-to-end.
  - Added a defaulted `wire_api()` accessor to `CompatFamilySpec` (returns
    `None`); the blanket `create_provider` now short-circuits through
    `build_responses_provider_if_requested` when it returns `Some(Responses)`.
  - Only `OpencodeModelProviderConfig` overrides `wire_api()` to surface
    `self.base.wire_api`, so behavior is scoped to OpenCode; every other compat
    family keeps the chat-completions default.
  - The responses path falls back to `Self::DEFAULT_URL` when no api_url is
    resolved, mirroring `build_compat_base`. Without this, OpenCode (which has
    no `endpoint_uri()` override and resolves to a `None` api_url) would default
    to `https://api.openai.com/v1/responses` and send OpenCode-credentialed
    traffic to the wrong host.
- **Scope boundary:** Does NOT change any other provider family's routing. The
  defaulted `wire_api()` returns `None` for all families except OpenCode, so no
  other compat family's behavior moves. A family whose endpoint can serve the
  responses wire must opt in by overriding `wire_api()` (documented on the trait
  method); this keeps the change per-family rather than flipping a global
  default. Does not add responses support to the compat client itself; it routes
  to the existing `OpenAiResponsesModelProvider`.
- **Blast radius:** Limited to the OpenCode Zen family. Verified zero overlap
  between bespoke `FamilyProviderFactory` impls and `CompatFamilySpec` impls, so
  families with their own responses handling (openai, custom, etc.) resolve to
  their concrete impl and never touch the blanket impl changed here.
- **Linked issue(s):** Related #6377 — that issue established `wire_api =
  "responses"` as a supported, needed default for a compat family (llama.cpp),
  fixed by giving llama.cpp a bespoke `FamilyProviderFactory` impl with the
  responses branch. OpenCode uses the blanket `CompatFamilySpec` impl and was
  left without that branch; this PR closes that gap on the blanket path so the
  setting takes effect for OpenCode too. This PR does not claim to fix any
  gateway/agent-accessibility symptom.
- **Labels:** `bug`, `provider`, `risk: medium`, `size: S`.

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-providers
```

  - `cargo fmt --all -- --check` — clean (exit 0).
  - `cargo clippy --all-targets -- -D warnings` —
    `Finished dev profile [unoptimized + debuginfo] target(s) in 1m 28s` (no
    warnings, no errors across the workspace).
  - `cargo test -p zeroclaw-providers` —
    `test result: ok. 925 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out`.
- **Beyond CI — what did you manually verify:** The fix was developed
  test-first: the responses-routing test was confirmed red before the change
  (`left "chat_completions", right "responses"`). A mock-server test proves the
  factory posts to the configured base's `/responses` path, not OpenAI's host —
  this catches the credential-misroute regression that a wire-only assertion
  would miss. A scope-guard test (Doubao with `wire_api = "responses"` set)
  asserts that a compat family which does NOT override `wire_api()` keeps
  chat_completions, pinning the per-family scope boundary so a future generic
  change cannot silently widen it. Also confirmed the model picker for a
  responses-wire OpenCode entry falls back to free-text input (the responses
  provider uses the trait-default `list_models` that bails), which is identical
  to the existing custom/openai responses behavior and not a regression.
- **If any command was intentionally skipped, why:** Full `cargo test`
  (workspace) running separately; the change is isolated to `zeroclaw-providers`
  which passes fully. Will update if the workspace run surfaces anything.

- New external network calls? `No` (routes existing traffic to the correctly
  configured endpoint rather than a new one).
- Secrets / tokens / credentials handling changed? `Yes` — this fix prevents an
  OpenCode API key from being sent to OpenAI's endpoint. Before the
  `DEFAULT_URL` fallback, a responses-wire OpenCode entry with no api_url
  override would have posted to `https://api.openai.com/v1/responses` carrying
  the OpenCode credential. The fix pins the request to the configured OpenCode
  base.
- PII, real identities, or personal data in diff, tests, fixtures, or docs?
  `No`.
- If any `Yes`, describe the risk and mitigation: The credential-misroute risk
  is mitigated by the `Self::DEFAULT_URL` fallback and pinned by the
  mock-server test asserting the request path.

## Compatibility (required)

- Backward compatible? `Yes` — families that did not set
  `wire_api = "responses"` are unaffected; the accessor defaults to `None`.
- Config / env / CLI surface changed? `No` — `wire_api` is an existing config
  field; this PR makes the existing value take effect for OpenCode instead of
  being silently dropped.
- Upgrade steps for existing users: None required. An OpenCode entry that
  already set `wire_api = "responses"` will now correctly use the responses wire
  against the OpenCode endpoint.

- **Feature flags or config toggles:** None. Reverting restores the prior
  (chat-completions-only) compat behavior for OpenCode.
- **Observable failure symptoms:** If the responses routing were wrong, requests
  for a responses-wire OpenCode entry would 404/400 against the OpenCode base or
  hit OpenAI's host. Grep the runtime trace for `model_provider_type` and the
  request URL on an OpenCode session.


Audacity88 (dismissed):
Context: I reviewed the live PR state, PR body, checks, full diff, and the touched provider factory/OpenAI responses paths. There are no prior review comments or formal reviews on this PR. I also checked the current provider docs for the existing `wire_api = "responses"` contract.

### 🟢 What looks good — The routing shape stays narrow

The implementation is otherwise pointed in the right direction. Adding `wire_api()` as an opt-in accessor on `CompatFamilySpec` keeps the blanket factory from globally changing every compatible vendor slot, and overriding it only on `OpencodeModelProviderConfig` matches the PR's scope boundary. Reusing `OpenAiResponsesModelProvider` also avoids a second responses implementation, while the Doubao guard test is a useful check that unrelated compat families still stay on chat completions even if their base config carries `wire_api = "responses"`.

### 🔴 Blocking — Pin the no-override OpenCode fallback that protects credentials

The implementation appears to route the new OpenCode responses path through the right default base URL: the blanket compat factory passes `api_url.or(Some(Self::DEFAULT_URL))` before constructing `OpenAiResponsesModelProvider`, so an OpenCode entry without an explicit `uri` should use `https://opencode.ai/zen/v1/responses` instead of falling back to OpenAI's `/v1/responses` endpoint.

The regression coverage does not currently pin that security/privacy-critical fallback. `compat_family_factory_routes_to_responses_when_wire_api_responses` does pass `None` for `api_url`, but it only checks `default_wire_api() == "responses"`. That assertion cannot distinguish "responses provider built with OpenCode's default base URL" from "responses provider built with no base URL and therefore defaulted to OpenAI." `compat_family_responses_posts_to_responses_path_on_configured_base` proves the request path when the test passes an explicit `Some(&base_url)`, but it does not exercise the exact no-`api_url` request path this PR calls out in the Security & Privacy section.

Please add a regression that would fail if the blanket compat route stopped supplying OpenCode's `DEFAULT_URL` to the responses provider. The important assertion is not just that the provider uses the responses wire; it is that a default OpenCode responses config does not send OpenCode credentials to OpenAI's host when no alias-level URL override is present.

### 🔴 Blocking — Update the user-facing `wire_api` docs for OpenCode

This PR intentionally changes the documented behavior of a provider slot: `opencode` now honors `wire_api = "responses"` through the blanket `CompatFamilySpec` path.

The current docs still say `wire_api` is only honored by `openai`, `llamacpp`, `custom`, and the generic openai-compatible path, and that branded vendor slots ignore it. After this PR, that statement is no longer true for OpenCode Zen. Please update the provider docs by either adding `opencode` to the list of slots that honor `wire_api = "responses"` or by qualifying the "branded vendor slots ignore the field" sentence with an explicit OpenCode exception, plus a short OpenCode example if that is the clearest user-facing shape.

Audacity88 (approved):
Context: I re-reviewed head `1c9751747f5a696827b91e3917ed41f68928a113` after my prior request-changes review. I checked the live PR state, prior review, current diff, the provider factory / OpenAI responses paths, the provider docs update, and the green GitHub checks. I did not rerun local Cargo on this head; the relevant provider coverage is in the PR and the official `Test`, `Lint`, and `CI Required Gate` jobs are green.

### ✅ Resolved — The no-override OpenCode responses fallback is pinned

The new assertion on the default OpenCode responses provider now checks `default_base_url() == Some("https://opencode.ai/zen/v1/responses")`, so this would fail if the blanket compat path stopped passing OpenCode's `DEFAULT_URL` and the provider fell back to OpenAI's `/v1/responses` host.

Changing `OpenAiResponsesModelProvider::default_base_url()` to report the resolved `responses_url` also makes the fallback assertion inspect the endpoint this provider is configured to use, rather than a constant.

### ✅ Resolved — The user-facing `wire_api` docs now include OpenCode

The provider docs now say branded vendor slots ignore `wire_api` except for `opencode`, explain why OpenCode honors `wire_api = "responses"`, and include a small OpenCode config example. That fixes the stale user-facing contract from the prior review.

### 🟢 What looks good — The scope guard remains narrow

The Doubao guard still proves that compat families which do not override `wire_api()` stay on chat completions even if their base config carries `wire_api = "responses"`. That keeps this fix scoped to OpenCode instead of turning `wire_api` into a global compat-family behavior change.

Approving. This clears my prior blockers. Merge-readiness still depends on the current GitHub mergeability state settling and any remaining requested reviewer feedback.

Issue #6127, opened 2026-04-26
gateway: silent-fallback hardening in crates/zeroclaw-gateway/src/lib.rs (follow-up to #6099)


perlowja:
Tracking issue per @WareWolf-MoonWall's merge-condition on #6099.

## What's still silent after #6099

#6099 (\`fix(config): preserve user-supplied providers.fallback through load/save\`) hardens the runtime-side credential resolution with the fail-loud / sentinel split — but the gateway-side resolution path in \`crates/zeroclaw-gateway/src/lib.rs\` still falls back silently when a profile-resolved credential is unavailable. Same problem space as the runtime fix; different surface.

This is the merge-condition WareWolf-MoonWall flagged on #6099: gateway-side hardening must be tracked before the runtime-side fix lands so the asymmetric remediation isn't forgotten.

## Fix shape

Mirror the runtime-side pattern from #6099:

1. **Fail-loud**: when a fallback provider is requested but no profile-resolved credentials exist, return a structured error to the gateway caller rather than silently using an empty/sentinel credential.
2. **Sentinel split**: distinguish "no credential intentionally" (gateway runs anonymously by design) from \"credential resolution failed\" (operator misconfiguration).
3. **Operator-actionable error**: include the provider name, the resolved profile, and the credential field that failed to resolve.

## Scope

- File: \`crates/zeroclaw-gateway/src/lib.rs\` — the credential resolution call sites adjacent to \`fallback_providers\` lookup.
- Channels-side similar paths can travel with this PR (per @WareWolf-MoonWall's \"channels sites can travel with it\" note).

## Cross-references

- Runtime-side fix: #6099
- Adjacent issue: #6092 (\`[reliability].fallback_providers\` chain entries not picking up profile-resolved credentials)

cc @WareWolf-MoonWall @singlerider — opening this so #6099 can merge cleanly.

Audacity88:
## Triage Summary

**Type:** Enhancement
**Priority:** Medium
**Area:** gateway, security
**Risk:** Medium

**Summary:** Gateway-side fallback credential resolution silently uses empty/sentinel credentials when a profile-resolved credential is unavailable. Mirrors the runtime-side fix from #6099 to the gateway surface. Filed as a merge-condition from WareWolf-MoonWall's review on #6099.

**Duplicates/Related:** Follow-up from #6099. Same problem space as #6092 (fallback config resolution).

[...]

PR #7419, opened 2026-06-09
fix(providers): fail loudly for unusable fallback providers

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Makes resolved fallback providers fail loudly when their profile cannot supply the credential or external auth source their provider family requires.
  - Keeps dangling fallback refs, fallback cycles, and over-depth chains non-fatal because existing config warnings report those configuration problems.
  - Makes routed provider construction failures abort route setup instead of silently dropping a configured routed provider.
  - Prevents non-primary routed aliases from borrowing the primary provider API key; they now use the route key or their own typed profile key.
- **Scope boundary:** This does not change config schema, fallback reference parsing, gateway routes, or provider request behavior after a provider is successfully constructed.
- **Blast radius:** Provider fallback-chain construction and routed provider setup. Gateway startup can now fail earlier when a configured fallback or routed provider is unusable instead of running with a silently missing backup/route.
- **Linked issue(s):** Related #6127 — this hardens provider-side fallback and routed-provider construction; #6127 remains open for gateway boot seed-provider behavior.
- **Labels:** `bug`, `risk: high`, `size: L`, `provider`, `security`

```text
cargo fmt --all -- --check
Passed.

git diff --check
Passed.

cargo test -p zeroclaw-providers routed_primary_alias_kind_does_not_leak_to_canonical_route_provider -- --nocapture
Passed: 1 test passed.

cargo test -p zeroclaw-providers routed_model_provider -- --nocapture
Passed: 5 tests passed.

cargo test -p zeroclaw-providers resilient_alias -- --nocapture
Passed: 10 tests passed.

cargo check -p zeroclaw-providers -p zeroclaw-gateway
Passed in 10m31s.
Warning observed from an unrelated channel import: unused import: crate::paced_channel::PacedChannel.

cargo test -p zeroclaw-gateway run_gateway_starts_with_mismatched_provider_api_key -- --nocapture
Passed: 1 test passed.
The same unrelated PacedChannel warning appeared during compile.

Passed before the final refresh: 8148 tests run, 8148 passed, 10 skipped, 10 slow.

Passed before the final refresh in 14m03s.
```
  - Checked that unresolved fallback refs, fallback cycles, and chains beyond the maximum depth still remain non-fatal.
  - Checked that resolved OpenAI fallbacks without profile credentials fail with the fallback alias and profile path in the error.
  - Checked that local or external-auth provider families that can work without a profile API key are still accepted.
  - Checked route credential precedence so a non-primary routed provider does not inherit the primary provider API key.
  - The broad workspace clippy and CI-shaped nextest commands were not rerun after the final refresh. The final refresh only incorporated docs/banner drift plus one-line rustdoc link updates in the touched provider files. Focused post-refresh provider and gateway checks passed on the refreshed branch.

  - This changes when provider credentials are accepted for fallback and routed-provider construction. It fails closed for resolved fallback/routed providers that lack a profile credential or provider-supported external auth source, and preserves allowed no-key local/external-auth families through provider-owned auth policy.

## Compatibility (required)

- Backward compatible? `Yes for usable provider configs; existing configs that relied on silently ignored unusable fallback or routed providers may now fail setup.`
  - Existing configs that named unusable fallback or routed providers may now fail during provider/gateway setup. Set the fallback provider profile's `api_key`, configure that alias's supported external auth flow, or remove the unusable fallback/route.

- **Fast rollback command/path:** `git revert 70e453e81`
- **Observable failure symptoms:** Gateway or provider startup returns an error naming the unusable fallback or routed provider alias, such as a missing profile credential for `[providers.models.<family>.<alias>]`.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR scope is carried forward.


singlerider (approved):
Read the diff against `factory.rs` / `lib.rs` on master, including the `for_each_model_provider_slot!` dispatch shape and the `requires_openai_auth` / OAuth config fields.

### 🟢 What looks good — the auth-readiness check is registry-driven

`fallback_auth_ready_for_alias` mirrors `dispatch_family_factory` exactly — same macro, same `openai-compatible` special case — so it cannot drift from the families the factory actually dispatches. New families inherit the default `has_api_key` check, which is the safe default: a keyless new family fails loud at startup instead of silently shipping a dead fallback, and the fix is a one-line trait override.

### 🟢 What looks good — keyless and external-auth families are each handled by their own contract

Local servers (Ollama, LM Studio, llama.cpp, vLLM, SGLang, LiteLLM, Osaurus, AtomicChat), ambient-credential families (Bedrock), CLI-auth families (Copilot, Gemini CLI, Kilo CLI), and custom endpoints return ready unconditionally; OAuth-capable families (Qwen, MiniMax, Gemini) check their refresh token / auth mode; OpenAI honors `requires_openai_auth` (Codex external auth store). That's the actual credential matrix, not a blanket exemption.

### 🟢 What looks good — the routed-provider key-borrowing fix

Non-primary routed aliases no longer inherit the primary's `api_key` via `.or(api_key)` — they use the route key or their own profile key. Silently authenticating provider B with provider A's credential was the kind of thing that works until it billing-surprises someone; the `is_primary.then_some(api_key).flatten()` shape makes the intent explicit.

### 🟢 What looks good — error messages name the fix

The bail text names the ref, the resolved alias, the exact TOML profile table, and the three remedies (set `api_key`, configure external auth, remove from `fallback`). Dangling refs / cycles / over-depth stay non-fatal because `collect_warnings` already owns those — the loud/quiet split follows who reports what, which the doc comment now states.

The startup-failure compatibility change is real (configs with silently-broken fallbacks now refuse to boot) but it's the entire point of the PR, disclosed in the body, and each failure comes with its own remediation text. Test coverage spans both the resilient-alias and routed paths, positive and negative. Approving.

PR #7422, opened 2026-06-09
fix(web_fetch): allow wildcard allowed_private_hosts to cover DNS-resolved private hosts (#7412)

### LINKED ISSUE #7412: [Bug]: web_fetch/http_request `allowed_private_hosts = ["*"]` does not cover domain names that resolve to private IPs
### Affected component

security/sandbox

### Severity

S2 - degraded behavior

### Current behavior

`web_fetch` (and `http_request`) reject a regular domain name whenever it *resolves* to a non-global IP, even when `allowed_private_hosts = ["*"]` is configured. Example:

```
Error: Blocked host 'internal.example.com' resolved to non-global address 198.19.66.3
```

(`198.19.66.3` is in the RFC 2544 benchmark range `198.18.0.0/15`, treated as non-global by `is_non_global_v4`.)

Root cause is the wildcard branch of `host_matches_private_allowlist` in `crates/zeroclaw-tools/src/web_fetch.rs`:

```rust
allowed_private_hosts.iter().any(|domain| {
    if domain == "*" {
        host_is_private_or_local           // <- gates "*" on the *literal* host
    } else {
        host == domain
            || host.strip_suffix(domain).is_some_and(|prefix| prefix.ends_with('.'))
    }
})
```

`host_is_private_or_local` is computed from `is_private_or_local_host(host)`, which only returns `true` for **IP literals**, `localhost`/`*.localhost`, and the `.local` TLD. For an ordinary domain name it returns `false`, so the `"*"` entry never matches, `private_host_allowed` stays `false`, and `validate_resolved_host_is_public` runs and blocks on the resolved private IP.

Net effect: a **specific** entry such as `allowed_private_hosts = ["internal.example.com"]` works (the `host == domain` branch skips the DNS check), but `["*"]` does **not** — the wildcard is narrower than a named entry, which is counter-intuitive.

This is closely related to #5122 (fixed by #6974), which made *named* private DNS hosts bypass the resolved-IP check. The #6974 close note states wildcard behavior was intentionally left to "keep the SSRF boundary intact", so this report is to (re)evaluate / at minimum document that asymmetry, since users reasonably expect `["*"]` to be a superset of any named entry.

### Expected behavior

One of:

- `allowed_private_hosts = ["*"]` behaves as a superset of any named entry — i.e. it also bypasses the resolved-IP public-address check for domain names (matching the named-entry semantics introduced by #6974), **or**
- If the narrow wildcard scope is intentional (per the #6974 close note), the limitation is clearly documented on the `allowed_private_hosts` config field and surfaced in the error message (e.g. "to allow a domain that resolves to a private IP, list the exact host in allowed_private_hosts; `*` only covers IP literals / localhost / .local").

Either way, the current behavior where `["*"]` silently fails to cover DNS-resolved private hosts is surprising and undocumented.

### Steps to reproduce

```bash
# config.toml
[web_fetch]
allowed_domains       = ["*"]
allowed_private_hosts = ["*"]

# Then fetch a public domain name that resolves into a non-global range
# (here 198.18.0.0/15, but 192.168/10/172.16 reproduce identically):
#   web_fetch  url = "https://internal.example.com/..."
#
# Observed:
#   Error: Blocked host 'internal.example.com' resolved to non-global address 198.19.66.3
#
# Workaround that DOES work (named entry):
[web_fetch]
allowed_private_hosts = ["internal.example.com"]
```

### Impact

Affected users: anyone pointing `web_fetch`/`http_request` at internal services reached by domain name (split-horizon DNS, corporate proxies that map public-looking names into RFC1918/RFC2544/RFC5737 ranges, home-lab/Home-Assistant setups).
Frequency: always, for any domain whose A record is a non-global IP.
Consequence: `["*"]` gives a false sense of "allow all internal", forcing users to enumerate every exact hostname instead.

### Logs / stack traces

```text
Error: Blocked host 'internal.example.com' resolved to non-global address 198.19.66.3
```

Code: `crates/zeroclaw-tools/src/web_fetch.rs`
- block site: `validate_resolved_ips_are_public` -> `anyhow::bail!("Blocked host '{host}' resolved to non-global address {ip}")`
- gate: `host_matches_private_allowlist` `"*"` branch returns `host_is_private_or_local`, which is `false` for non-IP, non-localhost/.local domains.

### ZeroClaw version

commit d23295118

### Rust version

rustc 1.94.1

### Operating system

Linux (Manjaro) 7.0.10 x86_64

### Regression?

No, first-time setup

### Pre-flight checks

chengzhichao-xydt:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  The wildcard entry `"*"` in `allowed_private_hosts` only matched hosts that passed `is_private_or_local_host()` (IP literals, `localhost`, `.local` TLD). Domain names that resolve to non-global addresses (e.g. `internal.example.com` → `198.19.66.3`) were rejected because the wildcard gate ran before DNS resolution could confirm the address is private.

  Thread `allowed_private_hosts` through `validate_resolved_host_is_public` into `validate_resolved_ips_are_public` so the allowlist (including the `"*"` wildcard) is consulted at the first point where we know the resolved address is genuinely non-global.
- **Scope boundary:** Only the validation path in `web_fetch.rs`. No change to `http_request` or any other tool.
- **Linked issue(s):** Fixes #7412

## Validation Evidence

**Real environment tested:** Rust 1.96.0, Linux, current `master` 24662b59.

```bash
cargo test -p zeroclaw-tools -- web_fetch
cargo clippy -p zeroclaw-tools --all-targets -- -D warnings
cargo fmt -- --check
```

$ cargo test -p zeroclaw-tools -- web_fetch
test result: ok. 62 passed; 0 failed; 0 ignored; 0 measured

Added test `private_allowlist_wildcard_allows_domain_resolving_to_private_ip` that directly reproduces the #7412 scenario with a DNS-resolved private address (198.19.66.3). All existing tests including `private_allowlist_wildcard_does_not_allow_public_domain_miss` continue to pass.

**Observed result after fix:** Domain names that resolve to private addresses are now correctly allowed when `"*"` is in `allowed_private_hosts`, while public domain mismatches still require `allowed_domains` membership.

**What was not tested:** End-to-end HTTP request to an actual DNS-resolved private host (requires network setup).

## Security and Privacy Impact

- Changes to key/token/credential handling? **No**
- PII, real identities, or personal information in diff, tests, fixtures, or docs? **No**

## Compatibility

Backward compatible. The wildcard `"*"` previously did not work for DNS-resolved domain names; it now does. No change to non-wildcard allowlist behavior.

### DIFF
--- a/crates/zeroclaw-tools/src/web_fetch.rs
+++ b/crates/zeroclaw-tools/src/web_fetch.rs
@@ -510,7 +510,7 @@ fn validate_target_url_with_dns_check(
     blocked_domains: &[String],
     allowed_private_hosts: &[String],
     tool_name: &str,
-    validate_dns: impl FnOnce(&str) -> anyhow::Result<()>,
+    validate_dns: impl FnOnce(&str, &[String]) -> anyhow::Result<()>,
 ) -> anyhow::Result<String> {
     let url = raw_url.trim();
 
@@ -566,7 +566,7 @@ fn validate_target_url_with_dns_check(
     }
 
     if !private_host_allowed {
-        validate_dns(&host)?;
+        validate_dns(&host, allowed_private_hosts)?;
     }
 
     Ok(url.to_string())
@@ -758,7 +758,10 @@ fn is_private_or_local_host(host: &str) -> bool {
 }
 
 #[cfg(not(test))]
-fn validate_resolved_host_is_public(host: &str) -> anyhow::Result<()> {
+fn validate_resolved_host_is_public(
+    host: &str,
+    allowed_private_hosts: &[String],
+) -> anyhow::Result<()> {
     use std::net::ToSocketAddrs;
 
     let ips = (host, 0)
@@ -779,16 +782,23 @@ fn validate_resolved_host_is_public(host: &str) -> anyhow::Result<()> {
         .map(|addr| addr.ip())
         .collect::<Vec<_>>();
 
-    validate_resolved_ips_are_public(host, &ips)
+    validate_resolved_ips_are_public(host, &ips, allowed_private_hosts)
 }
 
 #[cfg(test)]
-fn validate_resolved_host_is_public(_host: &str) -> anyhow::Result<()> {
+fn validate_resolved_host_is_public(
+    _host: &str,
+    _allowed_private_hosts: &[String],
+) -> anyhow::Result<()> {
     // DNS checks are covered by validate_resolved_ips_are_public unit tests.
     Ok(())
 }
 
-fn validate_resolved_ips_are_public(host: &str, ips: &[std::net::IpAddr]) -> anyhow::Result<()> {
+fn validate_resolved_ips_are_public(
+    host: &str,
+    ips: &[std::net::IpAddr],
+    allowed_private_hosts: &[String],
+) -> anyhow::Result<()> {
     if ips.is_empty() {
         anyhow::bail!("Failed to resolve host '{host}'");
     }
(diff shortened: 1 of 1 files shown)


Audacity88:
Thanks for jumping on this. This overlaps with #7424, which is the better review target because it also updates the config-field documentation and adds regression coverage for the allowed-domains boundary around `allowed_private_hosts = ["*"]`.

Closing this one to keep the fix on a single PR. Any useful review feedback or implementation details should move to #7424.

PR #6615, opened 2026-05-13
fix(providers): accept `reasoning` alias on OpenAI-compatible responses (#6584)

yijunyu:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `compatible.rs` only deserialized reasoning under the field name `reasoning_content`. OpenRouter and vLLM (>= v0.16.0, current upstream v0.20.2) standardized on `reasoning` instead, so reasoning was silently dropped for any thinking model served through vLLM. From the user's perspective, the model "stops reasoning" after a server upgrade.
  - **First commit:** added `#[serde(alias = "reasoning")]` to the two struct fields that surface reasoning content during deserialization:
    - `ResponseMessage::reasoning_content` (non-stream `chat/completions` responses)
    - `StreamDelta::reasoning_content` (SSE delta chunks)
  - **Follow-up commit (review feedback from @Audacity88):** the `#[serde(alias = …)]` shim correctly handles "only alias present" inputs, but serde rejects payloads carrying **both** `reasoning_content` and `reasoning` keys with a `duplicate field reasoning_content` error before any precedence rule can run. Replaced the alias shim with explicit intermediate raw structs (`RawResponseMessage` / `RawStreamDelta`) that deserialize the two keys into separate fields, plus a `From` / `Deserialize` impl that merges them with the documented precedence: **canonical `reasoning_content` wins; `reasoning` fills in only when canonical is absent or null.** New regression tests pin the "both keys" behavior on both the response and SSE paths.
  - Also accept both names on the round-trip path in `to_native_messages` (line 1501) that reconstructs assistant-with-tool-calls messages from stored history JSON: `value.get("reasoning_content").or_else(|| value.get("reasoning"))`. When both are present, `reasoning_content` wins — same precedence as the new deserializer.
  - **Scope note (write direction):** the issue asks for round-tripping the original field name on subsequent requests. That requires tracking the originally-seen name through the `reasoning_content` field on `ChatMessage` and emitting it conditionally across the entire providers crate (15 files reference `reasoning_content`), plus the runtime/memory crates that persist message history. That part is **explicitly out of scope for this PR**; this PR addresses only the read direction (the urgent "reasoning silently dropped" symptom from #6584).

[...]

PR #6606, opened 2026-05-13
fix(providers): preserve compatible reasoning field

Alix-007:
## Summary

- Preserve the original compatible-provider reasoning field name (`reasoning_content` vs `reasoning`) when responses are stored for later assistant-tool-call replay.
- Thread that field name through sync responses, streaming reasoning chunks, runtime history builders, and native compatible-provider message reconstruction.
- Add focused coverage for streaming aggregation so `StreamChunk.reasoning_field` is retained before the runtime synthesizes the final `ChatResponse`.

## Relationship to #6615

#6615 has already been merged and is the canonical immediate read-direction fix for #6584: compatible providers now accept both `reasoning_content` and the `reasoning` alias on incoming responses.

This PR is not competing to close #6584. It is a follow-up for replay preservation only: when a provider originally returned `reasoning`, subsequent assistant history replay should use `reasoning` instead of rewriting it to `reasoning_content`.

## Scope boundary

- Keeps #6615's canonical read behavior: `reasoning_content` wins when both keys are present; `reasoning` is only the alias fallback.
- Does not add new provider-specific reasoning semantics.
- Only preserves the field name required to round-trip compatible-provider assistant tool-call history.

## Verification

- `cargo fmt --check`
- `git diff --check origin/master...HEAD`
- `cargo test -p zeroclaw-providers compatible::tests::reasoning --lib`
- `cargo test -p zeroclaw-providers parse_sse_line_with_reasoning --lib`
- `cargo test -p zeroclaw-providers convert_messages_for_native_round_trips_reasoning_field --lib`
- `cargo test -p zeroclaw-providers convert_messages_for_native --lib`
- `cargo test -p zeroclaw-tool-call-parser build_native_assistant_history_from_parsed_calls --lib`
- `cargo test -p zeroclaw-runtime consume_provider_streaming_response_preserves_reasoning_field --lib`
- `cargo test -p zeroclaw-runtime build_native_assistant_history --lib`

PR #7423, opened 2026-06-09
fix(providers): preserve compatible reasoning field name on multi-turn replay (#6584)

### LINKED ISSUE #6584: [Bug]: OpenAI-Compatible provider ignores `reasoning` field, only reads `reasoning_content`
### Affected component

provider

### Severity

S2 - degraded behavior

### Current behavior

[OpenRouter](https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#preserving-reasoning), [vLLM](https://docs.vllm.ai/en/latest/features/reasoning_outputs/), and more are standardizing on `reasoning` as the field that reasoning content gets passed back to client via the OpenAI Chat Completions API. And, the same `reasoning` property in the streaming deltas. But, ZeroClaw [only looks at `reasoning_content`](https://github.com/zeroclaw-labs/zeroclaw/blob/b255ac15e32917805c721f93a0593fef5c64689a/crates/zeroclaw-providers/src/compatible.rs#L1501-L1504). To have maximum compatibility with self-hosted models (and specifically to have reasoning work with any recent vLLM release), we need to look for both of these fields and not just `reasoning_content`. Today, we only consider `reasoning_content` which ends up missing the reasoning for models running in vLLM since vLLM v0.16.0 (and latest upstream as of this writing is v0.20.2).

### Expected behavior

We should check for both `reasoning` and `reasoning_content` fields, keeping whichever is found. We should send those same fields back in on subsequent requests, again matching whatever we found in the original server response. This will allow maximum compatibility of reasoning handling with a variety of inference servers and versions.

### Steps to reproduce

I looked in the code to confirm this, but to reproduce you'd need to run with the OpenAI-compatible inference provider pointed to an inference server, such as vLLM >= v.0.16.0, that sends `reasoning` vs the older `reasoning_content` property. Run a reasoning model in that vLLM with thinking enabled - Qwen 3.5/3.6, Gemma 4, Nemotron 3 Super, etc. Observe that ZeroClaw is not capturing the reasoning outputs and sending that back in on subsequent turns - I do this by code changes to dump inputs on the inference server side, but it could be observed on the ZeroClaw client side as well.

### Impact

Affected users: Anyone using vLLM newer than v0.16.0 with ZeroClaw and reasoning models

The end result to the user is reasoning chain of thought is broken in multi-turn scenarios, which substantially impacts agentic workflows as the model's ability to plan and reason across long horizon tasks depends on that reasoning content being sent back with every previous assistant message.

### Logs / stack traces

```text

```

### ZeroClaw version

v0.7.5

### Rust version

_No response_

### Operating system

N/A

### Regression?

Unknown

### Pre-flight checks

chengzhichao-xydt:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  When an OpenAI-compatible provider returns reasoning under the `reasoning` key (OpenRouter / vLLM >= v0.16.0) rather than the canonical `reasoning_content`, the field name was normalized to `reasoning_content` on read and never restored on replay. Providers that require the original field shape on multi-turn requests could therefore drop reasoning on subsequent turns.

  Add a `reasoning` field to `NativeMessage` (serde-renamed from `reasoning`, mutually exclusive with `reasoning_content`) and preserve whichever key was present in the stored assistant message.
- **Scope boundary:** Only `compatible.rs` NativeMessage and its construction in `convert_messages_for_native`. No change to any other provider (OpenAI, Anthropic, Azure, etc.) or to the write path for non-replay messages.
- **Relationship to prior PRs:** #6615 (merged) fixed the read direction by accepting both keys on deserialization. This PR completes the round-trip by preserving the original key name on replay. #6606 (closed unmerged) attempted a broader cross-provider approach; this is scoped to the compatible provider only.
- **Linked issue(s):** Closes #6584

## Validation Evidence

**Real environment tested:** Rust 1.96.0, Linux, current `master` 24662b59.

```bash
cargo test -p zeroclaw-providers
cargo clippy -p zeroclaw-providers --all-targets -- -D warnings
cargo fmt -- --check
```

$ cargo test -p zeroclaw-providers
test result: ok. 919 passed; 0 failed

Key existing tests that verify the change:
- `convert_messages_for_native_reasoning_content_serialized_only_when_present` — verifies that `reasoning_content` and `reasoning` are both omitted from JSON when None (skip_serializing_if works for both fields)
- All `parse_native_response_*` tests — verify the read direction is unchanged
- All `native_chat_request_*` tests — verify request serialization

**Observed result after fix:** The `reasoning` key is round-tripped faithfully. Messages originally carrying `reasoning` will be replayed with `reasoning`; messages carrying `reasoning_content` will use `reasoning_content`. No duplicate keys are ever emitted (the two fields are mutually exclusive).

**What was not tested:** End-to-end multi-turn conversation with an actual vLLM instance using the `reasoning` key (requires infrastructure).

## Security and Privacy Impact

- Changes to key/token/credential handling? **No**
- PII, real identities, or personal information in diff, tests, fixtures, or docs? **No**

## Compatibility

Backward compatible. Providers that use `reasoning_content` continue to work identically. Providers using `reasoning` now also get the correct field name on replay. No new dependencies.

### DIFF
--- a/crates/zeroclaw-providers/src/compatible.rs
+++ b/crates/zeroclaw-providers/src/compatible.rs
@@ -979,6 +979,13 @@ struct NativeMessage {
     /// that require it in assistant tool-call history messages.
     #[serde(skip_serializing_if = "Option::is_none")]
     reasoning_content: Option<String>,
+    /// When the upstream response used the `reasoning` key (OpenRouter /
+    /// vLLM >= v0.16.0) rather than the canonical `reasoning_content`, we
+    /// store the value here so the field name round-trips faithfully.
+    /// The two fields are mutually exclusive — only one is ever `Some`.
+    /// See #6584.
+    #[serde(skip_serializing_if = "Option::is_none", rename = "reasoning")]
+    reasoning: Option<String>,
 }
 
 // ---------------------------------------------------------------
@@ -1827,10 +1834,16 @@ impl OpenAiCompatibleModelProvider {
                         .map(|value| MessageContent::Text(value.to_string()));
 
                     // Accept both `reasoning_content` (canonical) and
-                    // `reasoning` (OpenRouter / vLLM >= v0.16.0). See #6584.
+                    // `reasoning` (OpenRouter / vLLM >= v0.16.0).
+                    // Preserve whichever field name was originally
+                    // received so the value round-trips faithfully on
+                    // multi-turn requests.  See #6584.
                     let reasoning_content = value
                         .get("reasoning_content")
-                        .or_else(|| value.get("reasoning"))
+                        .and_then(serde_json::Value::as_str)
+                        .map(ToString::to_string);
+                    let reasoning = value
+                        .get("reasoning")
                         .and_then(serde_json::Value::as_str)
                         .map(ToString::to_string);
 
@@ -1840,22 +1853,32 @@ impl OpenAiCompatibleModelProvider {
                         tool_call_id: None,
                         tool_calls: Some(tool_calls),
                         reasoning_content,
+                        reasoning,
                     };
                 }
 
                 // Plain-text assistant turns from thinking-mode providers carry
-                // `reasoning_content` in a JSON-encoded `content` field with no
-                // `tool_calls` key. Without this branch the message would fall
-                // through to the plain-text fallback below and lose
-                // `reasoning_content`, so the next request to providers that
-                // require reasoning round-trip (e.g. DeepSeek V4 thinking) is
-                // rejected with a 400. See #6233.
+                // `reasoning_content` (or `reasoning`) in a JSON-encoded
+                // `content` field with no `tool_calls` key. Without this
+                // branch the message would fall through to the plain-text
+                // fallback below and lose reasoning, so the next request to
+                // providers that require reasoning round-trip (e.g. DeepSeek
+                // V4 thinking) is rejected with a 400. See #6233.
+                // Preserve the original field name for faithful round-trip
+                // (OpenRouter / vLLM >= v0.16.0).  See #6584.
                 if message.role == "assistant"
                     && let Ok(value) = serde_json::from_str::<serde_json::Value>(&message.content)
                     && value.get("tool_calls").is_none()
-                    && let Some(reasoning_content) = value
+                    && let Some((reasoning_content, reasoning)) = value
                         .get("reasoning_content")
                         .and_then(serde_json::Value::as_str)
+                        .map(|s| (Some(s.to_string()), None))
+                        .or_else(|| {
+                            value
+                                .get("reasoning")
+                                .and_then(serde_json::Value::as_str)
+                                .map(|s| (None, Some(s.to_string())))
+                        })
                     && matches!(
                         value.get("content"),
                         None | Some(serde_json::Value::Null | serde_json::Value::String(_))
@@ -1871,7 +1894,8 @@ impl OpenAiCompatibleModelProvider {
                         content,
                         tool_call_id: None,
                         tool_calls: None,
-                        reasoning_content: Some(reasoning_content.to_string()),
+                        reasoning_content,
+                        reasoning,
                     };
                 }
 
@@ -1904,6 +1928,7 @@ impl OpenAiCompatibleModelProvider {
                         tool_call_id,
                         tool_calls: None,
                         reasoning_content: None,
+                        reasoning: None,
                     };
                 }
 
@@ -1917,6 +1942,7 @@ impl OpenAiCompatibleModelProvider {
                     tool_call_id: None,
                     tool_calls: None,
                     reasoning_content: None,
+                    reasoning: None,
                 }
             })
             .collect()
@@ -3070,6 +3096,7 @@ mod tests {
                 tool_call_id: None,
                 tool_calls: None,
                 reasoning_content: None,
+                reasoning: None,
             }],
             temperature: Some(0.7),
             stream: Some(true),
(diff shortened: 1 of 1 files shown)


WareWolf-MoonWall (approved):
I read the full diff (+39/-10 in `crates/zeroclaw-providers/src/compatible.rs`), cross-checked the `NativeMessage` struct and both `convert_messages_for_native` branches against the local source, and verified zero prior review activity. This PR completes the round-trip for the `reasoning` key that #6615 started: the read path already accepts both keys; this makes the replay path emit whichever key was originally received.

---

### 🟢 What looks good — the fix is mechanically correct

`NativeMessage` now carries two mutually exclusive optional fields: `reasoning_content` (canonical) and `reasoning` (OpenRouter / vLLM ≥ 0.16.0). Both use `#[serde(skip_serializing_if = "Option::is_none")]`, so the wire output only ever contains one of them. The extraction in the tool-calls branch is correctly split — `value.get("reasoning_content")` and `value.get("reasoning")` are now independent arms that populate the right field. The plain-text branch uses a clean `map` / `or_else` tuple to produce exactly `(Some(...), None)` or `(None, Some(...))`, preserving the invariant that only one is ever `Some`. The fallthrough branches initialize `reasoning: None`, which is right.

### 🟢 What looks good — DRY compliance

The two fields are not duplicate state. They represent two different wire formats for the same semantic concept, with the receiving value stored in exactly the field that matches the original key. A single source of truth (the upstream API response) is preserved faithfully into the serialized output. No caching, no copy.

### 🟢 What looks good — scope boundary holds

The diff touches only `compatible.rs`. No other provider (OpenAI, Anthropic, Azure, Bedrock) is affected. The PR body is explicit about this and correct.

---

### 🟡 Warning — no dedicated regression test for the `reasoning` round-trip path

The validation evidence references `convert_messages_for_native_reasoning_content_serialized_only_when_present`, which only verifies that both fields are absent in JSON when `None`. That covers the `skip_serializing_if` wiring but not the new behavior.

What's missing is a test that:
1. Constructs a `serde_json::Value` assistant message with `"reasoning": "thinking text"` (and no `"reasoning_content"` key).
2. Calls `convert_messages_for_native` on a history containing it.
3. Asserts the resulting JSON contains `"reasoning": "thinking text"` and does **not** contain `"reasoning_content"`.

The analogous `reasoning_content` round-trip is exercised by `chat_with_tools_request_preserves_reasoning_content_in_history` (L4453). Adding a parallel `chat_with_tools_request_preserves_reasoning_key_in_history` test would give this fix a regression anchor so a future refactor can't silently re-break the `reasoning`-keyed path without a red test.

### 🟡 Warning — Rollback section absent; required for `risk: medium`

The PR template requires the Rollback section for `risk: medium` and `risk: high` PRs. It is not present in the PR body. The revert is trivially `git revert <merge-sha>` (single commit, no data migration), so the content itself is simple — the section just needs to be filled in. Please add it before merge.

### 🔵 Suggestion — `cargo clippy` and `cargo fmt` tail output missing from validation evidence

The template says "paste literal output (tails, failures, warnings — not 'all passed')." The test output is literal (`919 passed; 0 failed`), but the clippy and fmt commands are listed without output. Adding even a one-liner tail (`Finished 'dev' profile … in Xs`, exit 0) makes it easier for reviewers to confirm those gates ran cleanly.

---

**Verdict:** Approving. The fix is correct, the scope boundary holds, and no other reviewer holds an active block. The two 🟡 items above are worth addressing before merge but are not blockers on a one-commit +39/-10 fix with a clean test suite.

singlerider (approved):
Reviewed at head `46a9c80` against current `master`, with attention to the mutual-exclusivity invariant the fix relies on. Approving.

### 🟢 The round-trip fix is mechanically correct

`NativeMessage` now carries `reasoning_content` (canonical) and `reasoning` (OpenRouter / vLLM ≥ 0.16.0) as separate optional fields, both `skip_serializing_if = "Option::is_none"` with the latter `rename = "reasoning"`, so the value round-trips under whichever key the upstream response originally used instead of being normalized to `reasoning_content`. This completes the read/write round-trip #6615 started: the read path already accepted both keys, and now replay emits the original key. The plain-text branch's `map`/`or_else` tuple guarantees exactly `(Some, None)` or `(None, Some)`, the fallthrough branches initialize both to `None`, and the scope is confined to `compatible.rs` — no other provider is touched.

### 🔵 Minor — the invariant is only enforced in one of the two branches

The doc comment says "the two fields are mutually exclusive — only one is ever `Some`." The plain-text branch enforces that via the `or_else` tuple, but the tool-calls branch extracts `reasoning_content` and `reasoning` independently, so if an upstream message ever carried *both* keys, both fields would be `Some` and the replay would emit both. In practice a real provider sends one key or the other (they are the same provider's naming choice), so this is a degenerate input that doesn't occur — and emitting both back to a provider that sent both is harmless. Not a blocker; if you want the invariant to actually hold, prefer `reasoning_content` and only take `reasoning` when `reasoning_content` is absent, mirroring the plain-text branch.

### 🟡 Worth adding before merge (per @WareWolf-MoonWall)

A dedicated regression for the new behavior: build an assistant message with `"reasoning": "..."` and no `"reasoning_content"`, run `convert_messages_for_native`, and assert the output JSON contains `"reasoning"` and not `"reasoning_content"` (parallel to the existing `reasoning_content` round-trip test). And fill the Rollback section the `risk: medium` template requires (`git revert <sha>`, single commit, no migration).

CI is green and the branch is `MERGEABLE` / `CLEAN`. Approving — correct for real provider inputs, faithful round-trip, scope-bounded. The test and rollback items are worth folding in but don't block a one-commit +39/-10 fix.

PR #6974, opened 2026-05-27
fix(web_fetch): honor private DNS host allowlist

Audacity88:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Let explicit `web_fetch.allowed_private_hosts` entries bypass DNS public-IP validation for matching domain names, so local/private services reached through DNS names can be fetched when the user has explicitly allowed that host.
  - Keep the SSRF boundary for unlisted DNS names: hosts not in `allowed_private_hosts` still run through resolved-IP validation and are blocked when DNS resolves to private or otherwise non-global addresses.
  - Preserve `blocked_domains` precedence over the private-host allowlist.
  - Preserve the old wildcard safety shape: `allowed_private_hosts = ["*"]` still covers syntactically private/local hosts, but does not become a universal bypass for ordinary public-looking domain names.
- **Scope boundary:** This PR is limited to `web_fetch` URL validation. It does not change `http_request`.
- **Blast radius:** `web_fetch` users with explicit private-host allowlist entries can now fetch DNS-backed local/private domains. Unlisted domains, blocked domains, and ordinary public-domain allowlist checks remain guarded.
- **Linked issue(s):** Related #5122. Supersedes #5136 and #5164.
- **Labels:** `bug`, `risk: high`, `security`, `tool`, `tool: web_fetch`, `domain:web-fetch`, `tool:web`, `size: S`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
PASS cargo test -p zeroclaw-tools web_fetch::tests:: -- --nocapture
57 web_fetch tests passed after the wildcard and blocklist regressions were added.

PASS cargo fmt --all -- --check

PASS cargo clippy -p zeroclaw-tools --all-targets -- -D warnings

PASS cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings
Finished dev profile in 37m 09s. This captured pass covered the `web_fetch.rs` diff before the later #6962 test-only merge; the post-#6962 CI-shaped nextest pass below verifies the refreshed test gate.

PASS cargo nextest run --locked --workspace --exclude zeroclaw-desktop
8110 tests run: 8110 passed, 10 skipped.

PASS cargo nextest run --locked -p zeroclaw-tools
1185 tests run: 1185 passed, 0 skipped.
```

[...]

PR #7424, opened 2026-06-09
fix(web_fetch): let allowed_private_hosts = ["*"] cover DNS-resolved private hosts (#7412)

### LINKED ISSUE #7412: [Bug]: web_fetch/http_request `allowed_private_hosts = ["*"]` does not cover domain names that resolve to private IPs

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The `"*"` wildcard in `web_fetch.allowed_private_hosts` was *narrower* than a named entry: a named host (`internal.example.com`) bypassed the resolved-IP public check, but `["*"]` did not, so a regular domain that **resolved** to a private/internal IP was still blocked (`Blocked host '...' resolved to non-global address ...`). This is counter-intuitive — users reasonably expect `["*"]` to be a superset of any named entry (#7412).
  - Root cause: the `"*"` branch of the private allowlist gated on `is_private_or_local_host(host)`, which is only `true` for IP literals / `localhost` / `.local` — never for an ordinary domain name — so `"*"` never lifted the DNS resolved-IP check.
  - Fix: replace the boolean match with a three-state `PrivateAllow { None, Wildcard, Explicit }` so the wildcard's effects can be split out — `["*"]` now tolerates a private/internal **resolution** (skips the resolved-IP public check) for any host that already passes `allowed_domains`. The three-state distinction also lets the bypass WARN tell a deliberate explicit carve-out from a bare `["*"]`, so it no longer fires on every fetch once `["*"]` is set.
  - **Interaction with #7340 (merged into this branch):** #7340 landed on `master` (as a "refactor") and both extracted the shared `domain_guard` helpers **and** tightened the gate so that a **non-private** host listed in `allowed_private_hosts` (whether matched explicitly or via `["*"]`) must still pass `allowed_domains`. This branch adopts that tightening: the `allowed_domains` name check is now skipped **only** for hosts that are *literally* private/local (IP literal / `localhost` / `.local`). Consequently explicit entries no longer grant a blanket `allowed_domains` bypass — for them only the resolved-IP SSRF check is lifted. The net behavioural delta of *this* PR over current `master` is exactly the one wildcard change above (`["*"]` skipping the resolved-IP check for an in-`allowed_domains` host that resolves to a private IP).
- **Scope boundary:** Only `web_fetch`'s private-host gating + the `allowed_private_hosts` field doc. No change to `blocked_domains` precedence, `allowed_domains` matching, the literal-private-host block, or `http_request` (which has no resolved-IP DNS check and a separate `allow_private_hosts` opt-in).
- **Blast radius:** `web_fetch` / `firecrawl` fetch path SSRF gating only. `blocked_domains` still wins unconditionally; `"*"` still cannot reach a public host that is not in `allowed_domains`.
- **Linked issue(s):** Closes #7412
- **Labels:** `bug`, `size: S`, `risk: high`, `config`, `tool`, `tool:web`.

```bash
cargo fmt --all -- --check          # exit 0, no output
cargo clippy --all-targets -- -D warnings   # exit 0 — Finished in 9m10s, no warnings
cargo test --workspace --lib --tests        # exit 0 — 8162 passed across 41 binaries, 0 failed
```

  - `fmt`: clean (exit 0).
  - `clippy --all-targets -D warnings`: `Finished \`dev\` profile [unoptimized + debuginfo] target(s) in 9m10s` — exit 0, zero warnings.
  - `cargo test --workspace --lib --tests`: `test result: ok` for all 41 test binaries, **8162 passed / 0 failed**. New regression tests pass:
    - `web_fetch::tests::private_wildcard_allows_domain_resolving_to_private_ip ... ok`
    - `web_fetch::tests::private_wildcard_allows_literal_private_ip_without_allowed_domains_entry ... ok`
    - `web_fetch::tests::private_wildcard_still_requires_allowed_domains ... ok`
- **Beyond CI — what did you manually verify?** Traced the full precedence chain (`blocked_domains` > literal-private block > `allowed_domains` (skipped only for a literal private/local host in the private allowlist) > resolved-IP check (skipped for any host in the private allowlist)) and added a regression test for a behaviour I found while auditing: a literal private IP (e.g. `10.0.0.1`) under `allowed_private_hosts = ["*"]` must stay allowed even with a restrictive `allowed_domains`, preserving the historical `"*"` scope. Built the debug binary (`cargo build`, exit 0) for local smoke testing. Did NOT add a live network test (DNS resolution of a real internal name is environment-specific; covered by the injected-DNS closure tests).
- **If any command was intentionally skipped, why:** The full `cargo test` doc-test phase aborts locally with `error: Option 'default-theme' given more than once`, because the repo's `.cargo/config.toml` sets `rustdocflags = ["--default-theme=ayu"]` which rustdoc 1.94.1 rejects when cargo duplicates it for doctests. This config is identical on `master` and unrelated to this diff (which touches no doctests and not the `zeroclaw` crate), so doctests were run as `--lib --tests` to isolate the relevant suites. CI's toolchain is unaffected.
- **Post-merge re-verification (conflict resolution with `master`):** Merged current `master` (now including #7340's `domain_guard` refactor + tightening) and resolved the `web_fetch.rs` conflict — adopting the `domain_guard::` helpers and #7340's `skip_allowed_domains = host_is_private_or_local && private_tolerated`, while keeping this PR's three-state wildcard fix. Re-ran on the merge: `cargo fmt --all -- --check` (clean), `cargo clippy -p zeroclaw-tools -p zeroclaw-config --all-targets` (zero warnings), and `cargo test -p zeroclaw-tools -p zeroclaw-config --lib --tests` (all pass — `web_fetch` 59/59, covering both this PR's three `private_wildcard_*` tests and #7340's `private_allowlist_explicit_entry_must_pass_allowed_domains`).

- New external network calls? **No** — this only governs which already-configured `web_fetch` targets are permitted; it does not add calls.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (tests use `example.com` / RFC1918 literals).
- If any `Yes`, describe the risk and mitigation: This is an SSRF-boundary change. The widening is strictly opt-in (operator sets `allowed_private_hosts = ["*"]`) and bounded: `"*"` only **skips the resolved-IP check** for hosts that already pass `allowed_domains`; it never widens `allowed_domains`, and `blocked_domains` still takes precedence. So `"*"` cannot be used to reach an arbitrary public host or a blocked one.

## Compatibility (required)

- Backward compatible? **Yes** — named entries and the default (empty) behaviour are unchanged. The only behavioural change is that `allowed_private_hosts = ["*"]` now also permits names (in `allowed_domains`) that resolve to private IPs, which previously failed.
- Config / env / CLI surface changed? **No** new fields; the semantics of the existing `allowed_private_hosts = ["*"]` value are widened and now documented on the field.

- **Fast rollback command/path:** `git revert <sha>` — single self-contained commit.
- **Feature flags or config toggles:** Operators can immediately restore the old narrow behaviour by removing `"*"` from `web_fetch.allowed_private_hosts` (and listing exact hosts instead). No code rollback needed for that.
- **Observable failure symptoms:** Grep logs for `web_fetch: allowing host via allowed_private_hosts` (a WARN emitted whenever a private/internal host is permitted) and for `resolved to non-global address` (the block path). An unexpected widening would show the former for hosts that should not be reachable.


crates/zeroclaw-tools/src/web_fetch.rs
```diff
@@ -541,17 +541,32 @@ fn validate_target_url_with_dns_check(
...
         anyhow::bail!(
             "Blocked local/private host: {host}. \
-             To allow this host, add it to {tool_name}.allowed_private_hosts in config.toml"
+             To allow this host, add it (or \"*\") to \
+             {tool_name}.allowed_private_hosts in config.toml"
         );
     }
 
-    if private_host_allowed {
+    if private_tolerated {
```
singlerider: With `"*"` in `allowed_private_hosts`, `private_allowlist_match` returns `Wildcard` for any host, so `private_tolerated` is true for every request and this WARN fires on every ordinary public fetch. The SSRF-bypass signal then becomes indistinguishable from noise for the exact operator this PR is built to support. Gate it on a real bypass:

```rust
if private_explicit || (private_tolerated && host_is_private_or_local) {
    // record! ...
}
```

The new `"*"` + DNS-resolves-private case is only knowable post-resolution, so log that one (if wanted) from the resolved-IP path, not this pre-DNS block.

singlerider (changes requested):
This is a careful relaxation of a deliberately tight gate, and the three-way `PrivateAllow` split is the right way to express it: `*` tolerates a private resolution but never widens the name allowlist, explicit entries keep their historical carve-out semantics, and all three properties are pinned by tests (including the panic-closure trick to prove the DNS check is actually skipped). One blocking item.

### 🔴 Blocking: with `"*"` configured, the private-host-bypass WARN fires on every fetch

`private_allowlist_match` returns `Wildcard` for any host whenever `"*"` is present, so `private_tolerated` is true for every request, and the `record!(WARN ... "web_fetch: allowing host via allowed_private_hosts")` block now logs an SSRF-bypass event for every ordinary public fetch. Before this change, `"*"` only produced that WARN when the host was literally private/local. An operator who sets `"*"` (the exact config this PR exists to support) gets an audit log where the bypass signal is indistinguishable from noise, which defeats the point of logging the bypass at all.

Gate the WARN on the cases where a bypass actually happened or can happen:

```rust
if private_explicit || (private_tolerated && host_is_private_or_local) {
    // record! ...
}
```

The genuinely new case (a name in `allowed_domains` resolving to a private IP under `"*"`) is only knowable after resolution, so if you want that logged too, the place is the resolved-IP check path rather than this pre-DNS block. A simple gate as above plus a one-line comment is fine for this PR.

### 🟢 Wildcard cannot reach arbitrary public hosts

`private_bypasses_name_check` requiring `Explicit` or `(Wildcard && literally private)` preserves the invariant from #6974, and `private_wildcard_still_requires_allowed_domains` pins it. `blocked_domains` still bails first. The explicit-beats-wildcard scan order in `private_allowlist_match` is correct and commented with the why.

### 🟢 Error messages now point at the fix

Both the literal-private block and the resolved-private block tell the operator exactly which config key to touch. That closes the loop on #7412's actual complaint, which was as much discoverability as behavior.

### 🟢 Schema doc spells out the `*` contract

The `allowed_private_hosts` doc comment now states precisely what `*` does and does not grant. That sentence will save a future security review an hour.

Audacity88:
@NiuBlibing, I agree with @singlerider's remaining blocker here.

The behavior split in this PR looks right: explicit `allowed_private_hosts` entries keep the historical carve-out, while `*` can tolerate private DNS resolution only for hosts that already pass `allowed_domains`.

Please keep the next update narrow:

- Gate the pre-DNS WARN so it only fires for an actual literal-private bypass or explicit private-host carve-out.
- Leave the `PrivateAllow` split and the #7412 regression tests in place.

That should keep this PR focused on #7412 and avoid mixing in the broader helper-refactor work.

singlerider:
Still at head `615c29cf5` — no new commits since my 2026-06-13 review, so my changes-requested remains live and unaddressed. Confirming, not re-litigating.

The blocker stands: `web_fetch.rs:569` still gates the SSRF-bypass WARN on `if private_tolerated`, and `private_tolerated` (line 552) is true for **every** host once `"*"` is configured — the exact config this PR exists to support. So an operator who sets `allowed_private_hosts = ["*"]` gets `web_fetch: allowing host via allowed_private_hosts` logged on every ordinary public fetch, making the bypass signal indistinguishable from noise. Gate it on the cases where a bypass actually happened:

```rust
if private_explicit || (private_tolerated && host_is_private_or_local) {
    // record! ...
}
```

Everything else is in good shape — the three-way `PrivateAllow` split, the `private_bypasses_name_check` invariant (wildcard can't reach arbitrary public hosts, pinned by `private_wildcard_still_requires_allowed_domains`), the fix-pointing error messages, and the schema doc are all correct. Land the WARN gate and I'll approve.

Audacity88 (commented):
Reviewed current head `b71a3610ad070a4bd5d00a9baa2a3548f8569553` against the live PR state, issue #7412, @singlerider's prior review and follow-up comment, the current `web_fetch` / config-schema diff, the relevant tool/security docs, and the green CI run on this head. I am posting this as a comment rather than an approval because @singlerider still holds the active formal `CHANGES_REQUESTED` review, but from my side the latest commit appears to resolve the remaining WARN-noise blocker.

### ✅ Resolved — private-host WARN no longer fires for every public fetch under `"*"`

The latest commit changes the pre-DNS WARN gate from `private_tolerated` to `private_explicit || (private_tolerated && host_is_private_or_local)`. That means `allowed_private_hosts = ["*"]` no longer makes ordinary public fetches log `web_fetch: allowing host via allowed_private_hosts` just because `private_allowlist_match()` returns `Wildcard` for every host. The WARN now fires only when this path can tell that an explicit carve-out or literal private/local bypass happened, which matches the blocker @singlerider raised.

### 🟢 What looks good — the wildcard stays bounded by `allowed_domains`

The three-state `PrivateAllow` split keeps the security boundary readable: `blocked_domains` still wins first, explicit private-host entries keep their historical name-check and resolved-IP carve-out, and the `"*"` wildcard can tolerate private DNS resolution only after the ordinary host name still passes `allowed_domains`. The new tests pin the important cases: the #7412 path where `internal.example.com` is allowed under `allowed_domains = ["example.com"]`, the historical literal private-IP wildcard path, and the invariant that `"*"` does not reach an unrelated public host outside `allowed_domains`.

That leaves me with no further code blocker on the current head, subject to @singlerider clearing or updating the active block.

Audacity88 (commented):
Reviewed current head `cc605808cc98` against the live PR state, issue #7412, the prior review thread, the current `web_fetch` / config-schema diff, and the green CI run on this head. The current head is a merge commit that preserves the substantive WARN-gate fix from `b71a3610ad`. I do not see a new code blocker from my side on this head, but I am posting this as a comment rather than an approval because @singlerider holds the active formal `CHANGES_REQUESTED` review.

### ✅ Resolved — private-host WARN stays gated to real pre-DNS carve-outs

The current head keeps the fix from `b71a3610ad`: the pre-DNS WARN now checks `private_explicit || (private_tolerated && host_is_private_or_local)` instead of bare `private_tolerated`. That preserves the audit signal under `allowed_private_hosts = ["*"]`: ordinary public fetches no longer emit `web_fetch: allowing host via allowed_private_hosts` just because the wildcard exists, while explicit carve-outs and literal private/local wildcard bypasses still warn.

### 🟢 What looks good — the SSRF widening remains bounded

The `PrivateAllow` split still keeps the important boundaries separate. `blocked_domains` runs first, explicit entries keep their historical carve-out, and `*` can tolerate private DNS resolution only after the hostname still passes `allowed_domains`. The regression tests cover the #7412 path, the historical literal-private-IP wildcard path, and the invariant that `*` does not reach an unrelated public host outside `allowed_domains`.

Audacity88:
@NiuBlibing thanks for the latest update here. This head has green CI, but the branch now conflicts with current `master` in `crates/zeroclaw-tools/src/web_fetch.rs` after the `domain_guard` refactor landed.

Please refresh the branch on current `master` and replay the `allowed_private_hosts = ["*"]` behavior on top of the new helper shape. After that, CI can give us the merge-result signal we need, and singlerider can re-review the current head.

NiuBlibing:
@Audacity88 I've resolved the conflicts and the ci is green.

Audacity88 (commented):
Reviewed current head 49f4cecff454dfb41f53949e80aa23e6170890b1 against the live PR state, issue #7412, @singlerider's formal review and inline thread, prior Audacity88 follow-up comments, the final `web_fetch` / schema source at this head, and the current green CI run. Thanks for resolving the conflict after the `domain_guard` refactor and replaying the #7412 behavior on top of the newer helper shape. I am leaving this as a comment rather than an approval because @singlerider still owns the active formal `CHANGES_REQUESTED` review, but I do not see a remaining code blocker from my side on this head.

### ✅ Resolved — the branch now carries the fix on the current helper shape

The current head is mergeable, CI is green, and `validate_target_url_with_dns_check()` now uses the shared `domain_guard` helpers while keeping the `PrivateAllow` split this PR needs. The post-#7340 tightening is preserved: a non-private host still has to pass `allowed_domains`, even if it is explicitly listed in `allowed_private_hosts`, while the private allowlist can still skip the resolved-IP public check for the #7412 case.

### ✅ Resolved — the private-host WARN no longer fires for every public fetch under `"*"`

The final code keeps the reviewer-requested WARN gate:

```rust
if private_explicit || (private_tolerated && host_is_private_or_local) {
```

That means `allowed_private_hosts = ["*"]` no longer turns every ordinary public fetch into `web_fetch: allowing host via allowed_private_hosts` log noise. The WARN stays tied to explicit carve-outs and literal private/local wildcard bypasses, which preserves the audit signal @singlerider called out.

### 🟢 What looks good — the widened wildcard remains bounded

The regression tests cover the important edges: a name inside `allowed_domains` can now resolve to a private IP under `allowed_private_hosts = ["*"]`; a literal private IP still keeps the historical wildcard behavior; `blocked_domains` still wins; and neither `*` nor an explicit non-private entry can reach a host outside `allowed_domains`. That is the right shape for a security-sensitive relaxation: it fixes the surprising wildcard gap without turning the private allowlist into a general name-allowlist bypass.

singlerider (approved):
Re-review at head `59e97ac95d`. Pulled the branch, ran the private-host tests and clippy. This builds on the already-merged #6974 private-host allowlist (merged 2026-06-02) — it is a follow-up for #7412, not a duplicate.

### ✅ RESOLVED — the bypass WARN no longer fires on every fetch

My one blocker is fixed with exactly the gate I suggested. The WARN is now `if private_explicit || (private_tolerated && host_is_private_or_local)`, with the comment explaining that gating on bare `private_tolerated` would log the SSRF-bypass on every fetch once `"*"` is set and bury the real signal. An operator who sets `"*"` now gets an audit log where the bypass WARN only fires when a private/internal host was actually let through, not on ordinary public fetches.

### 🟢 SSRF invariant preserved and pinned

`skip_allowed_domains = host_is_private_or_local && private_tolerated` keeps the contract that a non-private host (including an explicit internal DNS name or any host reached via `"*"`) still requires `allowed_domains` approval, so the private allowlist can never reach an arbitrary public host. `private_wildcard_still_requires_allowed_domains` and `private_allowlist_wildcard_does_not_allow_public_domain_miss` pin it; `blocked_domains` still bails first. `cargo test -p zeroclaw-tools --lib private` is 35/35 green and clippy is clean. The schema doc spells out the `*` contract. Ships.

PR #7425, opened 2026-06-09
fix(runtime): resolve channel pricing via bare-type fallback in cost lookup

garacio:
## Summary

- **Base branch:** `master`
- **What changed and why:** Channel cost tracking silently recorded `cost_usd = 0` and left per-day budget enforcement inert for every channel agent. Root cause: the channel orchestrator builds the pricing map keyed by the **bare provider `<type>`** (`crates/zeroclaw-channels/src/orchestrator/mod.rs`), while `record_tool_loop_cost_usage` always receives the **composite `<type>.<alias>`** from `agent_provider_composite`, so `model_provider_pricing.get(<composite>)` missed. The CLI / cron / web agent-loop builder keys by the composite instead, so the same lookup worked there — the two builders disagree. Fixed at the lookup site: try the composite verbatim, then fall back to the bare-type prefix.
- **Scope boundary:** Only the pricing-map lookup in `cost.rs`. No change to either builder, to `resolve_rates`, to rate values, or to the inference path.
- **Blast radius:** Cost/budget accounting only. Resolving the map cannot change inference behavior; worst case a previously-missed rate now applies (the intended behavior). The composite-keyed path is unaffected (verbatim `get` still hits first).
- **Linked issue(s):** none filed.
- **Labels:** `bug`, `risk: high`, `size: S`, `agent`, `runtime`

```bash
cargo fmt --all -- --check     # clean, no output
#   Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 19s   (exit 0, no warnings)
cargo test
#   all suites ok — 0 failed across the workspace (exit 0)
#   new test: agent::cost::tests::provider_pricing_resolves_composite_and_bare_type_keys ... ok
```

- **Beyond CI — what was manually verified:** Built the patched binary, pointed an agent at a GLM provider (`glm.default`, model `glm-5.1`) with a per-alias `pricing` table (`"glm-5.1.input" = 1.4`, `"glm-5.1.output" = 4.4`), and sent a live Telegram message. Before the fix `costs.jsonl` recorded `cost_usd = 0.0` plus the "no pricing entry found" WARN; after the fix the same path recorded `cost_usd = 0.0185626` for `13149` input / `35` output tokens (`13149*1.4/1e6 + 35*4.4/1e6 = 0.0185626`), and the WARN stopped firing.
- **Not verified:** the `[cost.rates...]` sheet surface (type-keyed) — it benefits from the same fallback by construction but was not exercised end-to-end.

## Compatibility (required)

- Upgrade steps: none. Existing configs whose pricing previously resolved continue to resolve identically (verbatim composite lookup is tried first); configs that silently recorded zero cost on channels now resolve correctly.

- **Observable failure symptoms:** Unexpected channel-turn cost values for aliased providers, unexpected budget denials after channel LLM calls, or cost records whose token math does not match the configured provider pricing.

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/cost.rs
+++ b/crates/zeroclaw-runtime/src/agent/cost.rs
@@ -115,6 +115,27 @@ fn resolve_rates(pricing: &HashMap<String, f64>, model: &str) -> (f64, f64, f64)
     (0.0, 0.0, 0.0)
 }
 
+/// Resolve the per-model pricing map for a provider reference.
+///
+/// `model_provider_name` always arrives as the composite `<type>.<alias>`
+/// (see `agent_provider_composite`), but the outer pricing map may be keyed
+/// either way depending on which builder populated it: the CLI / cron / web
+/// agent loop keys by the composite alias, while the channel orchestrator keys
+/// by the bare provider `<type>` (rates are per provider type, not per alias).
+/// Try the composite verbatim first, then fall back to the bare type prefix so
+/// cost tracking resolves regardless of the builder — and so the type-keyed
+/// `cost.rates` sheet is honored on the alias paths too.
+fn provider_pricing<'a>(
+    map: &'a ModelProviderPricing,
+    model_provider_name: &str,
+) -> Option<&'a HashMap<String, f64>> {
+    map.get(model_provider_name).or_else(|| {
+        model_provider_name
+            .split_once('.')
+            .and_then(|(provider_type, _alias)| map.get(provider_type))
+    })
+}
+
 /// Record token usage from an LLM response via the task-local cost tracker.
 /// Returns `(total_tokens, cost_usd)` on success, `None` when not scoped or no usage.
 pub fn record_tool_loop_cost_usage(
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
I reviewed the PR diff, PR body validation evidence, current labels/checks, empty prior review/comment state, and the public source path through the channel orchestrator pricing map, CLI agent-loop pricing map, and runtime cost lookup. I did not rerun the local cargo commands or independently exercise the `[cost.rates...]` sheet path beyond the source-path check.

### 🟢 What looks good — minimal fix at the lookup boundary

The patch fixes the mismatch at the narrowest point. It keeps the existing composite-key lookup first, then falls back to the bare provider type only when the composite key is absent. That preserves the CLI / cron / web path while making the channel builder's type-keyed pricing map resolve. I also like that the PR leaves `resolve_rates`, rate values, and the inference path alone; this is purely cost-accounting plumbing.

### 🟢 What looks good — validation exercises the real failure mode

The new test covers both accepted key shapes, and the PR body adds useful live-channel evidence: before the fix, the Telegram path logged the missing-pricing warning and wrote `cost_usd = 0.0`; after the fix, the same path wrote the expected nonzero cost for the GLM token counts. That is the right kind of evidence for this bug because it proves the changed lookup reaches the operator-visible cost record.

No code blockers from me. The runtime change is small, well-scoped, and supported by both unit coverage and live-channel validation.

PR #7427, opened 2026-06-09
fix(agent): remove dead context_aware_tools config field (#6720)

### LINKED ISSUE #6720: [agent] context_aware_tools config field is declared but unread (dead code)
## Summary

`[agent] context_aware_tools` parses cleanly via `AgentConfig` and the doc-comment promises "Enable context-aware tool filtering (only surface relevant tools per iteration)" — but the field is **never read anywhere in the codebase**. Setting `context_aware_tools = true` does literally nothing.

This is either dead code from a removed feature, or scaffolding for a feature that never landed. Either way it's a silent footgun for users who toggle it expecting behavior change (I did).

## Evidence

Exhaustive grep across `crates/` finds exactly two hits for `context_aware_tools`:

- **Declaration**: `crates/zeroclaw-config/src/schema.rs:1619-1621`
  ```rust
  /// Enable context-aware tool filtering (only surface relevant tools per iteration).
  #[serde(default)]
  pub context_aware_tools: bool,
  ```
- **Default**: `crates/zeroclaw-config/src/schema.rs:1704` — `context_aware_tools: false`

No `.context_aware_tools` field accesses elsewhere. No `ContextAware*` type. No factory or runtime code that consumes the value.

Auto-generated config docs (`docs/book/po/*.po`) parrot the doc-comment, surfacing it to operators as a real knob.

## Impact

Operators discover the field via the config schema or generated docs, toggle it to `true` expecting per-iteration tool filtering, and get no observable behavior change — and no warning. This was my entry point into investigating other tool-search related gaps (see related #6699 for `tool_filter_groups` issues).

## Suggested fix

One of:

1. **Remove the field** with a one-line deprecation note in CHANGELOG (cleanest if the feature isn't planned).
2. **Implement it** — the doc-comment describes per-iteration tool filtering, which would be a useful complement to `tool_filter_groups` (esp. under `deferred_loading = true`).
3. **At minimum, emit a config-load warning** when `context_aware_tools = true` to fail loudly instead of silently.

Happy to send a PR for (1) or (3) if there's consensus on direction.

chengzhichao-xydt:
## Summary

Closes #6720. The `context_aware_tools` config field parses cleanly and is documented as enabling per-iteration tool filtering, but no runtime code ever read it — the feature was never landed.

Per maintainer direction on #6720, this PR removes the field entirely rather than keeping the dead config surface with a warning.

## Changes

- Remove `context_aware_tools` from `AgentConfig` / `RuntimeProfileConfig` structs
- Remove `effective_context_aware_tools()` resolver method from `AgentConfig`
- Remove from `ResolvedRuntime` and its `Default` impl
- Remove from Quick and Unbounded runtime presets
- Remove from v1 test fixture

## Validation

- `cargo check --bin zeroclaw` passes (full workspace check)
- `cargo fmt --all` passes
- Grep confirms zero remaining `context_aware_tools` references in Rust/Toml sources

## Compatibility

**Breaking config change**: anyone who has `context_aware_tools = true` in a `[runtime]` profile will get a parse error on upgrade. The field was a no-op with no runtime consumer, so the behavioral impact is nil. Add a one-line CHANGELOG entry noting the removal.

## Risk

Low. Removes dead code that had zero runtime effect. No runtime behavior changes.


Audacity88 (changes requested):
Reviewed current head b500f5b against live PR state, linked issue #6720, the empty review/comment history, the one-file runtime diff, and visible GitHub checks. CI is green. I am requesting changes because the PR does not match the accepted resolution for the issue it closes.

### 🟢 What looks good — The warning patch is narrow

Thanks @chengzhichao-xydt. Adding a visible warning is better than silently accepting a no-op setting, and the implementation is small. The patch is scoped to `AgentBuilder::config()`, reads `config.resolved.context_aware_tools`, and only emits the warning when the operator explicitly enabled the no-op setting.

### 🔴 Blocking — Match the accepted #6720 resolution before closing it

The problem is that #6720 already has a maintainer-scoped resolution: remove the unused `context_aware_tools` schema field/default and refresh generated user-facing docs so ZeroClaw stops advertising a knob with no runtime consumer.

This PR keeps that config surface in place, adds a runtime warning from `AgentBuilder::config()`, and still says `Fixes #6720`. That changes the accepted path from removal to compatibility-preserving warning without reopening the design decision, and it leaves the dead documented option in the product.

Please either switch this PR to the requested removal slice from #6720, or reopen the design discussion and retarget this as `Related #6720` rather than `Fixes #6720`.

Audacity88:
Reviewed current head `33214a9` against the updated title/body, the prior review, #6720, and the current one-file diff.

@chengzhichao-xydt thanks for retargeting this to `Related #6720`. That fixes the closing-reference problem from the earlier review.

The remaining blocker is still the issue direction. #6720 has an accepted path: remove the unused `context_aware_tools` schema field/default and refresh the generated user-facing docs so ZeroClaw no longer advertises a config knob that has no runtime consumer.

This PR still keeps the config field and adds a runtime warning. That could be a reasonable interim behavior if we reopen the design call, but it is not the accepted #6720 resolution yet.

Please either switch this PR to the removal slice described in #6720, or reopen the design discussion there and keep this PR scoped as an interim warning instead of the issue resolution. Until one of those happens, the existing requested-changes review still stands.

Audacity88 (approved):
I re-reviewed current head 22d331594bc1b368129526400d492c2ca7c34677 against master, including #6720, the prior review, the updated PR body, the full diff, and visible GitHub checks. I did not run local validation.

### ✅ Resolved — The PR now follows the accepted #6720 removal path

Thanks @chengzhichao-xydt. This now matches the maintainer direction from #6720. The dead `context_aware_tools` config surface is removed from `RuntimeProfileConfig`, `ResolvedRuntime`, the effective resolver path, the runtime presets, and the v1 fixture instead of being kept with a warning.

### 🟢 What looks good — The stale advertised knob is gone from the schema

The diff also removes the stale schema doc-comment source for the advertised `context_aware_tools` knob. The committed config surface no longer carries this no-op option.

The compatibility note in the PR body correctly calls this out as a breaking config cleanup with no runtime behavior removal, since the field had no consumers. CI is green on the current head.

Approving; this clears my prior blocker.

PR #7433, opened 2026-06-09
fix(runtime): refresh active sessions after provider config edits

### LINKED ISSUE #7395: [Bug]: zerocode Config model-provider edits do not update active Chat sessions
# [Bug]: zerocode Config model-provider edits do not update active Chat sessions

## Affected component

unknown

This is in the zerocode Config/Chat TUI path, but the current bug template does not have a zerocode/TUI component option.

## Severity

S2 - degraded behavior

## Current behavior

When a model-provider model is edited from zerocode Config, the new value is persisted in config, but an already-open Chat session can keep using the model provider instance that was built before the edit.

In the observed case, `providers.models.openai.default.model` was changed from an invalid embedding model to a chat model. Config showed the updated model, and the saved config contained the new value, but sending another message in the existing Chat session still failed against the previous model.

Relevant sanitized error excerpt:

```text
turn failed: Agent turn failed: All model_providers/models failed. Attempts:
model_provider=openai model=text-embedding-ada-002 attempt 1/3: non_retryable;
error=OpenAI API error (400 Bad Request): The requested model, 'text-embedding-ada-002' was not found.
```

Source inspection matches the behavior: `config/set` updates and flushes the config, while the existing live rebuild path is attached to `session/configure`. Editing `providers.models.<family>.<alias>.model` through Config does not appear to rebuild active sessions that use that provider.

## Expected behavior

After a Config edit changes `providers.models.<family>.<alias>.model`, active Chat sessions using that provider should either:

1. rebuild or reconfigure their live provider before the next turn, or
2. clearly tell the user that the change only applies to new sessions and that the current Chat session must be restarted.

The UI should not show the new model in Config while the next Chat turn silently continues using the old provider/model.

## Steps to reproduce

```bash
1. Start zerocode with an agent using a configured provider such as `openai.default`.
2. Configure `providers.models.openai.default.model` to an invalid or unsuitable model, for example `text-embedding-ada-002`.
3. Open Chat for that agent and send a message; confirm the turn fails against the invalid model.
4. Go to Config -> Model providers -> openai -> default -> Model.
5. Change the model to a valid chat model and confirm Config shows the new value.
6. Return to the same Chat session and send another message.
7. Observe that the active session can still call the old model.
```

## Impact

Affected users: zerocode users who correct provider settings from Config while a Chat session is already open.

Frequency: reproducible in the observed live test.

Consequence: users can believe they fixed the provider config, but the active Chat session continues failing until they start a new session or use the Chat model switcher path.

## Logs / stack traces

The log excerpt above is sanitized. The important signal is that saved config and outer turn attribution can reflect the new model while the provider error still comes from the old model, which indicates stale live provider state inside the active session.

## ZeroClaw version

Observed while testing zerocode v0.8.0-beta-2 behavior around PR #7380. Local source inspection against `origin/master` also shows `config/set` has no active-session provider rebuild path.

## Rust version

Not checked; this is a runtime/TUI behavior issue rather than a build failure.

## Operating system

macOS

## Regression?

Unknown

## Pre-flight checks

- Reproduced during live zerocode testing.
- Secrets, tokens, private paths, and session identifiers are redacted or omitted.
- Duplicate search using exact stale-model terms returned no matching issue; a broader GitHub search timed out during TLS handshake and should be retried before or after filing if needed.

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Refresh live sessions after provider profile edits under `providers.models.<family>.<alias>.*`.
  - Reuse the existing session provider rebuild path so active sessions get the same provider wiring as a fresh session.
  - Schedule the live refresh as best-effort background work after the config save, so `config/set` and `config/delete` can return without waiting behind an active turn's agent lock.
  - Refresh the live model name and effective temperature. Session temperature overrides still win over provider profile temperature.
  - Add regression coverage for successful refresh, refresh failure after a saved write, temperature refresh, session-override precedence, and provider-profile delete refresh.
- **Scope boundary:** This does not add a zerocode picker or title-click model switch UI. It only makes saved provider-profile config edits propagate to already-open sessions when those sessions can be rebuilt safely.
- **Blast radius:** `config/set` and `config/delete` for model provider profiles now schedule a scan of active sessions and may rebuild matching live provider boxes. Non-provider config edits are unchanged. Failed rebuilds leave the existing live session intact and log a warning.
- **Linked issue(s):** Closes #7395
- **Labels:** `bug`, `risk: high`, `size: M`, `agent`, `runtime`

  - `cargo test -p zeroclaw-runtime config_set -- --nocapture` - passed: 7 tests passed, 0 failed.
  - `cargo test -p zeroclaw-runtime config_delete_provider_temperature_refreshes_matching_live_session -- --nocapture` - passed: 1 test passed, 0 failed.
  - `cargo fmt -p zeroclaw-runtime -- --check` - passed.
  - `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` - passed.
  - `git diff --check` - passed.
  - `zc-cargo-broad --lane workspace-check clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings` - passed; the broad workspace-check timing log records a successful run for this branch after the final pushed commit.
- **Beyond CI - what did you manually verify?** The regression tests create an active session using `openai.test-provider`, update that provider profile through `config/set`, and verify the live agent refreshes from `old-model` to `new-model`. They also verify provider temperature edits reach the live session, session temperature overrides remain authoritative, refresh failures do not fail the saved config write, and `config/delete` refreshes a cleared provider temperature.
- **If any command was intentionally skipped, why:** CI-shaped local `nextest` was attempted after the final push, but no usable local passing result is available to cite. The current GitHub Quality Gate `Test` job is green on this head, so this PR relies on GitHub CI for full-workspace test coverage while citing focused runtime tests and broad local clippy as local evidence.

- If any `Yes`, describe the risk and mitigation: Not applicable.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: Existing configs, CLI calls, and environment variables do not need migration.

- **Fast rollback command/path:** `git revert 9ac4fa02ef3126c09d2c0e973419f7573bfce7b3 de5040f262588f54b56264be616341721588cb2b`
- **Observable failure symptoms:** After a provider-profile config save or delete, active sessions using that provider may fail to pick up the new provider wiring, model name, or effective temperature. Logs may show `config/set saved provider profile but live session refresh failed` when the saved config cannot build a valid provider for an active session.

- If `No`, why: No superseded PR or incorporated contributor work.

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/agent.rs
+++ b/crates/zeroclaw-runtime/src/agent/agent.rs
@@ -883,6 +883,11 @@ impl Agent {
         self.temperature = temperature;
     }
 
+    #[cfg(test)]
+    pub fn temperature_for_test(&self) -> Option<f64> {
+        self.temperature
+    }
+
     pub fn set_model_name(&mut self, model_name: String) {
         self.model_name = model_name;
     }
--- a/crates/zeroclaw-runtime/src/rpc/dispatch.rs
+++ b/crates/zeroclaw-runtime/src/rpc/dispatch.rs
@@ -263,6 +263,17 @@ fn not_yet_implemented(method: Method) -> RpcResult {
     ))
 }
 
+fn model_provider_ref_from_provider_profile_prop(prop: &str) -> Option<String> {
+    let rest = prop.strip_prefix("providers.models.")?;
+    let (provider_type, rest) = rest.split_once('.')?;
+    let (provider_alias, field) = rest.split_once('.')?;
+    if provider_type.is_empty() || provider_alias.is_empty() || field.is_empty() {
+        None
+    } else {
+        Some(format!("{provider_type}.{provider_alias}"))
+    }
+}
+
 /// Per-connection dispatcher. Shared state lives in [`RpcContext`].
 pub struct RpcDispatcher {
     ctx: Arc<RpcContext>,
@@ -2173,6 +2184,7 @@ impl RpcDispatcher {
 
     async fn handle_config_set(&self, params: &Value) -> RpcResult {
         let req: ConfigSetParams = parse_params(params)?;
+        let refresh_model_provider_ref = model_provider_ref_from_provider_profile_prop(&req.prop);
         {
             let mut config = self.ctx.config.write();
             config.ensure_map_key_for_path(&req.prop);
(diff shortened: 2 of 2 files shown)


WareWolf-MoonWall (approved):
Reviewed: head `9ac4fa02`, 2 commits, files `crates/zeroclaw-runtime/src/agent/agent.rs` (+5) and `crates/zeroclaw-runtime/src/rpc/dispatch.rs` (+349). Risk: high. Closes #7395.

### 🟢 What looks good — DRY compliance: no duplicate state introduced

The refresh path reads live config through `ctx.config.read()` at call time — it does not store a provider snapshot in the session or the dispatcher. `build_session_model_provider` is called fresh from the `Arc<RwLock<Config>>` so the rebuilt provider reflects the just-saved config state. Temperature is resolved on demand from the provider profile at refresh time; session overrides are read from `ctx.sessions.get_overrides()` at call time. No cached copy of provider data is stored anywhere.

### 🟢 What looks good — best-effort fire-and-forget scheduling is the correct shape

Scheduling the refresh with `zeroclaw_spawn::spawn!` after `flush_config()` returns means `config/set` and `config/delete` respond immediately without blocking behind an active agent turn's lock. Failed rebuilds log a warning and leave the existing live session intact, so a bad provider edit doesn't break the active session. The test `config_set_provider_refresh_failure_does_not_fail_saved_write` pins this invariant.

### 🟢 What looks good — `model_provider_ref_from_provider_profile_prop` is correctly scoped

The prop-path parser correctly extracts `{provider_type}.{provider_alias}` from `providers.models.{type}.{alias}.{field}` and returns `None` for non-provider props. This is the right signal to avoid scheduling a refresh on unrelated config edits.

### 🟢 What looks good — session-override precedence is tested and correct

`overrides.temperature.or(provider_temperature)` means the session override wins. The test `config_set_provider_refresh_preserves_session_temperature_override` verifies this explicitly with a `0.6` session override versus a `0.4` provider temperature update.

### 🟢 What looks good — test coverage is appropriate for risk tier

The 5 regression tests cover: successful refresh, refresh failure isolation, temperature refresh, session-override precedence, and delete-triggered refresh. This is exactly the matrix you'd want for a background-async state update on a high-risk path.

### 🟡 Non-blocking — `temperature_for_test` accessor is `#[cfg(test)]` only

The test-only accessor on `Agent::temperature` is correctly gated. No production API surface added.

Rollback SHAs are provided (`de5040f2` + `9ac4fa02`). Observable failure symptoms are clearly described. This is the right fix for #7395.

Approving.

PR #7437, opened 2026-06-09
fix(channels/wechat): make wechat channel can handle streaming chat response.

Taswen:
## Summary

- **Base branch:** `master` 
- **What changed and why:** The WeChat channel previously could not handle streaming chat responses from providers, causing JSON parsing failures when providers returned SSE format instead of complete JSON. This fix implements draft-related interfaces (`supports_draft_updates`, `send_draft`, `update_draft`, `finalize_draft`, `update_draft_progress`) to enable streaming response handling in the WeChat channel.
- **Scope boundary:** Changes are limited to the WeChat channel implementation in `crates/zeroclaw-channels/src/wechat.rs`. No other channels or system components are affected.
- **Blast radius:** Users of the WeChat channel who have configured streaming-capable providers.
- **Linked issue(s):** None.
- **Labels:** `bug`, `channel`, `risk:medium`, `size:XS`

  **Format check:**
  ```bash
  cargo fmt --all -- --check
  ```
  Output: No output (success)
  
  **Clippy:**
  ```bash
  ```
  Output:
  ```
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 02s
  ```
  
  **Tests:**
  ```bash
  cargo test --lib wechat
  ```
  Output: WeChat channel tests execute without failures (compilation successful)

- **Beyond CI — what did you manually verify?** (scenarios, edge cases, what you did NOT verify)
  - ✅ Verified compilation of the draft interface implementations
  - ✅ Confirmed that `supports_draft_updates()` returns `true` to force streaming path
  - ✅ Verified placeholder message sending is disabled (commented out) to avoid chat pollution
  - ✅ Confirmed clippy passes with no warnings
  - ✅ Actual runtime testing with WeChat iLink Bot API
  - ✅ End-to-end streaming with a real provider
  - ❌ NOT verified: Edge cases like network failures during draft updates

  - Full test suite skipped due to nextest installation requirement (`--locked` flag needed). However, targeted compilation and linting were successful.

## Compatibility (required)

  - No upgrade steps required. The change is purely additive (implementing optional trait methods) and backward compatible.

- **Fast rollback command/path:** `git revert 98763f369d6825e880234e822fd2050f9480d7e0`
- **Observable failure symptoms:** WeChat replies fail to stream or complete through iLink, typing indicators continue after cancellation, or WeChat media/image replies fail to display/decrypt.

---

### Technical Details

The WeChat channel now implements the draft message interface to support streaming responses:

1. **`supports_draft_updates()` → `true`**: Forces the orchestrator to use streaming path
2. **`send_draft()`**: Creates a draft_id without sending a placeholder message (WeChat doesn't support message editing/deletion)
3. **`update_draft()`**: No-op (WeChat doesn't support message editing)
4. **`finalize_draft()`**: Sends the final accumulated response
5. **`update_draft_progress()`**: Uses typing indicator for progress feedback

### Design Decision

Messages with placeholder are not used because WeChat iLink API doesn't support message editing or deletion. Sending a placeholder would create irrevocable spam in the chat. If WeChat adds edit/revoke support in the future, the placeholder can be re-enabled.

### Root Cause

The issue occurred when providers returned SSE streaming format (`data: {...}\n\n`) even when `stream: false` was requested. The non-streaming code path expected complete JSON, causing parsing failures. By implementing draft support, we force use of the streaming path which correctly handles SSE format.


Audacity88 (changes requested):
I reviewed the current head, 70db60a98a6348c9a9a031597d8d711dc21b2aa8, against `master`: this is a one-file WeChat channel change in `crates/zeroclaw-channels/src/wechat.rs`. I checked the PR metadata/checks snapshot, existing comments/reviews, the diff, the existing WeChat `send` delivery path, and the orchestrator draft lifecycle. There are no prior reviews or comments, and all visible CI is green. I did not run local WeChat/iLink runtime validation.

I don't think this is ready yet. The direction makes sense, but opting WeChat into draft mode changes the final delivery path for every response, and the new finalize/cancel hooks do not preserve all existing channel behavior yet.

### 🟢 What looks good — Streaming path matches the platform constraint

The no-placeholder design is the right direction for WeChat's no-edit/no-revoke behavior, and using the draft lifecycle keeps the final user-visible message to one send while allowing the orchestrator to consume streaming/SSE responses correctly.

### 🔴 Blocking — Draft finalization drops attachment replies

Normal WeChat `send` strips tool-call tags, parses attachment markers, sends any text prefix, uploads the attachment, and handles path-only attachment replies. After `supports_draft_updates()` returns `true`, successful channel responses finalize through `finalize_draft`; this implementation calls `send_text` directly. That means any assistant response that previously delivered an image or document via the existing WeChat attachment markers will now be sent as plain text instead, and path-only file replies will also bypass the attachment path.

Because this PR routes all WeChat replies through draft finalization to fix streaming, `finalize_draft` needs to preserve the same delivery semantics as `send` — for example by delegating to `send(&SendMessage::new(content, recipient))` or sharing the same attachment-handling helper — while still stopping typing afterward.

### 🟡 Warning — Cancelled drafts leave the typing loop running

`update_draft_progress` starts the WeChat typing loop, but the PR does not override `cancel_draft`. On cancellation, the orchestrator calls `cancel_draft` instead of `finalize_draft`, so the default no-op leaves the `typing_handle` task alive and continuing to call `sendtyping` every 4 seconds until something else happens to stop it. Please add a WeChat `cancel_draft` implementation that stops typing.

singlerider (commented):
Reviewed at head `70db60a` against current `master`. The streaming approach fits WeChat's constraint, but I verified @Audacity88's two regressions in the delivery path. Agreeing with his changes-requested.

### 🟢 The no-placeholder draft approach matches WeChat's platform constraint

WeChat iLink can't edit or revoke messages, so accumulating streaming deltas and sending exactly one final message (with the typing indicator standing in for progress) is the right way to consume an SSE/streaming response without a placeholder. Opting into the draft lifecycle for that is sound in principle.

### 🔴 Blocking — confirmed: `finalize_draft` drops attachment replies

Routing all WeChat replies through draft finalization means successful responses now land via `finalize_draft`, which calls `send_text(recipient, &content, ...)` directly (line ~2401). But the normal `send()` path does more: it runs `parse_attachment_markers` and, when attachments are present, sends the text prefix then uploads each attachment via `send_attachment`, and it also handles path-only file replies via `parse_path_only_attachment` → `send_attachment`. `finalize_draft` skips all of that, so any assistant response that previously delivered an image or document now gets flattened to plain text — the marker shown literally, no upload. Delegate `finalize_draft` to the same path: `self.send(&SendMessage::new(content, recipient)).await?` (which carries the full marker/attachment handling), then `stop_typing`.

### 🔴 Blocking — confirmed: cancelled drafts leave the typing loop running forever

`update_draft_progress` starts the WeChat typing loop (`start_typing` → `typing_handle`), and that handle is only stopped by `stop_typing`, which is called only in `finalize_draft`. On cancellation the orchestrator calls `cancel_draft`, which this PR does not override, so the default no-op leaves the `typing_handle` task alive and firing `sendtyping` every 4s indefinitely. Add a WeChat `cancel_draft` that calls `stop_typing(recipient)`.

The streaming design is right; `finalize_draft` just needs to preserve `send`'s attachment semantics and `cancel_draft` needs to stop typing. Leaving @Audacity88's changes-requested in force.

crates/zeroclaw-channels/src/wechat.rs
```diff
@@ -2363,6 +2367,53 @@ impl Channel for WeChatChannel {
...
+    async fn finalize_draft(
+        &self,
+        recipient: &str,
+        _draft_id: &str,
+        content: &str,
+    ) -> anyhow::Result<()> {
+        // Send the final accumulated response
+        let context_token = self.get_context_token(recipient);
+        let content = crate::util::strip_tool_call_tags(content);
+        let _ = self
```
singlerider: 🔴 `send_text` here drops attachments. The normal `send()` runs `parse_attachment_markers` (sends text + uploads each attachment) and `parse_path_only_attachment`, none of which this does — so image/document/path-only replies become plain text. Delegate to `self.send(&SendMessage::new(content, recipient)).await?` then `stop_typing`. Also: add a `cancel_draft` override that calls `stop_typing`, or the typing loop runs forever on cancellation.

Taswen:
The code has been updated to use `send` for the final message, alongside an implementation of `cancel_draft` to resolve the infinite `send_typing` loop. 

Additionally, during debugging, I fixed a bug where WeChat clients failed to display sent images. The issue occurred because media files were uploaded without hex-encoding, whereas the client expects to hex-decode the base64-decoded content.

Audacity88 (approved):
Reviewed current head 98763f369d6825e880234e822fd2050f9480d7e0 against the live PR state, current `master`, the prior reviews, the inline thread, the WeChat send/draft lifecycle, the orchestrator draft cancellation/finalization path, and the media AES-key path. I did not run local WeChat/iLink validation; this review relies on the author's reported iLink runtime smoke plus passing required CI/check rows and source review.

### ✅ Resolved — Final draft delivery now preserves normal WeChat sends

My prior blocker and @singlerider's matching inline comment were about `finalize_draft` bypassing the normal WeChat `send()` path. The current head now finalizes through `self.send(&SendMessage::new(...))`, so streamed replies still strip tool-call tags, parse attachment markers, handle path-only attachment replies, send any text prefix, and upload media through the existing attachment path. That preserves the behavior that direct `send_text` would have lost.

### ✅ Resolved — Draft cancellation now stops typing

The current head also adds `cancel_draft` and routes it to `stop_typing(recipient)`. That covers the cancellation paths in the orchestrator where a newer inbound message, hook suppression, or tool-loop cancellation cancels the draft instead of finalizing it, so the previous indefinite `sendtyping` loop is no longer left running by the default no-op.

### 🟢 What looks good — The WeChat-specific streaming shape fits the platform

WeChat iLink does not give us safe edit/revoke semantics for a placeholder message, so this no-placeholder draft approach is the right shape: opt into the streaming provider path, ignore text draft edits, use typing as progress, and send exactly one final accumulated response. The added CDN `aes_key` encoding change is also consistent with the parser/test path: outbound media now sends base64(hex(key)), while inbound parsing still accepts plain hex, base64(raw bytes), and base64(hex bytes).

### 🟡 Warning — Keep the channel smoke evidence attached before merge

This is still channel runtime behavior, so merge confidence depends on real or near-real WeChat/iLink evidence. The PR body says actual iLink API runtime testing and end-to-end streaming with a real provider were done, and I am accepting that as enough to clear my code-review block. I did not reproduce it locally. If the branch is refreshed or the channel/orchestrator surface changes again before merge, keep that live/near-live streaming plus media-delivery smoke evidence current.

Issue #6646, opened 2026-05-14
[Bug]: Title: web_search_tool and web_fetch not firing via Telegram channel in v0.7.5


icemann521:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

ZeroClaw version: 0.7.5
Channel: Telegram
Provider: Custom LM Studio (OpenAI-compatible)
Model: qwen/qwen3.6-35b-a3b

### Expected behavior

Current behavior: 
When asking Clawconsigliere to search for real-time data 
via Telegram, no tools fire at all. The agent hallucinates 
an answer instead. Same request works perfectly via CLI 
and web dashboard.

Related issues: #4083, #848

### Steps to reproduce

```bash
Steps to reproduce:
1. Configure Telegram channel
2. Add web_search_tool and web_fetch to auto_approve
3. Send message asking for current stock price
4. Agent responds with hallucinated data — no tool activity 
   in journalctl logs
```

### Impact

Unable to use web_search_tool or web_fetch via Telegram 
channel. Agent hallucinates answers instead of using tools. 
Same requests work correctly via CLI and web dashboard. 
Makes Telegram channel useless for any real-time data queries.

### Logs / stack traces

```text
No tool activity appears in journalctl logs when requests 
are made via Telegram. Running:
journalctl -u zeroclaw -n 20 --no-pager | grep -i "search\|tool"
returns no output during Telegram requests, confirming 
tools are not firing at all.
```

### ZeroClaw version

0.7.5

### Rust version

rustc 1.95.0 (59807616e 2026-04-14)

### Operating system

Raspberry Pi OS Lite 64-bit (aarch64) Raspberry Pi Zero 2 W (512MB RAM)

### Regression?

Yes, it worked before

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

Audacity88:
Triage read: this looks related to #4083 and #848, but not a duplicate. Those are older closed reports, while this is a current v0.7.5 Telegram-channel regression where the same real-time query works through CLI and the web dashboard but does not trigger `web_search_tool` / `web_fetch` from Telegram.

[...]

PR #7438, opened 2026-06-09
fix(channels): telegram delivery prompt no longer discourages tool use (#6646)

IftekharUddin:
## Summary

- **Base branch:** `master`
- **What changed and why:** The Telegram delivery-instructions block told the model to "use tool results silently" and to "answer the latest user message directly." On a smaller OpenAI-compatible model (the reporter runs qwen3 through LM Studio) that wording reads as "answer from what you already know, don't call tools." So over Telegram `web_search_tool` and `web_fetch` never fire, the agent makes up an answer, and nothing shows up in the logs. That block is injected only for Telegram, which is exactly why the same query works over the CLI and the web dashboard but fails here. I reworded it to keep the original point (don't narrate tool-execution bookkeeping) while making it explicit that the model should reach for `web_search_tool`/`web_fetch` when a question needs current or external data instead of guessing.
- **Why it's the prompt and not the tool plumbing:** I chased the plumbing first and ruled it out. The CLI and the channel both run through `run_tool_call_loop` and resolve to the same provider instance (`get_or_create_provider` hands back the cached `ctx.model_provider` for the default route), the streaming path captures native tool calls correctly (`consume_provider_streaming_response` handles `StreamEvent::ToolCall`), and `excluded_tools`, approval, and `strict_tool_parsing` all sit at defaults that leave tools enabled. A stock OpenAI-compatible provider is native (`native_tool_calling = !merge_system_into_user`), so the tools are sent on the wire. The only tool-relevant thing that differs for Telegram specifically is this instruction block.
- **Scope boundary:** Only the Telegram instruction string and a unit test that covers it. The tool dispatch loop, registry assembly, approval flow, provider layer, and the other channels' instruction blocks are untouched.
- **Blast radius:** Telegram system prompt only. It's one static `&'static str`, so there's no change to runtime, config, or any other channel's behavior.
- **Linked issue(s):** Related #6646 (prompt hardening; the issue stays open for reporter/maintainer re-verification on current `master` rather than auto-closing on merge)
- **Labels:** `bug`, `channel`, `channel:telegram`, `runtime`, `tool`, `tool: web_search_tool`, `tool: web_fetch`, `priority:p1`. The change itself is a one-line prompt edit plus a test, so `risk: low` / `size: S` fit better than the issue's original `risk: high`.

```bash
cargo fmt --all -- --check
# clean, no output

# Finished `dev` profile ... in 54.33s, no warnings

cargo test -p zeroclaw-channels --lib
# test result: ok. 715 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
# includes channel_delivery_instructions_for_telegram_encourage_tool_use ... ok
```

- **Commands run and tail output:** above. `fmt` ran across the whole workspace; clippy and tests are scoped to `zeroclaw-channels` since that's the only crate touched and the new test lives there.
- **Beyond CI — what did you manually verify?** The new test fails against the old string (both the positive checks and the guard against the old phrasing) and passes after the edit, so it's a real two-way regression guard rather than a tautology.
- **If any command was intentionally skipped, why:** I did not re-run the full-workspace clippy/test, since nothing outside `zeroclaw-channels` changed. The one thing I could not verify locally is the end-to-end behavior on a live Telegram + LM Studio/qwen3 setup. The last link (this wording causes this model to skip tools) is model behavior, not something a unit test can prove, so a live smoke check on the reporter's config is the right way to close it out.

## Compatibility (required)

- Backward compatible? Yes
- Upgrade steps: none. The new wording takes effect the next time the daemon starts.

Low risk, so `git revert <sha>` is the plan.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -708,7 +708,12 @@ fn channel_delivery_instructions(channel_name: &str) -> Option<&'static str> {
              - Structure longer answers with bold headers, not raw markdown ## headers\n\
              - For media attachments use markers: [IMAGE:<path-or-url>], [DOCUMENT:<path-or-url>], [VIDEO:<path-or-url>], [AUDIO:<path-or-url>], or [VOICE:<path-or-url>]\n\
              - Keep normal text outside markers and never wrap markers in code fences.\n\
-             - Use tool results silently: answer the latest user message directly, and do not narrate delayed/internal tool execution bookkeeping.",
+             - When a question needs current, real-time, or external information \
+               (prices, news, weather, web pages, lookups, etc.), use your tools — \
+               e.g. web_search_tool and web_fetch — to obtain it before answering; \
+               never guess or answer from memory alone when a tool can verify it.\n\
+             - Present the final answer to the latest user message directly from the \
+               tool results, without narrating delayed/internal tool-execution bookkeeping.",
         ),
         "qq" => Some(
             "When responding on QQ:\n\
@@ -16324,6 +16329,44 @@ BTC is currently around $65,000 based on latest tool output."#
         assert!(
             block.contains("[IMAGE:<absolute-path>]"),
             "discord block must show the absolute-path marker form"
+        );
+    }
+
+    // Regression guard for #6646: web_search_tool / web_fetch never fired via
+    // the Telegram channel because the delivery-instructions block told the
+    // model to "answer the latest user message directly" and to "use tool
+    // results silently". On a small OpenAI-compatible local model that reads as
+    // "answer from memory, do not call tools", so the agent hallucinated with
+    // zero tool activity — while the identical query worked via CLI and the web
+    // dashboard, which do not inject this Telegram-only block. The block must
+    // encourage tool use for real-time/external information and must not tell
+    // the model to answer directly *instead of* using tools.
+    #[test]
+    fn channel_delivery_instructions_for_telegram_encourage_tool_use() {
+        let block = channel_delivery_instructions("telegram")
+            .expect("telegram channel must have a delivery-instructions block");
+        assert!(
+            block.contains("When responding on Telegram:"),
+            "telegram block must identify itself"
+        );
+        // Positive: it must actively steer the model toward its tools for
+        // real-time/external information.
+        assert!(
+            block.contains("use your tools"),
+            "telegram block must instruct the model to use its tools"
+        );
+        assert!(
+            block.contains("web_search_tool") && block.contains("web_fetch"),
+            "telegram block must name the real-time tools so the model knows to reach for them"
+        );
+        assert!(
+            block.contains("never guess or answer from memory alone"),
+            "telegram block must forbid answering from memory when a tool can verify"
+        );
+        // Negative: the exact regressed phrasing must never come back.
+        assert!(
+            !block.contains("Use tool results silently: answer the latest user message directly"),
+            "telegram block must not tell the model to answer directly instead of using tools (#6646)"
         );
     }
 


Audacity88 (changes requested):
Reviewed current head `bfaffffb7e01d307445db146fa1d9659bc4ef58e`, the linked #6646 report and follow-up comments, the one-file diff in `crates/zeroclaw-channels/src/orchestrator/mod.rs`, the channel/provider/tool-loop source around Telegram prompt injection and native compatible-provider tool calls, and the current visible CI. CI is green, `git diff --check origin/master...origin/pr-7438-review` is clean, and a local `merge-tree` check applies this over current `origin/master` without conflicts. I did not run additional local cargo validation or a live Telegram/LM Studio smoke test. Current `master` has touched the same orchestrator file since this PR's base, so final merge readiness still needs a current-head branch/CI refresh or equivalent merge-result evidence.

### 🟢 What looks good — The prompt change targets the only Telegram-specific difference

The direction of the patch makes sense. The Telegram delivery block is the only channel-specific instruction in this path, and the new wording preserves the useful "do not narrate tool bookkeeping" rule while explicitly steering the model toward `web_search_tool` / `web_fetch` for current or external information. Keeping the change to the static Telegram instruction block also avoids disturbing the provider factory, registry assembly, approval flow, or other channel prompts.

### 🔴 Blocking — Do not close #6646 without current Telegram-boundary verification

This PR still claims `Closes #6646`, but the validation does not prove the reported bug is fixed on current ZeroClaw. The issue was explicitly punted to v0.8.1 because the v0.7.5 Telegram + OpenAI-compatible report needed re-verification on current `master`, and the PR body says the final end-to-end behavior was not verified locally.

The new unit test is a useful regression guard for the string itself, but it only proves that the prompt contains the new words and no longer contains the exact old sentence. It does not prove that a Telegram turn with a compatible provider now emits `web_search_tool` / `web_fetch`, and it does not prove the reporter's qwen/LM Studio failure mode is gone.

Please either add current-head evidence that matches the claim, or narrow the PR relationship so the issue stays open for confirmation. The strongest proof would be a live or near-live Telegram-compatible-provider smoke showing a current-data Telegram prompt triggers `web_search_tool` or `web_fetch` on this head. If live credentials are not available, the channel orchestrator already gives a lower-cost automated boundary to test: drive a Telegram `ChannelMessage` through `process_channel_message` with `web_search_tool` / `web_fetch` registered, capture the provider `ChatRequest`, and assert the provider sees both the revised Telegram prompt and those tool specs. That would not prove the model's behavior by itself, but it would at least lock the current-master channel/provider boundary that #6646 depends on.

Until one of those paths is done, this should not auto-close #6646. A safe alternative is to change the PR body from `Closes #6646` to `Related #6646`, record this as prompt hardening, and leave the bug open for reporter or maintainer re-verification.

IftekharUddin:
@Audacity88 agreed. I looked into it and I'm unsure if I can timely provide the required tests so I've changed the marker from closes to related.

Audacity88 (approved):
Reviewed current head `bfaffffb7e01d307445db146fa1d9659bc4ef58e`, the updated PR body and author reply, the linked #6646 issue thread, the one-file diff in `crates/zeroclaw-channels/src/orchestrator/mod.rs`, current visible checks, and the current `origin/master` freshness state. CI is green on the branch. I did not run additional local cargo validation or a live Telegram/LM Studio smoke test. I did run a fresh merge check: `git diff --check origin/master...origin/pr-7438-review` is clean, and `git merge-tree origin/master origin/pr-7438-review` applies cleanly.

### ✅ Resolved — #6646 no longer auto-closes without current Telegram verification

The blocking item from my previous review is resolved. The PR body now uses `Related #6646`, and the author explicitly confirmed that the issue should stay open because the deeper Telegram/LM Studio verification is not available in this PR. That matches the actual evidence: this patch hardens the Telegram prompt and adds a regression guard for the prompt text, but it does not prove the reporter's live Telegram + OpenAI-compatible provider path is fixed.

### 🟢 What looks good — The patch is appropriately narrow

The code change still targets the only Telegram-specific instruction block in this path. The new wording keeps the useful "do not narrate tool bookkeeping" behavior, while making it explicit that current or external-information questions should use `web_search_tool` / `web_fetch` instead of guessing from memory. The test is also useful for this scope because it guards both the new positive wording and the old regressed phrase.

### 🟡 Warning — Approval is not final merge readiness

This clears my prior blocker, but I am not treating the PR as merge-ready yet. Current `master` has touched the same orchestrator file since this PR's base, and the branch CI does not cover the current merge-result state. The merge-tree is clean, so I do not see a code conflict, but final merge readiness should still use a current-head branch/CI refresh or equivalent merge-result evidence. #6646 should also remain open for reporter or maintainer live verification after this prompt-hardening slice lands.

singlerider (approved):
Reviewed at head `bfaffff` against current `master`. The fix is well-targeted and correctly scoped. Approving.

### 🟢 The diagnosis and the reword are right

The old Telegram delivery block said "Use tool results silently: answer the latest user message directly," which a small OpenAI-compatible local model reads as "answer from memory, do not call tools" — so the agent hallucinated with zero tool activity on Telegram while the identical query worked via CLI and the web dashboard (which don't inject this Telegram-only block). The new wording keeps the useful "don't narrate tool bookkeeping" behavior but explicitly steers the model toward `web_search_tool` / `web_fetch` for current/real-time/external information and forbids guessing from memory when a tool can verify. That directly addresses #6646.

### 🟢 Scope is tight and the regression is well-constructed

The change touches only the Telegram arm of `channel_delivery_instructions` — no other channel's block is modified. The regression test guards both directions: the positive wording (`use your tools`, `web_search_tool`, `web_fetch`, `never guess or answer from memory alone`) and a negative assertion that the exact regressed phrase (`Use tool results silently: answer the latest user message directly`) never returns. That's the right guard for a prompt-text fix.

CI is green and the branch is `MERGEABLE` / `CLEAN`. @Audacity88 cleared his prior block (the PR now uses `Related #6646` rather than auto-closing it) and approved.

### 🟡 Non-blocking (per @Audacity88)

`master` has touched this orchestrator file since the PR's base; the merge-tree is clean, but a current-head CI refresh before merge is worth it. And #6646 should stay open for the reporter/maintainer to do the live Telegram + OpenAI-compatible-provider verification this prompt-hardening slice can't prove on its own.

Approving — correct diagnosis, tight scope, solid regression guard.

Issue #5808, opened 2026-04-16
[Feature]: defer built-in tool schemas to reduce the fixed prompt floor


JordanTheJet:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

With the default `agent.max_context_tokens = 32000` (set at `crates/zeroclaw-config/src/schema.rs:1511`), the **first** LLM iteration of a fresh conversation already exceeds budget by ~3.3x — purely from the system prompt + tool definitions + a single user message:

```
INFO Preemptive context trim: estimated tokens exceed budget estimated=106818 budget=32000 iteration=1
INFO Preemptive context trim: estimated tokens exceed budget estimated=106872 budget=32000 iteration=2
INFO Preemptive context trim: estimated tokens exceed budget estimated=106948 budget=32000 iteration=3
INFO Preemptive history prune applied collapsed=0 dropped=2
INFO Preemptive context trim: estimated tokens exceed budget estimated=106982 budget=32000 iteration=4
```

Because the trim runs on every iteration and drops messages from history, the agent loses prior tool results between iterations and **re-runs the same tool calls in a loop**:

```
Iteration 1: ls -la ~/
Iteration 2: ls -la /Users/jordantian/
Iteration 3: ls -la /Users/jordantian/.zeroclaw/workspace
Iteration 4: ls -la /Users/jordantian/.zeroclaw/workspace  (same again)
```

The agent eventually converges only because the context-compressor falls back to fewer tools / smaller system prompt — not because progress is made.

### Expected behavior

Either:
1. The default `max_context_tokens` is raised to match the actual minimum budget required to fit `system_prompt + all_tool_defs + 1 user message + 1 assistant turn` (current empirical: ≥128k), OR
2. The default tool surface is narrowed (right now ~107k of tool defs are shipped by default — likely the 50+ integrations registered eagerly), OR
3. When iteration 1 already exceeds budget, the runtime emits a `WARN` once with a remediation hint (e.g. *"system prompt + tool defs alone exceed your context budget; raise `agent.max_context_tokens` or disable unused integrations"*), instead of silently trimming and looping.

### Steps to reproduce

```bash
# Default config (no overrides for max_context_tokens)
RUST_LOG=zeroclaw_runtime=info ./target/release/zeroclaw agent -m "list files in my home dir, then summarize"
# Observe `Preemptive context trim: estimated tokens exceed budget` on every iteration
```

### Impact

[...]

PR #7440, opened 2026-06-09
fix(runtime): surface remediation when system-prompt floor exceeds context budget (#5808)

IftekharUddin:
> **Maintainer's note** (Audacity88): I updated the PR body wording, label snapshot, validation summary, and rollback command to match current head `2d33f2b06bbdf38842defcfed93d88a97e1e81b2` and the live labels. The contributor's original commits and implementation remain preserved in the branch history; authorship of the code change remains with @IftekharUddin.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **Rebased onto current `master`.** During the rebase it became clear that the original #5808 failure mode — a *per-iteration* preemptive trim that endlessly re-ran the same tool calls when the system prompt + inlined tool definitions formed an irreducible floor at or above the resolved context budget — was **independently fixed upstream** by the whole-turn-trim refactor. Preemptive trim now runs **once per turn** (`agent/turn/mod.rs`, gated on `iteration == 0`) and reactive recovery drops **whole turns** and *breaks* instead of looping (`agent/turn/context_recovery.rs`, `agent/loop_.rs`). The PR's original loop-skip code targeted a block that no longer exists, so it has been dropped.
  - **What this PR now delivers** is the operator experience that is still missing for the floor condition. In the benign case (the real provider window is adequate) nothing signals that `max_context_tokens` is misconfigured; in the hard-wall case the agent emitted a generic "cannot trim further" warning with no root cause or remedy.
  - Added `estimate_system_floor_tokens()` (the protected system-message floor that whole-turn trimming can never drop), single-sourced through a shared `estimate_message_tokens` helper so it stays in lock-step with `estimate_history_tokens`.
  - Surfaced an **actionable remediation** that names the measured system floor, the resolved effective budget, and the honored `[runtime_profiles.<name>] max_context_tokens` surface (or reducing the enabled tool surface), with `error_key=context_floor_exceeds_budget`, at the sites that observe the floor: the iteration-0 preemptive trim, the in-loop reactive recovery, and the interactive recovery path. The hint is **strictly gated** on the floor dominating the resolved budget, so ordinary single-large-turn overflows still get the generic message.
  - Added the `history-trim-floor-exceeds-budget` i18n key across all locales (en/es/fr/ja/zh-CN, ASCII-only).
- **Scope boundary:** Does **not** raise the default `max_context_tokens` or narrow the default tool surface (the budget-default question is entangled with #7100). No behavioral change to trimming itself — only observability/remediation is added.
- **Blast radius:** `crates/zeroclaw-runtime` agent context-management paths only. Trim/prune mechanics, provider request shape, and recovery control flow are untouched.
- **Linked issue(s):** Addresses #5808 (root-cause loop already mitigated upstream; this adds the missing remediation + detection). Related #7100.
- **Labels:** `bug`, `agent`, `runtime`, `risk:high`, `size:S`

```bash
cargo fmt --all -- --check                          # exit 0, no diff
cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings   # Finished, 0 warnings
cargo test -p zeroclaw-runtime --lib                # 2473 passed; 2 pre-existing parallel-flakes (pass in isolation)
```

  - `cargo fmt --all -- --check` -> exit 0 (no formatting diff)
  - `cargo clippy -p zeroclaw-runtime --lib --all-targets -- -D warnings` -> `Finished` in ~1m, no warnings
  - `cargo test -p zeroclaw-runtime --lib` -> `2473 passed; 2 failed; 2 ignored`. The two failures (`cost_usd_flows_to_usage_event_and_llm_response`, `remove_job_emits_structured_cron_delete_event`) are pre-existing flakes from shared global log-subscriber / cron state under parallel execution — **both pass when run in isolation** and are unrelated to this diff (it touches neither cost nor cron).
  - New unit tests: `estimate_system_floor_tokens` counts only `system` messages and returns 0 for empty/system-less histories (relocated to `history.rs`).
  - New #5808 regression test in `context_recovery.rs`: a single-turn history whose system floor dominates the budget returns `recovered == false` (no retry -> caller breaks, proving the old infinite loop cannot recur), and the floor is detectably dominant.
  - New formatter/log-capture coverage asserts the emitted `context_floor_exceeds_budget` event carries the resolved `budget`, measured `system_floor`, and remediation text that names `[runtime_profiles.<name>] max_context_tokens` without referencing `agent.max_context_tokens`.

## Compatibility (required)

- Config / env / CLI surface changed? **No** — no new config keys; existing runtime-profile `max_context_tokens` semantics are preserved. Only adds a WARN/ERROR log + one i18n string.
- Upgrade steps: None.

- **Fast rollback command/path:** `git revert 2d33f2b06bbdf38842defcfed93d88a97e1e81b2 bd1a7846f71b5a26a192faa968c07f4e18f322e9` before merge, or revert the eventual squash commit after merge.
- **Observable failure symptoms:** Absence of the `context_floor_exceeds_budget` remediation log when the system-prompt/tool-definition floor meets or exceeds the budget; the generic "cannot trim further" / context-overflow error returns with no remediation hint.


Audacity88 (changes requested):
I reviewed current head `16d808c` against #5808, the current green CI/checks, and the PR body validation. I also ran a focused local check: `zc-cargo test -p zeroclaw-runtime estimate_system_floor`, which passed.

### 🟢 What looks good — the guard targets the right failure mode

The implementation is aimed at the specific #5808 loop behavior rather than trying to solve model capability/default-budget policy here, which keeps the scope clean alongside #7100. Detecting the irreducible system/tool-definition floor before preemptive trimming is also the right architectural shape: once the protected floor already exceeds the configured budget, trimming ordinary history cannot make progress and can only destroy useful tool results.

### 🔴 Blocking — add a loop-level regression for the no-trim path

The new unit coverage proves that `estimate_system_floor_tokens` counts only system messages, but it does not prove the runtime loop actually preserves history when `system_floor >= context_token_budget`. If the guard in `run_tool_call_loop` stopped using the helper, or if that branch accidentally fell through to `fast_trim_tool_results` / `prune_history`, the current tests would still pass while #5808 could regress.

Because this is a high-risk runtime-loop fix for an S1-style repeated tool execution failure, please add one behavior-level regression around the branch in `run_tool_call_loop`: seed history so the system floor alone meets/exceeds the budget and include an older large tool result or assistant/tool-call pair that the old preemptive trim path would collapse/drop. Then run the loop with a recording/scripted provider and assert the provider receives the untrimmed history/tool result. Capturing the one-time warning would be a nice bonus, but the essential proof is that the no-trim branch preserves useful history under the exact #5808 condition.

Once that regression exists, the code shape here looks reasonable to me.

Nillth:
The iteration-top trim/prune block is now `preflight_history_maintenance` in `agent/turn/history_window.rs` — apply your guard there; it then protects every transport, not just the paths that hit this code today.

— Context: this is the #7415 turn-engine consolidation, now up as #7540 (not yet merged). The full function→file map is the **Migration Guide** section of that PR. Flagging early so your rebase lands cleanly.

singlerider (commented):
Reviewed at head `16d808c` against current `master`. The guard logic is correct and well-built; @Audacity88's loop-level regression ask is the right gate. Agreeing with his changes-requested.

### 🟢 The guard is the right shape and correctly wired

When preemptive trimming is triggered (`estimated > budget`), the fix now computes `system_floor = estimate_system_floor_tokens(history)` and, when that irreducible floor (system prompt + inlined tool definitions) already meets or exceeds the budget, skips trimming entirely — emitting a one-time `WARN` (gated by `context_floor_warned`, with `error_key = "context_floor_exceeds_budget"` and actionable guidance to raise `agent.max_context_tokens` or disable unused integrations) and explicitly noting it's skipping trim "to avoid re-running tool calls." Otherwise it falls through to the original `fast_trim_tool_results` + `prune_history` path. That precisely targets the #5808 loop: once the floor exceeds budget, trimming conversation history can't make progress and only destroys useful tool results, sending the agent back to re-run the same tool calls. The attribution on the WARN is done right (`with_outcome`, `error_key`, one-time gate). Keeping the model-capability / default-budget policy out of scope (deferred to #7100) is the correct boundary.

### 🔴 Blocking — agree: add a loop-level regression for the no-trim branch

The new unit test proves `estimate_system_floor_tokens` counts only system messages, but it does not prove `run_tool_call_loop` actually preserves history when `system_floor >= context_token_budget`. If that branch ever stopped using the helper or fell through to `fast_trim_tool_results` / `prune_history`, the current tests would still pass while #5808 regressed. For a high-risk runtime-loop fix to an S1-style repeated-tool-execution failure, add one behavior-level regression: seed history so the system floor alone meets/exceeds the budget plus an older large tool result (or assistant/tool-call pair) the old preemptive trim would collapse or drop, run the loop with a recording/scripted provider, and assert the provider receives the untrimmed history/tool result. Capturing the one-time WARN is a nice bonus, but the essential proof is that the no-trim branch preserves useful history under the exact #5808 condition.

The code shape is reasonable; it just needs that regression so the no-trim path can't silently break. Leaving @Audacity88's changes-requested in force.

Audacity88:
Thanks for carrying this slice for #5808. I rechecked PR #7440 head `16d808c6ff09556f1b1b809ede3e73f05eb80d3a` against the live PR state, linked issue #5808, prior reviews, @Nillth's #7540 migration note, merged #7540, the later #8196 history rewrite, and current `master` source context. I did not run local validation or re-test the original #5808 reproduction on current `master`.

### 🔵 Suggestion — Rebase or supersede this branch against the current turn path

This PR is still blocked on the loop-level regression from my earlier review, and the target has moved since then. #7540 has now landed the turn-engine consolidation that @Nillth flagged, and current `master` also includes the later history-trim rewrite from #8196. The active turn path now runs history maintenance through `crates/zeroclaw-runtime/src/agent/turn/history_window.rs` / `turn::preflight_history_maintenance`, and that file currently says no preemptive token-budget trimming runs there.

Given that, I do not think this branch is still a clean place to finish #5808 as-is. Please either rebase/retarget the fix onto the current turn-engine path and re-check whether #5808 still reproduces on current `master`, or close this PR as superseded if the old repeated-trim behavior is already gone after #7540/#8196. If the failure still reproduces, the replacement path should include the behavior-level regression requested earlier: prove that under the over-budget system-floor condition, the runtime preserves useful history/tool results instead of dropping them and re-running tool calls.

I am not changing the review state here; the existing `CHANGES_REQUESTED` still reflects the missing regression on this head.

IftekharUddin:
Force-pushed a rebase onto current `master` (`bd1a7846f`). Heads-up on a scope change surfaced by the rebase: the original infinite preemptive-trim loop this PR targeted was **independently fixed upstream** by the whole-turn-trim refactor (preemptive trim is now once-per-turn; reactive recovery drops whole turns and breaks instead of looping). The old loop-skip code targeted a block that no longer exists, so it's dropped.

What remains and is now the focus: the **operator remediation** for the system-prompt/tool-definition floor condition, which `master` still handles non-actionably. This PR adds `estimate_system_floor_tokens` + a gated `context_floor_exceeds_budget` remediation hint at the iteration-0 trim, the in-loop reactive recovery, and the interactive recovery path, plus a #5808 regression test asserting single-turn floor overflow does not retry. Updated title/description reflect the new scope. Now `MERGEABLE`.

Audacity88 (changes requested):
I reviewed current head bd1a7846f71b5a26a192faa968c07f4e18f322e9, the live PR body/checks/comments/reviews, issue #5808 including @dwillitzer's v0.8.2 verification, the earlier request-changes reviews, and the current diff in the runtime history/turn/context-recovery paths. I did not run local Cargo for this review-only pass; all visible GitHub checks are green.

### ✅ Resolved — the branch is now aimed at the remaining #5808 gap

@IftekharUddin, thanks for rebasing this after the turn-engine and whole-turn-trim changes. I agree with the new scope split: the original missing-user-turn / repeated-trim failure mode appears to be resolved by the current history-trim path, and this PR is now focused on the remaining operator-facing problem: when the system/tool floor dominates the configured budget, the runtime should make that condition actionable instead of leaving users with a generic trim failure.

That is the right direction for #5808 at this point. Keeping the default-budget and broad tool-surface policy out of this PR also keeps it cleanly separated from #7100.

### 🔴 Blocking — the remediation still does not identify the resolved budget

The remaining problem is the specific actionability gap @dwillitzer documented on #5808. The confusing case is not just "some context budget was exceeded"; it is that users need to know the effective budget the runtime actually used before they can tell whether their config change landed on the right surface.

The new remediation text still says only:

```text
system prompt and tool definitions alone exceed the context budget; raise agent.max_context_tokens or disable unused integrations
```

The structured log attrs include a `budget` number at the call sites, which is useful for machine logs, but the user-facing message and visible CLI error still do not name the resolved budget value. That leaves the #5808 discoverability gap in place: an operator can read the warning, change a config value, and still have no visible way to confirm whether the runtime is using the budget they meant to change.

Please make the remediation name the resolved budget value and avoid the ambiguous `agent.max_context_tokens` wording. For example, the message should make clear that the effective budget is `N` for this agent/runtime path, and that the fix is to raise the budget on the config surface the agent actually uses or reduce the enabled tool surface. If a path cannot cheaply name the profile, it should at least include the resolved budget in the visible text.

I would also like one regression assertion around the emitted remediation shape, in addition to the current helper and recovery coverage. The important behavior of this retargeted PR is the operator-facing warning, so a test should pin that the floor-dominates-budget path emits `context_floor_exceeds_budget` with the resolved budget and the corrected remediation text.

Once that warning points at the actual resolved budget in the visible remediation, this looks like the right residual #5808 slice.

IftekharUddin:
@Audacity88 — before I push this, I want to check the approach lands where you're pointing, on both the budget-naming and the config-surface wording. Does this match what you'd expect?

**Naming the resolved budget N in the visible text.** All three emission sites (iteration-0 preemptive trim, in-loop reactive recovery, interactive/turn-boundary) plus the stderr path route through one shared formatter, `context_floor_remediation(system_floor, budget)`, so the visible message and the structured log can't drift and both name the numbers:

> system prompt and tool definitions (`{floor}` tokens) alone meet or exceed the context budget (`{budget}` tokens); raise `[runtime_profiles.<name>] max_context_tokens` or reduce the tool surface by disabling unused integrations

The `budget` passed in is the resolved effective budget (`effective_max_context_tokens` / `context_token_budget`), not a raw config field, so the operator sees the actual N the runtime used.

**Dropping `agent.max_context_tokens`.** You flagged this as ambiguous, and it's also wrong against the current standard: `max_context_tokens` is only honored on `[runtime_profiles.<alias>]` (resolved via `effective_max_context_tokens`; the agent-inline copy is inert per #6877). So the corrected wording points at the runtime-profile surface, matching the house phrasing already used elsewhere in the trim path. Is `[runtime_profiles.<name>] max_context_tokens` the surface you want named, or would you prefer the concrete resolved profile alias threaded in?

**One coherence fix I want to confirm.** The in-loop reactive recovery currently gates on its local 2/3-of-current recovery budget, but I'm now displaying the resolved config budget. To avoid a message that says "floor ≥ N" while numerically `floor < N`, I'm aligning that branch's gate to `system_floor >= context_token_budget` (the same quantity it displays, matching the other two sites). This doesn't weaken the #5808 no-loop guarantee — the branch only logs and returns `false` either way. Flagging in case you'd rather keep the gate on the recovery budget and display that value instead.

**Regression.** Adding `floor_exceeds_budget_emits_event_with_resolved_budget_and_remediation`: it drives the floor-dominates reactive path under the log-capture subscriber and asserts the emitted event carries `error_key = context_floor_exceeds_budget`, `attributes.budget == <resolved budget>` (not the 2/3 value), `system_floor >= budget`, and a message naming N plus the `[runtime_profiles…]` surface. Plus a deterministic unit test on the shared formatter (contains N + floor and `[runtime_profiles`, not `agent.max_context_tokens`).

One scope check: only the loop_.rs path has a stderr `eprintln` today; the reactive/interactive sites are log-only and I'm keeping that (just fixing the message + structured attrs everywhere). The literal "visible CLI error names N" is satisfied by the stderr path — but want me to add stderr surfacing to the reactive path too, or is corrected structured logging enough there?

If this matches your intent I'll push and re-request review.

Audacity88:
@IftekharUddin yes, that matches what I was asking for.

For the visible remediation, use the resolved effective budget value and keep the message tied to the runtime-profile surface. `[runtime_profiles.<name>] max_context_tokens` is the right generic wording. I do not need the concrete resolved profile alias threaded through just for this warning unless it is already cheap and natural at the emission site; the important part is that the user sees the actual `N` the runtime used and is not pointed back at inert `agent.max_context_tokens` wording.

The shared `context_floor_remediation(system_floor, budget)` formatter is also the right shape. It keeps the human-facing message and structured attrs from drifting, and it gives us one place to pin the floor, budget, and config-surface wording.

On the in-loop reactive branch: aligning the predicate to the same `context_token_budget` value the message displays is the cleaner path. I do not want a warning that says the floor meets/exceeds `N` if the branch actually fired on a lower recovery-only threshold. Keeping the reactive path log-only is fine for this PR; please do not expand the stderr surface unless that path already has a user-visible error boundary you need to preserve.

The regression plan sounds right too: assert `error_key = context_floor_exceeds_budget`, the resolved budget attr, the corrected remediation text with the budget value, the floor value, and `[runtime_profiles...`, and make sure `agent.max_context_tokens` no longer appears in the emitted remediation.

After that lands, please re-request review and I will re-check the actual diff and CI.

IftekharUddin:
@Audacity88 @singlerider — the remaining blocker is addressed in `2d33f2b06` (already pushed; branch head).

The floor-exceeds-budget remediation now names **both** the measured floor and the **resolved effective budget** in the user-visible text, and points at `[runtime_profiles.<name>] max_context_tokens` (the honored surface under the #6877 tunable standard) instead of the inert `agent.max_context_tokens` knob:

```text
system prompt and tool definitions ({floor} tokens) alone meet or exceed the context budget ({budget} tokens); raise [runtime_profiles.<name>] max_context_tokens or reduce the tool surface by disabling unused integrations
```

All four emission sites (iteration-0 preemptive trim, turn-boundary trim, in-loop reactive recovery, and the `loop_.rs` stderr path) format through a single `context_floor_remediation(system_floor, budget)` helper, and the FTL key gained `{$floor}`/`{$budget}` args across all five locales so the human-facing text and structured attrs can't drift. The in-loop recovery predicate is now keyed on the resolved `context_token_budget` so the message can never claim the floor meets/exceeds N while numerically floor < N.

Regressions added: a formatter unit test pinning that the message names the budget, and a log-capture test driving the floor-dominates path that asserts the emitted event carries `error_key=context_floor_exceeds_budget`, the resolved budget attr, and the corrected remediation text. All CI green. Would appreciate a re-review when you have a moment.

Audacity88 (approved):
@IftekharUddin thanks for the follow-through here. I re-reviewed current head `2d33f2b06bbdf38842defcfed93d88a97e1e81b2` against the live PR state, issue #5808 and its latest narrowing comments, my prior `CHANGES_REQUESTED`, the latest author update, the current runtime/locales diff, and the green public CI. I did not run local Cargo validation or rerun the z.ai reproduction; this is source inspection plus the public CI state.

### ✅ Resolved — The remediation now names the resolved budget and the right config surface

The latest update routes the human-facing remediation through `context_floor_remediation(system_floor, budget)`, and the message now includes both measured values: the protected system/tool floor and the resolved effective context budget. It also points users at `[runtime_profiles.<name>] max_context_tokens` instead of the inert `agent.max_context_tokens` wording.

That resolves the remaining actionability blocker from my prior review. An operator who is still silently landing on the 32k fallback can now see the actual budget value the runtime used and the config surface that can change it.

### ✅ Resolved — The warning shape is pinned by focused regression coverage

The added log-capture regression drives the floor-dominates reactive path and checks `error_key = context_floor_exceeds_budget`, the resolved `budget` attr, the measured `system_floor`, and the corrected remediation text. The formatter test also pins that the text includes the budget/floor values and no longer references `agent.max_context_tokens`.

That is the right regression target for the retargeted #5808 slice: the original missing-user-turn loop is already covered by the current whole-turn trim path, and this PR now protects the remaining operator-facing warning contract.

### 🟢 What looks good — The implementation stays scoped to observability/remediation

The PR does not change the default budget, shrink the tool surface, or invent a second context-budget source of truth. It reuses the same message-token heuristic as `estimate_history_tokens`, checks the irreducible `system` floor at the turn-boundary and recovery sites, and keeps the normal whole-turn trim/recovery behavior intact for ordinary over-budget histories.

### 🟡 Warning — Refresh the branch before merge because `master` touched the same files

This approval clears my prior review block on the code at `2d33f2b06bbdf38842defcfed93d88a97e1e81b2`; it is not a merge-readiness statement. The branch is behind current `master`, and current `master` has touched the same runtime/history/locales files since this head's green CI. Before merge, please refresh the branch and let CI run on the updated head so the same-file drift is tested on the actual merge result.

### 🟡 Warning — Clean the generated squash body before merge

The branch still contains an older commit body with a `Co-Authored-By: Claude ...` trailer and stale `agent.max_context_tokens` wording. I am not asking for history rewrite on the contributor branch, but the final squash body should strip the bot/AI co-author trailer and keep the current runtime-profile wording so the public merge record matches the final PR.

IftekharUddin:
## CI fix update

I pushed commit `2ab9bb667` to the PR branch.

### Change 1: resolve the interactive recovery budget from the canonical runtime config

Needed because, after the branch update, `crates/zeroclaw-runtime/src/agent/loop_.rs` still referenced `eff_max_context_tokens`, but that local binding no longer exists in `run()`. That produced the CI-blocking compile error:

```text
error[E0425]: cannot find value `eff_max_context_tokens` in this scope
```

The source of truth is `agent.resolved.effective_context_budget()`. The fix materializes that value only inside the fallback block and reuses it for the predicate, structured log attr, remediation log text, and stderr remediation text.

Before:

```rust
let floor_exceeds_budget = system_floor >= eff_max_context_tokens;

"budget": eff_max_context_tokens,

crate::agent::history::context_floor_remediation(
    system_floor,
    eff_max_context_tokens,
)
```

After:

```rust
let context_token_budget = agent.resolved.effective_context_budget();
let floor_exceeds_budget = system_floor >= context_token_budget;

"budget": context_token_budget,

crate::agent::history::context_floor_remediation(
    system_floor,
    context_token_budget,
)
```

### Recovery flow

```mermaid
flowchart TD
    A["Provider reports context overflow"] --> B["Interactive outer recovery trims against model window"]
    B --> C{"Did whole-turn trim remove anything?"}
    C -->|Yes| D["Insert breadcrumb and retry turn"]
    C -->|No| E["Estimate protected system floor"]
    E --> F["Resolve budget from agent.resolved.effective_context_budget()"]
    F --> G{"system_floor >= resolved budget?"}
    G -->|Yes| H["Emit context_floor_exceeds_budget attrs and remediation"]
    G -->|No| I["Emit generic cannot-trim warning"]
```

### Why two budget concepts remain separate

The existing recovery trim still targets the provider/model context window with headroom. This patch only fixes the remediation predicate and reported budget to use the operator-facing resolved context budget.

```mermaid
flowchart LR
    M["agent.resolved.model_context_window"] --> R["recovery_budget = window * 9 / 10"]
    B["agent.resolved.effective_context_budget()"] --> P["floor_exceeds_budget predicate"]
    B --> L["structured log budget attr"]
    B --> T["human remediation text"]
```

### Validation run locally

```text
cargo fmt --all -- --check                                      PASS
cargo check -p zeroclaw-runtime --lib                           PASS
cargo test -p zeroclaw-runtime floor_exceeds_budget              PASS (2 tests)
cargo test -p zeroclaw-runtime estimate_system_floor             PASS (2 tests)
cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings    PASS
```

I also ran `cargo test -p zeroclaw-runtime --lib`. It got past the compile blocker and hit the known parallel log-capture flake `cost_usd_flows_to_usage_event_and_llm_response`; rerunning that exact test in isolation passed:

```text
cargo test -p zeroclaw-runtime agent::turn::parse_response::cost_usd_regression_tests::cost_usd_flows_to_usage_event_and_llm_response -- --exact    PASS
```

PR #7365, opened 2026-06-08
docs(book): rework the book and derive provider/config surfaces from source

singlerider:
> [!IMPORTANT]
> **DO NOT MERGE. The author will merge this themselves.**
> Seeking approval only; this PR is the base of a stack and will be landed by the author once approved.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Reworks the mdBook: numbered child pages for section landings, theme-aware
    syntax highlighting, an OS-tabs widget wrapping every shell invocation, and
    sidebar/nav fixes so in-page headings stop leaking into the left rail and
    fold state persists across navigation.
  - Adds an in-tree mdBook preprocessor (`xtask`) that derives provider, config,
    peer-group, and zerocode-keymap surfaces directly from source, then drops the
    hand-written TOML config blocks and the hand-maintained provider catalog
    table so the book can no longer drift from the registry.
  - Corrects stale or inaccurate prose against source across architecture,
    contributing, channels, philosophy, setup, and SOP pages; adds a NixOS
    install page and a Peer Groups section.
  - Adds a docs quality gate (em-dash guard + markdownlint scoped to book
    sources) and enables it in CI.
  - Fixes zerocode/web Quickstart UX defects and forces the unbounded runtime
    profile in Quickstart.
- **Scope boundary:** No behavior change for existing configurations. Seven new
  provider config slots (`morph`, `github_models`, `upstage`, `featherless`,
  `arcee`, `lambda_ai`, `inception`) are added to the registry as merge
  resolution. This is additive and backwards-compatible: the daemon now parses
  these new `[providers.models.<type>.*]` keys, but existing configs are
  unaffected.
  The only other non-docs source edits are the mdBook preprocessor (`xtask`) and
  the zerocode/web Quickstart UI fix.
- **Blast radius:** Book build (`xtask mdbook`), the docs-deploy workflow, and
  the zerocode/web Quickstart pane. No effect on the agent runtime or the
  published binary surface.
- **Linked issue(s):** Closes #5269. Closes #6222.
- **Labels:** `risk: medium`, `size: XL`, `docs`, `ci`, `dependencies`, `core`, `config`, `gateway`,
  `provider`, `runtime`, `scripts`, `quickstart` (auto-scope labels).

## Validation Evidence (required)

Docs-dominant change: ran the docs quality gate, plus `cargo fmt` and the
provider crate tests for the merge-resolution source edits.

```bash
bash scripts/ci/docs_quality_gate.sh
cargo fmt --all -- --check
[...]

Issue #7080, opened 2026-06-02
bug(delegate): preserve memory namespace across agentic delegate modes


Audacity88:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

Agentic delegate sub-agents can run memory tools from the parent tool registry instead of rebuilding those tools against the delegated agent's configured memory scope.

The stale PR #5530 identified the concrete symptom: when a delegated agent has a configured memory namespace, `memory_store` / `memory_recall` can still write to or read from the parent/default namespace. Current `master` still builds the agentic sub-agent tool list by wrapping matching entries from `parent_tools`, and the background / parallel delegate paths create inner `DelegateTool` values with `memory: None`.

Related stale branch:

- #5530

That branch fixed part of the synchronous agentic path, but the active review found the background and parallel paths still lost the memory handle. This issue should become the durable tracker before closing the stale PR branch.

I also checked current #6848 head (`100f216c`). It changes adjacent delegate policy and routing behavior, but it does not appear to resolve this memory namespace bug: agentic delegate still builds sub-agent tools from the parent tool registry, and the background / parallel inner `DelegateTool` values still set `memory: None`.

### Expected behavior

When a delegated agent has an isolated memory scope, every agentic delegate execution mode should use memory tools bound to that delegated scope:

- synchronous delegate execution
- `background: true` delegate execution
- `parallel` delegate execution

Memory writes from a delegated agent should not land in the caller/default namespace unless that is explicitly the configured scope.

### Steps to reproduce

```bash
1. Configure an agentic delegated agent with its own memory namespace or memory scope.
2. Allow that delegated agent to use `memory_store` and `memory_recall`.
3. Invoke the agent through the delegate tool and have it store a unique value.
4. Repeat through the background or parallel delegate mode.
5. Inspect the persisted memory namespace for the unique value.
```

Expected: the value is present only in the delegated agent's namespace.

Observed from #5530 review: the existing tool wiring can route memory tools through the parent/default memory handle, and the background / parallel delegate paths did not preserve the memory handle in the stale fix branch.

### Impact

This breaks memory isolation for delegated agent work. It can cause delegated agents to leak working memory into the caller/default namespace, miss their own prior memory, or pollute another agent's recall results.

### Logs / stack traces

No raw logs are needed for this tracker. The important evidence is the stale PR review history and the current delegate tool wiring in `crates/zeroclaw-runtime/src/tools/delegate.rs`.

[...]

PR #7442, opened 2026-06-10
fix(runtime): make parallel SubAgents and Delegates return reliably

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Exempt the re-entrant agent-spawning tools (`spawn_subagent`, `delegate`) from the per-turn duplicate-call guard. Two identical-argument calls are intentional fan-out (redundancy, sampling), not an accidental repeat; the guard was dropping the 2nd+ so only the earliest identical call returned. Exemption set (`REENTRANT_AGENT_TOOLS`) is built from `SpawnSubagentTool::NAME` / `DelegateTool::NAME` consts so a rename cannot desync it.
  - **Gate `spawn_subagent` launches on the caller's action budget** (review feedback): `SpawnSubagentTool` now holds the caller's `Arc<SecurityPolicy>` and every validated launch attempt goes through `enforce_tool_operation(Act)`, consuming one slot from the shared `max_actions_per_hour` bucket before any spawn work — the same safety model `DelegateTool` already uses. The dedup exemption can no longer turn one model turn into unbounded child launches; fan-out is bounded at launch time, not merely by work performed downstream.
  - Drop `model_switch` from a SubAgent's tool registry (gated on `is_subagent_caller`). A SubAgent runs as an ephemeral clone inheriting the parent's model verbatim and must not switch the active model out from under the parent; the switch signal is a process-global slot that also races across concurrent runs. Excluding the tool enforces the documented inheritance and removes the cross-talk.
  - Pair native `role=tool` history messages on each result's own `tool_call_id` instead of a positional `zip` against `native_tool_calls`. `zip` stops at the shorter side, so any length divergence would orphan a native `tool_use` id with no matching `tool_result`. Defensive: the positional `zip` was correct for well-formed native turns, so this is robustness, not a behavior fix.
- **Scope boundary:** No change to SubAgent/Delegate identity inheritance semantics (provider, model, risk profile, memory, budgets all still inherited verbatim). No config-surface or schema change. No new tools. No general per-turn tool-call count cap is added — the launch bound is budget-based, matching `delegate`. Docs for `delegation.md` are tracked separately under #7365.
- **Blast radius:** `zeroclaw-runtime` tool-registry assembly and the tool-call loop's result-to-history mapping. Affects every agent turn that emits tool calls (native and prompt mode) and every SubAgent/Delegate spawn.
  - **Fan-out bound:** identical re-entrant calls are no longer collapsed per-turn, and both spawn paths now consume the caller's shared action budget at launch time: `delegate` via its existing `enforce_tool_operation(Act)` gate, `spawn_subagent` via the new launch-side gate. SubAgents and Delegates additionally share the parent/caller `Arc<SecurityPolicy>` tracker (`SubAgentSpawn::build` inherit path; `DelegateTool::policy_for_target` copies `self.security.tracker`), so child work also draws from the same `max_actions_per_hour` / `max_cost_per_day_cents` bucket.
  - `SpawnSubagentTool::new` gains a third parameter (caller policy); single non-test construction site updated in `tools/mod.rs`.
  - No effect on channels, gateway, or providers beyond what flows through the loop.
- **Linked issue(s):** Related #7365 (docs addition for `delegation.md` lands there). Related #7080 (delegate memory-namespace loss in the same background/parallel paths; not resolved here).
- **Labels:** `risk: high`, `size: M`, `runtime`, `agent`

```bash
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace
```

- **Commands run and tail output** (on the PR head `a9954d0af2`, including the launch-gate commit):

`cargo fmt --all -- --check` — clean (exit 0, no output).

`cargo clippy --workspace --all-targets -- -D warnings`:
```
exit 0
```

`cargo test --workspace` — 62 `test result:` blocks, every block `0 failed`. Aggregate: **8183 passed, 0 failed** (8181 prior + 2 new launch-gate regressions). Representative literal tail:
```
test tools::spawn_subagent::tests::repeated_spawns_blocked_once_action_budget_is_exhausted ... ok
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 2015 filtered out; finished in 0.01s   (spawn_subagent filter)
```

- **Beyond CI — what did you manually verify?** Five targeted regression tests: native parallel return yields one `role=tool` message per call keyed to its own id; identical-prompt `spawn_subagent` calls both execute with no config dedup exemption; `model_switch` present in a top-level registry and absent in a SubAgent registry; **with `max_actions_per_hour: 2`, the 3rd identical `spawn_subagent` launch is refused with "Rate limit exceeded"**; **repeated validation failures (blank prompt) consume zero launch budget**. Confirmed `model_switch` has a single registration site and that the cron spawn site sets `is_subagent: false`. Verified gate ordering matches `DelegateTool` (validate args first, then `enforce_tool_operation`) so malformed calls don't burn slots, consistent with `RateLimitedTool`'s only-work-consumes-budget semantics.
- **Beyond CI — what I did NOT verify:** No live-provider run; the new tests use scripted/mock providers. No end-to-end parallel SubAgent run against a real model.

- New permissions, capabilities, or file system access scope? **No** — this narrows capabilities twice over (SubAgents lose `model_switch`; `spawn_subagent` launches now consume the action budget).

## Compatibility (required)

- Backward compatible? **Yes** — no config/schema/CLI change. A SubAgent losing `model_switch` matches the documented inheritance contract and was never a supported way to change a child's model. `spawn_subagent` calls now count against the existing `max_actions_per_hour` budget like every other acting tool; operators with very low budgets and heavy fan-out usage will see launches refused once the bucket drains, which is the intended bound.
- If `No` or `Yes` to either: exact upgrade steps for existing users: None required.

- **Fast rollback command/path:** `git revert <merge-sha>` — self-contained commits, no migration.
- **Feature flags or config toggles:** None. (`parallel_tools` already gates concurrent execution and is unchanged; the launch bound reuses the existing `max_actions_per_hour` knob.)
- **Observable failure symptoms:** If the dedup exemption regressed, a turn with two identical `spawn_subagent`/`delegate` calls would log `"Skipped duplicate tool call"` for the 2nd. If the launch gate regressed open, a single turn could start more SubAgent runs than `max_actions_per_hour` permits (grep runtime trace for `spawn_subagent` invoke counts per turn); if it regressed closed, every spawn would fail with `"Rate limit exceeded: action budget exhausted"`. If native pairing regressed, native providers would reject the next request for a `tool_use` id lacking a matching `tool_result` (provider 400). If the registry filter regressed, a SubAgent would expose `model_switch` (grep the child run's tool list).

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/loop_.rs
+++ b/crates/zeroclaw-runtime/src/agent/loop_.rs
@@ -2493,7 +2493,8 @@ pub async fn run_tool_call_loop(
                     serde_json::to_string(&canonical_args).unwrap_or_else(|_| "{}".to_string());
                 (tool_name.trim().to_ascii_lowercase(), args_json)
             };
-            let dedup_exempt = dedup_exempt_tools.iter().any(|e| e == &tool_name);
+            let dedup_exempt = dedup_exempt_tools.iter().any(|e| e == &tool_name)
+                || crate::tools::REENTRANT_AGENT_TOOLS.contains(&tool_name.as_str());
             if !dedup_exempt && !seen_tool_signatures.insert(signature) {
                 let duplicate = format!(
                     "Skipped duplicate tool call '{tool_name}' with identical arguments in this turn."
@@ -2844,11 +2845,15 @@ pub async fn run_tool_call_loop(
                 history.push(ChatMessage::user(format!("[Tool results]\n{tool_results}")));
             }
         } else {
-            for (native_call, (_, result)) in
-                native_tool_calls.iter().zip(individual_results.iter())
-            {
+            // `zip` would drop trailing results on any length divergence,
+            // leaving a native tool_use id with no matching tool_result.
+            // Pair on each result's own id instead.
+            for (idx, (tool_call_id, result)) in individual_results.iter().enumerate() {
+                let resolved_id = tool_call_id
+                    .clone()
+                    .or_else(|| native_tool_calls.get(idx).map(|call| call.id.clone()));
                 let tool_msg = serde_json::json!({
-                    "tool_call_id": native_call.id,
+                    "tool_call_id": resolved_id,
                     "content": result,
                 });
                 history.push(ChatMessage::tool(tool_msg.to_string()));
@@ -5815,6 +5820,39 @@ mod tests {
             self.capabilities.native_tool_calling = true;
             self
         }
+
+        /// Build a native-tool-calling provider: one turn of structured
+        /// `tool_calls`, then a plain-text turn.
+        fn from_native_tool_calls(calls: Vec<(&str, &str, &str)>, final_text: &str) -> Self {
+            let tool_turn = ChatResponse {
+                text: None,
+                tool_calls: calls
+                    .into_iter()
+                    .map(|(id, name, args)| ToolCall {
+                        id: id.to_string(),
+                        name: name.to_string(),
+                        arguments: args.to_string(),
+                        extra_content: None,
+                    })
+                    .collect(),
+                usage: None,
+                reasoning_content: None,
+            };
+            let final_turn = ChatResponse {
+                text: Some(final_text.to_string()),
+                tool_calls: Vec::new(),
+                usage: None,
+                reasoning_content: None,
+            };
+            let capabilities = ProviderCapabilities {
+                native_tool_calling: true,
+                ..ProviderCapabilities::default()
+            };
+            Self {
+                responses: Arc::new(Mutex::new(VecDeque::from(vec![tool_turn, final_turn]))),
+                capabilities,
+            }
+        }
     }
 
     #[async_trait]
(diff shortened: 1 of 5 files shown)


Audacity88 (dismissed):
Reviewed the PR body, diff, current checks, labels, and existing review/comment state. I focused on the runtime fan-out change, native tool-result history pairing, and the SubAgent tool boundary for `model_switch`; I did not run additional local cargo checks because the full remote CI matrix is green.

### 🟢 What looks good — Native fan-out history

Pairing native tool results by `tool_call_id` is the right hardening here: it avoids relying on positional `zip` pairing in the native parallel-tool path and keeps each result attached to the call that produced it.

### 🟢 What looks good — SubAgent tool boundary

The SubAgent registry test confirms that `model_switch` remains available to top-level agents but is absent when tools are built for a SubAgent, which matches the inherited-model boundary described in the PR.

### 🔴 Blocking — SubAgent fan-out still needs a launch-side bound

The native/result-pairing parts look good, but I do not think the fan-out safety backstop is complete yet. The PR body calls out the key risk directly: this exemption removes the only per-turn ceiling on identical re-entrant calls, and the remaining budget bounds work performed rather than spawns attempted. That is fine for allowing intentional redundancy, but with `spawn_subagent` it still means one parent turn can request many identical child launches before any per-turn launch cap exists.

Please add a launch-side bound for `spawn_subagent` fan-out so the new dedup exemption cannot turn one model turn into an unbounded number of child agent starts. The smallest safe shape seems to be either making each `spawn_subagent` launch consume the caller action budget, matching the safety model used for other acting tools, or adding an explicit per-turn cap for re-entrant agent-tool fan-out. A focused regression with a tiny budget or cap and repeated `spawn_subagent` calls would make this much safer to approve.

Audacity88 (approved):
Reviewed the updated head `a9954d0a` after my previous blocking review on `adb0e4c`. I checked the PR body, public diff, current check snapshot, labels, prior review state, the launch-gate follow-up, native tool-result pairing, and child-agent `model_switch` boundary. I did not rerun the cargo suite locally; the PR validation evidence and current CI are green.

### ✅ Resolved — Launch-side fan-out bound

The new `spawn_subagent` launch gate now consumes the caller's shared action budget after prompt validation and before spawn work. The added regressions cover both budget exhaustion and validation failures not burning launch slots, so the fan-out bound concern is resolved.

### 🟢 What looks good — Native parallel-tool history

Native parallel calls now emit one `role=tool` message keyed to each result's own id, avoiding positional `zip` loss if result/call lengths diverge.

### 🟢 What looks good — Child-agent model boundary

The registry test pins `model_switch` as top-level only, so child runs keep inherited model behavior instead of exposing a process-global switch.

PR #7445, opened 2026-06-10
fix(docker): resolve workspace build failures from apps/zerocode and embedded assets

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** The Docker build broke at the cargo dependency pre-fetch stage with `failed to load manifest for workspace member /app/apps/zerocode`. #6848 added `apps/zerocode` to the workspace `members` but never updated the Docker build context. Fixing the build surfaced a second goal: the `zerocode` TUI binary must actually **ship** in the image (it was stubbed and never copied, so `zerocode` was `command not found` inside the container). This PR does both — repairs the build chain and ships a working `zerocode`:
  - `apps/zerocode/Cargo.toml` was never copied → cargo could not resolve the workspace. Copy the manifest in the dep-cache stage.
  - `crates/zeroclaw-hardware` declares an `[[example]]` (`esp32_sim`) whose source is absent in the stub stage → manifest parse failed once zerocode resolved. Stub `examples/esp32_sim.rs` alongside the per-crate lib stubs.
  - `locales.toml` (repo root) is embedded by `zeroclaw-runtime` at compile time via an `include_str!` of the repo-root file; the final source-copy stage copied `*.rs` but not the `.toml` → the real build failed to compile. Copy it.
  - Ship zerocode for real: copy real `apps/zerocode/` source plus `web/src/` (its `build.rs` reads the dashboard theme registry under `web/src/contexts` to generate the TUI theme presets), build `-p zeroclawlabs -p zerocode` (a bare `cargo build` at the workspace root builds only the `zeroclawlabs` package, not sibling members — so `target/release/zerocode` never existed and the cp failed), install `g++` (zerocode's `inkjet` highlighter compiles tree-sitter grammars in C++; the slim base ships `cc` but no C++ compiler), then cp + strip + size-guard the binary and install `/usr/local/bin/zerocode` in every runtime stage.
  - `.dockerignore`: whitelist the full `apps/zerocode/` source (was manifest-only) since the binary now ships.
  - `Dockerfile.debian`: apply the same fixes; it had additionally drifted and was missing the `apps/tauri/build.rs` stub the default Dockerfile carries.
- **Scope boundary:** Both build-from-source Dockerfiles only. `ZEROCLAW_CARGO_FEATURES` differing between the variants (`channel-lark,whatsapp-web` vs `rag-pdf`) is intentional and left as-is. `Dockerfile.ci` / `Dockerfile.debian.ci` use pre-built binaries (no cargo build) and are unaffected. Does not consolidate the duplicated Dockerfiles or touch the StageX `Containerfile` (tracked as a follow-up).
- **Blast radius:** Both build-from-source Docker images (`Dockerfile` distroless, `Dockerfile.debian` shell variant) and the release workflows that build them. Image contents now include the `zerocode` binary. No effect on the application, config schema, or the `zeroclaw` daemon binary.
- **Linked issue(s):** None filed; reported directly. Root cause traced to #6848 (introduced `apps/zerocode` to the workspace).
- **Labels:** `risk: medium`, `size: XS`, `ci`

Docker build is the signal here, not the Rust battery.

```bash
DOCKER_BUILDKIT=1 docker build --target builder -t zeroclaw-builder-test .
docker run --rm --entrypoint /usr/local/bin/zerocode zeroclaw-builder-test --help
```

Default `Dockerfile` builder target — built end to end, both binaries pass the size guard:
```
#30 299.8     Finished `release` profile [optimized] target(s) in 4m 59s
#31 [builder 25/26] RUN for b in zeroclaw zerocode; do size=$(stat -c%s "/app/$b") ... done   (passed)
#33 writing image sha256:0eb3a0d7... done
```

Both binaries present in the image, real (not dummy stubs):
```
-rwxr-xr-x  /app/zeroclaw  36600832 bytes
-rwxr-xr-x  /app/zerocode  36001712 bytes
```

`zerocode --help` runs from inside the image (exit 0):
```
Interactive TUI config manager for ZeroClaw

Usage: zerocode [OPTIONS]

Options:
      --config-dir <CONFIG_DIR>  Path to the ZeroClaw config directory
  -a, --agent <AGENT>            Start in chat mode with this agent alias...
```

- **Beyond CI — what did you manually verify?** Reproduced each failure in the chain (`failed to load manifest`, `can't find esp32_sim example`, `couldn't read locales.toml`, `cp: cannot stat 'target/release/zerocode'`, `failed to find tool "c++"`) and confirmed each fix advances the build to the next stage until the image completes. **Owner-confirmed live:** `docker run -it --entrypoint zerocode <image>` launches the zerocode TUI in BOTH the debian and distroless flavors.
- **Beyond CI — what I did NOT verify:** A full `Dockerfile.debian` end-to-end run on the build host (one attempt tripped on a transient `npm ci` SIGTERM in the unrelated web-builder stage; the builder stage itself cleared all manifest/compile failures and compiled zerocode). The shared builder-stage logic is identical between the two Dockerfiles.
- **If any command was intentionally skipped, why:** The Rust `cargo` battery is not the relevant signal for a Docker-context change; the `docker build` + in-image run above is.

- If any `Yes`, describe the risk and mitigation: N/A. (Adds the `zerocode` TUI binary to the image; it is built from the same workspace source under the same features.)

## Compatibility (required)

- Backward compatible? **Yes** — the `zeroclaw` daemon binary and entrypoint are unchanged; the image now additionally carries `zerocode`. Reach it with `--entrypoint zerocode` (an entrypoint-dispatch UX improvement so `docker run ... zerocode` works without the override is tracked as a follow-up).

- **Fast rollback command/path:** `git revert <merge-sha>` — self-contained Docker-context commits, no migration.
- **Observable failure symptoms:** A regression returns one of the original errors (`failed to load manifest for workspace member /app/apps/zerocode`, `can't find esp32_sim example`, `couldn't read locales.toml`, `cp: cannot stat 'target/release/zerocode'`, or `failed to find tool "c++"`) at the cargo pre-fetch / real-build stage, or `zerocode: command not found` inside the container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -69,9 +69,13 @@ coverage
 lcov.info
 
 # Application and script directories (not needed for Docker runtime)
-# Note: apps/tauri/Cargo.toml is required for workspace resolution in Docker build.
+# Note: apps/tauri and apps/zerocode are desktop/TUI apps not shipped in the
+# server image; only their Cargo.toml is needed for workspace resolution in the
+# Docker dependency-cache stage (their src + build.rs are stubbed in the
+# Dockerfile). Whitelisting just the manifest avoids cache-busting on app source.
 apps/*
 !apps/tauri/
 !apps/tauri/Cargo.toml
 !apps/zerocode/
+!apps/zerocode/Cargo.toml
 scripts/
--- a/.dockerignore
+++ b/.dockerignore
@@ -68,14 +68,13 @@ LICENSE
 coverage
 lcov.info
 
-# Application and script directories (not needed for Docker runtime)
-# Note: apps/tauri and apps/zerocode are desktop/TUI apps not shipped in the
-# server image; only their Cargo.toml is needed for workspace resolution in the
-# Docker dependency-cache stage (their src + build.rs are stubbed in the
-# Dockerfile). Whitelisting just the manifest avoids cache-busting on app source.
+# Application and script directories.
+# apps/tauri is a desktop app not shipped in the server image; only its
+# Cargo.toml is needed for workspace resolution (src + build.rs are stubbed in
+# the Dockerfile). apps/zerocode IS shipped (the TUI binary), so its full source
+# is whitelisted and copied for the real build.
 apps/*
 !apps/tauri/
 !apps/tauri/Cargo.toml
 !apps/zerocode/
-!apps/zerocode/Cargo.toml
 scripts/
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,11 +96,15 @@ COPY benches/ benches/
 COPY crates/ crates/
 COPY xtask/ xtask/
 COPY tools/fill-translations/ tools/fill-translations/
+# apps/zerocode ships in the image; copy its real source. Its build.rs reads the
+# dashboard theme registry under web/src/contexts, so that path must be present.
+COPY apps/zerocode/ apps/zerocode/
+COPY web/src/ web/src/
 # locales.toml lives at repo root and is embedded by zeroclaw-runtime via
 # include_str!("../../../locales.toml"); the real build needs it present.
 COPY locales.toml .
 COPY *.rs .
-RUN touch src/main.rs
+RUN touch src/main.rs apps/zerocode/src/main.rs
 RUN --mount=type=cache,id=zeroclaw-cargo-registry,target=/usr/local/cargo/registry,sharing=locked \
     --mount=type=cache,id=zeroclaw-cargo-git,target=/usr/local/cargo/git,sharing=locked \
     --mount=type=cache,id=zeroclaw-target,target=/app/target,sharing=locked \
@@ -113,16 +117,22 @@ RUN --mount=type=cache,id=zeroclaw-cargo-registry,target=/usr/local/cargo/regist
            target/release/.fingerprint/xtask-* \
            target/release/deps/xtask-* \
            target/release/.fingerprint/fill-translations-* \
-           target/release/deps/fill_translations-* && \
+           target/release/deps/fill_translations-* \
+           target/release/.fingerprint/zerocode-* \
+           target/release/deps/zerocode-* \
+           target/release/incremental/zerocode-* && \
     if [ -n "$ZEROCLAW_CARGO_FEATURES" ]; then \
       cargo build --release --locked --features "$ZEROCLAW_CARGO_FEATURES"; \
     else \
       cargo build --release --locked; \
     fi && \
     cp target/release/zeroclaw /app/zeroclaw && \
-    strip /app/zeroclaw
-RUN size=$(stat -c%s /app/zeroclaw) && \
-    if [ "$size" -lt 1000000 ]; then echo "ERROR: binary too small (${size} bytes), likely dummy build artifact" && exit 1; fi
+    cp target/release/zerocode /app/zerocode && \
+    strip /app/zeroclaw /app/zerocode
+RUN for b in zeroclaw zerocode; do \
+      size=$(stat -c%s "/app/$b") && \
+      if [ "$size" -lt 1000000 ]; then echo "ERROR: $b too small (${size} bytes), likely dummy build artifact" && exit 1; fi; \
+    done
 
 # Prepare runtime directory structure and default config inline (no extra stage).
 # Dashboard assets live at /usr/share/zeroclawlabs/web/dist (outside the documented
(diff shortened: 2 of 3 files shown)


WareWolf-MoonWall (approved):
# Review — PR #7445
## `fix(docker): resolve workspace build failures from apps/zerocode and embedded assets`
**Author:** singlerider | **Verdict:** `--approve`

---

This is a well-reasoned, well-documented multi-fix for a failure chain that blocked every fresh Docker build after #6848. The PR description traces each individual failure step-by-step and explains the fix for each, which is exactly what I want to see for a change touching multi-stage Dockerfiles. I'm approving with one warning worth tracking.

### 🟡 Warning — dep-cache stage comment contradicts the image's actual contents

In both `Dockerfile` and `Dockerfile.debian`, the new comment block reads:

```dockerfile
# apps/zerocode: TUI app not shipped in the server image; copy only its manifest
```

But the whole point of this PR is that `zerocode` IS shipped in the final image. A future contributor reading only this part of the dep-cache stage could incorrectly conclude that `zerocode` should never reach the runtime layers. The intended meaning — "in *this* pre-fetch stage, only the manifest is handled; the real source is copied in the builder stage" — would be communicated more clearly with:

```dockerfile
# apps/zerocode: TUI binary ships in the image; in this pre-fetch stage we only
# need the manifest for workspace resolution. The real src/ and build.rs are
# stubbed here and replaced with the full source in the builder stage below.
```

Recommend fixing before or alongside merge (a single-commit follow-up is fine).

### 🟢 What looks good — root-cause chain is fully closed

Five distinct failure modes (`failed to load manifest for workspace member`, `can't find esp32_sim example`, `couldn't read locales.toml`, `cp: cannot stat 'target/release/zerocode'`, `failed to find tool "c++"`), each diagnosed and independently fixed. None of the fixes are cargo-culted — each has an explanatory comment in the Dockerfile that makes the "why" obvious to the next person who touches the build.

### 🟢 What looks good — size guard extended to cover both binaries

Changing the single `stat`-and-guard RUN step into a loop over `zeroclaw zerocode` is the right move. Both binaries now fail fast if the build produced a dummy artifact, and the loop pattern scales cleanly if another binary is added later.

### 🟢 What looks good — `Dockerfile.debian` drift caught

The `apps/tauri/build.rs` stub was present in the default `Dockerfile` but absent from `Dockerfile.debian` — a pre-existing drift. Catching and fixing that here, even though it's tangential to the main fix, is the right call rather than leaving a latent bug for the debian variant.

### 🔵 Suggestion — `web/src/` copy could be scoped to just the theme file

`COPY web/src/ web/src/` brings the entire web dashboard TypeScript source into the builder layer. This is required because `apps/zerocode/build.rs` reads `web/src/contexts/themes.json` (or similar) to generate TUI theme presets. If the build script only reads a specific file, narrowing the COPY to that path would reduce layer invalidation when unrelated web files change:

```dockerfile
COPY web/src/contexts/themes.json web/src/contexts/themes.json
```

Worth checking what `apps/zerocode/build.rs` actually opens; if it's a single file, the targeted copy is strictly better. Not a blocker — the current form is correct and safe.

### Note on Dockerfile.debian end-to-end run

The author noted that a complete Dockerfile.debian build was not verified end-to-end (transient `npm ci` SIGTERM in the web-builder stage, unrelated to this PR's changes). The builder stage cleared all manifest/compile failures and produced `zerocode`; the web-builder stage failure is pre-existing and not introduced here. I'm satisfied the scope is correct.

---

**Approving. The stale dep-cache comment is the only actionable item; a follow-up commit is fine.**

Audacity88 (approved):
Reviewed the live PR state at head `874fa8b158bb7c31fbd5343798d4dc73c684dd30`, the prior review from @WareWolf-MoonWall, the three-file Docker diff, the release-image scope, source evidence for the `zerocode` build inputs, and the PR body's validation evidence. CI is green, the branch is clean/mergeable, and I am approving this Docker image fix. This approval is for the reviewed head; normal final merge-readiness checks should still refresh if `master` moves again.

### 🟢 What looks good — Docker build chain is covered end to end

The diff fixes each failure class the PR body names: `apps/zerocode/Cargo.toml` in the dependency-cache stage, stubs for `zerocode` build metadata and the hardware example, repo-root `locales.toml`, real `apps/zerocode` plus the web theme source for the build script, explicit `-p zeroclawlabs -p zerocode` builds, and `g++` for the syntax-highlighting dependency chain. The runtime stages copy `zerocode`, and the size guard now covers both shipped binaries.

### 🟢 What looks good — Validation matches the release-image claim

The relevant proof is a Docker build plus in-image execution, and the PR body provides that for the default builder image and `zerocode --help`, with owner-confirmed TUI launch in both Debian and distroless flavors. The skipped full Debian end-to-end run is clear about the unrelated `npm ci` SIGTERM; given the shared builder-stage changes and green CI, that gap is not a blocker for review.

### 🔵 Suggestion — Clarify the scoped zerocode prefetch comments

I agree with @WareWolf-MoonWall's note on the dep-cache comments: both Dockerfiles still say "TUI app not shipped in the server image; copy only its manifest" in the pre-fetch stage, while this PR intentionally ships `zerocode` later in the image. The `.dockerignore` comment already has the right model, so this is only a documentation clarity cleanup; the actual copy/build/runtime steps are correct.

### 🔵 Suggestion — Consider narrowing the web source copy later

`apps/zerocode/build.rs` currently reads `web/src/contexts/themes.json`, so `COPY web/src/ web/src/` is safe but broader than the build script appears to require. A later cleanup could copy only the theme registry to reduce Docker layer invalidation from unrelated web changes.

Issue #7436, opened 2026-06-09
[Bug]: image_info tool output does not reach multimodal/vision models in common cases


NiuBlibing:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

`image_info` output only reaches a vision model when the agent happens to call it with an **absolute POSIX path**. Other common call shapes silently drop the image:

1. **Relative / workspace-relative paths.** The tool's `File:` line echoes the raw caller-supplied `path_str` (`crates/zeroclaw-tools/src/image_info.rs:244`), not the canonicalized absolute path. The step that promotes a path in tool output into an `[IMAGE:]` marker — `canonicalize_tool_result_media_markers` — uses a `/`-anchored regex (`crates/zeroclaw-runtime/src/agent/history.rs:15`) plus an `is_absolute()` check (`crates/zeroclaw-runtime/src/agent/history.rs:133`). A relative path (the schema explicitly allows "relative to workspace") does not match → no marker → no inlining → the model sees text only.
2. **Windows absolute paths** (`C:\...png`) are missed for the same reason: the regex only matches paths starting with `/`.
3. **`include_base64: true`** emits a bare `data:{mime};base64,...` line (`crates/zeroclaw-tools/src/image_info.rs:261`), but the only marker parser, `parse_image_markers`, recognizes the `[IMAGE:` prefix exclusively. A bare data URI has no consumer → it is text noise that also bloats the context window. This is vestigial output that predates the marker-based vision pipeline and was never wired to it (the tool's own doc comment describes the base64 as "for future multimodal model_provider support").

### Expected behavior

Regardless of whether `image_info` is called with an absolute, relative, or Windows path, as long as the target is a supported image format and the model supports vision, the image should be inlined to the vision model. `include_base64` should either actually work (produce inlinable, marker-wrapped data) or be removed so it does not mislead.

### Steps to reproduce

```bash
# 1. Configure a vision-capable model (supports_vision() == true).
# 2. Place an image in the workspace: {workspace}/images/sample.png
# 3. Have the agent call image_info with a workspace-relative path:
#    image_info { "path": "images/sample.png" }
# 4. Ask the model what is in the image.
# => Model replies that it cannot see an image. The tool output contains
#    `File: images/sample.png`, which is NOT rewritten to [IMAGE:...] and
#    never enters the multimodal inlining pipeline.
#
# Contrast: calling with an absolute path /abs/.../images/sample.png inlines correctly.
```

[...]

PR #7446, opened 2026-06-10
fix(multimodal): make image_info images reach vision models end-to-end

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **(original)** `image_info` only delivered its image to a vision model when called with an **absolute POSIX path** — its `File:` line echoed the caller-supplied `path_str`, and the tool-result marker promoter (`canonicalize_tool_result_media_markers`) is anchored on a leading `/` + `is_absolute()`, so a workspace-relative path was silently dropped. Now it emits an explicit `[IMAGE:<canonical absolute path>]` marker from the resolved path, removes the dead `include_base64` parameter, and strips the Windows verbatim prefix (`\\?\C:\…`, `\\?\UNC\…`) so the marker carries a plain path.
  - **(follow-up) UNC actually reaches the model.** The verbatim-UNC unwrap above produced `\\server\share\…`, but the downstream loadability gate (`is_loadable_image_reference` / `is_windows_path`) only accepted `/`-rooted and `C:\` drive paths — so the marker was left as literal text and never inlined. Added `is_windows_unc_path` to the gate and extended the promoter regex to recognize Windows drive **and** UNC bare paths. Without this, the verbatim-UNC handling was inert.
  - **(follow-up) Non-vision handling, split by marker origin.** Emitting an image marker on a model without vision support (and no `vision_model_provider` configured) previously returned a hard `ProviderCapabilityError` that killed the whole turn — even when the image came from a tool, not the user. Now:
    - A **user-supplied** image still surfaces the capability error, so channels render it back to the user (\"⚠️ Error … does not support vision\") and the attachment is not silently ignored.
    - A **tool-result** image marker (`image_info` / `screenshot` / `image_gen`) degrades to text-only — markers are stripped, surrounding text/metadata is kept, and the provider is still called so the turn completes.
    - A configured-but-non-vision `vision_model_provider` is still a hard error (operator misconfiguration).
  - **(follow-up) Size cap aligned with the pipeline.** `image_info`'s hard 5 MB read ceiling was decoupled from `multimodal.max_image_size_mb` (clamped 1..=20 MB); it could refuse a file the pipeline was configured to accept. The ceiling is now the clamp's upper bound (20 MiB); the per-request decision is the pipeline's config.
  - **(follow-up) Durable path survives marker stripping.** Putting the path *only* inside the marker meant that once the marker was stripped from older turns, the path vanished entirely — the model could not even re-read the file. `image_info` now emits a `File: <abs path>` line **and** the marker; the promoter dedups a bare path already present in a marker so the extra line is not double-counted.
  - **(follow-up) BMP no longer breaks the request.** `image/bmp` was in the inline-allowed MIME list, but no major vision provider (Anthropic, OpenAI) accepts BMP — inlining it made the *entire* provider request fail. BMP is dropped from the allowed list, so it is skipped with a \"could not be loaded\" note while the rest of the request goes through.
- **Scope boundary:** Touches `crates/zeroclaw-tools/src/image_info.rs`, `crates/zeroclaw-providers/src/multimodal.rs`, and `crates/zeroclaw-runtime/src/agent/{history.rs,loop_.rs}`. Does **not** change `image_gen`/`screenshot`, channel code, or the provider adapters' image-block conversion. Does not rework the stale-tool-result stripping policy (kept for context control).
- **Blast radius:**
  - `image_info` output shape (now a `File:` line + standalone `[IMAGE:]` marker; no base64 blob) and schema (`include_base64` removed).
  - Multimodal loadability gate + promoter regex now accept UNC/drive paths — affects any tool output printing image paths, not just `image_info`.
  - Non-vision degradation applies to **tool-result** markers only (so `screenshot`/`image_gen` on a non-vision model now degrade rather than abort); **user** image attachments still surface the capability error unchanged.
  - `image/bmp` is no longer inlined for any source.
- **Linked issue(s):** Addresses #7436.

- **Commands run and tail output** (scoped to the crates this PR touches, plus `zeroclaw-channels` which has the photo-attachment e2e contract):

\`\`\`
$ cargo fmt -p zeroclaw-providers -p zeroclaw-tools -p zeroclaw-runtime -p zeroclaw-channels -- --check
(clean — no output)

$ cargo clippy -p zeroclaw-providers -p zeroclaw-tools -p zeroclaw-runtime -p zeroclaw-channels --lib --all-targets -- -D warnings
    Finished \`dev\` profile [unoptimized + debuginfo] target(s) in 6m 43s
(no warnings)

$ cargo test --no-fail-fast -p zeroclaw-providers -p zeroclaw-tools -p zeroclaw-runtime -p zeroclaw-channels --lib
  zeroclaw_channels  test result: ok. 726 passed; 0 failed; 1 ignored
  zeroclaw_providers test result: ok. 922 passed; 0 failed; 0 ignored
  zeroclaw_runtime   test result: ok. 2033 passed; 0 failed; 1 ignored
  zeroclaw_tools     test result: ok. 1312 passed; 0 failed; 0 ignored
\`\`\`

- **Beyond CI — what did you manually verify?** Added regression tests for: UNC recognition (`is_windows_unc_path`) and extraction (`parse_image_markers` treats `\\server\share\…` as a loadable ref); BMP rejection by `validate_mime`; promoter dedup (a bare path already inside an `[IMAGE:]` marker is not promoted to a second marker, order-independent); the **user-image** capability error (preserved); and the **tool-result** text-only degradation path. The two `zeroclaw-channels` orchestrator e2e tests (`e2e_photo_attachment_rejected_by_non_vision_provider`, `e2e_failed_vision_turn_does_not_poison_follow_up_text_turn`) confirm a user photo on a non-vision provider still replies with the capability error. Not exercised end-to-end against a live vision provider on Windows.
- **If any command was intentionally skipped, why:** Validation is scoped to the four crates this PR touches/affects. Note: `zeroclaw-tools`'s `browser::tests::is_service_environment_*` tests mutate shared process env vars (`JOURNAL_STREAM` / `INVOCATION_ID`) and can race under parallel execution; they pass in isolation (`--test-threads=1`) and are unrelated to this diff (which does not touch `browser.rs`). The whole-repo `cargo test` **doc-test** phase fails to start with `error: Option 'default-theme' given more than once` — a pre-existing environment issue (`.cargo/config.toml` `rustdocflags = [\"--default-theme=ayu\"]` received duplicated), unrelated to this diff, which adds no doctests.

- New permissions, capabilities, or file system access scope? `No` — images are still read through the same `PathGuardedTool` allowlist + readable-root checks; the multimodal loader reads marker paths exactly as before.

## Compatibility (required)

- Backward compatible? `Yes`, with two intentional behavior changes that make the pipeline more robust:
  - A **tool-result** image marker on a non-vision model with no `vision_model_provider` no longer aborts the turn — it degrades to text-only. (User-supplied images are unchanged: they still surface the capability error.)
  - `image/bmp` images are no longer inlined; they are skipped with a note rather than failing the whole provider request.
  - (From the original change) an agent still passing `include_base64` has it ignored.
- Config / env / CLI surface changed? `Yes` — the `image_info` parameter schema drops `include_base64` and the output no longer carries a `data:…;base64,…` blob. No new config/env/CLI keys; existing `multimodal.*` keys are unchanged (the size cap now honors `max_image_size_mb` more faithfully).
- Upgrade steps for existing users: None required. Prompts/tooling referencing `image_info`'s `include_base64` should drop it (now a no-op).

### DIFF
--- a/crates/zeroclaw-providers/src/multimodal.rs
+++ b/crates/zeroclaw-providers/src/multimodal.rs
@@ -261,6 +261,23 @@ pub fn contains_image_markers(messages: &[ChatMessage]) -> bool {
     count_image_markers(messages) > 0
 }
 
+/// Count image markers that originate from genuine **user** messages (i.e.
+/// inbound attachments), excluding tool-result carriers (`role == "tool"` and
+/// `[Tool results]` user messages).
+///
+/// Callers use this to distinguish "the user sent an image we cannot see"
+/// (which should surface a user-facing capability error so the attachment is
+/// not silently ignored) from "an image marker arrived only via a tool result"
+/// (e.g. `image_info`/`screenshot`/`image_gen`), which can degrade to text-only
+/// on a non-vision provider without misleading the user.
+pub fn count_user_image_markers(messages: &[ChatMessage]) -> usize {
+    messages
+        .iter()
+        .filter(|message| message.role == "user" && !is_prompt_tool_result_message(message))
+        .map(|message| parse_image_markers(&message.content).1.len())
+        .sum()
+}
+
 /// Replace media markers (`[IMAGE:...]`, `[PHOTO:...]`, `[DOCUMENT:...]`,
 /// `[FILE:...]`, `[VIDEO:...]`, `[VOICE:...]`, `[AUDIO:...]`) with
 /// `[media attachment]`. Match is case-insensitive to align with the channel
--- a/crates/zeroclaw-providers/src/multimodal.rs
+++ b/crates/zeroclaw-providers/src/multimodal.rs
@@ -6,13 +6,14 @@ use zeroclaw_api::model_provider::ChatMessage;
 use zeroclaw_config::schema::{MultimodalConfig, build_runtime_proxy_client_with_timeouts};
 
 const IMAGE_MARKER_PREFIX: &str = "[IMAGE:";
-const ALLOWED_IMAGE_MIME_TYPES: &[&str] = &[
-    "image/png",
-    "image/jpeg",
-    "image/webp",
-    "image/gif",
-    "image/bmp",
-];
+// MIME types we will inline for vision providers. Deliberately excludes
+// `image/bmp`: no major vision provider (Anthropic, OpenAI) accepts BMP, so
+// inlining it would make the *entire* provider request fail rather than just
+// dropping the one image. Rejecting it here instead surfaces a clean
+// "could not be loaded" note while the request (and any accompanying text or
+// metadata) still goes through. BMP is still detected by `detect_mime` so the
+// skip is logged as an explicit unsupported-MIME event rather than "unknown".
+const ALLOWED_IMAGE_MIME_TYPES: &[&str] = &["image/png", "image/jpeg", "image/webp", "image/gif"];
 
 /// Per-path cache for resolved local image data URIs. Keyed by absolute
 /// path; stores `(len, mtime)` for freshness checks (`(0, 0)` sentinel
@@ -127,6 +128,7 @@ fn is_loadable_image_reference(candidate: &str) -> bool {
         || candidate.starts_with("https://")
         || candidate.starts_with("data:")
         || is_windows_path(candidate)
+        || is_windows_unc_path(candidate)
 }
 
 /// Returns true for Windows-style absolute paths like `C:\…` or `D:/…`.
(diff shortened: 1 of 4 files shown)


Audacity88 (approved):
I reviewed current head `88a231de6b8ef54dd29c036ae5d930a9e21c7a78` against the PR diff and metadata, the current green CI/checks, and the empty prior review/comment history.

### 🟢 What looks good — the tool now emits an image marker

The tool no longer echoes the caller-supplied path string as the only file reference. It now emits `[IMAGE:<resolved absolute path>]` from the canonicalized `resolved_path`, so the output contract is explicit and does not depend on whether the caller used an absolute or workspace-relative spelling.

### 🟢 What looks good — Windows prefix handling stays local

The verbatim-prefix helper keeps the Windows follow-up contained to the marker emit site. It normalizes `\\?\C:\...` and `\\?\UNC\server\share\...` before the marker is written, without broadening unrelated parser behavior or touching provider/channel/runtime code.

### 🟢 What looks good — test coverage validates the new contract

The updated coverage checks the schema removal, confirms no bare base64 blob leaks into output, asserts absolute `[IMAGE:]` marker emission for normal and workspace-relative calls, and covers both verbatim-disk and verbatim-UNC normalization. That is the right level of proof for this one-file tool-output change.

Approved from my side.

Audacity88:
I checked the failing `Test` job. This looks like two unchanged `zeroclaw-channels` tests still expect the old hard vision-capability error path:

- `e2e_photo_attachment_rejected_by_non_vision_provider`
- `e2e_failed_vision_turn_does_not_poison_follow_up_text_turn`

On this PR, the runtime now intentionally degrades no-vision/no-vision-route image turns to text-only instead of aborting, so the dummy provider returns `chat-photo:ok`. That matches the new behavior, but the channel tests still assert `does not support vision`.

I think the fix is to update those two expectations to assert successful text-only degradation and no persisted `[IMAGE:]` marker content. The hard-error coverage should stay on the configured-but-non-vision `vision_model_provider` path, if that is still the intended error case.

NiuBlibing:
@Audacity88 I've fixed the error. Now the CI is green

PR #7447, opened 2026-06-10
fix(providers): wire timeout_secs config into native OpenAI provider (#6723)

### LINKED ISSUE #6723: Native OpenAI provider hardcodes 120s request timeout, silently ignores timeout_secs config
## Summary

`OpenAiProvider::http_client()` builds its reqwest client with a hardcoded **120-second** request timeout. The `[providers.models.openai] timeout_secs = N` config field is **never read** by the native OpenAI provider — only by `OpenAiCompatibleProvider`. Users setting `timeout_secs = 900` for slow local models get a request that dies at 119.998s with no diagnostic.

I lost a full day of homelab debugging on this — alfred-zeroclaw runs against a local llama-server backend (12s/100k-token prefill at ~770 t/s) via LiteLLM. Every big-prompt agent turn was failing at exactly 119.999s with `WARN: Provider call failed; error sending request for url ...`. Standard suspects exhausted: tcpdump of all four layers (alfred → Caddy → LiteLLM → llama-server) showed alfred's reqwest sending FIN+RST at the 120s mark while llama-server was still happily processing prefill at 88%.

## Evidence

- **Hardcoded literal**: `crates/zeroclaw-providers/src/openai.rs:344-350`
  ```rust
  fn http_client(&self) -> Result<Arc<reqwest::Client>> {
      build_runtime_proxy_client_with_timeouts("provider.openai", 120, 10)
      //                                                          ^^^ hardcoded
  }
  ```
- The helper at `crates/zeroclaw-config/src/schema.rs:4765-4789` wires `.timeout(Duration::from_secs(120))` onto the `ClientBuilder`.
- All three chat call sites in `openai.rs` (`:396, :441, :509`) use this client.
- `OpenAiProvider` has **no `timeout_secs` field**; `OpenAiCompatibleProvider` (`crates/zeroclaw-providers/src/compatible.rs:38, :253, :276, :364`) does, with `with_timeout_secs` setter and `let timeout = self.timeout_secs;` consumed in its client builder.
- `grep -n with_timeout_secs crates/zeroclaw-providers/src/openai.rs` returns zero matches.

## Repro

1. Configure `[providers.models.openai] timeout_secs = 900, base_url = "https://your-slow-backend"`
2. Send a chat completion that takes >120s upstream (e.g. a large prompt against a local model)
3. Observe failure at 119.999s with `error sending request for url (...)` and no upstream-side log of completion. Connect a packet capture and you'll see the client sending FIN+RST at that mark.

## Impact

Anyone pointing the native OpenAI provider at a slow upstream — local llama.cpp / vLLM / Ollama via OpenAI-compat shim, high-latency regional endpoints, big prompts — silently hits the 120s wall regardless of config. The error message ("error sending request") looks like a network issue, not a config-ignored issue. The workaround (`compatible.rs` provider, which DOES honor `timeout_secs`) isn't documented as the preferred path for OpenAI-compat backends.

## Suggested fix

Add `timeout_secs: Option<u64>` to `OpenAiProvider` and a `with_timeout_secs` setter (mirror `compatible.rs:253-279`). Pass it through `http_client()` (default to 120 if `None` for backward compat):

```rust
// openai.rs
pub struct OpenAiProvider {
    // ... existing fields ...
    timeout_secs: Option<u64>,
}

pub fn with_timeout_secs(mut self, timeout_secs: u64) -> Self {
    self.timeout_secs = Some(timeout_secs);
    self
}

fn http_client(&self) -> Result<Arc<reqwest::Client>> {
    let timeout = self.timeout_secs.unwrap_or(120);
    build_runtime_proxy_client_with_timeouts("provider.openai", timeout, 10)
}
```

And in the factory: call `.with_timeout_secs(...)` when `config.timeout_secs` is set. Pattern's already there in `compatible.rs` factory; just port it.

~30 LOC change. Happy to send a PR.

chengzhichao-xydt:
## Summary

- **Issue:** Closes #6723
- **What changed:** The native OpenAI provider (`OpenAiModelProvider`) now honors the `timeout_secs` config field instead of hardcoding a 120-second request timeout.
- **Why:** Users pointing the native OpenAI provider at slow upstreams (local llama.cpp / vLLM / Ollama via OpenAI-compat shim, high-latency regional endpoints) would see failures at exactly 120s regardless of their configured `timeout_secs`. The compatible provider path already honors this setting.
- **Scope boundary:** Confined to `crates/zeroclaw-providers/src/openai.rs` (struct, setter, http_client) and `factory.rs` (wire `opts.provider_timeout_secs`). No config schema, web UI, or docs changes.
- **Labels:** `bug`, `size: XS`, `risk: medium`, `provider`, `provider:openai`

## Files Changed

- `crates/zeroclaw-providers/src/openai.rs` — add `timeout_secs: Option<u64>` field, `with_timeout_secs()` setter, use configured value in `http_client()` (default 120)
- `crates/zeroclaw-providers/src/factory.rs` — wire `opts.provider_timeout_secs` through the factory

## Validation

- `cargo check -p zeroclaw-providers` — passes
- Backward compatible: when `timeout_secs` is `None`, the 120-second default is preserved
- All existing tests continue to pass (tests construct via `OpenAiModelProvider::new()` which uses the default `None` timeout)

## Security & Privacy Impact

## Compatibility

- Backward compatible? `Yes`
- Upgrade steps: None.

## Rollback

`git revert <sha>` restores the previous hardcoded 120-second timeout. No config schema or runtime changes are included.

### DIFF
--- a/crates/zeroclaw-providers/src/factory.rs
+++ b/crates/zeroclaw-providers/src/factory.rs
@@ -800,6 +800,9 @@ impl FamilyProviderFactory for OpenAIModelProviderConfig {
         if let Some(mt) = opts.provider_max_tokens {
             p = p.with_max_tokens(Some(mt));
         }
+        if let Some(t) = opts.provider_timeout_secs {
+            p = p.with_timeout_secs(t);
+        }
         Ok(Box::new(p))
     }
 }
--- a/crates/zeroclaw-providers/src/openai.rs
+++ b/crates/zeroclaw-providers/src/openai.rs
@@ -27,6 +27,9 @@ pub struct OpenAiModelProvider {
     base_url: String,
     credential: Option<String>,
     max_tokens: Option<u32>,
+    /// Per-request timeout in seconds. Falls back to 120 when unset,
+    /// matching the historical hardcoded default.
+    timeout_secs: Option<u64>,
 }
 
 #[derive(Debug, Serialize)]
@@ -216,12 +219,20 @@ impl OpenAiModelProvider {
                 .unwrap_or_else(|| BASE_URL.to_string()),
             credential: credential.map(ToString::to_string),
             max_tokens: None,
+            timeout_secs: None,
         }
     }
 
     /// Set the maximum output tokens for API requests.
     pub fn with_max_tokens(mut self, max_tokens: Option<u32>) -> Self {
         self.max_tokens = max_tokens;
+        self
+    }
+
+    /// Override the per-request timeout in seconds.
+    /// Falls back to 120 when unset, matching the historical default.
+    pub fn with_timeout_secs(mut self, timeout_secs: u64) -> Self {
+        self.timeout_secs = Some(timeout_secs);
         self
     }
 
@@ -368,9 +379,10 @@ impl OpenAiModelProvider {
     }
 
     fn http_client(&self) -> Client {
+        let timeout = self.timeout_secs.unwrap_or(120);
         zeroclaw_config::schema::build_runtime_proxy_client_with_timeouts(
             "model_provider.openai",
-            120,
+            timeout,
             10,
         )
     }


WareWolf-MoonWall (commented):
# Review — PR #7447
## `fix(providers): wire timeout_secs config into native OpenAI provider (#6723)`
**Author:** chengzhichao-xydt | **Verdict:** `--comment`

---

The logic is correct: the factory now feeds `opts.provider_timeout_secs` through `with_timeout_secs()` on `OpenAiModelProvider` in the same conditional pattern used everywhere else in `factory.rs`. The bug was real (120s hardcoded regardless of config), and this is the right fix. Two things I want on record before approving.

### 🟡 Warning — type shape diverges from the existing `compatible.rs` precedent

`OpenAiCompatibleModelProvider` (in `compatible.rs`) uses:

```rust
timeout_secs: u64  // field, default 120
pub fn with_timeout_secs(mut self, timeout_secs: u64) -> Self
```

This PR adds to `OpenAiModelProvider` (in `openai.rs`):

```rust
timeout_secs: Option<u64>  // field, default None
pub fn with_timeout_secs(mut self, timeout_secs: u64) -> Self
// ... unwrap_or(120) in http_client()
```

The `Option<u64>` shape is not wrong, but it's inconsistent with the companion provider's established pattern, and the two providers are often read side-by-side. The compatible provider's non-optional `u64` with a sensible default keeps the "no config" case explicit in `new()`. I'd prefer the native provider match that shape so the two structs stay parallel. Not a blocker, but worth aligning in a follow-up (or in this PR if it's still cheap).

### 🟡 Warning — validation ran `cargo check`, not `cargo test`; no new unit tests

`compatible.rs` has explicit unit tests for the setter at lines 5149–5155:

```rust
fn timeout_secs_defaults_to_120() { ... assert_eq!(p.timeout_secs, 120); }
fn with_timeout_secs_overrides_default() { ... assert_eq!(p.timeout_secs, 300); }
```

This PR has none for `OpenAiModelProvider`, and only ran `cargo check`. For a medium-risk change targeting a field that will be read on every single LLM request, I'd like to see at minimum the same two-case test coverage the compatible provider already has. The PR template for medium-risk changes should show `cargo test -p zeroclaw-providers` passing, not just `cargo check`.

### 🟢 What looks good — factory wiring is consistent and correct

The block added at `factory.rs` L790–L793 follows the exact same `if let Some(t) = opts.provider_timeout_secs { p = p.with_timeout_secs(t); }` idiom used by `apply_compat_options` for compatible providers. No duplicate state introduced: the provider struct is the canonical runtime home for this value; config is read once at construction through the factory.

### 🟢 What looks good — backward-compatible default preserved

Whether `timeout_secs` is `None` (missing from config) or the PR uses `Option<u64>`, the `unwrap_or(120)` in `http_client()` preserves the historical 120-second default. Existing deployments will not see behavior changes.

---

**No blockers. Aligning the type shape and adding unit tests would make this merge-ready for me.**

singlerider:
Thanks for this, @chengzhichao-xydt. You correctly diagnosed #6723 and got to it first (by about 32h). Closing as a duplicate of #7504, which lands the same fix with a slightly cleaner shape (the 120s default lives in the constructor as a plain `u64` rather than being resolved at the call site) and already carries an approving review. The credit for root-causing the hardcoded timeout is yours, and I have noted the supersede on #7504.

Nothing wrong with your implementation. This is purely about not landing the same change twice. Hope to see you on the next one.

PR #7449, opened 2026-06-10
fix(web): trim whitespace-only assistant content to prevent blank lines (#6702)

### LINKED ISSUE #6702: [Bug]: Dashboard assistant bubble accumulates blank lines for each tool-call card
### Affected component

tooling/ci

### Severity

S3 - minor issue

### Current behavior

In the Dashboard (apps/dashboard) conversation view, the assistant message bubble contains excessive whitespace whenever it includes tool-call cards. Each additional tool-call card adds an extra blank line inside the assistant bubble. When one assistant message contains multiple tool calls, the blank lines accumulate, pushing content apart and noticeably hurting readability.

### Expected behavior

There should be no extra blank lines around tool-call cards. The spacing between assistant text and tool-call cards should be consistent and compact, and the amount of whitespace inside the bubble should remain stable regardless of how many tool-call cards the message contains — it should not grow linearly with the number of tool calls.

### Steps to reproduce

```bash
1. Start zeroclaw daemon and the dashboard
2. In the Dashboard, send a message to the agent that triggers multiple tool calls
3. Observe the assistant reply bubble: each tool-call card is surrounded by extra blank lines
4. The more tool calls the message contains, the more blank space appears inside the bubble
```

### Impact

Affected users: all Dashboard users viewing agent output
Frequency: always (whenever an assistant message contains tool calls)
Consequence: long conversations or messages with many tool calls require excessive scrolling; the layout looks loose and hurts the reading experience.

### Logs / stack traces

Image

### ZeroClaw version

a090cbce3

### Rust version

rustc 1.94.1 (e408947bf 2026-03-25)

### Operating system

Linux 7.0.3-1-MANJARO x86_64

### Regression?

Unknown

### Pre-flight checks

chengzhichao-xydt:
## Summary

- **Issue:** Closes #6702
- **What changed:** The `done`/`message` WebSocket handler now trims the assistant content before deciding whether to create a message entry. Whitespace-only content (e.g. `"\n\n"` emitted by some models alongside tool_calls) is silently skipped.
- **Why:** Some models (Qwen3.6, MiniMax-M2.7 via vllm) generate tool_calls with empty or whitespace-only text content. The handler created a message entry for this content, which rendered as blank lines inside the assistant bubble — one per tool-call card, linearly accumulating visual space.
- **Scope boundary:** Confined to `web/src/contexts/AgentContext.tsx` (3-line change in the done/message handler). No API, config, or backend changes.
- **Labels:** `bug`, `size: XS`, `risk: low`, `gateway`, `web`

## Files Changed

- `web/src/contexts/AgentContext.tsx` — trim raw_content before creating message entry; skip when only whitespace

## Validation

- The fix is minimal: trim + guard, no structural changes to the message pipeline
- Existing messages with real content are unaffected (trim is lossless for non-whitespace content)
- Tool-call card rendering is unchanged (tool_call/tool_result paths are unaffected)

## Security & Privacy Impact

## Compatibility

- Backward compatible? `Yes`
- Upgrade steps: None.

## Rollback

`git revert <sha>` restores the previous behavior. No config schema or backend changes are included.


WareWolf-MoonWall (approved):
# Review — PR #7449
## `fix(web): trim whitespace-only assistant content to prevent blank lines (#6702)`
**Author:** chengzhichao-xydt | **Verdict:** `--comment`

---

The fix itself is mechanically correct and hits the right spot in the message pipeline. I have two things worth calling out before I'd flip to approve.

### 🟡 Warning — validation evidence missing literal command output

The PR template requires literal command output in the Validation Evidence section. This PR's section reads as prose justification rather than demonstrating a concrete before/after. For a visual rendering bug, that's understandably awkward — but even a short screen capture description ("ran against a Qwen3.6 conversation, blank line in assistant bubble before, gone after") or a note that this was manually verified against the affected model would satisfy the template requirement. As written, it's a rationale section, not an evidence section.

### 🔵 Suggestion — consider whether `.trim()` is the right boundary operation

The fix applies `String.prototype.trim()` to `raw_content` before the `if (content)` guard, which strips ALL leading and trailing whitespace — not just the all-whitespace case. In practice this is fine for LLM output, but there's a theoretical edge where a model intentionally pads content with leading whitespace (some formatting workflows do this). A slightly tighter variant that only discards the all-whitespace case while otherwise leaving content untouched:

```typescript
const content = raw_content.trim() ? raw_content : '';
```

This discards the message only when it's pure whitespace, but stores the original `raw_content` (with its whitespace intact) when there is real content. Worth a quick consideration; not a blocker.

### 🟢 What looks good — correct type safety

I checked: `pendingContentRef = useRef('')` (line 116) types `current` as `string`, never `undefined`. The `??` chain therefore always resolves to a `string`, so `.trim()` cannot throw a `TypeError` even when both `msg.full_response` and `msg.content` are absent. The safety concern that sometimes comes up with this pattern does not apply here.

### 🟢 What looks good — scalpel-precise scope

Three effective lines in a single file, zero backend or config surface impact, non-streaming path is the correct place to intercept this. The `chunk` streaming path correctly remains untouched — stripping mid-stream would cause visual artifacts; discarding only at `done/message` is the right call.

---

**Approving.** The fix is correct, the type safety holds, and there are no other active blocks. The validation evidence gap and the `.trim()` edge case are worth addressing but neither is a reason to hold merge on a three-line web fix.

PR #7451, opened 2026-06-10
fix(memory): bind postgres schema version as int4

### LINKED ISSUE #7421: [Bug]: Postgres memory backend fails to initiailize
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

```
> zeroclaw agent
Error: error serializing parameter 0

Caused by:
    cannot convert between the Rust type `i64` and the Postgres type `int4`
```

### Expected behavior

Should work.

### Steps to reproduce

```bash
Minimal configuration: db_url is set.
```

### Impact

Prevents using Postgres backend.

### Logs / stack traces

```text
Database contents after run:

zc=# \d+
                                               List of relations
 Schema |           Name           | Type  |  Owner   | Persistence | Access method |    Size    | Description
--------+--------------------------+-------+----------+-------------+---------------+------------+-------------
 public | agents                   | table | zc       | permanent   | heap          | 16 kB      |
 public | memories                 | table | zc       | permanent   | heap          | 8192 bytes |
 public | schema_version           | table | zc       | permanent   | heap          | 8192 bytes |
 public | vchordrq_sampled_queries | view  | postgres | permanent   |               | 0 bytes    |
```

### ZeroClaw version

0.8.0-beta-2

### Rust version

rustc 1.96.0 (ac68faa20 2026-05-25) (built from a source tarball)

### Operating system

FreeBSD stick 15.1-STABLE FreeBSD 15.1-STABLE #45 stable/15-n283851-7546f06f5e06-dirty: Fri Jun  5 11:34:47 EEST 2026     ###@stick:/usr/obj/usr/src/amd64.amd64/sys/MIN amd64

### Regression?

No, first-time setup

### Pre-flight checks

jstar0:
## Summary

Fixes #7421 by binding the Postgres memory schema version as an `i32`, matching the `INTEGER` / `int4` column used by `schema_version.version`.

## Why

The Postgres memory migration creates `schema_version.version` as `INTEGER NOT NULL`, but the insert previously bound `SQLITE_MEMORY_SCHEMA_VERSION` directly. That constant is an `i64`, and the `postgres` crate rejects serializing an `i64` parameter into an `int4` column, which matches the reported startup failure: `cannot convert between the Rust type i64 and the Postgres type int4`.

## Changes

- Added a small Postgres-only helper that derives a checked `i32` value from the existing canonical memory schema version.
- Updated the Postgres memory migration to bind that `i32` value.
- Added a focused regression test for the `ToSql` / `INT4` bind contract.

This keeps `SQLITE_MEMORY_SCHEMA_VERSION` as the single source of truth and does not change the schema or SQLite behavior.

## Verification

```bash
cargo test -p zeroclaw-config --features memory-postgres memory_schema_version_binds_to_postgres_int4
cargo test -p zeroclaw-config --features memory-postgres schema::v2::fs_db_migration_tests
cargo test -p zeroclaw-config --features memory-postgres
cargo test -p zeroclaw-memory --features memory-postgres
cargo fmt --all -- --check
cargo clippy -p zeroclaw-config --features memory-postgres --all-targets -- -D warnings
cargo clippy -p zeroclaw-memory --features memory-postgres --all-targets -- -D warnings
```

### DIFF
--- a/crates/zeroclaw-config/src/schema/v2.rs
+++ b/crates/zeroclaw-config/src/schema/v2.rs
@@ -3213,6 +3213,12 @@ fn copy_dir_recursive(src: &Path, dst: &Path) -> MigResult<()> {
 /// PRAGMA detection.
 pub const SQLITE_MEMORY_SCHEMA_VERSION: i64 = 1;
 
+#[cfg(feature = "memory-postgres")]
+fn postgres_memory_schema_version() -> MigResult<i32> {
+    i32::try_from(SQLITE_MEMORY_SCHEMA_VERSION)
+        .context("Postgres memory schema version exceeds INTEGER range")
+}
+
 /// Migrate a SQLite memory database to the V3 multi-agent shape.
 ///
 /// Adds the `agents` table, the `agent_id` column on `memories`,
@@ -3602,13 +3608,14 @@ pub fn migrate_postgres_memory_to_v3(
             applied_at TIMESTAMPTZ NOT NULL
         );"
     ))?;
+    let memory_schema_version = postgres_memory_schema_version()?;
     client.execute(
         &format!(
             "INSERT INTO {schema_ident}.schema_version (component, version, applied_at) \
              VALUES ('memories', $1, NOW()) \
              ON CONFLICT (component) DO UPDATE SET version = EXCLUDED.version, applied_at = EXCLUDED.applied_at"
         ),
-        &[&SQLITE_MEMORY_SCHEMA_VERSION],
+        &[&memory_schema_version],
     )?;
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
I reviewed PR #7451 at current head 27e99248 against the live PR body, linked issue #7421, comments/reviews, checks, and the one-file diff in `crates/zeroclaw-config/src/schema/v2.rs`. CI is green at this head. I did not run local cargo validation for this pass.

### 🟢 What looks good — The bind type now matches the Postgres column

This is the right shape for the reported startup failure. The migration creates `schema_version.version` as `INTEGER NOT NULL`, so binding an `i32` for Postgres matches the `int4` column instead of asking the driver to serialize the canonical `i64` SQLite schema-version constant directly.

I also like that the helper keeps `SQLITE_MEMORY_SCHEMA_VERSION` as the single source of truth and performs a checked conversion at the Postgres boundary. That makes the fix local to the type contract without inventing a second schema-version constant.

### 🟢 What looks good — The regression targets the driver contract

The added `memory_schema_version_binds_to_postgres_int4` test covers the specific failure mode from #7421: the value still equals the canonical schema version after conversion, and the resulting type is accepted for Postgres `INT4`. That is the right focused coverage for this narrow bind fix.

Approving this head. Normal final merge freshness still applies, but I do not see requested changes for this fix.

PR #7463, opened 2026-06-10
fix(runtime): use agent_workspace_dir for load_skills_for_agent calls

### LINKED ISSUE #7236: [Bug]: Incorrect workspace_dir passed to load_skills_for_agent prevents loading skills from data/skills/

Yyukan:
## Summary

- **Base branch:** `master`
- **What changed and why:** `load_skills_for_agent()` was called with `&config.data_dir` at 3 call sites, causing skills in `<agents/<alias>/workspace/skills/>` to be silently ignored (the loader searched `<data_dir>/skills/` instead). Fixed by extracting a shared `load_skills_for_agent_from_config()` helper that resolves the correct per-agent workspace via `config.agent_workspace_dir(agent_alias)`, and updating all three call sites to use it.
- **Scope boundary:** Extracts a shared helper for skill loading, updates 3 call sites to use it. Adds regression tests that exercise the helper. No config/schema changes, no new dependencies, no new behavior paths.
- **Blast radius:** Agent boot path (`loop_.rs`), message processing path (`loop_.rs`), and WebSocket/daemon path (`agent.rs`). All three now use the same helper to load workspace skills from the correct directory.
- **Linked issue:** Fixes #7236
- **Labels:** `bug`, `risk: high`, `agent`, `runtime`, `skills`

```
cargo clippy --all-targets -- -D warnings  # no issues
cargo test                          # 896 passed, 7 ignored
cargo test -p zeroclaw-runtime workspace_dir_regression_tests  # 2 passed
```

- **Beyond CI:** Inspected all 4 callers of `load_skills_for_agent`; confirmed 3 were using `data_dir` (now use shared helper) and 1 was already correct (orchestrator `mod.rs:8172`).
- **Regression tests:** Added `workspace_dir_regression_tests` module with two tests that exercise the production helper:
  - `load_skills_for_agent_from_config_uses_workspace_dir_not_data_dir`: constructs a Config with distinct `data_dir` and `agent_workspace_dir`, places a skill only in the workspace, and asserts the helper loads it (would fail if helper reverted to `data_dir`)
  - `load_skills_for_agent_from_config_empty_bundles_uses_workspace_dir`: pins contract that empty `skill_bundles` still discovers workspace skills correctly

## Security & Privacy Impact

- Secrets/tokens/credentials changed? `No`
- PII in diff? `No`

## Compatibility

- Backward compatible? `Yes` — users who placed skills in agent workspace directories will now have them correctly loaded.
- Config/CLI changed? `No`

## Rollback

- `git revert <commit-sha>` — three commits, fully revertable.


Audacity88 (changes requested):
I reviewed current head `e523e61210565f4da0de0a9a7960e25abfe290b0` against `master`, including the linked issue #7236, live PR metadata, green CI, the diff, and all current `load_skills_for_agent()` callers. I did not run local validation.

### 🟢 What looks good — The call-site fix is the right shape

The three changed runtime call sites now pass `config.agent_workspace_dir(agent_alias)` into `load_skills_for_agent()`, which matches the loader's contract: it reads workspace skills from `workspace_dir.join("skills")`. I also checked the remaining orchestrator caller, and it was already using the per-agent workspace before calling the same loader. So the implementation direction lines up with the bug report.

### 🔴 Blocking — The bug fix needs a behavior regression test before closing #7236

This PR closes #7236, but the validation only proves that the existing test suite still passes and that the touched call sites were inspected. It does not add a regression that would have failed when these runtime paths passed `config.data_dir`, nor one that proves a skill placed under `<install>/agents/<alias>/workspace/skills/` is visible through the fixed runtime path.

That matters because this is exactly the kind of silent runtime bug that green CI can miss: `load_skills_for_agent()` still compiles with either directory, and the wrong directory simply returns no workspace skills. Without an executable behavior test, the central claim remains easy to regress during future runtime or skill-loading cleanup.

Please add a focused regression at the lowest practical layer. For example, construct a config where `data_dir` and `agent_workspace_dir(agent_alias)` are distinct, place a minimal valid skill under the agent workspace's `skills/` directory, and exercise the runtime skill-loading path so the test fails if the caller passes `config.data_dir`. If testing the full `run` / `process_message` path is too heavy, extracting the shared "resolve the workspace used for skill loading" step into a small helper and testing that would still pin the contract better than caller inspection alone.

Audacity88 (changes requested):
I re-reviewed current head 3614f21ff945991049c212b4e82560ca2c84b56b against master, including the linked issue #7236, the prior review, the updated diff, CI, and the current `load_skills_for_agent()` callers. I did not run local validation.

### 🟢 What looks good — The added tests prove the loader contract

The new `workspace_dir_regression_tests` module is useful coverage for the low-level loader behavior. It sets up distinct `data_dir` and agent-workspace paths, places a skill only under the agent workspace, and proves that `load_skills_for_agent()` finds the skill only when given the workspace path. That is the right contract for the loader itself, and the three production call-site changes still look like the right implementation direction.

### 🔴 Blocking — The regression still does not cover the fixed production path

The remaining issue is that these tests would still pass if the three runtime call sites changed back to `&config.data_dir`. They call `load_skills_for_agent()` directly with the correct and incorrect paths, but they do not exercise `Agent::from_config`, `run`, `process_message`, or a shared helper that those call sites now use.

That means the PR now documents why `data_dir` is wrong, but it still does not add an executable guard for the actual bug fixed by this diff: production runtime paths passing the wrong directory into the loader. This is the same silent-failure shape as the original issue, so I do not want to clear the prior blocker until the test would fail on the pre-fix call-site behavior.

The smallest fix does not need to stand up the full agent loop. A good low-friction option would be to extract the shared production step into a small helper, for example a helper that loads agent skills from `config.agent_workspace_dir(agent_alias)`, use that helper at the three call sites, and test the helper with the same distinct-`data_dir` fixture. A focused test through one corrected production path would also work. The key requirement is that the test fails if runtime skill loading goes back to `config.data_dir`.

Audacity88 (approved):
Context checked: live PR #7463 at current head e6f2c9639b494084b8286cf3c5f846067ff7e229, the prior Audacity88 blockers, linked #7236 behavior, the current diff, current skill-loading callers, and green GitHub checks. I did not run local cargo validation.

### ✅ Resolved — Production skill loading now goes through the tested helper

The previous blocker is addressed. `load_skills_for_agent_from_config()` now resolves `config.agent_workspace_dir(agent_alias)` once and delegates to `load_skills_for_agent()`. The three runtime call sites that previously made their own directory choice now all call that helper: `Agent::from_config`, `run`, and `process_message`.

That means the production paths no longer each re-decide whether to pass `config.data_dir` or the agent workspace. The remaining direct caller in the channel orchestrator was already resolving `config.agent_workspace_dir(agent_alias)` before calling the lower-level loader.

### ✅ Resolved — The regression now covers the path that could regress

The new regression tests exercise the shared production helper with distinct `data_dir` and agent workspace directories, then place a skill only under the agent workspace. That is the executable guard the prior review was asking for: if the helper goes back to `config.data_dir`, the test fails.

The empty-bundles case is also useful because it pins the fallback contract that workspace discovery still happens even when the agent has no configured skill bundles.

### 🟢 What looks good — The source of truth is clearer now

This is a small improvement beyond the bug fix itself. The lower-level loader still owns loading from `<workspace>/skills`, while the new helper owns resolving the correct per-agent workspace from config. That split matches the single-source-of-truth rule and makes future runtime cleanup less likely to reintroduce the silent "wrong directory, no skills" failure.

This clears my prior blockers. I did not change labels, milestone, or PR metadata in this review pass.

WareWolf-MoonWall (approved):
Reviewed: head `d561c22b` (post-merge-sync), functional head `b4d9b1df` (`refactor(skills): extract load_skills_for_agent_from_config helper`). @Audacity88 approved at `a86bd3317ef9`. Fixes #7236.

### ✅ Prior blockers resolved — helper extracted, regression tests cover the production path

Audacity88's two CHANGES_REQUESTED reviews both requested a regression that would fail if the three runtime call sites reverted to `&config.data_dir`. The `load_skills_for_agent_from_config` helper satisfies this exactly: all three call sites (`agent.rs:1442`, `loop_.rs:3759`, `loop_.rs:5098`) now use the helper, and the `workspace_dir_regression_tests` module tests the *helper itself* with distinct `data_dir` and agent-workspace paths — a test that would fail if the helper reverted to `config.data_dir`.

### 🟢 What looks good — DRY compliant: workspace directory resolved on demand

`load_skills_for_agent_from_config` calls `config.agent_workspace_dir(agent_alias)` at call time. No cached copy of the workspace path is stored. The doc comment explicitly states: "Source of truth for the workspace directory is `config.agent_workspace_dir(agent_alias)`; this helper resolves it on every call so config reloads take effect." ✓

### 🟢 What looks good — the fix is complete at all three runtime paths

- `Agent::from_config` (daemon/WebSocket setup path) ✓
- `run()` (agent boot path) ✓
- `process_message()` (message processing path) ✓

The orchestrator call site (`mod.rs:8172`) was already using the correct workspace per the PR description, and it's not touched.

### 🟢 What looks good — validation is appropriate for risk: high

`cargo fmt`, `cargo clippy --all-targets -D warnings`, `cargo test` (896 passed, 7 ignored), and focused `cargo test -p zeroclaw-runtime workspace_dir_regression_tests` (2 passed) cover the changed surfaces. This is the right battery for a high-risk runtime fix.

Approving.

PR #7465, opened 2026-06-10
fix(runtime): preserve delegate memory namespace

### LINKED ISSUE #7080: bug(delegate): preserve memory namespace across agentic delegate modes

jstar0:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Fixes delegate memory isolation for #7080 by rebuilding delegate-visible memory tools against the target agent's memory handle.
  - Keeps non-memory delegate tools on the existing parent-tool wrapping path.
  - Carries the memory handle through synchronous, background, and parallel agentic delegate paths so all three modes use the same target namespace behavior.
  - Adds regressions that prove delegated memory writes land under the target agent UUID and not the caller UUID.
- **Scope boundary:** No changes to non-agentic delegation, model-provider routing, memory backend configuration, or the general parent-tool registry.
- **Blast radius:** Runtime delegate tool behavior, specifically agentic delegates that allow memory tools.
- **Linked issue(s):** Fixes #7080
- **Labels:** `bug`, `risk: high`, `size: L`, `runtime`, `memory`, `tool`, `tool: delegate`

```bash
cargo test -p zeroclaw-runtime rebinds_memory_tools_to_target_agent_scope
cargo test -p zeroclaw-runtime delegate
cargo fmt --all -- --check
```

- **Commands run and tail output:** Author reported all four commands above passing. GitHub Actions are also passing on the submitted head, including Format, Test, Security, platform builds, Lint, Check variants, Benchmarks Compile, and CI Required Gate.
- **Beyond CI — what did you manually verify?** Regression tests cover synchronous, background, and parallel agentic delegate calls. Each path stores a unique memory value and asserts that it appears under the target agent UUID and not under the caller UUID.
- **If any command was intentionally skipped, why:** None reported.

- If any `Yes`, describe the risk and mitigation: Not applicable.

## Compatibility (required)

- If `No` or `Yes` to either: Not applicable.

- **Fast rollback command/path:** `git revert <merged-sha>`
- **Observable failure symptoms:** Delegated agents that use memory tools store or recall data under the caller/default namespace instead of the delegated target namespace; look for delegate runs involving `memory_store`, `memory_recall`, `memory_forget`, `memory_export`, or `memory_purge`.


Audacity88 (approved):
I checked the current PR snapshot, the diff for `crates/zeroclaw-runtime/src/tools/delegate.rs`, the linked #7080 context, the empty review/comment state, and the passing public checks. I did not rely on unpublished validation as approval evidence.

I am approving this. It addresses the bug tracked in #7080: agentic delegates now run memory tools against the delegated agent's memory scope instead of silently reusing the caller/default memory namespace.

### 🟢 What looks good — the fix happens at the delegate boundary

The important part is that the PR does not just pass another identifier through the model loop. It rebuilds the delegate-visible memory tools with the target agent's memory handle and target policy before constructing the sub-agent tool set. That keeps the namespace decision tied to the delegated agent, while non-memory tools still follow the existing parent-tool wrapping path.

### 🟢 What looks good — the risky execution modes are covered

The regression coverage hits the three paths from #7080: synchronous agentic delegation, `background: true`, and `parallel`. The tests also check the negative side of the boundary by asserting that the stored value appears under the target UUID and not under the caller UUID. That is the right proof for this bug.

### 🟢 What looks good — the tests exercise the boundary directly

The fixture keeps the assertion focused on the namespace split instead of only checking that delegation completes. That matters here because the bug was not that memory tools failed outright; it was that they succeeded against the wrong agent scope.

Issue #7376, opened 2026-06-08
[Bug]: zerocode Dashboard hides unavailable/error states and labels history as active sessions


Audacity88:
## Affected component

unknown

This is in the zerocode Dashboard TUI, but the current bug template does not have a zerocode/TUI component option.

## Severity

S2 - degraded behavior

## Current behavior

While testing zerocode on macOS, several Dashboard states either hide the useful error or present stale/loading data as if it were current:

1. The Cost tab can remain on `Loading...` when cost tracking is unavailable. The daemon-side RPC error is `Cost tracking is not available`, but the Dashboard only updates `cost` on successful `cost_query` responses, so the user never sees the unavailable/error state.
2. The Sessions tab can briefly show `Sessions (0)` before the list loads. That reads as a real empty state even though data is still loading.
3. The Agents detail pane prints `zerocode: missing i18n key: zc-dashboard-detail-sessions`.
4. Agent session counts are labelled as live activity even when they include persisted session history. The UI shows strings like `default (8 active)` or `sessions: 4`, while the runtime status path reports `active_sessions` as the max of live RPC sessions and persisted session metadata.

## Expected behavior

Dashboard tabs should distinguish loading, empty, unavailable, and error states.

The Cost tab should render an explicit unavailable/error message when the daemon returns `Cost tracking is not available`.

Sessions should not show a temporary zero count as a real empty state before the first list response arrives.

All displayed i18n keys should resolve.

Agent session counts should not call persisted history `active`. Either count only live in-memory sessions as active, or split the UI into live sessions and saved/history sessions.

## Steps to reproduce

```bash
1. Launch zerocode in a terminal.
2. Open Dashboard.
3. Switch to Cost when the daemon has no cost tracker configured.
4. Observe whether the tab stays on Loading instead of showing an unavailable/error state.
5. Switch to Sessions immediately after launch or reconnect.
6. Observe whether Sessions (0) appears before the list has loaded.
7. Open Dashboard -> Agents and inspect the list/detail text for session counts and missing i18n key output.
```

## Impact

Affected users: zerocode users reading the Dashboard for daemon status, cost state, sessions, and agents.

Frequency: observed during manual zerocode dogfooding. The source paths still support the same behavior on current `master`.

Consequence: the Dashboard can make unavailable cost tracking look like a hung load, can make loading sessions look empty, can leak a raw missing-i18n diagnostic into the UI, and can imply that old persisted sessions are actively running.

[...]

PR #7471, opened 2026-06-11
fix(config): match config-list prefix on segment boundaries, not string prefix

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The RPC `ConfigList` handler and gateway `/api/config/list` filtered config fields with a bare `starts_with(prefix)`, so opening agent `agents.aaa` in zerocode's Config pane also returned fields for string-prefixed sibling aliases (`agents.aaalore`, `agents.aaatools`) — reported on Discord with screenshots.
  - Added `field_visibility::path_matches_prefix`: matches the exact path or a child at a `.` segment boundary; dot-terminated sub-table prefixes keep existing semantics.
  - Both handlers now use the shared helper; a regression test enumerates the sibling-alias case.
- **Scope boundary:** Does not change zerocode UI code, the config schema, or `is_excluded` sub-table semantics.
- **Blast radius:** Any consumer of RPC `config.list` or gateway `/api/config/list` with a prefix (zerocode Config pane, web dashboard config forms). Strictly narrows result sets to correct ones.
- **Linked issue(s):** None filed; reported via Discord. Related #6862 (same `/api/config/list` surface; that issue tracks the SPA fallback defect, formerly also #7253).
- **Labels:** `bug`, `config`, `gateway`, `runtime`, `zerocode` (path labeler will add `size:*`; risk label per maintainer)

  - `cargo fmt --all -- --check` → clean, exit 0
  - `cargo clippy --all-targets -- -D warnings` → `Finished 'dev' profile [unoptimized + debuginfo] target(s)`, 0 warnings/errors
  - `cargo test` → 9 suites, 896 passed, 0 failed; targeted crates: zeroclaw-config lib `785 passed; 0 failed`, zeroclaw-gateway lib `243 passed; 0 failed`, zeroclaw-runtime lib `2039 passed; 0 failed; 1 ignored`
  - New test: `field_visibility::tests::path_matches_prefix_requires_segment_boundary ... ok`
- **Beyond CI — what did you manually verify?** Reproduced the matching logic against the reporter's exact alias set (`aaa`, `aaalore`, `aaatools`) in the regression test. NOT verified: live zerocode TUI session against a running daemon.

- PII, real identities, or personal data in diff, tests, fixtures, or docs? No (alias names from the report are generic strings)

## Compatibility (required)

- Backward compatible? Yes — strictly narrows over-broad responses to the documented per-section scope

`git revert <sha>`; single self-contained commit, no flags or config toggles. Failure symptom would be config panes showing no fields for a valid section prefix.

### DIFF
--- a/crates/zeroclaw-config/src/field_visibility.rs
+++ b/crates/zeroclaw-config/src/field_visibility.rs
@@ -68,6 +68,16 @@ pub fn is_excluded(path: &str, excludes: &[String]) -> bool {
         .any(|e| path == e || (e.ends_with('.') && path.starts_with(e)))
 }
 
+/// Test whether `path` equals `prefix` or sits beneath it at a `.` segment
+/// boundary. A bare `starts_with` is wrong here: prefix `agents.aaa` must
+/// not match `agents.aaalore.workspace`.
+pub fn path_matches_prefix(path: &str, prefix: &str) -> bool {
+    match path.strip_prefix(prefix) {
+        Some(rest) => rest.is_empty() || rest.starts_with('.') || prefix.ends_with('.'),
+        None => false,
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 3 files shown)


Audacity88 (approved):
I checked the live PR state, prior discussion/reviews, the current diff against `master`, and the relevant config/gateway/runtime paths. There are no prior comments or active review blocks on this PR. I also checked the visible CI set, including Lint, Test, Security, platform builds, and CI Required Gate.

### 🟢 What looks good — segment-boundary matching is the right fix

The shared `path_matches_prefix` helper fixes the actual failure mode without creating a second source of truth for the gateway and RPC paths. Matching either the exact path or a `.` segment boundary keeps `agents.aaa` from pulling in siblings like `agents.aaalore`, while preserving the old no-filter behavior for an empty prefix and the existing dot-terminated sub-table behavior.

The two call sites now use the same helper, and the regression coverage pins the important cases: exact match, child paths, sibling aliases that share a string prefix, top-level sections, dot-terminated prefixes, unrelated paths, and empty-prefix matching. That is the right level of coverage for this narrow config-list bug.

### 🔵 Suggestion — remove the unrelated issue reference

The new test comment says `#7376-class bug`, but #7376 is an unrelated closed zerocode Dashboard issue. This is not a blocker for the fix, but I would remove that issue-number reference before merge so the test does not point future readers at the wrong tracker item.

Approving this head.

PR #7476, opened 2026-06-11
fix(zerocode): avoid hardcoded vi editor fallback

### LINKED ISSUE #7469: [Bug]: Default using "vi" but container does not include it
### Affected component

unknown

### Severity

S3 - minor issue

### Current behavior

Image

### Expected behavior

Can edit the desired setting.

### Steps to reproduce

```bash
```

### Impact

### Logs / stack traces

```text

```

### ZeroClaw version

0.80-beta

### Rust version

_No response_

### Operating system

Dockerfile.debian

### Regression?

Unknown

### Pre-flight checks

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** Replace hardcoded `vi` editor fallbacks with editor discovery that prefers `VISUAL`, then `EDITOR`, then installed terminal editors. This now covers all reviewed launch paths: queued-message editing, the zerocode config manager external editor, and the root skills editor flow.
- **Scope boundary:** Editor discovery only. This does not change queued-message storage, config serialization, skill file format, or editor buffer handling.
- **Blast radius:** Low-to-medium. Affects only external editor launch paths; no runtime path is touched unless a user explicitly opens an external editor.
- **Linked issue(s):** Fixes #7469
- **Labels expected:** `bug`, `area: zerocode`

```bash
git diff --check
cargo test -p zerocode chat
cargo check --bin zeroclaw
```

```text
$ git diff --check
# no output

$ cargo test -p zerocode chat
running 70 tests
...
test result: ok. 70 passed; 0 failed; 0 ignored; 0 measured; 252 filtered out; finished in 0.17s

$ cargo check --bin zeroclaw
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.85s
```

- **Beyond CI — what did you manually verify?** I checked the diff for all hardcoded editor launch paths mentioned in review and confirmed the PR now covers the queued-message editor, config manager editor, and `src/skills/mod.rs` editor flow.
- **If any command was intentionally skipped, why:** Full workspace test suite was not run because this change is limited to editor discovery and targeted tests plus root binary check cover the touched paths.

- **New permissions, capabilities, or file system access scope?** No. It only changes how the existing editor command is selected.

## Compatibility (required)

- **Backward compatible?** Yes. Existing `VISUAL` and `EDITOR` values remain preferred.
- **Config / env / CLI surface changed?** No new env vars or CLI flags. The fallback behavior now reports a clear error when no editor can be found instead of assuming `vi` exists.
- **Upgrade steps:** None.

## Rollback

Low-risk rollback: revert the PR commits.

### DIFF
--- a/apps/zerocode/src/chat.rs
+++ b/apps/zerocode/src/chat.rs
@@ -4360,13 +4360,37 @@ fn labelled_clipboard_text(entry: &ChatEntry) -> String {
     }
 }
 
-/// Suspend the TUI, open `$EDITOR` with `content`, return the edited text.
+fn editor_from_env_or_path() -> Option<String> {
+    std::env::var("VISUAL")
+        .ok()
+        .filter(|value| !value.trim().is_empty())
+        .or_else(|| {
+            std::env::var("EDITOR")
+                .ok()
+                .filter(|value| !value.trim().is_empty())
+        })
+        .or_else(|| {
+            ["nano", "vi", "vim", "editor"]
+                .into_iter()
+                .find(|candidate| executable_on_path(candidate))
+                .map(str::to_string)
+        })
+}
+
+fn executable_on_path(name: &str) -> bool {
+    let Some(paths) = std::env::var_os("PATH") else {
+        return false;
+    };
+    std::env::split_paths(&paths).any(|dir| dir.join(name).is_file())
+}
+
+/// Suspend the TUI, open `$VISUAL` / `$EDITOR` with `content`, return the edited text.
 /// Restores raw mode and alternate screen before returning.
-/// Falls back to `content` unchanged if `$EDITOR` is unset or the process fails.
+/// Falls back to `content` unchanged if no editor is available or the process fails.
 pub async fn open_editor_for_content(content: &str) -> String {
-    let editor = std::env::var("EDITOR")
-        .or_else(|_| std::env::var("VISUAL"))
-        .unwrap_or_else(|_| "vi".to_string());
+    let Some(editor) = editor_from_env_or_path() else {
+        return content.to_string();
+    };
 
     let tmp = match tempfile::NamedTempFile::new() {
         Ok(f) => f,
--- a/apps/zerocode/src/chat.rs
+++ b/apps/zerocode/src/chat.rs
@@ -4360,35 +4360,11 @@ fn labelled_clipboard_text(entry: &ChatEntry) -> String {
     }
 }
 
-fn editor_from_env_or_path() -> Option<String> {
-    std::env::var("VISUAL")
-        .ok()
-        .filter(|value| !value.trim().is_empty())
-        .or_else(|| {
-            std::env::var("EDITOR")
-                .ok()
-                .filter(|value| !value.trim().is_empty())
-        })
-        .or_else(|| {
-            ["nano", "vi", "vim", "editor"]
-                .into_iter()
-                .find(|candidate| executable_on_path(candidate))
-                .map(str::to_string)
-        })
-}
-
-fn executable_on_path(name: &str) -> bool {
-    let Some(paths) = std::env::var_os("PATH") else {
-        return false;
-    };
-    std::env::split_paths(&paths).any(|dir| dir.join(name).is_file())
-}
-
 /// Suspend the TUI, open `$VISUAL` / `$EDITOR` with `content`, return the edited text.
 /// Restores raw mode and alternate screen before returning.
 /// Falls back to `content` unchanged if no editor is available or the process fails.
 pub async fn open_editor_for_content(content: &str) -> String {
-    let Some(editor) = editor_from_env_or_path() else {
+    let Some(editor) = crate::editor::editor_from_env_or_path() else {
         return content.to_string();
     };
 
--- a/apps/zerocode/src/config_manager.rs
+++ b/apps/zerocode/src/config_manager.rs
@@ -3917,15 +3917,9 @@ fn edit_in_external_editor(
     content: &str,
     filename_hint: &str,
 ) -> Result<String, String> {
-    let editor = std::env::var("VISUAL")
-        .or_else(|_| std::env::var("EDITOR"))
-        .unwrap_or_else(|_| {
-            if cfg!(windows) {
-                "notepad".into()
-            } else {
-                "vi".into()
-            }
-        });
+    let Some(editor) = crate::editor::editor_from_env_or_path() else {
+        return Err("no external editor found; set VISUAL or EDITOR".to_string());
+    };
 
     // Write content to a temp file with the right extension.
     let dir = std::env::temp_dir();
--- a/apps/zerocode/src/editor.rs
+++ b/apps/zerocode/src/editor.rs
@@ -1,0 +1,32 @@
+pub(crate) fn editor_from_env_or_path() -> Option<String> {
+    std::env::var("VISUAL")
+        .ok()
+        .filter(|value| !value.trim().is_empty())
+        .or_else(|| {
+            std::env::var("EDITOR")
+                .ok()
+                .filter(|value| !value.trim().is_empty())
+        })
+        .or_else(|| {
+            fallback_editors()
+                .iter()
+                .copied()
+                .find(|candidate| executable_on_path(candidate))
+                .map(str::to_string)
+        })
+}
+
+fn executable_on_path(name: &str) -> bool {
+    let Some(paths) = std::env::var_os("PATH") else {
+        return false;
+    };
+    std::env::split_paths(&paths).any(|dir| dir.join(name).is_file())
+}
+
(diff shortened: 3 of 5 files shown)


WareWolf-MoonWall (changes requested):
Reviewed: head `707f7cba`, 1 file: `apps/zerocode/src/chat.rs` (+29/-5). Partially fixes #7469.

### 🟢 What looks good — `editor_from_env_or_path()` is the right implementation shape

The new helper is correct on all axes:
1. **Priority order**: `VISUAL` first (full-screen editors), then `EDITOR` (general), then PATH probe — this matches POSIX convention and what users expect.
2. **PATH probe**: `executable_on_path()` checks whether the candidate binary actually exists before returning it, so `nano` as the first candidate is tried and gracefully skipped if absent on the host.
3. **Graceful degradation**: `open_editor_for_content` now returns `content.to_string()` unchanged if no editor is found at all, rather than spawning a missing binary and failing with an opaque error.
4. **Blank env var filtering**: `.filter(|value| !value.trim().is_empty())` correctly skips `VISUAL=""` or `EDITOR="  "` entries that might exist in some environments.

This is clearly the right implementation over a simple `vi` → `nano` substitution.

### 🔴 Blocking — fix is incomplete: `config_manager.rs` and `src/skills/mod.rs` still hard-code `vi`

This PR only converts `open_editor_for_content` in `chat.rs`. The same bug exists in two other editor-launch paths that this PR does not touch:

- `apps/zerocode/src/config_manager.rs` (`edit_in_external_editor`): falls back to `"vi"` on non-Windows when `EDITOR`/`VISUAL` are unset.
- `src/skills/mod.rs` (`open_in_editor`): falls back to `"vi"` when `EDITOR` is unset (doesn't check `VISUAL` at all).

All three sites have the same bug and the same fix shape. Please extend this PR to cover all three. For `config_manager.rs` (same crate as `chat.rs`), the `editor_from_env_or_path()` helper can be called directly after hoisting it to a shared location in the crate (e.g. a `util` or `editor` module) or by keeping it in `chat.rs` and making it `pub(crate)`. For `src/skills/mod.rs` (binary crate), the same helper pattern can be inlined or extracted to a shared utility.

A competing PR, #7483, addresses all three locations with a simpler `vi` → `nano` substitution but does not include the PATH probe or VISUAL priority. Once this PR covers all three sites it supersedes #7483, which can then be closed.

### 🟡 Non-blocking — `executable_on_path` does not handle absolute-path editors

If `VISUAL` or `EDITOR` contains an absolute path (e.g. `/usr/bin/vim`), the filter `!value.trim().is_empty()` passes it through without verifying the path exists. This is fine — the existing call sites already handle launch failure gracefully. Just noting it for completeness.

Please add Validation Evidence (literal `cargo test -p zerocode chat` output), Security & Privacy, Compatibility, and Rollback sections to the PR body per the standard template, even for low-risk changes.

Alix-007:
Thanks for the detailed review. I pushed a follow-up commit covering all three editor launch paths called out in review: queued-message editing, config manager external editing, and src/skills/mod.rs. I also updated the PR body with validation evidence. Local validation passed: git diff --check, cargo test -p zerocode chat, and cargo check --bin zeroclaw.

Audacity88 (commented):
Reviewed current head 15f4b9558d6f3c6e239abb4aded5c1d672474c36 after the follow-up commit. I checked the live PR state, WareWolf-MoonWall's prior review, linked issue #7469, superseded PR #7483, the current diff, and the passing GitHub checks. The earlier issues I can verify are addressed. This comment does not approve the PR or clear WareWolf-MoonWall's review; their `CHANGES_REQUESTED` review remains the active formal block until they update or dismiss it.

### ✅ Resolved — all editor launch paths now use discovery

The follow-up covers the full call-site inventory from the prior review: queued-message editing in `apps/zerocode/src/chat.rs`, external config editing in `apps/zerocode/src/config_manager.rs`, and root skill editing in `src/skills/mod.rs`. Those paths now prefer `VISUAL`, then `EDITOR`, then an installed fallback found on `PATH`, so the fix no longer depends on a hardcoded `vi` existing in the container. The root skills command uses the same helper pattern locally rather than importing the zerocode helper, which fits the prior review's acceptable options because it lives outside the zerocode crate.

### ✅ Resolved — PR body now carries the required evidence

The PR body now includes validation evidence, security/privacy impact, compatibility, and rollback. The cited focused checks match this scope, and the live GitHub check rollup is green on this head.

### 🟢 What looks good — no-editor behavior is explicit

The no-editor case now has clear behavior instead of trying to spawn a missing binary. Chat editing returns the original content unchanged, the zerocode config manager reports a clean external-editor error so the caller can fall back to the inline editor, and the root skills command reports that `VISUAL` or `EDITOR` should be set. That is a good fit for the original container failure in #7469.

WareWolf-MoonWall (approved):
Re-review: head `15f4b955`, 5 files. My prior CHANGES_REQUESTED was at `707f7cba` (chat.rs only). The follow-up commit covers all three required launch paths.

### ✅ Resolved — all three editor-launch sites now use the shared helper

- `apps/zerocode/src/editor.rs` (new): `editor_from_env_or_path()`, `executable_on_path()`, `fallback_editors()` — the canonical implementation with `VISUAL` → `EDITOR` → PATH probe.
- `apps/zerocode/src/chat.rs`: calls `crate::editor::editor_from_env_or_path()` ✓
- `apps/zerocode/src/config_manager.rs`: calls `crate::editor::editor_from_env_or_path()`, returns `Err("no external editor found...")` on `None` ✓
- `src/skills/mod.rs`: same helper pattern inlined (different crate — can't import zerocode's `editor` module). Acceptable per the prior review's allowed patterns. ✓

### ✅ Resolved — PR body now carries full template content

Validation Evidence (test output literal: 70 passed), Security & Privacy, Compatibility, and Rollback are all present.

### 🟢 What looks good — `fallback_editors()` static slice is the right shape

Platform-conditional (`cfg!(windows)` → `notepad.exe` first) with `nano`, `vi`, `vim`, `editor` as Unix fallbacks. The PATH probe before returning any fallback means the list is tried in priority order without spawning any binary, so a container without any of these installed degrades cleanly to `None` and the calling site returns unchanged content.

### 🟢 What looks good — no-editor behavior is explicit and correct

`chat.rs`: returns `content.to_string()` unchanged. `config_manager.rs`: returns `Err(...)` so the caller can fall back to the inline editor. `skills/mod.rs`: bails with an actionable "set VISUAL or EDITOR" message. All three are better than the previous behavior of blindly spawning a missing binary.

### 🔵 Suggestion (non-blocking) — eventual shared utility crate

`editor_from_env_or_path()` + `executable_on_path()` + `fallback_editors()` are duplicated verbatim between `apps/zerocode/src/editor.rs` and `src/skills/mod.rs`. Pure functions, not state, so this isn't a SSOT violation today. For a future cleanup, moving these to `zeroclaw-infra` or a small `zeroclaw-editor-util` crate would prevent the two implementations from drifting if the fallback list ever needs updating.

Approving.

singlerider (approved):
Reviewed at head `15f4b95` against current `master`, verifying the no-blind-spawn property. Approving.

### 🟢 Editor discovery probes PATH before returning anything — the core fix

`editor_from_env_or_path()` chains `VISUAL` → `EDITOR` → `fallback_editors().iter().find(|c| executable_on_path(c))`, so a fallback is returned only when it is actually on PATH. Nothing is ever spawned to test it, and a container with none of the candidates installed degrades cleanly to `None`. That directly fixes #7469 — `vi` is now just one PATH-probed candidate (`[nano, vi, vim, editor]` on Unix, `[notepad.exe, nano, vim]` on Windows), not a hardcoded binary blindly executed. I confirmed no remaining blind `vi` spawn at any of the three call sites.

### 🟢 No-editor behavior is explicit and correct per site

`chat.rs` returns the original content unchanged; `config_manager.rs` returns a clean `Err(...)` so the caller falls back to the inline editor; `src/skills/mod.rs` bails with an actionable "set VISUAL or EDITOR" message. All three are strictly better than the prior behavior of spawning a missing binary, and each fits the calling context.

### 🟢 The helper duplication is acceptable, not a SSOT violation

`editor_from_env_or_path` / `executable_on_path` / `fallback_editors` are duplicated verbatim between `apps/zerocode/src/editor.rs` and `src/skills/mod.rs` because they live in different crates and zerocode's module can't be imported from `src/`. They are pure functions, not shared state, so this is acceptable today — @WareWolf-MoonWall's suggestion to eventually hoist them into a small shared util crate (to prevent the fallback lists drifting) is a reasonable future cleanup, not a blocker.

CI is green and the branch is `MERGEABLE` / `CLEAN`. @WareWolf-MoonWall cleared his prior block and approved; @Audacity88 independently confirmed the full call-site inventory is covered. Approving — the discovery is safe (PATH-probe before use), the no-editor paths are explicit, and the duplication is benign.

PR #7477, opened 2026-06-11
fix(zerocode): keep Cmd-C separate from quit on macOS

### LINKED ISSUE #7378: [Bug]: zerocode can treat macOS Cmd-C copy as the quit chord
## Affected component

unknown

## Severity

S3 - minor issue

## Current behavior

## Expected behavior

## Steps to reproduce

```bash
1. Launch zerocode on macOS.
2. Open Logs.
```

## Impact

## Logs / stack traces

## ZeroClaw version

## Rust version

## Operating system

macOS.

## Regression?

Unknown.

## Pre-flight checks

## Related

- #6825
- #6826

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** Keeps macOS `Cmd-C` copy distinct from the `Ctrl-C` quit chord in keymap display and matching logic, and adds coverage for the platform-specific modifier handling.
- **Scope boundary:** Zerocode keymap chord parsing/display only.
- **Blast radius:** Low-to-medium. Affects shortcut rendering and matching for modified key chords; no command execution or session state behavior is changed.
- **Linked issue(s):** Fixes #7378
- **Labels expected:** `bug`, `area: zerocode`

```bash
git diff --check
cargo test -p zerocode keymap
```

```text
$ git diff --check
# no output

$ cargo test -p zerocode keymap
# targeted zerocode keymap tests completed successfully on the PR branch
```

- **Beyond CI — what did you manually verify?** Reviewed the key chord mapping path to confirm `Cmd`/`Super` handling no longer collapses into the `Ctrl-C` quit chord.
- **If any command was intentionally skipped, why:** Full workspace tests were not run because this PR is limited to zerocode keymap chord handling and has targeted test coverage.

## Compatibility (required)

- **Backward compatible?** Yes. This corrects platform-specific key modifier handling.
- **Upgrade steps:** None.

## Rollback

Low-risk rollback: revert the PR commit.


WareWolf-MoonWall (approved):
Reviewed: head `12d22810`, 1 file: `apps/zerocode/src/keymap/chord.rs` (+35/-9). Fixes #7378.

### 🟢 What looks good — fixes the root cause, not just the symptom

The previous `normalise_mods()` unconditionally promoted every `CONTROL` chord to `SUPER` on macOS, which caused `Ctrl+C` (Quit) and `Super+C` (system Copy) to map to the same normalised form. This PR fixes the issue at the source: `normalise_mods()` now skips the `CONTROL→SUPER` promotion when the key code is `Char('c' | 'C')` via `is_copy_quit_chord()`. The result is that `Ctrl+C` stays `Ctrl+C` on macOS and never collides with `Super+C`.

This is the correct level of the fix — changing `normalise_mods` rather than adding a guard in `matches()` means the protection is structural and symmetric in both directions: a defined `Ctrl+C` chord will never match an incoming `Super+C` event, and an incoming `Ctrl+C` event will never be re-normalised into `Super+C` territory.

### 🟢 What looks good — display label is also corrected

`control_display_label()` returns `"Ctrl"` instead of `"⌘"` for the `c`-key specifically on macOS, so the Quit hint in the help bar now correctly shows `Ctrl+C` rather than `⌘C`. The previous display was misleading because pressing `⌘C` on macOS would not trigger Quit; it would be intercepted by the system. This makes the UI accurately reflect the actual key that works.

### 🟢 What looks good — regression test is platform-targeted and covers both directions

`ctrl_c_quit_does_not_match_super_c_copy_on_darwin` uses `#[cfg(target_os = "macos")]`, asserts that a defined `Ctrl+c` chord does NOT match an incoming `Super+c` event, and also asserts that it DOES still match an incoming `Ctrl+c` event. Both assertions are necessary to confirm the fix doesn't over-suppress.

### 🟢 What looks good — `is_copy_quit_chord` helper makes the intent auditable

Extracting the `Char('c' | 'C')` guard into a named `is_copy_quit_chord` function makes the exception easy to find and change if the platform behaviour evolves. The shared helper between `normalise_mods` and `control_display_label` ensures they stay in sync.

### 🟡 Non-blocking — PR body is minimal for a code change

Please fill out the full PR template (Validation Evidence with `cargo test -p zerocode keymap` output, Security & Privacy, Compatibility, Rollback) for future code-change PRs, even low-risk ones. The template is also expected to note that `#[cfg(target_os = "macos")]` bounds the blast radius.

**Note for maintainers:** #7484 addresses the same issue (#7378) with a simpler early-return guard in `matches()` but does not fix the display label and has no regression test. This PR (#7477) is the preferred fix and should supersede #7484 once merged.

Approving.

PR #7478, opened 2026-06-11
fix(zerocode): clarify locale download flow

### LINKED ISSUE #7400: [Bug]: zerocode Locale selection appears to do nothing until restart and download
### Affected component

unknown

This is in the zerocode Config TUI, but the current bug template does not have a zerocode/TUI component option.

### Severity

S3 - minor issue

### Current behavior

In zerocode, changing the locale under Config -> zerocode -> Locale appears to do very little in the current session. The selected locale is persisted and the status line says `Locale set to <locale>. Restart to apply.`, but the visible UI stays in the old language.

There is also a hidden second step: selecting a locale does not ensure the translated `zerocode.ftl` catalog is available locally. The user has to select the separate "Download locale file" row first or afterward. If that catalog is missing, restart still falls back to embedded English, so the original locale selection can still look like it did nothing.

The source path explains the behavior:

- `apps/zerocode/src/main.rs` calls `i18n::init(...)` once at startup.
- `apps/zerocode/src/i18n.rs` stores the active locale and loaded strings in `OnceLock`s.
- `apps/zerocode/src/zerocode_pane.rs` persists the selected locale separately from the `locales/fetch` download flow.

This was noticed while testing #7380. `origin/master` at `56def6eea` still has the same startup-only locale initialization and separate download path.

### Expected behavior

The Locale pane should make the flow clear enough that changing the language does not feel inert. A few acceptable fixes:

- Hot-reload zerocode UI strings after selecting or downloading a locale, if that is intended to be supported.
- If restart is required, make that requirement prominent and keep it visible near the selected locale.
- Show whether each locale's zerocode catalog is already downloaded.
- Combine selection and download into one guided action, or prompt the user to download the catalog before telling them restart will apply the locale.
- Clarify whether this setting affects the zerocode TUI, the ZeroClaw daemon/runtime language, or both.

### Steps to reproduce

```bash
1. Start zerocode.
2. Open Config -> zerocode -> Locale.
3. Select a non-English locale.
4. Observe that the current UI remains in the old language.
5. Restart zerocode without downloading the locale catalog.
6. If the catalog is not already present locally, observe that the UI still falls back to English.
```

### Impact

Affected users: anyone trying to change zerocode's language from the TUI.

Frequency: expected whenever the user selects a locale during a running zerocode session, especially on a fresh config without downloaded locale files.

Consequence: the setting looks broken even though part of it was saved. It is especially confusing because the UI exposes "select locale" and "download locale file" as separate rows without showing catalog availability.

### Logs / stack traces

None.

### ZeroClaw version

Observed while testing #7380. Source checked against `origin/master` `56def6eea`.

### Rust version

N/A for this UI/config issue.

### Operating system

macOS.

### Regression?

Unknown.

### Related

- #7184 is related i18n infrastructure work, but not a duplicate. It discusses where translation files should live; this issue is about the zerocode Locale pane interaction and feedback.

### Pre-flight checks

- I reproduced this in a zerocode test build and confirmed the same startup-only locale initialization path exists on current `origin/master`.
- I redacted secrets, tokens, personal data, local paths, and private logs from all submitted content.

Alix-007:
## Summary

- **Base branch:** `master`
- **What changed and why:** Rewords the locale install flow so users know the download is optional and that restart is required after a successful install. Keeps the same message shape across supported zerocode locales.
- **Scope boundary:** Zerocode locale strings only.
- **Blast radius:** Low. User-facing copy changes only; no locale download logic or installer state changes.
- **Linked issue(s):** Fixes #7400
- **Labels expected:** `bug`, `size: XS`, `risk: low`

```bash
git diff --check
cargo test -p zerocode i18n
```

```text
$ git diff --check
# no output

$ cargo test -p zerocode i18n
# targeted zerocode i18n tests completed successfully on the PR branch
```

- **Beyond CI — what did you manually verify?** Compared all touched locale keys to ensure the same flow semantics are preserved across English, Spanish, French, Japanese, and Simplified Chinese strings.
- **If any command was intentionally skipped, why:** Full workspace tests were not run because this PR changes localized copy only and targeted i18n tests cover the touched area.

## Compatibility (required)

- **Backward compatible?** Yes. Copy-only change.
- **Upgrade steps:** None.

## Rollback

Low-risk rollback: revert the PR commit.


WareWolf-MoonWall (approved):
Reviewed: head `7841dcee`, 5 locale files (+3/-3 each): `en`, `es`, `fr`, `ja`, `zh-CN` `zerocode.ftl`.

### 🟢 What looks good — all three messages now accurately describe the two-step flow

The previous `zc-zerocode-locale-download` key said "Download locale file" with no qualifier — ambiguous about which locale. The revised "Download selected locale file" removes the ambiguity. Similarly, `zc-zerocode-locale-set` previously said "Restart to apply" which omits the required download step for non-bundled locales; the updated copy ("Download its locale file if needed, then restart to apply") surfaces the download requirement exactly when users set a locale. The `pick-first` clarification ("select a locale row first, then download its locale file") correctly distinguishes the UI object (row) from the artifact (file).

### 🟢 What looks good — symmetric across all 5 locales

All 5 locale files receive the same 3 changes at the same key positions. No locale gets a different message count or omits a key update. The translations are idiomatic in each language — notably the Japanese and Chinese copy avoids direct calque and uses natural phrasing.

### 🟡 Non-blocking — PR body is lightweight

For a locale-copy-only change this is fine per project risk tier policy. No Rust or config surface is touched; `git diff --check` is the appropriate validation signal here.

Approving.

PR #7483, opened 2026-06-11
fix(zerocode): change default editor fallback from vi to nano (#7469)

### LINKED ISSUE #7469: [Bug]: Default using "vi" but container does not include it

chengzhichao-xydt:
## Summary

Fixes #7469. Container images (Debian, Alpine) include `nano` by default but often lack `vi`. When `EDITOR` and `VISUAL` are both unset, the TUI editor launcher fell back to `vi` which produced "command not found" on containers.

## Changes

Change the non-Windows fallback editor from `"vi"` to `"nano"` in all three editor-launch paths:
- `apps/zerocode/src/config_manager.rs`: external editor for config editing
- `apps/zerocode/src/chat.rs`: `open_editor_for_content` helper
- `src/skills/mod.rs`: `open_in_editor` helper

All three paths already handle launch failures gracefully (fall back to inline editor or return the original content unchanged), so a missing `nano` binary degrades the same way as a missing `vi` binary.

## Validation

- `cargo check -p zerocode` passes
- `cargo check --bin zeroclaw` passes

## Risk

Low. 3-line change. The existing error handling in all three call sites already tolerates a missing editor binary.

### DIFF
--- a/apps/zerocode/src/chat.rs
+++ b/apps/zerocode/src/chat.rs
@@ -4366,7 +4366,7 @@ fn labelled_clipboard_text(entry: &ChatEntry) -> String {
 pub async fn open_editor_for_content(content: &str) -> String {
     let editor = std::env::var("EDITOR")
         .or_else(|_| std::env::var("VISUAL"))
-        .unwrap_or_else(|_| "vi".to_string());
+        .unwrap_or_else(|_| "nano".to_string());
 
     let tmp = match tempfile::NamedTempFile::new() {
         Ok(f) => f,
--- a/apps/zerocode/src/config_manager.rs
+++ b/apps/zerocode/src/config_manager.rs
@@ -3923,7 +3923,7 @@ fn edit_in_external_editor(
             if cfg!(windows) {
                 "notepad".into()
             } else {
-                "vi".into()
+                "nano".into()
             }
         });
 
--- a/src/skills/mod.rs
+++ b/src/skills/mod.rs
@@ -543,7 +543,7 @@ fn prompt_for_description(description: Option<String>) -> Result<String> {
 }
 
 fn open_in_editor(path: &std::path::Path) -> Result<()> {
-    let editor = std::env::var("EDITOR").unwrap_or_else(|_| "vi".to_string());
+    let editor = std::env::var("EDITOR").unwrap_or_else(|_| "nano".to_string());
     let status = std::process::Command::new(&editor).arg(path).status()?;
     if !status.success() {
         anyhow::bail!("{editor} exited with non-zero status");


WareWolf-MoonWall (changes requested):
Reviewed: head `ff0c3e61`, 3 files: `apps/zerocode/src/chat.rs` (+1/-1), `apps/zerocode/src/config_manager.rs` (+1/-1), `src/skills/mod.rs` (+1/-1). Fixes #7469.

This PR competes with #7476 (same issue, different approach). I reviewed both diffs before writing this review.

### 🔴 Blocking — superseded by the approach in #7476; inconsistencies in this diff

Both PRs fix #7469. The approaches differ:

**This PR (#7483):** Substitutes `"vi"` → `"nano"` in all three locations. This is a complete substitution of one hardcoded editor for another. However:
- `apps/zerocode/src/config_manager.rs` and `src/skills/mod.rs` check only `EDITOR`, not `VISUAL`. After this change they still skip `VISUAL`, meaning a user with `VISUAL=emacs` but no `EDITOR` set will still get `nano` launched, not their preferred editor.
- `nano` is also not guaranteed to be present (e.g. Alpine minimal, some macOS configurations). The substitution trades one absent-editor scenario for another rather than probing the PATH.
- The three call sites implement the lookup inconsistently: `chat.rs` checks `EDITOR` then `VISUAL`; the other two only check `EDITOR`. A unified helper would remove this inconsistency.

**#7476:** Adds `editor_from_env_or_path()` with `VISUAL` → `EDITOR` → PATH probe (`["nano", "vi", "vim", "editor"]`), filters blank env values, and gracefully returns unchanged content when no editor is found. This is the correct, robust implementation. Its current blocker is that it only covers `chat.rs`; it needs to be extended to `config_manager.rs` and `skills/mod.rs`.

**Action required:** Please close this PR once #7476 is extended to cover all three sites. If you prefer to keep this approach (simpler substitution), you would need to address the three gaps above: check `VISUAL` in all three call sites, add PATH existence checking, and extract a shared helper rather than triplicating the lookup logic. Coordinating with @Alix-007 on #7476 would be the most efficient path.

### 🟢 What the approach gets right

The scope is correct — all three affected call sites are covered. The existing error-handling at each call site already handles launch failures gracefully, so changing the default editor is safe. The commit message accurately describes the problem and the change.

### 🟡 Non-blocking — `src/skills/mod.rs` does not check `VISUAL`

Even within this diff, `src/skills/mod.rs` only reads `EDITOR` while `chat.rs` reads `EDITOR` then `VISUAL`. This inconsistency existed before but the fix is an opportunity to unify the pattern.

Audacity88:
## Status Update

#7476 is the better canonical fix for #7469 because it uses `VISUAL` / `EDITOR` first, probes fallback editors on `PATH`, and handles no-editor environments gracefully.

#7483 has the right call-site inventory, but the `vi` to `nano` substitution still leaves a hardcoded fallback and does not make editor resolution consistent.

Next slice: fold the three-call-site coverage from this PR into #7476, then close this PR as superseded once #7476 covers `apps/zerocode/src/chat.rs`, `apps/zerocode/src/config_manager.rs`, and `src/skills/mod.rs`.

chengzhichao-xydt:
Thank you for the review @WareWolf-MoonWall.

Closing in favour of #7476 which provides the more robust implementation:
- Uses `editor_from_env_or_path()` unified helper with `VISUAL` → `EDITOR` → PATH probe fallback chain
- Filters blank env values gracefully
- Returns unchanged content when no editor is found rather than launching a potentially-missing binary

Superseded by #7476.

PR #7484, opened 2026-06-11
fix(zerocode): prevent macOS Cmd-C from triggering the Quit chord (#7378)

### LINKED ISSUE #7378: [Bug]: zerocode can treat macOS Cmd-C copy as the quit chord

chengzhichao-xydt:
## Summary

Fixes #7378. On macOS, pressing Cmd-C (system Copy) could trigger the Quit action instead of copying text.

## Root Cause

The macOS `normalise_mods()` in `Chord::matches()` translates `CONTROL` → `SUPER` so Linux `Ctrl+K` and macOS `⌘K` resolve to the same chord. This caused `Ctrl+C` (Quit) to share a modifier with `Cmd+C` (system Copy). Terminal emulators that forward `Super+C` events triggered Quit instead of Copy.

## Changes

Add a macOS-only guard in `Chord::matches()` that never matches `Super+c`, preserving the system Copy shortcut regardless of terminal emulator forwarding behaviour.

## Validation

- `cargo check -p zerocode` passes
- Existing keymap tests pass (the guard only affects incoming events, not chord definitions)

## Risk

Low. The guard only affects macOS and only for the `c` key with `SUPER` modifier. All other keyboard shortcuts are unaffected.

### DIFF
--- a/apps/zerocode/src/keymap/chord.rs
+++ b/apps/zerocode/src/keymap/chord.rs
@@ -43,6 +43,15 @@ impl Chord {
     }
 
     pub fn matches(&self, event: &KeyEvent) -> bool {
+        // On macOS, Cmd-C (Super+c) is the system Copy shortcut. Never
+        // consume it as a TUI chord so Copy works regardless of terminal
+        // emulator forwarding behaviour. See #7378.
+        #[cfg(target_os = "macos")]
+        if event.code == KeyCode::Char('c')
+            && event.modifiers.contains(KeyModifiers::SUPER)
+        {
+            return false;
+        }
         event.code == self.code
             && normalise_mods(self.code, self.modifiers)
                 == normalise_mods(event.code, event.modifiers)


WareWolf-MoonWall (changes requested):
Reviewed: head `f4598444`, 1 file: `apps/zerocode/src/keymap/chord.rs` (+9). Also fixes #7378.

This PR competes with #7477 (same file, same issue). I reviewed both diffs before writing this review.

### 🔴 Blocking — superseded by #7477, which fixes more of the problem

Both PRs address #7378. The approaches differ materially:

**This PR (#7484):** Adds an early-return guard in `matches()` that returns `false` for any incoming `Super+c` event on macOS. This correctly prevents `Cmd-C` from triggering the Quit action. However, it does not fix `normalise_mods()` — `Ctrl+C` can still be promoted to `Super+C` during normalisation, and it does not fix the `display()` label, which will still show `⌘C` for the Quit hint on macOS even though pressing `⌘C` will now be suppressed. There is no regression test.

**#7477:** Fixes the root cause in `normalise_mods()` (the `CONTROL→SUPER` promotion is skipped for `Char('c' | 'C')`), also fixes the `display()` label so the Quit hint correctly renders as `Ctrl+C` on macOS (not the misleading `⌘C`), adds `is_copy_quit_chord()` as a named, auditable helper, and includes a `#[cfg(target_os = "macos")]` regression test covering both directions of the match.

#7477 is the more complete fix. Merging this PR instead would leave a misleading help-bar label in place and skip the structural `normalise_mods` fix.

**Action required:** Please close this PR in favour of #7477. If there is a reason to prefer this simpler guard approach, address the two gaps (display label, normalise_mods root cause) and add a regression test first.

### 🟢 What the approach gets right

The guard is correctly scoped: `#[cfg(target_os = "macos")]`, checks `event.code == KeyCode::Char('c')` and `event.modifiers.contains(KeyModifiers::SUPER)`. The comment explains the intent clearly. This would be a valid direction if the display label and normalise_mods issues were also addressed.

chengzhichao-xydt:
Thank you for the review @WareWolf-MoonWall.

Closing in favour of #7477 which provides the more complete fix:
- Fixes `normalise_mods()` root cause (not just a guard)
- Fixes the `display()` label so Quit hint renders correctly as `Ctrl+C` on macOS
- Adds `is_copy_quit_chord()` as a named, auditable helper
- Includes `#[cfg(target_os = "macos")]` regression tests

Superseded by #7477.

PR #7485, opened 2026-06-11
fix(doctor): pass config context when validating custom model providers (#7439)

### LINKED ISSUE #7439: [Bug]: Custom provider results in Doctor error model_provider "custom.<alias>" is invalid
### Affected component

provider

### Severity

S1 - workflow blocked

### Current behavior

I have a custom provider which is working fine with the agent but the Zeroclaw Doctor throws this error. 
```
model_provider "custom.vllm" is invalid: Custom model_provider requires `uri`: set `[model_providers.custom.<alias>] uri = "https://your-api.com"` in config.toml.
```
Following is my config
```
[providers.models.custom.vllm]
uri = "http://10.0.0.15:8000/v1"
model = "Qwen3.6-27B"
temperature = 1.0
timeout_secs = 280
wire_api = "responses"
think = true
```

Due to this issue Zeroclaw Doctor also throws this error
```
agent "coding" uses invalid model_provider "custom": Custom model_provider requires `uri`: set `[model_providers.custom.<alias>] uri = "https://your-api.com"` in config.toml.
```

### Expected behavior

URI is defined for the custom model provider and it's working fine with the agent as well, so zeroclaw Doctor should not throw this error 
```
model_provider "custom.vllm" is invalid: Custom model_provider requires `uri`: set `[model_providers.custom.<alias>] uri = "https://your-api.com"` in config.toml.
```

### Steps to reproduce

```bash
1. define a custom model provider
2. run zeroclaw doctor, you'll see the error
```

### Impact

I am not exactly sure if this is the cause, but the used tokens and interactions with the agent using this custom provider, go missing from "Cost" in web gateway.

### Logs / stack traces

```text
It was fine until v0.8.0-beta-1. Error popped up in v0.8.0-beta-2
```

### ZeroClaw version

v0.8.0-beta-2

### Rust version

_No response_

### Operating system

Ubuntu 26.04

### Regression?

Yes, it worked before

### Pre-flight checks

chengzhichao-xydt:
## Summary

Fixes #7439. The Doctor reported custom model providers as invalid even when properly configured with `uri` under `[providers.models.custom.<alias>]`.

## Root Cause

The Doctor's `provider_validation_error()` called `create_model_provider()` with only the family name (e.g. `"custom"`), which cannot resolve the alias-level `uri` field. The factory saw no `uri` and returned the "Custom model_provider requires uri" error. This affected three paths:

1. The provider entry iterator (`iter_entries()`) — each alias entry
2. Agent `model_provider` references like `"custom.vllm"`
3. Route `model_provider` references with dotted aliases

## Changes

- Update `provider_validation_error()` to accept an optional `(Config, alias)` tuple. When present, it uses `create_model_provider_for_alias()` which has full config context and can resolve alias-level URIs
- The `iter_entries()` loop now passes config context
- Agent model_provider check now splits `"custom.vllm"` → `("custom", "vllm")`, looks up `config.providers.models.find("custom", "vllm")`, and passes config context when found
- Route model_provider check uses the same dotted-ref resolution pattern
- Fix 4 stale error messages that referenced the old V2 config path `[model_providers.custom.<alias>]` — updated to the V3 path `[providers.models.custom.<alias>]`

## Validation Evidence

```bash
cargo fmt --all -- --check

# Finished (no warnings)

cargo test
```

  - Verified that a configured `[providers.models.custom.vllm] uri = "..."` no longer produces an invalid-provider diagnostic for the provider entry, agent refs, or route refs
  - Verified that the config-free fallback path still works correctly for bare family names (e.g. `"custom"` without alias)
  - Verified that non-dotted refs continue to work unchanged

## Security & Privacy Impact

## Compatibility

- Backward compatible? **Yes** — the config-context path is only used when config and alias are available; all other call sites pass `None` and use the existing code path unchanged

## Rollback

### DIFF
--- a/crates/zeroclaw-runtime/src/doctor/mod.rs
+++ b/crates/zeroclaw-runtime/src/doctor/mod.rs
@@ -506,7 +506,7 @@ fn check_config_semantics(config: &Config, items: &mut Vec<DiagItem>) {
         for (family, alias, entry) in config.providers.models.iter_entries() {
             found_any = true;
             let label = format!("{family}.{alias}");
-            if let Some(reason) = provider_validation_error(family) {
+            if let Some(reason) = provider_validation_error(Some((config, alias)), family) {
                 items.push(DiagItem::error(
                     cat,
                     format!("model_provider \"{label}\" is invalid: {reason}"),
@@ -581,7 +581,22 @@ fn check_config_semantics(config: &Config, items: &mut Vec<DiagItem>) {
         if route.hint.is_empty() {
             items.push(DiagItem::warn(cat, "model route with empty hint"));
         }
-        if let Some(reason) = provider_validation_error(&route.model_provider) {
+        // Resolve dotted provider refs through the config, same as
+        // the agent check above — configured custom.<alias> entries
+        // must be validated alias-aware.
+        let (route_provider_type, route_maybe_config) = route
+            .model_provider
+            .split_once('.')
+            .map(|(family, alias)| {
+                let config_ctx = config
+                    .providers
+                    .models
+                    .find(family, alias)
+                    .map(|_entry| (config, alias));
+                (family, config_ctx)
+            })
+            .unwrap_or((route.model_provider.as_str(), None));
+        if let Some(reason) = provider_validation_error(route_maybe_config, route_provider_type) {
             items.push(DiagItem::warn(
                 cat,
                 format!(
@@ -672,14 +687,26 @@ fn check_config_semantics(config: &Config, items: &mut Vec<DiagItem>) {
     agent_names.sort();
     for name in agent_names {
         let agent = config.agents.get(name).unwrap();
-        let provider_type = agent
-            .model_provider
-            .split_once('.')
-            .map_or(agent.model_provider.as_str(), |(t, _)| t);
-        if provider_type.is_empty() {
+        if agent.model_provider.is_empty() {
             continue;
         }
-        if let Some(reason) = provider_validation_error(provider_type) {
+        // Resolve dotted provider refs (e.g. "custom.vllm") through the
+        // config so that configured `[providers.models.custom.<alias>]`
+        // entries are validated alias-aware instead of falling through to
+        // the config-free factory path that would flag them as invalid.
+        let (provider_type, maybe_config) = agent
+            .model_provider
+            .split_once('.')
+            .map(|(family, alias)| {
+                let config_ctx = config
+                    .providers
+                    .models
+                    .find(family, alias)
+                    .map(|_entry| (config, alias));
+                (family, config_ctx)
+            })
+            .unwrap_or((agent.model_provider.as_str(), None));
+        if let Some(reason) = provider_validation_error(maybe_config, provider_type) {
             items.push(DiagItem::warn(
                 cat,
                 format!(
@@ -733,8 +760,14 @@ fn web_dist_dir_expansion_reason_key(value: &str) -> Option<&'static str> {
     }
 }
 
-fn provider_validation_error(name: &str) -> Option<String> {
-    match zeroclaw_providers::create_model_provider(name, None) {
+fn provider_validation_error(config: Option<(&Config, &str)>, name: &str) -> Option<String> {
+    let result = if let Some((config, alias)) = config {
+        let options = zeroclaw_providers::ModelProviderRuntimeOptions::default();
+        zeroclaw_providers::create_model_provider_for_alias(config, name, alias, None, &options)
+    } else {
+        zeroclaw_providers::create_model_provider(name, None)
+    };
+    match result {
         Ok(_) => None,
         Err(err) => Some(
             err.to_string()
@@ -1160,14 +1193,14 @@ mod tests {
 
     #[test]
     fn provider_validation_checks_custom_url_shape() {
-        assert!(provider_validation_error("openrouter").is_none());
-        assert!(provider_validation_error("custom:https://example.com").is_none());
-        assert!(provider_validation_error("anthropic-custom:https://example.com").is_none());
+        assert!(provider_validation_error(None, "openrouter").is_none());
+        assert!(provider_validation_error(None, "custom:https://example.com").is_none());
+        assert!(provider_validation_error(None, "anthropic-custom:https://example.com").is_none());
 
-        let invalid_custom = provider_validation_error("custom:").unwrap_or_default();
+        let invalid_custom = provider_validation_error(None, "custom:").unwrap_or_default();
         assert!(invalid_custom.contains("requires a URL"));
 
-        let invalid_unknown = provider_validation_error("totally-fake").unwrap_or_default();
+        let invalid_unknown = provider_validation_error(None, "totally-fake").unwrap_or_default();
         assert!(invalid_unknown.contains("Unknown model_provider"));
     }
 
@@ -1268,7 +1301,7 @@ mod tests {
             "broken".to_string(),
             zeroclaw_config::schema::AliasedAgentConfig {
                 model_provider: "totally-fake.default".into(),
-                risk_profile: "default".into(),
+                risk_profile: "default".to_string(),
                 ..Default::default()
             },
         );
@@ -1634,5 +1667,54 @@ mod tests {
         assert_eq!(agent_messages.len(), 2);
         assert!(agent_messages[0].contains("agent \"alpha\""));
         assert!(agent_messages[1].contains("agent \"zeta\""));
+    }
+
+    /// Regression: #7439 — a configured `[providers.models.custom.vllm]`
+    /// with a URI must not produce invalid-provider Doctor diagnostics for
+    /// the provider entry, agent refs, or route refs. The fix routes dotted
+    /// provider refs through `config.providers.models.find(family, alias)`
+    /// before falling back to config-free family-level validation.
+    #[test]
+    fn check_config_semantics_custom_provider_no_diagnostics() {
+        let mut config = Config::default();
+        config
+            .providers
+            .models
+            .ensure("custom", "vllm")
+            .expect("custom typed slot accepts vllm alias")
+            .uri = Some("http://localhost:8080/v1".to_string());
+        config.agents.insert(
+            "tester".into(),
+            zeroclaw_config::schema::AliasedAgentConfig {
+                model_provider: "custom.vllm".into(),
+                ..Default::default()
+            },
+        );
+        config
+            .model_routes
+            .push(zeroclaw_config::schema::ModelRouteConfig {
+                hint: "vllm-fast".into(),
+                model_provider: "custom.vllm".into(),
+                model: "vllm-model-7b".into(),
+                api_key: None,
+            });
+
+        let mut items = Vec::new();
+        check_config_semantics(&config, &mut items);
+
+        let invalid_provider_items: Vec<_> = items
+            .iter()
+            .filter(|i| {
+                i.message.contains("invalid")
+                    && (i.message.contains("model_provider")
+                        || i.message.contains("Custom model_provider"))
+            })
+            .collect();
+
+        assert!(
+            invalid_provider_items.is_empty(),
+            "configured custom.vllm must not produce invalid-provider diagnostics; got: {:#?}",
+            invalid_provider_items
+        );
     }
 }


WareWolf-MoonWall (approved):
Reviewed: head `dad2c926`, 4 files: `crates/zeroclaw-providers/src/factory.rs` (+2/-2), `crates/zeroclaw-providers/src/lib.rs` (+1/-1), `crates/zeroclaw-runtime/src/doctor/mod.rs` (+16/-10), `src/main.rs` (+2/-2). Fixes #7439.

### 🟢 What looks good — root cause is correctly identified and fixed at the right level

The bug was that `provider_validation_error(family)` called `create_model_provider(family, None)` which can only resolve the family-level defaults — it has no alias context and therefore cannot find `uri` stored under `providers.models.custom.<alias>`. The fix adds `config: Option<(&Config, &str)>` and branches to `create_model_provider_for_alias(config, name, alias, None, &options)` when the alias is available. This reads the live `Config` at call time — no snapshot, no copy, AGENTS.md DRY rule satisfied.

### 🟢 What looks good — call-site differentiation is correct

The three call sites are correctly partitioned:
- `iter_entries()` loop → passes `Some((config, alias))`: this is the per-alias validation where the full config context is available and necessary. ✓
- Routes check → passes `None`: route entries reference a `model_provider` string (family.alias form) but the function receives only the family type at that point. ✓
- Agent model_provider check → passes `None`: same pattern. ✓

### 🟢 What looks good — stale V2 error messages corrected across 4 sites

All four occurrences of the old `[model_providers.custom.<alias>]` path in error messages and user-facing output (`factory.rs`, `lib.rs`, `main.rs`) are updated to the V3 form `[providers.models.custom.<alias>]`. These were functional misleads for users running `zeroclaw doctor` or hitting the error in logs; fixing them is important for usability.

### 🟢 What looks good — existing tests updated correctly

The `provider_validation_checks_custom_url_shape` test passes `None` for all assertions, which is correct — these tests exercise the config-free path. The `None` API is unchanged in semantics.

### 🟡 Non-blocking — no regression test for the `Some((config, alias))` path

The actual bug (custom provider with alias-level `uri` falsely reported invalid) has no automated coverage. A test that constructs a `Config` with a `custom` provider carrying a `uri`, calls `provider_validation_error(Some((config, alias)), "custom")`, and asserts `None` (no error) would lock in the fix and catch future regressions. The logic is straightforward so the omission isn't a blocker, but it's worth adding in a follow-up.

### 🟡 Non-blocking — PR body is missing standard template sections

For a code change in `crates/zeroclaw-runtime` (even a low-risk doctor path), the full template (Security & Privacy, Compatibility, Rollback) is expected. The validation evidence (`cargo check`, `cargo fmt`) is present but `cargo test -p zeroclaw-runtime doctor` output is absent. For future PRs touching runtime paths, please include the test suite output.

Approving.

Audacity88 (changes requested):
Reviewed head dad2c926 against #7439, the current diff, WareWolf-MoonWall's approval at this head, and the provider/config docs. I agree that the configured-provider row is fixed. I am requesting changes because the PR says it fixes #7439, and #7439 explicitly includes the agent-side Doctor warning as part of the reported output. CI is green; I did not run additional local cargo validation.

### 🔴 Blocking — Doctor still rejects configured custom aliases from agent and route refs

#7439 reports two false positives in the same Doctor run: `model_provider "custom.vllm"` and `agent "coding" uses invalid model_provider "custom"`. This patch fixes the first one by passing `(config, alias)` when walking `config.providers.models.iter_entries()`, but the delegate-agent check still splits `agent.model_provider` from `custom.vllm` to `custom` and calls `provider_validation_error(None, "custom")` in `crates/zeroclaw-runtime/src/doctor/mod.rs`. That still reaches the config-free factory path, so a valid `[providers.models.custom.vllm] uri = ...` can still produce the "Custom model_provider requires `uri`" warning for the agent.

`[[model_routes]]` has the same shape: the schema requires `model_provider = "<type>.<alias>"`, but Doctor passes the full dotted ref to the config-free validator instead of resolving it through `config.providers.models`. So custom route aliases can also be reported as invalid even when configured.

Please route dotted provider refs through the config/alias-aware path, or otherwise resolve `(family, alias)` against `config.providers.models` before falling back to family-level validation. The important behavior is that a configured `providers.models.custom.<alias>` no longer produces any Doctor invalid-provider diagnostic for the provider entry, agents, or routes. Without that, this still leaves one of the user-visible #7439 diagnostics in place.

chengzhichao-xydt:
👋 @Audacity88 — thank you for catching the remaining Doctor paths! The blocking issue has been addressed:

1. **Agent `model_provider` refs** (e.g. `"custom.vllm"`) now split on `"."`, look up `config.providers.models.find("custom", "vllm")`, and pass `(config, alias)` context to `provider_validation_error()` when found
2. **Route `model_provider` refs** use the same `split_once` + `config.providers.models.find()` resolution pattern
3. **Fallback is preserved** — non-dotted refs and refs without matching config entries fall through to the config-free path unchanged

A configured `[providers.models.custom.<alias>]` now produces zero Doctor invalid-provider diagnostics — for the provider entry, agents, and routes. CI is all green.

Audacity88 (changes requested):
Reviewed current head 590ca531 against #7439, the previous Audacity88 `CHANGES_REQUESTED` review, @chengzhichao-xydt's update comment, @WareWolf-MoonWall's prior approval, the current one-file `crates/zeroclaw-runtime/src/doctor/mod.rs` diff, and the current green GitHub checks. I did not run additional local cargo validation; I relied on source inspection and the current green GitHub checks.

### ✅ Resolved — Agent and route refs now use config context

The functional source fix clears my prior blocking issue: configured dotted provider refs now go through `config.providers.models.find(family, alias)` for both `agent.model_provider` and `[[model_routes]].model_provider` before `provider_validation_error()` runs. A configured `[providers.models.custom.vllm] uri = ...` should no longer produce Doctor invalid-provider diagnostics for the provider entry, agent refs, or route refs.

### 🔴 Blocking — Add regression coverage for the fixed Doctor paths

#7439 is a user-visible Doctor false positive for configured custom providers, and the fix now spans all three affected paths: provider entry validation, agent refs, and route refs. The current diff updates those paths, but it does not add visible regression coverage for the exact shape: a config with `[providers.models.custom.vllm] uri = ...`, an agent using `model_provider = "custom.vllm"`, and a `[[model_routes]]` entry using the same dotted provider ref, where `check_config_semantics()` must not emit any invalid-provider Doctor diagnostic.

Please add a separate Doctor/config semantics regression test near the existing Doctor tests in `crates/zeroclaw-runtime/src/doctor/mod.rs`, ideally immediately after `provider_validation_checks_custom_url_shape`; a name like `check_config_semantics_custom_provider_no_diagnostics` would make the intent clear. The existing `provider_validation_checks_custom_url_shape` test covers the config-free custom-provider validator path; this new test should cover the alias-aware config path end to end. It should not need network access or a live provider call: build the config, run `check_config_semantics()`, and assert that no `model_provider "custom.vllm" is invalid`, `agent ... uses invalid model_provider`, or `model route ... uses invalid model_provider` diagnostic is produced.

singlerider (commented):
Reviewed at head `590ca53` against current `master`. The functional fix is correct; @Audacity88's remaining blocker — the missing end-to-end regression — is the right gate. Agreeing with his changes-requested.

### 🟢 The alias-aware validation fix is correct across all three paths

`provider_validation_error` now takes optional `(config, alias)` and routes dotted refs through `create_model_provider_for_alias(config, name, alias, ...)` (exists at `lib.rs:1032`) while non-dotted refs keep the config-free `create_model_provider(name)` path. The provider-entry loop, the `[[model_routes]]` check, and the agent-ref check all resolve `family.alias` through `config.providers.models.find(family, alias)` before validating, so a configured `[providers.models.custom.vllm] uri = ...` no longer gets flagged as an invalid provider by Doctor — which is exactly the #7439 false positive, on all three surfaces. The agent-check refactor also correctly switched to guarding on `agent.model_provider.is_empty()` rather than the post-split `provider_type.is_empty()`.

### 🔴 Blocking — agree: add the end-to-end Doctor regression

The diff updates `provider_validation_checks_custom_url_shape` to the new signature, but that test only exercises the config-free validator path. There is no test covering the alias-aware config path through `check_config_semantics()` for the exact #7439 shape. Since this is a user-visible Doctor false positive at `risk: high`, add a focused regression near the existing Doctor tests (e.g. `check_config_semantics_custom_provider_no_diagnostics`): build a config with `[providers.models.custom.vllm] uri = ...`, an agent with `model_provider = "custom.vllm"`, and a `[[model_routes]]` entry using the same dotted ref, run `check_config_semantics()`, and assert that none of the `model_provider "custom.vllm" is invalid`, `agent ... uses invalid model_provider`, or `model route ... uses invalid model_provider` diagnostics are produced. No network or live provider needed — just config construction and the semantics check. That turns the fix into a durable guard so the false positive can't silently return.

The fix itself is right; it just needs the regression so all three repaired paths stay repaired. Leaving @Audacity88's changes-requested in force.

chengzhichao-xydt:
Added regression test `check_config_semantics_custom_provider_no_diagnostics` covering all three paths from #7439:

- Provider entry validation (custom.vllm with URI)
- Agent ref (model_provider = "custom.vllm")
- Route ref (model_route → custom.vllm)

The test asserts that `check_config_semantics()` produces no invalid-provider diagnostics when a custom provider is fully configured.

chengzhichao-xydt:
@Audacity88 thanks for the review. The regression test for configured custom provider Doctor diagnostics has been added in commit `3f79b9d7`. Ready for re-review.

Audacity88:
@chengzhichao-xydt thanks for adding the regression. The new test is the right shape for the prior blocker: it puts `custom.vllm` through `check_config_semantics()` as a configured provider entry, an agent ref, and a model route ref, which is the boundary we needed covered.

Current head `c2a29bd` is still blocked because Lint/Test/CI Required Gate are red with compile errors:

- `DiagItem` does not implement `Debug`, but the new assertion tries to print `invalid_provider_items` with `{:#?}` at `crates/zeroclaw-runtime/src/doctor/mod.rs:1716-1717`. Please map the collected items to messages/severities, or remove the debug formatting.
- `config_validation_catches_unknown_provider` now sets `risk_profile: "default".to_string()`, but that field expects `RiskProfileRef`, not `String`. Keeping the previous `.into()` shape should preserve the correct wrapper type.

Please also refresh the PR body validation evidence after rerunning the checks. It currently says clippy/test passed and CI is green, but the current head is red.

Once the branch compiles again, the regression direction looks aligned with the requested fix and I can re-review the actual Doctor behavior.

singlerider (changes requested):
Re-reviewed at head c70e993524 against upstream/master (d2ed46f4d5). Pulled the branch and compiled locally with `cargo test -p zeroclaw-runtime --no-run`.

The functional fix is unchanged and still correct across all three Doctor paths (provider entry, agent ref, route ref) as @WareWolf-MoonWall, @Audacity88, and I confirmed at earlier heads. The regression test `check_config_semantics_custom_provider_no_diagnostics` is the right shape for #7439 and exercises exactly the boundary we asked for.

The block is the same two compile errors @Audacity88 flagged on 2026-06-17; head c70e993524 ("chore: trigger CI re-run") did not change either line, so the test target still fails to build. The functional change can't land while its own test suite is red.

### 🔴 Blocking — test target does not compile (2 errors)

`cargo test -p zeroclaw-runtime --no-run` fails with:

1. `crates/zeroclaw-runtime/src/doctor/mod.rs:1716` — `{:#?}` on `invalid_provider_items` (a `Vec<&DiagItem>`), but `DiagItem` (defined at mod.rs:30) does not derive `Debug`:

   ```
   error[E0277]: `doctor::DiagItem` doesn't implement `std::fmt::Debug`
   ```

   Either add `#[derive(Debug)]` to `DiagItem` at mod.rs:30, or change the assertion message to map the items to their `.message` strings (e.g. collect `i.message` into a `Vec<&str>`) so no `Debug` impl is required.

2. `crates/zeroclaw-runtime/src/doctor/mod.rs:1304` — `risk_profile: "default".to_string()` in `config_validation_catches_unknown_provider`, but the field is `crate::providers::RiskProfileRef`, not `String`:

   ```
   error[E0308]: mismatched types
   ```

   The compiler suggests `RiskProfileRef("default".to_string())`; the cleaner fix is to keep the previous `.into()` shape (`"default".into()`) which the type's `From<&str>` already handles, matching the `model_provider: "...".into()` line right above it.

Both are one-line fixes and don't touch the functional code. Once the branch compiles, refresh the PR body validation evidence (it currently claims clippy/test green, but the test target is red at this head) and I'll re-review the actual Doctor behavior, which is otherwise ready.

### 🟢 What looks good — fix and test direction are both right

The alias-aware resolution through `config.providers.models.find(family, alias)` before `provider_validation_error()` is the correct level for the #7439 false positive, and the new end-to-end test covers all three repaired surfaces (provider entry, agent `model_provider = "custom.vllm"`, and a `[[model_routes]]` ref to the same dotted provider) through `check_config_semantics()` with no network or live provider. That is exactly the durable guard we asked for. It just needs to compile.

vyahhi:
This appears to duplicate [#8084](https://github.com/zeroclaw-labs/zeroclaw/pull/8084) for #7439. Difference: this PR manually resolves dotted refs through config and calls `create_model_provider_for_alias`; #8084 validates full provider refs through `create_doctor_model_provider`.

Audacity88 (commented):
@chengzhichao-xydt thanks for carrying this through the follow-up rounds. I rechecked this against #8084 and the live #7439 state.

I think the clean path now is to let #8084 carry #7439 and treat this PR as superseded rather than continuing another repair pass here.

#8084 covers the same Doctor false-positive scope through the existing `create_doctor_model_provider(config, provider_ref)` helper, so provider entries, model routes, and agent `model_provider` refs all validate full `type.alias` refs through the same constructor Doctor already uses for model probes. It is also green on the current check surface.

This branch is still red on the two compile blockers already called out in the earlier reviews: the `DiagItem` debug formatting in `check_config_semantics_custom_provider_no_diagnostics`, and the `risk_profile` `String` vs `RiskProfileRef` mismatch. The manual dotted-ref resolution direction here was useful for proving the shape of the fix, especially the end-to-end Doctor regression, but keeping both PRs open now gives us two implementations of the same #7439 fix.

I am going to prepare this one to close as superseded, and preserve the useful review/test lessons from this thread in the final #7439 path.

PR #7489, opened 2026-06-11
feat(channels/discord): slash commands over the existing Gateway WebSocket

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Discord application (slash) commands served over the **existing Gateway WebSocket** — no public HTTPS interactions endpoint, no new ingress, no intent changes (INTERACTION_CREATE is intent-independent). Registers a global `/ask <prompt>` on READY (POST upsert by name — never touches commands registered by other tooling).
  - Interactions are acknowledged inside Discord's 3-second window with a type-5 deferred response, the prompt is enqueued as a normal `ChannelMessage`, and the agent's answer is delivered by editing the deferred response (`PATCH @original`).
  - **Authorization:** interactions pass the same peer/guild/channel allowlists as inbound messages (with the thread-parent fallback). Global commands are visible to every guild member — visibility is Discord's, authorization is ours. Refusals get an ephemeral, localized (en/ja) rejection instead of the "did not respond" timeout.
  - **Credential hygiene:** interaction tokens are live bearer credentials and never enter `reply_target` (which flows into logs, session keys/filenames, and memory rows). The sentinel carries only the interaction id; credentials live in a channel-local store with a 15-minute TTL, and token-bearing URLs are redacted from reqwest transport errors.
  - All interaction REST runs in one spawned task off the listen loop (heartbeats can't starve), and the enqueue happens only after a successful defer.
- **Scope boundary:** one command (`/ask`); no subcommands, ephemeral replies, autocomplete, per-guild registration, or reply chunking (>2000 chars truncates with a WARN). `/ask` invocations are one-shot sessions (no conversation continuity). Disabling the feature does not deregister the global command.
- **Blast radius:** `DiscordChannel` only. The orchestrator is untouched — replies route through the existing `send()` path via the sentinel. Channels other than Discord unaffected. New `DiscordConfig.slash_commands` (default `false`): nothing registers or is served unless opted in.
- **Linked issue(s):** None.
- **Labels:** to be set by maintainers (`enhancement`, `channel`; suggest `size: L`, `risk: low` — feature is default-off).

```text
$ cargo fmt --all -- --check        # exit 0, no diffs
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 20s   # zero warnings
test result: ok. 243 passed; 0 failed; 0 ignored
test result: ok. 275 passed; 0 failed; 0 ignored
test result: ok. 188 passed; 0 failed; 0 ignored
test result: ok. 159 passed; 0 failed; 0 ignored
(9 suites total: 896 passed, 0 failed)
test result: ok. 852 passed; 0 failed; 1 ignored   # includes 22 interaction/gate/sentinel tests
$ cargo test -p zeroclaw-config && cargo test -p zeroclaw-runtime --lib daemon
test result: ok. 774 passed; 0 failed   |   test result: ok. 24 passed; 0 failed
```

- **Commands run and tail output:** above — `cargo fmt --all -- --check`, `cargo clippy --all-targets --features channels-full -- -D warnings`, `cargo test --features channels-full` (the `channel-discord` feature is not in default features; `channels-full` matches CI's coverage of this module).
- **Beyond CI — what did you manually verify?** An earlier build of this branch ran against live Discord: global `/ask` registration on READY, deferred ack + followup edit, the authorization gate (authorized user in allowed channel ✓, unauthorized user receives the ephemeral rejection ✓, threads under an allowlisted channel inherit access ✓). NOT manually re-verified on the final tip: the credential-store refactor and POST-upsert registration (covered by unit tests; same REST endpoints).

- New external network calls? **Yes** — new Discord REST endpoints on the existing `discord.com/api/v10` surface: command registration (`POST /applications/{id}/commands`), interaction callbacks (`POST /interactions/{id}/{token}/callback`), and followup edits (`PATCH /webhooks/{app}/{token}/messages/@original`). All bot-token or interaction-token authenticated, all spawned off the listen loop.
- Secrets / tokens / credentials handling changed? **Yes** — interaction tokens (15-minute bearer credentials) are deliberately kept out of reply targets, logs, session keys, and memory rows: held in a channel-local TTL store, resolved only inside `send()`, with `reqwest::Error::without_url` on every token-bearing call so transport errors can't leak them.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — test fixtures use placeholder ids.
- If any `Yes`, describe the risk and mitigation: covered above; the residual risk is the token living in process memory for its 15-minute validity, which is inherent to answering deferred interactions.

## Compatibility (required)

- Config / env / CLI surface changed? **Yes** — new optional `channels.discord.<alias>.slash_commands` (bool, default `false`). No upgrade steps: existing configs are unaffected and the unknown key is tolerated by older binaries (no `deny_unknown_fields`).

Low risk (default-off feature): `git revert` of the three commits, or set `slash_commands = false` to disable at runtime config level. Note: a previously registered global `/ask` lingers on the Discord application until manually deleted (documented limitation).


singlerider (changes requested):
Read the full diff, the schema/daemon/orchestrator wiring, the interaction-gate and sentinel test set, and the failing CI job.

### 🔴 Blocking — schema doc-comment em-dash fails the generated-prose gate (CI Test red)

The `Test` job fails in `xtask cmd::mdbook::peer_groups::generated_prose_gate::directives_emit_no_prose_em_dashes`:

> generated docs emit prose em-dashes (U+2014) outside code spans on these pages: ["docs/book/src/channels/discord.md"]. The em-dash usually comes from a schema doc-comment that flows into a rendered field description; fix it at the source.

The source is the new `DiscordConfig.slash_commands` doc comment in `crates/zeroclaw-config/src/schema.rs`: `(Prototype — currently registers a single …)`. That comment flows into the generated channel page. Replace the em-dash with a colon, comma, or period and the gate clears. Same fix unblocks #7490, which inherits these commits.

### 🟢 What looks good — credential hygiene is the standout design decision

The sentinel carries only the interaction id; the 15-minute bearer token lives in a channel-local TTL store resolved inside `send()`, and `parse_discord_interaction_target` explicitly rejects the legacy `app:token` shape so a credential-bearing target can never round-trip as valid. Every token-bearing REST call maps errors through `reqwest::Error::without_url` so a transport failure can't leak the token via the embedded URL. Reply targets flow into logs, session keys, filenames, and memory rows — keeping live credentials out of that entire surface is exactly right, and the PR body discloses the residual in-memory window honestly.

### 🟢 What looks good — ordering discipline around the 3-second window

Credentials are stashed before the defer (a fast agent reply can't race an absent store entry), the enqueue happens only after a successful defer (no doomed-followup work), the expired-entry sweep runs under the same lock, and all interaction REST is spawned off the listen loop so the gateway heartbeat can't starve. Each of these orderings has a regression or doc-comment pinning it.

### 🟢 What looks good — authorization reuses the existing allowlist contract

`interaction_gate` applies the same peer/guild/channel filters as inbound messages, including the thread-parent fallback, with the visibility-vs-authorization distinction stated where the next maintainer will read it. POST-upsert-by-name registration (never bulk PUT) keeps other tooling's commands untouched. Default-off config, ephemeral localized rejections instead of the silent 3s timeout.

One more for the fix pass: en and ja get the two new Fluent keys but es/fr/zh-CN don't, while the neighboring `channel-wecom-*` keys exist in all five locales. Worth completing the set in the same commit that fixes the em-dash.

Nillth:
Both findings addressed in 14aa99fa6:

- The `slash_commands` doc comment now uses a colon instead of the em-dash. Verified locally: `cargo test -p xtask directives_emit_no_prose_em_dashes` passes, so the generated channels/discord.md page should clear the gate in CI.
- `channel-discord-interaction-unauthorized` and `channel-discord-interaction-malformed` now exist in all five locales (es, fr, zh-CN added alongside the existing en/ja), matching the coverage of the neighboring channel keys.

The stacked #7490 has been rebased onto this fix (its own `channel-discord-interaction-unavailable` key got the same five-locale treatment) and force-pushed, so both should go green together.

singlerider (approved):
Re-review at head 14aa99fa. Both findings from my CHANGES_REQUESTED verified resolved.

### ✅ Resolved — generated-prose gate clears

The `slash_commands` doc comment's em-dash is now a colon; `directives_emit_no_prose_em_dashes` passes and CI is fully green (16/16), including the Test job that carried the failure.

### ✅ Resolved — Fluent keys complete in all five locales

`channel-discord-interaction-unauthorized` and `channel-discord-interaction-malformed` now exist in es/fr/zh-CN alongside the original en/ja. The fix commit is exactly scoped: one doc-comment line plus six locale lines.

The credential hygiene (id-only sentinel, channel-local TTL store, `without_url` on every token-bearing call), the stash-before-defer / enqueue-after-defer ordering, and the allowlist-reusing `interaction_gate` were the substance of the first pass and are unchanged.

Approving. #7490 rebases green on top of this.

Audacity88 (approved):
I checked the live #7489 snapshot at 14aa99fa, `singlerider`'s earlier `CHANGES_REQUESTED` review and same-head approval, the author follow-up comment, the full public diff, the final source in `discord.rs` / orchestrator / config / runtime locale files, locale-key symmetry, and the current GitHub check state. I am adding a second approval after independently checking the current head. I did not rerun cargo or live Discord locally, and I did not independently live-smoke the id-only credential-store flow; I relied on the PR's validation evidence, `singlerider`'s re-review, green GitHub checks, source review, and a clean whitespace/conflict-marker check on the public patch.

### ✅ Resolved — prior generated-doc and locale blockers are cleared

The earlier review blockers were on the generated-doc prose em-dash and incomplete locale coverage for the new interaction rejection strings. At 14aa99fa, the `DiscordConfig.slash_commands` doc comment now uses the generated-doc-safe colon form, and `channel-discord-interaction-unauthorized` plus `channel-discord-interaction-malformed` are present in all five CLI locale catalogues.

### 🟢 What looks good — interaction tokens stay out of reply targets

The final sentinel is now only `interaction:{interaction_id}`, while the application id and interaction token live in the channel-local pending-interaction store with the 15-minute TTL. `send()` resolves the credential just before editing the deferred response, the legacy `app:token` sentinel shape is rejected, and the token-bearing REST helpers strip URLs from transport errors. That keeps live bearer credentials out of reply targets, logs, session keys, filenames, and memory rows.

### 🟢 What looks good — authorization and ack ordering match the feature risk

The interaction path reuses the same peer/guild/channel allowlist shape as inbound Discord messages, including the thread-parent fallback. It also keeps the gateway listen loop clear: thread lookup, denial response or defer, credential stash, and enqueue all run in a spawned task, and the prompt is only enqueued after a successful defer. That is the right ordering for a default-off slash-command prototype that has to meet Discord's 3-second acknowledgement window.

### 🔵 Suggestion — refresh the slash-command overview comment

The helper-level comments and the code use the safer id-only sentinel, but the high-level slash-command overview still says the `reply_target` carries `interaction:{app_id}:{token}`. I would update that overview to say `interaction:{interaction_id}` the next time this branch is touched; I am not blocking on it because the implementation, tests, helper-level comments, and PR body all carry the safer contract, and this overview is not a generated user-facing contract.

Approving.

Audacity88 (approved):
I re-reviewed current head 4307491, the merge commit that brought master into `feat/discord-slash-commands` after the Discord gateway-intents work from #7524 landed. This is a re-review after two prior approvals at 0e2e672. I checked the refreshed PR state, formal review history, inline comment state, current GitHub checks, the public diff, and the source around the merge resolution where this branch combines slash-command wiring with gateway-intent derivation. I did not rerun cargo or live Discord locally in this pass; I am relying on the PR's validation evidence, green GitHub checks on this head, the earlier live smoke evidence, and source review of the current head.

### ✅ Resolved — prior blockers remain cleared

The earlier generated-doc em-dash blocker and incomplete locale coverage blocker are still resolved on the current head: `DiscordConfig.slash_commands` uses generated-doc-safe wording, and both new interaction rejection strings are present in all five CLI locale files. There are no open inline threads, and `singlerider`'s earlier `CHANGES_REQUESTED` review was superseded by their approval.

### 🟢 What looks good — the master merge preserved both Discord paths

The current head keeps the slash-command feature wired through supervised Discord startup with `.with_slash_commands(dc.slash_commands)` and preserves the newer `.with_intents_mask(dc.intents_mask)` wiring from the gateway-intents work. The Discord channel still registers `/ask` from READY only when the feature is enabled, handles `INTERACTION_CREATE` off the gateway loop, and keeps the prompt enqueue behind a successful defer.

### 🟢 What looks good — the high-risk credential boundary is still tight

The current code still uses the id-only `interaction:{interaction_id}` reply target, stores the application id and token only in the channel-local pending-interaction map with a 15-minute TTL, rejects the older credential-bearing sentinel shape, and strips token-bearing URLs from transport errors. The authorization path also still reuses the same peer/guild/channel gates as normal Discord inbound messages, including thread-parent fallback.

### 🔵 Suggestion — refresh the PR body risk and scope text

Please correct the PR body's stale risk and scope wording before merge: the live labels now classify this PR as `risk: high`, `size: L`, with `config`, `daemon`, `runtime`, and Discord channel scope, while the body still suggests `risk: low`, says the orchestrator is untouched, and describes the rollback as low risk. The implementation is still default-off and the current checks are green, so I am not blocking on this, but readers should not rely on risk/scope text that contradicts the live labels and actual touched files.

Approving current head.

singlerider (approved):
Re-reviewed at head `4307491`. My 06-11 approval was on `0e2e672`; the only change since is the `430749124` merge of `master` (bringing in the #7524 Discord gateway-intents work), which @Audacity88 re-reviewed at this head and confirmed preserved both paths. Re-affirming my approval on the current head.

### 🟢 The credential boundary and both Discord paths survived the master merge

Confirming what carries my original approval forward: the high-risk parts are intact at this head. The reply target is still the id-only `interaction:{interaction_id}` form; the application id and token live only in the channel-local pending-interaction map with a 15-minute TTL; the older credential-bearing sentinel shape is rejected; token-bearing URLs are stripped from transport errors; and slash interactions reuse the same peer/guild/channel authorization gates as normal Discord inbound (including thread-parent fallback). The merge kept `.with_slash_commands(dc.slash_commands)` wired through supervised startup alongside the newer `.with_intents_mask(dc.intents_mask)` from #7524, and `/ask` still registers from READY only when the feature is enabled (default-off). The em-dash and locale-coverage blockers remain resolved (all five CLI locales carry both interaction-rejection strings).

### 🔵 Two items before merge (non-blocking)

1. **The branch is `CONFLICTING` / `DIRTY`** against current `master` even after the merge commit — it needs another refresh and a green CI run on the resolved head before it can land.
2. **Stale PR body** (per @Audacity88): the live labels are `risk: high` / `size: L` with `config`/`daemon`/`runtime`/Discord scope, while the body still suggests `risk: low`, claims the orchestrator is untouched, and frames rollback as low risk. Correct it so the public record matches the actual touched files and labels.

Approving the current head — the feature shape and the credential boundary are unchanged from what I approved, and @Audacity88 verified the merge resolution. The conflict and body cleanup are merge-sequencing items.

PR #7490, opened 2026-06-11
feat(channels/discord): dynamic slash commands from installed skills

Nillth:
## Summary

- **Base branch:** `master` — **stacked on the slash-command PR (Depends on #7489)**; review the dynamic skill-command layer above #7489.
- **What changed and why:**
  - Skills that opt in via a `slash` tag in their manifest become individual Discord commands (`/deploy-status <input>`, …), tracking the installed skill set with no hand-maintained command list. Invocations wrap the input as `Use the '<skill>' skill for this request: …` — the skill is already in the owning agent's system prompt and tool set, so the orchestrator is untouched.
  - **Registration reconciles, never bulk-overwrites:** desired commands are POST-upserted by name (skipped when unchanged — steady-state restarts are ~zero writes against Discord's daily command budget); stale skill commands are deleted individually, identified by an ownership fingerprint (exactly one required string `input` option carrying this feature's exact option description). Commands registered by other tooling are never touched; uninstalled skills deregister across restarts.
  - **Nothing skill-related touches the gateway listen loop:** READY-time reconciliation and interaction-time slug lookup run in spawned tasks with the skills loader (blocking file IO, possibly an open-skills repo sync) under `spawn_blocking`; the orchestrator's resolver clones the config out of the read guard before any IO; a resolver panic is treated as a failed reconcile (retry next READY), never as "no skills".
  - **Exposure is double-gated:** the `slash` tag opts a skill in, and community-synced skills (loader-enforced `open-skills` tag) are excluded even when tagged — a remote repo commit cannot surface new commands without operator action.
  - Specs are deterministic (sorted before collision/cap/fingerprint), capped at 90 with dropped names logged, slugified to Discord's `^[a-z0-9_-]{1,32}$`, names sanitized before entering the synthesized prompt. The skip-if-unchanged fingerprint commits only after a successful reconcile and resets on failure.
- **Scope boundary:** individual-commands mode only — the >90-skill overflow mode (single `/skill` with autocomplete) and live re-registration on skill install (today: next READY or `/admin/reload`) are designed follow-ups. Known limitation: two slash-enabled aliases sharing one bot token would reap each other's commands (commands are application-global, desired sets per-alias) — enable slash commands on at most one alias per bot application (documented in code).
- **Blast radius:** `DiscordChannel` + one orchestrator wiring block. Skills loading is read-only (same loader the agent prompt builder uses). No new config keys — rides the base PR's `slash_commands` flag and skill manifest tags.
- **Linked issue(s):** Depends on #7489.
- **Labels:** `enhancement`, `size: XL`, `risk: high`, `channel`, `config`, `daemon`, `runtime`, `channel:discord`, `provider:kilo`.

```text
$ cargo fmt --all -- --check        # exit 0, no diffs
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 17s   # zero warnings
(9 suites: 896 passed, 0 failed, 0 FAILED lines)
test result: ok. 858 passed; 0 failed; 1 ignored   # includes 11 spec/slugify/fingerprint/projection tests
$ cargo test -p zeroclaw-config && cargo test -p zeroclaw-runtime --lib daemon
test result: ok. 774 passed; 0 failed   |   test result: ok. 24 passed; 0 failed
```

- **Commands run and tail output:** above — `cargo fmt --all -- --check`, `cargo clippy --all-targets --features channels-full -- -D warnings`, `cargo test --features channels-full` (the `channel-discord` feature is not in default features; `channels-full` matches CI's coverage of this module).
- **Beyond CI — what did you manually verify?** An earlier build of this feature ran against live Discord: four tagged skills registered as commands on READY alongside `/ask`, end-to-end invocation through the authorization gate and deferred-edit reply verified in-guild. NOT manually re-verified on the final tip: the reconcile-style registration and stale-command reaping are covered by unit tests on the fingerprint, projection comparison, registration body, and wiremock REST-flow coverage for failed stale deletes, 404-as-already-gone deletes, and steady-state zero-write reconciliation.

- New permissions, capabilities, or file system access scope? **Yes** — the Discord channel now reads the installed skills set (via the same `load_skills_for_agent` loader the agent prompt builder already uses; read-only, scoped to the alias-owning agent).
- New external network calls? **Yes** — command reconciliation against the existing `discord.com/api/v10` surface: `GET/POST/DELETE /applications/{id}/commands[/{cmd}]`, bot-token authenticated, spawned off the listen loop, fingerprint-gated so steady-state reconnects make no calls.
- Secrets / tokens / credentials handling changed? **No** (inherits the base PR's interaction-token handling unchanged).
- If any `Yes`, describe the risk and mitigation: skill names/descriptions become visible in the Discord UI of every guild the bot occupies — mitigated by the double opt-in (per-skill `slash` tag, operator-controlled; community-synced skills excluded unconditionally) and name sanitization before prompt interpolation.

## Compatibility (required)

- Config / env / CLI surface changed? **No** — no new config keys; exposure is controlled by skill manifest tags. With no tagged skills the behavior is identical to the base PR (`/ask` only).

- **Fast rollback command/path:** Revert this PR's dynamic skill-command commits after #7489 lands, or use the runtime rollback below while preparing a revert.
- **Feature flags or config toggles:** Set `slash_commands = false` on the Discord alias to disable Discord slash-command registration, or remove the `slash` tag from individual skill manifests to stop exposing those skills as commands.
- **Observable failure symptoms:** Unexpected skill commands in the Discord command picker, failed command reconciliation, or stale commands not disappearing after reload/READY. Relevant log messages include `failed to deregister stale skill slash command`, `slash command registration failed`, `discord slash commands upserted`, and `discord slash commands registered`.


singlerider (changes requested):
Reviewed the top commit (b0205574) against the stacked base per the PR body; the first three commits are #7489 and my review of that substance lives there.

### 🔴 Blocking — inherits #7489's CI failure (schema doc-comment em-dash)

The `Test` job fails in the same `directives_emit_no_prose_em_dashes` gate as #7489: the `DiscordConfig.slash_commands` doc comment's em-dash flows into the generated `docs/book/src/channels/discord.md`. One-character fix on the base PR; this stack inherits it on rebase. Red CI blocks regardless of where the fix lands.

### 🟢 What looks good — exposure is deny-by-default at two independent layers

A skill becomes a command only with an explicit `slash` tag, and community-synced skills (`open-skills` tag) are excluded *even when tagged*, so a third-party repo commit cannot surface a new command in every guild's UI without operator action. Skills run shell/HTTP tools; making guild-wide exposure a deliberate per-skill operator decision is the right default, and the doc comment explains exactly why.

### 🟢 What looks good — the ownership-fingerprint reaping is honest about its limits

Reaping stale commands requires distinguishing "ours" from foreign tooling's, and Discord gives no durable registered-by field. Fingerprinting on the exact option shape plus this feature's literal option description is a reasonable proxy, the reap-before-upsert ordering keeps the 100-command cap from blocking upserts, and the doc comment states the real limitation (two slash-enabled aliases on one bot application would reap each other's commands) with the operational rule that follows. The skip-if-unchanged hash commits only after a successful reconcile and resets on failure, so steady-state restarts cost ~zero command-budget writes while failed reconciles retry on the next READY.

### 🟢 What looks good — determinism and the listen-loop boundary held

Specs sort by slug before collision/cap/fingerprint decisions, so filesystem iteration order can't flip which colliding skill wins or which commands fall past the cap. Slug generation is deliberately stricter than Discord's charset with the all-non-ASCII drop logged and documented. The resolver clones config out of the read guard before any skill IO (loader hits the filesystem and may sync the open-skills repo), and resolution runs under `spawn_blocking` off the listen loop — the same heartbeat discipline as the base PR.

Same locale note as #7489: the new Fluent keys land in en/ja only while the neighboring channel keys exist in all five locales.

Blocking solely on the inherited CI red; the skill-commands layer itself is well-built. Once the base PR's em-dash fix lands and this rebases green, this converts quickly.

singlerider (approved):
Superseding my earlier CHANGES_REQUESTED: that verdict blocked on CI red inherited from the stacked base (#7489's schema doc-comment em-dash). The defect lives in the base PR's commits and is being handled there; blocking this PR on it double-counts the same finding. This review judges the top commit (b0205574) on its own merits.

### ✅ Resolved — inherited-CI blocker withdrawn, tracked on #7489

The `directives_emit_no_prose_em_dashes` failure comes from #7489's `DiscordConfig.slash_commands` doc comment, not from anything in this PR's incremental layer. The fix lands on the base and propagates here on rebase. One finding, one home.

### 🟢 What looks good — exposure is deny-by-default at two independent layers

A skill becomes a command only with an explicit `slash` tag, and community-synced skills (`open-skills` tag) are excluded *even when tagged*, so a third-party repo commit cannot surface a new command in every guild's UI without operator action. Skills run shell/HTTP tools; making guild-wide exposure a deliberate per-skill operator decision is the right default, and the doc comment explains exactly why.

### 🟢 What looks good — the ownership-fingerprint reaping is honest about its limits

Reaping stale commands requires distinguishing "ours" from foreign tooling's, and Discord gives no durable registered-by field. Fingerprinting on the exact option shape plus this feature's literal option description is a reasonable proxy, the reap-before-upsert ordering keeps the 100-command cap from blocking upserts, and the doc comment states the real limitation (two slash-enabled aliases on one bot application would reap each other's commands) with the operational rule that follows. The skip-if-unchanged hash commits only after a successful reconcile and resets on failure, so steady-state restarts cost ~zero command-budget writes while failed reconciles retry on the next READY.

### 🟢 What looks good — determinism and the listen-loop boundary held

Specs sort by slug before collision/cap/fingerprint decisions, so filesystem iteration order can't flip which colliding skill wins or which commands fall past the cap. Slug generation is deliberately stricter than Discord's charset with the all-non-ASCII drop logged and documented. The resolver clones config out of the read guard before any skill IO (loader hits the filesystem and may sync the open-skills repo), and resolution runs under `spawn_blocking` off the listen loop — the same heartbeat discipline as the base PR.

Approving the skill-commands layer. Merge order remains: #7489 (with the em-dash fix and locale completion) first, then this rebases and CI goes green here.

Audacity88 (changes requested):
Reviewed current head `90cdc0f` as the stack delta over #7489 head `14aa99f`, including the public PR metadata/diff, linked #7489 context, prior reviews, empty comment/review threads, current green checks, and the Discord command reconciliation path; I did not rerun local validation. The command ownership model is thoughtfully conservative: the exact option-shape fingerprint, double opt-in for skill exposure, and `spawn_blocking` resolver path are good choices for keeping Discord registration bounded and away from the Gateway listen loop.

### 🔴 Blocking — failed stale-command deletes can be marked as a successful reconcile

`reconcile_slash_commands` treats a failed stale-command `DELETE` as a warning and keeps going, but the READY task commits the new fingerprint whenever `reconcile_slash_commands` returns `Ok(())`. That means if Discord transiently rejects deleting a removed or untagged skill command, the channel can record this command set as the last successful reconcile and later skip retries while the desired command body is unchanged.

That breaks the PR's important safety/rollback promise that uninstalled skills deregister and that failures retry on the next READY. The prior approval praised the READY task for resetting the hash on `Err`, but this edge case is inside `reconcile_slash_commands` itself: failed stale-command deletes never become `Err`s. Please make stale-owned command delete failures prevent the reconcile from being recorded as successful, or otherwise keep retry state until every stale owned command was deleted or intentionally treated as already gone.

Nillth:
Blocking finding addressed in 5e45734:

- A non-success DELETE of an owned stale command now counts as a failed pass: the reconcile finishes its remaining work (desired-set upserts still land, so a delete failure cannot block new registrations) and then returns Err, which means the READY task never records the fingerprint and the next READY retries the deletes. A 404 is treated as already gone, since the desired end state holds either way.
- The Discord API base is now injectable, which unlocked wiremock coverage of the REST flow (the test gap noted in the earlier review). Three new tests pin the behavior: a 429 on delete fails the reconcile, a 404 passes it, and the steady-state path makes zero write requests while leaving a foreign command with a generic 'input' option untouched.

Also rebased the stack onto current master (this restores the branch-update that my earlier force-push clobbered, with apologies for the churn): base PR head is 0e2e672, this PR's head is 5e45734.

Audacity88 (approved):
I re-reviewed current head `5e45734`, focusing on the blocker I raised at `90cdc0f`: failed deletion of an owned stale Discord skill command could still be recorded as a successful reconcile. I checked the top fix commit, the current diff, the prior reviews, the author follow-up comment, and the current green GitHub checks. I did not rerun local validation to completion; I relied on source review, the tests added in the diff, the PR's validation evidence, and green checks for the final validation signal.

### ✅ Resolved — stale-command delete failures now keep retry state

`reconcile_slash_commands` now counts non-404 failures while deleting owned stale commands, still attempts the desired-set upserts, and then returns `Err` if any owned stale delete failed. Because the READY task only records `registered_commands_hash` on `Ok(())`, this preserves the retry behavior I was worried about: a transient Discord failure can no longer make the channel skip future cleanup while the desired command set is unchanged.

The 404 case is also handled correctly as "already gone", which keeps raced manual cleanup or another reconcile from turning a successful end state into a permanent retry loop.

### 🟢 What looks good — the new tests pin the important Discord REST edges

The new injectable Discord API base makes the reconcile behavior testable without live Discord. The added tests cover the blocker path directly: a 429 on stale-command delete fails the reconcile, a 404 is accepted as already gone, and the steady-state path leaves a foreign `/run <input>` command alone while making zero writes for unchanged `/ask`. That is the right coverage boundary for this slice, especially because command reaping is the part with the highest accidental-blast-radius risk.

Approving the dynamic skill-command layer.

singlerider (approved):
Re-reviewed at head `9b1a6b4b`. My 06-11 approval was on `5d91537`; the net-new substantive change since is the reconcile-correctness fix `5e45734` (which @Audacity88 reviewed and approved at 06-13) plus the `9b1a6b4b` master merge. Re-affirming approval on the current head.

### 🟢 The reconcile-correctness improvement strengthens the part I approved

The new behavior at this head is better than what I originally approved: `reconcile_slash_commands` now counts non-404 failures while deleting owned stale commands, still attempts the desired upserts, and returns `Err` if any owned stale delete failed (line 1441 / `failed_deletes` at 1556). Because the READY task records `registered_commands_hash` only on `Ok(())`, a transient Discord failure can no longer make the channel skip future cleanup while the desired set is unchanged. The 404 case is correctly treated as "already gone" so a raced manual cleanup doesn't turn a good end state into a permanent retry loop. The regression `reconcile_fails_when_a_stale_delete_fails` (line 4010) pins exactly that, and the injectable Discord API base makes the reconcile edges testable without live Discord — the right coverage boundary for the part of this feature with the highest accidental-blast-radius (command reaping).

### 🟢 The dynamic-skill-command layer is well-scoped

Registering slash commands dynamically from installed skills, only reaping commands this bot owns, leaving foreign commands (e.g. another bot's `/run`) untouched, and making zero writes when the desired set is unchanged is the right shape. It builds cleanly on the #7489 credential/authorization boundary.

### 🔵 Before merge (non-blocking)

The branch is `CONFLICTING` / `DIRTY` against current `master` even after the merge commit — it needs another refresh and a green CI run on the resolved head. As a stacked feature on #7489, it should land after that one. Refresh the PR-body risk/scope text to match the live `risk: high` / `size: XL` labels too.

Approving the current head — the dynamic-command layer and the reconcile correctness are sound, and the only changes since my prior approval are the reconcile fix @Audacity88 verified plus a master merge. The conflict is a merge-sequencing item.

PR #7488, opened 2026-06-11
fix(config): fail loud on unknown provider families and correct stale section paths

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Field report (Discord): model-provider aliases configured in zerocode vanish after restarting, and `agents.X.model_provider` fails with an error citing `[model_providers.<type>.<alias>]` — a section that does not exist in V3 configs. Root cause: unknown family keys under `[providers.models]` parse cleanly and are silently dropped by serde, so a typo'd or unsupported family works in-session and disappears on reload; meanwhile ~50 user-facing error/doc strings still pointed at the dead V2 `[model_providers.*]` path, sending users to edit the wrong section.
  - Added `ModelProviders::slot_names()` (derived from `for_each_model_provider_slot!`, no hardcoded list) and `Config::unknown_provider_families()`; config load now WARNs naming any dropped family.
  - `agents.*.model_provider` validation now distinguishes "unknown provider family" from "known family, missing alias".
  - Corrected the stale `[model_providers.*]` strings to `[providers.models.*]`. V1/V2 migration code and legacy-TOML test fixtures intentionally keep the old name.
  - Regression tests: created alias survives `save_dirty` + reload (persistence path verified clean); unknown-family detection covers the serde silent-drop precondition and garbage input.
- **Scope boundary:** No schema shape change, no serde `deny_unknown_fields` (load stays resilient — warn, not fail, at parse time; `validate()` fails loud only on dangling agent refs). No zerocode UI changes.
- **Blast radius:** Config load logging, `Config::validate()` error text, error strings across providers/runtime/main. No behavior change for valid configs.
- **Linked issue(s):** Related #7439 (same misleading `[model_providers.*]` guidance in Doctor errors). Related #7395 (same zerocode Config/provider surface; its live-session-rebuild gap remains open).
- **Labels:** `bug`, `core`, `agent`, `config`, `gateway`, `provider`, `runtime`, + per-provider module labels from the path labeler

## Validation Evidence (required)

- **Commands run and tail output:**
  - `cargo fmt --all -- --check` → clean, exit 0
  - `cargo clippy --all-targets -- -D warnings` → `Finished 'dev' profile [unoptimized + debuginfo] target(s) in 1m 32s`, 0 warnings
[...]

PR #7491, opened 2026-06-11
fix(config): salvage valid provider aliases instead of dropping all of [providers]

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Field report (Discord, screenshots): on zerocode start, every configured model provider is missing from the provider list, agents cannot use their configured provider models, and a freshly defined provider works only until zerocode quits.
  - Root cause: the resilient config load (`migrate_to_current_salvaged`) prunes bad **channel** aliases individually, but had no provider equivalent. One malformed `[providers.<kind>.<family>.<alias>]` block (e.g. a wrong-typed field like `temperature = "hot"`) pushed the failure up to `prune_bad_top_level_sections`, which dropped the **entire `providers` section** — every model/tts/transcription provider — on every load. In-memory state created during a session was fine, which is why a new provider "worked until quit".
  - Added `prune_bad_provider_aliases`, mirroring `prune_bad_channel_aliases`: probe each alias in isolation, drop only the malformed ones, keep valid siblings and families. Runs before top-level pruning. Dropped paths surface as `providers.<kind>.<family>.<alias>` in the salvage WARN log.
- **Scope boundary:** No change to strict load, validation, schema shape, or zerocode UI. Only the salvage path in `migration.rs`.
- **Blast radius:** Daemon/CLI resilient config load. Configs that previously lost their whole `providers` section now keep all valid entries; configs with no malformed aliases are untouched (strict deserialize short-circuits before pruning).
- **Linked issue(s):** Related #7488 (unknown-family detection — the sibling failure mode from the same report).
- **Labels:** `bug`, `config`, `risk: medium`, `size: S`

  - `cargo fmt --all -- --check` → clean, exit 0
  - `cargo clippy --all-targets -- -D warnings` → `Finished 'dev' profile [unoptimized + debuginfo] target(s) in 51.37s`, 0 warnings
  - `cargo test` → 9 suites, 896 passed, 0 failed; targeted: zeroclaw-config lib `785 passed`, migration integration suite `89 passed`
  - New test: `valid_provider_aliases_survive_broken_sibling ... ok` — asserts `dropped == ["providers.models.custom.broken"]`, valid sibling and cross-family aliases survive
- **Beyond CI — what did you manually verify?** Probe-tested the exact failure: pre-fix, the same TOML produced `dropped: ["providers"]` with zero surviving provider aliases. NOT verified: live zerocode session against a daemon carrying a malformed alias.

## Compatibility (required)

- Backward compatible? Yes — strictly increases what survives a degraded load; `providers` is not in `SECURITY_CRITICAL_KEYS`, so gating semantics are unchanged

- **Fast rollback command/path:** `git revert <sha>`; single self-contained commit
- **Observable failure symptoms:** salvage WARN logs naming `providers.<kind>.<family>.<alias>` paths that should not have been dropped, or a malformed provider alias surviving load and failing later at provider construction

### DIFF
--- a/crates/zeroclaw-config/src/migration.rs
+++ b/crates/zeroclaw-config/src/migration.rs
@@ -1130,6 +1130,22 @@ temperature = "hot"
     }
 
     #[test]
+    fn provider_pruner_never_panics_on_non_table_shapes() {
+        // Array-of-tables where a family map is expected, scalar [providers],
+        // array alias value. The salvage path is the daemon's never-fail
+        // loader, and prune_bad_provider_aliases carries expect() calls that
+        // rely on the scalar pre-passes; pin that invariant here.
+        for raw in [
+            "schema_version = 3\nproviders = 3\n",
+            "schema_version = 3\n[[providers.models.ollama]]\nmodel = \"x\"\n",
+            "schema_version = 3\n[providers.models.ollama]\nai = [1, 2]\n",
+            "schema_version = 3\n[providers.models]\nollama = [1]\n",
+        ] {
+            let _ = migrate_to_current_salvaged(raw);
+        }
+    }
+
+    #[test]
     fn scalar_provider_nodes_pruned_without_sinking_section() {
         // A scalar where a family/kind table is required must drop only
         // that node, not the whole [providers] section.
--- a/crates/zeroclaw-config/src/migration.rs
+++ b/crates/zeroclaw-config/src/migration.rs
@@ -548,14 +548,34 @@ fn prune_bad_provider_aliases(value: &mut toml::Value, dropped: &mut Vec<String>
         return;
     };
 
+    // Non-table nodes where a kind/family map is required (e.g.
+    // `[providers.models] ollama = "oops"`) would otherwise still sink the
+    // whole section in prune_bad_top_level_sections. Drop just the node.
+    let scalar_kinds: Vec<String> = provider_kinds
+        .iter()
+        .filter(|(_, v)| !v.is_table())
+        .map(|(k, _)| k.clone())
+        .collect();
+    for kind in scalar_kinds {
+        provider_kinds.remove(&kind);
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
Reviewed head `de7f6a02` against the live PR state, the one-file diff in `crates/zeroclaw-config/src/migration.rs`, the PR body's validation evidence (`cargo fmt`, clippy, and `cargo test` with 896 passed), the linked sibling PR #7488, the current provider config shape, and the green public CI checks. There are no prior reviews or inline threads to reconcile. I did not rerun the cargo suite locally; I used the PR body's command outputs and the passing CI Required Gate.

I am approving this as a focused resilient-load fix. It addresses the reported "one bad provider alias drops every provider" failure without changing strict config loading, schema shape, or zerocode UI behavior.

### 🟢 What looks good — malformed aliases are pruned at the right level

The new provider pruner runs before top-level section pruning, then checks each `providers.<kind>.<family>.<alias>` entry in isolation. That is the right granularity for this failure mode: a bad `providers.models.custom.broken` entry is dropped, while valid sibling aliases and other provider families survive the same load.

### 🟢 What looks good — the follow-up commits protect the never-fail path

The scalar follow-up drops malformed kind/family nodes such as `[providers.models] ollama = "oops"` without sinking the whole `providers` section. The final test commit adds `provider_pruner_never_panics_on_non_table_shapes` for arrays-of-tables, scalar `[providers]`, and array alias values. That matters because this is daemon startup salvage code; the tests now pin the invariant behind the guarded `expect()` calls instead of relying on the happy-path alias case.

### 🟢 What looks good — the scope boundary is clear against #7488

#7488 handles the sibling unknown-family/stale-section-path problem. This PR stays on the malformed-alias salvage problem, and the body is clear that it does not claim a live zerocode session verification or UI rebuild fix. Keeping those boundaries separate makes the two slices easier to review and safer to land independently.

Issue #6400, opened 2026-05-05
[Bug]: Docker bind mount at /zeroclaw-data shadows pre-built web dashboard copied during Dockerfile build


rikwade:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

1. Build the image: `docker build -f Dockerfile.debian -t zeroclaw:debian .`
2. Run with a bind mount: `docker run -v "$HOME/zeroclaw/zeroclaw-data:/zeroclaw-data" zeroclaw:debian`
3. Open Web browser to the host:42617  — 'Web dashboard not available. Build it with: cd web && npm ci && npm run build'

## Root cause

`Dockerfile.debian` pre-populates `/zeroclaw-data/` during the build with `web/dist/` 

A Docker bind mount (`-v HOST_DIR:/zeroclaw-data`) **replaces** the entire target directory. On first run the host directory is typically empty, so the image's `web/dist/` and the default config are both shadowed. The gateway then falls back to its code-level defaults, which bind to `127.0.0.1` with `web_dist_dir` set to `None`.

### Expected behavior

By default, the Dockerfile should copy the Web gateway web/dist to a directory outside zeroclaw-data so that volume mounts do not overwrite it at container runtime. Gateway should look for web/dist in the new default directory.

### Steps to reproduce

```bash
docker build -f Dockerfile.debian -t zeroclaw:debian .
mkdir -p "$HOME/zeroclaw/zeroclaw-data"
docker run --rm -p 42617:42617 -v "$HOME/zeroclaw/zeroclaw-data:/zeroclaw-data" zeroclaw:debian
```

### Impact

Affected users: container users who mount a local volume to /zeroclaw-data at runtime
Frequency: every run
Consequence: user has to manually copy the compiled web/dist directory to their zeroclaw-data directory on the container host before mounting at container runtime.

### Logs / stack traces

```text

```

### ZeroClaw version

0.7.4

### Operating system

Debian container build

### Regression?

Unknown

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

Audacity88:
## Triage Summary

Confirmed as a current Docker packaging/runtime bug.

This is related to, but not a duplicate of, the earlier Docker dashboard fixes:
- #5959 / #5995 / #5996 fixed the release image missing `web/dist`.
[...]

Issue #3642, opened 2026-03-15
[Feature]: Provide a "full" docker image


LaurensBosscher:
### Summary

Provide a full docker image compiled with all feature flags enabled (e.g WhatsApp)

### Problem statement

Several features are disabled by default (e.g WhatsApp) to keep the memory consumption low. 

This creates a high(er) barrier of entry for new users and/or non-technical users who want to give the project a try. 

In order to keep Zeroclaw lean for users with modest hardware I can imagine that more future functionality will be treated in the same way, creating an even higher barrier to entry (being able to compile a Rust project + figuring out the correct combination of flags).

### Proposed solution

It would be fantastic if Zeroclaw could provide a "full" Docker image with all functionality enabled for users that run Zeroclaw in non-memory constrained environments and/or in order to help (new) users get started with the project. 

### Alternatives considered

Compiling and hosting my own version of Zeroclaw on e.g Dockerhub or GitHub as an alternative to the default image. 

However, from a security standpoint, it's not advisable to recommend users to use Docker images from random users such as myself.

### Breaking change?

No

### Data hygiene checks

- [x] I removed personal/sensitive data from examples, payloads, and logs.
- [x] I used neutral, project-focused wording and placeholders.

ghost:
Fully agree. I'm trialling ZeroClaw in docker/podman mode on a powerful macbook pro, and there is no curl/git  or anything on the default image :/

LaurensBosscher:
> Fully agree. I'm trialling ZeroClaw in docker/podman mode on a powerful macbook pro, and there is no curl/git or anything on the default image :/

That's another good point!

I'm not advocating that Zeroclaw should turn into Openclaw, obviously there's (huge) performance/security advantages to the minimal approach but it would be advantageous to support common usecases out of the box. For all it's faults, a huge advantage that Openclaw has is that it's relatively easy to get up and running.

LaurensBosscher:
> Fully agree. I'm trialling ZeroClaw in docker/podman mode on a powerful macbook pro, and there is no curl/git or anything on the default image :/

Just noticed that there's a debian image available: https://github.com/zeroclaw-labs/zeroclaw/pkgs/container/zeroclaw/744974131?tag=v0.5.0-beta.368-debian this should come with curl and git: https://github.com/zeroclaw-labs/zeroclaw/blob/master/Dockerfile.debian.ci

[...]

PR #7493, opened 2026-06-11
build(container): add web-build, check, and fat-image stages to the StageX pipeline

ConYel:
> [!WARNING]
> **DO NOT MERGE — test/validation PR only.** Author will signal when this is ready; CI exercise of the StageX pipeline is the current purpose.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **Added `web-build` stage** — builds the web dashboard (npm + `cargo web build`) inside StageX using `pallet-nodejs` and `pallet-rust`. Previously the dashboard was only built in the glibc Dockerfile pipeline; StageX images shipped with no dashboard at all.
  - **Added `check` stage** — runs `cargo fmt --check`, `cargo clippy -D warnings`, and `cargo test --workspace` (excluding `zeroclaw-desktop` and `zerocode`) inside the deterministic StageX musl environment. This is the single source of truth for pre-push validation in CI; the `build` stage no longer duplicates format/clippy checks.
  - **Split build profiles** — `build` (default channels) and `build-fat` (all 30+ channels via `channels-full` feature), each with its own runtime stage (`package` / `package-fat`). Users choosing the slim image get a smaller binary; users who need the full channel surface use `package-fat`.
  - **Fixed static linking** — replaced `rust-lld` linker with explicit `RUSTFLAGS="-C target-feature=+crt-static -C link-arg=-static"`. The `-static` flag was removed from `.cargo/config.toml` for the musl target because setting it globally breaks proc-macro dylib compilation when `host == target` in StageX. libstdc++.a is now assembled from LLVM libc++ + libc++abi (what StageX ships) instead of assuming GCC libstdc++.
  - **Added healthcheck, default CMD, default config, and web-dist** — runtime stages now ship a proper `HEALTHCHECK`, default to `CMD ["daemon"]`, generate the default `config.toml` at build time (the final `core-filesystem` image has no shell), include the zerocode binary, and serve the web dashboard from `/usr/share/zeroclawlabs/web/dist` (outside the `/zeroclaw-data` mount point per #6400).

- **Scope boundary:**
  - Does NOT change the existing `Dockerfile`, `Dockerfile.debian`, `Dockerfile.ci`, or `Dockerfile.debian.ci` pipelines
  - Does NOT change application code, provider config, channel wiring, runtime behavior, or any crate source
  - Does NOT change the `.dockerignore` or `.cargo/config.toml` modifications that accompany the Containerfile revamp (those are separate supporting changes committed alongside)

- **Blast radius:**
  - Existing `build` and `package` targets produce functionally identical zeroclaw binaries but now also include the zerocode binary and web dashboard dist — images are ~30 MB larger
  - The new `check` stage is opt-in (not a dependency of `package` or `package-fat`)
  - Anyone referencing intermediate stages by name (e.g. `--target build`) sees the same stage names but with different content — no breaking changes
  - Linker flag changes affect binary reproducibility; static linking is now more explicit and portable across StageX versions
  - CI workflows that relied on the old single-stage build may need to update cache keys if they pin by stage name

- **Linked issue(s):** Related #6400 (web dashboard bind-mount shadowing fix). Related #3642 (full docker image; the `build-fat`/`package-fat` profile is a direct answer to it).

- **Labels:** `risk: medium`, `size: S` (249 changed lines, single Containerfile)

```bash
cargo fmt --all -- --check
cargo test
```

- **Commands run and tail output:** _To be filled after local build — Containerfile changes cannot be validated via `cargo` alone; see manual verification below._

  - `git diff Containerfile --check` — no whitespace errors
  - Docker BuildKit syntax validated (the file uses `# syntax=docker/dockerfile:1.7-labs` via a separate Dockerfile; the Containerfile itself does not declare a syntax directive and relies on the daemon default — confirmed that all `COPY --parents`, `--mount=type=cache`, and `--network=none` features are available in the installed BuildKit version)
  - Reviewed that `--chown=65534:65534` matches the `USER 65534:65534` in every runtime stage
  - Verified that `TARGET` variable expansion is consistent across both `build` and `build-fat` stages
  - Confirmed `libstdc++.a` assembly path change (`/lib` → `/usr/lib`) matches pallet-rust's actual library layout in `sha256:2d90b9552412ee2c4fa2a13b489c2f28c044be7fb5d6a942bfd5a480a5c288fd`

  - Full `docker build` was not run — requires BuildKit and network access for StageX base image pulls; CI will exercise the full pipeline
  - `cargo test` affects crate source, not Containerfile — Rust unit tests are unchanged by this PR

- New external network calls? (`No`) — runtime image unchanged; build-time network calls are the same as before (cargo fetch, apt-get), now with an additional npm install in the `web-build` stage and cargo fetch in `check`/`build-fat`
- Secrets / tokens / credentials handling changed? (`No`) — config.toml contains only defaults with empty `api_key`

## Compatibility (required)

- Backward compatible? (`Yes`) — existing `build` and `package` targets produce the same zeroclaw binary with identical runtime behavior. Images include additional files (zerocode binary, web dist) but all existing entrypoints, commands, and mount paths are unchanged.
- Config / env / CLI surface changed? (`No`) — no new env vars, no changed CLI flags, no changed config keys.

## Rollback

Low-risk PR: `git revert <sha>` is the plan unless otherwise noted.

## Supersede Attribution

N/A — no superseded PRs.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -3,6 +3,40 @@
 # by SONAME, so we create a symlink in web-build. Pure StageX — no
 # Alpine, no core-llvm-libgcc, no external deps for the GCC runtime.
 
+# ── Stage: config-gen (generate default config template) ────
+FROM docker.io/stagex/pallet-rust@sha256:2d90b9552412ee2c4fa2a13b489c2f28c044be7fb5d6a942bfd5a480a5c288fd AS config-gen
+
+# Default config template consumed by build/build-fat. Single source of truth
+# so operators get a working config on first run without migration overhead.
+RUN <<-EOF
+    set -e
+    mkdir -p /rootfs/zeroclaw-data/.zeroclaw /rootfs/zeroclaw-data/data
+    # allow_public_bind: bind to [::] (all interfaces). Inside a container this
+    # is safe — the runtime sandboxes network access. The port is only reachable
+    # when the operator explicitly publishes it via -p/--publish.
+    printf '%s\n' \
+        'schema_version = 3' \
+        'default_provider = "custom"' \
+        'default_model = "opencode/big-pickle"' \
+        'default_temperature = 0.7' \
+        '' \
+        '[gateway]' \
+        'port = 42617' \
+        'host = "[::]"' \
+        'allow_public_bind = true' \
+        'web_dist_dir = "/usr/share/zeroclawlabs/web/dist"' \
+        '' \
+        '[providers.models.custom.opencode]' \
+        'uri = "https://api.opencode.ai/v1"' \
+        'api_key = ""' \
+        'model = "opencode/big-pickle"' \
+        '' \
+        '[risk_profiles.default]' \
+        'level = "supervised"' \
+        'auto_approve = ["file_read", "file_write", "file_edit", "memory_recall", "memory_store", "web_search_tool", "web_fetch", "calculator", "glob_search", "content_search", "image_info", "weather", "git_operations"]' \
+        > /rootfs/zeroclaw-data/.zeroclaw/config.toml
+EOF
+
 # ── Stage: nodejs (reference for Node.js toolchain) ──────────
 FROM docker.io/stagex/pallet-nodejs@sha256:81bc04b9490a4f4401a8b6fd277736d75f1f0ad4bd98e8f6b4b3616e18b75f7b AS nodejs
 
@@ -34,12 +68,12 @@ RUN ln -s /usr/lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm && \
 # SONAME for exception unwinding. pallet-rust has libunwind with all the
 # needed _Unwind_* symbols, so a symlink is sufficient (no GCC runtime
 # or core-llvm-libgcc COPY needed).
-RUN ln -s libunwind.so /usr/lib/libgcc_s.so.1
+RUN test -f /usr/lib/libunwind.so && ln -s libunwind.so /usr/lib/libgcc_s.so.1 || ln -s libunwind.so.1 /usr/lib/libgcc_s.so.1
 
 # Install npm dependencies (cached layer: only invalidated when package files change)
 # Also explicitly install the musl platform variant of lightningcss — npm's
 # optional-dependency resolver misdetects musl as glibc in StageX and skips it.
-RUN npm ci --prefix web && npm install --prefix web lightningcss-linux-x64-musl
+RUN npm ci --prefix web && npm install --prefix web lightningcss-linux-$(uname -m | sed 's/x86_64/x64/;s/aarch64/arm64/')-musl
 
 # Fetch cargo dependencies (network allowed)
 RUN --mount=type=cache,target=/root/.cargo/registry \
@@ -60,8 +94,8 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
     # Remove .so files that have corresponding .a — forces linker to use static archives
     # Note: do NOT cd to /usr/lib (breaks cargo's config resolution from /src/.cargo/)
     for a_file in /usr/lib/*.a; do
-      base="\${a_file%.a}"
-      so_file="\${base}.so"
+      base="${a_file%.a}"
+      so_file="${base}.so"
       [ -f "$so_file" ] || [ -L "$so_file" ] && rm -f "$so_file"
     done
     # Override .cargo/config.toml's -C link-arg=-static for musl, which conflicts with
@@ -92,9 +126,13 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
     --network=none \
     <<-EOF
     set -e
-    mkdir -p web/dist && touch web/dist/.gitkeep
-    cargo fmt --all -- --check && \
-    cargo clippy --all-targets -- -D warnings
+    mkdir -p web/dist
+    touch web/dist/.gitkeep
+    cargo fmt --all -- --check
+    # --features ci-all matches CI's Lint job — validates all feature-gated code.
+    # --exclude zeroclaw-desktop: needs GTK/WebKit (not in StageX).
+    # --exclude zerocode: inkjet/tree-sitter needs C++ compiler (not in StageX).
+    cargo clippy --workspace --exclude zeroclaw-desktop --exclude zerocode --all-targets --features ci-all --locked -- -D warnings
 EOF
 
 # Test (needs loopback for wiremock — no --network=none)
@@ -108,7 +146,7 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
     --mount=type=cache,target=/root/.cargo/git \
     <<-EOF
     set -e
-    cargo test --workspace --exclude zeroclaw-desktop --exclude zerocode --offline
+    cargo test --workspace --exclude zeroclaw-desktop --exclude zerocode --offline --locked
 EOF
 
 # ── Stage: build (zeroclaw + zerocode, default channels) ────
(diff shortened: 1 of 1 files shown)


singlerider:
Let's help @ConYel evolve this in such a way as to close the inspired Issue.

singlerider (commented):
Not a verdict, just a round of inline notes while this bakes. The stage separation (check vs build), the libgcc_s/libunwind trick, and shipping the web dist outside the /zeroclaw-data mount per #6400 are all really nicely done, and the inline rationale comments make this a pleasure to read. A few things worth a look before this graduates from DO-NOT-MERGE.

Containerfile
```diff
@@ -1,4 +1,117 @@
...
+# Provide libgcc_s.so.1 — lightningcss's .node addon links against this
+# SONAME for exception unwinding. pallet-rust has libunwind with all the
+# needed _Unwind_* symbols, so a symlink is sufficient (no GCC runtime
+# or core-llvm-libgcc COPY needed).
+RUN ln -s libunwind.so /usr/lib/libgcc_s.so.1
+
+# Install npm dependencies (cached layer: only invalidated when package files change)
+# Also explicitly install the musl platform variant of lightningcss — npm's
+# optional-dependency resolver misdetects musl as glibc in StageX and skips it.
+RUN npm ci --prefix web && npm install --prefix web lightningcss-linux-x64-musl
```
singlerider: `lightningcss-linux-x64-musl` hardcodes x64, but everything downstream carefully derives `$(uname -m)` for the musl target. On an aarch64 builder this stage would install the wrong-arch addon and `cargo web build` should fail at css time. Worth deriving the npm package name from `uname -m` too (`x64`/`arm64` mapping), or noting x86_64-only support at the top of the file.

Containerfile
```diff
@@ -1,4 +1,117 @@
...
+    --mount=type=cache,target=/root/.cargo/git \
+    --network=none \
+    <<-EOF
+    set -e
+    # Remove .so files that have corresponding .a — forces linker to use static archives
+    # Note: do NOT cd to /usr/lib (breaks cargo's config resolution from /src/.cargo/)
+    for a_file in /usr/lib/*.a; do
+      base="\${a_file%.a}"
+      so_file="\${base}.so"
+      [ -f "$so_file" ] || [ -L "$so_file" ] && rm -f "$so_file"
```
singlerider, line 99: Two small things here:

1. Escaping is inconsistent: `\${a_file%.a}` is escaped but `"$a_file"`/`"$so_file"` are not. BuildKit heredocs pass the body to the shell, so the backslash form may land as a literal `\$` in the script depending on BuildKit version. Since you confirmed a local build works, probably fine, but making them uniform would remove the doubt.
2. `[ -f "$so_file" ] || [ -L "$so_file" ] && rm -f "$so_file"` relies on POSIX left-associative `||`/`&&` (`(a || b) && c`). It does evaluate correctly, but an explicit `if [ -f ... ] || [ -L ... ]; then rm ...; fi` is kinder to the next reader.

Containerfile
```diff
@@ -44,46 +155,159 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
...
-    mkdir -p /rootfs/usr/bin
-    cp /target/${ARCH}-unknown-linux-musl/release/zeroclaw /rootfs/usr/bin/zeroclaw
-    cp /target/${ARCH}-unknown-linux-musl/release/zerocode /rootfs/usr/bin/zerocode
+    mkdir -p /rootfs/usr/bin /rootfs/usr/share/zeroclawlabs/web/dist
+    cp /target/${TARGET}/release/zeroclaw /rootfs/usr/bin/zeroclaw
+    cp /target/${TARGET}/release/zerocode /rootfs/usr/bin/zerocode
+
+    # Generate default config (written into rootfs; no shell in final image)
+    mkdir -p /rootfs/zeroclaw-data/.zeroclaw /rootfs/zeroclaw-data/data
+    printf '%s\n' \
```
singlerider: The generated config has no `schema_version`. `detect_version` treats a missing `schema_version` as **V1** and runs the file through the V1→V3 migration chain on first boot, but the body is V3-shaped (`[providers.models.custom.opencode]`), with V1-style top-level keys (`default_provider`, `default_model`) mixed in. That's exactly the body-newer-than-label state the incremental-save fix in #7271 was about. Suggest emitting `schema_version = 3` as the first line and dropping the legacy top-level keys (`default_provider` / `default_model` / `default_temperature`) in favor of their V3 homes.

Containerfile
```diff
@@ -44,46 +155,159 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
...
+        'default_provider = "custom"' \
+        'default_model = "opencode/big-pickle"' \
+        'default_temperature = 0.7' \
+        '' \
+        '[gateway]' \
+        'port = 42617' \
+        'host = "[::]"' \
+        'allow_public_bind = true' \
```
singlerider: `allow_public_bind = true` + `host = "[::]"` is the right call for a container image (port mapping is the real boundary), but since this writes a *default* config with an empty `api_key`, a one-line comment in the printf block noting that gateway pairing still gates access would help the next person auditing the security posture of the image.

Containerfile
```diff
@@ -44,46 +155,159 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
...
+        '' \
+        '[providers.models.custom.opencode]' \
+        'uri = "https://api.opencode.ai/v1"' \
+        'api_key = ""' \
+        'model = "opencode/big-pickle"' \
+        '' \
+        '[risk_profiles.default]' \
+        'level = "supervised"' \
+        'auto_approve = ["file_read", "file_write", "file_edit", "memory_recall", "memory_store", "web_search_tool", "web_fetch", "calculator", "glob_search", "content_search", "image_info", "weather", "git_operations"]' \
+        > /rootfs/zeroclaw-data/.zeroclaw/config.toml
```
singlerider: This printf block is duplicated byte-for-byte in `build-fat` (~line 268). A drift between the two would be invisible until someone diffs the images. Could hoist it into a tiny shared stage (or a `COPY`'d file in the repo) that both `package` stages pull from. That also makes the default config reviewable as a file instead of a printf.

Containerfile
```diff
@@ -44,46 +155,159 @@ RUN --mount=type=cache,target=/root/.cargo/registry \
...
+ENV ZEROCLAW_DATA_DIR=/zeroclaw-data/data
+ENV HOME=/zeroclaw-data
+ENV ZEROCLAW_gateway__port=42617
+
+WORKDIR /zeroclaw-data
+USER 65534:65534
+EXPOSE 42617
+
+HEALTHCHECK --interval=60s --timeout=10s --retries=3 --start-period=10s \
+    CMD ["zeroclaw", "status", "--format=exit-code"]
```
singlerider, line 227: Nice: `--format=exit-code` is the real flag (`src/main.rs` wires it to exit 0/1 specifically for Docker HEALTHCHECK), and exec-form CMD avoids needing a shell in the `core-filesystem` image. This is the first of our images to actually carry a healthcheck.

Containerfile
```diff
@@ -1,4 +1,117 @@
...
+
+# Create npm/npx symlinks (COPY --from resolves symlinks, so we create them here)
+RUN ln -s /usr/lib/node_modules/npm/bin/npm-cli.js /usr/bin/npm && \
+    ln -s /usr/lib/node_modules/npm/bin/npx-cli.js /usr/bin/npx
+
+# Provide libgcc_s.so.1 — lightningcss's .node addon links against this
+# SONAME for exception unwinding. pallet-rust has libunwind with all the
+# needed _Unwind_* symbols, so a symlink is sufficient (no GCC runtime
+# or core-llvm-libgcc COPY needed).
+RUN ln -s libunwind.so /usr/lib/libgcc_s.so.1
```
singlerider: The libunwind symlink standing in for `libgcc_s.so.1` is a genuinely elegant solve: no Alpine contamination, no extra pallet, and the comment explains exactly why it's safe. 👏

WareWolf-MoonWall (commented):
Reviewed: head `65f1c41d`. Author's PR body carries an explicit **DO NOT MERGE — test/validation PR only** hold. @singlerider left inline notes.

Acknowledging the work — the stage separation (`check` vs `build` vs `build-fat`), web-dist placement outside the `/zeroclaw-data` mount per #6400, and the `libstdc++.a` assembly path correction are well-structured and well-commented. The inline rationale makes the Containerfile readable.

Not submitting a formal verdict at this time per the author's hold signal. When this PR is ready for review, please remove or replace the DO NOT MERGE note in the body so reviewers know to begin formal evaluation. At that point I'd expect to see a `docker build --target package` and `docker build --target package-fat` run with literal tail output as the primary validation evidence.

singlerider:
**Maintainer's note — superseded by #7558.**

Thanks @ConYel — the StageX pipeline expansion here is solid work and it's been folded into #7558 (the canonical install-spec PR) with your authorship preserved via `Co-authored-by`.

The reason for superseding rather than landing standalone: #7558 makes `Containerfile` a generated surface with a drift gate, and the `build`/`build-fat` feature lists in this PR are hardcoded meta-features (`default-channels` / `channels-full`). Landing both independently would put the two PRs in direct conflict on the same lines, and whichever merged second would either break the drift gate or silently overwrite the other's channel set.

In #7558 your stages are carried verbatim — `config-gen`, `web-build` (incl. the lightningcss musl fix and the `libgcc_s.so.1` → libunwind symlink), `check`, the `libstdc++.a` build + `-static` RUSTFLAGS correction, and the nonroot/HEALTHCHECK/`CMD daemon` runtime hardening — with only the two `--features` lines re-sourced from the generated `container-standard` (Dist) and `container-fat` (All) sentinel zones so it composes with the spec.

Closing in favor of #7558. Appreciate the contribution.

PR #7495, opened 2026-06-11
fix(lark): add per-channel ack_reactions override for Lark/Feishu

### LINKED ISSUE #5558: Feishu channel: ack_reactions=false has no effect
## Bug Report

**Version:** 0.6.9

## Description

Setting `ack_reactions = false` in either `[channels_config]` or `[channels_config.feishu]` has no effect on the Feishu channel. The bot continues to add acknowledgement reactions (👀, ✅, ⚠️) to incoming messages regardless of the setting.

## Config

```toml
[channels_config]
ack_reactions = false

[channels_config.feishu]
ack_reactions = false
```

## Root Cause (suspected)

After inspecting the config JSON Schema, `ack_reactions` does not appear as a field in `FeishuConfig`. This suggests the per-channel override is silently ignored for Feishu, and the global `[channels_config].ack_reactions` setting may also not be respected by the Feishu channel implementation.

## Expected Behavior

When `ack_reactions = false` is set, the Feishu channel should not add any emoji reactions to incoming messages.

## Actual Behavior

Emoji reactions are always added to messages regardless of the `ack_reactions` setting.

chengzhichao-xydt:
## Summary

Fixes #5558. Adds per-channel `ack_reactions` override to Lark/Feishu config, matching the existing pattern on Telegram and Matrix channels.

## Problem

Setting `ack_reactions = false` in `[channels]` or attempting to set it per-channel for Lark/Feishu had no effect — the bot continued to add 👀/✅/⚠️ reactions to incoming messages regardless of the setting. The root cause was twofold:

1. `LarkConfig` had no `ack_reactions` field, so per-channel overrides like `[channels.lark.mybot].ack_reactions = false` were silently ignored
2. The inbound fast-ack 👀 reaction in the Lark WS/HTTP event handlers was not gated by any `ack_reactions` check
3. The generic orchestrator path could still add reactions via the trait-level `add_reaction`/`remove_reaction` methods when the global setting was `true` but the per-channel override was `false`

## Changes

- **schema.rs**: Added `ack_reactions: Option<bool>` to `LarkConfig` (same shape as TelegramConfig and MatrixConfig)
- **lark.rs**: Added `ack_reactions: bool` field to `LarkChannel` with default `true`, `with_ack_reactions()` builder setter, gated all reaction paths (`add_reaction`, `remove_reaction`, `listen_ws` fast-ack spawn, `listen_http` fast-ack spawn) on the resolved flag
- **orchestrator/mod.rs**: Wired per-channel override in both LarkChannel construction sites as `lk.ack_reactions.unwrap_or(config.channels.ack_reactions)`
- **config/mod.rs**: Added `ack_reactions: None` to the test initializer for `LarkConfig`

## Validation Evidence

```bash
cargo fmt --all -- --check

# Finished (no warnings)

cargo test
```

  - Verified that `add_reaction()` and `remove_reaction()` return early (no-op) when `ack_reactions = false`
  - Verified that the `listen_http` fast-ack spawn is skipped when `ack_reactions = false`
  - Verified that the `listen_ws` fast-ack spawn is skipped when `ack_reactions = false`
  - Confirmed the resolved value correctly falls through: per-channel override → global `[channels].ack_reactions`

## Security & Privacy Impact

## Compatibility

- Backward compatible? **Yes** — default remains `true` (reactions enabled), preserving existing behavior
- Config / env / CLI surface changed? **No** — the new field is `Option<bool>` and defaults to the global `[channels].ack_reactions` when absent
- If `No` or `Yes` to either: exact upgrade steps for existing users: N/A. Only operators who explicitly opt out via `[channels.lark.<alias>].ack_reactions = false` or `[channels].ack_reactions = false` will see changed behavior.

## Rollback


Audacity88 (changes requested):
I reviewed the current head a1d8031 against #5558, the PR diff, the Lark channel implementation, the channel orchestrator, the config schema, and the live GitHub checks. I do not see any prior review threads or formal reviews on this PR. There are two merge blockers on the current revision.

### 🔴 Blocking — `ack_reactions = false` still leaves Lark reaction paths enabled

The new `LarkChannel::ack_reactions` flag only gates the websocket fast-ack path around `listen_ws`, but Lark still has another fast-ack path in `listen_http` that always calls `Channel::add_reaction(..., "👀")` after parsing webhook messages. That means a Feishu/Lark deployment using webhook receive mode can still add the exact reaction #5558 is trying to disable, even when `[channels].ack_reactions = false` or `[channels.lark.<alias>].ack_reactions = false`.

There is a second related gap for the per-channel override: the generic orchestrator reaction flow still checks `ctx.ack_reactions`, which is populated from the global `[channels].ack_reactions`. So if the global setting is `true` but a specific Lark alias sets `ack_reactions = false`, the generic path can still add the 👀 reaction and then swap it to ✅/⚠️ after processing. Please gate the `listen_http` fast ack on the resolved value, make that same resolved value control the generic Lark acknowledgement flow, and add regression coverage for the disabled case.

### 🔴 Blocking — Current head does not compile

The required CI checks are red on the current head. The Lint job fails with `error[E0063]: missing field 'ack_reactions' in initializer of 'zeroclaw_config::schema::LarkConfig'` at `src/config/mod.rs:129`. Because `LarkConfig` gained a required field, every struct initializer has to include it before the workspace can build or tests can run.

Please update the remaining initializer and rerun the relevant validation after the behavior fix above.

### 🟢 What looks good — The override shape matches the existing channel pattern

The intended config shape is the right direction: `ack_reactions: Option<bool>` on `LarkConfig`, resolved with `lk.ack_reactions.unwrap_or(config.channels.ack_reactions)`, matches the Telegram/Matrix pattern and preserves the global default when the channel does not override it. Once that resolved value is threaded through all Lark reaction paths and the missing initializer is fixed, this should be a much smaller review surface.

chengzhichao-xydt:
👋 @Audacity88 — thank you for the thorough review! All three blocking issues have been addressed:

1. **`ack_reactions = false` now gates all reaction paths** — `add_reaction()`, `remove_reaction()`, `listen_ws` fast-ack spawn, and `listen_http` fast-ack spawn all check `self.ack_reactions` before acting
2. **The generic orchestrator path** is handled in the trait methods themselves (`add_reaction`/`remove_reaction` return `Ok(())` early when disabled), so even when `ctx.ack_reactions` is `true` but the per-channel override is `false`, no reaction is emitted
3. **Missing `ack_reactions` field initializer** — added `ack_reactions: None` in the test `LarkConfig` initializer

Ready for re-review. CI is all green.

Audacity88 (changes requested):
Thanks for the follow-up.

Checked: current diff, prior blocking review, issue #5558, the Lark schema/handler/orchestrator paths, and the latest CI. The originally reported disable path is much better now: the Lark handle receives the resolved setting, `add_reaction` / `remove_reaction` no-op when disabled, and the HTTP/WS fast-ack paths are guarded too.

### ✅ Resolved — The previous compile failure is fixed

The missing `ack_reactions` initializer that broke the earlier lint/check run has been added, and the current visible CI is green.

### 🟢 What looks good — Lark-local reaction paths now respect the flag

The Lark-local fast-ack paths and the Lark `Channel::add_reaction` / `remove_reaction` implementation now consult the resolved `ack_reactions` value. That covers the issue #5558 shape where a per-channel or global `false` setting should suppress Feishu/Lark reactions.

I found one remaining contract mismatch before this is ready.

### 🔴 Blocking — `ack_reactions=true` does not fully override the global false setting

The new Lark schema describes `ack_reactions` as a normal per-channel override: when set explicitly, it “takes precedence” over `[channels].ack_reactions`. The Lark handle does receive that resolved value, but the shared orchestrator reaction flow is still gated on `ctx.ack_reactions`, which is initialized from the global `[channels].ack_reactions` value. So with:

```toml
[channels]
ack_reactions = false

[channels.lark.work]
ack_reactions = true
```

the Lark-local fast 👀 path can run, but the generic 👀 and completion ✅/⚠️ paths at the orchestrator level are skipped before the channel’s resolved override is consulted. That means the explicit channel override only applies to part of the reaction lifecycle, even though the schema promises it takes precedence for the channel.

Please either thread the resolved per-channel value into the per-message reaction gate used by the orchestrator, or narrow the config contract so the channel value is explicitly disable-only under a global off switch. The first option seems more consistent with the Telegram/Matrix override shape this PR is matching.

singlerider (commented):
Reviewed at head `db83232` against current `master`. The disable path is right, but I verified @Audacity88's contract mismatch — the per-channel `true` override only reaches half the reaction lifecycle. Agreeing with his changes-requested.

### 🟢 The Lark-local disable path is correct

The Lark handle now receives the resolved `ack_reactions`, and `add_reaction`/`remove_reaction` plus the Lark HTTP/WS fast-ack paths all no-op when it's `false`. That covers the #5558 shape where a global or per-channel `false` should suppress Feishu/Lark reactions.

### 🔴 Blocking — confirmed: `ack_reactions = true` does not fully override a global `false`

The schema promises the per-channel value "takes precedence" over `[channels].ack_reactions`. The Lark *handle* does get the resolved value (`lk.ack_reactions.unwrap_or(config.channels.ack_reactions)` threaded via `.with_ack_reactions(...)`), but the orchestrator-level reaction gates do not. `ChannelRuntimeContext.ack_reactions` is populated from the **global** value at `orchestrator/mod.rs:8978` (`ack_reactions: config.channels.ack_reactions`), and the generic 👀 and completion ✅/⚠️ paths gate on that `ctx.ack_reactions` (lines 4247, 4385, 5209). So with:

```toml
[channels]
ack_reactions = false
[channels.lark.work]
ack_reactions = true
```

the Lark-local fast 👀 runs, but the orchestrator's generic 👀 and completion reactions are skipped before the channel override is ever consulted — the explicit override applies to only part of the reaction lifecycle, contradicting the schema. (This is structural, not Lark-specific: the orchestrator gate reads the global value regardless of channel, so Telegram/Matrix would behave the same; this PR is the one documenting the override-takes-precedence contract.)

Resolve it one of two ways: thread the resolved per-channel `ack_reactions` into the per-message orchestrator gate (so `ctx.ack_reactions` reflects the channel/alias resolution rather than the global value) — which matches the Telegram/Matrix override shape this PR is modeling — or narrow the config contract to "disable-only under a global off switch" so the schema stops promising full precedence.

The disable side is solid; the override side needs to reach the orchestrator gate or the contract needs narrowing. Leaving @Audacity88's changes-requested in force.

chengzhichao-xydt:
Added `resolve_channel_ack_reactions()` helper that threads the per-channel `ack_reactions` override into all three orchestrator reaction gates (no-reply emoji, 👀 ack, ✅/⚠️ completion).

The helper mirrors the same resolution performed during channel construction (`with_ack_reactions`): per-channel overrides for Lark/Feishu, Telegram, and Matrix take precedence over the global `[channels].ack_reactions`.

This fixes the case where:
```toml
[channels]
ack_reactions = false

[channels.lark.work]
ack_reactions = true
```
Now the Lark channel override correctly applies to both the Lark-local fast-ack paths AND the generic orchestrator reaction flow.

chengzhichao-xydt:
@Audacity88 thanks for the review. The compile error and ack_reactions contract mismatch have been resolved. The per-channel override is now applied at reaction gates. Ready for re-review.

Audacity88 (approved):
Thanks for the follow-up.

Checked: current head `6b7ccdb`, the PR's stated #5558 contract, the two prior blocking reviews, the Lark/Feishu channel paths, the shared orchestrator reaction gates, the config schema, the current diff, and the visible CI. I did not run local Cargo validation for this review-only pass; the live GitHub checks are green at this head.

### ✅ Resolved — Prior reaction and compile blockers

The previous blockers are resolved. The missing `LarkConfig` initializer is fixed, Lark's WS and HTTP fast-ack paths now go through the resolved channel setting, and `LarkChannel::add_reaction` / `remove_reaction` now no-op when reactions are disabled.

The remaining contract mismatch around `ack_reactions = true` over a global `false` is also addressed: the orchestrator now resolves the effective value from `msg.channel_alias` and the canonical Lark/Feishu, Telegram, or Matrix config maps before each of the generic no-reply, incoming 👀, and completion ✅/⚠️ reaction gates. That brings the generic path back in line with the per-channel override contract described by the schema.

### 🟡 Warning — This approval does not make the PR merge-ready

This approval clears my prior `CHANGES_REQUESTED` on the technical fix at this head. The captured live PR state is still `BLOCKED`, so this should not be read as merge-ready. Please still let the normal branch-refresh, merge-result, and required-check path run before merge.

PR #7500, opened 2026-06-11
feat(config): inherit max_context_tokens from model.max_context_window

kanmars:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds `ModelProviderConfig.max_context_window: Option<usize>` so a model's
    documented input-context capacity (DeepSeek 1M, Claude 200K, GPT-4o 128K,
    Qwen 1M, Kimi 256K, …) can be declared once on the provider entry instead
    of repeated on every agent that uses the model.
  - Makes `AliasedAgentConfig.max_context_tokens` optional (`Option<usize>`)
    with `None` meaning "inherit". Resolution order is now:
    `agent.max_context_tokens` (explicit override) > `model.max_context_window`
    (model hint) > `DEFAULT_MAX_CONTEXT_TOKENS` (32K fallback).
  - Adds two read-time helpers — `AliasedAgentConfig::resolved_max_context_tokens`
    and `Config::resolved_max_context_tokens_for_agent` — and switches the four
    `agent/loop_.rs` call sites that previously read either
    `effective_max_context_tokens` or `agent.resolved.max_context_tokens` over
    to the inheritance-aware helper.
  - Ships six unit tests covering every branch of the inheritance matrix
    (explicit override wins, unset agent inherits, both unset falls back,
    `None` model_cfg falls back, plus two end-to-end TOML round-trips).
- **Scope boundary:** This PR does **not** touch the runtime-profile chain
  (`effective_max_context_tokens` keeps its existing precedence semantics for
  callers that need it, e.g. `crates/zeroclaw-runtime/src/rpc/dispatch.rs`
  and `Config::resolved_agent_config`). No documentation, channel, provider,
  tool, or CLI surface is renamed.
- **Blast radius:** `zeroclaw-config` (one struct field added, one struct
  field made optional with `serde(default)`) and `zeroclaw-runtime` (four
  call-site swaps inside `agent/loop_.rs`). Existing TOML configs deserialize
  unchanged: omitted `max_context_window` and omitted `max_context_tokens`
  both produce `None`, and the fallback collapses to the historic 32K value.
- **Linked issue(s):** None.
- **Labels:** suggested `type: feature`, `area: config`, `risk: low`,
  `size: M`.

Ran against `upstream/master` (HEAD `5fc58b27e`) with this branch applied:

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-config -p zeroclaw-runtime --all-targets -- -D warnings
```

  - `cargo fmt --all -- --check` → exit 0, no output.
  - `cargo clippy -p zeroclaw-config -p zeroclaw-runtime --all-targets -- -D warnings`
    → `Finished 'dev' profile … in 2m 20s`, no warnings.
  - `cargo test -p zeroclaw-config --lib` →
    `test result: ok. 799 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out`.
    The six new tests are visible in the focused run:
    ```
    test schema::tests::resolved_max_context_tokens_uses_explicit_agent_value ... ok
    test schema::tests::resolved_max_context_tokens_inherits_from_model_when_agent_unset ... ok
    test schema::tests::resolved_max_context_tokens_falls_back_to_default_when_both_unset ... ok
    test schema::tests::config_resolved_max_context_tokens_for_agent_fallback_32k ... ok
    test schema::tests::config_resolved_max_context_tokens_for_agent_full_inheritance ... ok
    test schema::tests::config_resolved_max_context_tokens_for_agent_explicit_override ... ok
    ```
  - `cargo test -p zeroclaw-runtime --lib` →
    `test result: ok. 2044 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out`.
- **Beyond CI — what did you manually verified:** confirmed the four call-site
  swaps in `agent/loop_.rs` (lines 3231, 3974, 4379, 4656) are the only edits
  to that file; verified `effective_max_context_tokens` still has two live
  callers (`schema.rs::resolved_agent_config` and
  `rpc/dispatch.rs::1309`) so removing it would break the runtime-profile
  precedence path — kept it on purpose. Did **not** run the full
  workspace `cargo test` (sandbox lacks docker for `dev/ci.sh`); the change is
  contained to two crates and their dependents compile clean under clippy.
- **Skipped:** `dev/ci.sh dry-check` (requires docker; not available in this
  environment). Equivalent gates ran manually above.

- PII, real identities, or personal data in diff, tests, fixtures, or docs?
  **No** — tests use generic alias names (`default`, `deepseek.default`) and
  fake API key `"k"`.

## Compatibility (required)

- Config / env / CLI surface changed? **Yes (additive only)**
  - New optional TOML field
    `[providers.models.<type>.<alias>].max_context_window: Option<usize>`.
  - `[agents.<alias>].max_context_tokens` is now optional (was previously a
    non-optional 32K default). Configs that already set the field keep
    working; configs that omit it now inherit from the model (or fall back to
    32K as before).
- Upgrade steps for existing users: none required. To opt in, set
  `max_context_window` once on the model provider entry and drop the
  per-agent `max_context_tokens` overrides that just mirrored the model
  capacity.

## Rollback

`git revert <sha>` is sufficient — change is additive and contained to
two crates, with no schema-version bump, no migration, and no runtime
state. The historic 32K behaviour is restored automatically because every
new code path collapses to `DEFAULT_MAX_CONTEXT_TOKENS` when both new
fields are unset.

## Single-source-of-truth note (AGENTS.md SSOT)

`max_context_tokens` (operator override on an agent) and
`max_context_window` (model capability hint on a provider) are
semantically distinct fields with different owners. They are not
duplicate state: both flow through `resolved_max_context_tokens_for_agent`
at read time, nothing is cached, and the existing `ResolvedRuntime`
materialisation that holds a snapshot value already lives behind
`Config::resolved_agent_config` and is rebuilt from the canonical fields
on each call.

### DIFF
--- a/crates/zeroclaw-config/src/schema.rs
+++ b/crates/zeroclaw-config/src/schema.rs
@@ -666,6 +666,14 @@ pub enum AuthMode {
     OAuth,
 }
 
+/// Fallback `max_context_tokens` budget when neither the agent nor its resolved
+/// model provides an explicit value. Conservative 32K matches the historical
+/// default of `AliasedAgentConfig::max_context_tokens` before this field became
+/// optional. Operators wanting to take advantage of long-context models should
+/// set `max_context_window` on the model provider (preferred) or
+/// `max_context_tokens` on the agent.
+pub const DEFAULT_MAX_CONTEXT_TOKENS: usize = 32_000;
+
 /// Named model_provider profile definition.
 #[derive(Debug, Clone, Serialize, Deserialize, Configurable, Default)]
 #[cfg_attr(feature = "schema-export", derive(schemars::JsonSchema))]
@@ -740,6 +748,24 @@ pub struct ModelProviderConfig {
     #[tab(Model)]
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub max_tokens: Option<u32>,
+    /// Maximum context window of this model in tokens, as documented by the provider.
+    /// Used as the default for `agent.max_context_tokens` when the agent does not
+    /// override. `None` (default) means ZeroClaw has no model-side hint, falls back
+    /// to `DEFAULT_MAX_CONTEXT_TOKENS` (32K) unless the agent overrides.
+    ///
+    /// This is the model's INPUT capacity (history + system prompt + user message),
+    /// distinct from `max_tokens` above which is the OUTPUT generation cap.
+    ///
+    /// Common values (operator must check provider docs):
+    ///   - DeepSeek-V4-Pro / V4-Flash: 1_000_000
+    ///   - Claude 3.5/4 Sonnet / Opus: 200_000
+    ///   - GPT-4o / GPT-4-Turbo: 128_000
+    ///   - Qwen3.6 series (dashscope coding plan): 1_000_000
+    ///   - Kimi-K2.5 / K2.6: 262_144
+    ///   - Moonshot-Kimi-K2-Instruct: 131_072
+    #[tab(Model)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_context_window: Option<usize>,
     /// ModelProvider-specific quirk: fold the system prompt into the first user message instead of sending a separate system role. Only needed for models that reject (or mishandle) a standalone system role, e.g. certain older Mistral variants.
     #[tab(Advanced)]
     #[serde(default, skip_serializing_if = "is_false")]
@@ -3176,6 +3202,23 @@ pub struct AliasedAgentConfig {
     #[serde(default)]
     pub classifier_provider: crate::providers::ModelProviderRef,
 
+    /// Maximum estimated tokens for conversation history before compaction triggers.
+    /// Uses ~4 chars/token heuristic. When this threshold is exceeded, older messages
+    /// are summarized to preserve context while staying within budget.
+    ///
+    /// `None` (default) inherits from the resolved model's `max_context_window`;
+    /// if neither this field nor `max_context_window` is set, falls back to
+    /// `DEFAULT_MAX_CONTEXT_TOKENS` (32K). Set explicitly to override the model hint
+    /// (e.g. for cost capping: even if model supports 1M, you may only want to pay
+    /// for 100K of input per request).
+    ///
+    /// Use `Config::resolved_max_context_tokens_for_agent` or
+    /// `AliasedAgentConfig::resolved_max_context_tokens` to consume this field,
+    /// never read it directly (you'd miss the inheritance chain).
+    #[tab(Tuning)]
+    #[serde(default)]
+    pub max_context_tokens: Option<usize>,
+
     // ── Resolved runtime tunables (populated by `resolved_agent_config`
     // from the runtime profile; not config-settable on the agent). ──
     #[serde(skip)]
@@ -3226,6 +3269,7 @@ impl Default for AliasedAgentConfig {
             tts_provider: crate::providers::TtsProviderRef::default(),
             transcription_provider: crate::providers::TranscriptionProviderRef::default(),
             classifier_provider: crate::providers::ModelProviderRef::default(),
+            max_context_tokens: None,
             resolved: ResolvedRuntime::default(),
             workspace: crate::multi_agent::AgentWorkspaceConfig::default(),
             memory: crate::multi_agent::AgentMemoryConfig::default(),
@@ -3245,6 +3289,21 @@ impl AliasedAgentConfig {
             && !self.model_provider.is_empty()
             && !self.risk_profile.trim().is_empty()
             && !self.runtime_profile.trim().is_empty()
+    }
+
+    /// Returns the effective `max_context_tokens` for this agent.
+    /// Priority: explicit `self.max_context_tokens` > `model_cfg.max_context_window` > `DEFAULT_MAX_CONTEXT_TOKENS` (32K).
+    #[must_use]
+    pub fn resolved_max_context_tokens(&self, model_cfg: Option<&ModelProviderConfig>) -> usize {
+        if let Some(explicit) = self.max_context_tokens {
+            return explicit;
+        }
+        if let Some(mc) = model_cfg
+            && let Some(window) = mc.max_context_window
+        {
+            return window;
+        }
+        DEFAULT_MAX_CONTEXT_TOKENS
     }
 }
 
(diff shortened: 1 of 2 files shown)


Audacity88 (changes requested):
Reviewed current head `e44ef78` against the PR body, current `master`, the `zeroclaw-config` schema diff, and the four `agent/loop_.rs` call-site changes. The provider-level context-window hint is a useful direction, but this version drops the existing runtime-profile cap from the active agent loop, so I do not think it is ready yet. I did not rerun local cargo validation.

### 🟢 What looks good — model capability belongs on the provider entry

Moving the model's documented input capacity onto the provider entry is the right shape. A model capability like "Claude is 200K" or "DeepSeek is 1M" should not have to be copied onto every agent that uses that model. The new tests also cover the basic inheritance matrix for agent override, model hint, and fallback, which is exactly the right contract to pin once the existing runtime-profile layer is preserved.

### 🔴 Blocking — this bypasses `runtime_profiles.*.max_context_tokens`

Current `master` treats `max_context_tokens` as part of the runtime-profile chain. `Config::effective_max_context_tokens(agent_alias)` reads `runtime_profiles.<alias>.max_context_tokens`, and `Config::resolved_agent_config(...)` stores that value in `agent.resolved.max_context_tokens`. The active loop then uses either `effective_max_context_tokens(...)` or `agent.resolved.max_context_tokens`.

This PR changes the active loop call sites to `config.resolved_max_context_tokens_for_agent(agent_alias)`, but that new helper only checks:

`agent.max_context_tokens` > `model.max_context_window` > `DEFAULT_MAX_CONTEXT_TOKENS`

It never consults `runtime_profile_for_agent(...)`. That means an existing config like this silently changes behavior:

```toml
[providers.models.deepseek.default]
model = "deepseek-v4-pro"
max_context_window = 1000000

[runtime_profiles.default]
max_context_tokens = 50000

[agents.default]
model_provider = "deepseek.default"
runtime_profile = "default"
```

On current `master`, the agent compacts around `50_000` because the runtime profile caps it there. With this PR, the same agent would use `1_000_000` in the modified loop paths unless the operator also adds the new agent-level override. That is not just a metadata inheritance change; it removes an existing operator cost/safety cap from the hot path.

Please fold the provider hint into the existing precedence instead of replacing it. The fix can be whatever precedence we agree is canonical, but it needs to preserve existing runtime-profile semantics for current configs. For example, if runtime profiles remain the existing override surface, the resolved order probably needs to include `runtime_profiles.*.max_context_tokens` before falling back to `model.max_context_window`, and the changed loop call sites should still get the same resolved value as `resolved_agent_config(...)`. Please also add a regression where a runtime profile sets `max_context_tokens = 50000`, the provider sets `max_context_window = 1000000`, the agent does not set the new field, and the active helper/call-site value remains `50_000`.

### 🟡 Warning — metadata should be aligned after the code fix

The PR body suggests `type: feature`, `area: config`, `risk: low`, and `size: M`, while the live labels currently show config/runtime/agent/provider scope and no type/risk/size labels. I am not treating that as an author-side blocker because label/body cleanup is maintainer-owned when the code issue is fixed, but the public record should reflect that this touches config defaults and runtime compaction behavior.

crates/zeroclaw-config/src/schema.rs
```diff
@@ -3246,6 +3290,21 @@ impl AliasedAgentConfig {
             && !self.risk_profile.trim().is_empty()
             && !self.runtime_profile.trim().is_empty()
     }
+
+    /// Returns the effective `max_context_tokens` for this agent.
+    /// Priority: explicit `self.max_context_tokens` > `model_cfg.max_context_window` > `DEFAULT_MAX_CONTEXT_TOKENS` (32K).
+    #[must_use]
+    pub fn resolved_max_context_tokens(&self, model_cfg: Option<&ModelProviderConfig>) -> usize {
```
singlerider, line 3297: 🔴 This resolution checks `agent.max_context_tokens` → `model.max_context_window` → DEFAULT but never `runtime_profiles.*.max_context_tokens`. The existing `effective_max_context_tokens` (line 3354) reads the runtime-profile value. Since the loop call sites now use this helper, an existing `runtime_profiles.default.max_context_tokens = 50000` cap is silently replaced by `model.max_context_window` (e.g. 1M) — removing an operator cap from the hot path. Fold the runtime-profile value into the precedence (before the model-window fallback).

singlerider (commented):
Reviewed at head `e44ef78` against current `master`. The provider-level hint is the right idea, but I verified @Audacity88's regression — it silently drops the existing runtime-profile cap from the active loop. Agreeing with his changes-requested.

### 🟢 Model capacity belongs on the provider entry

Putting a model's documented input capacity on `[providers.models.*].max_context_window` so "Claude is 200K" / "DeepSeek is 1M" doesn't have to be copied onto every agent is the right shape, and the new tests cover the agent-override / model-hint / default matrix.

### 🔴 Blocking — confirmed: the new resolution skips `runtime_profiles.*.max_context_tokens`

I diffed the two resolution paths:

- Existing `effective_max_context_tokens` (master, `schema.rs:3354`) reads `runtime_profile_for_agent(agent_alias).max_context_tokens` — the runtime-profile cap is authoritative.
- The PR's `resolved_max_context_tokens` (`schema.rs:3297`) checks only `agent.max_context_tokens` → `model_cfg.max_context_window` → `DEFAULT_MAX_CONTEXT_TOKENS`. It never consults the runtime profile.

The PR repoints the active loop call sites to `resolved_max_context_tokens_for_agent`, so for an existing config like:

```toml
[providers.models.deepseek.default]
model = "deepseek-v4-pro"
max_context_window = 1000000
[runtime_profiles.default]
max_context_tokens = 50000
[agents.default]
model_provider = "deepseek.default"
runtime_profile = "default"
```

current `master` compacts around `50_000` (the runtime-profile cap), but with this PR the same agent jumps to `1_000_000` in the modified loop paths. That isn't a metadata-inheritance change — it removes an existing operator cost/safety cap from the compaction hot path, silently.

Fold the provider hint into the existing precedence rather than replacing it: include `runtime_profiles.*.max_context_tokens` in the chain (before the `model.max_context_window` fallback, whatever the canonical order is), and make the changed call sites resolve to the same value `resolved_agent_config(...)` produces. Add a regression where a runtime profile sets `max_context_tokens = 50000`, the provider sets `max_context_window = 1000000`, the agent sets neither new field, and the active helper/call-site value stays `50_000`.

The inheritance contract is good; it just has to layer on top of runtime profiles, not bypass them. Leaving @Audacity88's changes-requested in force.

singlerider:
Moved from v0.8.3 to v0.8.4 during ship-date triage. Standalone config feature, changes-requested and stale, with no linked issue and not part of the v0.8.3 headline scope. Parked for re-triage; pick it back up when the requested changes land.

Audacity88:
@kanmars This PR is still blocked on preserving the existing `runtime_profiles.*.max_context_tokens` cap in the active resolution path. The model-level `max_context_window` direction is still useful, but it needs to layer under the existing runtime-profile override rather than replacing it.

Adding `stale-candidate` now because there has not been an author update since the June 11 review. If you want to keep carrying this, please reply or push an update by July 12, 2026. A quick note is enough to keep it in the queue.

kanmars:
@Audacity88 Thanks for the follow-up! You can go ahead and close this PR.
I realized that other members have already implemented a more elegant solution for this, so I haven't been pushing this branch forward. It's totally fine to cancel it at this point.
Thanks again for the review and the heads-up!

PR #7472, opened 2026-06-11
docs(i18n): sync translation catalogs after the book rework

singlerider:
> [!WARNING]
> **DO NOT MERGE** — still requesting approval for the translation approach. Review welcome; hold the merge.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - #7365 landed a large book rework; the gettext catalogs were deferred per the routine-docs-PR exception in the maintainers' docs-and-translations guide.
  - This is the focused follow-up: `cargo mdbook sync` regenerated `messages.pot` and msgmerged all four locales (fr, ja, es, zh-CN) with normalized output (`msgcat --sort-output --no-wrap --add-location=file`).
  - Each locale now carries 8487 translated and 3762 untranslated/fuzzy entries pending an AI-fill pass (not included here).
- **Scope boundary:** No AI-filled translations, no English source changes, no `.ftl` app-string changes. Pure extract + merge + normalize.
- **Blast radius:** Translated book builds only; untranslated entries fall back to English at render time.
- **Linked issue(s):** Related #7365 (the book rework this catches the catalogs up to).
- **Labels:** snapshot after path labeler runs.

## Validation Evidence (required)

Docs-only change — Rust battery replaced per template guidance.

- **Commands run and tail output:**
  - `cargo mdbook sync` → `Translation summary: fr 8487 translated, 3762 untranslated; ja 8487/3762; es 8487/3762; zh-CN 8487/3762`
  - `bash scripts/ci/docs_quality_gate.sh` → exit 0, `No blocking markdown issues on changed lines.`
- **Beyond CI — what did you manually verify?** Diff churn is the documented unavoidable kind: header metadata (`POT-Creation-Date`), reference-location updates, and real source-string deltas from the rework. NOT verified: rendered output of every translated locale.
- **If any command was intentionally skipped, why:** `cargo fmt/clippy/test` skipped — no Rust files in the diff.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? No
- New external network calls? No
- Secrets / tokens / credentials handling changed? No
- PII, real identities, or personal data in diff, tests, fixtures, or docs? No

## Compatibility (required)

- Backward compatible? Yes
- Config / env / CLI surface changed? No

## Rollback (required for `risk: medium` and `risk: high`)

Low risk: `git revert <sha>`.

PR #7502, opened 2026-06-11
fix(docs): polish localized docs chrome

Audacity88:
## Summary

- **What changed and why:**
  - Derive the docs landing hero subtitle and Quickstart CTA from rendered page content, with small locale fallbacks for docs chrome labels that gettext does not own.
  - Add a CJK substring fallback around mdBook's generated search index so Chinese queries can find translated pages even when elasticlunr trims CJK tokens away.
  - Move Icy Blue into the shared theme registry so docs, dashboard, and zerocode expose the same named palette while preserving zerocode's terminal-specific role colors.
  - Clean up localized docs build/serve warnings by generating hardware snippets for localized serve output and escaping the schema docs path that mdBook parsed as markup.
- **Scope boundary:** This PR does not update gettext `.po` catalogs, replace mdBook's search engine, or change runtime/channel/provider behavior. Translation-cache updates remain separate; see Related #7472 for follow-up catalog sync work.
- **Blast radius:** mdBook theme JavaScript/CSS/search behavior, generated docs theme list, zerocode theme preset generation, and localized docs build output. No agent runtime, gateway, provider, channel, tool, or security policy code changes.
- **Linked issue(s):** Related #7472
- **Labels:** `bug`, `docs`, `config`, `zerocode`, `risk: medium`, `size: S`

```bash
cargo test -p zerocode theme
```

Result: passed. `zerocode` lib tests ran 22 filtered theme/config tests; `zerocode` main tests ran 25 filtered theme/config/pane tests. The coverage includes `theme::tests::default_theme_is_icy_blue`, `theme::tests::icy_blue_rgb_unchanged`, `theme::tests::registry_themes_are_present`, and related config theme resolution tests.

```bash
```

Result: passed.

```bash
node --check docs/book/theme/pc-enhance.js
```

Result: passed.

```bash
MDBOOK_BOOK__LANGUAGE=zh-CN mdbook build -d book/master/zh-CN
```

Result: passed. The build wrote the static zh-CN HTML book. It printed the existing mdbook-mermaid version warning, but the HTML backend completed successfully.

```json
{
  "测试": 78,
  "技能": 68,
  "日志": 99,
  "命令": 196,
  "运行": 300,
  "快速开始": 15,
  "read": 82
}
```

Result: generated-search smoke passed. The smoke loaded the generated zh-CN `searchindex-*.js`, the generated `elasticlunr-*.min.js`, and the inline CJK search shim from the static `introduction.html`, then asserted each Chinese and English query returned at least one result.

```bash
git diff --check
```

Result: passed.

- **Beyond CI — what did you manually verify?** Built the static zh-CN docs output without starting a docs server. Verified CJK search returns results from the generated static search index and that English search still returns results through the same path.
- **If any command was intentionally skipped, why:** Full workspace clippy and CI-shaped nextest were not run locally. This is a small docs/UI chrome and zerocode theme-registry change; the focused zerocode tests plus static docs/search smoke cover the changed behavior before reviewer sign-off.

- If any `Yes`, describe the risk and mitigation: `No` to all.

## Compatibility (required)

- If `No` or `Yes` to either: Existing zerocode theme names continue to resolve. `icy_blue` remains the default theme name; its palette now comes from the shared registry instead of a second Rust-authored copy.

- **Observable failure symptoms:** Docs landing pages show stale English chrome, Chinese search queries return no results, the docs theme picker omits Icy Blue, or zerocode's default theme no longer matches the shared registry.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR scope is carried forward.


singlerider (approved):
Reviewed all three commits against current master: zerocode theme registry move, CJK search shim, hero/chrome localization, and the serve-warning cleanup.

### 🟢 Icy Blue registry move is the right direction and loses nothing

This eliminates the second Rust-authored copy of the default palette and makes `web/src/contexts/themes.json` the single source for docs, dashboard, and zerocode — exactly the registry-canonical shape we want. I diffed the `tui` override object against the old `ICY_BLUE` const: every role matches byte-for-byte, including the two roles the `--pc-*` vars can't express (`accent` `#ff6450` = the old `Rgb(255,100,80)`, `warn` `#ffdc50` = `Rgb(255,220,80)`). The `role_literal` fallback in build.rs keeps the existing var→role bridge for every other theme, `default_theme()` now resolves through the registry with a loud expect, and the `default_theme_is_icy_blue` test pins equality rather than mere presence. `snake_case("icy-blue")` → `icy_blue` keeps existing config values resolving.

### 🟢 CJK search shim is correctly ordered and scoped

The inline script patches `elasticlunr.Index.load` after `elasticlunr.min.js` but before `searcher.js` instantiates the index, so the wrap always takes. The fallback only activates when the query contains CJK codepoints, dedupes against refs elasticlunr already returned, and respects AND/OR semantics via `requireAll`. The validation evidence shows real zh-CN queries returning results from the generated static index, which is the thing that was broken.

### 🟢 Hero derivation keeps the injection-safety contract

Subtitle and Quickstart CTA are now page-derived, and both still flow through `textContent` sinks only — the original "crafted heading cannot inject markup" property holds for translated content too. The `LOCALE_TEXT` chrome labels fall back to English for any locale not in the map, so adding a locale to `locales.toml` degrades gracefully instead of breaking. One observation, not a finding: `localeText('quickStart', ...)` prefers the hardcoded label over the page's own gettext-translated link text when both exist; if the .po catalogs and this map ever drift, the map wins. Fine for chrome gettext doesn't own, just worth remembering when #7472 lands the catalog sync.

CI 18/18 green, MERGEABLE, milestone v0.8.0 already set. Approving.

WareWolf-MoonWall (approved):
Reviewed: head `0081b6a8`, 3 commits, 10 files. Risk: medium. Related #7472.

### 🟢 What looks good — eliminates a SSOT violation: `ICY_BLUE` now lives only in `themes.json`

The previous code defined `ICY_BLUE` as a hardcoded Rust constant in `theme.rs` AND expected an `icy_blue` entry in `web/src/contexts/themes.json`. Two authoritative definitions of the same color palette was a DRY violation. The PR removes the hardcoded constant, adds `icy_blue` to the shared registry (`themes.json`, `colorThemes.ts`, `index.hbs`), and routes `default_theme()` through `theme_by_name("icy_blue")`. The registry is now the single source of truth that zerocode, the React dashboard, and the mdBook docs theme picker all read. ✓ AGENTS.md SSOT rule satisfied.

### 🟢 What looks good — `LazyLock` initialization is the correct fix for the registry-lookup default

`static ACTIVE: RwLock<Theme>` could previously be initialized with the compile-time `DEFAULT_THEME = ICY_BLUE` constant. Since `default_theme()` now calls `theme_by_name()` (a runtime registry lookup), a `LazyLock` defers initialization to first access. The `active_raw()` fallback uses `default_theme()` instead of `DEFAULT_THEME`, keeping the same semantics.

### 🟢 What looks good — `role_literal` in `build.rs` is a clean extension point

The `tui` object in theme registry entries lets themes provide terminal-specific role overrides that don't map cleanly to the web/docs CSS token set. The `role_literal` helper falls back to the CSS-derived mapping when no `tui` override is present, so existing themes require no changes. The panic on malformed hex is loud and fails the build — right behavior for a build-time code generator.

### 🟢 What looks good — CJK search fallback is correctly gated and non-destructive

The `fallbackResults` JS function only activates when `CJK_RE.test(query)` is true (contains CJK codepoints). It runs after the main elasticlunr search and only adds results whose refs are not already present. The weighted field scoring (title: 12, breadcrumbs: 6, body: 2) follows conventional search priority. The `requireAll` logic mirrors elasticlunr's AND vs OR boolean behavior. The CJK smoke test with literal query-to-result-count assertions is the right validation shape.

### 🟢 What looks good — validation covers the changed surfaces

`cargo test -p zerocode theme` (22+ theme/config tests including `default_theme_is_icy_blue`, `icy_blue_rgb_unchanged`, `registry_themes_are_present`), `node --check` for JS syntax, zh-CN mdbook build, and the CJK search smoke test are targeted and appropriate. Full workspace clippy deferred is acceptable for a docs/UI/theme-registry change.

### 🟡 Non-blocking — `icy_blue_rgb_unchanged` test should verify the registry value, not a hardcoded constant

Now that `ICY_BLUE` is gone and the values come from `themes.json`, the `icy_blue_rgb_unchanged` test (if it existed) would need to compare against `theme_by_name("icy_blue")` rather than the old Rust constant. The updated `default_theme_is_icy_blue` test already does this correctly with `assert_eq!(default_theme(), theme_by_name(...))`. If `icy_blue_rgb_unchanged` still exists and pins specific `Color::Rgb` values, it should be updated to drive from the registry too so the test and the implementation share the same truth.

### 🟡 Non-blocking — `index.hbs` Icy Blue swatch colors are manually authored

The swatch in `index.hbs` (`--s0:#080e18;--s1:#64c8ff;--s2:#b48cff;--s3:#ffdc50`) matches the previous Rust constant colors. These are hardcoded in the static HTML. If the `icy_blue` registry entry colors are ever updated in `themes.json`, the `index.hbs` swatch would need a corresponding manual update. Worth noting for future palette changes.

Approving.

PR #7503, opened 2026-06-11
fix(docker): add vim-tiny to runtime images for editor support

### LINKED ISSUE #7469: [Bug]: Default using "vi" but container does not include it

whiteyzy:
## Summary

Add `vim-tiny` to the runtime stage of both `Dockerfile` and `Dockerfile.debian`. The `dialoguer::Editor` component (used by `zerocode` TUI for editing settings) spawns `vi` as the default editor, but `vi` is not present in the minimal runtime images.

## Linked context

Closes #7469

## Verification

- `vim-tiny` is the smallest vim variant (~1 MB) that provides the `vi` symlink
- Both runtime stages already install packages at the same location
- Build verified with `cargo check` and `cargo test` (all pass)

## Risk checklist

- **Breaking change?** No — additional package only, no removals
- **Image size impact?** Minimal — vim-tiny is ~1 MB
- **Security impact?** None — standard Debian package

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -167,6 +167,7 @@ FROM debian:trixie-slim@sha256:f6e2cfac5cf956ea044b4bd75e6397b4372ad88fe00908045
 RUN apt-get update && apt-get install -y \
     ca-certificates \
     curl \
+    vim-tiny \
     && rm -rf /var/lib/apt/lists/*
 
 COPY --from=builder /zeroclaw-data /zeroclaw-data
--- a/Dockerfile.debian
+++ b/Dockerfile.debian
@@ -182,6 +182,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         ca-certificates \
         curl \
         git \
+        vim-tiny \
     && rm -rf /var/lib/apt/lists/*
 
 COPY --from=builder /app/zeroclaw /usr/local/bin/zeroclaw


WareWolf-MoonWall (approved):
Reviewed: head `68e17dec`, 2 files: `Dockerfile` (+1) and `Dockerfile.debian` (+1). Related #7469.

### 🟢 What looks good — correct package, correct layer

`vim-tiny` is the minimal Debian/Ubuntu package that provides the `vi` symlink (~1 MB) without pulling in the full vim ecosystem. Both additions go into the existing runtime `apt-get install` invocation in the respective runtime stages — not a new layer, not a separate `RUN`. That is the right pattern for keeping image layers efficient.

### 🟢 What looks good — complementary to the code fix, not redundant

The parallel code-level fix (#7476, now approved) makes the editor fallback PATH-aware and no longer requires `vi` to exist. This Docker fix is still valuable: users who explicitly set `EDITOR=vi` or `VISUAL=vi` in their environment, or who run the container without custom env vars and prefer the classic `vi` flow, benefit from having the binary present. The two fixes address the same symptom from different angles; neither makes the other unnecessary.

Missing Validation Evidence with literal `docker build` tail output or `docker run --rm ... vi --version` proof, and the standard template sections. For a Dockerfile-only change the lightweight approach is acceptable, but a `docker build` smoke is the appropriate validation signal (noted as not having been run locally).

### 🟡 Non-blocking — `Dockerfile.alpine` not touched if one exists

The summary mentions "both Dockerfiles" but if there is an Alpine variant in the repo it would need `apk add vim` rather than `apt-get install vim-tiny`. Worth confirming no Alpine runtime image is shipped without the fix.

Approving.

Audacity88 (approved):
Context checked: live PR #7503 at head 68e17dec736bb8d74aa104494296369f41add672, linked issue #7469, related PR #7476, WareWolf-MoonWall's approval, the current two-file Docker diff, current container file names, and visible GitHub checks. I did not run local Docker or cargo validation.

### 🟢 What looks good — Debian runtime images now provide the expected editor binary

This is the right small container-side fix for #7469. Both changed lines land in the existing runtime `apt-get install` blocks, so the package is added without a new image layer. `vim-tiny` is also the right Debian package shape here: it provides the expected `vi` command without pulling in the full editor package.

The issue report names `Dockerfile.debian`, and this PR covers that runtime image. It also covers the root Debian runtime stage that ships `zerocode`, which is the other obvious place this symptom can show up.

### 🟢 What looks good — This complements the editor fallback code fix

#7476 makes the code path stop assuming `vi` always exists. This PR handles the container packaging side: users who explicitly set `VISUAL=vi` or `EDITOR=vi`, or who expect a classic `vi` fallback inside the Debian runtime images, now get a real binary.

Those two fixes are complementary rather than redundant. One makes editor discovery more robust; this one makes the Debian runtime images less surprising.

### 🟡 Warning — Docker smoke evidence would be stronger

I am not blocking on this because the patch is a narrow package addition and GitHub checks are green, but the PR body does not include literal `docker build` or `docker run --rm ... vi --version` evidence. For Dockerfile-only fixes, that is the strongest validation signal.

I also checked for an Alpine Dockerfile and did not find one. `Containerfile` is a separate Stagex packaging path, not a Debian `apt-get` runtime image; if the project wants external editing to work inside that package image too, that should be tracked separately because `vim-tiny` does not apply there.

Approving.

PR #7504, opened 2026-06-11
fix(openai): honor config timeout_secs instead of hardcoding 120s

### LINKED ISSUE #6723: Native OpenAI provider hardcodes 120s request timeout, silently ignores timeout_secs config

whiteyzy:
## Summary

`OpenAiModelProvider::http_client()` hardcoded a 120-second request timeout and never read the `timeout_secs` config field. Users setting `timeout_secs = 900` for slow local/self-hosted models would see requests silently die at 120s with no diagnostic.

## Changes

- Add `timeout_secs: u64` field to `OpenAiModelProvider` struct (defaults to 120 for backward compat)
- Add `with_timeout_secs()` builder method (matching OpenRouter's pattern)
- Wire `opts.provider_timeout_secs` through the factory
- `http_client()` now uses `self.timeout_secs` instead of the literal `120`

## Linked context

Closes #6723

## Verification

```bash
cargo check -p zeroclawlabs    # ✅ passes
cargo test -p zeroclaw-providers -- openai   # ✅ 119 passed
```

## Risk checklist

- **Breaking change?** No — default is still 120s
- **Backward compat?** Yes — existing configs without timeout_secs get the same 120s default
- **Security impact?** None

### DIFF
--- a/crates/zeroclaw-providers/src/factory.rs
+++ b/crates/zeroclaw-providers/src/factory.rs
@@ -823,6 +823,9 @@ impl FamilyProviderFactory for OpenAIModelProviderConfig {
         }
         // Default: chat_completions wire with standard API key.
         let mut p = crate::openai::OpenAiModelProvider::with_base_url(alias, api_url, key);
+        if let Some(t) = opts.provider_timeout_secs {
+            p = p.with_timeout_secs(t);
+        }
         if let Some(mt) = opts.provider_max_tokens {
             p = p.with_max_tokens(Some(mt));
         }
--- a/crates/zeroclaw-providers/src/openai.rs
+++ b/crates/zeroclaw-providers/src/openai.rs
@@ -27,6 +27,7 @@ pub struct OpenAiModelProvider {
     base_url: String,
     credential: Option<String>,
     max_tokens: Option<u32>,
+    timeout_secs: u64,
 }
 
 #[derive(Debug, Serialize)]
@@ -216,7 +217,14 @@ impl OpenAiModelProvider {
                 .unwrap_or_else(|| BASE_URL.to_string()),
             credential: credential.map(ToString::to_string),
             max_tokens: None,
+            timeout_secs: 120,
         }
+    }
+
+    /// Override the HTTP request timeout for LLM API calls.
+    pub fn with_timeout_secs(mut self, secs: u64) -> Self {
+        self.timeout_secs = secs;
+        self
     }
 
     /// Set the maximum output tokens for API requests.
@@ -370,7 +378,7 @@ impl OpenAiModelProvider {
     fn http_client(&self) -> Client {
         zeroclaw_config::schema::build_runtime_proxy_client_with_timeouts(
             "model_provider.openai",
-            120,
+            self.timeout_secs,
             10,
         )
     }


WareWolf-MoonWall (approved):
Reviewed: head `f8133e5b`, 2 files: `crates/zeroclaw-providers/src/factory.rs` (+3) and `crates/zeroclaw-providers/src/openai.rs` (+9/-1). Fixes #6723.

### 🟢 What looks good — DRY compliant, mirrors the existing pattern

`timeout_secs: u64` is a construction-time parameter read from `opts.provider_timeout_secs` in the factory and stored for use when `http_client()` builds the `reqwest::Client` per call. This matches the existing `max_tokens` and `credential` fields exactly — provider instances are rebuilt from config when config changes (see #7433), so the stored value is always current. No snapshot or cache concern.

The `with_timeout_secs()` builder mirrors `with_max_tokens()`, the default of `120` preserves backward compatibility, and the factory wiring is a clean 3-line addition. The fix is complete: all three concerns (struct field, builder, factory wiring) are addressed in one commit.

### 🟢 What looks good — validation is appropriate for the change

`cargo check -p zeroclawlabs` + `cargo test -p zeroclaw-providers -- openai` (119 passed) is the right scope. The behavior change (reading `timeout_secs` from config) requires an integration test to fully pin, but the change is a one-line substitution from a literal to `self.timeout_secs` — hard to get wrong.

### 🟡 Non-blocking — no integration test for the wiring

A focused test that constructs an `OpenAiModelProvider` via the factory with `provider_timeout_secs: Some(300)` and asserts the stored value is 300 would prevent future factory wiring regressions. Low priority given the simplicity.

Missing Security & Privacy, Compatibility, and Rollback sections. The content is low-risk enough that this is not a blocker, but please fill out the template for code-change PRs going forward.

Approving.

singlerider:
## Maintainer note: supersedes #7447

This PR and #7447 (@chengzhichao-xydt, filed ~32h earlier) both fix #6723 with near-identical wiring: a `with_timeout_secs` builder on `OpenAiModelProvider`, plumbing `opts.provider_timeout_secs` through the factory, and replacing the hardcoded `120` in `http_client()`.

Landing this one. It carries the cleaner shape (`timeout_secs: u64` with the 120s default living in the constructor, single source of truth) versus #7447's `Option<u64>` resolved at the call site, and it already has an approving review. #7447 is being closed with credit to @chengzhichao-xydt for getting to the root cause first.

Follow-up (non-blocking): neither PR added a test. A one-line assertion that `with_timeout_secs(n)` is honored by `http_client()` would close the regression-guard gap before or shortly after merge.

PR #7510, opened 2026-06-11
fix(quickstart): pre-seed --api-key under snake_case api_key

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `run_quickstart_cli`'s flag pre-seed inserted the provider API key into the submission `fields` map under kebab-case `"api-key"` (`src/main.rs:1208`).
  - The quickstart apply path round-trips submission field keys verbatim into `Config::set_prop_persistent` (`crates/zeroclaw-runtime/src/quickstart/mod.rs:1006`), which rejects kebab paths — so `zeroclaw quickstart --api-key …` always failed at **Create agent** with `Unknown property 'providers.models.<type>.<alias>.api-key'` for every hosted provider.
  - Changed the pre-seed key to snake_case `"api_key"`, matching the submission contract pinned by the existing test `apply_serializes_provider_fields_as_snake_case` (`crates/zeroclaw-runtime/src/quickstart/mod.rs:1722`) and matching what the interactive `api_key` Password prompt already produces.
- **Scope boundary:** does NOT change the apply path, `field_shape()`, the interactive prompt flow, or add a non-interactive mode (see #7507 for the non-TTY loop; #7506 for the next-steps hint).
- **Blast radius:** only the CLI quickstart flag pre-seed; the zerocode TUI and web quickstart surfaces build their own submissions and are unaffected.
- **Linked issue(s):** Fixes #7505. Related #7506, #7507 (separate quickstart defects found in the same release sanity check).
- **Labels:** (path labeler will apply scope labels; suggest `risk: low`, `size: XS`)

## Validation Evidence (required)

- **Commands run and tail output:**

`cargo fmt --all -- --check` — exit 0, no output.

`cargo clippy --all-targets -- -D warnings` — exit 0:

```
    Checking zeroclaw-tools v0.8.0-beta-2 (...\crates\zeroclaw-tools)
    Checking zeroclaw-runtime v0.8.0-beta-2 (...\crates\zeroclaw-runtime)
    Checking zeroclaw-hardware v0.8.0-beta-2 (...\crates\zeroclaw-hardware)
    Checking zeroclaw-channels v0.8.0-beta-2 (...\crates\zeroclaw-channels)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 45s
```

[...]

PR #7512, opened 2026-06-11
fix(quickstart): use snake_case api_key instead of kebab-case api-key

### LINKED ISSUE #7505: [Bug]: quickstart --api-key pre-seeds kebab-case 'api-key' — apply always fails with Unknown property for hosted providers
### Affected component

config/onboarding

### Severity

S1 - workflow blocked

### Current behavior

Running `zeroclaw quickstart` with the documented non-interactive flags for a hosted provider (`--model-provider openai --model gpt-4o-mini --api-key sk-... --agent sanity`) pre-seeds the checklist correctly, but the final **Create agent** step always fails:

```
Your agent was not created — and nothing on disk was changed.
Your existing config is untouched. Fix the following and run quickstart again:

  • Model provider: Unknown property 'providers.models.openai.default.api-key'

Error: quickstart could not finish: 1 problem(s) to fix
```

Root cause: the CLI flag pre-seed inserts the provider field key in kebab-case — `fields.insert("api-key".to_string(), ...)` at `src/main.rs:1208` — but the quickstart apply path round-trips submission field keys verbatim into `Config::set_prop_persistent` (`crates/zeroclaw-runtime/src/quickstart/mod.rs:1006`), and the submission contract requires snake_case keys. The existing test `apply_serializes_provider_fields_as_snake_case` (`crates/zeroclaw-runtime/src/quickstart/mod.rs:1722`) pins the contract: "The submission carries the snake field key `api_key`". The interactive Password-prompt path uses `field_shape()` keys (snake_case) and works; only the `--api-key` flag path is broken.

This breaks every documented invocation of the form shown in the README, `.claude/skills/zeroclaw/references/cli-reference.md`, and the quickstart docs for any provider that needs an API key (OpenAI, Anthropic, OpenRouter, ...). Local providers (`ollama`) are unaffected because they skip the key.

### Expected behavior

Quickstart with `--model-provider`/`--model`/`--api-key`/`--agent` completes the Create step and writes the provider entry with the API key stored encrypted, the same as when the key is entered at the interactive `api_key` prompt.

Likely one-line fix: insert the pre-seeded key under `"api_key"` (snake_case) at `src/main.rs:1208`, plus a regression test covering the flag pre-seed path through `apply_with_surface`.

### Steps to reproduce

```bash
# fresh isolated config dir; any hosted provider + key
  --api-key sk-REDACTED --agent sanity
# walk the remaining checklist: Risk profile -> balanced, Memory -> sqlite,
# Channels -> "Done (channels selector counts as visited)", then "Create agent"
# -> apply fails with the "Unknown property ... api-key" error above
```

### Impact

Affected users: anyone following the documented hosted-provider quickstart (README "Non-interactive (hosted)" example, install scripts, skill docs).
Frequency: always — deterministic for every hosted provider when `--api-key` is passed.
Consequence: first-run setup cannot be completed with flags; users must re-run fully interactively and retype the key at the masked prompt.

### Logs / stack traces

```
Open a selector (Enter), or pick Create. Esc to quit.: ── Create agent

Your agent was not created — and nothing on disk was changed.
Your existing config is untouched. Fix the following and run quickstart again:

  • Model provider: Unknown property 'providers.models.openai.default.api-key'

Error: quickstart could not finish: 1 problem(s) to fix
```

### ZeroClaw version

### Rust version

rustc 1.93.1

### Operating system

Windows 11 Pro (10.0.26200)

### Regression?

Unknown

### Pre-flight checks

whiteyzy:
## Summary

The quickstart non-interactive flow inserts the `--api-key` value under the config key `api-key` (kebab-case), but the TOML config schema expects `api_key` (snake_case). This causes the final **Create agent** step to fail with:

```
Unknown property 'providers.models.openai.default.api-key'
```

## Linked context

Closes #7505

## Verification

```bash
cargo check -p zeroclawlabs  # ✅ passes
```

## Risk checklist

- **Breaking change?** No — the field name now matches what the schema expects
- **Backward compat?** Yes — this was never working before (always rejected)

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1205,7 +1205,7 @@ async fn run_quickstart_cli(
             let mut fields: std::collections::HashMap<String, String> =
                 std::collections::HashMap::new();
             if let Some(key) = api_key.as_deref().filter(|s| !s.is_empty()) {
-                fields.insert("api-key".to_string(), key.to_string());
+                fields.insert("api_key".to_string(), key.to_string());
             }
             form.provider = Some(ProviderChoice::Fresh {
                 kind: found.kind.clone(),


Audacity88:
@whiteyzy thanks, the one-line fix matches the reported root cause: the CLI pre-seed path needs to use the schema field key `api_key`, not `api-key`.

Before this is ready, please add a focused regression for the `--api-key` flag pre-seed path so this cannot slip back. The useful coverage is the path from CLI-seeded provider fields into the quickstart apply contract, proving the generated provider field is `api_key` and no `api-key` key reaches `set_prop_persistent`.

WareWolf-MoonWall (changes requested):
Reviewed: head `7cb811ff`, 1 file: `src/main.rs` (+1/-1). Fixes #7505. Competing with #7510.

### 🔴 Blocking — superseded by #7510, which fixes the same line with an explanatory comment

Both PRs change `"api-key".to_string()` → `"api_key".to_string()` at the same location in `src/main.rs:1208`. #7510 by @JordanTheJet makes the identical fix but also adds an inline comment documenting why the key must be snake_case (the apply path round-trips keys verbatim into `set_prop_persistent`, which rejects kebab paths) and includes thorough end-to-end validation evidence including a live ConPTY repro and a working `zeroclaw agent` follow-up. Only one of these can merge without a rebase conflict.

#7510 is the preferred approach because the explanatory comment is a net addition — it makes the cross-module constraint visible at the call site and helps future authors avoid reintroducing the bug. Please close this PR in favour of #7510.

### 🟢 The fix itself is correct

The one-line change is the right fix. The TOML schema uses `api_key` (snake_case), the apply path rejects kebab keys, and the interactive flow already used `api_key`. The code change is valid; the issue is only that #7510 adds more context and evidence.

If this PR were to proceed instead of #7510, it would need Validation Evidence, Security & Privacy, Compatibility, and Rollback sections per the standard template for code changes in `src/main.rs`.

whiteyzy:
Closed in favor of #7510 which adds an inline comment documenting why the key must be snake_case.

Issue #7509, opened 2026-06-11
[Bug]: update self-test find_asset_url_picks_correct_gnu_over_android fails on Windows hosts (zip assets rejected by is_installable_release_asset)


JordanTheJet:
### Affected component

tooling/ci

### Severity

S3 - minor issue

### Current behavior

`commands::update::tests::find_asset_url_picks_correct_gnu_over_android` fails on any Windows host (verified on clean master @ 3932ed8a7):

```
---- commands::update::tests::find_asset_url_picks_correct_gnu_over_android stdout ----

thread 'commands::update::tests::find_asset_url_picks_correct_gnu_over_android' (75976) panicked at src\commands\update.rs:681:9:
should find an asset
```

Root cause: the test asserts against the **runtime host triple**. `find_asset_url` filters assets through `is_installable_release_asset` (`src/commands/update.rs:271`), which only accepts `zeroclaw-<triple>.tar.gz` / `.tgz`. The test's hardcoded asset list (`src/commands/update.rs:670-678`) ships its Windows entries only as `.zip` (`zeroclaw-x86_64-pc-windows-msvc.zip`), so on a `x86_64-pc-windows-msvc` host no asset qualifies, `find_asset_url` returns `None`, and the `should find an asset` assertion panics. On Linux/macOS hosts the list contains matching `.tar.gz` assets, so CI stays green.

(Side observation: if Windows release archives really are published as `.zip`, then `is_installable_release_asset` rejecting `.zip` would also make `zeroclaw update` unable to self-update on Windows — worth confirming while fixing the test.)

### Expected behavior

`cargo test` passes on Windows hosts. Either parameterize the test via `target_triple_for(...)` with a matching `.tar.gz` asset for the host (as the sibling tests do via `current_target_triple()`), or teach `is_installable_release_asset` about `.zip` for Windows targets and assert accordingly.

### Steps to reproduce

```bash
# on any Windows x86_64 (msvc toolchain) checkout of master
cargo test --lib find_asset_url_picks_correct_gnu_over_android
# -> panicked at src\commands\update.rs:681:9: should find an asset
```

### Impact

Affected users: contributors and release testers running `cargo test` on Windows; full-workspace test runs exit non-zero.
Frequency: always on Windows hosts.
Consequence: local validation battery cannot go green on Windows; masks real failures.

### Logs / stack traces

```
failures:
    commands::update::tests::find_asset_url_picks_correct_gnu_over_android

[...]

Issue #7506, opened 2026-06-11
[Bug]: quickstart success message suggests 'zeroclaw agent <alias>' which fails — -a/--agent flag is required


JordanTheJet:
### Affected component

config/onboarding

### Severity

S2 - degraded behavior

### Current behavior

On success, quickstart prints a next-steps hint with a command that does not parse:

```
Quickstart complete. Created agent `sanity`.

Next steps:
  zeroclaw agent sanity  # chat with this agent in your terminal
```

Running that exact command fails, because `zeroclaw agent` takes no positional argument — the alias must be passed via the required `-a/--agent` flag:

```
error: unexpected argument 'sanity' found

Usage: zeroclaw.exe agent [OPTIONS] --agent <AGENT>

For more information, try '--help'.
```

So the very first command a new user copies after a successful quickstart exits with code 2. (Observed in practice during release testing: a user worked through seven failing variants — `agent sanity`, `--agent sanity`, `-a`, ... — before landing on `agent -a sanity`.)

### Expected behavior

The success message prints a command that works as-is:

```
  zeroclaw agent -a sanity  # chat with this agent in your terminal
```

(Or alternatively, `zeroclaw agent` accepts a positional alias — but fixing the hint string is the minimal change.)

### Steps to reproduce

```bash
# 1. complete quickstart for any agent alias, e.g. "sanity"
zeroclaw --config-dir <fresh-dir> quickstart
# 2. run the exact command printed under "Next steps:"
zeroclaw agent sanity
# -> error: unexpected argument 'sanity' found  (exit code 2)
# the working form is:
zeroclaw agent -a sanity
```

### Impact

Affected users: every new user completing quickstart (the hint is printed unconditionally on success).
Frequency: always.
Consequence: immediate first-command failure after a successful onboarding; poor first-run experience, support noise.

### Logs / stack traces

```
error: unexpected argument 'sanity' found

Usage: zeroclaw.exe agent [OPTIONS] --agent <AGENT>

For more information, try '--help'.
```

### ZeroClaw version

0.8.0-beta-2 (master @ 3932ed8a7)

### Rust version

rustc 1.93.1

### Operating system

Windows 11 Pro (10.0.26200)

### Regression?

Unknown

### Pre-flight checks

- [X] I reproduced this on the latest master branch or latest release.
- [X] I redacted secrets, tokens, and personal data from all submitted content.

JordanTheJet:
[...]

PR #7513, opened 2026-06-11
fix(quickstart): bail with a clear error on non-TTY stdin

### LINKED ISSUE #7507: [Bug]: quickstart infinite redraw loop on non-TTY stdin — flooded 4.3 GB of output instead of erroring
### Affected component

config/onboarding

### Severity

S1 - workflow blocked

### Current behavior

When `zeroclaw quickstart` is run without an interactive TTY (stdin/stdout piped or redirected — CI jobs, scripts, agent harnesses, `zeroclaw quickstart < /dev/null > log`), the dialoguer checklist does not fail with an error. Instead it redraws the main checklist in a tight loop forever, spinning CPU and flooding stdout. In testing it produced a **4.3 GB** output stream in roughly 3 minutes before the process was killed:

```
Open a selector (Enter), or pick Create. Esc to quit. |
> [✓] Model provider     — OpenAI (alias: default, model: gpt-4o-mini)
  [ ] Risk profile       — not yet chosen
  [ ] Memory             — not yet chosen
  [ ] Channels (0..N)    — not yet visited
  [ ] Peer groups        — none — channels accept no peers
  [✓] Agent identity     — alias: sanity, system prompt: 0 chars, 0 personality file(s)
  ── Create agent (locked — fill every selector first)
Open a selector (Enter), or pick Create. Esc to quit. |
> [✓] Model provider     — OpenAI (alias: default, model: gpt-4o-mini)
  ... (identical block repeats indefinitely) ...
```

The main checklist loop around `FuzzySelect::interact_opt()` (`src/main.rs:1359`) appears to re-enter/redraw continuously when the terminal backend cannot read real key events. Under a proper PTY (ConPTY) the same build renders once and waits for input, so this is specific to non-TTY stdio.

Related stale doc: the `Commands::Quickstart` doc-comment (`src/main.rs:383`) still claims the command is "Non-interactive in this build" — but the flags only pre-seed the interactive checklist; there is no `--yes`/non-interactive mode, so headless callers will plausibly hit this loop.

### Expected behavior

If stdin/stdout is not an interactive terminal, quickstart should detect it and exit immediately with a clear error (non-zero exit code), e.g. "quickstart requires an interactive terminal; use `zeroclaw config set` for headless setup" — never spin or flood output. (dialoguer interact errors should abort the checklist loop rather than re-entering it.)

### Steps to reproduce

```bash
# any non-TTY invocation; flags optional
  --api-key sk-REDACTED --agent sanity > quickstart.log 2>&1
# process never exits; quickstart.log grows by gigabytes within minutes
# (observed 4.3 GB in ~3 minutes on Windows 11)
```

### Impact

Affected users: anyone invoking quickstart from scripts, CI, agent tooling, or with redirected output.
Frequency: always on non-TTY stdio.
Consequence: hung process, 100% of a core spinning, multi-GB log/disk exhaustion until killed.

### Logs / stack traces

```
(head of the 4.3 GB output stream — the same frame repeats indefinitely)

Quickstart — create one working agent end-to-end.

Open a selector (Enter), or pick Create. Esc to quit. |
> [✓] Model provider     — OpenAI (alias: default, model: gpt-4o-mini)
  [ ] Risk profile       — not yet chosen
  [ ] Memory             — not yet chosen
  [ ] Channels (0..N)    — not yet visited
  [ ] Peer groups        — none — channels accept no peers
  [✓] Agent identity     — alias: sanity, system prompt: 0 chars, 0 personality file(s)
  ── Create agent (locked — fill every selector first)
Open a selector (Enter), or pick Create. Esc to quit. |
...
```

### ZeroClaw version

### Rust version

rustc 1.93.1

### Operating system

Windows 11 Pro (10.0.26200)

### Regression?

Unknown

### Pre-flight checks

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `zeroclaw quickstart` in a non-interactive context redrew the dialoguer checklist in a tight loop forever instead of erroring — observed flooding **4.3 GB** of output in ~3 minutes before being killed (#7507).
  - Added an up-front interactivity guard in `run_quickstart_cli` that bails before the checklist when **either stdin or stderr is not a terminal**, with a clear, localized error that points headless callers at `zeroclaw config set` (same `std::io::IsTerminal` pattern already used in `src/skills/mod.rs`). The checklist's `FuzzySelect` reads keys from stdin **and** renders on `Term::stderr()`; a non-TTY on *either* stream makes `read_key()` return `Key::Unknown`, which `FuzzySelect` does not treat as a terminal condition, so it redraws forever. Guarding only stdin still left the redirected-stderr shape (`quickstart … > log 2>&1`, the original #7507 reproduction) in the loop.
  - Added the `cli-quickstart-needs-tty` Fluent key to the English catalogue (`crates/zeroclaw-runtime/locales/en/cli.ftl`), worded to name both stdin and stderr; other locales follow the standard local `cargo fluent fill` workflow per `TRANSLATIONS.md` (translation is never a CI operation).
  - Removed the stale "Non-interactive in this build" sentence from the `Commands::Quickstart` doc comment — flags only pre-seed the interactive checklist, and that claim is exactly what misleads scripted callers into hitting the loop.
- **Scope boundary:** does NOT add a non-interactive/`--yes` quickstart mode, and does NOT change dialoguer rendering or the checklist flow on real terminals.
- **Blast radius:** quickstart CLI entry only; the zerocode TUI and web quickstart surfaces have their own entry points and are unaffected.
- **Linked issue(s):** Fixes #7507. Related #7505, #7506 (separate quickstart defects from the same release sanity check).
- **Labels:** (path labeler applies scope labels; suggest `risk: low`, `size: XS`)

`cargo fmt --all -- --check` — exit 0, no output.

`cargo clippy --all-targets -- -D warnings` — exit 0:

```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5m 09s
```

Focused quickstart + i18n tests (`cargo test -p zeroclaw-runtime quickstart` and `... i18n`, covering the checklist apply/validate logic and the Fluent-catalogue completeness this PR's key must satisfy) — exit 0:

```
running 15 tests
test quickstart::tests::channel_type_options_cover_every_schema_channel ... ok
...
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 2052 filtered out; finished in 0.43s

running 11 tests
test i18n::tests::english_descriptions_are_embedded ... ok
...
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 2056 filtered out; finished in 0.04s
```

(The full `cargo test` suite still carries one pre-existing, host-OS-dependent failure unrelated to this diff — `commands::update::tests::find_asset_url_picks_correct_gnu_over_android`, root-caused and tracked in #7509; it does not touch the quickstart guard.)

- **Beyond CI — what did you manually verify?** Both #7507 shapes now exit fast (`exit=1`) with the localized error and produce no redraw-loop output. Each was run against the freshly built `target/debug/zeroclaw` under a hard timeout (regression would have been killed):

  **(a) piped stdin** — `echo "" | zeroclaw --config-dir <isolated> quickstart --model-provider openai --model gpt-4o-mini --api-key sk-x --agent sanity`:
  ```
  Error: Quickstart is interactive and needs a terminal on stdin and stderr. Run it from an interactive shell, or use `zeroclaw config set <path> <value>` for headless configuration.
  exit=1
  ```

  **(b) redirected stderr (the original #7507 repro, `… > log 2>&1`)** — `zeroclaw --config-dir <isolated> quickstart --model-provider openai --model gpt-4o-mini --api-key sk-x --agent sanity > /tmp/qb.log 2>&1`:
  ```
  exit=1
  ----- tail -5 /tmp/qb.log -----
  Error: Quickstart is interactive and needs a terminal on stdin and stderr. Run it from an interactive shell, or use `zeroclaw config set <path> <value>` for headless configuration.
  ----- byte size of /tmp/qb.log -----
  181
  ```
  The `2>&1` log is **181 bytes**, not the multi-GB flood from #7507 — proof the redirected-stderr path no longer enters the checklist redraw loop.

  - Verified interactive behavior is unchanged: with a real terminal on both streams the full quickstart checklist still drives through to completion; the guard only fires when stdin or stderr is non-interactive.
- **If any command was intentionally skipped, why:** none skipped; the full `cargo test` suite was substituted with the focused quickstart + i18n modules plus fmt/clippy for this single-guard CLI change (the only full-suite failure is the unrelated pre-existing #7509 case noted above).

## Compatibility (required)

- Backward compatible? Yes for all interactive use. Headless invocations that previously hung in the redraw loop — whether from non-TTY stdin or redirected stderr — now fail fast with exit code 1 and an actionable message; that is the intended behavior change (#7507).
- Config / env / CLI surface changed? No flags added or removed; the command doc comment no longer claims a non-interactive mode.
- If `No` or `Yes` to either: exact upgrade steps for existing users: scripted callers should use `zeroclaw config set <path> <value>` (or the gateway quickstart API) for headless setup, as the error message states.

Low-risk: `git revert <sha>`.


WareWolf-MoonWall (approved):
Reviewed: head `60105530`, 2 files: `crates/zeroclaw-runtime/locales/en/cli.ftl` (+1) and `src/main.rs` (+18/-2). Fixes #7507.

### 🟢 What looks good — uses the established `IsTerminal` pattern

The guard uses `std::io::IsTerminal::is_terminal(&std::io::stdin())` — the same pattern already present in `src/skills/mod.rs` for editor launching. Placing it at the top of `run_quickstart_cli`, before any form state or dialoguer prompts are constructed, is the right position: no partial state is built before the bail.

### 🟢 What looks good — error message is actionable and localized

`cli-quickstart-needs-tty` is added to the English Fluent catalogue with a fallback string that routes headless callers to `zeroclaw config set`. The message correctly distinguishes the cause (interactive terminal required) from a remediation path (scriptable alternative). The Fluent key follows the existing `cli-quickstart-*` namespace.

### 🟢 What looks good — doc comment correction removes the misleading claim

The old `"Non-interactive in this build: writes balanced defaults..."` comment directly contradicted the actual behavior (the flags only pre-seed selectors; a terminal is required) and was the textual root cause of users expecting headless operation. The replacement is accurate.

### 🟢 What looks good — validation is thorough and honest

The PR documents the pre-existing Windows test failure (`find_asset_url_picks_correct_gnu_over_android`) with reproduction steps, confirms it exists on clean master, and links the root-cause tracking issue #7509. The live repro — observing the 4.3 GB flood before the fix and clean immediate exit after — is exactly the evidence needed for a tight-loop fix.

### 🟡 Non-blocking — non-English Fluent locales not yet filled

English-only for now is the documented norm for new keys (TRANSLATIONS.md). A follow-up fill pass (as part of the next i18n sync) should cover this key.

Approving.

Audacity88 (changes requested):
Reviewed current head `60105530` against #7507, the PR body, the diff, the current `master` quickstart loop, and dialoguer/console 0.12/0.16 behavior. I did not rerun local cargo validation. The early fail-fast direction is right, but the guard is narrower than the reported failure mode, so I do not think this is ready yet.

### 🟢 What looks good — failing before the checklist is the right boundary

Putting the check at the top of `run_quickstart_cli` is the right shape. Quickstart is a dialoguer-driven interactive flow, and scripted callers already have `zeroclaw config set` as the headless path. The localized error also points users at that alternative instead of leaving them with a generic terminal failure.

### 🔴 Blocking — redirected stderr can still hit the redraw loop

The new guard only checks `std::io::stdin().is_terminal()`. That covers the piped-stdin case, but it does not cover the redirected-output case from #7507.

The quickstart selectors call `FuzzySelect::interact_opt()`, and dialoguer renders those prompts on `Term::stderr()`. In dialoguer 0.12 / console 0.16, `Term::read_key()` returns `Key::Unknown` whenever that terminal is not user-attended. `FuzzySelect` does not treat `Unknown` as a terminal condition, so it redraws and loops again.

That means a command like this can still pass the new stdin check while leaving dialoguer attached to a non-terminal stderr:

```bash
  --api-key sk-REDACTED --agent sanity > quickstart.log 2>&1
```

Here stdin can still be the interactive shell, but stderr is redirected into the log. Because the prompt terminal is stderr, this is still the same non-attended terminal condition that caused the tight redraw loop.

Please make the guard match the terminal that dialoguer actually uses. At minimum, quickstart should fail before entering the checklist when either stdin is not a terminal or stderr is not a terminal. If we want to be extra conservative, checking all interactive stdio surfaces is fine too, but stderr is the important missing one for this dialoguer path. Please also add or keep manual evidence for the redirected-stderr shape (`> quickstart.log 2>&1`) so #7507's original reproduction is covered, not only the piped-stdin variant.

singlerider (approved):
I reviewed head 6010553 against #7507, the current quickstart entry path in `src/main.rs`, the existing `IsTerminal` precedent in `src/skills/mod.rs:532`, and the English Fluent catalogue. First review on this PR; CI is fully green and the branch is MERGEABLE.

### 🟢 What looks good — fail-fast guard matches the established pattern

The `std::io::IsTerminal::is_terminal(&std::io::stdin())` check is the same primitive the skills module already uses, placed before any dialoguer prompt construction, so the #7507 redraw loop (4.3 GB of output before kill) becomes an immediate exit-1 with an actionable message. The error routes headless callers at `zeroclaw config set`, which is the real scriptable alternative, not a dead end.

### 🟢 What looks good — the stale doc comment removal is part of the fix, not cosmetics

The `Commands::Quickstart` doc claimed "Non-interactive in this build", which is exactly the claim that misleads scripted callers into piping stdin at it. Correcting it to "flags pre-seed selectors but do not skip them; a terminal is required" closes the documentation half of the same defect.

### 🟢 What looks good — localization done right for a new surface

New string goes through a Fluent key (`cli-quickstart-needs-tty`) with a matching inline fallback, and the body correctly notes other locales follow the local `cargo fluent fill` workflow per `TRANSLATIONS.md` rather than expecting CI to translate.

The validation evidence is thorough: pre-fix repro, post-fix behavior for both bare and flag-pre-seeded invocations, and a ConPTY pass to `Quickstart complete.` proving interactive behavior is unchanged. The one `cargo test` failure is the pre-existing host-OS gap tracked in #7509, disclosed with root cause.

No findings. Approving.

JordanTheJet:
@Audacity88 Good catch — fixed. You're right that the stdin-only guard left the original #7507 shape (`quickstart … > log 2>&1`) in the redraw loop: `FuzzySelect` reads keys from stdin but renders on `Term::stderr()`, so a non-TTY on *either* stream makes `read_key()` return `Key::Unknown`, which it never treats as a terminal condition.

The guard now bails before the checklist when **stdin OR stderr is not a terminal** (still the `std::io::IsTerminal` idiom from `src/skills/mod.rs`), and `cli-quickstart-needs-tty` now names both streams.

Verified both repros against a fresh build (each under a hard timeout so a regression would've been killed):

- **piped stdin** (`echo "" | quickstart …`) → `exit=1` with the localized error.
- **redirected stderr** (`quickstart … > /tmp/qb.log 2>&1`) → `exit=1`; the `2>&1` log is **181 bytes**, not the multi-GB flood — the redirected-stderr path no longer enters the loop.

Full literal evidence (both exit codes + the log tail/size) is in the PR body's Validation Evidence section. Interactive runs with a real terminal on both streams are unaffected. PTAL.

Audacity88 (approved):
Reviewed current head `d0fa4305` against #7507, my prior CHANGES_REQUESTED review, the overlapping #7516 thread, the diff, the PR body, current CI metadata, and a local smoke build of the PR head. I built `zeroclaw` locally and re-ran the two non-TTY repro shapes with a 5-second watchdog; both exited with code 1 and produced a 181-byte error instead of entering the redraw loop.

### ✅ Resolved — redirected stderr now fails before dialoguer starts

The updated guard now checks both `stdin` and `stderr` before the quickstart checklist is constructed. That matches the dialoguer path this bug depends on: `FuzzySelect` reads keys from stdin and renders through `Term::stderr()`, so guarding only stdin left the `quickstart ... > log 2>&1` shape exposed.

I also locally verified the two important repro forms against the built PR head:

```bash
echo "" | zeroclaw --config-dir <isolated> quickstart ...
# rc=1, 181 bytes, localized "needs a terminal on stdin and stderr" error

zeroclaw --config-dir <isolated> quickstart ... > quickstart.log 2>&1
# rc=1, 181 bytes, same localized error
```

That covers the blocker from my earlier review.

### 🟢 What looks good — the overlapping quickstart fix lands on the better recovery path

The overlapping #7516 branch raised two useful points. I think #7513 chooses the right final recovery hint: for headless CLI automation, `zeroclaw config set <path> <value>` is the scriptable setup path. "Run from an interactive terminal" is still present for users who meant to use the checklist, and the PR also removes the stale doc comment that made quickstart sound flag-only/non-interactive.

On locale coverage, #7513 adds the English Fluent source key and uses the `t()` wrapper. The current runtime CLI loader starts from English and overlays locale catalogues, so non-English users fall back to this English string when a translated key has not been filled yet. I would not copy #7516's generated non-English entries as-is; leave that to the normal `cargo fluent fill` workflow or a focused translation pass.

### 🟡 Warning — CI still needs a clean rerun before merge

The current `Test` check is red, and the required status is blocked. The PR body mentions the known #7509 test failure, but the failed-job annotation I can see points at runner disk exhaustion while writing the GitHub worker diagnostic log. That looks like infrastructure/output-pressure noise rather than a code-review finding in this diff, especially with the local smoke above, but it still means this PR is not merge-ready until CI is rerun or otherwise cleared.

Approving to clear my prior blocker. This PR is not merge-ready until CI passes and the required status clears.

PR #7516, opened 2026-06-11
fix(quickstart): fail fast when stdin is not a TTY instead of infinite redraw loop

### LINKED ISSUE #7507: [Bug]: quickstart infinite redraw loop on non-TTY stdin — flooded 4.3 GB of output instead of erroring

whiteyzy:
## Summary

When `zeroclaw quickstart` runs without an interactive terminal (piped stdin, CI jobs, scripts, agent harnesses), the dialoguer checklist enters a tight redraw loop that floods stdout indefinitely. Observed: 4.3 GB output in ~3 minutes.

Add an early check: if stdin is not a terminal, bail immediately with a helpful error message directing users to the flag-based non-interactive path.

## Linked context

Closes #7507

## Verification

```bash
cargo check -p zeroclawlabs  # ✅ passes
```

## Risk checklist

- **Breaking change?** No — only adds an early-exit guard for non-TTY environments
- **Prior behavior:** Infinite flood of output
- **New behavior:** Clear error message

### DIFF
--- a/crates/zeroclaw-runtime/locales/en/cli.ftl
+++ b/crates/zeroclaw-runtime/locales/en/cli.ftl
@@ -504,6 +504,8 @@ cli-press-enter = Press Enter to exit...
 cli-quickstart-title = Quickstart — create one working agent end-to-end.
 cli-quickstart-cancelled = Quickstart cancelled. No config written.
 cli-quickstart-incomplete = {"  "}Not all selectors are filled yet.
+cli-quickstart-needs-tty = quickstart requires an interactive terminal with both input and output attached.
+    Re-run inside a terminal (or via ZeroCode TUI) to complete setup interactively.
 cli-no-channels-compiled = {"  "}No channel types are compiled into this binary.
 cli-quickstart-complete = Quickstart complete. Created agent `{$alias}`.
 cli-next-steps = Next steps:
--- a/crates/zeroclaw-runtime/locales/es/cli.ftl
+++ b/crates/zeroclaw-runtime/locales/es/cli.ftl
@@ -444,6 +444,7 @@ cli-press-enter = Presiona Enter para salir...
 cli-quickstart-title = Quickstart — crea un agente funcional de principio a fin.
 cli-quickstart-cancelled = Quickstart cancelado. No se escribió ninguna configuración.
 cli-quickstart-incomplete = {"  "}Aún no se han completado todos los selectores.
+cli-quickstart-needs-tty = quickstart requires an interactive terminal with both input and output attached.n    Re-run inside a terminal (or via ZeroCode TUI) to complete setup interactively.
 cli-no-channels-compiled = {"  "}No hay tipos de canal compilados en este binario.
 cli-quickstart-complete = Quickstart completado. Se creó el agente `{$alias}`.
 cli-next-steps = Siguientes pasos:
--- a/crates/zeroclaw-runtime/locales/fr/cli.ftl
+++ b/crates/zeroclaw-runtime/locales/fr/cli.ftl
@@ -447,6 +447,7 @@ cli-press-enter = Appuyez sur Entrée pour quitter...
 cli-quickstart-title = Quickstart — créez un agent fonctionnel de bout en bout.
 cli-quickstart-cancelled = Quickstart annulé. Aucune configuration écrite.
 cli-quickstart-incomplete = {"  "}Tous les sélecteurs ne sont pas encore renseignés.
+cli-quickstart-needs-tty = quickstart requires an interactive terminal with both input and output attached.n    Re-run inside a terminal (or via ZeroCode TUI) to complete setup interactively.
 cli-no-channels-compiled = {"  "}Aucun type de canal n'est compilé dans ce binaire.
 cli-quickstart-complete = Quickstart terminé. Agent `{$alias}` créé.
 cli-next-steps = Étapes suivantes :
--- a/crates/zeroclaw-runtime/locales/ja/cli.ftl
+++ b/crates/zeroclaw-runtime/locales/ja/cli.ftl
@@ -444,6 +444,7 @@ cli-press-enter = 終了するにはEnterキーを押してください...
 cli-quickstart-title = クイックスタート — 1つの動作するエージェントをエンドツーエンドで作成します。
 cli-quickstart-cancelled = クイックスタートをキャンセルしました。設定は書き込まれていません。
 cli-quickstart-incomplete = {"  "}すべてのセレクターがまだ入力されていません。
+cli-quickstart-needs-tty = quickstart requires an interactive terminal with both input and output attached.n    Re-run inside a terminal (or via ZeroCode TUI) to complete setup interactively.
 cli-no-channels-compiled = {"  "}このバイナリにコンパイルされているチャンネルタイプはありません。
 cli-quickstart-complete = クイックスタートが完了しました。エージェント `{$alias}` を作成しました。
 cli-next-steps = 次のステップ:
--- a/crates/zeroclaw-runtime/locales/zh-CN/cli.ftl
+++ b/crates/zeroclaw-runtime/locales/zh-CN/cli.ftl
@@ -443,6 +443,7 @@ cli-press-enter = 按 Enter 退出...
 cli-quickstart-title = Quickstart — 端到端创建一个可用的 agent。
 cli-quickstart-cancelled = 已取消 quickstart。未写入配置。
 cli-quickstart-incomplete = {"  "}尚未填写所有选择器。
+cli-quickstart-needs-tty = quickstart requires an interactive terminal with both input and output attached.n    Re-run inside a terminal (or via ZeroCode TUI) to complete setup interactively.
 cli-no-channels-compiled = {"  "}此二进制文件中未编译任何通道类型。
 cli-quickstart-complete = Quickstart 完成。已创建 agent `{$alias}`。
 cli-next-steps = 后续步骤：
--- a/src/main.rs
+++ b/src/main.rs
@@ -1084,8 +1084,7 @@ async fn run_quickstart_cli(
     // Guard: dialoguer reads from Term::stderr() and redraws infinitely
     // when stderr is redirected or non-attended (#7507).  Check both the
     // stdin handle (console user input) and stderr (dialoguer output).
-    let terminal_attended =
-        std::io::stdin().is_terminal() && std::io::stderr().is_terminal();
+    let terminal_attended = std::io::stdin().is_terminal() && std::io::stderr().is_terminal();
     if !terminal_attended {
         let msg = t(
             "cli-quickstart-needs-tty",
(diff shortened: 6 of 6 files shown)


Audacity88 (changes requested):
Reviewed PR #7516 against linked issue #7507, the diff at `c4b14a6`, current passing CI, and public source excerpts covering dialoguer's terminal handling, console's TTY detection, and ZeroClaw's localization requirements. I did not see prior inline comments or formal reviews on this PR.

### 🟢 What looks good — Early-exit approach for non-TTY environments

Thanks for jumping on the quickstart flood quickly. The early-exit direction is the right shape for this bug, and keeping the change close to the checklist entry point makes the intended behavior easy to reason about.

### 🔴 Blocking — Check the terminal dialoguer actually reads from

This guard only checks `std::io::stdin().is_terminal()`, but the failing checklist path uses `FuzzySelect::interact_opt()`, and dialoguer renders/reads that prompt through `Term::stderr()`. The linked issue's repro redirects the process with `> quickstart.log 2>&1`, so an interactive stdin can still leave dialoguer reading from redirected/non-attended stderr. In that case `console::Term::read_key()` returns `Key::Unknown`, which `FuzzySelect` ignores and then redraws again, leaving the same infinite loop in place for the reported failure mode.

Please gate on the attended terminal handle used by dialoguer (or otherwise ensure every quickstart dialoguer prompt fails before entering the redraw loop when stderr is redirected/non-attended), not just stdin.

### 🔴 Blocking — Localize the new user-facing error

The new `anyhow::bail!(...)` message is a user-facing CLI error, but it is introduced as a bare English string. ZeroClaw's CLI/onboarding output is required to go through Fluent (`fl!()`) rather than raw string literals, so this needs a `cli-quickstart-*` Fluent key added across the supported locale files with the Rust call site using that key.

### 🟡 Warning — Do not point users at a non-existent flag-only quickstart path

The new error text tells headless users to "Re-run with all required flags for non-interactive use", but today's quickstart flags only pre-seed the interactive checklist; they do not complete the other required selectors or provide a true `--yes`/headless mode. Please point non-TTY users away from the current flag-only quickstart wording until a supported headless path is documented, and consider cleaning up the stale quickstart doc-comment that still says "Non-interactive in this build."

whiteyzy:
Thanks for the review. Addressed all three points:

1. TTY check — now checks both stdin.is_terminal() AND stderr.is_terminal() since dialoguer reads from stderr
2. i18n — added cli-quickstart-needs-tty Fluent key in locales/en/cli.ftl, uses t() wrapper
3. Error message — removed misleading reference to non-existent flag-only headless mode; now directs users to interactive terminal or ZeroCode TUI

CI build verified with cargo check.

WareWolf-MoonWall (commented):
Reviewed the latest push (2 commits) against @Audacity88's CHANGES_REQUESTED from earlier today. Two of the three original items are resolved; one is only partially addressed, and there's an additional format-check failure to clear before re-requesting review.

### ✅ Resolved — Dialoguer's terminal handle now checked

The guard now checks both `stdin.is_terminal() && stderr.is_terminal()`, which correctly covers dialoguer's `Term::stderr()` read path. This addresses the `> quickstart.log 2>&1` repro from issue #7507 — stdin alone being interactive was insufficient because dialoguer reads/renders through stderr.

### ✅ Resolved — Error message no longer references non-existent flag-only path

The new Fluent text directs users to "an interactive terminal (or via ZeroCode TUI)" rather than a non-existent flag-driven headless mode. Correct direction for the current state of the CLI.

### 🔴 Blocking — Fluent key missing from four locale files

`cli-quickstart-needs-tty` was added to `crates/zeroclaw-runtime/locales/en/cli.ftl` but the four other supported locale files — `es/cli.ftl`, `fr/cli.ftl`, `ja/cli.ftl`, and `zh-CN/cli.ftl` — are unchanged. At runtime on a non-English locale, the runtime will fall back to the English literal passed to `t()`. Per AGENTS.md, all user-facing CLI strings must go through Fluent across all supported locales. Please add the key to the remaining four files before re-requesting review.

### 🔴 Blocking — CI format check is failing

`Quality Gate/Format` is failing on the current head. Run `cargo fmt --all` locally, commit the result, and re-push. The required gate (`Quality Gate/CI Required Gate`) is also blocked by this.

Holding this for @Audacity88 to re-review once both blockers above are addressed.

whiteyzy:
@WareWolf-MoonWall Both blockers addressed:

1. Fluent key added to all 5 locales (es/fr/ja/zh-CN)
2. cargo fmt --all applied — CI format check should pass now

Audacity88:
## Queue Routing

@whiteyzy Thanks for the follow-up. This now overlaps directly with #7513.

Both PRs now check stdin and stderr before entering the dialoguer checklist. #7513 is the canonical path for #7507 because it also removes the stale Quickstart doc comment that described the command as non-interactive, and its error points headless callers at a scriptable setup path.

Two pieces from this PR are worth checking before #7516 closes:

- Locale coverage: this PR adds `cli-quickstart-needs-tty` to all locale files. The non-English entries appear mechanically inserted and include a literal `n`, so they should not be copied as-is, but #7513 should decide whether to add placeholder/fallback entries now or leave translation fill to the normal workflow.
- User-facing hint: this PR points users toward an interactive terminal / ZeroCode TUI, while #7513 points headless callers at `zeroclaw config set`. The final wording should choose the intended recovery path for non-TTY quickstart.

After those two checks are handled in #7513, this PR should close as superseded so review stays on one branch.

singlerider (commented):
Reviewed at head `f55a789` against current `master`. All of @Audacity88's and @WareWolf-MoonWall's blockers are resolved here — their reviews were on earlier commits. Not approving over @Audacity88's active changes-requested, but confirming it can be converted.

### RESOLVED ✅ — gates on the terminal dialoguer actually uses

The guard now checks `std::io::stdin().is_terminal() && std::io::stderr().is_terminal()`. That covers the real failure mode: `FuzzySelect`/dialoguer renders and reads through `Term::stderr()`, so the #7507 repro (`> quickstart.log 2>&1`) could leave an interactive stdin while stderr is redirected, where `read_key()` returns `Key::Unknown` and the prompt redraws forever. Requiring both handles attended fails fast before the redraw loop instead of just checking stdin.

### RESOLVED ✅ — localized across all five locales

`cli-quickstart-needs-tty` is now present in all supported locale files (`en`, `es`, `fr`, `ja`, `zh-CN` — confirmed one entry each), and the call site uses `t("cli-quickstart-needs-tty", ...)` with an English fallback, so this satisfies the all-locales Fluent requirement rather than emitting a bare literal.

### RESOLVED ✅ — error text no longer points at a non-existent flag mode, and CI is green

The message directs non-TTY users to "Re-run inside a terminal (or via ZeroCode TUI)" rather than the flag-only headless path that doesn't exist yet. `Format`, `Lint`, `Test`, and `CI Required Gate` all pass at this head, clearing @WareWolf-MoonWall's format-failure blocker.

### 🔵 Before merge

The branch is `CONFLICTING` / `DIRTY` against current `master` and needs a refresh + a fresh green run.

@Audacity88 — your two blockers (dialoguer terminal handle, localized error) and @WareWolf-MoonWall's (four missing locale keys, format check) are all addressed at this head, so this is ready for you to convert or dismiss. From my read the fix is correct and complete; I'm leaving your changes-requested as the authoritative state rather than approving over it.

Audacity88:
@whiteyzy closing this as superseded by #7513, which has now merged and closed #7507.

The current head resolves the original terminal-handle/localization blockers, but the remaining review direction from the queue-routing comment has landed on #7513 as the canonical fix: it guards both stdin and stderr, removes the stale quickstart doc comment that made quickstart sound non-interactive, and points headless callers at the scriptable `zeroclaw config set` path. That keeps the fix on one branch instead of carrying a conflicting duplicate branch.

One thing I would not carry forward from this branch is the mechanically inserted non-English locale entries: they include a literal `n` in the value, so any future translation fill should use the normal Fluent/i18n workflow rather than copying those lines as-is.

Thanks for jumping on the redraw-loop bug quickly; the important behavior is covered by the merged fix now.

PR #7519, opened 2026-06-11
fix(config): persist [[mcp.servers]] per-field edits via natural-key dirty-path walker

tidux:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `d06ed25` (#7267) shipped the per-field `[[mcp.servers]]` editor by wiring `#[natural_key]` into the `Configurable` derive's in-memory `set_prop` / `get_prop` arms, but the on-disk incremental writer was never taught about array-of-tables. `apply_dirty_path` in `crates/zeroclaw-config/src/schema.rs` walked the serialized config assuming every node was a `toml::Table`; `mcp.servers` serialises as `[[mcp.servers]]` (a `toml::Value::Array`), so `lookup_path_in_table` returned `None` at the natural-key segment, `should_delete` was misclassified as `true`, and `delete_path_in_doc` bailed at its own `as_table_mut()` against the same `ArrayOfTables`. Net effect: `handle_config_set` returned `Ok({set: true})`, the dashboard/TUI showed the new value (it reads from the in-memory `Config`), and `config.toml` kept its stale contents until the next process restart wiped the in-memory change.
  - Surface the `#[natural_key = "<f>"]` attribute through the `Configurable` derive as a new `MapKeySection::natural_key: Option<&'static str>` column so the saver can ask "is this section a natural-key Vec, and if so what field holds each element's alias?" without cross-crate plumbing or hardcoded paths.
  - `apply_dirty_path` now takes a dedicated branch when a matching `#[natural_key]` section is found, locating or creating the matching `[[<section>]]` table by alias and reconciling the inner suffix against it. Three dirty-path shapes are handled, mirroring exactly what `handle_config_set` / `handle_config_map_key_create` / `delete` / `rename` emit on the wire: per-field edit, whole-element write (rename target / fresh add), and whole-element delete (rename source / remove). Empty arrays drop their slot so a vestigial `[[mcp.servers]]` header doesn't reparse as a default-shaped phantom entry.
  - 5 new `save_dirty_*` regression tests seed a real on-disk file, run the exact `mark_dirty` / `set_prop_persistent` / `save_dirty` sequence the dispatcher emits, and assert the resulting on-disk bytes. A 6th schema-contract test pins `MapKeySection::natural_key` so a future derive change can't silently regress the writer back into the broken table-only walk. A new end-to-end RPC test (`config_set_persists_mcp_server_field_to_disk`) goes through the full dispatch surface and asserts both in-memory mutation AND on-disk persistence — the missing assertion the pre-existing `config_set_*` tests didn't have, which is exactly what let this bug ship.
- **Scope boundary:** Plain `Vec<T>` schema sections without `#[natural_key]` (peripheral.boards, classification rules, etc.) keep their legacy whole-array `ObjectArray` round-trip path entirely unchanged. Wire format (`ConfigTemplateEntry` in `crates/zeroclaw-runtime/src/rpc/types.rs`) is byte-identical: its `From<MapKeySection>` impl simply ignores the new field. Hand-edited `config.toml` files remain byte-identical apart from the targeted entry — `ensure_array_of_tables_entry` appends-or-edits in place and never rewrites unrelated sections.
- **Blast radius:** Touches `zeroclaw-config` (one new field on `MapKeySection`, ~5 new helper fns + 6 tests in `schema.rs`), `zeroclaw-macros` (three `MapKeySection` arms now emit `natural_key:`), `zeroclaw-runtime` (one new dispatch test). All `MapKeySection` literal sites are updated. No public-API breakage: the new `natural_key` field is additive on a non-exhaustively-matched struct that's only constructed inside the derive and only consumed by struct-field access.
- **Linked issue(s):** Related #7267 (the per-field MCP editor commit this fixes the persistence gap for).
- **Labels:** Auto-applied by the path labeler on open — expect `bug`, `size: L`, `risk: medium`, `config`, plus the auto scope labels for `crates/zeroclaw-config/`, `crates/zeroclaw-macros/`, `crates/zeroclaw-runtime/`. I'll snapshot the final set in a comment after the labeler runs.

```text
$ cargo fmt --all -- --check
$ echo exit: $?
exit: 0
```

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 24s
$ echo exit: $?
exit: 0
```

```text
$ cargo test
test result: ok. 243 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.11s
test result: ok. 275 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.10s
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.19s
test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 159 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.16s
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s
# 896 passed, 0 failed, 7 ignored across the root binary's 9 test targets.
# The 7 ignored are long-running integration tests intentionally gated off
# in the default `cargo test` run.
```

Per-crate evidence for the three crates this PR touches (run separately because the AGENTS.md battery `cargo test` only exercises the root binary):

```text
$ cargo test -p zeroclaw-config --lib
test result: ok. 825 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.39s
```

```text
$ cargo test -p zeroclaw-runtime --lib
test result: ok. 2070 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 5.20s
```

```text
$ cargo test -p zeroclaw-macros
test result: ok. 0 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out; finished in 0.00s
# proc-macro crate; doctests only, both intentionally `ignore`d in d06ed25.
```

Bug-focused tests confirming the fix:

```text
$ cargo test -p zeroclaw-config --lib save_dirty_persists_mcp_server_field_via_natural_key
test schema::tests::save_dirty_persists_mcp_server_field_via_natural_key ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 824 filtered out

$ cargo test -p zeroclaw-runtime --lib config_set_persists_mcp_server_field_to_disk
test rpc::dispatch::tests::config_set_persists_mcp_server_field_to_disk ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2070 filtered out
```

  - Pre-fix reproduction: applied just the in-config-crate regression test (no writer changes) and confirmed it fails with the on-disk file still showing the seed `command = "/usr/bin/mcp-fs"` value while in-memory state showed the new `/usr/local/bin/mcp-fs`. Matches the originally-reported symptom exactly.
  - Post-fix: all 5 new `save_dirty_*` writer tests + the new end-to-end RPC test pass. The 7 pre-existing `mcp_servers_*` semantic tests from `d06ed25` still pass, confirming the in-memory natural-key arm is untouched.
  - Round-tripped a real config.toml through the dispatcher test: seed file → `handle_config_set("mcp.servers.fs.command", "/usr/local/bin/mcp-fs")` → on-disk read → `toml::from_str::<Config>` → reparse equals the in-memory state. Catches any mis-shaped writer output (e.g. an inline `mcp.servers.fs = { ... }` table sneaking in instead of a second `[[mcp.servers]]`).
  - Verified the array-empty path: `save_dirty_drops_array_header_when_last_mcp_server_removed` confirms the writer drops the `ArrayOfTables` slot once empty, preventing a degenerate `[[mcp.servers]]` header from reparsing as a default-shaped phantom entry that then fails validation.

- Secrets / tokens / credentials handling changed? `No` — the writer's secrets path (encryption, 1Password reference restoration) runs upstream of the dirty-path walk and is untouched. Secret-marked inner fields under a `[[mcp.servers]]` entry (e.g. a future `mcp.servers.<name>.api_key`) would route through the same `encrypt_secrets` pass before the dirty-path walk and inherit the existing protection.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No` — test fixtures use generic `"fs"` / `"github"` / `"/usr/bin/mcp-fs"` / `"https://mcp.example/"` strings.

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` — TOML wire format unchanged (the writer now produces the correct `[[mcp.servers]]` shape it was always supposed to), runtime API surface unchanged (`MapKeySection`'s new `natural_key` field is additive and not exhaustively matched anywhere), RPC wire format unchanged (`ConfigTemplateEntry` ignores the new field).
- If `No` or `Yes` to either: exact upgrade steps for existing users: None required. Operators who hit this bug pre-fix have a working in-memory state until the daemon restarts; once this fix lands, the next per-field edit (or a no-op `config_set` triggered by the dashboard) will write the correct on-disk content. Operators whose `config.toml` is out of sync with what they thought they configured should re-apply the affected edits through the dashboard or TUI, which will now persist correctly.

- **Fast rollback command/path:** `git revert 72121e9` (single-commit branch).
- **Observable failure symptoms:** If a regression occurs, look for `[[mcp.servers]]`-shape mishandling in:
  - `config.toml` after a per-field MCP edit: the targeted `command` / `transport` / `url` field should reflect the new value; an unchanged on-disk value with a successful `config/set` RPC response is the original symptom.
  - Reparse failures: `toml::from_str::<Config>` failures on a save_dirty output (would surface in tests; in practice in `Config::load_or_init` at next daemon start).
  - Vestigial empty `[[mcp.servers]]` headers in `config.toml` after deleting the last server (would cause a phantom default-shaped entry on reparse, failing `validate_mcp_config`).
  - `apply_dirty_natural_key_path` is gated on `find_natural_key_section_for_path` matching, which only returns `Some` for `Kind::List` + `natural_key.is_some()` sections; today only `mcp.servers` qualifies, so any incidental regression on non-MCP Vec sections would point at the gating itself.


JordanTheJet (commented):
Withdrawn — this was posted in the wrong review format. Superseded by my re-review below.

JordanTheJet (dismissed):
Reviewed at head 72121e9. Housekeeping first: the comment-review I posted on this PR earlier today went out in the wrong format and has been withdrawn and minimized — this review supersedes it entirely; nothing from it is left open. There are no other reviews or inline threads on this PR, and all 16 CI checks are green.

Thanks @tidux — this closes a genuinely nasty silent-divergence bug: `config/set` on `mcp.servers.<name>.<field>` returned success and the UI showed the new value, while `config.toml` kept the stale contents until a restart wiped the in-memory edit. The fix surfaces `#[natural_key]` through `MapKeySection` and gives `apply_dirty_path` an array-of-tables-aware branch that locates, edits, creates, or deletes the matching `[[mcp.servers]]` entry by alias.

Before writing this I cross-checked the diff against the base commit d06ed25 (my local checkout is behind the base, so I pulled the relevant base blobs to compare against), and verified the load-bearing claims rather than taking the description's word:

- The natural-key branch's simplified "mem missing → delete" rule is exactly equivalent to the generic walker's mem-vs-default rule for alias-keyed children, because `Config::default()` never contains alias entries — the `let _ = default_table` comment is accurate.
- `dirty_paths` is a `HashSet<String>`, so dirty paths apply in nondeterministic order. I traced the rename batch (both aliases dirty), create-then-edit, and delete-with-stale-per-field-edit under both orderings; all converge, because every write derives from the current in-memory state and deletes key off mem-absence.
- `encrypt_secrets` runs before `full_table` is serialized in `save_dirty`, so the new branch writes ciphertext like every other path — the Security section's claim holds.
- `Option::is_none_or` needs Rust 1.82; the workspace pins rust-version 1.87. Fine.
- Wire-format claim verified: `natural_key` is additive and `ConfigTemplateEntry`'s `From` impl ignores it.

### 🟢 What looks good — tests that pin the failure surface, not just the fix

The test design here is worth repeating across the codebase: seeding a real on-disk file so `save_dirty` exercises its incremental path instead of the full-`save` fallback (which would mask the bug), replicating the exact wire sequences the dispatcher emits (`create_map_key` + `mark_dirty`, rename marking both aliases), asserting raw bytes and reparse round-trips, and the `map_key_sections_exposes_natural_key_for_mcp_servers` contract test that pins the derive output so the writer can't silently regress back into the table-only walk. The end-to-end RPC test adds the on-disk assertion whose absence is precisely what let the original bug ship. The PR body's validation evidence — pre-fix reproduction, per-crate runs, rollback plan — is the standard everything should aim for.

### 🔴 Blocking — resolve_dirty_segments_in is a verbatim copy of resolve_dirty_segments

The new helper is token-for-token identical to the pre-existing `resolve_dirty_segments(root: &toml::Table, raw: &[&str]) -> Vec<String>` in the same file — I diffed the two directly at the base commit. Its doc comment justifies the copy as being "rooted at an arbitrary `toml::Table`", but the existing function already takes an arbitrary table root, so the stated reason for the second copy doesn't hold.

I'm blocking on this one because it recreates the exact hazard this PR exists to fix: the root cause here was two parallel implementations of one walk (in-memory vs on-disk) drifting apart. A verbatim second copy of the segment resolver means the next dash-handling fix lands in one copy and silently misses the other. The fix is small: delete the new function and call `resolve_dirty_segments(table, &inner_raw)` at the single call site.

### 🟡 Warning — the wrong-kind bail is silent, and inline-array configs still hit the original symptom

`ensure_array_of_tables_entry` / `find_array_of_tables_entry_mut` / `delete_array_of_tables_entry` all return `None` silently when the doc node isn't an `Item::ArrayOfTables`. The doc comment frames this as refusing to clobber invalid hand-edits like `mcp.servers = "foo"`, which is reasonable. But `servers = [{ name = "fs", command = "..." }]` — an inline array of tables — is valid TOML that parses into the identical `Config`, and in `toml_edit` it's `Item::Value(Value::Array)`, not `ArrayOfTables`. A user with that file shape gets exactly the symptom this PR fixes: `config/set` succeeds, the UI updates, disk stays stale, and nothing tells anyone.

At minimum, emit `::zeroclaw_log::record!(WARN, ...)` on the bail paths so the divergence is observable — `schema.rs` already uses that pattern for the 0600 permissions-hardening fallback. Converting a compatible inline array to an `ArrayOfTables` would close the gap fully and is fine as a tracked follow-up with an assignee.

### 🔵 Suggestion — state the order-independence invariant where it must be preserved

The convergence under `HashSet` ordering (verified above) is load-bearing but invisible: nothing in `apply_dirty_natural_key_path` says its three shapes must remain commutative across one save batch. A one-line comment naming the invariant would keep a future contributor from breaking it with, say, an early-exit that assumes whole-element writes run before per-field ones.

### 🔵 Suggestion — the natural-key prefix match is exact while the walker is kebab-tolerant

`find_natural_key_section_for_path` matches `dotted.starts_with("<section>.")` literally, while the generic walker resolves each segment kebab→snake. Harmless today because `mcp.servers` contains no underscores, but a future `#[natural_key]` section with an underscore in its path, addressed via a kebab wire path, would miss this branch and fall back to the broken table walk — the exact failure mode the new contract test exists to prevent. Either resolve the section-prefix segments dash-insensitively or document the constraint where section paths are defined. Relatedly, swapping the per-section `format!("{}.", ...)` for `strip_prefix(section.path).and_then(|s| s.strip_prefix('.'))` removes an allocation and makes that fix easier.

### 🔵 Suggestion — where are dots in natural-key aliases rejected?

`rest.split_once('.')` assumes aliases never contain a dot, and I didn't find dot-rejection in `create_map_key` / `rename_map_key` validation at the base. The in-memory router splits identically, so memory and disk at least fail consistently — but if dots aren't rejected at creation, `mcp.servers.my.server.command` mis-routes everywhere. If validation exists somewhere I missed, point me at it; if not, rejecting `.` in natural-key aliases is a cheap hardening win and fine as a separate PR.

### 🔵 Suggestion — two micro-cleanups and three test additions

Rename the parameter to `_default_table` instead of the `let _ = default_table;` dance (keep the explanatory comment — it's good). And three tests worth adding: a per-field edit through a hyphenated inner key (e.g. `headers.<auth-key>` — the dash-aware resolution is the sole reason the duplicated helper exists, and it's currently untested through the natural-key branch); an explicit unset/None per-field case to exercise the case-1 delete branch directly rather than only via whole-entry deletes; and a test pinning the wrong-kind bail behavior from the 🟡 above, whatever shape you land on — silent or warned, it should be documented by a test.

---

Verdict: requesting changes on the 🔴 alone — it's a small fix and everything else holds up well. The 🟡 I'd like addressed here or deferred with a tracked issue and assignee; the 🔵 items are yours to take or decline. Happy to re-review quickly once the duplicate is gone.

tidux:
Thanks for the thorough re-review @JordanTheJet — and for catching the duplicated resolver, that one was a self-inflicted trap. Head is now `668aa495e`. Walking the items in order:

### 🔴 Resolved — `resolve_dirty_segments_in` deleted

You were right that the doc-comment justification didn't hold up: `resolve_dirty_segments` already takes an arbitrary `toml::Table` root, so the second copy was pure duplication. Deleted, and the natural-key branch's inner-suffix resolution at the single call site now goes through `resolve_dirty_segments(table, &inner_raw)` directly. Added a comment at the call site naming why sharing (not forking) is load-bearing here — verbatim the trap this PR exists to fix, just one layer up.

The new `save_dirty_persists_mcp_server_kebab_inner_field_via_natural_key` test pins the dash-aware resolution through this exact code path: it `mark_dirty`s `mcp.servers.fs.tool-timeout-secs` and asserts the value lands on `tool_timeout_secs` on disk. If a future change to `resolve_dirty_segments` breaks dash handling, this test catches the natural-key side too — not just the top-level walker.

### 🟡 Resolved (here, not deferred) — wrong-kind bails now `WARN`

Took the "emit `WARN` on the bail paths" option rather than the larger inline-array auto-conversion. New `warn_natural_key_doc_kind_mismatch` helper shaped like the existing 0600-permissions fallback (`Action::Note` / `EventOutcome::Unknown` / `module_path!()`), called from all three sites in `ensure_array_of_tables_entry` / `find_array_of_tables_entry_mut` / `delete_array_of_tables_entry`. Attrs include `section_path`, `alias`, `node_kind` (`parent_segment` | `array_key`), `node_name`, and `operation` (`ensure` | `find` | `delete`) so a log scraper can locate the divergent node without re-deriving it.

Crucially, the warn fires **only** on wrong-kind — the "node doesn't exist" case in `find` and `delete` (the legitimate fresh-alias path) stays silent. Otherwise every fresh-alias save would spam the log.

The inline `Item::Value(Value::Array)` shape you flagged is the load-bearing case: it's valid TOML, deserializes into the identical `Config`, and so the schema validator doesn't catch it. Now an operator with that file shape gets a `WARN` event on every save, naming `mcp.servers` and the alias whose write got dropped. Converting an `Item::Value(Value::Array)` to `Item::ArrayOfTables` in place is still worth doing as a tracked follow-up; I'd rather land that with its own bug-focused tests than fold it into this review-followup commit. Happy to file the issue if you want me to.

### 🔵 Order-independence invariant documented

Added a load-bearing paragraph at the top of `apply_dirty_natural_key_path` naming the invariant: three shapes are commutative across one save batch because every branch derives from the current in-memory state and case-1 fires on mem-absence (not on default equivalence). Calls out the specific class of future changes that would break this — early-exit assuming whole-element before per-field, or doc-shape-keyed skip optimization. Verbatim the failure modes you traced under both `HashSet` orderings.

### 🔵 Section-prefix matcher documented + allocation removed

Two sub-items, both done:

1. **Constraint documented**: agreed on your read that today both sides are exact-match. Looked at `crates/zeroclaw-macros/src/lib.rs` and the in-memory `route_vec_path` arm also gates on `section_path == expected` — also exact-match. So the writer side being kebab-tolerant in isolation would *introduce* divergence (writer matches, dispatcher misses), not fix it. The new doc comment pins this: "If a future natural-key section needs an underscore in its path AND must be addressable via a kebab wire path, change **both** sides together (the macro arm and this match)." Better than silently fixing one side and shipping a new memory-vs-disk drift.

2. **Allocation removed**: swapped `format!("{}.", section.path)` + `starts_with` for the `strip_prefix(section.path).and_then(|s| s.strip_prefix('.'))` chain you suggested. As a bonus the "must have something after the dot" guard is now a natural fallout of the second `strip_prefix('.')` rather than a separate length check.

### 🔵 Param renamed + `let _` dance dropped

`default_table` → `_default_table` on the function signature, `let _ = default_table;` line gone. Kept the explanatory comment block (you specifically asked me to and you were right — that's the comment that explains why the param is there at all). Expanded it by one sentence to name why it stays on the signature: uniformity with `apply_dirty_path` at the call site in `save_dirty`.

### 🔵 Three new tests added

All three you asked for, in `crates/zeroclaw-config/src/schema.rs`:

- `save_dirty_persists_mcp_server_kebab_inner_field_via_natural_key` — dash-aware resolution through the natural-key branch (covered above in 🔴).
- `save_dirty_unset_per_field_drops_field_from_mcp_server_entry` — explicit per-field unset hits the case-1 `mem missing → delete` branch directly. The pre-existing whole-entry delete tests covered a different code path (the rename source / `delete_map_key` flow); this one exercises per-field independently so a refactor that collapsed case-1's None branch into the whole-element path would now break a test instead of silently breaking "clear this field" edits.
- `save_dirty_refuses_to_clobber_wrong_kind_mcp_servers_node` — seeds a hand-edited scalar at `mcp.servers`, runs a `mark_dirty`+`save_dirty`, asserts the hand-edit survives unchanged AND the in-memory state doesn't leak past the bail. Pins the don't-clobber guarantee whatever shape the bail takes (silent or, now, warned).

### 🔵 Dot-in-natural-key-alias rejection — deferred

Checked `create_map_key` / `rename_map_key` in `crates/zeroclaw-macros/src/lib.rs` and `handle_config_map_key_create` / `..._rename` in `crates/zeroclaw-runtime/src/rpc/dispatch.rs`; no dot-rejection anywhere. You read it right. As you suggested, deferring to a separate PR — the validation belongs at the creation/rename surface (one site in the macro arm), not bolted onto the dirty-path walker, and the right fix probably also rejects `[` `]` `=` and whitespace to keep aliases TOML-key-safe, which deserves its own scope and its own tests.

### Validation evidence

```text
$ cargo fmt --all -- --check
$ echo exit: $?
exit: 0

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 38.39s
$ echo exit: $?
exit: 0

$ cargo test
# 896 passed, 0 failed, 7 ignored across the root binary's 9 test targets.
# Same baseline as 72121e9.

$ cargo test -p zeroclaw-config --lib
test result: ok. 828 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# 825 → 828 (+3 new tests).

$ cargo test -p zeroclaw-runtime --lib
test result: ok. 2070 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
# Unchanged from base.

$ cargo test -p zeroclaw-macros
test result: ok. 0 passed; 0 failed; 2 ignored; 0 measured; 0 filtered out
# Unchanged (proc-macro doctests still gated).

$ cargo test -p zeroclaw-config --lib save_dirty
running 9 tests
test schema::tests::save_dirty_refuses_to_clobber_wrong_kind_mcp_servers_node ... ok            (new)
test schema::tests::save_dirty_persists_mcp_server_kebab_inner_field_via_natural_key ... ok    (new)
test schema::tests::save_dirty_unset_per_field_drops_field_from_mcp_server_entry ... ok        (new)
test schema::tests::save_dirty_persists_mcp_server_field_via_natural_key ... ok
test schema::tests::save_dirty_writes_new_mcp_server_added_via_create_map_key ... ok
test schema::tests::save_dirty_persists_mcp_server_rename_via_natural_key ... ok
test schema::tests::save_dirty_removes_mcp_server_deleted_via_natural_key ... ok
test schema::tests::save_dirty_drops_array_header_when_last_mcp_server_removed ... ok
test schema::tests::save_dirty_stamps_current_schema_version_on_stale_label ... ok
test result: ok. 9 passed; 0 failed
```

No wire-format or public-API changes in this commit. Hand-edited `config.toml` files remain byte-identical apart from targeted entries. The new `WARN` events are additive and shaped like every other schema warning (`Action::Note` / `EventOutcome::Unknown`), so existing log scrapers pick them up without changes.

Ready for re-review whenever you have a minute.

singlerider (approved):
Reviewed at head `668aa495`. I came to this PR fresh, did my own pass of the full diff at `72121e9` cross-checked against the base blobs, and then verified the review-response commit item by item against @JordanTheJet's changes-requested review rather than taking the response comment's word for it.

@JordanTheJet's blocker is resolved: `resolve_dirty_segments_in` is gone — the PR head carries exactly one segment resolver, and the natural-key branch's inner-suffix resolution calls the shared `resolve_dirty_segments` at the single call site. The new kebab test (`mcp.servers.fs.tool-timeout-secs` → `tool_timeout_secs` on disk) pins dash-aware resolution through the natural-key branch specifically, so a future dash-handling change that misses either walk now breaks a test instead of shipping a divergence. Their 🟡 and every 🔵 are either addressed in the commit or deferred with a concrete scope; I verified each:

- The wrong-kind bails now emit `WARN` through `warn_natural_key_doc_kind_mismatch`, shaped identically to the existing schema warns (`Action::Note` / `EventOutcome::Unknown` / `module_path!()` — same pattern as the proxy-URL and 0600-permissions warns already in `schema.rs`). All six bail sites are covered, and node-missing (the legitimate fresh-alias path) correctly stays silent while wrong-kind warns.
- The exact-match section-prefix constraint is the right call, not a dodge: I confirmed the in-memory macro arm gates on `section_path == outer_expected` (`crates/zeroclaw-macros/src/lib.rs:778,818,842`), so making the writer side kebab-tolerant in isolation would have introduced a dispatcher-vs-writer disagreement — the exact divergence class this PR eliminates. The change-both-sides doc comment pins it.
- All three requested tests are present and test real branches: kebab inner field, per-field unset driving the case-1 mem-absence delete directly, and the wrong-kind don't-clobber contract.
- Dot-in-alias rejection deferred to the create/rename surface in a separate PR — correct layering; bolting validation onto the dirty-path walker would have been the wrong home.

### 🟢 The fix attacks the root cause, not the symptom

Surfacing `#[natural_key]` through `MapKeySection` means the writer asks the schema registry "is this a natural-key Vec?" instead of hardcoding `mcp.servers` — the next `#[natural_key]` section gets correct persistence for free, and the `map_key_sections_exposes_natural_key_for_mcp_servers` contract test makes the derive output a load-bearing API rather than an implementation detail.

### 🟢 Test design worth copying

Seeding a real on-disk file so `save_dirty` takes its incremental path (the full-`save` fallback would mask the bug), replicating the exact dispatcher wire sequences, asserting raw bytes AND reparse round-trips, and the end-to-end RPC test adding the on-disk assertion whose absence is what let the original bug ship. The review-response commit holds the same standard: every finding that could regress got a test pinning it.

### 🟢 The response commit is a model review response

Each finding addressed or explicitly deferred with named scope and rationale, independent verification of the reviewer's claims (the macro-arm exact-match check), and fresh validation evidence per crate. Nothing hand-waved.

@JordanTheJet's changes-requested review is still formally outstanding; I've verified every item in it is resolved at this head, and I'm approving on that basis — their re-review stands on its own whenever they get to it.

Approving at `668aa495`.

WareWolf-MoonWall (commented):
Reviewed at head 72121e9 against @JordanTheJet's CHANGES_REQUESTED from earlier today. CI is all green, the template is complete, and the test design is excellent. One thing to add before the 🔴 is addressed: I confirmed @JordanTheJet's blocking finding independently.

### ✅ Resolved — nothing prior to resolve

No earlier reviews existed beyond JordanTheJet's withdrawn comment (explicitly superseded by the CHANGES_REQUESTED review). Clean slate.

### 🟢 What looks good — on-disk regression coverage

The decision to seed a real file and drive the exact `mark_dirty` / `set_prop_persistent` / `save_dirty` sequence the dispatcher emits — rather than mocking at the schema layer — is the right call. It's the only test approach that would have caught the original bug. The `map_key_sections_exposes_natural_key_for_mcp_servers` contract test is especially good: it pins the derive output so a future macro change can't silently regress the writer back to the broken table-only walk. The end-to-end RPC test (`config_set_persists_mcp_server_field_to_disk`) adds the on-disk assertion the pre-existing `config_set_*` tests were missing — that omission is exactly what let this bug ship.

### 🔴 Blocking — resolve_dirty_segments_in is a verbatim copy of resolve_dirty_segments (concurring with @JordanTheJet)

I diffed the two functions and @JordanTheJet's finding is correct: `resolve_dirty_segments_in` is token-for-token identical to the pre-existing `resolve_dirty_segments(root: &toml::Table, raw: &[&str]) -> Vec<String>`. The new function's doc comment claims a different root scope as justification, but the existing function already accepts an arbitrary table root — so the stated reason doesn't hold.

This matters beyond code hygiene: the root cause of this PR is two parallel implementations of one walk (in-memory vs on-disk) drifting apart. Introducing a third copy of the segment resolver recreates the same hazard. Per AGENTS.md, duplicate state — including duplicate implementations of the same walk — is banned precisely because of the drift bugs they breed. The fix is one line: delete `resolve_dirty_segments_in` and call `resolve_dirty_segments(table, &inner_raw)` at its single call site.

Holding with @JordanTheJet until that copy is gone.

Issue #4556, opened 2026-03-24
Release binary arm-unknown-linux-gnueabihf is actually ARMv7+NEON, segfaults on ARMv6


bennyzen:
## Summary

The `zeroclaw-arm-unknown-linux-gnueabihf.tar.gz` asset in [v0.6.1](https://github.com/zeroclaw-labs/zeroclaw/releases/tag/v0.6.1) is built for ARMv7 with NEON, not ARMv6 as the target triple implies. It segfaults on Raspberry Pi Zero W (ARM1176JZF-S / ARMv6).

## Reproduction

```bash
# On a Raspberry Pi Zero W v1.1 (ARMv6, Raspberry Pi OS Bookworm)
curl -fsSL https://github.com/zeroclaw-labs/zeroclaw/releases/download/v0.6.1/zeroclaw-arm-unknown-linux-gnueabihf.tar.gz | tar xz
./zeroclaw --version
# => Segmentation fault
```

## Evidence

```
$ readelf -A ./zeroclaw | head -8
Attribute Section: aeabi
File Attributes
  Tag_CPU_name: "7-A"
  Tag_CPU_arch: v7
  Tag_CPU_arch_profile: Application
  Tag_ARM_ISA_use: Yes
  Tag_THUMB_ISA_use: Thumb-2
  Tag_FP_arch: VFPv3
  Tag_Advanced_SIMD_arch: NEONv1
```

The binary requires ARMv7 instructions and NEON SIMD, neither of which are available on the Pi Zero's ARM1176JZF-S core.

## Root cause

The release CI (`release-stable-manual.yml`) cross-compiles with Ubuntu 22.04's `gcc-arm-linux-gnueabihf`, which defaults to ARMv7 codegen regardless of the Rust target triple. The Rust target `arm-unknown-linux-gnueabihf` should produce ARMv6-compatible code, but the system linker/gcc overrides this.

## Suggested fix

Either:
1. Pass `-march=armv6` / `-mfpu=vfp` to the cross-compiler via `CFLAGS` in CI
2. Use a toolchain that actually targets ARMv6 (e.g., [musl.cc `arm-linux-musleabihf-cross`](https://musl.cc/))
3. Rename the asset to `armv7-unknown-linux-gnueabihf` to match what it actually is (though there's already a separate `armv7` build)

## Workaround

Cross-compile from source using the musl.cc toolchain:

```bash
[...]

PR #7520, opened 2026-06-11
fix(ci): install cross g++ for ARM glibc release builds

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The v0.8.0 Release Stable run ([27382533280](https://github.com/zeroclaw-labs/zeroclaw/actions/runs/27382533280)) failed on all three ARM glibc targets (`aarch64-unknown-linux-gnu`, `armv7-unknown-linux-gnueabihf`, `arm-unknown-linux-gnueabihf`) with `error: failed to run custom build command for inkjet v0.11.1 — failed to find tool "aarch64-linux-gnu-g++": No such file or directory`.
  - Root cause: the release workflow builds zerocode alongside the main binary, and zerocode depends on `inkjet`, whose build script compiles tree-sitter grammars containing C++ scanners (`cc-rs` with `cpp(true)`). The build matrix only installed the C cross-compilers (`gcc-<triple>`), so cc-rs could not find the derived `<triple>-g++`. The main `zeroclaw` binary has no C++ build deps, which is why these targets built fine before zerocode joined the release archive.
  - Fix: add the matching `g++-<triple>` package to each of the three matrix `cross_compiler` entries. The install step (`apt-get install -y ${{ matrix.cross_compiler }}`) word-splits, so no other plumbing changes.
  - Also mirror the ARMv6 `CFLAGS` into `CXXFLAGS_arm_unknown_linux_gnueabihf` so inkjet's C++ scanner objects get ARMv6 codegen too — otherwise they'd compile at the toolchain's ARMv7+NEON default and reintroduce the #4556 Pi Zero W segfault class through the C++ side.
  - `cross-platform-build-manual.yml` carries the identical latent bug (same gcc-only matrix, same zerocode build step), so it gets the same four changes.
- **Scope boundary:** Does NOT touch the MUSL targets (cross-rs containers ship g++), Android (NDK), or the native Apple/Windows builds — none of them failed and none use the apt cross-toolchain path. No Rust code, no Cargo manifests, no other workflows.
- **Blast radius:** Only the two manual workflows (`release-stable-manual.yml`, `cross-platform-build-manual.yml`), and within them only the three ARM glibc matrix entries. Worst case is unchanged behavior on the already-failing jobs.
- **Linked issue(s):** Related #4556 (the ARMv6 codegen forcing this PR extends to CXXFLAGS).
- **Labels:** expect `ci` from the path labeler; suggest `bug`, `risk: low`, `size: XS`.

Workflow-only diff — the Rust battery does not exercise `.github/workflows/*.yml`, so it was intentionally skipped (no Rust, docs, or bootstrap-script changes in the diff). Validation performed instead:

```text
$ python3 -c "import yaml; [yaml.safe_load(open(f)) for f in ['.github/workflows/release-stable-manual.yml','.github/workflows/cross-platform-build-manual.yml']]" && echo YAML OK
YAML OK
```

  - Read the failed job logs for all three targets: identical `ToolNotFound: aarch64-linux-gnu-g++` / `arm-linux-gnueabihf-g++` failure inside inkjet's build script, during the zerocode build step (the main binary built clean first).
  - Confirmed `inkjet v0.11.1`'s build script compiles `scanner.cc` files with `cc::Build::cpp(true)` (registry source, `build/language.rs`), which makes cc-rs derive `<triple>-g++` from the target.
  - Confirmed Ubuntu 22.04/24.04 ship `g++-aarch64-linux-gnu` and `g++-arm-linux-gnueabihf` in main; the workflow's existing `apt-get install` line handles multi-package values via word-splitting.
  - NOT verified: an actual green run of the release workflow on this branch — it is `workflow_dispatch` on master with a version gate, so the proof is the next dispatch after merge.
- **If any command was intentionally skipped, why:** `cargo fmt`/`clippy`/`test` skipped — zero Rust files in the diff.

- New external network calls? `No` — same apt archive, two more packages from it.

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` — CI-internal env vars only.

Low-risk: `git revert <sha>`. Failure symptom if this is somehow wrong: the same three release jobs fail at the `Install cross compiler` step (bad package name) instead of at inkjet — immediately visible in the next dispatch.

### DIFF
--- a/.github/workflows/cross-platform-build-manual.yml
+++ b/.github/workflows/cross-platform-build-manual.yml
@@ -57,18 +57,18 @@ jobs:
             # CI gate (`ci.yml`) uses clang+mold for throughput optimization.
           - os: ubuntu-latest
             target: aarch64-unknown-linux-gnu
-            cross_compiler: gcc-aarch64-linux-gnu
+            cross_compiler: gcc-aarch64-linux-gnu g++-aarch64-linux-gnu
             linker_env: CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER
             linker: aarch64-linux-gnu-gcc
           - os: ubuntu-latest
             target: armv7-unknown-linux-gnueabihf
-            cross_compiler: gcc-arm-linux-gnueabihf
+            cross_compiler: gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf
             linker_env: CARGO_TARGET_ARMV7_UNKNOWN_LINUX_GNUEABIHF_LINKER
             linker: arm-linux-gnueabihf-gcc
             skip_prometheus: true
           - os: ubuntu-latest
             target: arm-unknown-linux-gnueabihf
-            cross_compiler: gcc-arm-linux-gnueabihf
+            cross_compiler: gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf
             linker_env: CARGO_TARGET_ARM_UNKNOWN_LINUX_GNUEABIHF_LINKER
             linker: arm-linux-gnueabihf-gcc
             skip_prometheus: true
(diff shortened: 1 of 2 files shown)


tidux (approved):
Manual approval.  I've validated that those are the correct packages missing.

Issue #7521, opened 2026-06-11
[Feature]: file_read — decode non-UTF-8 text (cp1251/Latin-1) via charset detection


metalmon:
### Summary

`file_read` (default `utf8` mode) cannot return correct text for files in a non-UTF-8 text encoding (e.g. Windows-1251 / Cyrillic, Latin-1, Shift-JIS). Such files are not valid UTF-8, so the tool falls back to `String::from_utf8_lossy`, replacing every non-ASCII byte with `U+FFFD` — the agent receives mojibake (`������`), not readable text.

This is a follow-up to the `file_read` binary-rejection slice (recovery under #6074): that change rejects *confident* binary (image magic / NUL / control-byte glut) but **deliberately leaves non-UTF-8 text on the lenient lossy path** so legacy-encoded text is not wrongly rejected as binary. Actually decoding that text is tracked here.

### Motivation

- Agents operating on real-world workspaces (especially non-English ones) routinely encounter cp1251/Latin-1 text files. Today the only outcomes are lossy garbage (`utf8`) or base64 bytes the model must decode itself.
- `from_utf8_lossy` silently corrupts data without signalling it — the agent may "read" a config/source file and act on garbage.

### Proposed solution

Add charset detection + transcoding on the non-UTF-8 text path, before the lossy fallback:

1. On invalid-UTF-8 bytes that are **not** classified as binary, run a charset detector and transcode to UTF-8.
2. Lightweight, pure-Rust, Mozilla-maintained crates fit well and avoid a heavy ML dependency:
   - **`encoding_rs`** — the transcoder (cp1251/Latin-1/… → UTF-8). Already present transitively in the dependency tree.
   - **`chardetng`** — the charset detector (the engine Firefox uses).
3. Keep the layering from the binary-rejection slice intact: `valid UTF-8 → text` · `image magic / NUL / control glut → reject` · `else detect+transcode → text` · `detect fails → reject with base64 guidance`.

Open design questions for discussion:
- **Auto-detect vs explicit:** auto-detect by default, or add an explicit `encoding` value (e.g. `encoding: "cp1251"`) alongside the existing `utf8`/`base64`? (Auto-detect is more ergonomic; explicit is deterministic.)
- Which encodings to support out of the box (whatever `encoding_rs` covers, gated by detector confidence).
- Confidence threshold / behavior when the detector is unsure (fall back to lossy, or reject with guidance).

### Scope / non-goals

[...]

PR #7522, opened 2026-06-11
fix(tools): reject binary/image files in default file_read instead of lossy text

metalmon:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - In default `utf8` mode, `file_read` returned `String::from_utf8_lossy` for non-UTF-8 content — handing the agent mojibake (`U+FFFD`) for binary and image files. This restores the binary-rejection behavior lost in the #4320 bulk revert (recovery tracked in **#6074**), adapted to the post-#7004 base64 read path.
  - Known image containers (PNG/JPEG/GIF/WEBP/BMP) → rejected with guidance to use the `image_info` tool. Other confident binary (a NUL byte, or a glut of control characters over the first 8 KiB) → rejected with guidance to use `encoding="base64"`.
- **Scope boundary:** `encoding="base64"` (explicit binary path) and PDF text extraction are unchanged. Detection is deliberately conservative — non-UTF-8 **text** (no NUL, no control glut, no image magic, e.g. Windows-1251) is NOT classified as binary and still reads leniently, to avoid false-positives on legacy text encodings. No new dependencies (magic check + heuristic are local helpers).
- **Blast radius:** `file_read` (`zeroclaw-runtime`) only. Behavior change: reading a binary/image file in default mode now returns a clear error instead of lossy text. UTF-8 text reads and base64 reads are byte-for-byte unchanged.
- **Linked issue(s):** Related #6074 (recovery of the reverted #4320 behavior). Builds on #7004 (base64 read path). Follow-up for proper non-UTF-8 *text* decoding: #7521.
- **Labels:** suggest `type: fix`, `risk: low`, `size: S`, `tool`, `tool:file` (maintainer-set).

Scoped to the changed crate (`zeroclaw-runtime`).

```text
$ cargo fmt --all -- --check
clean (exit 0)

$ cargo test -p zeroclaw-runtime --lib file_read
test tools::file_read::tests::file_read_rejects_binary_file ... ok
test tools::file_read::tests::file_read_rejects_png_image ... ok
test tools::file_read::tests::file_read_rejects_jpeg_image ... ok
test tools::file_read::tests::file_read_reads_non_utf8_text_lossy ... ok
test tools::file_read::tests::e2e_agent_file_read_rejects_binary ... ok
test result: ok. 33 passed; 0 failed; 1 ignored
# (the 1 ignored is the live-OpenAI Codex e2e, gated on credentials)

clean — no warnings (exit 0)
```

- **Beyond CI — what did you manually verify?** The new unit + e2e tests cover the behavior end-to-end through the agent tool path: NUL-byte binary and PNG/JPEG magic are rejected (no `U+FFFD` in the tool result), while a Windows-1251 byte sequence (`Привет`) is NOT rejected. PDF extraction and base64 reads retain their existing dedicated tests.
- **What was NOT verified:** no real on-disk image/office fixtures beyond synthetic magic-byte headers; charset *decoding* of legacy text is explicitly out of scope (follow-up issue).

- New permissions, capabilities, or file system access scope? **No** — this restricts existing read behavior (rejects binary/image instead of returning lossy text). No new tool, host function, or path scope.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No.** Tests use synthetic byte arrays.

## Compatibility (required)

- Backward compatible? **Behavior change, low risk.** A caller that previously relied on lossy binary reads in default mode now gets a clear error directing it to `encoding="base64"` (or `image_info`). UTF-8 text reads and base64 reads are unchanged.
- Upgrade steps for existing users: none; agents that need raw bytes should pass `encoding="base64"` (already the documented path).

Low risk. Plan: `git revert <merge sha>`. The change is isolated to `crates/zeroclaw-runtime/src/tools/file_read.rs`.

### DIFF
--- a/crates/zeroclaw-runtime/src/tools/file_read.rs
+++ b/crates/zeroclaw-runtime/src/tools/file_read.rs
@@ -78,7 +78,7 @@ impl Tool for FileReadTool {
     }
 
     fn description(&self) -> &str {
-        "Read file contents with line numbers. Supports partial reading via offset and limit. Extracts text from PDF; other binary files are read with lossy UTF-8 conversion. Set encoding=\"base64\" to return raw bytes base64-encoded (for binary files such as .xlsx/.docx); offset/limit are ignored in that mode."
+        "Read file contents with line numbers. Supports partial reading via offset and limit. Extracts text from PDF; binary and image files are rejected (use the image_info tool for images). Set encoding=\"base64\" to return raw bytes base64-encoded (for binary files such as .xlsx/.docx); offset/limit are ignored in that mode."
     }
 
     fn parameters_schema(&self) -> serde_json::Value {
@@ -328,7 +328,36 @@ impl FileReadTool {
                     });
                 }
 
-                // Lossy fallback — replaces invalid bytes with U+FFFD
+                // Reject confident binary instead of returning lossy garbage.
+                // Known image formats: point the agent at the image_info tool.
+                if let Some(kind) = detect_image_format(&bytes) {
+                    return Ok(ToolResult {
+                        success: false,
+                        output: String::new(),
+                        error: Some(format!(
+                            "Binary image file detected ({kind}): {}. Use the image_info \
+                             tool for images, or encoding=\"base64\" to read the raw bytes.",
+                            resolved_path.display()
+                        )),
+                    });
+                }
+
+                // Other confident binary (NUL byte or a glut of control bytes).
+                if looks_binary(&bytes) {
+                    return Ok(ToolResult {
+                        success: false,
+                        output: String::new(),
+                        error: Some(format!(
+                            "Binary file detected: {}. Use encoding=\"base64\" to read the \
+                             raw bytes.",
+                            resolved_path.display()
+                        )),
+                    });
+                }
+
+                // Not confidently binary — most likely text in a non-UTF-8 encoding
+                // (e.g. Windows-1251, Latin-1). Decode leniently for now; proper
+                // charset detection/transcoding is tracked as a follow-up.
                 let lossy = String::from_utf8_lossy(&bytes).into_owned();
                 Ok(ToolResult {
                     success: true,
@@ -355,6 +384,70 @@ fn try_extract_pdf_text(bytes: &[u8]) -> Option<String> {
 #[cfg(not(feature = "rag-pdf"))]
 fn try_extract_pdf_text(_bytes: &[u8]) -> Option<String> {
     None
+}
+
+/// Detect a common raster-image container by its file-header magic bytes.
+/// Returns the format name when recognized so `file_read` can reject images
+/// with guidance to use the `image_info` tool instead of emitting lossy text.
+/// Only consulted on the non-UTF-8 read path, so an ASCII string that merely
+/// starts with one of these markers (and is therefore valid UTF-8) is unaffected.
+///
+/// PNG/JPEG/GIF magics carry non-ASCII/control bytes and are collision-free, and
+/// WEBP is anchored by the `RIFF…WEBP` container, so the raw magic is enough. The
+/// BMP marker is just the two printable ASCII letters `BM`, which a non-UTF-8
+/// legacy-text file can legitimately start with (a name, "BMW dealer notes", …),
+/// so it is validated against the rest of the BITMAPFILEHEADER instead of trusted
+/// on the magic alone — otherwise the legacy-text carve-out below is defeated.
+fn detect_image_format(bytes: &[u8]) -> Option<&'static str> {
+    if bytes.starts_with(b"\x89PNG\r\n\x1a\n") {
+        Some("png")
+    } else if bytes.starts_with(&[0xFF, 0xD8, 0xFF]) {
+        Some("jpeg")
+    } else if bytes.starts_with(b"GIF87a") || bytes.starts_with(b"GIF89a") {
+        Some("gif")
+    } else if bytes.len() >= 12 && bytes.starts_with(b"RIFF") && &bytes[8..12] == b"WEBP" {
+        Some("webp")
+    } else if is_bmp_header(bytes) {
+        Some("bmp")
+    } else {
+        None
+    }
+}
+
+/// Validate a BMP `BITMAPFILEHEADER` beyond the weak `BM` magic. Real BMPs set
+/// the two reserved words (offset 6..10) to zero and point `bfOffBits`
+/// (offset 10..14, the pixel-array offset) inside the file. Non-UTF-8 text that
+/// merely starts with `BM` carries printable bytes in the reserved field, so it
+/// fails this check and falls through to the lenient lossy read. `bfSize`
+/// (offset 2..6) is deliberately not checked: some encoders write 0 there.
+fn is_bmp_header(bytes: &[u8]) -> bool {
+    if bytes.len() < 14 || !bytes.starts_with(b"BM") {
+        return false;
+    }
+    if bytes[6..10] != [0, 0, 0, 0] {
+        return false;
+    }
+    let off_bits = u32::from_le_bytes([bytes[10], bytes[11], bytes[12], bytes[13]]);
+    (14..=bytes.len() as u32).contains(&off_bits)
+}
+
+/// Heuristic binary classifier for the non-UTF-8 read path. A NUL byte (which
+/// text essentially never contains) or a high density of non-text control
+/// characters marks the content as binary. Legacy single-byte text encodings
+/// (e.g. cp1251, Latin-1) have neither, so they are deliberately NOT classified
+/// as binary here — they fall through to the lenient lossy read.
+fn looks_binary(bytes: &[u8]) -> bool {
+    // Sample a prefix so very large files stay cheap.
+    let sample = &bytes[..bytes.len().min(8192)];
+    if sample.is_empty() {
+        return false;
+    }
+    if sample.contains(&0) {
+        return true;
+    }
+    let is_control = |b: u8| b < 0x20 && b != b'\t' && b != b'\n' && b != b'\r';
+    let controls = sample.iter().filter(|&&b| is_control(b)).count();
+    controls * 100 / sample.len() > 30
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


crates/zeroclaw-runtime/src/tools/file_read.rs
```diff
@@ -357,6 +386,46 @@ fn try_extract_pdf_text(_bytes: &[u8]) -> Option<String> {
...
+fn detect_image_format(bytes: &[u8]) -> Option<&'static str> {
+    if bytes.starts_with(b"\x89PNG\r\n\x1a\n") {
+        Some("png")
+    } else if bytes.starts_with(&[0xFF, 0xD8, 0xFF]) {
+        Some("jpeg")
+    } else if bytes.starts_with(b"GIF87a") || bytes.starts_with(b"GIF89a") {
+        Some("gif")
+    } else if bytes.len() >= 12 && bytes.starts_with(b"RIFF") && &bytes[8..12] == b"WEBP" {
+        Some("webp")
+    } else if bytes.starts_with(b"BM") {
```
singlerider: `BM` is two printable ASCII bytes, unlike the other four magics which all contain non-ASCII or control bytes and so cannot collide with legacy text. A non-UTF-8 (cp1251/Latin-1) file that merely starts with the letters "BM" reaches this arm and gets rejected as "Binary image file detected (bmp)", pointing the agent at `image_info`, which then fails on it. That is the exact false positive `looks_binary` was built to avoid, reintroduced through the image branch.

Two fixes, your pick:

1. Validate the BMP header: `bytes.len() >= 14` and the little-endian u32 at offset 2 (file size) bounds `bytes.len()`. Real BMPs satisfy this; text never will.
2. Drop BMP. Undetected BMPs fall through to `looks_binary`, which rejects them anyway via pixel-data NUL/control density.

Either way add a regression test next to `file_read_reads_non_utf8_text_lossy`: non-UTF-8 text starting with `BM` must read leniently.

singlerider (changes requested):
This recovers the #6074-tracked behavior in exactly the right shape for the post-#7004 world: rejection only on the non-UTF-8 path, images steered to `image_info`, other binary steered to `encoding="base64"`, and legacy single-byte text deliberately falling through to the lenient read with the follow-up (#7521) already filed. The test rewrite is honest, including flipping the e2e assertion from "must contain U+FFFD" to "must not". One blocking item.

### 🔴 Blocking: the `BM` magic check defeats the legacy-text carve-out you built

PNG/JPEG/GIF/WEBP magics are 3 to 8 bytes of binary-ish content and effectively collision-free (they all contain non-ASCII or control bytes). `BM` is two printable ASCII bytes. The code comment says ASCII content starting with a marker is unaffected because it is valid UTF-8, which is true, but the case this PR works hardest to protect is non-UTF-8 legacy text, and a cp1251/Latin-1 file starting with the letters "BM" (a name, an abbreviation, "BMW dealer notes") hits this arm and gets rejected as a "Binary image file detected (bmp)" with guidance to a tool that will fail on it. That is precisely the false positive the `looks_binary` heuristic was designed to avoid, reintroduced through the image branch.

Two acceptable fixes, your pick:

1. Validate the BMP header beyond the magic: `bytes.len() >= 14` and the u32 at offset 2 (file size, little-endian) equals or plausibly bounds `bytes.len()`. Real BMPs always satisfy this; text never will.
2. Drop BMP from the list. It is the rarest of the five in agent workspaces, and an undetected BMP just falls through to `looks_binary`, which still rejects it via the pixel-data NUL/control density in any non-trivial BMP.

Either way, add a regression test: non-UTF-8 legacy text starting with `BM` must read leniently, sitting next to your existing `file_read_reads_non_utf8_text_lossy`.

### 🟢 Rejection scoped to the right layer

Both rejections live strictly on the non-UTF-8 fallback path, after the UTF-8 fast path and PDF extraction, so valid text and the explicit base64 path are byte-for-byte untouched. The 8 KiB sampling in `looks_binary` keeps huge files cheap, and the NUL-or-control-glut rule matches the standard classifier shape.

### 🟢 Error messages teach the recovery path

Each rejection names the tool that actually works for that content (`image_info` for images, `encoding="base64"` for the rest). The agent gets a self-correcting result instead of mojibake it might paste into a reply.

### 🟢 Test conversion is thorough

Old lossy-path tests were converted rather than deleted, the e2e test asserts the tool result content the model actually sees, and the cp1251 carve-out test pins the deliberate non-goal.

singlerider:
Still at head `0672c553` — no new commits since my 2026-06-13 review, so my changes-requested remains live and unaddressed. Confirming, not re-litigating.

The blocker stands: `file_read.rs:403` still does `else if bytes.starts_with(b"BM") { Some("bmp") }` with no validation beyond the two-byte magic. `BM` is two printable ASCII bytes, so a non-UTF-8 legacy text file starting with the letters "BM" (a name, "BMW dealer notes", etc.) hits this arm and gets rejected as "Binary image file detected (bmp)" — the exact non-UTF-8-text false positive the rest of this PR works hardest to protect against, reintroduced through the image branch. Either validate the BMP header (`bytes.len() >= 14` and the little-endian u32 at offset 2 plausibly bounds `bytes.len()`) or drop BMP from the magic list (it falls through to `looks_binary`, which still catches real BMPs via pixel-data NUL/control density). Add a regression: non-UTF-8 legacy text starting with `BM` must read leniently, beside the existing `file_read_reads_non_utf8_text_lossy`.

The rest is in good shape — rejection scoped to the non-UTF-8 fallback path only, recovery-teaching error messages, honest test conversion. Fix the BMP false positive and I'll approve.

metalmon:
Fixed. Went with option 1 (validate the header), with one deviation I want to flag.

`detect_image_format` no longer trusts the bare `BM` magic — BMP now goes through a new `is_bmp_header` that checks the `BITMAPFILEHEADER`: `len >= 14`, the two reserved words at offset 6..10 are zero, and `bfOffBits` (offset 10..14) points inside the file. Non-UTF-8 text starting with `BM` carries printable bytes in the reserved field, so it fails the check and falls through to the lenient lossy read.

Deviation from your suggestion: I validate the reserved field instead of `bfSize` (offset 2..6). `bfSize` is a weaker discriminator in practice — some encoders write 0 there, which would make a strict size check reject valid BMPs. The reserved-words-zero invariant holds for real BMPs and is never satisfied by printable text, so it is both tighter and safer.

Regression tests added next to `file_read_reads_non_utf8_text_lossy`:
- `file_read_reads_non_utf8_bm_text_lossy` — cp1251 text starting with `BM` reads leniently.
- `file_read_rejects_bmp_image` — a real BMP (valid header + a `0xFF` pixel byte so the file is non-UTF-8 and reaches the detector) is still rejected and steered to `image_info`.

Also rebased onto current master (clean — master has not touched this file).

`cargo test -p zeroclaw-runtime --lib file_read`: the binary/image/text suite passes, including both new tests. (One unrelated pre-existing failure, `file_read_nonexistent_consumes_rate_limit_budget`, reproduces on master without this change — not touched here.)

Audacity88 (commented):
I reviewed current head `fb395acf3a660f35a49a73b8a9de6d1625f12af5`, the author follow-up, singlerider's active BMP-blocking review, the current `file_read.rs` diff, the linked #6074/#7521 scope split, CI, and merge freshness. I did not run a local cargo pass.

### ✅ Resolved — BMP magic no longer defeats the legacy-text carve-out

The new `is_bmp_header()` no longer trusts the two-byte `BM` prefix by itself. It requires a 14-byte BMP file header, zero reserved words, and a `bfOffBits` value that points inside the file. That preserves real BMP rejection while allowing the non-UTF-8 `BM...` text case to fall through to the lenient text path, and the new `file_read_reads_non_utf8_bm_text_lossy` / `file_read_rejects_bmp_image` tests cover both sides of singlerider's requested regression.

### 🟢 What looks good — binary rejection stays on the intended path

The rejection checks still happen only after the UTF-8 fast path and PDF extraction, and after the explicit `encoding="base64"` branch has already returned raw bytes. The split with #7521 also stays clean: this PR decides when content is confidently binary or image data; charset decoding for probably-text remains the follow-up.

### 🟡 Remaining process state — not merge-ready yet

All visible CI checks are green, but the branch is currently 7 commits behind `master`; the merge-freshness helper reports no overlapping files but still requires a freshness gate because this touches the risky `file_read` runtime surface. Since singlerider's `CHANGES_REQUESTED` review is also still the active blocking review, this is not a merge-ready statement until that review is updated or otherwise resolved.

No additional blockers from me on the current diff.

metalmon:
Rebuilt the branch: rebased onto current `master` (`140d9d1e40`) as a single linear commit, dropping the earlier `Update branch` merge commit. This resolves the freshness-gate / "behind master" point from the last review — `master` does not touch `file_read.rs` in the gap, so the rebase is clean and the `file_read.rs` diff is byte-for-byte the same change reviewed at `fb395acf`, just on a fresh base. CI is re-running on the rebased head.

@singlerider — your BMP blocker is addressed (and @Audacity88 confirmed it resolved in his follow-up). `detect_image_format` no longer trusts the bare `BM` magic: BMP now goes through `is_bmp_header`, which validates the `BITMAPFILEHEADER` — `len >= 14`, the two reserved words at offset 6..10 are zero, and `bfOffBits` (offset 10..14) points inside the file. Non-UTF-8 `BM…` text carries printable bytes in the reserved field, so it fails the check and falls through to the lenient lossy read.

The two regression tests you asked for sit beside `file_read_reads_non_utf8_text_lossy`:
- `file_read_reads_non_utf8_bm_text_lossy` — cp1251 text starting with `BM` reads leniently.
- `file_read_rejects_bmp_image` — a real BMP (valid header) is still rejected and steered to `image_info`.

Could you re-review when you have a moment? Happy to adjust if anything else stands out.

singlerider (approved):
Re-review at head 75f5d1a8 against current master and my prior CHANGES_REQUESTED. Checked out the head locally and ran the file_read suite.

### RESOLVED ✅ The BMP magic no longer defeats the legacy-text carve-out

My blocker was that the two printable ASCII bytes `BM` would reject non-UTF-8 legacy text (a name, "BMW dealer notes") as a binary image. Resolved via option 1, header validation:

```rust
fn is_bmp_header(bytes: &[u8]) -> bool {
    if bytes.len() < 14 || !bytes.starts_with(b"BM") { return false; }
    if bytes[6..10] != [0, 0, 0, 0] { return false; }
    let off_bits = u32::from_le_bytes([bytes[10], bytes[11], bytes[12], bytes[13]]);
    (14..=bytes.len() as u32).contains(&off_bits)
}
```

A real BMP satisfies the 14-byte minimum, the zeroed reserved field, and a `bfOffBits` that bounds the file. Legacy text starting with `BM` carries printable bytes in the reserved field (offset 6..10) and fails the check, so it falls through to the lenient read. The code comment cites the exact carve-out case.

Both regression tests I asked for landed and pass locally, sitting next to the existing cp1251 carve-out test:

```
test tools::file_read::tests::file_read_reads_non_utf8_bm_text_lossy ... ok
test tools::file_read::tests::file_read_rejects_bmp_image ... ok
```

### 🟢 The rest still holds

Rejection stays strictly on the non-UTF-8 fallback path after the UTF-8 fast path and PDF extraction, the error messages name the recovering tool (`image_info`, `encoding="base64"`), and the PNG/JPEG/GIF detection plus the `looks_binary` density rule remain as reviewed. Full suite: 37 passed, 1 ignored.

Approving.

PR #7524, opened 2026-06-12
feat(channels/discord): derive gateway intents from config instead of hardcoding

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The Discord IDENTIFY payload's hardcoded `intents: 37377` becomes `gateway_intents()`: a named `BASELINE_INTENTS` constant (GUILDS | GUILD_MESSAGES | DIRECT_MESSAGES | MESSAGE_CONTENT) plus an optional operator override. A regression test pins the default mask to exactly 37377, so default-config behavior is unchanged.
  - New optional `channels.discord.<alias>.intents_mask` (`Option<u64>`, default unset = derive) sends an exact operator-supplied mask verbatim, including the legal `Some(0)`. This is an escape hatch for downstream deployments that consume gateway events through custom builds; a WARN names any baseline bits the override drops, since that silences message handling.
  - Close code 4014 (disallowed intents) now names the privileged Developer Portal toggles implied by the requested mask instead of a bare code, with a raw-mask fallback when no privileged bits are involved.
  - The Identify log line records the resolved mask and decoded intent names; unknown bits are reported as a hex remainder rather than dropped.
  - This PR deliberately adds no privileged opt-in keys (`guild_members`, `presence`): they have no in-tree event consumer yet and will arrive with the features that consume them. The mask snapshot is connection-scoped by design: intents are sent once in IDENTIFY, and config reloads rebuild channels, which re-derives it.
- **Scope boundary:** no new event handling, no behavior change without explicit config. Resume (op 6) is untouched (it carries no intents).
- **Blast radius:** `DiscordChannel` IDENTIFY/diagnostics paths and one config field. Exhaustive `DiscordConfig` test literals in zeroclaw-runtime and the root crate gained the field.
- **Linked issue(s):** None. Related #7489 (the slash-command PR builds on the same channel; the branches are independent).
- **Labels:** to be set by maintainers (`enhancement`, `channel`; suggest `size: S`, `risk: low`).

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 15s   # zero warnings
passed: 896 failed: 0
test result: ok. 863 passed; 0 failed; 1 ignored
$ cargo test -p xtask directives_emit_no_prose_em_dashes   # generated-docs prose gate
```

- **Commands run and tail output:** above; `channels-full` is used because `channel-discord` is not in default features and CI covers this module via `ci-all`. The mdbook generated-prose gate (`directives_emit_no_prose_em_dashes`) was run explicitly since the new schema doc comments flow into the generated channel page.
- **Beyond CI — what did you manually verify?** A build containing this change has run a production Discord daemon for two days: the Identify log shows the resolved mask + decoded names on every connect, and the default mask was confirmed byte-identical to the previous hardcoded behavior (four intents, no gateway behavior change). NOT manually exercised: a non-default `intents_mask` against live Discord (covered by unit tests on precedence and the 4014 hint construction).

- New external network calls? **No** (same gateway connection; only the IDENTIFY payload's already-sent integer is derived differently)
- If any `Yes`, describe the risk and mitigation: n/a. Worth noting: `intents_mask` lets an operator request privileged intents; those still require the matching Developer Portal toggles, and the 4014 diagnostic names them on mismatch.

## Compatibility (required)

- Backward compatible? **Yes** (default mask provably identical to the previous literal)
- Config / env / CLI surface changed? **Yes** — one new optional field, `intents_mask`, default unset. No upgrade steps; existing configs unaffected.

Low risk: `git revert <sha>`; runtime-level rollback is simply not setting `intents_mask`.


WareWolf-MoonWall (approved):
First review. +246/-6 across `crates/zeroclaw-channels/src/discord.rs`, `crates/zeroclaw-channels/src/orchestrator/mod.rs`, `crates/zeroclaw-config/src/schema.rs`, `crates/zeroclaw-runtime/src/daemon/mod.rs`, and `src/config/mod.rs`. CI is all green (16 checks). I read the full diff, traced the `BASELINE_INTENTS` derivation against the Discord API intent table, cross-checked `intents_mask_override` against the existing `DiscordChannel` field pattern, and reviewed all 4 new tests. No prior reviews on this PR; no active blocks from other reviewers.

**Context:** This PR replaces the hardcoded `37377` gateway intents literal in the Discord IDENTIFY payload with a `BASELINE_INTENTS` constant assembled from six named `INTENT_*` constants, and adds an operator `intents_mask` override field in config. The regression test `gateway_intents_default_matches_legacy_mask` pins the computed mask to exactly `37377`. The PR has been running in production for 2 days with intent names logging correctly on connect.

**DRY analysis:** `intents_mask_override: Option<u64>` is a construction-time snapshot from `dc.intents_mask`. This is the same lifecycle pattern as every other `DiscordChannel` struct field (`listen_to_bots`, `mention_only`, `proxy_url`, `stall_timeout_secs`, etc.). The doc comment makes the invariant explicit: channels are rebuilt on config reload, so the snapshot can never drift from the config value it was derived from. This satisfies the AGENTS.md "Materialized views built ON-DEMAND from canonical state" exception when the construction lifecycle bounds the view. No DRY concern.

### 🟢 What looks good — `gateway_intents_default_matches_legacy_mask` is the right regression anchor

This test is the most important artifact in the diff. It proves that `BASELINE_INTENTS` computes to exactly `37377` — byte-identical to the literal it replaced — so existing deployments see no behavior change on upgrade. Any future edit to an `INTENT_*` constant, an inadvertent addition to `BASELINE_INTENTS`, or a default-override path that silently changed the IDENTIFY payload would immediately break this test before it reaches CI. The 2-day production evidence (intent names logged correctly on connect) corroborates the correctness of the assembled value in a live gateway session.

### 🟢 What looks good — `BASELINE_INTENTS` is self-documenting and future-proof

Replacing the bare `37377` with `INTENT_GUILDS | INTENT_GUILD_MESSAGES | INTENT_DIRECT_MESSAGES | INTENT_MESSAGE_CONTENT` makes the intent composition readable without a lookup table — each constant name maps directly to the Discord API documentation. Adding or removing an intent from the default set in the future is a one-line, auditable change with an obvious diff scope. The named constants also make the WARN log for operator overrides that drop baseline bits immediately interpretable: the log names what's missing, not just a changed hex value.

### 🟢 What looks good — `disallowed_intents_hint` turns 4014 into an actionable operator message

Naming the specific Developer Portal toggles ("Message Content", "Server Members", "Presence") in the 4014 close-code handler converts one of the most common Discord bot setup failures from an opaque close code into an immediately actionable error. The fallback to raw mask hex for non-privileged-bit overrides is also the right call: it surfaces the exact bits the operator configured so they can diagnose misconfiguration without reading source. This is a meaningful operator experience improvement with no behavior risk.

### 🟢 What looks good — `intents_mask_override` DRY pattern is consistent and auditable

The snapshot field follows the same construction-scoped pattern as `listen_to_bots`, `mention_only`, `proxy_url`, and every other `DiscordChannel` config field. The doc comment makes the lifecycle invariant explicit at the declaration site: "Intents are connection-scoped (sent once in IDENTIFY), so a construction-time snapshot matches the connection lifecycle exactly: config reloads rebuild channels, which re-derives the mask." This is not a hidden cache — the lifetime is visible in the type and documented at the field, making it auditable on sight.

### 🟡 Warning — `size:` and `risk:` labels not applied

The auto-labeler didn't apply `size:` or `risk:` labels. I agree with the PR body's self-assessment: `size: S` (one channel's IDENTIFY behavior, default mask provably unchanged, new field defaults to `None`) and `risk: low` (no behavior change on default path, production-validated, CI green). Applying `enhancement`, `size: S`, and `risk: low` directly.

### 🔵 Suggestion — confirm `intents_mask` appears in the Discord channel reference docs

The `intents_mask` field carries a `#[tab(Advanced)]` attribute, so it should surface in the config TUI/dashboard. It may not be picked up automatically by the docs generation for the Discord channel reference prose page. Low-stakes follow-up: worth confirming the field appears in the generated channel reference, and if not, adding a brief entry with a note about the gateway privilege model and when an override is needed. No action required before merge.

---

Approving. Default behavior is byte-identical to the prior hardcoded literal, CI is all green, production evidence is present, test coverage is thorough, and the DRY pattern is explicitly documented. Applying `enhancement`, `size: S`, and `risk: low` labels now.

tidux (approved):
First pass at `3df10a35`. Read the full diff, traced `BASELINE_INTENTS` against the Discord API intent table (https://discord.com/developers/docs/events/gateway#gateway-intents), cross-checked the orchestrator wiring at both build sites, ran the 4 new tests + the surrounding Discord channel suite locally, and walked the docs surface for the `intents_mask` follow-up @WareWolf-MoonWall called out. Approving on top of his APPROVED on the same head — branch is `CLEAN` against current master, 16/16 CI green, no active blocks from any other reviewer.

### 🟢 What looks good — `gateway_intents_default_matches_legacy_mask` is the regression that lets this PR be `risk: low`

`INTENT_GUILDS | INTENT_GUILD_MESSAGES | INTENT_DIRECT_MESSAGES | INTENT_MESSAGE_CONTENT` = `1 + 512 + 4096 + 32768 = 37377`, byte-identical to the old hardcoded literal. The test asserts both `ch.gateway_intents() == 37377` and `ch.gateway_intents() == BASELINE_INTENTS`, which is the right belt-and-braces: the first form catches a future edit to an `INTENT_*` constant or a structural drift in `BASELINE_INTENTS` even if `gateway_intents()` is refactored, the second catches a regression in `gateway_intents()` even if the literal mask gets updated for a real reason. Combined with the test for the privileged-intent path, this PR has zero behavior change on the default IDENTIFY for every existing deployment — which is what lets it ship as `risk: low` despite touching the gateway-handshake hot path.

### 🟢 What looks good — `intent_names` treats unknown bits as data, not noise

`intent_names()` decodes the six known bits and reports any remainder as one `unknown(0x...)` entry rather than masking it off silently. That's the right call for the operator override path: if someone sets `intents_mask` to a value that includes a bit zeroclaw doesn't recognize (a newer Discord intent, a typo, or a custom build's extension), the Identify log and the 4014 hint both surface the unknown bit instead of pretending it isn't there. The test at `discord.rs:2832-2836` pins this with the explicit `(1 << 21)` case.

### 🟢 What looks good — `intents_mask_override` snapshot lifecycle is correct *and* documented at the declaration site

The doc comment on the struct field (`discord.rs:39-46`) makes the construction-time-snapshot invariant load-bearing rather than hidden: "Intents are connection-scoped (sent once in IDENTIFY), so a construction-time snapshot matches the connection lifecycle exactly: config reloads rebuild channels, which re-derives the mask." A reader who hits this field and asks "why doesn't this drift from the config?" gets the answer at the field. The orchestrator wires `.with_intents_mask(dc.intents_mask)` at both build sites (`orchestrator/mod.rs:5681` and `:6616`), so the snapshot can never be skipped on the `build_channel_by_id` or `collect_configured_channels` path.

### 🟢 What looks good — `Some(0)` is treated as a legal IDENTIFY value, not coerced

`with_intents_mask(Some(0))` keeps `intents_mask_override = Some(0)`, and the test at `discord.rs:2877` asserts `gateway_intents() == 0`. That's the right semantics — `0` is a valid Discord intent mask (a bot that only listens for gateway-required events, no message/guild events) — and it's an easy thing to get wrong by `unwrap_or(BASELINE_INTENTS)`-ing the override on read. The current `if let Some(mask) = self.intents_mask_override { return mask }` shape avoids that trap.

### 🟢 What looks good — 4014 close code finally tells the operator what to toggle

The most common Discord-bot setup failure I see is operators not enabling "Message Content" in the Developer Portal and then hitting a bare 4014 close code with no hint about where to look. `disallowed_intents_hint` (`discord.rs:1523-1545`) walks the privileged bits in the *requested* mask and names them by their Developer Portal label ("Message Content", "Server Members", "Presence") — which is the form the operator will see in the UI. Pairing that with the IDENTIFY-time WARN when an override drops baseline bits (`discord.rs:1770-1783`) gives defense in depth: silent-handler-loss is logged proactively at connect, and the Portal-toggle hint fires on the upstream rejection.

### 🔵 Suggestion — second @WareWolf-MoonWall's docs follow-up, and it's actionable now

I checked `docs/book/src/channels/discord.md` — it's a hand-written prose page (118 lines), not a generated one, and `intents_mask` doesn't appear anywhere in it. `grep -rn intents_mask docs/` returns no matches at all. The `#[tab(Advanced)]` attribute means it'll surface in the config TUI/dashboard, but the prose page is what an operator reads when they hit a 4014 and Google their way to it. A short Advanced subsection (~5-10 lines) under the existing Discord guide would close the loop: name the field, link to the Developer Portal Privileged Gateway Intents page, mention that `Some(0)` is a legal value, and call out that dropping baseline bits silences message handling (mirroring the WARN). Not blocking — the diff itself doesn't regress docs since the field is new — but the natural moment to write that paragraph is now while the rationale is fresh.

### 🔵 Suggestion — `disallowed_intents_hint` fallback wording is self-contradictory in the rare path

If an operator sets `intents_mask` to a value with no privileged bits at all (e.g. `INTENT_GUILDS` only, the `bare` case in the test at `discord.rs:2849-2851`), the hint still produces the *prefix* "a privileged intent is not enabled for this bot in the Discord Developer Portal", then the body reads "Requested: mask 0x1". A reader who hits this path sees the function tell them a privileged intent isn't enabled when none was requested. Unreachable in real life (Discord won't 4014 on a no-privileged-bits mask), but if it ever does fire it'll send the operator hunting for a portal toggle that isn't the problem. Two ways to tidy:

```rust
if requested.is_empty() {
    return format!(
        " — gateway rejected the requested intent mask {mask:#x}; \
         no privileged bits requested, so this may indicate a malformed override."
    );
}
format!(
    " — a privileged intent is not enabled for this bot in the Discord \
     Developer Portal (Bot → Privileged Gateway Intents). Requested: {}",
    requested.join(", ")
)
```

Or just accept that this branch is dead code in practice and add an inline note saying so. Author's call — not blocking.

---

**Verification at `3df10a35` (local):**

- `cargo fmt -p zeroclaw-channels -p zeroclaw-config -p zeroclaw-runtime -- --check` — clean.
- `cargo clippy -p zeroclaw-channels --features channel-discord --lib --all-targets -- -D warnings` — clean (`Finished` in 1m 01s).
- `cargo test -p zeroclaw-channels --features channel-discord --lib -- gateway_intents intent_names disallowed_intents_hint intents_mask_override` — `4 passed; 0 failed` covering `gateway_intents_default_matches_legacy_mask`, `intent_names_decode_the_mask`, `disallowed_intents_hint_names_privileged_toggles`, `intents_mask_override_wins_verbatim`.
- Cross-checked the math: `1 + 512 + 4096 + 32768 = 37377`. Confirmed `BASELINE_INTENTS` equals the legacy literal.
- Grepped the workspace: the only remaining references to `37377` are the regression-test comment and assertion. No other site still uses the hardcoded mask.

Approving. Solid `risk: low` Discord-channel cleanup that makes the IDENTIFY payload auditable, gives operators an actionable error on 4014, and ships with a regression that makes a future silent default-mask drift impossible.

Audacity88 (approved):
@Nillth, review context: I checked 3df10a35 against the live PR state, the two existing approvals from @WareWolf-MoonWall and @tidux, the current diff/source around Discord IDENTIFY, the config schema field, the two Discord channel construction paths, the PR body validation, current green GitHub checks, and current master freshness. I did not run an additional local cargo pass or live Discord smoke.

Verdict: approving the reviewed head. This clears my requested-review slot only; @singlerider is still requested, and this is not a merge-readiness signal until the freshness note below is handled.

### 🟢 What looks good — default IDENTIFY stays byte-identical

The old default sent the hardcoded mask `37377`; this head derives `BASELINE_INTENTS` from named Discord intent bits and then pins the default path back to `37377` in `gateway_intents_default_matches_legacy_mask`. That is the right regression for this slice because existing configs still send the same IDENTIFY payload unless an operator explicitly sets the new override.

### 🟢 What looks good — the override path preserves operator intent without hidden drift

`intents_mask` defaults to `None`, both Discord construction paths pass it into `with_intents_mask()`, and `Some(0)` remains a real override instead of being collapsed back to the baseline. The connect log reports the resolved mask and decoded names, and the WARN when an override drops baseline bits makes the main foot-gun visible before message handling goes quiet.

### 🟢 What looks good — 4014 diagnostics now point at the actual operator knob

The 4014 close-code path now includes the privileged Developer Portal toggles implied by the requested mask, with a raw-mask fallback for unusual overrides. That turns a bare fatal Discord code into something an operator can fix without reading the source.

Current master is 15 commits ahead of this head, and `crates/zeroclaw-channels/src/orchestrator/mod.rs` changed after this PR's merge base. I do not see that invalidating the reviewed diff, but before merge this needs a branch refresh with CI rerun, an equivalent targeted merge-result smoke for the Discord/channel construction path, or explicit stale-risk acceptance.

PR #7525, opened 2026-06-12
feat(channels/discord): sync archive on message edit, delete, and bulk delete

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The gateway already delivers MESSAGE_UPDATE, MESSAGE_DELETE, and MESSAGE_DELETE_BULK under the intents the Discord channel requests, but the listen loop dropped all three: an archived message stayed frozen at its original text after edits and lingered with no trace after deletion. This PR keeps the `discord.db` archive in sync.
  - Edits append `[edited at <edited_timestamp>: <new content>]`. Discord sends the full message object on every MESSAGE_UPDATE (embed unfurls, pins, flag changes, with `content` present and unchanged), so only events carrying a non-null `edited_timestamp` count as edits; that timestamp also keys idempotency, making redelivered events no-ops.
  - Edits re-run the author gates (bot self, bot policy, peer allowlist) against current config. Archive-time authorization is not durable: a peer removed from the allowlist must not keep writing into agent-readable memory by editing old messages.
  - Deletes append a `[deleted at <ts>]` tombstone once (idempotent under redelivery and delete/bulk-delete races); MESSAGE_DELETE_BULK, the moderation sweep event, tombstones every archived id.
  - Only messages that were actually archived are touched: the lookup on the `discord_{message_id}` key gates everything, and category plus session attribution survive the re-store. Entry growth is bounded at 16 KiB by dropping the middle of the edit history (original text and latest edit retained), since edits are remote-controlled and unlimited.
- **Scope boundary:** agent routing is untouched; MESSAGE_CREATE remains the only event that reaches the orchestrator. No new intents, no config changes. Markers are in-band text (same convention as the existing attachment markers) and are documented as advisory context, not tamper-proof provenance.
- **Blast radius:** the Discord archive path only. Channels without `archive = true` see zero behavior change.
- **Linked issue(s):** None.
- **Labels:** to be set by maintainers (`enhancement`, `channel`, `memory`; suggest `size: S`, `risk: low`).

```text
    Finished `dev` profile   # zero warnings
passed: 896 failed: 0
test result: ok. 868 passed; 0 failed; 1 ignored
$ cargo test -p xtask directives_emit_no_prose_em_dashes
```

- **Commands run and tail output:** above; `channels-full` because `channel-discord` is not in default features (CI covers it via `ci-all`).
- **Beyond CI — what did you manually verify?** A build containing this change has run a production Discord daemon with `archive = true` aliases for two days without archive corruption; the unfurl case (link message gaining an embed) was observed live producing no phantom edit marker. NOT manually exercised: MESSAGE_DELETE_BULK against live Discord (unit-covered; it shares the tombstone path with single deletes).

- New permissions, capabilities, or file system access scope? **No** (writes to the existing archive sidecar only)
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (fixtures use placeholder ids)
- If any `Yes`, describe the risk and mitigation: n/a. Note the deliberate access-control hardening: per-edit author re-checks close a path where de-authorized users could keep writing into agent-readable memory.

## Compatibility (required)

Low risk: `git revert <sha>`. Existing archive entries with markers remain valid plain text either way.


singlerider (approved):
This closes a real archive-integrity gap: the events were already arriving under the existing intents and getting dropped on the floor, so the cost of this feature is exactly the listen-loop dispatch plus the sync logic, no new gateway surface. Approving.

### 🟢 The deauthorization re-check is the security insight of this PR

Re-running bot-self, bot-policy, and allowlist gates against the UPDATE payload, with the explicit rationale that archive-time authorization is not durable, closes an edit-as-write-primitive hole that a naive implementation would have shipped: a removed peer editing old messages to keep injecting text into agent-readable memory. `deauthorized_author_cannot_write_via_edits` pins it.

### 🟢 Edit detection is keyed on the right field

Gating on `edited_timestamp` rather than content diffing means embed unfurls, pins, and flag changes (which redeliver the full object with unchanged content) produce no phantom markers, and the same timestamp doubles as the idempotency key for gateway redelivery. The empty-content early return also makes this safe against Discord's partial MESSAGE_UPDATE payloads: an update without usable content is skipped rather than misrecorded.

### 🟢 Growth is bounded against a remote-controlled input

Edits are attacker-paced and unlimited; the 16 KiB cap with middle-of-history truncation (original + `[edit history truncated]` + latest edit) keeps a hostile editor from inflating the archive, and `edit_history_growth_is_bounded` proves the cap plus the at-most-one-marker overshoot. The doc comment honestly stating markers are advisory in-band text, not tamper-proof provenance, is the right disclosure.

### 🟢 The get-gate keeps scope airtight

Everything hangs off the `discord_{message_id}` lookup, so messages that failed any create-time filter can never be conjured into the archive by their edit/delete events, bulk deletes only touch archived ids, and category plus session attribution survive the re-store. The test set covers redelivery, the delete/bulk race, edit-then-delete, and unarchived-id no-ops.

Agent routing staying MESSAGE_CREATE-only keeps the blast radius to the archive path.

Audacity88 (approved):
I reviewed PR #7525 at current head `ac23767cec0673916c87297e28174f73b8c81f3b`, checking the one-file Discord diff, the PR body, @singlerider's prior approval, live merge/check state, and the existing archive/write gates. This second approval confirms the same core read: the archive sync path keeps agent routing untouched, preserves the existing create-time archive boundary, and re-checks current author policy before edit writes. I also ran a focused local Discord module test slice on this head:

`cargo test -p zeroclaw-channels --features channel-discord --lib discord::tests::`

Result: `125 passed; 0 failed; 0 ignored; 744 filtered out`.

Approving. I do not see a code blocker here.

### 🟢 What looks good — archive sync is gated by the existing archive key

The edit/delete handlers only mutate entries that already exist under `discord_{message_id}`. That keeps the scope tight: a message that failed the original create-time filters cannot be created later by an edit, single delete, or bulk delete event.

### 🟢 What looks good — edits re-check current author policy

The edit path re-runs the bot-self, bot-policy, and peer allowlist checks before appending a marker. That is the important security property in this change: an author who was allowed when the original message was archived does not get a permanent write path into agent-readable memory after being removed from the allowlist.

### 🟢 What looks good — gateway noise stays out of the archive

Using `edited_timestamp` as the edit signal avoids treating embed unfurls, pin/flag updates, or full-object redeliveries with unchanged content as real edits. The timestamp also gives the idempotency key, so duplicate update delivery does not keep adding markers.

### 🟢 What looks good — deletion and growth behavior are bounded

The tombstone path is idempotent across redelivery and delete/bulk-delete races, and bulk delete still only touches archived ids. Edit history growth is capped by preserving the original text and latest edit while dropping the middle history once the entry grows too large, which is the right tradeoff for remote-controlled edits.

PR #7526, opened 2026-06-12
feat(channels/discord): record inbound reactions behind a scope config

Nillth:
## Summary

- **Base branch:** `master` — **stacked on the intent-resolver PR (Depends on #7524)**; review only the top commit (the first commit is that PR).
- **What changed and why:**
  - New `channels.discord.<alias>.reaction_notifications` scope (`off` | `own` | `all`, default `off`): anything other than `off` adds the two unprivileged reaction intents to the IDENTIFY mask via the resolver from the base PR and records inbound reactions to the `discord.db` archive sidecar, where `discord_search` finds them.
  - `own` records reactions to the bot's own messages only (via `message_author_id`, which Discord includes on ADD); `all` records reactions to any message passing the channel's filters. MESSAGE_REACTION_REMOVE forgets the same key, so the archive tracks live reaction state. Removal events carry no author and rely on the key-existence check: a reaction that was never recorded can't be un-recorded.
  - Archive keys use the custom emoji `id` when present, since emoji names are mutable guild state (a rename or deletion between ADD and REMOVE would orphan entries, and two same-named emoji would collide). The human-readable name appears only in the entry content.
  - The bot's own ack/failure reactions echo back as gateway events and are skipped; reactors outside the peer allowlist and events outside the guild/channel allowlists (with the same thread-parent fallback the message path uses) are filtered identically to inbound messages. Reactions from other bots are recorded deliberately; the peer allowlist already governs who is recorded at all.
  - With reactions enabled the resolved mask is 46593. Default config remains 37377: reaction events are not even requested from the gateway.
- **Scope boundary:** no agent wake-ups; recorded reactions are archive/observability only (surfacing them to the agent is deliberately deferred to a planned channel-event-ingestion design). Bulk clears (REMOVE_ALL / REMOVE_EMOJI) are not yet swept and are documented as such on the config field.
- **Blast radius:** `DiscordChannel` event handling plus one config field and the orchestrator wiring. `off` (default) leaves the IDENTIFY mask and event handling byte-identical to the base PR.
- **Linked issue(s):** Depends on #7524.
- **Labels:** to be set by maintainers (`enhancement`, `channel`; suggest `size: M`, `risk: low`).

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 16s   # zero warnings
passed: 896 failed: 0
test result: ok. 871 passed; 0 failed; 1 ignored
$ cargo test -p xtask directives_emit_no_prose_em_dashes   # generated-docs prose gate
```

- **Commands run and tail output:** above; `channels-full` because `channel-discord` is not in default features (CI covers it via `ci-all`). The generated-prose gate was run explicitly since the new schema doc comment renders into the channel docs page.
- **Beyond CI — what did you manually verify?** Nothing live yet: the feature is default-off and has not been enabled on a production alias. The handler is covered by unit tests against a real sqlite archive (scope gating incl. the `own` author gate, custom-emoji id keying with rename/delete round-trip, peer/guild filter parity, identity-less emoji, remove-without-add) and the mask arithmetic is pinned to 46593.

- New permissions, capabilities, or file system access scope? **No** (writes to the existing archive sidecar; both new intents are unprivileged, no Developer Portal toggles involved)
- New external network calls? **No** (reaction events arrive on the existing gateway connection; the only REST involved is the pre-existing thread-parent lookup, reused with its cache)
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — note that enabling `all` records which user reacted with which emoji into the local archive; that is the feature's documented purpose and stays within the operator's existing archive data boundary.
- If any `Yes`, describe the risk and mitigation: n/a.

## Compatibility (required)

- Backward compatible? **Yes** (default `off` requests no new intents and handles no new events)
- Config / env / CLI surface changed? **Yes** — one new optional enum field, `reaction_notifications`, default `off`. No upgrade steps.

Low risk: `git revert` the top commit, or set `reaction_notifications = "off"`. Recorded reaction entries are plain archive rows; they remain harmless either way.


singlerider (approved):
Reviewed the full diff against master since the stack includes #7524's resolver; the reaction work composes with it cleanly. Approving.

Merge order: this PR carries both commits (the #7524 intent resolver `3df10a35` plus the reaction feature on top), so land #7524 first or this brings it along.

### 🟢 No-silent-change parity is pinned by arithmetic

`gateway_intents_default_matches_legacy_mask` asserting the default derived mask equals the old hardcoded `37377`, and the reactions-on mask equaling `46593`, means neither the resolver refactor nor the new feature can shift the IDENTIFY mask for an existing deployment without a test failing. This is exactly how to retire a magic number.

### 🟢 Off means off at the gateway, not just at the handler

`reaction_notifications = off` (the default) doesn't request the reaction intents at all, so the events never arrive; and the handler still re-checks scope for the override/resume edge cases where a wider mask delivers events anyway. Both unprivileged intents, so no new Developer Portal toggles, which the schema doc states.

### 🟢 Filter parity with the message path, deliberately reasoned

Reactor allowlist, bot-self skip (the ack emoji echo), guild allowlist, channel allowlist with the same thread-parent fallback, and an explicit documented decision that other-bot reactions are governed by the peer allowlist rather than `listen_to_bots`. The `reaction_filters_mirror_the_message_path` test walks all three gates. `Own` scope keyed on ADD-only `message_author_id`, with REMOVE relying on key existence, is correctly reasoned about Discord's payload shapes.

### 🟢 Emoji identity keyed on the stable id

Keying custom emoji by `id` instead of mutable name, with the rename/delete-then-REMOVE orphan case actually tested (`name: null` on remove still forgets), shows the failure mode was thought through and not just commented. Identityless emoji payloads are dropped rather than mis-keyed.

### 🟢 Honest limitation disclosure

REMOVE_ALL / REMOVE_EMOJI bulk clears not being swept yet is stated in the config doc where an operator will read it, not buried. Acceptable for an advisory archive; a follow-up issue when this lands would be the natural home.

The `intents_mask` escape hatch is well guarded: `Some(0)` semantics defined and tested, baseline-drop warning logged with named missing intents, and the 4014 close now names the exact portal toggles. The `disallowed_intents_hint` diagnostics alone would justify a PR.

Audacity88 (approved):
Context: I reviewed the top commit over #7524 against the live PR state, the prior approval, the Discord channel path, the config/schema wiring, and the current archive/search behavior. This is stacked on the intent resolver, so #7524 should still land first or this PR will bring that resolver along with it.

### 🟢 What looks good — default-off remains a true gateway boundary

The default path still resolves to the old 37377 mask, and `reaction_notifications = "off"` does not request reaction events at all. When the scope is `own` or `all`, the test pins the expected 46593 mask, so the feature cannot silently widen an existing Discord deployment without an explicit config change.

### 🟢 What looks good — reaction filtering mirrors the message path

The handler reuses the existing peer allowlist resolver, guild allowlist, channel allowlist, and thread-parent fallback before it writes anything to the archive. It also drops the bot's own ack/failure reaction echoes. The `own` scope uses Discord's `message_author_id` on add events, and remove events only forget the same key, so a reaction that was never admitted cannot create new archive state.

### 🟢 What looks good — archive identity and search semantics are practical

Using the custom emoji ID for the archive key avoids the rename/delete orphan case, while the stored content keeps the human-readable emoji name for `discord_search`. The sqlite-backed tests cover add, remove, custom emoji rename/delete, identityless emoji, filter gates, and remove-without-add. That is the right boundary for this default-off archive feature.

### 🟡 Warning — live Discord smoke is still the final confidence gap

The PR body is honest that no production/live alias has been enabled yet. I am comfortable approving without making that an author blocker because this is default-off, archive-only, unprivileged-intent behavior, and the tests exercise the gateway payload shape and sqlite sidecar behavior directly. Before merge, a cheap live or near-live Discord smoke would still be the best final confidence check if a test bot is available: enable `reaction_notifications = "own"` or `"all"` with `archive = true`, add and remove one reaction, then confirm `discord_search` can see and forget the archive entry.

singlerider (approved):
Re-reviewed at head `483e37bf5`. My 06-13 approval was on `f5da53e10`; the only change since is the `483e37bf5` merge of `master`, which preserved the feature (the diff still shows the reaction-scope config wiring: discord.rs +437, schema.rs +33, the orchestrator/daemon/config hookups). CI is green on the merge head. Re-affirming approval.

The inbound-reaction recording behind the scope config is unchanged from what both @Audacity88 and I approved — the master merge is the only delta. 

🔵 One merge-sequencing note: the branch is `CONFLICTING` / `DIRTY` against current `master` even after the merge commit, so it needs another refresh and a fresh green run before it can land (and, as a Discord feature sharing files with the slash-command stack, it should be sequenced with those). Approval carries over once the conflict is resolved cleanly.

PR #7532, opened 2026-06-12
fix(config): align serde defaults with struct Default to prevent save round-trip loss

### LINKED ISSUE #7498: [Bug]: config save default-pruning breaks round-trip for fields whose serde default differs from struct Default
### Affected component

config/onboarding

### Severity

S1 - workflow blocked

### Current behavior

`Config::save()` prunes every key whose value equals `Config::default()` (`prune_default_values`, `crates/zeroclaw-config/src/schema.rs:14834`), and `sync_table` (`crates/zeroclaw-config/src/migration.rs:981`) physically deletes those keys from the on-disk file. The save-path comment claims `#[serde(default = ...)]` re-supplies the pruned values on load, but load deserializes with **per-field serde defaults** — not the struct `Default` impl. A full-schema audit found 9 fields where the two disagree:

**Group A — value flips silently (4 fields):**

| field | struct `Default` | serde refill after prune |
|---|---|---|
| `http_request.enabled` | `true` | `false` |
| `http_request.allowed_domains` | `["*"]` | `[]` (= deny-all) |
| `web_fetch.enabled` | `true` | `false` |
| `web_search.enabled` | `true` | `false` |

**Group B — required fields, section bricked (5 fields, no `#[serde(default)]` at all):**

`memory.backend`, `observability.backend`, `hooks.enabled`, `hooks.builtin.command_logger`, `tunnel.tunnel_provider`

When a section partially survives pruning (any non-default sibling key — a secret's fresh ciphertext, a timeout change, etc.), Group A fields flip on next load (tools silently disabled; `http_request` becomes deny-all). Group B fields make the section fail strict deserialize (`missing field`), so the daemon's resilient loader (`deserialize_resilient`) drops the **entire section** back to `Default`, and the next full save persists that loss to disk. `save_dirty` has the same defect ("dirty paths whose value equals the schema default are removed from the doc instead of written"), so e.g. enabling `web_fetch` from the dashboard writes nothing and reads back `false`.

Note the whole-section prune case is safe (the parent field's `#[serde(default)]` uses the type's `Default::default()`); only **partially surviving** sections trigger the bug.

There is also **no test coverage for this invariant**: nothing verifies that prune-on-save round-trips identically through serialize → prune → deserialize, which is why 9 mismatched fields accumulated unnoticed.

### Expected behavior

Pruning a field that equals `Config::default()` must be a no-op after reload:

1. Every field's serde default must equal its struct-`Default` value (Group A: use `default_true` / a `["*"]` default fn).
2. Required fields must never be pruned, or must gain serde defaults matching the `Default` impl (Group B).
3. A regression test should enforce the invariant schema-wide: serialize `Config::default()` to a TOML table, remove each path one at a time (simulating the pruner), deserialize, and assert the value refills identically. (Exclude `SAVE_PRESERVE_KEYS` paths such as `schema_version`, which the pruner already protects.)

### Steps to reproduce

```bash
# 1. Config with two sections that each have one default-valued key
#    plus one non-default key (keeps the section alive through pruning):
cat > ~/.zeroclaw/config.toml <<'EOF'
schema_version = 3

[web_fetch]
enabled = true                  # == Config::default() -> pruned on save
blocked_domains = ["x.example"] # non-default -> section survives

[memory]
backend = "sqlite"              # == Config::default() -> pruned on save
hygiene_enabled = false         # non-default -> section survives
EOF
# 2. Trigger any full save (Config::save), e.g. change any setting via dashboard/CLI
# 3. Inspect config.toml: the `enabled = true` and `backend = "sqlite"` lines are gone
# 4. Restart the daemon:
#    - web_fetch loads with enabled=false (serde default) -> tool silently disabled
#    - [memory] fails strict deserialize (missing field `backend`) -> resilient
#      loader drops the whole [memory] section to Default; a later save persists the loss
```

### Impact

Affected users: anyone whose config has a partially-non-default `[web_fetch]` / `[http_request]` / `[web_search]` / `[memory]` / `[observability]` / `[hooks]` / `[tunnel]` section — including deployments shipping explicit `enabled = true` + `allowed_private_hosts` overrides.
Frequency: deterministic on every save→load cycle.
Consequence: tools silently disabled; `http_request` becomes deny-all even after re-enabling; Group B causes permanent loss of operator customizations in the affected section (borderline S0 data loss).

### Logs / stack traces

```text
Verbatim output of the exhaustive audit test (serialize Config::default(), remove each
TOML path to simulate pruning, deserialize, compare). 10 hits = 9 real defects plus one
false positive (schema_version, already protected by SAVE_PRESERVE_KEYS):

10 field(s) whose serde default differs from the struct Default (pruning these on save changes behavior on next load):
  hooks.builtin.command_logger: REQUIRED (deserialize fails when pruned: missing field `command_logger` in `hooks.builtin`)
  hooks.enabled: REQUIRED (deserialize fails when pruned: missing field `enabled` in `hooks`)
  http_request.allowed_domains: default = ["*"], serde refill = []
  http_request.enabled: default = true, serde refill = false
  memory.backend: REQUIRED (deserialize fails when pruned: missing field `backend` in `memory`)
  observability.backend: REQUIRED (deserialize fails when pruned: missing field `backend` in `observability`)
  schema_version: default = 3, serde refill = 0   <- false positive: protected by SAVE_PRESERVE_KEYS
  tunnel.tunnel_provider: REQUIRED (deserialize fails when pruned: missing field `tunnel_provider` in `tunnel`)
  web_fetch.enabled: default = true, serde refill = false
  web_search.enabled: default = true, serde refill = false
```

### ZeroClaw version

v0.8.0-beta-2 (master @ 0a3ec7b640)

### Rust version

rustc 1.96.0 (ac68faa20 2026-05-25)

### Operating system

Manjaro Linux (kernel 7.0.10-1-MANJARO, x86_64)

### Regression?

Unknown

### Pre-flight checks

chengzhichao-xydt:
## Summary

Fixes #7498. When `save()` prunes config fields whose value equals the serde-default, a field whose `#[serde(default)]` and `impl Default` disagree silently flips to the serde-default on the next load — a config save round-trip loss.

## Root Cause

Nine fields across eight structs had `#[serde(default)]` (which produces the type's zero-value — `false`, `""`, `0`) while their `impl Default` set a different value (`true`, `"sqlite"`, `"none"`). When all fields happen to equal the serde-default after pruning, save writes an empty table; on reload the fields are re-initialized to serde's default instead of the struct's Default.

## Changes

- **schema.rs**: Replace bare `#[serde(default)]` with named default functions (`default_true`, `default_memory_backend`, `default_observability_backend`, `default_tunnel_provider`, `default_allowed_domains_star`) that return the same value as `impl Default`
- **schema.rs (tests)**: Add 8 empty-table round-trip regression tests — one per affected struct — that deserialize an empty TOML table and assert every field matches the struct Default. An empty table is the extreme case of default-pruning; these tests turn any future serde/struct default drift into a test failure naming the struct

## Validation Evidence

```bash
cargo fmt --all -- --check

# Finished (no warnings)

cargo test -p zeroclaw-config
# 98 tests pass, including 8 new empty_table_round_trips tests

cargo test
```

  - All 9 annotations verified against their struct Default impls
  - The 5 no-serde-default required fields (`memory.backend`, `observability.backend`, `hooks.enabled`, `hooks.builtin.command_logger`, `tunnel.tunnel_provider`) that would break load on a pruned save are now covered
  - Helper functions (`default_true`, `default_memory_backend` etc.) keep the annotation greppable next to the Default impl it must mirror

## Security & Privacy Impact

## Compatibility

- Backward compatible? **Yes** — the fields now serialize to their correct Default values instead of type zero-values
- If `No` or `Yes` to either: exact upgrade steps for existing users: N/A. Configs that were already saved with the correct values are unchanged. Configs that were affected by the pruning bug will now correctly preserve their intended values after a save-load cycle.

## Rollback


crates/zeroclaw-config/src/schema.rs
```diff
@@ -5735,6 +5735,18 @@ fn default_false() -> bool {
     false
 }
 
+fn default_memory_backend() -> String {
```
singlerider, line 6075: Nothing in the suite detects serde-default vs struct-Default drift on the save-pruning path, which is exactly how these nine fields drifted (the 90 passing tests passed before this change too). The next contributor who adds a `#[serde(default)]` field with a non-trivial `Default` reintroduces #7498 with no signal.

Add a round-trip regression test that guards the class, not just these nine fields. `Config` does not derive `PartialEq`, so compare the re-serialized TOML strings:

```rust
#[test]
fn config_save_pruning_round_trips_to_default() {
    let default = Config::default();
    let pruned_toml = /* the same serialize+prune path save() uses */;
    let reloaded: Config = toml::from_str(&pruned_toml).unwrap();
    assert_eq!(
        toml::to_string(&reloaded).unwrap(),
        toml::to_string(&default).unwrap(),
    );
}
```

That turns every future serde/struct default mismatch on the save path into a failure that names the field. Per-struct empty-table tests (`toml::from_str::<HttpRequestConfig>("")` equals default) are an acceptable narrower alternative, but the whole-Config round trip is the one that actually encodes the bug report.

singlerider (changes requested):
Good catch and a correct fix. I verified all nine annotations against their `impl Default` blocks: `enabled: true` x3, `allowed_domains: vec!["*"]`, `backend: "sqlite"`, `backend: "none"`, `hooks.enabled: true`, `command_logger: false`, `tunnel_provider: "none"` all match. The failure mode was nasty too: a user with a hand-edited partial `[http_request]` section silently got the tool disabled, and `save()`'s default-pruning manufactured exactly such partial sections. One blocking item.

### 🔴 Blocking: no test pins the fix, so the drift this PR fixes will recur

The validation evidence is `cargo test -p zeroclaw-config` passing 90 tests, but those 90 passed before this change too; nothing in the suite detects serde-default vs struct-Default drift, which is how nine fields drifted in the first place. The next contributor who adds a field with `#[serde(default)]` and a non-trivial `Default` impl reintroduces #7498 with no signal.

Add a round-trip regression test that guards the class, not just these nine fields:

```rust
#[test]
fn config_save_pruning_round_trips_to_default() {
    let default = Config::default();
    let toml = /* the same serialize path save() uses, with pruning */;
    let reloaded: Config = toml::from_str(&toml).unwrap();
    assert_eq!(reloaded, default);
}
```

If `Config` doesn't implement `PartialEq`, comparing the re-serialized TOML strings of `reloaded` and `default` works without adding the derive. Either shape turns every future serde/struct default mismatch on the save path into a test failure naming the field. Per-struct empty-table tests (`toml::from_str::<HttpRequestConfig>("")` equals default) are an acceptable narrower alternative, but the whole-Config round trip is the one that actually encodes the bug report.

### 🔴 Blocking: PR body does not follow the required template

The body uses a custom Summary/Fix/Validation layout. The repo template (`.github/pull_request_template.md`) requires Validation Evidence with the literal tail of `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` (the body shows only `cargo test -p zeroclaw-config` and `cargo check`, paraphrased counts), plus the Security & Privacy, Compatibility, and Rollback sections. Please refill the body against the template. The change itself is config-default-only and low-blast, so these sections will be short, but they are required for code PRs and a reviewer relies on the literal command tails to diff against their own run.

### 🟢 Helper functions over literals

Reusing `default_true` and adding named helpers (`default_memory_backend` etc.) instead of inline closures keeps the annotation greppable next to the `Default` impl it must mirror.

### 🟢 Group B was the silent killer and is now covered

The five no-serde-default required fields (`memory.backend`, `observability.backend`, `hooks.enabled`, `hooks.builtin.command_logger`, `tunnel.tunnel_provider`) made a pruned save fail to load at all or section-drop. Covering them in the same pass rather than just the four value-flip fields shows the bug was chased to the class boundary.

chengzhichao-xydt:
👋 @singlerider — thank you for the thorough review! Both blocking items addressed:

1. **Regression tests added** — 8 empty-table round-trip tests (`empty_table_round_trips_to_*_config_default`) that deserialize an empty TOML table for every affected struct and assert field equality with the struct Default. An empty table is the extreme case of default-pruning — any future `#[serde(default)]` vs `impl Default` drift will fail one of these tests and name the struct.
2. **PR body now follows the required template** — Validation Evidence, Security & Privacy, Compatibility, and Rollback sections all filled with literal command output.

Rebased on latest master (c48b9ead). All 98 zeroclaw-config tests pass including the 8 new guards.

Audacity88 (commented):
I reviewed current head `558bf37708af0ca67d66bb45e40b00f182813f9a` against #7498, the earlier changes-requested review from `singlerider`, the updated PR body, the one-file `crates/zeroclaw-config/src/schema.rs` diff, and the current green CI checks for Format, Lint, Test, and the required Quality Gate.

### ✅ Resolved — Serde defaults now match the affected struct defaults

The nine mismatches called out in #7498 are now aligned with their `impl Default` values: `http_request.enabled`, `http_request.allowed_domains`, `web_fetch.enabled`, `web_search.enabled`, `memory.backend`, `observability.backend`, `hooks.enabled`, `hooks.builtin.command_logger`, and `tunnel.tunnel_provider`. That addresses both failure shapes from the bug report: the value-flip fields no longer reload to the type zero-value, and the previously required fields can survive a pruned/empty section reload.

### ✅ Resolved — Regression coverage now pins the accepted narrower path

The new `empty_table_round_trips_to_*_config_default` tests cover each affected struct by deserializing an empty TOML table and comparing the affected fields with the struct `Default`. The earlier review preferred a whole-`Config` save-pruning round trip, but it also called this per-struct empty-table shape an acceptable narrower alternative. For this patch, that is enough to catch the drift class that caused #7498.

### ✅ Resolved — PR body now carries the required review sections

The body now includes validation evidence, manual verification notes, security/privacy impact, compatibility, and rollback. One small non-blocking wording nit: the PR body says `save()` prunes fields equal to "the serde-default," while the implementation prunes against `Config::default()` and then relies on serde defaults when loading. The code and tests are aligned; tightening that wording would just make the explanation match the implementation more precisely.

I am leaving this as a comment rather than a formal approval. The technical issues from `singlerider`'s changes-requested review are addressed in the current head, but `singlerider`'s `CHANGES_REQUESTED` state remains active, so this comment does not clear that review or make the PR merge-ready.

singlerider (approved):
Re-reviewed at head `558bf3770`. Both my blockers are resolved by the follow-up commits; @Audacity88 independently verified the same. Converting my changes-requested to an approval.

### RESOLVED ✅ — regression coverage now pins the drift class

The follow-up adds per-struct round-trip tests (`empty_table_round_trips_to_http_request_config_default` and siblings) that deserialize an empty TOML table and compare the affected fields against the struct `Default`. My review preferred the whole-`Config` save-pruning round trip but explicitly called this per-struct empty-table shape an acceptable narrower alternative — and it is, for this patch: it turns the exact serde-default vs struct-Default drift that caused #7498 into a test failure naming the field. Each of the nine aligned fields is covered.

### RESOLVED ✅ — PR body refilled to the template

The body now carries Validation Evidence, Security & Privacy, Compatibility, and Rollback. (Minor, non-blocking: the wording says `save()` prunes against "the serde-default" while it actually prunes against `Config::default()` and relies on serde defaults on load — the code and tests are correct; tightening that sentence would just match the prose to the implementation.)

### 🟢 The fix itself was already right

All nine annotations match their `impl Default` values (`enabled: true` ×3, `allowed_domains: vec!["*"]`, the `memory`/`observability` backends, `hooks.enabled`, `command_logger`, `tunnel_provider`), the named helper functions keep the annotation greppable next to the `Default` impl it mirrors, and the previously-required Group-B fields (which made a pruned save fail to load) are covered in the same pass. CI is green.

Approving — this clears my outstanding review. 🔵 The branch is now `CONFLICTING` against current `master` and needs a refresh before merge.

Audacity88:
@chengzhichao-xydt thanks for the follow-up here. This is approved and the visible checks are green, but the branch has gone DIRTY/CONFLICTING against current master. The overlap is in `crates/zeroclaw-config/src/schema.rs`, which is also the file this PR changes, so I do not want to merge it stale.

Could you please refresh against current master, resolve the schema conflict, and let CI rerun? Once it is clean/green again, this can go back into the merge lane.

chengzhichao-xydt:
Rebased on current master (5afcf06c). Resolved conflicts in schema.rs. The ObservabilityConfig change is no longer applicable as backend has been refactored to an enum type in current master. All other serde-default alignments and round-trip tests are preserved.

@singlerider @Audacity88 ready for re-review.

chengzhichao-xydt:
The `turn_cache_hit_emits_agent_end_with_none_tokens` test passes locally (Rust stable, Linux). The CI failure appears to be a transient/flaky test. Could this be re-run?

NiuBlibing:
@chengzhichao-xydt Is it possible to write a test which emurate every key and do the round-trip test, so that we can cover every filed without writting for every config.

singlerider:

Audacity88 (approved):
I re-reviewed current head c2c0c740862713fb8d6f8646beb86e661ef708d4 against live PR state, linked issue #7498, the current `master` merge refresh, the prior reviews and follow-up comments, the one-file `crates/zeroclaw-config/src/schema.rs` diff, the latest green Quality Gate checks, and merge freshness. I did not run local Cargo validation for this review-only pass.

### ✅ Resolved — Branch refresh and CI are clean again

The earlier conflict state is gone. The branch now merges cleanly against current `master`, and the visible checks are green after the maintainer-side master refresh. That addresses the stale-branch concern I raised after the first approval path.

### ✅ Resolved — The current serde defaults match the current struct defaults

The current diff preserves the #7498 fixes that remain applicable on this head: `http_request.enabled`, `http_request.allowed_domains`, `web_fetch.enabled`, `web_search.enabled`, `memory.backend`, `hooks.enabled`, `hooks.builtin.command_logger`, and `tunnel.tunnel_provider` now deserialize to the same values as their struct `Default` implementations.

### ✅ Resolved — Regression coverage still pins the accepted narrower guard

The empty-table tests now cover the affected structs that still need explicit serde defaults on current `master`. That keeps the accepted narrow test shape from the earlier review: if a future edit makes one of these serde defaults drift away from the struct default again, the corresponding test fails at the config boundary.

### 🔵 Suggestion — Clean up the stale public metadata before merge

The code path is correct, but a few public-record details are stale after the master refresh. The PR body still says the patch added 8 empty-table tests and mentions the old observability default helper, while current head has 7 test functions covering the 8 serde-default-aligned fields that remain in this diff. The body and new test comment also say the save path prunes fields whose value equals the serde default; the implementation prunes against `Config::default()` and then relies on serde defaults when loading. Finally, the live label set is `risk: medium`, so the rollback section should use the medium-risk template fields: fast rollback path, feature flags or config toggles, and observable failure symptoms. None of that changes the correctness of the fix, but tightening it would keep the public record aligned with the current diff.

Approving from my side.

PR #7536, opened 2026-06-12
fix(channels/whatsapp-web): forward quoted media attachments

rifuki:
## Summary

- **What changed and why:**
  - Forward accepted WhatsApp Web image, video, and sticker messages as `MediaAttachment`s instead of dropping them at the channel boundary.
  - Forward consumable media from quoted/replied-to WhatsApp messages with a `quoted-` filename prefix, including quoted audio for the existing audio transcription pipeline.
  - Read `context_info` from media message variants, not only `extended_text_message`, so media replies and structured mentions still satisfy bot-addressing checks.
  - Add lightweight `[Sticker]` / `[Image]` / `[Video]` / `[Document]` placeholders only for media-only messages, preserving the existing empty-message drop for non-media messages.
- **Scope boundary:** This PR does not change media-pipeline behavior, provider vision support, document parsing/download semantics, session/history scoping, read receipts, or outbound media sending.
- **Blast radius:** WhatsApp Web inbound handling only, behind the existing `whatsapp-web` feature. Accepted consumable media messages can now trigger existing media pipeline processing when configured.
- **Linked issue(s):** None.
- **Labels:** `bug`, `size: M`, `risk: medium`, `channel`, `channel:whatsapp`.

```bash
$ cargo fmt --all -- --check
# exit 0

$ cargo test -p zeroclaw-channels --features whatsapp-web
...
test result: ok. 878 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 7.75s
...
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.01s
...
Doc-tests zeroclaw_channels

$ cargo clippy -p zeroclaw-channels --features whatsapp-web --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.20s

$ cargo test -p zeroclaw-channels --features whatsapp-web whatsapp_web::tests::
running 53 tests
...
test result: ok. 53 passed; 0 failed; 0 ignored; 0 measured; 826 filtered out; finished in 0.02s
```

- **Beyond CI — what did you manually verify?** Reviewed the WhatsApp Web inbound path to confirm media downloads occur after allowlist / mention gating, and before `ChannelMessage` delivery to the orchestrator. Verified regression tests cover media `context_info`, reply-to-bot detection for sticker replies, quoted-message extraction, and empty non-media messages remaining dropped. After review feedback, also verified that collected attachments now line up with media kinds consumed by the existing media pipeline: image, video, sticker-as-image, and quoted audio.
- **If any command was intentionally skipped, why:** Full workspace `cargo clippy --all-targets -- -D warnings` and `cargo test` were not run locally; this is a WhatsApp Web channel-only patch, so I ran the channel crate with the `whatsapp-web` feature enabled. Full workspace CI should still exercise the broader matrix.

- New external network calls? (`Yes`)
- If any `Yes`, describe the risk and mitigation: Accepted WhatsApp Web image/video/sticker messages and quoted audio/image/video/sticker messages now trigger WhatsApp media download calls so the existing media pipeline can consume the bytes. The downloads are gated after existing sender policy and bot-addressing checks, and attachments remain in memory instead of being written to disk. Documents are not downloaded because the current media pipeline has no document consumer.

## Compatibility (required)

- If `No` or `Yes` to either: No upgrade steps. Existing text-only behavior remains unchanged; media-only accepted messages now carry a compact placeholder and consumable attachments where supported by the existing media pipeline.

- **Fast rollback command/path:** `git revert ad571a1ed 95fde54a8`
- **Feature flags or config toggles:** Existing `whatsapp-web` feature and `[media_pipeline]` configuration still control the surrounding behavior; no new toggle added.
- **Observable failure symptoms:** Unexpected WhatsApp Web logs containing `failed to download WhatsApp media attachment`, increased media-download latency for accepted image/video/sticker or quoted audio messages, or media messages reaching the agent when they should have been filtered.

- `Co-authored-by` trailers added in commit messages for incorporated contributors? (`Yes/No`): N/A

### DIFF
--- a/crates/zeroclaw-channels/src/whatsapp_web.rs
+++ b/crates/zeroclaw-channels/src/whatsapp_web.rs
@@ -729,6 +729,7 @@ impl WhatsAppWebChannel {
         client: &whatsapp_rust::Client,
         msg: &waproto::whatsapp::Message,
         file_prefix: &str,
+        include_audio: bool,
         attachments: &mut Vec<MediaAttachment>,
     ) {
         use wacore::proto_helpers::MessageExt;
@@ -774,24 +775,18 @@ impl WhatsAppWebChannel {
             .await;
         }
 
-        if let Some(ref document) = base.document_message {
-            let mime = document
+        if include_audio && let Some(ref audio) = base.audio_message {
+            let mime = audio
                 .mimetype
                 .clone()
-                .unwrap_or_else(|| "application/octet-stream".to_string());
-            let file_name = document
-                .file_name
-                .as_deref()
-                .map(|name| format!("{file_prefix}{name}"))
-                .unwrap_or_else(|| {
-                    format!(
-                        "{file_prefix}whatsapp-document.{}",
-                        Self::mime_extension(&mime, "bin")
-                    )
-                });
+                .unwrap_or_else(|| "audio/ogg".to_string());
+            let file_name = format!(
+                "{file_prefix}whatsapp-audio.{}",
+                Self::mime_extension(&mime, "ogg")
+            );
             Self::push_downloaded_attachment(
                 client,
-                document.as_ref() as &dyn Downloadable,
+                audio.as_ref() as &dyn Downloadable,
                 file_name,
                 Some(mime),
                 attachments,
@@ -1603,6 +1598,7 @@ impl Channel for WhatsAppWebChannel {
                                     &client,
                                     msg,
                                     "",
+                                    false,
                                     &mut attachments,
                                 )
                                 .await;
@@ -1611,6 +1607,7 @@ impl Channel for WhatsAppWebChannel {
                                         &client,
                                         quoted,
                                         "quoted-",
+                                        true,
                                         &mut attachments,
                                     )
                                     .await;
--- a/crates/zeroclaw-channels/src/whatsapp_web.rs
+++ b/crates/zeroclaw-channels/src/whatsapp_web.rs
@@ -36,6 +36,8 @@ use std::sync::Arc;
 use tokio::select;
 use waproto::whatsapp::device_props::PlatformType;
 use zeroclaw_api::channel::{Channel, ChannelMessage, SendMessage};
+#[cfg(feature = "whatsapp-web")]
+use zeroclaw_api::media::MediaAttachment;
 #[cfg(not(feature = "whatsapp-web"))]
 use zeroclaw_runtime::i18n;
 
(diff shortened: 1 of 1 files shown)


crates/zeroclaw-channels/src/whatsapp_web.rs
```diff
@@ -620,6 +622,228 @@ impl WhatsAppWebChannel {
...
+
+        attachments.push(MediaAttachment {
+            file_name,
+            data,
+            mime_type,
+        });
+    }
+
+    #[cfg(feature = "whatsapp-web")]
+    async fn collect_media_attachments(
```
singlerider, line 728: Collected media doesn't line up with what the pipeline consumes. `MediaPipeline::process` handles `MediaKind::Audio`, `Image`, `Video`; there is no `Document` variant (`zeroclaw-api/src/media.rs`), so document mimes classify as `MediaKind::Unknown` and hit `_ => {}`. Two mismatches:

1. Audio is extracted for `context_info` (`extract_context_info` reads `audio_message`) but `collect_media_attachments` has no `audio_message` arm, so replying to a voice note forwards nothing even though the pipeline already handles `MediaKind::Audio`. Add the arm mirroring the others with an `audio/ogg` default mime.
2. The document arm downloads the full file via `push_downloaded_attachment`, but `MediaKind::Unknown` is ignored downstream, so the bytes are dead weight (download + memory cost, potentially large) with zero agent-visible effect. Either drop the document arm until the pipeline grows a document path, or wire it to a real consumer.

The sticker arm is fine: `image/webp` classifies as `MediaKind::Image` and is consumed. Keep the invariant that every arm here maps to a `MediaKind` the pipeline processes.

singlerider (changes requested):
The shape of this is right: consolidating the six per-variant `context_info` extractions into one helper (and retiring the stale NOTE about media mentions being filtered upstream), the `quoted-` prefix so the agent can distinguish reply media from direct media, and placeholders gated to media-only messages so the empty-message drop survives for everything else. The `mime_extension` sanitizer with an explicit `image/../../png` traversal test is exactly the kind of paranoia channel boundaries need. One blocking item, in two halves of the same problem.

### 🔴 Blocking: collected media doesn't line up with what the pipeline consumes

Downstream, `MediaPipeline::process` handles `MediaKind::Audio`, `Image`, and `Video`; there is no `Document` variant, so document mimes classify as `MediaKind::Unknown` and fall into the `_ => {}` arm producing nothing, and `msg.attachments` has no other consumer on this path. The sticker arm is fine (`image/webp` classifies as `MediaKind::Image` and is consumed). Two mismatches remain:

1. Audio is extracted for `context_info` but never collected. `extract_context_info` reads `audio_message.context_info`, but `collect_media_attachments` has no `audio_message` arm, so replying to a voice note (arguably the most common WhatsApp reply-to-media case) forwards nothing, while the PR summary says quoted media is forwarded. The transcription pipeline downstream already handles `MediaKind::Audio`, so the arm is cheap to add, mirroring the others with an `audio/ogg` default mime.
2. Documents are downloaded but never consumed. The document arm pulls the full file into memory, but document mimes classify as `MediaKind::Unknown` which the pipeline ignores, so the bytes are dead weight on every document message: pure download and memory cost (WhatsApp documents can be large) with zero agent-visible effect beyond the `[Document]` placeholder that would exist without the download. Either drop the document arm until the pipeline grows a document path, or wire the document branch to whatever consumption you intend; downloading into a void is the one option that shouldn't ship.

Fixing both keeps the invariant simple: every arm in `collect_media_attachments` corresponds to a `MediaKind` the pipeline actually processes.

### 🟢 The context_info consolidation fixes real addressing bugs

Media replies and media mentions now satisfy bot-addressing checks instead of being invisible, `extract_reply_participant` and `extract_mentioned_jids` share the single extraction path, and both are pinned by tests including the lid-JID reply case.

### 🟢 mime_extension is defensive in the right ways

Parameter stripping, subtype charset allowlist, `+` suffix handling, jpeg/svg normalization, and the traversal-attempt test. Filenames derived from attacker-controlled mimetypes are a classic footgun and this closes it.

### 🟢 Placeholder gating preserves existing drop semantics

`media_fallback_content` only fills in for media-bearing messages and leaves genuinely empty messages empty, with both directions tested. Download failures degrade to a WARN plus missing attachment rather than killing the message.

crates/zeroclaw-channels/src/whatsapp_web.rs
rifuki: Addressed in 95fde54a8.

- Added quoted-audio collection with an `audio/ogg` fallback so quoted voice notes map to the existing `MediaKind::Audio` pipeline path.
- Dropped document downloading because documents currently classify as `MediaKind::Unknown` and have no downstream consumer on this path.
- Kept document-only messages as lightweight `[Document]` placeholders without downloading bytes.

Local validation and CI are green again.

Audacity88 (commented):
@rifuki, I re-checked 95fde54a8 against the live PR state, @singlerider's prior review, the current diff, the public media-kind classifier, the media pipeline consumer, the author validation in the PR body, and the current green GitHub checks. I did not run a live WhatsApp Web smoke. The blocker I checked is resolved from my pass, but this is a comment rather than an approval because @singlerider's `CHANGES_REQUESTED` review is still active and should be cleared by them.

### ✅ Resolved — collected media now matches the pipeline consumers

The latest head adds quoted-audio collection with an `audio/ogg` fallback, so quoted voice notes now map to the existing `MediaKind::Audio` path. It also drops document downloads while keeping document-only messages as lightweight `[Document]` placeholders. That restores the useful invariant from the prior review: every downloaded attachment arm now maps to a media kind the current pipeline can consume (`Audio`, `Image`, or `Video`) instead of downloading bytes that would be ignored as `Unknown`.

### 🟢 What looks good — addressability and placeholder behavior stay bounded

The shared `context_info` extraction now covers media variants for mentions, reply-participant checks, and quoted-message lookup. The placeholder helper keeps media-only accepted messages visible to the agent while preserving the empty-message drop for non-media messages. The tests cover those edges, including media `context_info`, reply-to-bot handling for a sticker reply, quoted-message extraction, safe MIME-derived extensions, and the non-media empty-message case.

### 🔵 Merge-readiness note — this is not a merge-ready signal

This comment only records my pass on the updated blocker. The PR still has an active `CHANGES_REQUESTED` review from @singlerider, so it is not merge-ready until that review is cleared. Since `master` has also moved since the last green CI run, do a normal freshness check after the active review block is cleared and before any merge action.

singlerider (approved):
Re-reviewed at head `95fde54a` after the follow-up commit. Both halves of my blocker are resolved with exactly the fixes I asked for; @Audacity88 independently verified the same. Converting my changes-requested to an approval.

### RESOLVED ✅ — collected media now matches what the pipeline consumes

The two mismatches are both fixed at this head:

1. **Quoted audio is now collected** — the new arm (`if include_audio && let Some(ref audio) = base.audio_message`, line 778) collects with an `audio/ogg` fallback (line 782), so replying to a voice note forwards it and it maps to the existing `MediaKind::Audio` transcription path instead of forwarding nothing.
2. **Document downloads are dropped** — `document_message` now returns the lightweight `[Document]` placeholder (line 835-836) without pulling the (potentially large) file into memory. That removes the dead-weight download for content the pipeline would classify as `Unknown` and ignore.

Together this restores the invariant: every arm in `collect_media_attachments` corresponds to a `MediaKind` the pipeline actually processes (`Audio`/`Image`/`Video`).

### 🟢 The rest was already right

The `context_info` consolidation fixes the real addressing bugs (media replies/mentions now satisfy bot-addressing checks; `extract_reply_participant` / `extract_mentioned_jids` share one extraction path, including the lid-JID reply case), `mime_extension` is defensive in the right ways (parameter stripping, subtype allowlist, `+`-suffix handling, jpeg/svg normalization, traversal-attempt test), and placeholder gating preserves the empty-message drop for non-media while keeping media-only messages visible. CI is green and the branch is `MERGEABLE`.

Approving — this clears my outstanding review. Per @Audacity88, do a normal freshness check before merge since `master` has moved.

PR #7538, opened 2026-06-12
fix(zerocode): prevent Cmd-C from triggering Quit on macOS

### LINKED ISSUE #7378: [Bug]: zerocode can treat macOS Cmd-C copy as the quit chord

whiteyzy:
## Summary

On macOS, `normalise_mods` translated `CONTROL→SUPER` for ALL key chords (so Linux `Ctrl+K` and macOS `⌘K` map to the same binding). This also mapped Quit (`Ctrl+C`) to the copy chord (`⌘C`), causing `Cmd+C` to trigger the quit flow instead of terminal copy.

## Fix

Skip the `CONTROL→SUPER` translation for the `c`/`C` key. `Ctrl+C` still quits but `⌘C` remains available for system copy.

## Linked context

Closes #7378

## Verification

```bash
cargo check -p zerocode   # ✅ passes
```

## Risk checklist

- **Breaking change?** No — only affects the 'c' key
- **Side effects:** Other chord bindings (`Ctrl+K`→`⌘K`) unaffected
- **macOS only:** `#[cfg(target_os = "macos")]` gated

### DIFF
--- a/apps/zerocode/src/keymap/chord.rs
+++ b/apps/zerocode/src/keymap/chord.rs
@@ -234,7 +234,12 @@ impl<'de> Deserialize<'de> for Chord {
 
 #[cfg(target_os = "macos")]
 fn normalise_mods(code: KeyCode, mut m: KeyModifiers) -> KeyModifiers {
-    if m.contains(KeyModifiers::CONTROL) {
+    // Translate Ctrl→⌘ on macOS so Linux `Ctrl+K` and macOS `⌘K`
+    // map to the same chord.  Skip this translation for the `c` key
+    // because ⌘C is the system copy chord — mapping it to Ctrl+C
+    // would cause Quit to fire on a copy gesture. (#7378)
+    let is_c_key = matches!(code, KeyCode::Char('c') | KeyCode::Char('C'));
+    if m.contains(KeyModifiers::CONTROL) && !is_c_key {
         m.remove(KeyModifiers::CONTROL);
         m.insert(KeyModifiers::SUPER);
     }


WareWolf-MoonWall (approved):
# PR Review — #7538 fix(zerocode): prevent Cmd-C from triggering Quit on macOS

**PR:** #7538 — fix(zerocode): prevent Cmd-C from triggering Quit on macOS (+6/-1, `apps/zerocode/src/keymap/chord.rs`)
**Round:** 1 (first review)
**Prior blocks:** none
**CI:** passing
**Verdict:** approve

---

## Summary

Targeted guard inside `#[cfg(target_os = "macos")] fn normalise_mods` that skips the CONTROL→SUPER translation for the `'c'`/`'C'` key, preventing ⌘C from being misread as the Quit chord. Closes #7378.

---

## Findings

### 🟢 Fix is targeted and correct

Skipping the CONTROL→SUPER translation for `'c'`/`'C'` stops ⌘C from triggering the Quit binding while leaving all other ⌘+key mappings intact. The `#[cfg(target_os = "macos")]` gate ensures zero cross-platform impact. The two-line diff is the minimum necessary change.

### 🟡 PR template incomplete

The Security & Privacy Impact and Compatibility sections are absent. For risk:low/size:XS these are short answers (all No / all Yes), but they must be present per the PR template before merge. Please fill them in.

### 🟡 Validation evidence is cargo check only

The template asks for literal output from the full battery:

```
cargo fmt --all -- --check
cargo test
```

CI is green so this is not blocking, but the PR body should include that output (or a note that `./dev/ci.sh all` was run clean) rather than just `cargo check`.

### 🔵 Suggested regression test

`chord.rs` already has a test block. Adding two cases here would lock in the invariant cheaply:

- `normalise_mods(KeyCode::Char('c'), KeyModifiers::CONTROL)` → returns `KeyModifiers::CONTROL` (no translation)
- `normalise_mods(KeyCode::Char('k'), KeyModifiers::CONTROL)` → returns `KeyModifiers::SUPER` (translated as before)

Not blocking for a fix this targeted, but worth doing to prevent accidental reversion.

Audacity88:
Thanks @whiteyzy. Closing this as superseded by #7477, which has already merged and closed #7378.

#7477 covers the same macOS Ctrl-C/Cmd-C keymap fix in `apps/zerocode/src/keymap/chord.rs`, and also includes the display-label correction plus macOS regression coverage. Thanks for the focused fix here.

PR #6428, opened 2026-05-06
feat(channels/slack): backfill thread context on first agent encounter

tw-360vier:
## Summary

- **Base branch:** `master`
- **What changed and why:** When an agent is @-mentioned mid-thread today,
  zeroclaw forwards only the triggering message — the thread parent and any
  earlier replies are invisible to the agent loop, so it answers blind.
  This change adds an automatic `[Thread context]` preamble that fetches the
  thread on first encounter and renders parent + recent replies into the
  agent payload, so the agent can answer with full thread context.
- **Scope boundary:** Slack channel only. No changes to outbound message
  shape, thread-replies polling cadence, file-attachment hydration, or any
  other channel.
- **Blast radius:** `crates/zeroclaw-channels/src/slack.rs` only — adds one
  field (`seen_threads: Mutex<HashSet<String>>`), one orchestrator method
  (`maybe_render_thread_backfill`), one async renderer
  (`render_thread_backfill_block`), and five pure helpers
  (`precheck_thread_backfill`, `record_thread_seen`,
  `filter_backfill_messages`, `strip_bot_mentions`,
  `compose_thread_backfill_block`). Adds a `channel_id: &str` parameter on
  `build_incoming_content` (threaded through 3 call sites). No new
  constants, no public API surface changes, no impact on consumers of
  `zeroclaw-channels` outside the Slack module.
- **Linked issue(s):** none upstream.

## Validation Evidence

Local validation in the order CI runs them:

```
$ cargo fmt --all -- --check
(no output — clean)

$ scripts/ci/rust_quality_gate.sh
==> rust quality: cargo fmt --all -- --check
==> rust quality: cargo clippy --locked --all-targets -- -D clippy::correctness
    Checking zeroclaw-channels v0.8.0-beta-1 (...)
    Checking zeroclaw-gateway v0.8.0-beta-1 (...)
    Checking zeroclawlabs v0.8.0-beta-1 (...)
    Finished `dev` profile [unoptimized + debuginfo] target(s)

$ cargo test --locked -p zeroclaw-channels --lib slack::tests::thread_backfill
running 10 tests
test slack::tests::thread_backfill_block_is_prepended_to_payload ... ok
test slack::tests::thread_backfill_compose_renders_allow_list_gap_marker ... ok
test slack::tests::thread_backfill_compose_renders_reply_cap_marker ... ok
test slack::tests::thread_backfill_compose_truncates_long_text ... ok
test slack::tests::thread_backfill_compose_with_only_dropped_senders ... ok
[...]

PR #7540, opened 2026-06-12
refactor(runtime): consolidate the three agent turn engines onto run_tool_call_loop — Implements #7415

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The three agent turn engines — `run_tool_call_loop` (channels/CLI/cron/delegate/SubAgents), `Agent::turn_streamed_with_steering_state` (gateway WS, RPC/zerocode, ACP), and `Agent::turn` (embedded) — are consolidated onto **one engine**: `run_tool_call_loop`, per the maintainer's all-at-once direction on #7415. FND-001's measured-facts table says `loop_.rs`'s job is to "Orchestrate a single agent turn" (`docs/book/src/foundations/fnd-001-intentional-architecture.md:124`) — the new `crates/zeroclaw-runtime/src/agent/turn/` module is that intent made literal.
  - **Group 1 (19 commits, `5ca421c55c..b3bec1ea22`): verbatim mechanical extraction** of the engine from `loop_.rs` (12.9k lines) into `agent/turn/` — an orchestrator plus 22 single-purpose step/type files. Zero behavior change; public paths preserved via `loop_.rs` re-exports.
  - **Group 2 (`5d9cdf4b90..1ccb0c48cb`): the consolidation.** The loop gains four optional capabilities (`event_tx: Option<Sender<TurnEvent>>`, `steering`, `new_messages_out`, `LoopKnobs`); the two private duplicate engines are removed and rebuilt as thin policy wrappers (~225 / ~415 lines) over the loop. A fix made in the engine now applies to every transport instead of needing the same change in three places.
  - **i18n (`600b83ce94`):** the turn engine's user-visible interruption markers (`[interrupted by user]`, `[stream interrupted]`, `[interrupted by user before this tool produced a result]`) now resolve through the Fluent catalogs at every emitting surface (turn engine, gateway WS compensation, ACP server), translation-ready via the standard catalog-sync flow. Also adds the missing `channel-runtime-malformed-tool-output` catalog entry — the lookup existed but no catalog defined it, so users received a raw `{channel-runtime-malformed-tool-output}` placeholder as the fallback reply. English values are byte-identical to the old literals.
- **Pre-push review fixes (`792025c85a..03275a2b69`):** a 7-angle review with adversarial verification was run against this branch before opening; it confirmed 8 regressions in the new glue code (none in the verbatim moves). All are fixed, each pinned by a test first verified red on the pre-fix tree. Details in the commit messages.
- **Scope boundary:** does NOT change `process_message`/`run`/`agent_turn` channel orchestration, `apply_policy_tool_filter` or any security gating, tool execution (`tool_execution.rs` signatures frozen), providers, or the wire schema. No test was moved; no test assertion was modified.
- **Blast radius:** every agent turn on every transport runs through the consolidated loop. The per-caller knobs table below is the honest record of what changes vs what is deliberately preserved.
- **Linked issue(s):** Related #7415 (the RFC this PR implements; see the implementation-update comment there for the direction record). Built and validated fork-first: NNet-Dev/zeroclaw#9 is the test-vehicle PR (same commits) where CI and the full validation record live.
- **Labels:** path labeler applies scope labels; maintainers set `risk:*`/`size:*` (suggest `type: refactor`, `risk: high`, `size: XL`).

## Per-caller knobs (what changes vs what is preserved)

| Knob | Channels / CLI / daemon / cron / delegate / spawn | WS / RPC / ACP (via wrappers) | `Agent::turn` |
|---|---|---|---|
| dedup_enabled | **true** (today) | **false** (today — RPC relies on idempotent retries) | **false** |
| max_iteration_behavior | GracefulSummary (today) | GracefulSummary (NEW — intended universal protection) | **ErrorAtCap preserved** (embedder control signal) |
| detect_protocol_without_tools | **true** (today — protocol junk never reaches end users) | **false** (E2's verbatim-text contract) | **false** (verbatim-text contract, oracle-pinned) |
| event_tx | None | Some(caller's) | None |
| steering | None | **None at the loop level** (the wrapper owns round-level steering; WS injects via the wrapper's `steering_rx`). **Granularity divergence (documented in CHANGELOG):** the old engine drained per tool ITERATION; the wrapper drains per ROUND so each message persists as an enriched user turn — mid-chain steering defers to chain end, cancel stays responsive | None |
| on_delta | Some (draft path) | **None — no relay task** | None |
| pacing (not a LoopKnob) | config | default with **loop_detection_enabled: false** (the streaming engine never had pattern loop detection; multi-round steering turns repeat similar calls by design) | default with **loop_detection_enabled: false** (oracle-pinned: an embedder's identical-args 5-step chain completes) |
| silent | config | **true** (the streaming engine never printed narration; daemon stdout stays clean) | **false** (embedded `print!` path preserved) |

Behavior gains on the streaming/embedded paths (from the shared engine): malformed tool-protocol retry, context-overflow recovery, per-iteration history maintenance/pruning, the `fire_llm_input` hook, and uniform task-local scoping (`TOOL_LOOP_THREAD_ID`/`SESSION_KEY`/`TOOL_CHOICE_OVERRIDE` — closes the `TOOL_CHOICE_OVERRIDE` provider leak). Full plain-language statement in `CHANGELOG-next.md` (this branch).

## Reviewing the commits

- `31d4abeca6` — safety-net tests, written against the UNMODIFIED engines and green there first.
- **`5ca421c55c..b3bec1ea22` (19 commits) are verbatim moves** — review with `git diff --color-moved=dimmed-zebra --color-moved-ws=ignore-all-space`; only visibility/`use`/re-export/doc lines show as non-moved.
- `5d9cdf4b90..1ccb0c48cb` — the consolidation, one concern per commit (plumbing → events → knobs → `Agent::turn` wrapper → streaming wrapper → dead-helper removal).
- `9c77ba484f` — CHANGELOG + architecture docs.
- `792025c85a..03275a2b69` — pre-push review fixes (each commit message names the finding, the old-engine parity argument, and the red-verified test).

- **Commands run and tail output** (local, this branch @ `03275a2b69`):

```
$ cargo fmt --all -- --check     # exit 0, no output
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 41.59s
$ cargo test --workspace --all-targets
    41 suites: 8,499 passed; 0 failed; 10 ignored
```

  Suite enumeration of record: `run_tool_call_loop` **44/44** · `agent::agent` **69/69** (53 upstream + 16 safety-net; the 3 steering oracles green by exact name) · `turn::events::tests` 2/2 · integration 159 · architecture 3 · system 9 · component 188 · migration 90+4 · pacing 3 · ACP bridge 19.
- **Beyond CI — what was manually verified:** every pre-push review fix was first reproduced as a failing test on the pre-fix tree (e.g. the `new_messages` slice panic reproduces as `range start index 41 out of range for slice of length 7`), then fixed, then re-verified green. What was NOT yet verified: live end-to-end turns (below).
- **Step-6 real-world validation — 4 of 5 legs done** (local daemon + binary built from this branch; scratch `--config-dir` with all channels/MCP/cron disabled; free-tier model `kilo-auto/free`; scratch credentials dir removed after the run):
  - **CLI one-shot + tool round** PASS: `agent -a <alias> -m ...` ran a shell tool round through the consolidated loop; reply `zc7415-cli-leg-ok 2026`.
  - **Delegate round** PASS: parent agent delegated to a second agent (child ran its own loop via `crate::agent::run`); reply `zc7415-delegate-leg-ok`. Two earlier attempts surfaced tool-arg validation correctly through the loop (risk-profile refusal, then a malformed `parallel: []` from the free model), with `tool_call_start`/`tool_call_result` trace events emitted from the new `turn/call_prep.rs` / `turn/post_exec.rs`.
  - **RPC (zerocode socket) turn** PASS: `initialize` → `session/new` → `session/prompt` over `<data_dir>/daemon.sock`; observed `tool_call`/`tool_result` updates sharing a correlation id, streamed text chunks, and outcome `completed` with `zc7415-rpc-leg-ok`.
  - **WS turn with mid-turn steering** PASS: raw RFC6455 client on `/ws/chat?agent=...`; a 3-step sequential shell chain; steering injected immediately after the first `tool_call` frame; ONE `done` frame whose `full_response` committed round-1 text AND the steered continuation containing the requested token (commit-before-drain + round loop, live). `thinking`, `tool_call`, `tool_result`, `chunk` frames all observed.
  - **ACP stdio smoke incl. deny-with-edit** PASS: `initialize` → `session/new` (explicit `agentAlias`) → `session/prompt`; a `file_write` call raised `session/request_permission` offering `reject_with_edit` with a `proposedEdit` payload; client answered reject-with-edit with `replacementContent`; the deny was honored (file not written), the turn completed `end_turn`, and `agent_thought_chunk`/`agent_message_chunk`/`tool_call`/`tool_call_update` notifications flowed.
  - **Daemon turn on the live production deployment** PASS: the branch was merged into the production integration build and deployed via the staged-swap flow (backup + atomic replace + graceful restart). A live webhook turn through a production agent (minimax-m3) ran a real shell tool round through the consolidated engine and returned the marker (~12.5s), with all Discord channels registered and healthy on the same binary. Deployment note for the record: the first swap used a default-features build, which silently dropped the opt-in Discord channel feature (~6 min channel outage, restored by the cutover script's one-command rollback; unrelated to the turn engine — the engine turn passed during that window too). The second swap used `--features channels-full`, verified to contain the channel code before deploy, and is the build kept in production.

- Secrets / tokens / credentials handling changed? **No** (`scrub_credentials` moved verbatim to `turn/redact.rs`; same call sites)
- Notes for reviewers: the approval flow is unchanged — `apply_policy_tool_filter` is untouched in the shared setup path; sub-agent non-escalation and depth guards live above `run_turn` and are unchanged. The one approval-adjacent change: the streaming wrapper passes an `AskUserApprovalBridge` so the loop's single-channel approval param reproduces the old multi-back-channel iteration (DenyWithEdit pinned by safety-net seam 5); `record_decision` now attributes `"cli"` instead of the winning channel name — observability-only.

## Compatibility (required)

- Backward compatible? **Yes** — public paths preserved via `loop_.rs` re-exports (`run_tool_call_loop`, `StreamDelta`, `ToolLoopCancelled`, `scrub_credentials`, …); `Agent::turn`/`turn_streamed*` signatures unchanged; WS/RPC/ACP wire schema unchanged (`TurnEvent`/`StreamDelta` untouched).
- Behavior changes are deliberate and documented (CHANGELOG entry + knobs table above).

## Rollback (required)

- **Fast rollback command/path:** revert the PR (single merge revert). No migration, no persisted-state change.
- **Feature flags or config toggles:** None — all-at-once per maintainer direction on #7415.
- **Observable failure symptoms:** grep `tool_loop` / `turn_final_response` / `llm_stream_fallback` log events; per-transport turn failures would surface as WS `done` frames without text, RPC `SessionPromptResult` errors, or ACP prompt failures.

### What happened

1. The turn engine (`run_tool_call_loop` and its private machinery) moved out of `crates/zeroclaw-runtime/src/agent/loop_.rs` into **`crates/zeroclaw-runtime/src/agent/turn/`** — a thin orchestrator plus 22 single-purpose files. Pure moves; behavior unchanged.
2. `Agent::turn` and `Agent::turn_streamed_with_steering_state` (in `agent.rs`) lost their private duplicate engines and became thin **wrappers** over `run_tool_call_loop`. The loop gained `TurnEvent` emission, steering, `new_messages` capture, and explicit `LoopKnobs`.

**Public paths did not change.** Everything you imported from `crate::agent::loop_::` still resolves — `loop_.rs` re-exports from `turn/`. No test was moved and no test assertion was modified; test call sites of `run_tool_call_loop` gained the four new trailing params (behavior-preserving defaults), and two mock providers gained streaming-capability flags.

### Moved into `agent/turn/` (apply your change in the new file)

| Code you touched (old `loop_.rs`) | New location |
|---|---|
| `run_tool_call_loop` skeleton: signature, iteration/cancel/budget control, model-switch check, malformed-retry, final exits | `agent/turn/mod.rs` |
| history trim/prune/orphan-removal at iteration top | `agent/turn/history_window.rs` |
| tool-spec build (`known_tool_names`, `use_native_tools`) | `agent/turn/tool_specs.rs` |
| vision provider routing, image-marker handling | `agent/turn/vision_route.rs` |
| observer `LlmRequest`, llm log, `fire_llm_input`, budget check, the provider call (streaming decision, step-timeout, cancel) | `agent/turn/provider_call.rs` |
| `consume_provider_streaming_response`, `StreamedChatOutcome` | `agent/turn/stream_consume.rs` |
| `StreamTextGuard`, `StreamThinkTagStripper` | `agent/turn/stream_guard.rs` |
| protocol/fence detection (`detect_*`, `find_*`, `starts_suspicious_*`, `complete_*`, `json_fence_body`) | `agent/turn/protocol_detect.rs` |
| response interpretation: observer `LlmResponse`, cost recording, tool-call parsing, `build_native_assistant_history`, `resolve_display_text` | `agent/turn/parse_response.rs` |
| LLM-failure observer event, context-overflow recovery | `agent/turn/context_recovery.rs` |
| per-call prep: `before_tool_call` hook, dedup gate, `set_runtime_approved_arg`, tool_call_start logging | `agent/turn/call_prep.rs` |
| approval flow (CLI prompt / channel `request_approval` / deny / replace) | `agent/turn/approval_gate.rs` |
| `maybe_inject_channel_delivery_defaults` + `AUTO_DELIVERY_DEFAULT_CHANNELS` | `agent/turn/delivery_defaults.rs` |
| post-execution result log, `after_tool_call` hook, completion status | `agent/turn/post_exec.rs` |
| results collection, LoopDetector feed, truncation, receipts, identical-output abort | `agent/turn/results_collect.rs` |
| assistant + tool-result history append | `agent/turn/history_append.rs` |
| max-iteration summary / error-at-cap | `agent/turn/max_iter.rs` |
| `StreamDelta`/`DraftEvent`, progress consts, post-hoc chunker | `agent/turn/events.rs` |
| `ToolLoopCancelled`, `ModelSwitchRequested` (+`is_*`), `ModelSwitchCallback` | `agent/turn/outcome.rs` |
| `scrub_credentials` + `SENSITIVE_KV_REGEX` | `agent/turn/redact.rs` |

### Replaced in `agent.rs` (re-express your change)

| Code you touched (old `agent.rs`) | What happened |
|---|---|
| `Agent::turn` body (its private iteration loop, `execute_tools` path) | Now a wrapper: preamble (cache/classify/memory/TurnGuard) → `run_tool_call_loop` → history replay. Engine-behavior changes go in `turn/*`; pre/post-turn policy goes in the wrapper |
| `turn_streamed_with_steering_state` body (its private loop, event emission, per-call usage, steering rounds) | Now a wrapper: round loop with commit-before-drain steering. `TurnEvent` emission moved INTO the loop (`turn/stream_consume.rs`, `turn/parse_response.rs`, `turn/post_exec.rs`) |
| `Agent::execute_tools` (the mirrored exec pipeline) | **Deleted** — superseded by the loop's `turn/approval_gate.rs` + `agent/tool_execution.rs` |
| `Agent::execute_tool_call` | **Kept, lib-dead** (`#[allow(dead_code)]`) — no production caller remains; survives only as the target of six direct unit tests (ACP approval / shell-`approved`-arg trust). Porting those tests to the loop path and deleting it is a flagged follow-up |
| `drain_steering_messages`, `send_turn_event`, `marked_partial_response`, `append_streamed_assistant_message_to_history`, `prepare_provider_messages`, `should_send_tool_specs`, `parse_response_for_effective_tools` | **Deleted** (superseded by loop equivalents / `turn/steering.rs`) |

### Stayed where it was (your change applies as-is; conflicts are line-drift only)

`run`/`process_message`/`agent_turn`, `apply_policy_tool_filter` + tool filtering, channel/peripheral registrars, model-switch statics, task-local scopes, `build_context`/`build_tool_instructions*`/`AgentRunOverrides` — all still in `agent/loop_.rs`. The entire loop `mod tests` stays in `loop_.rs` (assertions byte-identical; call sites pass `None, None, None, &LoopKnobs::default()`). `TurnGuard`, `classify_model`, response cache, `AgentBuilder`, `StreamedTurnSuccess/Error`, `run_single`/`run_interactive`, `seed_history`/`trim_history` stay in `agent/agent.rs`. `execute_tools_parallel`/`execute_tools_sequential` stay in `agent/tool_execution.rs` (signatures frozen).

### How to resolve your branch

1. **Conflict in `loop_.rs` on something in the "Moved" table:** take upstream's `loop_.rs`, apply your change to the mapped `turn/*` file. Functions kept their names — `grep -rn 'fn <name>' crates/zeroclaw-runtime/src/agent/turn/` finds the new home.
2. **Conflict in the "Stayed" table:** normal resolution; only line numbers drifted.
3. **Conflict inside the old wrapper bodies:** engine behavior → the mapped `turn/*` file (now applies to every transport); pre/post-turn policy → the wrapper (~225/~415 lines, policy-only).
4. **You called `run_tool_call_loop` directly:** pass `None, None, None, &LoopKnobs::default()` for the four new trailing params to preserve exact pre-#7415 behavior.
5. **New tests:** `loop_.rs`'s test module still sees everything via re-exports; step-specific unit tests can sit in the step's own `turn/*` file.

## Known follow-ups (documented, deliberately not in this PR)

- Port the six `execute_tool_call` unit tests to the loop path, then delete the kept lib-dead helper.
- Steering granularity: the loop has an iteration-level drain; giving it the wrapper's enrichment/persistence semantics would restore the old per-iteration responsiveness.
- `ToolCall.extra_content` (e.g. provider thought signatures) is not round-tripped through the loop's assistant-history encoding — the channel path always behaved this way; the embedded/streaming paths now inherit it.
- Loop-emitted `LlmRequest`/`LlmResponse` observer events carry `turn_id: None` on the wrapper paths (loop standard; `AgentStart`/`AgentEnd` keep full attribution).
- Efficiency on the wrapper paths (all match existing channel-path behavior, flagged by review): per-iteration tool-spec rebuild, per-chunk clone when only one sink is attached, and the unused per-session `LocalImageCache` (image-bearing embedded turns re-encode per iteration).


crates/zeroclaw-runtime/src/agent/agent.rs
```diff
@@ -2462,737 +2497,259 @@ impl Agent {
...
+                        self.model_provider.as_ref(),
+                        &mut loop_history,
+                        &self.tools,
+                        self.observer.as_ref(),
+                        &self.model_provider_name,
+                        &effective_model,
+                        self.temperature,
+                        true,
+                        self.approval_manager.as_deref(),
+                        "cli",
```
singlerider, line 2271: This passes the literal `"cli"` as the loop's `channel_name`, which `approval_gate.rs:94` records into the approval audit trail via `mgr.record_decision(...)`. On master the streaming path recorded the **actual deciding back-channel's name** (`decision_channel_name = ch_name.clone()`); the `AskUserApprovalBridge` here iterates the back-channels but discards which one answered (the `Ok(Some(response)) => return Ok(Some(response))` arm drops `channel_name`). Net result: an approval granted from a WS dashboard or ACP editor session is logged as `"cli"` in a security record, which is a regression from master's per-channel attribution.

Restore the old behavior: surface the winning channel name out of the bridge and thread it here in place of `"cli"`, so a WS approval logs WS and an ACP approval logs ACP. Pin the logged value with a safety-net seam-5 assertion.

CHANGELOG-next.md
```diff
@@ -1,3 +1,18 @@
+# Unreleased
```
singlerider: Drop the `CHANGELOG-next.md` edit. Same call as #6428/#6478/#6479: this file is owned by the release process and regenerated at release cut, so contributor edits get clobbered or conflict with every PR that lands before the cut (this branch already pays that tax). The content is excellent and already lives where it should: the PR body and the `docs/book/src/architecture/request-lifecycle.md` update. The release-notes pass will pick up the steering-granularity disclosure from the PR record.

singlerider (dismissed):
This is the most carefully constructed large refactor I've reviewed on this repo: safety-net tests written green against the unmodified engines first, verbatim moves isolated into their own reviewable commit group, the knobs table as an honest record of deliberate divergence, red-verified regression tests for every pre-push finding, and a five-leg live validation including production. I traced the approval gate old-vs-new line by line (decision mapping, deny logging, ReplaceWith path, record_decision placement all match) and checked every remaining consumer of the literal interruption markers for matcher breakage after the i18n switch; the remaining literal sites are emitters being replaced here or en-locale tests, so nothing greps history for a now-localized string. Two blocking items, both small relative to the PR.

### 🔴 Blocking: the approval audit log now records a false channel (regression)

The PR body discloses this as observability-only, but the approval audit trail is a security record: `record_decision` writes an `ApprovalLogEntry` whose `channel` field answers "who approved this tool execution, where". This is a regression, not pre-existing coarseness. On master the streaming path (`agent.rs` old `decision_channel_name = ch_name.clone()`) recorded the **actual deciding back-channel's name**. The consolidated wrapper passes the literal `"cli"` as `channel_name` to `run_tool_call_loop` (`agent.rs:2608`), and the `AskUserApprovalBridge` discards which channel answered (`request_approval` returns only the `ChannelApprovalResponse` at the `Ok(Some(response))` arm, dropping `channel_name`). So an approval granted from an ACP editor or WS dashboard session is logged as `"cli"`, which is affirmatively wrong. An operator auditing a damaging approved call is steered at the wrong surface, and the audit is now less precise than it was on master.

The fix is to restore what master did: surface the winning channel name out of the bridge (return it alongside the response, or have the bridge set it on a field the wrapper reads) and thread it into `record_decision` as `ctx.channel_name`, so a WS approval logs the WS channel and an ACP approval logs the ACP channel. Passing the bridge's own name (`"agent-approval-bridge"`) would be honest-but-coarse and acceptable as a floor, but it is strictly worse than the per-channel attribution master already had. A safety-net seam-5 assertion pinning the logged channel value closes it.

### 🔴 Blocking: drop the CHANGELOG-next.md edit

Same call as #6428/#6478/#6479: `CHANGELOG-next.md` is owned by the release process and regenerated at release cut; contributor edits get clobbered or conflict with every PR that lands before the cut (this branch already pays that tax). The content is excellent, and it should live exactly where you've also put it: the PR body and the architecture doc update. Keep `docs/book/src/architecture/request-lifecycle.md`, drop the changelog hunk, and the steering-granularity disclosure will be picked up by the release-notes pass from the PR record.

### 🟢 The safety-net methodology is the standard now

Oracle tests green against the unmodified engines before any move, every glue regression reproduced red first (the `range start index 41` slice panic in the body is a real reproduction, not a narrative), and no test moved or assertion modified across a 12.9k-line extraction. This is what makes an all-at-once consolidation reviewable at all.

### 🟢 The knobs table converts silent divergence into policy

`dedup_enabled`, `MaxIterationBehavior::ErrorAtCap` preserved for embedders, `detect_protocol_without_tools` honoring E2's verbatim-text contract, `loop_detection_enabled: false` on the steering path with the multi-round rationale: each divergence between the three old engines is now a named, documented, defaulted knob instead of three drifting copies. The streaming wrapper's commit-before-drain round loop with `single_text_exchange` cache gating mirrors the old semantics precisely where I checked them.

### 🟢 The approval bridge preserves multi-back-channel iteration

First-decisive-answer-wins over the registered ask_user handles, per-channel failure logging, DenyWithEdit round-trip pinned by seam 5. The granularity change on steering (per-round vs per-iteration) is disclosed, bounded (cancel stays per-iteration), and has its restoration hook already in the loop.

### 🟢 The i18n pass fixes a live defect on the way through

`channel-runtime-malformed-tool-output` rendering as a raw Fluent placeholder to end users was a real bug; catching it because the consolidation forced an audit of every emitting surface is the refactor paying for itself early.

Fix the audit attribution, drop the changelog hunk, and this is ready. Given the size, I'd also ask for a second maintainer's eyes on group 2 before merge regardless of my approval.

Nillth:
Both blocking items are addressed in the latest push.

1. **Approval-audit channel attribution** — restored master's per-channel attribution; the approval bridge now surfaces which back-channel answered and the gate records that instead of the static `"cli"`. Pinned by a safety-net seam-5 assertion (red as `"cli"`, green as `"edit-channel"`). See the thread on `agent.rs`.
2. **CHANGELOG-next.md** — dropped; the disclosure stays in the PR body and `request-lifecycle.md`. See the thread on `CHANGELOG-next.md`.

On your second point — a second maintainer's eyes on group 2 before merge — agreed, and flagged.

crates/zeroclaw-runtime/src/agent/agent.rs
Nillth: Fixed — per-channel attribution restored, the bridge no longer drops which back-channel answered.

- `Channel` gains a defaulted `last_decision_channel(&self) -> Option<String>` (returns `None` for ordinary single channels, whose own `name()` already identifies the deciding surface).
- `AskUserApprovalBridge` clears it at the top of each `request_approval` fan-out and sets it to the winning back-channel's registered name on the first decisive `Ok(Some(_))`.
- `gate_tool_approval` records `ctx.channel.and_then(|c| c.last_decision_channel()).unwrap_or_else(|| ctx.channel_name.to_string())`, so a WS approval logs the WS channel and an ACP approval logs the ACP channel; the CLI prompt path (`ctx.channel == None`) and single channels keep `channel_name` unchanged.

seam 5 now pins the logged value — it asserts the audit entry's `channel` is the deciding back-channel (`"edit-channel"`); that assertion fails as `"cli"` against the old wrapper and passes after the fix.

One deliberate edge: when the bridge fans out and *no* back-channel answers (`Ok(None)` everywhere → auto-deny), `last_decision_channel()` stays `None` and the record falls back to the loop's `channel_name`. Master left that case as an empty string; I kept the loop name rather than reintroduce the empty value, since the affirmative misattribution you flagged — a real WS/ACP approval logged as `"cli"` — is what the bridge now gets right. Happy to switch the no-answer case to `""` or the bridge's own name if you'd rather match master exactly there.

CHANGELOG-next.md
Nillth: Done — dropped the `CHANGELOG-next.md` hunk; the file is back to its base state. The behavior-change disclosure stays in the PR body and in `docs/book/src/architecture/request-lifecycle.md` (kept), so the release-notes pass can pick the steering-granularity change up from the PR record. Agreed on the #6428/#6478/#6479 precedent.

Nillth:
Follow-up push: removed every `#[allow(dead_code)]` this branch introduced (AGENTS.md bans them in production code), plus the four pre-existing ones in the files we touched — the branch is now net-zero on dead-code suppressions vs master.

- **TurnCtx**: dropped 10 fields the steps never read off `ctx` (threaded as explicit args).
- **Four dead Agent fields** (`tool_specs`, `allowed_tools`, `exclude_memory`, `image_cache`): removed; the two `builder_allowed_tools_*` tests repointed from the dead `tool_specs` to the live `agent.tools`.
- **`execute_tool_call`** (the lib-dead approval-pipeline mirror): deleted, and its six ACP/shell approval + `approved`-arg trust oracles **ported to the real loop path** (`turn_streamed_with_steering_state` → `gate_tool_approval`) as safety-net seam 15 — coverage preserved, now exercising production wiring.
- **`synthesize_cancelled_tool_results`** + dead test scaffolding: deleted (cancel-persistence is covered by the mid-batch-cancel seams).

**`image_cache` was wired in rather than deleted**: the consolidation had routed embedded `Agent` turns through the loop's *uncached* multimodal prep, re-reading+re-encoding local images every iteration. The per-session cache is now threaded through `run_tool_call_loop` into `prepare_messages_for_provider_cached`; a vision_route test pins populate-and-reuse.

Full battery 8,499 / 0, fmt + clippy `-D warnings` clean.

WareWolf-MoonWall (commented):
Re-review at 7df617598, following @singlerider's CHANGES_REQUESTED at 600b83ce.

### ✅ Resolved — CHANGELOG-next.md edit dropped

`CHANGELOG-next.md` is not in the current file list. The release-process and PR-body/architecture-doc content are preserved exactly where they belong — the PR body and `docs/book/src/architecture/request-lifecycle.md`.

### ✅ Resolved — Approval audit attribution restored

`agent.rs` and `turn/approval_gate.rs` are both in the changed file set at the current head, indicating the channel attribution fix was applied. The full diff is too large to trace line-by-line here, but the file-level evidence aligns with the fix direction singlerider described: restoring the winning channel name out of the bridge rather than passing the literal `"cli"`.

---

@singlerider — both blocking items appear addressed at 7df617598. Asking you to verify the attribution fix and dismiss your CHANGES_REQUESTED if satisfied. I'm posting as a comment because your block remains active. Per your own review: given the size of the change, a second maintainer's eyes on group 2 before merge still makes sense.

### 🟢 What remains strong

singlerider's praise from the prior review stands: the safety-net methodology (oracle tests green against unmodified engines before any move, every regression reproduced red first), the knobs table converting silent divergence into explicit policy, and the i18n pass fixing the live `channel-runtime-malformed-tool-output` placeholder defect as a side effect. The live production validation leg — five legs including a production deploy with staged-swap rollback — is an unusually thorough evidence record for a refactor of this scope.

crates/zeroclaw-runtime/src/agent/turn/mod.rs
```diff
@@ -0,0 +1,756 @@
...
+//! → append_tool_round_to_history(&mut history, ..)       history_append.rs
+//! [loop exhausted] → finish_after_max_iterations(..)     max_iter.rs
+//! ```
+//!
+//! Leaf/type modules: `context` (TurnCtx), `events` (StreamDelta/DraftEvent,
+//! pacing consts, post-hoc chunker), `outcome` (ToolLoopCancelled,
+//! ModelSwitchRequested), `redact` (scrub_credentials), `stream_guard` +
+//! `protocol_detect` (streaming protocol suppression), `delivery_defaults`.
+
+pub(crate) mod approval_gate;
```
singlerider, line 50: This is the file-consolidation anchor. The 23 modules here carry ~943 lines of per-file scaffolding (use-headers, doc lines, attributes, blanks), which is the bulk of the remaining +587 production gap. Fold the smallest single-purpose modules into this orchestrator or their one sibling caller, keeping the functions and tests intact: `steering` (16), `knobs` (39), `tool_specs` (45), `history_append` (54), `history_window` (68), `context` (41), `redact` (54), `delivery_defaults` (101). That alone covers the gap and brings the consolidation net-negative against master's 13,378 production lines.

crates/zeroclaw-runtime/src/agent/turn/steering.rs
```diff
@@ -0,0 +1,16 @@
+//! Mid-turn steering: non-blocking drain of caller-pushed messages between
```
singlerider, line 1: Representative of the file-granularity cost: this module is ~16 production lines paying a full file's use-header/doc/attribute tax. A 16-line steering helper does not earn its own compilation unit. Fold it into the orchestrator (`mod.rs`) or whichever single step consumes it. Same treatment for the other sub-60-line single-caller modules (`knobs`, `tool_specs`, `history_append`, `context`, `redact`). The goal is fewer, denser files so the scaffolding overhead drops and the PR lands net-negative.

singlerider (dismissed):
Re-reviewed at `bef597a3`. The de-bloat push is a real, direct response, and I owe a correction on my own numbers before the verdict.

### Correction to my prior production-line figure

My earlier review put production growth at +2,623. That was wrong: it miscounted `safety_net.rs` (~2,000 lines) and `tests.rs` as production because they are gated `#[cfg(test)]` at their `mod` declaration in `agent.rs` rather than with an in-file `#[cfg(test)]` block. Counting the test-only modules correctly:

| | master | this PR | delta |
|---|---|---|---|
| production lines | 13,378 | 13,965 | **+587** |

The real growth was +933 before this push and is **+587** now. The objection holds in direction; the magnitude is a fraction of what I claimed, and I should have caught it before filing the larger number.

### Dead code: fully resolved

Every item I cited is gone, commits mapping one-to-one: `execute_tool_call` (the 309-line orphan) deleted with its oracles ported to the loop (`3eed4732`); the four dead `Agent` fields and `synthesize_cancelled_tool_results` removed (`28b1268`, `aab6c8d`); `TurnCtx`'s ten unread fields dropped and the image cache actually wired in (`13bb732`, `bef597a3`). **Zero `#[allow(dead_code)]` remains in `agent/`.** That is exactly the right kind of change, and it removed 346 production lines.

### 🔴 Blocking: finish the job and land net-negative in this PR

The remaining +587 is not hidden logic; it is file-granularity scaffolding. `turn/` is 23 modules totaling ~3,986 lines, ~943 of which is per-file overhead (use-headers, doc lines, attributes, blanks). Several modules are a single short item paying a whole file's import/header tax: `steering` (16 prod lines), `knobs` (39), `tool_specs` (45), `history_append` (54), `history_window` (68), `context` (41), `redact` (54), `delivery_defaults` (101).

A consolidation has to come out net-negative on production to justify itself, and the only thing now standing between this PR and that bar is the file count. Fold the smallest single-purpose modules into the orchestrator (`mod.rs`) or their one sibling caller, keeping the same functions and tests, until production is at or below master's 13,378. That is roughly the ~600-line gap, and the scaffolding alone covers it. No "land now, consolidate later" - this is the last step and it belongs in this PR. When the production delta is at or below zero, restate the test-excluded accounting in the PR body and I will convert.

### 🟢 Both prior correctness reds remain resolved

Approval-audit attribution records the deciding back-channel (seam-5 green); `CHANGELOG-next.md` is master-clean with the disclosure in the architecture doc.

### 🟢 Lean where it counts

`agent.rs` production is down to 2,539 lines, the old engines are thin wrappers over `run_tool_call_loop`, no dead production code remains, and the new universal capabilities (event_tx, steering, knobs) are small and carry their weight.

---
The dead-code blocker is cleared and my size figure was overstated; what is left is one file-consolidation pass to cross into net-negative, which I am holding for because it belongs in this PR, not a follow-up. A second maintainer on the group-2 commits is still worth having before merge given the blast radius.

singlerider (commented):
**What I need from you: nothing blocking. No code changes requested.**

Concretely:

1. **No action required from you on the size objection.** I withdrew it (details below). Do not spend time folding `turn/` files; I proved it does not work.
2. **The only open item is not yours to do:** a second maintainer should review the group-2 consolidation commits before merge, purely because every transport's turn now flows through this engine (blast-radius precaution, not a defect). I am requesting that review; you do not need to change anything for it.
3. If you want, update the PR body's accounting line to the corrected +587 production-line figure so the record matches, but that is optional housekeeping, not a gate.

I have dismissed both of my prior changes-requested reviews. I am not holding this PR.

---

Correction and withdrawal of my prior blocker.

I implemented the file-consolidation plan I demanded in my last review, to verify it was achievable before holding you to it. It is not, and I owe you a straight correction.

I folded `steering.rs` (the 16-line module I anchored on) into `mod.rs`: deleted the file, moved the function in, dropped the `mod` decl and the re-export. It compiled green. Net production lines saved: **about 3**. Deleting a 17-line file but moving the function and its doc back into `mod.rs` returns almost everything. I then measured the ceiling: collapsing all 23 `turn/` modules into one file removes at most **~112 lines** (90 module-doc lines plus 22 `mod` declarations). The `use` imports do not vanish on a merge; they move with the code. So the maximum achievable by the consolidation I asked for is roughly a fifth of the gap, at the cost of worse structure.

That means my "land net-negative by folding files" demand was wrong, and so was the reasoning behind it. The remaining +587 production lines are not removable scaffolding. They are genuine new code: the consolidated orchestration layer plus the universal capabilities this PR adds to every transport (`event_tx` 106 lines, `knobs` 39, `steering` 16, and the wiring that threads them through the loop). Those are the point of the consolidation, not waste. A refactor that unifies three engines and also gives the streaming and embedded paths capabilities they never had (uniform events, malformed-protocol retry, context-overflow recovery, task-local scoping) legitimately costs some lines. I conflated pure deduplication, which should be net-negative, with dedup-plus-feature-unification, which is what this actually is.

The two inline comments on `mod.rs` and `steering.rs` no longer apply; please disregard them.

For the record, where this PR now stands after your de-bloat push:

- Dead code: fully eliminated. The 309-line `execute_tool_call` orphan deleted with its oracles ported, the dead `Agent` fields and `synthesize_cancelled_tool_results` gone, `TurnCtx`'s unread fields dropped, the image cache wired in. Zero `#[allow(dead_code)]` left in `agent/`. That was the real bloat, and it is gone.
- Production delta corrected: +587 vs master (my earlier +2,623 was wrong, it miscounted the test-only `safety_net.rs` as production; +587 is the honest figure and it is mostly new capability).
- Both correctness items resolved: approval-audit attribution records the deciding back-channel (seam-5 green), `CHANGELOG-next.md` is master-clean with the disclosure in the architecture doc.

This is approvable on the consolidation merits, pending the second-maintainer review noted above.

singlerider (approved):
Approving at `bef597a3`.

@WareWolf-MoonWall, closing the loop on your re-review. You asked me to verify the attribution fix and dismiss if satisfied: done on both counts. I traced the channel-attribution fix to source rather than at file level (the part your review noted was too large to trace): the bridge now captures the deciding back-channel in `last_decision_channel()`, the gate reads `ctx.channel.and_then(|ch| ch.last_decision_channel()).unwrap_or_else(|| ctx.channel_name)` (`turn/approval_gate.rs:100`), and the seam-5 test asserts `entry.channel == "edit-channel"` through a real registered channel and a full streaming turn (`safety_net.rs:687`). It is correct, and it survived the de-bloat push intact at the current head. Both my prior changes-requested reviews are dismissed.

Note for the record: the head moved from `7df61759` (where you reviewed) to `bef597a3` after a five-commit de-bloat push. That push deleted the 309-line `execute_tool_call` orphan and ported its oracles to the loop, removed four dead `Agent` fields and `synthesize_cancelled_tool_results`, dropped `TurnCtx`'s unread fields, and wired the image cache in. Zero `#[allow(dead_code)]` remains in `agent/`.

Why I am approving rather than holding for a separate second maintainer: your re-review is that second set of eyes on the consolidation, and it concurs. Between your review and my line-by-line trace of the attribution fix, the group-2 commits have had the scrutiny the blast radius warranted.

State of the blocking history, all resolved:

- CHANGELOG-next.md edit dropped; disclosure lives in the PR body and `request-lifecycle.md`. (Your ✅ stands.)
- Approval-audit attribution records the deciding back-channel, not the static `"cli"`. (Traced to source, seam-5 green.)
- Dead code fully eliminated; no `#[allow(dead_code)]` left.
- The size objection I raised was wrong and is withdrawn: I implemented the file-consolidation I demanded and it saves ~3 lines per fold (~112 max), not the +587 gap. The remaining delta is genuine new-capability code (uniform events, steering, knobs across every transport), which is the point of the consolidation. My earlier +2,623 figure miscounted the test-only `safety_net.rs` as production; the honest delta is +587.

The strengths you and I both noted hold: safety-net oracle methodology, the knobs table as explicit divergence policy, the i18n fix of the live malformed-tool-output placeholder, and an unusually thorough five-leg validation including a production deploy. This is a clean consolidation. Approving.

Audacity88 (approved):
Approving at `bef597a3`.

I re-reviewed the current diff, linked issue #7415, prior review threads, live PR metadata, and the green check snapshot. I focused on the high-risk runtime turn-loop seams in `crates/zeroclaw-runtime/src/agent.rs` and `crates/zeroclaw-runtime/src/agent/turn/**`, especially approval attribution, streamed turn behavior, image-cache routing, and the cleanup after the earlier de-bloat pass.

### 🟢 Approval — The consolidation keeps the risky seams explicit

The parts I wanted to see pinned are covered on current head: `Agent::turn` and the streamed wrapper now both delegate through `run_tool_call_loop` with their prior divergences expressed as explicit knobs; the streamed path keeps its event/steering/cancellation append-log behavior; the image cache is wired into both loop consumers; and the approval gate now records the deciding back-channel rather than the static loop channel.

### ✅ Resolved — Prior blockers are closed on current head

The stale `CHANGELOG-next.md` edit is gone, the approval-audit attribution fix is traced through the bridge/gate/seam test, and the de-bloat follow-up removed the orphaned agent-side tool executor and dead `Agent`/`TurnCtx` fields. I also rechecked that the remaining `#[allow(dead_code)]` hits in the touched surface are pre-existing ACP-server suppressions rather than new agent-loop debt.

### 🔵 Suggestion — Drop the private run-sheet path from the module docs

One small cleanup I would still make: the new module docs in `crates/zeroclaw-runtime/src/agent/turn/mod.rs` point to `/opt/notes/work/zeroclaw/unification_modular/RUN_SHEET.md`. The condensed run sheet is already copied into the in-tree comment below that line, so the local path does not help future contributors and is out of place in public source. This is not a runtime blocker, but it is worth removing before this becomes the canonical turn-engine module.

### 🔵 Suggestion — Keep the label snapshot tied to labels that exist

The PR body still suggests `type: refactor`, but that label does not exist in this repo. If the body gets one more cleanup pass, I would either list the live labels directly or say `enhancement`, `risk: high`, `size: XL`, `docs`, `agent`, `channel`, `gateway`, and `runtime`. This is maintainer-side metadata polish, not an author-side correctness issue.

### 🟢 What looks good — The validation matches the blast radius

This is still a very large runtime refactor, but the evidence is unusually strong for that shape of change: the PR keeps the scope bounded to turn-engine consolidation, carries regression coverage for the high-risk compatibility seams, and current CI is green across format, lint, tests, platform builds, docs style, security, feature combinations, and the required gate. I do not see a remaining technical blocker.

PR #7546, opened 2026-06-12
fix(runtime): unify SopEngine construction — single instance per daemon

### LINKED ISSUE #6687: Two independent SopEngine instances per daemon: MQTT-started runs are invisible to agent sop_status
## Summary

A single ZeroClaw daemon constructs two independent `Arc<Mutex<SopEngine>>` instances — one for the agent's tool stack, one for the MQTT listener. They share no state. A SOP run started by an MQTT publish lives in a different `active_runs` HashMap than the one queried by `sop_status`/`sop_advance`/`sop_approve` from inside an agent turn.

## Evidence

```
$ grep -rn 'SopEngine::new' --include="*.rs" | grep -v target | grep -v worktrees | grep -v "tests"
crates/zeroclaw-runtime/src/tools/mod.rs:819:        let mut engine = crate::sop::SopEngine::new(root_config.sop.clone());
src/main.rs:1813:                                let mut e = SopEngine::new(sop_config.clone());
src/main.rs:1817:                                SopEngine::new(SopConfig::default());
```

Production construction site 1 — agent tools ([tools/mod.rs:817-827](https://github.com/zeroclaw-labs/zeroclaw/blob/master/crates/zeroclaw-runtime/src/tools/mod.rs#L817-L827)):

```rust
if root_config.sop.sops_dir.is_some() {
    let mut engine = crate::sop::SopEngine::new(root_config.sop.clone());
    engine.reload(workspace_dir);
    let sop_engine = Arc::new(std::sync::Mutex::new(engine));
    tool_arcs.push(Arc::new(SopListTool::new(Arc::clone(&sop_engine))));
    tool_arcs.push(Arc::new(SopExecuteTool::new(Arc::clone(&sop_engine))));
    tool_arcs.push(Arc::new(SopAdvanceTool::new(Arc::clone(&sop_engine))));
    tool_arcs.push(Arc::new(SopApproveTool::new(Arc::clone(&sop_engine))));
    tool_arcs.push(Arc::new(SopStatusTool::new(Arc::clone(&sop_engine))));
}
```

Production construction site 2 — MQTT listener ([main.rs:1804-1830](https://github.com/zeroclaw-labs/zeroclaw/blob/master/src/main.rs#L1804-L1830)):

```rust
mqtt_start: Some(Box::new({
    ...
    move |mqtt_config| {
        let engine = if sop_config.sops_dir.is_some() {
            let mut e = SopEngine::new(sop_config.clone());
            e.reload(&workspace_dir);
            e
        } else { SopEngine::new(SopConfig::default()) };
        let engine = Arc::new(Mutex::new(engine));
        ...
        zeroclaw_channels::orchestrator::mqtt::run_mqtt_sop_listener(&mqtt_config, engine, audit).await
    }
})),
```

Each site owns its own `Arc<Mutex<SopEngine>>`. There is no shared state.

## Behavioural consequences

- **Visibility:** `sop_status` (which queries the agent's engine) cannot see runs started by MQTT.
- **Approvals:** A WaitApproval state created on the MQTT engine cannot be released by `sop_approve` (which talks to the agent's engine).
- **Concurrency limits:** `max_concurrent` and `max_concurrent_total` are enforced per-engine. Two engines can each launch up to the limit, doubling effective concurrency vs. what operators configure.
- **Cooldown:** `cooldown_secs` is per-engine. The same SOP can be re-fired within the cooldown if alternating between sources.
- **Future HTTP fan-in** (#6685) and **future cron fan-in** (#6686) will need a third engine if this is not unified first.

## Context

[#6534](https://github.com/zeroclaw-labs/zeroclaw/pull/6534) (merged 2026-05-10) added the missing `engine.reload(...)` calls at both sites — a real bug fix — but consciously stopped short of unifying them into a single shared engine.

## Suggested resolution

Construct one `Arc<Mutex<SopEngine>>` (and one `Arc<SopAuditLogger>`) at daemon startup, hand the same handles to `build_tools` and to `mqtt_start`. If/when HTTP and cron fan-in land, they receive the same handles. This also fixes the audit-no-op problem tracked separately (the agent path can adopt the wired audit logger).

Yyukan:
## Summary

- **What changed and why:** Unifies SopEngine construction to a single `Arc<Mutex<SopEngine>>` instance per daemon, eliminating the duplicate instance problem where agent tools and MQTT listener each had their own engine with separate state. MQTT-triggered SOP runs are now visible to agent `sop_status`/`sop_approve`/`sop_advance` tools. Engine only registered when `sops_dir` is configured — users without SOPs do not pay the memory cost.
- **Scope boundary:** SOP subsystem only — agent tools, MQTT listener, daemon startup. No changes to channel orchestration, security policy, or tool execution surface.
- **Blast radius:** `crates/zeroclaw-runtime/src/sop/mod.rs` (factory), `crates/zeroclaw-runtime/src/tools/mod.rs` (singleton registration), `src/main.rs` (daemon wiring), `crates/zeroclaw-gateway/src/lib.rs` (test initializers). Affects runtime state sharing and concurrency enforcement.
- **Linked issue(s):** Closes #6687
- **Labels:** `runtime`, `core`, `risk: high`, `size: M`

```
$ cargo fmt --all -- --check
(no output — clean)

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s

$ cargo test -p zeroclaw-runtime -- sop
running 184 tests
test result: ok. 184 passed; 0 failed; 0 ignored; 0 measured; 1901 filtered out
```

CI Quality Gate: green (Format, Lint, Build, Check, Test, Security all pass).

- New permissions, capabilities, or file system access scope? (No)
- New external network calls? (No)
- Secrets / tokens / credentials handling changed? (No)
- PII, real identities, or personal data in diff, tests, fixtures, or docs? (No)

## Compatibility (required)

- Backward compatible? Yes

`git revert <sha>`
- **Feature flags or config toggles:** None — singleton is always active once registered
- **Observable failure symptoms:** SOP runs invisible across channels, `max_concurrent` enforcement bypassed, `WaitApproval` stuck on MQTT-triggered runs. MQTT listener logs INFO skip and returns Ok(()) when no SOPs directory is configured.

N/A


Yyukan:
WIP — plan only. Implementation pending.

Audacity88 (changes requested):
Checked the current diff, linked issue #6687, the PR body validation, comments/reviews, visible CI checks, and current `master` drift on the touched files.

### 🟢 What looks good — The shared-state target is the right fix

The PR is aimed at the right #6687 failure mode. The old code really did create separate SOP engines for the agent tools and the MQTT listener, so MQTT-started SOP runs could be invisible to `sop_status` / `sop_approve` / `sop_advance`. Moving toward one shared engine/audit pair is the right direction, and the current checks are green.

### 🔴 Blocking — SOP tools now disappear outside the daemon path

`all_tools_with_runtime()` no longer registers SOP tools from `root_config.sop.sops_dir`; it only registers them when `resolve_sop_engine()` returns a process-global engine. The only production registration this PR adds is in the `zeroclaw daemon` branch before the reload loop. Other tool-registry callers still go through `all_tools_with_runtime()` without that daemon setup, including the `zeroclaw agent` path that calls `agent::run(...)`.

That means a non-daemon agent run with `sop.sops_dir` configured now builds a tool registry where `resolve_sop_engine()` is `None`, so `sop_list`, `sop_execute`, `sop_advance`, `sop_approve`, and `sop_status` are absent even though the config enables SOPs. That regresses the documented contract that `sop_*` tools are registered when `sop.sops_dir` is configured.

Please keep the shared-engine fix, but make the SOP handle available to every tool-registry construction path that should expose SOP tools. A daemon-scoped/runtime-scoped handle passed into tool construction would avoid making daemon startup the only way SOP tools can exist.

### 🔴 Blocking — The process-wide singleton is stale across daemon reloads

The new daemon registration is done once from the initial `config` before entering the reload loop. On `DaemonExit::Reload`, the code re-reads `current_config` and says the loop will start fresh subsystems, but the SOP singleton remains whatever was registered at first boot. If the daemon starts without `sops_dir` and the operator enables it before reload, no engine is registered. If the daemon starts with one SOP config/path and reloads after the operator changes it, the tools and MQTT listener keep resolving the old engine/audit handles.

That loses a behavior the previous per-iteration construction had: SOP configuration followed the current daemon config on reload. It also shows up in the new tests: `shared_sop_engine_resolve_returns_singleton()` permanently sets the same process-global `OnceLock`, while `sop_tools_absent_when_engine_not_registered()` explicitly depends on running before any test that registers the engine. That makes the regression coverage order-dependent rather than isolated.

Please make the shared SOP state reload-aware and test-isolated. For example, build the shared engine/audit pair per daemon iteration from `current_config` and pass those handles to both tool construction and MQTT registration, instead of storing them in a process-wide `OnceLock` that cannot be replaced.

WareWolf-MoonWall (commented):
This PR is currently in draft state and @Audacity88 already holds an active CHANGES_REQUESTED covering the two architectural concerns that matter most here. I'm not going to run a full parallel review pass on a draft with an open block — that creates noise for @Yyukan without moving anything forward. I'll do a full pass once the PR is marked ready and the draft blockers are addressed.

### 🟢 What looks good — the shared-state direction for SopEngine is the right fix for the #6687 failure mode

The root cause in #6687 was SOP state being re-initialized in contexts that should have shared a single engine, leading to audit gaps and inconsistent tool visibility. Unifying construction around a single daemon-scoped instance is the correct structural response; the alternative approaches (re-initializing per-call, or passing the engine through deep call stacks) either reproduce the original bug or spread construction logic across too many sites. The goal is right and worth getting through review.

Before requesting re-review, the two 🔴 blockers in @Audacity88's active CHANGES_REQUESTED are the ones to resolve: (1) SOP tools disappear outside the daemon path because `all_tools_with_runtime()` only registers them when the daemon has already set the global `OnceLock`, meaning `zeroclaw agent` runs without SOP tools even when `sop.sops_dir` is configured; and (2) the process-wide `OnceLock` singleton can't be replaced across daemon reloads, so SOP config changes on reload are silently ignored and test isolation becomes order-dependent. Both issues point to the same structural question: a `OnceLock` is not the right shape for something that needs to be refreshable. Once that's resolved, tag @Audacity88 for re-review before requesting a broader look.

Yyukan:
## How requested changes were addressed

1. **Duplicate SopEngine instances** — Replaced per-construction-site `build_sop_engine()` calls with a single engine stored in `DaemonRegistry`. The daemon reload loop builds one `(Arc<Mutex<SopEngine>>, SopAudit)` pair and stores it in the registry; `RpcContext` receives the same handle, so all RPC/TUI agent sessions and the MQTT starter share one engine.

2. **Refreshability across daemon reloads** — The engine lives in `DaemonRegistry` (not a process-wide `OnceLock`). On config reload the registry is rebuilt with a fresh engine/audit pair, so SOP config changes take effect immediately without restart.

3. **Test isolation** — No global singleton state. Tests construct `RpcContext` (or pass `None` engine) per-test, so test order cannot affect tool registration.

4. **MQTT listener decoupled from `EngineHandle`** — Listener receives the engine via `DaemonRegistry::take_sop_engine()` and logs an INFO skip when unregistered, returning `Ok(())` instead of panicking.

5. **Memory overhead for users without SOPs** — Engine construction is gated behind `sops_dir.is_some()`. If no SOPs directory is configured, the registry stores `None` and no engine is allocated.

6. **SOP tools available to all daemon paths** — `all_tools_with_runtime()` takes `Option<Arc<Mutex<SopEngine>>>` and `Option<Arc<SopAuditLogger>>` as parameters. Both `dispatch.rs` (RPC/TUI sessions) and the MQTT starter pass the engine from `RpcContext`, so SOP tools are registered wherever the daemon provides an engine — no longer limited to the daemon startup path.

7. **Test coverage** — Added `shared_sop_engine_resolve_returns_singleton` (verifies two sessions receive the same engine from the registry) and `sop_tools_absent_when_engine_not_registered` (verifies SOP tools are absent when engine is `None`).

8. **Gateway test AppState field ordering** — Reordered `sop_engine`/`sop_audit`/`webauthn` fields in 5 test initializers so the `#[cfg(feature = "webauthn")]` guard applies only to the `webauthn` field.

Audacity88 (changes requested):
Checked the updated head ecdc92ab90e7ffbcb3095dacf340c12bf68f18dd against the earlier #6687 blockers, the author follow-up, linked issue #6687, the current diff, current reviews/comments, PR body/labels/checks, and the relevant public source for `create_memory_for_agent`. I did not run local build or test validation for this pass. The two prior blockers are addressed, but a new blocker showed up in the updated construction path.

### ✅ Resolved — daemon SOP state is no longer a process-global OnceLock

The updated shape fixes the main reload concern from my first review. `src/main.rs` now builds the SOP engine/audit pair from `current_config` inside the reload loop, then passes those handles into gateway, channels, MQTT, and the daemon registry. `daemon::run` then moves the same handles into `RpcContext`, so RPC/TUI sessions use the reload-iteration state instead of a process-global singleton. That keeps the shared daemon state refreshable when `sop.sops_dir` is added, changed, or removed.

### ✅ Resolved — direct agent sessions no longer depend on daemon startup

The direct `zeroclaw agent` / standalone agent construction paths now build a local SOP engine when no daemon handle is supplied and `sop.sops_dir` is configured. The relevant call sites in `crates/zeroclaw-runtime/src/agent/agent.rs` and `crates/zeroclaw-runtime/src/agent/loop_.rs` now pass an engine into `all_tools_with_runtime()`, so this addresses the earlier "SOP tools disappear outside the daemon path" issue for direct agent runs.

### 🔴 Blocking — SOP-enabled configs now require an `agents.default` entry

The new SOP construction path hardcodes the audit memory to `create_memory_for_agent(..., "default", None)` in the daemon loop and the standalone agent/process-message paths. That helper looks up `config.agents["default"]` and returns `agents.default is not configured` when the alias does not exist. In the v3/multi-agent config model, `default` is not guaranteed to be an agent alias; users can have only `agents.alice`, `agents.ops`, etc.

That means a normal config with `sop.sops_dir` set but no `agents.default` will now fail before the daemon, `zeroclaw agent`, or channel/process-message SOP wiring can start. The shared-engine refactor correctly removed the process-global singleton, but the replacement should not introduce a magic agent alias dependency.

Please remove the hardcoded `default` agent dependency. For per-agent agent/tool paths, use the current `agent_alias` memory that was already resolved for that session. For daemon-wide SOP state, use a real daemon/global audit memory source that does not require a particular agent alias, or make the intended global SOP audit owner explicit and validated. A regression test should cover `sop.sops_dir = Some(...)` with only a non-`default` enabled agent so this cannot silently come back.

### 🟡 Warning — standalone `channel start` still passes no SOP handle

The daemon-supervised channel path now passes the shared SOP handle into `start_channels`, but the standalone `zeroclaw channel start` command path in `src/main.rs` still calls `start_channels(config, None, cancel, None, None)`. Since `all_tools_with_runtime()` now registers SOP tools only when a handle is supplied, standalone channel runs with `sop.sops_dir` configured will not expose the `sop_*` tools.

I am marking this as a warning rather than the main blocker because the daemon-supervised channel path is the production path the PR is explicitly fixing, while `zeroclaw channel start` is the standalone command. It is still the same contract edge: if a serving path starts from config and should expose normal agent tools, it needs either a shared handle from the daemon or a local handle built from `sop.sops_dir`.

singlerider (commented):
Reviewed current head `20ca520c` (one commit newer than @Audacity88's 06-14 review on `ecdc92ab`). His blocker is resolved here; flagging it so he can convert. Not approving over his active changes-requested on a `risk: high` daemon refactor he should re-verify.

### RESOLVED ✅ — the hardcoded `agents.default` dependency is gone

@Audacity88's blocker was that the SOP construction hardcoded `create_memory_for_agent(..., "default", ...)`, which fails on multi-agent configs without an `agents.default`. At this head:

- There are no production `create_memory_for_agent(..., "default", ...)` calls left. The per-agent paths now use the resolved `agent_alias` (lines ~406/435/464), and `"default"` appears only in a new negative regression test (around line 940) that asserts `create_memory_for_agent("default")` fails when `agents.default` is absent and exercises a non-`default` (`ops`) alias.
- The daemon-wide SOP audit no longer keys on an agent alias at all: `build_sop_engine` takes `audit_memory: Arc<dyn Memory>` as a parameter, and the daemon loop passes the **global** memory via `create_memory(&config.memory, &config.data_dir, None)` (src/main.rs ~line 162), not a per-agent memory. So a config with `sop.sops_dir` set and only non-`default` agents no longer fails to start.

That removes the magic-alias dependency while keeping the process-global-singleton fix from the earlier rounds.

### 🟢 The standalone `channel start` 🟡 also looks addressed

The standalone `zeroclaw channel start` path now builds a local SOP engine/audit from `config.sop` (src/main.rs ~line 164-166) and threads the handle into `start_channels`, so standalone channel runs with `sop.sops_dir` configured expose the `sop_*` tools instead of passing `None`.

### 🟢 The refactor shape is right

Single `build_sop_engine` constructor (callers never call `SopEngine::new`/`SopAuditLogger::new` directly), per-reload-iteration construction from `current_config` so daemon SOP state refreshes on `sops_dir` changes, and shared handles moved into `RpcContext` so RPC/TUI use reload-iteration state. CI is green.

@Audacity88 — your hardcoded-`default` blocker and the standalone-channel warning both appear resolved at `20ca520c`, with a regression test for the no-`default` case. This is ready for your re-verify/convert. Given it's a `risk: high` daemon-wide change I'm leaving your changes-requested as the authoritative state rather than approving over it, but from my read the blocker is cleared.

Audacity88 (approved):
Checked current head ec147bc55d5b78624993e7dc4ea518387c0770fe against my prior #6687 blocker, the follow-up commits, @singlerider's re-check, the current diff, visible CI, and the relevant SOP/tool/gateway construction paths. I did not run local cargo validation for this pass; visible CI is green and `git diff --check` is clean. This approval clears my prior blocker on the reviewed head, but it is not a merge-ready statement: the branch is still behind current `master`, and current `master` has same-file churn in `crates/zeroclaw-runtime/src/agent/agent.rs`.

### ✅ Resolved — SOP construction no longer depends on `agents.default`

The latest update removes the hardcoded `create_memory_for_agent(..., "default", None)` dependency that blocked the previous head. The per-agent construction paths now use the resolved `agent_alias`, so configs with `agents.ops` or another non-`default` alias can build SOP audit memory for that session. The daemon path now builds daemon-wide audit memory with `create_memory(&current_config.memory, &current_config.data_dir, None)`, so it no longer requires any specific agent alias.

The new regression test also pins the important shape: a config with `sop.sops_dir` and only `agents.ops` succeeds for the real session alias and fails for the old hardcoded `default` alias. That is the right guard for the multi-agent config model.

### ✅ Resolved — standalone `channel start` now receives SOP handles

The warning from my previous review is addressed too. The standalone `zeroclaw channel start` path now builds an engine/audit pair from config when `sop.sops_dir` is set and passes those handles into `start_channels()`. That keeps the standalone channel path aligned with the daemon-supervised channel path instead of silently registering tools without SOP state.

### 🟢 What looks good — the shared-state shape is now reload-scoped

The PR has moved away from the original process-wide singleton and now builds the daemon SOP engine/audit pair inside the reload loop from `current_config`. That pair is then cloned into gateway, channels, MQTT, and `DaemonRegistry`, with `RpcContext` carrying the same handles into RPC/TUI sessions. That preserves the #6687 fix, where MQTT-started runs need to be visible to `sop_status`, `sop_approve`, and `sop_advance`, without making the state stale across daemon reloads.

The direct agent and `process_message` paths also build a local SOP engine when no daemon handle is supplied and SOPs are configured, so the earlier "SOP tools disappear outside daemon startup" blocker is resolved on the current diff.

### 🔵 Suggestion — keep the standalone gateway tool listing aligned

One small edge still looks worth cleaning up, though I am not blocking on it. `run_gateway_if_enabled()` still calls `run_gateway(..., None, None)` for the SOP handles, so the standalone `zeroclaw gateway` seed registry used by `/api/tools` will omit the `sop_*` tool specs even when `sop.sops_dir` is configured. The actual gateway chat/WebSocket/ACP execution paths now build or receive SOP handles, so this looks like a listing/metadata drift rather than the runtime break we had before.

If the dashboard tools page is meant to stay authoritative for standalone gateway mode, mirror the standalone `channel start` fix here too: build a local engine/audit pair before calling `run_gateway`, or otherwise make `/api/tools` derive from the same per-session construction path.

This review approves the technical fix at ec147bc55d5b78624993e7dc4ea518387c0770fe, but the branch is still two commits behind current `master`, and current `master` changed `crates/zeroclaw-runtime/src/agent/agent.rs` after this PR's merge base. Before merge, update the branch and wait for CI, run an exact merge-result smoke that covers the affected runtime/agent surface, or make an explicit maintainer call to accept that stale-risk.

PR #7464, opened 2026-06-10
fix(mcp): make configured servers usable by default

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Enable configured MCP by default: missing `[mcp].enabled` now parses as enabled, while explicit `enabled = false` still disables MCP.
  - Make eager MCP schema loading the default, with `deferred_loading = true` kept as the opt-in large-catalog path.
  - Add `tool_search` to the default `auto_approve` list so deferred discovery can work in non-interactive sessions.
  - Update MCP docs and regression coverage for the new defaults.
- **Scope boundary:** This does not auto-approve MCP server tools, change MCP transports, change tool filtering, or remove explicit opt-outs.
- **Blast radius:** MCP config defaults, default risk-profile discovery approval, and MCP docs. Existing configured MCP servers become active unless users set `enabled = false`.
- **Linked issue(s):** Closes #6721, Closes #6873, Closes #6874.
- **Labels:** `bug`, `risk: high`, `size: XS`, `config`, `runtime`, `docs`, `tool:mcp`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
git diff --check
PASS

cargo fmt --all -- --check
PASS

cargo test -p zeroclaw-config default_auto_approve_includes_tool_search
PASS: 1 passed; 776 filtered out

cargo test -p zeroclaw-config mcp_config_
PASS: 18 passed; 759 filtered out

cargo test -p zeroclaw-runtime non_interactive_tool_search_is_auto_approved
PASS: 1 passed; 2045 filtered out

cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings
PASS: finished in 25m52s

cargo nextest run --locked --workspace --exclude zeroclaw-desktop
NOT GREEN: stopped after 5307/8195 tests run; 5304 passed, 3 delegate background-task timeout tests failed, 10 skipped. The three failed delegate tests passed when rerun focused under nextest.
```

- **Beyond CI — what did you manually verify?** Checked the final branch diff after rebase. The serde default for missing `mcp.enabled` uses the new enabled default, explicit `enabled = false` and `deferred_loading = true` are still honored, and docs still state that MCP tool execution follows the normal approval policy.
- **If any command was intentionally skipped, why:** No planned local validation command was skipped. The CI-shaped nextest run did not complete green because of unrelated delegate timeout failures under broad load.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? `Yes`
[...]

PR #7547, opened 2026-06-12
fix(runtime): auto-include discovered MCP tools in risk_profile allowed_tools

tidux:
## Summary

- **Base branch:** `master`
- **What changed and why:** After #7464 flipped `mcp.enabled` default to `true` and `deferred_loading` to `false`, ZeroClaw connects and enumerates MCP tools but they never appear in an agent's visible tool list when the agent's `risk_profile` has an explicit `allowed_tools` list. Three code paths needed fixing:
  1. `DelegateTool::resolve_allowed_tools` (web gateway / sub-agent path) — initial commit.
  2. `ToolAccessPolicy::is_tool_allowed` + eager-MCP registration (`mcp_tool_access_policy` / `register_eager_mcp_tool_if_allowed`) used by the main agent loop (Zerocode Chat/Code panes) — second commit.
  3. `DelegateTool::resolve_allowed_tools` deny-list pass (added after review): the auto-admit union step ignored the target profile's `excluded_tools`, so operators had no way to block individual destructive MCP tools like `filesystem__write_file` in the agentic-delegate path. Now `excluded_tools` always subtracts from the resolved set.
- **Scope boundary:** Augments the risk-profile allow-list resolution for MCP tools (names containing `__`), keeps caller-supplied per-run `allowed_tools` as a strict second gate, and adds the missing `excluded_tools` subtraction to the delegate agentic path. Risk-profile `allowed_tools` empty or omitted remains "no authorization constraint"; the TOML config cannot distinguish omitted from `allowed_tools = []`, so this PR no longer documents that as a deny-all escape hatch. Built-in tool gating outside the MCP wrapper paths and the approval pipeline are unchanged.
- **Blast radius:** Affects every code path that builds a `ToolAccessPolicy` or calls `resolve_allowed_tools` for agents with non-empty `allowed_tools` lists. Top-level agents with empty/omitted `allowed_tools` (the "all tools" case) are unaffected. Agents that already relied on `excluded_tools` to block tools in the agentic-delegate path will start seeing those tools actually blocked — they were previously silently ignored on that path, so the new behavior matches the documented contract.
- **Linked issue(s):** Fixes the regression reported after the MCP default change; resolves Zerocode Chat/Code pane invisibility.
- **Labels:** `type: fix`, `risk: high`, `size: S` (matches the auto-applied labels).

```bash
cargo fmt --all -- --check
cargo test --workspace
```

```
$ cargo fmt --all -- --check
    Checking zeroclaw-config v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-config)
    Checking zeroclaw-providers v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-providers)
    Checking zeroclaw-memory v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-memory)
    Checking zeroclaw-tools v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-tools)
    Checking zeroclaw-runtime v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-runtime)
    Checking zeroclaw-hardware v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-hardware)
    Checking zeroclaw-channels v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-channels)
    Checking zeroclaw-gateway v0.8.0 (/home/tidux/src/zeroclaw/crates/zeroclaw-gateway)
    Checking zeroclawlabs v0.8.0 (/home/tidux/src/zeroclaw)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 35.46s
$ cargo test --workspace
    ... (per-crate suites omitted; sum below) ...
    TOTAL: 8485 passed, 0 failed, 22 ignored
```
  - `ToolAccessPolicy` now keeps the risk-profile allow-list and caller-supplied per-run allow-list as separate gates. The `__` MCP auto-admit exception applies only to the risk-profile gate; the caller gate remains an exact-name intersection.
  - Eager MCP registration and deferred `tool_search` registration both reject `filesystem__write_file` when the caller-supplied per-run list narrows to `allowed_tools = ["cron_add"]`. Regression tests cover direct policy evaluation, eager registration, and deferred `mcp_allowed_tool_count`.
  - `DelegateTool` now uses `resolve_tool_policy` plus `delegate_admits_with_mcp` after the #7608 merge-shape change; three delegate regression tests still pin auto-admit, `excluded_tools` against auto-admitted MCP names, and `excluded_tools` against explicit allow-list entries.
  - `SecurityPolicy::from_profiles` is now documented and tested to map risk-profile `allowed_tools = []` to `allowed_tools: None` (unrestricted), because the public TOML config cannot distinguish omitted from explicit empty.

This PR changes the risk-profile security contract for runtime-discovered MCP tools from strict exact-name allow-listing under a non-empty `allowed_tools` list to default-accept unless denied for MCP-style names containing `__`. The old behavior did not work in practice: operators were manually maintaining MCP tool names in their allow-lists and tools were still being silently dropped from the agent's visible registry. The mitigation is `excluded_tools`: it subtracts from both the main agent loop path and the delegate-agentic path, so operators can block individual destructive MCP capabilities like `filesystem__write_file` while keeping the rest of the server reachable. Caller-supplied per-run `allowed_tools` remains a strict exact-name second gate, so cron/job-level narrowing is not weakened. Docs are updated in this branch: `crates/zeroclaw-config/src/schema.rs` rustdoc for `RiskProfileConfig::{allowed_tools,excluded_tools}`, `docs/book/src/tools/overview.md`, and the "Authorization" subsection in `docs/book/src/tools/mcp.md`.

## Compatibility (required)

- Backward compatible? (`Yes`, with one behavior-corrects caveat)
  - No config changes are required.
  - Operators who set `excluded_tools` in a risk profile and use the agentic-delegate path will start seeing those entries actually honored on that path. The previous behavior silently ignored `excluded_tools` for agentic delegates, which contradicted the documented contract. If any agent relied on that bug (i.e. expected an excluded tool to still be reachable from a delegate), remove the entry from `excluded_tools` to restore the old behavior.

Core pre-merge commits in this PR (oldest first, all on `fix-mcp-tool-visibility-v2` before the later master merge and review follow-ups):

1. 6365127f7 — `fix(runtime): auto-include discovered MCP tools in risk_profile allowed_tools` (DelegateTool path)
2. 105b9fcf8 — `fix(runtime): auto-admit runtime-discovered MCP tools in ToolAccessPolicy` (main loop / eager-MCP path)
3. 5a8c81147 — `fix(tools): honor explicit empty allowed_tools in ToolAccessPolicy` (programmatic `ToolAccessPolicy` state, not reachable from risk-profile TOML config)
4. 16b7fc559 — `style(tools): collapse nested if in ToolAccessPolicy::is_tool_allowed`
5. 6691bba90 — `test(tools): align policy_filters_select_results with new MCP auto-admit semantics`
6. e1d4a0b68 — `fix(runtime): honor risk_profile.excluded_tools in delegate agentic path` (post-review deny-list fix)
7. 7f6a07093 — `docs(security): document MCP allow-list contract change from #7547`
8. 4feb12ef5 — `docs(mcp): drop em-dashes flagged by docs style gate`

Later review follow-ups on the current head include the #7608 master merge, `d259ce74d` (per-run allow-lists stay a strict second gate), `567fd2aa` (docs style), and the latest docs clarification that risk-profile `allowed_tools = []` is unrestricted rather than deny-all.

Rollback options, in order of preference:

- **Targeted revert of the post-review deny-list fix:** `git revert e1d4a0b68` rolls back just the `excluded_tools` subtraction in the delegate-agentic path. The auto-admit branch (commits 1–3) and the docs stay in place. This is the right move if honoring `excluded_tools` in the delegate path turns out to break a downstream agent that relied on the previous silent-ignore behavior.
- **Targeted revert of just the auto-admit contract:** `git revert 105b9fcf8 6365127f7` backs out only the MCP auto-admit behavior on both the main loop and delegate paths, leaving the programmatic `ToolAccessPolicy` empty-list handling (commit 3) intact. Use this if the auto-admit policy itself turns out to be too permissive.
- **Full revert:** `git revert 4feb12ef5 7f6a07093 e1d4a0b68 6691bba90 16b7fc559 5a8c81147 105b9fcf8 6365127f7` (or revert the squash-merge commit on `master` once merged) backs out the entire MCP-visibility change. Combine with reverting #7464 to restore the pre-`mcp.enabled = true` default if the regression is rooted there.
- **Forward-fix:** if a specific MCP tool name turns out to be wrongly auto-admitted, list it in the affected profile's `excluded_tools`. That is the operator-facing escape hatch this PR adds and the docs now describe.


Audacity88 (changes requested):
Reviewed 6691bba against the full diff, current green CI, the merged #7464 MCP-defaults context, and the runtime/tool-policy paths for eager MCP, deferred `tool_search`, and agentic delegates. There are no prior reviews or comment threads on this PR.

### 🟢 What looks good — Covers both eager and deferred MCP visibility

The branch follows the regression through both main-agent MCP registration and deferred `tool_search`, instead of fixing only one surface. The tests around deny-list filtering and deferred result filtering also point in the right direction: if MCP tools are going to be discovered dynamically, the block path needs to be operator-visible and exercised directly.

### 🔴 Blocking — `excluded_tools` is not honored for agentic delegates

The PR body says MCP tools can still be blocked with `excluded_tools`, but I do not think that is true in the delegate path. `DelegateTool::resolve_allowed_tools` appends every parent tool whose name contains `__` to the target risk profile's allow-list, and `execute_agentic` then builds `sub_tools` by checking only that computed allow set before calling `run_tool_call_loop`. I do not see the target profile's `excluded_tools` subtracted before the sub-agent receives the tool list.

That means a target profile that allows some tool but excludes `filesystem__write_file` can still receive that MCP wrapper in an agentic delegate, which is the exact escape hatch the PR body says remains available. Please apply the same deny-list semantics in this path and add a regression that uses a real `<server>__<tool>` MCP name. The current `mcp_tools_included_in_subagent_tool_list` fixture uses `mcp_fake`, so it proves explicit allow-listing still works, but it does not exercise the new double-underscore auto-admit branch.

### 🔴 Blocking — Document the new MCP allow-list contract before merge

This PR changes runtime-discovered MCP tool visibility from default-deny under explicit `allowed_tools` to default-accept unless denied. That may be the right product direction, but it is a high-risk security contract change and the current docs/schema still describe `allowed_tools` as the authorization list where `Some(list)` admits only listed names and `excluded_tools` subtracts from that set.

Please update the MCP/security docs, and any generated/reference text that mirrors the config contract, to say how `<server>__<tool>` names behave with `allowed_tools` and `excluded_tools`. The PR body should also stop sending mixed risk/release signals here: it still says `risk: low` and "Marking as draft pending docs and user validation," while the live PR is non-draft and correctly labeled `risk: high`. #7464 explicitly did not change tool filtering, so this PR is the place where users need to learn that configured MCP tools now have a special allow-list exception and must be blocked individually with `excluded_tools`.

tidux:
Thanks for the careful review — you were right on both blocking comments. Pushed two commits addressing them:

- **e1d4a0b** `fix(runtime): honor risk_profile.excluded_tools in delegate agentic path` — `DelegateTool::resolve_allowed_tools` now subtracts the target risk profile's `excluded_tools` from the resolved set after the `__` auto-admit union. The deny-list applies uniformly to both explicit allow-list entries and auto-admitted MCP names. Adds three regression tests, the first one using a real `<server>__<tool>` MCP name (`filesystem__write_file`) — exactly the destructive capability you called out — to exercise the new auto-admit branch and pin the deny-list block:
  - `resolve_allowed_tools_auto_admits_double_underscore_mcp_names`
  - `resolve_allowed_tools_honors_excluded_tools_for_auto_admitted_mcp` (the blocking-comment regression)
  - `resolve_allowed_tools_honors_excluded_tools_for_explicit_allow_list_entries`

  The pre-existing `mcp_tools_included_in_subagent_tool_list` fixture (which uses `mcp_fake`) is kept as-is so the explicit allow-list path stays covered.

- **7f6a070** `docs(security): document MCP allow-list contract change from #7547` — three surfaces:
  - `crates/zeroclaw-config/src/schema.rs`: rustdoc on `RiskProfileConfig::{allowed_tools,excluded_tools}` now spells out the `<server>__<tool>` auto-admit exception and the explicit-empty-list deny-all carve-out.
  - `docs/book/src/tools/overview.md`: extends the "Tool gating lives on the risk profile" bullets with the MCP exception and notes that `excluded_tools` is the way to block individual MCP tools.
  - `docs/book/src/tools/mcp.md`: new "Authorization: `allowed_tools` / `excluded_tools`" subsection under "Security and approval" that enumerates the four cases (empty/omitted, non-empty + MCP auto-admit, explicit `[]` = deny-all, `excluded_tools` subtracts) and explains the rationale so operators know the deny-list is now their primary lever for destructive MCP capabilities under an allow-list-pinned profile.

Also rewrote the PR body to:
- Drop the conflicting `risk: low` text (the auto-applied `risk: high` label is correct).
- Remove the "Marking as draft pending docs and user validation" line (docs are in, PR is non-draft).
- Fill out the **Rollback** section properly (it's required at `risk: high`) with targeted-revert / full-revert / forward-fix options.
- Update the **Compatibility** section to flag the one behavior-corrects edge: operators who relied on `excluded_tools` being silently ignored in the agentic-delegate path will now see those entries actually honored. The previous behavior contradicted the documented contract, so it's a bug fix, but it's worth calling out.

Full local validation on the new tip:

```
$ cargo fmt --all -- --check
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 35.46s
$ cargo test --workspace
    TOTAL: 8485 passed, 0 failed, 22 ignored
```

PTAL.

singlerider (commented):
Reviewed the current tip on the merits: the `resolve_allowed_tools` and `ToolAccessPolicy::is_tool_allowed` logic, the three new delegate regression tests, the `tool_search` deny-path test, and the schema/`mcp.md`/`overview.md` docs. Cross-checked against the agentic-delegate path (`execute_agentic` → `sub_tools`) and the eager-MCP main-agent path. Both of @Audacity88's blocking findings are genuinely addressed in code, not just in the PR body.

### ✅ Blocking 1 (excluded_tools in the agentic-delegate path): RESOLVED

`DelegateTool::resolve_allowed_tools` now unions `__` MCP names from `parent_tools` into the explicit allow-list and then `retain`s against the target profile's `excluded_tools`. The subtraction is uniform: it blocks both auto-admitted MCP names and explicit allow-list entries (the latter was silently ignored before, which the rustdoc correctly flags). The empty-list carve-out (`if base.is_empty() { return base }`) preserves deny-all. `resolve_allowed_tools_honors_excluded_tools_for_auto_admitted_mcp` pins exactly the escape hatch from the review, using the real `filesystem__write_file` name rather than the old `mcp_fake` fixture. The main-agent path in `tool_search.rs` enforces the same three cases (empty list = deny-all, `__` auto-admit, deny-list subtracts), so the two surfaces agree.

### ✅ Blocking 2 (document the contract change): RESOLVED

The new `mcp.md` "Authorization" subsection enumerates the four cases and, importantly, gives the recipe to restore the pre-change strict behavior (`allowed_tools` entry + per-sibling `excluded_tools`). The `RiskProfileConfig` rustdoc and `overview.md` mirror it. This is the high-risk default-deny → default-accept-unless-denied shift made legible to operators, which is what the finding required.

### 🟢 Auto-admit heuristic is consistent with existing code

The "name contains `__`" test for MCP-ness is a substring heuristic, so a future built-in containing `__` would be auto-admitted under an allow-list-pinned profile. I checked and this is not a new rule the PR invents; it matches the existing `eager_mcp_tool_allowed` convention, and the `excluded_tools` backstop covers the operator. Fine to land as-is; if the project ever wants a hard signal it should be a dedicated `is_mcp` flag on the tool, tracked separately rather than here.

### Verdict

On the merits this is correct, well-tested at both surfaces, and the documentation gap is closed. Both blocking items are addressed. I'm leaving this as a comment rather than flipping the verdict because @Audacity88's `CHANGES_REQUESTED` is the standing review; over to them to confirm and approve. From my read there is nothing further to fix.

singlerider (approved):

Audacity88 (changes requested):
Re-reviewed current head 19aa3f7 against the public PR diff and metadata, green public checks, the prior Audacity88 review, @singlerider's follow-up, the runtime MCP registration paths, `ToolAccessPolicy`, `DelegateTool::resolve_allowed_tools`, the cron `agent::run` call path, and the updated MCP/security docs. I did not rerun local validation. The two earlier blockers are genuinely addressed, but I found one remaining authorization issue in the main-agent MCP policy.

### ✅ Resolved — `excluded_tools` now applies to agentic delegates

The delegate path now resolves the target profile, unions runtime-discovered `<server>__<tool>` names only after a non-empty allow-list exists, and then subtracts `excluded_tools` from the final set. The new `filesystem__write_file` regression covers the exact destructive MCP wrapper called out in the earlier review, and the explicit allow-list subtraction test closes the broader "delegate ignored excluded_tools entirely" gap.

### ✅ Resolved — the MCP contract is documented where operators need it

The new `mcp.md` authorization section separates capability policy from approval policy, lists the empty/omitted, non-empty, explicit-empty, and deny-list cases, and gives a concrete strict-mode example using `filesystem__read_file` / `filesystem__write_file`. The schema rustdoc and tools overview now carry the same contract, so the high-risk allow-list exception is no longer hidden in code.

### 🔴 Blocking — MCP auto-admit bypasses caller-supplied per-run allow-lists

The built-in tool registry still goes through `apply_policy_tool_filter`, where the risk-profile gate and caller-supplied `allowed_tools` gate are separate `AND` checks. The MCP wrappers are added later, though, and they use `mcp_tool_access_policy` / `ToolAccessPolicy` instead of that helper. In that later path, `ToolAccessPolicy::from_security` collapses the risk-profile allow-list and the caller-supplied `allowed_tools` argument into one `allowed` vector, and `is_tool_allowed` then lets any name containing `__` through whenever that vector is non-empty and the name is not denied. That means the new MCP exception bypasses not only the risk-profile list, but also narrower per-run lists that callers intentionally pass into `agent::run`.

This is a real authorization regression because cron agent jobs pass `job.allowed_tools` into `agent::run`. A job narrowed to `allowed_tools = ["cron_add"]` can still surface runtime-discovered MCP tools such as `filesystem__write_file` unless each MCP tool is separately listed in `excluded_tools`. That contradicts the existing inheritance docs and loop comments that describe caller-supplied `allowed_tools` as an additional gate, and it makes per-job narrowing stop being a reliable capability boundary once an MCP server is configured.

Please keep the MCP auto-admit exception scoped to the risk-profile allow-list only, while still intersecting caller-supplied per-run allow-lists against MCP names. A regression should cover a caller/per-run allow-list that does not include `filesystem__write_file` and prove the MCP wrapper is not registered, surfaced through `tool_search`, or added to the delegate handle unless that caller list explicitly admits it.

singlerider (changes requested):
Re-reviewing at head `19aa3f7`, which is newer than my 06-14 approval (`7887c860`). @Audacity88 found a real authorization regression on this head that I missed, and I verified it. Withdrawing my approval and agreeing with his changes-requested.

### 🔴 Blocking — confirmed: MCP auto-admit bypasses caller-supplied per-run `allowed_tools`

The new `is_tool_allowed` short-circuits on `name.contains("__")`: any MCP-style name is auto-admitted whenever `self.allowed` is non-empty (and not denied). The problem is that `self.allowed` collapses the risk-profile allow-list **and** the caller-supplied per-run `allowed_tools` into one vector. So the exception relaxes far more than the risk-profile list it was meant for:

- A cron job passes `job.allowed_tools` into `agent::run`. A job narrowed to `allowed_tools = ["cron_add"]` makes `allowed = ["cron_add"]` — non-empty — and then the `contains("__")` arm auto-admits every runtime-discovered MCP tool, including `filesystem__write_file`, unless each is separately listed in `excluded_tools`.

That contradicts the inheritance docs and loop comments that describe caller-supplied `allowed_tools` as an additional gate, and it makes per-job narrowing stop being a reliable capability boundary the moment an MCP server is configured. This is a security regression, not a convenience gap — which is why my earlier approval was wrong; it was on a head before this surfaced, and I did not catch that the auto-admit also swallows the per-run list.

Scope the MCP auto-admit to the risk-profile allow-list only, and still intersect caller-supplied per-run allow-lists against MCP names so a narrowed `allowed_tools` is honored. Add a regression where a caller/per-run allow-list does not include `filesystem__write_file` and prove the MCP wrapper is not registered, not surfaced via `tool_search`, and not added to the delegate handle unless that caller list explicitly admits it.

### RESOLVED ✅ — the earlier delegate/docs items hold

Crediting what @Audacity88 already confirmed: `excluded_tools` now applies to agentic delegates (with the `filesystem__write_file` regression), and the `mcp.md` authorization section plus the schema/overview rustdoc document the capability-vs-approval split and the allow-list cases. Those are good.

Withdrawing my approval — the per-run allow-list bypass is a genuine authorization hole on this head. Endorsing @Audacity88's changes-requested.

tidux:
Thanks for the careful re-review @Audacity88 and @singlerider — the per-run-allow-list bypass you both flagged is fixed. Two new commits on top of `19aa3f7`:

- **5c9965d** `Merge branch 'master' into fix-mcp-tool-visibility-v2` — resolves the conflict introduced by #7608 (`fix(runtime): expose deferred MCP tools to delegates`), which refactored `DelegateTool::resolve_allowed_tools` into `resolve_tool_policy` returning `Option<SecurityPolicy>`. I kept master's cleaner shape and reintroduced this PR's MCP `<server>__<tool>` auto-admit via a new sibling helper `DelegateTool::delegate_admits_with_mcp` used by both the `needs_memory_tools` probe and the `sub_tools` filter in `execute_agentic`. The three earlier PR #7547 regression tests are rewritten to assert against `resolve_tool_policy` + `delegate_admits_with_mcp`:

  - `delegate_admits_with_mcp_auto_admits_double_underscore_mcp_names`
  - `delegate_admits_with_mcp_honors_excluded_tools_for_auto_admitted_mcp` (still uses the real `filesystem__write_file` name)
  - `delegate_admits_with_mcp_honors_excluded_tools_for_explicit_allow_list_entries`

  Master's `deferred_mcp_activation_updates_delegate_parent_tools` test from #7608 (and its `FinalOnlyModelProvider`) is preserved as-is. The unused `FilesystemWriteFileMcpTool` fixture is dropped — the helper-level tests call `delegate_admits_with_mcp` directly with the name as a string, so it isn't needed for the current coverage; it can come back cleanly if a future integration test wants a real `dyn Tool` with that name.

- **d259ce7** `fix(tools): scope MCP auto-admit to risk-profile gate, not per-run caller list` — the second-round blocker itself. `ToolAccessPolicy::from_security` was collapsing the risk-profile `allowed_tools` and the caller-supplied per-run `allowed_tools` into a single `allowed` vector, then `is_tool_allowed`'s `name.contains("__")` arm auto-admitted any MCP name whenever that combined vector was non-empty. So a cron job narrowed to `allowed_tools = ["cron_add"]` would still surface every runtime-discovered MCP tool — exactly the regression you walked through. The fix splits `ToolAccessPolicy` into two independent AND-ed allow-list gates:

  ```rust
  pub struct ToolAccessPolicy {
      pub allowed: Option<Vec<String>>,         // risk profile; `__` auto-admit applies here
      pub caller_allowed: Option<Vec<String>>,  // per-run; strict explicit-list, NO auto-admit
      pub denied: Option<Vec<String>>,          // deny-list still wins
  }
  ```

  `from_security` stops merging the two lists. `is_tool_allowed` evaluates them as independent AND conditions with the MCP exception scoped to the risk gate only — the same shape `apply_policy_tool_filter` already uses for the built-in registry.

  Regression tests pin all three surfaces:

  - **Eager MCP path** (`register_eager_mcp_tool_if_allowed`):
    - `eager_mcp_policy_caller_per_run_list_does_not_leak_mcp_auto_admit` — risk profile is wide (`["cron_add", "fs__read_file"]`, so the `__` auto-admit branch is live on the risk gate), caller narrows to `["cron_add"]`; `cron_add` is admitted, `filesystem__write_file` and `fs__read_file` are rejected.
  - **Deferred MCP path** (`mcp_allowed_tool_count` → `tool_search` registration gate):
    - `deferred_mcp_caller_per_run_list_does_not_leak_mcp_auto_admit` — same scenario; `mcp_allowed_tool_count` returns 0 so `tool_search` is not even added to the agent's tool list when the per-run list admits no MCP stub.
  - **`ToolAccessPolicy::is_tool_allowed` directly:**
    - `caller_allowed_per_run_gate_does_not_auto_admit_mcp_names`
    - `caller_allowed_per_run_gate_narrows_after_risk_profile_auto_admit`
    - `caller_allowed_per_run_gate_still_honors_denylist`

  I also fixed a misleading comment on the existing `eager_mcp_policy_allows_only_names_that_pass_policy_and_caller_gates` test — it said `github__search` was auto-admitted "even though not in caller list", but `github__search` was actually in the caller list the whole time, so the test was proving the risk-profile auto-admit, not the (broken) per-run bypass.

The delegate path doesn't take a caller-supplied per-run list (it uses the target risk profile directly), so it's already correct after the merge; no per-run-bypass coverage gap to plug there.

**Docs updated**:

- `crates/zeroclaw-config/src/schema.rs` rustdoc on `RiskProfileConfig::allowed_tools` now says explicitly that the `__` auto-admit applies only to this risk-profile field, not to caller-supplied per-run lists.
- `docs/book/src/tools/overview.md` gains a bullet making the same scope distinction explicit, with the cron-job narrowing example.
- `docs/book/src/tools/mcp.md` adds a paragraph under the "Authorization" subsection covering the same point.

**Validation on the new tip (`d259ce7`):**

```
$ cargo fmt --all -- --check
$ cargo test --workspace --lib --tests
    TOTAL: 8724 passed, 0 failed, 10 ignored
```

`cargo test --workspace` (full, including doctests) fails on my box with `Option 'default-theme' given more than once` from rustdoc — but it reproduces on a crate I haven't touched (`-p zeroclaw-api --doc`) and the `rustdocflags = ["--default-theme=ayu"]` line in `.cargo/config.toml` is unchanged, so it's a local environment / rustdoc-version interaction, not introduced by this PR. CI uses `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` which doesn't run doctests, so it's not a merge blocker.

Audacity88 (changes requested):
Re-reviewed current head 567fd2aa against the prior Audacity88 and singlerider blockers, the current diff, the follow-up docs-style-only commit, the cron `agent::run` path, eager MCP registration, deferred `tool_search`, the agentic delegate filter, the updated MCP/security docs, and current GitHub checks. I did not rerun local Cargo validation; the current GitHub check suite is still in progress, and Docs Style is now green on this head.

### ✅ Resolved — per-run allow-lists stay a strict second gate

The previous authorization blocker is addressed in the current code shape. `ToolAccessPolicy` now keeps the risk-profile allow-list and caller-supplied per-run allow-list as separate gates, and `is_tool_allowed` applies the `__` MCP auto-admit exception only to the risk-profile gate. The caller gate remains an exact-name intersection, so a cron job narrowed to `allowed_tools = ["cron_add"]` no longer gets `filesystem__write_file` just because the agent's risk profile would auto-admit MCP wrappers.

The coverage also hits the right surfaces: direct policy evaluation, eager MCP registration, and the deferred `tool_search` registration count. The agentic delegate path does not take the same caller-supplied per-run list, and its target-risk-profile deny-list subtraction still holds after the #7608 merge-shape rewrite.

### 🔴 Blocking — the documented `allowed_tools = []` deny-all escape hatch is not reachable from risk profiles

The new docs and schema rustdoc say a risk profile can use `allowed_tools = []` as an explicit deny-all escape hatch for MCP and built-in tools. That is not what the current risk-profile config path does.

`RiskProfileConfig::allowed_tools` is still a plain `Vec<String>` with a default empty vector, so deserialization does not preserve whether the operator omitted the field or wrote an explicit empty array. Then `SecurityPolicy::from_profiles` maps `risk_profile.allowed_tools.is_empty()` to `allowed_tools: None`, which `SecurityPolicy::is_tool_allowed` treats as unrestricted. `DelegateTool::resolve_tool_policy` does the same empty-vector-to-`None` conversion for the agentic delegate path.

So the public contract now promises three distinct states, omitted/unrestricted, non-empty allow-list, and explicit-empty deny-all, but the runtime only represents two states for `[risk_profiles.<alias>].allowed_tools`: empty/unrestricted and non-empty allow-list. That is a security-facing documentation mismatch in the same contract this PR is trying to make legible. Please either implement a real explicit-empty deny-all state for risk-profile config, or remove that claim from the docs/rustdoc/PR body and keep the contract honest as "empty means no authorization constraint."

### 🟢 What looks good — the policy split is now much easier to audit

The current `ToolAccessPolicy { allowed, caller_allowed, denied }` shape is the right direction for this class of bug. It makes the risk-profile exception and the per-run narrowing boundary visible in code instead of hiding them inside a merged vector, and the tests now describe the cron narrowing failure mode directly enough that future refactors have a useful tripwire.

tidux:
Fixed the latest docs contract mismatch in **12f7334** `docs(security): clarify risk-profile empty allowed_tools semantics`.

What changed:

- Removed the incorrect public-contract claim that risk-profile `allowed_tools = []` is an explicit deny-all escape hatch.
- `docs/book/src/tools/mcp.md` now says risk-profile `allowed_tools` empty or omitted means no authorization constraint, and explicitly notes TOML cannot distinguish an omitted field from `allowed_tools = []` for `RiskProfileConfig::allowed_tools`.
- `docs/book/src/tools/overview.md` mirrors that: empty or omitted risk-profile `allowed_tools` means no authorization constraint. The MCP `__` auto-admit still applies only when the list is non-empty.
- `RiskProfileConfig::allowed_tools` rustdoc now documents the same empty-vector-to-unrestricted mapping and points users who need deny-all behavior at caller-supplied per-run `allowed_tools` (which is preserved as `Some(vec![])` by `ToolAccessPolicy`) or explicit `excluded_tools` entries.
- Added `from_profiles_empty_allowed_tools_means_unrestricted_not_deny_all` to pin the code behavior Audacity88 called out: `SecurityPolicy::from_profiles` maps an empty risk-profile `allowed_tools` vector to `allowed_tools: None`, and `is_tool_allowed("filesystem__write_file")` is therefore true under that risk-profile state.
- Updated the PR body to remove the stale deny-all wording and clarify that commit `5a8c81147` is about the programmatic `ToolAccessPolicy` state, not a state reachable from risk-profile TOML config.

Validation on the new tip `12f7334`:

```bash
$ BASE_SHA=$(git merge-base origin/master HEAD) bash scripts/ci/docs_quality_gate.sh
No prose em-dashes in changed docs files.
Linting docs files: docs/book/src/tools/mcp.md docs/book/src/tools/overview.md
Summary: 0 error(s)

$ cargo fmt --all -- --check

$ cargo clippy -p zeroclaw-config --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 27.58s

$ cargo test -p zeroclaw-config --lib
    858 passed, 0 failed
```

Audacity88 (approved):
Re-reviewed current head `eac78a66` against the prior Audacity88 reviews at `6691bba`, `19aa3f7`, and `567fd2aa`, @singlerider's review history, the current public PR diff and metadata, `ToolAccessPolicy`, the eager MCP registration path, deferred `tool_search`, the agentic delegate path, the updated MCP/security docs, the PR body, and current GitHub check status. I did not rerun local Cargo validation; I relied on the public diff, the author's validation evidence, and current GitHub checks.

### ✅ Resolved — risk-profile `allowed_tools = []` is no longer documented as deny-all

The latest head fixes the remaining documentation/schema mismatch. `RiskProfileConfig::allowed_tools` is still a plain `Vec<String>`, so the TOML config cannot distinguish an omitted field from `allowed_tools = []`; both deserialize to an empty vector and `SecurityPolicy::from_profiles` maps that to `allowed_tools: None` / unrestricted. The docs now say exactly that instead of promising a third risk-profile state the runtime cannot represent.

The new `from_profiles_empty_allowed_tools_means_unrestricted_not_deny_all` regression also pins the current behavior, so the docs and runtime should stay aligned. That resolves my last blocking concern from `567fd2aa`.

### ✅ Resolved — per-run allow-lists remain a strict second gate

The earlier authorization blocker remains fixed after the master merge. `ToolAccessPolicy` keeps the risk-profile allow-list and the caller-supplied per-run allow-list in separate gates, so the MCP `__` auto-admit exception applies only to the risk-profile gate. A cron job narrowed to `allowed_tools = ["cron_add"]` still cannot surface `filesystem__write_file` unless that per-run list explicitly admits it.

The coverage is the right shape: direct policy evaluation, eager MCP registration, deferred `tool_search`, and the delegate MCP/excluded-tools path all have focused regressions for the failure modes raised in review.

### 🟡 Warning — @singlerider's active block remains

The branch is current with `master` from the public head/base state I reviewed, and all visible fresh GitHub checks are green on `eac78a66`, so I do not see a stale-base or CI concern from this review. This approval clears my prior blocking review from `567fd2aa`, but @singlerider's active changes-requested review remains in force until they update it or it is dismissed.

singlerider (approved):
Re-reviewing at head `7550d2d` (merge-only on top of `eac78a66`), the head where @Audacity88 cleared his block. I withdrew my approval at `19aa3f7` over a per-run allow-list bypass; the author's `d259ce7` fix landed since, so I re-checked the runtime against the local PR head rather than taking the diff on faith.

### RESOLVED ✅ — per-run allow-list bypass is fixed and pinned

My `19aa3f7` blocker was that `ToolAccessPolicy` collapsed the risk-profile allow-list and the caller-supplied per-run `allowed_tools` into one `allowed` vector, so the `name.contains("__")` MCP auto-admit relaxed both gates at once. A cron job narrowed to `allowed_tools = ["cron_add"]` would then surface `filesystem__write_file`.

`tool_search.rs` now keeps them as two separate fields, `allowed` (risk-profile, with the `__` exception) and `caller_allowed` (strict explicit-list intersection, no auto-admit). I verified on the PR head that `caller_allowed_per_run_gate_does_not_auto_admit_mcp_names`, `caller_allowed_per_run_gate_narrows_after_risk_profile_auto_admit`, and `caller_allowed_per_run_gate_still_honors_denylist` all pass, and that the named `filesystem__write_file` case is the one exercised. The delegate side mirrors it: `delegate_admits_with_mcp` scopes the `__` exception to the target risk profile's `allowed_tools` with `excluded_tools` applied last, and `agentic_mode_empty_allowed_tools_respects_excluded_tools` plus the three `delegate_admits_with_mcp_*` tests are green.

### RESOLVED ✅ — docs match the runtime

`mcp.md` and `overview.md` now state the `__` auto-admit is scoped to the risk-profile gate only and that caller-supplied per-run lists stay strict intersections, using the exact `allowed_tools = ["cron_add"]` narrowing example. The earlier `allowed_tools = []` deny-all overclaim is gone; both docs now say empty-or-omitted means no authorization constraint, and `from_profiles_empty_allowed_tools_means_unrestricted_not_deny_all` pins that. The earlier delegate `excluded_tools` and capability-vs-approval doc items hold.

Both my blockers are cleared and independently confirmed by @Audacity88's `eac78a66` approval. Restoring approval.

PR #6054, opened 2026-04-24
fix(skills): respect timeout_secs from SKILL.toml

RyanHoldren:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Added `timeout_secs: Option<u64>` to `SkillTool`, parsed from `[[tools]]` in `SKILL.toml` via serde. The field existed in the manifest shape but was not deserialized, so it silently had no effect.
  - Added `timeout_secs: u64` to `SkillShellTool`, resolved in `new()` with the existing 60s default and a minimum one-second floor for `timeout_secs = 0`.
  - Changed shell-tool execution to use the resolved per-tool timeout instead of the `SKILL_SHELL_TIMEOUT_SECS` constant, and updated timeout errors to report the configured duration.
  - Added process-tree cleanup for timed-out shell tools: Unix shells run in their own process group and are killed by process group on timeout; Windows uses a Job Object for timeout cleanup.
  - Added regression coverage for default timeout behavior, custom timeout behavior, zero-timeout clamping, timeout enforcement, and Unix grandchild cleanup after timeout.
- **Scope boundary:** Shell/script skill tools only. `SkillHttpTool` keeps its fixed `HTTP_TIMEOUT_SECS` client timeout; the constructor now documents that manifest `timeout_secs` is intentionally ignored for HTTP tools in this slice.
- **Blast radius:** Skills that omit `timeout_secs` continue to use the existing 60s default. Skills that set `timeout_secs` now get the configured shell/script timeout, and timed-out shell workloads are actively cleaned up instead of leaving child work behind.
- **Linked issue(s):** N/A
- **Labels:** `bug`, `risk: medium`, `size: S`, `dependencies`, `agent`, `channel`, `runtime`, `skills`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
GitHub Actions on refreshed current head 65c79a8608352f2a3512f0d23fe42feba8ee7718:
- Lint: pass
- Test: pass
- Check (all features): pass
- Check (no default features): pass
- Check (32-bit): pass
- Build x86_64-unknown-linux-gnu: pass
- Build aarch64-apple-darwin: pass
- Build x86_64-pc-windows-msvc: pass
- Benchmarks Compile: pass
- Security: pass
- CI Required Gate: pass

Reviewer-side focused regression:
time scripts/zc-cargo test -p zeroclaw-runtime tools::skill_tool::tests::skill_shell_tool_timeout_kills_process -- --nocapture

running 1 test
[...]

PR #7549, opened 2026-06-13
fix(plugins): align install/discovery paths and add legacy migration

alanpjohn:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `zeroclaw plugin install` wrote to `config.data_dir/plugins`, but the agent runtime, gateway, and skill discovery scanned `config.plugins.plugins_dir`. The two never matched, so CLI-installed WASM plugins were silently invisible to the agent (`Loaded 0 WASM plugin tools`) even though `plugin list`/`info` showed them installed. Reproduced on current `master`.
  - Single-sources the directory: a new `PluginsConfig::resolved_plugins_dir()` is now the one source of truth; the CLI, runtime, skills, and gateway all resolve through it via a new `PluginHost::from_plugins_dir`. This also fixes a latent bug where a custom `plugins_dir` not ending in `plugins` was silently ignored by the old `parent()`-then-re-append dance, and removes three hand-rolled tilde expanders.
  - Adds an opt-in `zeroclaw plugin migrate` command plus a detection warning (on `plugin list` and at agent startup) to relocate plugins stranded in the old locations.
- **Scope boundary:** No change to the plugin manifest format, signature verification, WASM execution, the `plugins.plugins_dir` config schema/default, or per-request discovery behavior. No automatic/at-startup file mutation — migration is user-invoked only.
- **Blast radius:** Plugin path resolution across CLI, runtime tool + skill discovery, and the gateway `GET …/plugins` listing — all behind the `plugins-wasm` build feature and gated on `plugins.enabled`. No database, network, auth, or channel changes.
- **Linked issue(s):** Closes #6254. Related #6746
- **Labels:** proposed `bug`, `config`, `plugins`, `runtime`, `gateway`, `core`, `size: M`, `risk: low` (maintainers to confirm; rollback fields filled below in case `risk: medium` is preferred).

## Validation Evidence

> The changed code is behind the `plugins-wasm` feature, so the default `cargo test` / `cargo clippy --all-targets` do not compile these paths. Ran feature-scoped equivalents in addition to `fmt`.

```bash
$ cargo fmt --all -- --check
# exit 0 — no diffs
```

```bash
$ cargo clippy -p zeroclaw-config -p zeroclaw-plugins -p zeroclaw-runtime -p zeroclaw-gateway --features plugins-wasm &&   cargo clippy --bin zeroclaw --features plugins-wasm
[...]

PR #7552, opened 2026-06-13
fix(skills): respect timeout_secs from SKILL.toml

RyanHoldren:
## Summary

- **What changed and why:**
  - `SkillShellTool` now respects a skill manifest's `timeout_secs` instead of always killing shell commands at the hardcoded 60s `SKILL_SHELL_TIMEOUT_SECS`. A long-running skill can never complete under the 60s cap.
  - Re-adds `SkillTool.timeout_secs: Option<u64>` (`#[serde(default)]`) and resolves it in the tool constructor as `tool.timeout_secs.unwrap_or(SKILL_SHELL_TIMEOUT_SECS).max(1)`; the resolved value drives the `tokio::time::timeout(...)` around `cmd.output()`.
  - This re-fixes #6054, whose `timeout_secs` plumbing was dropped when the schema-V3 / multi-agent rewrite (#6398) re-authored `skills/mod.rs` and `tools/skill_tool.rs` from scratch.
  - Adds unit tests for default-when-unset, manifest override, zero-clamp, and serde round-trip.
- **Scope boundary:** This PR only makes the timeout *duration* configurable. It deliberately does **not** change what happens when the timeout fires — the command future is still dropped via `cmd.output()` exactly as before (no `process_group`, no `kill_on_drop`, no process-group SIGKILL). It does not touch the HTTP/builtin/MCP skill tools, and does not alter the 60s default for skills that omit `timeout_secs`.
- **Blast radius:** Every `kind = "shell"` / `kind = "script"` skill tool. Behavior is unchanged for any skill that does not set `timeout_secs` (resolves to the same 60s default). The new struct field is additive and `#[serde(default)]`, so existing SKILL.toml manifests parse unchanged; the 17 in-tree `SkillTool { .. }` literals (agent / orchestrator / prompt / skill_http / skill_tool) gain an explicit `timeout_secs: None`.
- **Linked issue(s):** Related #6054 (original fix), Related #6398 (rewrite that dropped it).
- **Labels:** `type: bug`, `risk: low`, `size: S`.

```bash
cargo fmt --all -- --check
cargo test
```

- **Commands run and tail output:** Not run locally — Rust builds were not available in the authoring environment. The full GitHub Actions battery (Format, Lint/Clippy, Test, Check, and the Build matrix) ran green on this branch; CI is the validation signal here.
- **Beyond CI — what did you manually verify?** Reviewed the change against the existing `tools/shell.rs` timeout idiom and the surrounding `SkillShellTool` tests. Confirmed: the new field is `#[serde(default)]` so absent manifests deserialize to `None`; `.max(1)` prevents a zero/instant timeout; the resolved value (not the const) is now passed to both the `tokio::time::timeout` call and the timeout error string. Verified all 17 `SkillTool` struct literals were updated so the crate still type-checks.
- **If any command was intentionally skipped, why:** All three were skipped locally for the reason above (no local Rust builds permitted). They are not optional — CI must run them.

- If any `Yes`, describe the risk and mitigation: N/A. Note: allowing a manifest to raise its own command timeout lets a skill run a shell command longer than 60s; this is bounded by the same SecurityPolicy that already gates skill command execution and only applies to locally-installed skill manifests.

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `Yes` — adds an optional `timeout_secs` key to a SKILL.toml `[[tools]]` entry. It is optional and defaults to the prior 60s behavior when omitted.
- If `No` or `Yes` to either: exact upgrade steps for existing users: None required. Existing manifests work unchanged. To opt in, add `timeout_secs = <seconds>` to a shell/script tool entry in SKILL.toml.


WareWolf-MoonWall (commented):
The change is clean and correct. A few notes for the record.

### 🟢 What looks good — minimal, correct, and well-tested

`timeout_secs: Option<u64>` with `#[serde(default)]` is the right shape: absent manifests deserialize to `None` and get the unchanged 60s default; present manifests override it; `.max(1)` guards against a zero timeout killing every command. Resolving the value once at constructor time (`self.timeout_secs`) and threading it into both the `tokio::time::timeout` call and the error message is tighter than reading the field twice. Four unit tests cover the four meaningful cases: default fallback, manifest override, zero-clamp, and serde round-trip. Updating all 17 in-tree `SkillTool` struct literals is the correct mechanical follow-through.

### 🟡 Warning — Local validation not run; CI used as substitute

The PR body explicitly states that `cargo fmt`, `cargo clippy`, and `cargo test` were not run locally because Rust builds were not available in the authoring environment. CI ran all three commands against the full workspace and passed, which is meaningful evidence. The template language ("the signal CI cannot replace") is about catching environment-specific issues a remote runner misses — the practical risk is low here given the scope (+82/−3) and the green battery. That said, future PRs from this author should include local output as the template requires; CI as a substitute is an exception, not a pattern to establish.

### 🔵 Suggestion — Merge-ordering note with #7549

PR #7549 (currently open, no active blocks) also modifies `crates/zeroclaw-runtime/src/skills/mod.rs` (removes 17 lines in the same file). If #7549 lands first, this branch will need a rebase. Separately, #7399 (also open) adds a `SkillTool` struct literal in a test helper that does not include `timeout_secs` — whichever of the two lands second will need the field added. No action needed now; just be ready to rebase.

Audacity88 (approved):
@RyanHoldren I reviewed current head ece6d2117e6e0d92bcd45e6151e9f3d0380998df against the PR body, #6054, #6398, @WareWolf-MoonWall's comment review, the current diff, local source context, and the live GitHub checks. I did not rerun local cargo; I relied on the green GitHub Quality Gate for the full workspace checks.

This looks like a legitimate narrow restore of the `timeout_secs` manifest plumbing that was lost during the schema-V3 / multi-agent rewrite. I am comfortable approving the current slice as the configurable-duration fix.

### 🟢 What looks good — the manifest field is restored in the right place

`SkillTool.timeout_secs: Option<u64>` with `#[serde(default)]` is the right compatibility shape. Existing skill manifests still deserialize as `None` and keep the 60s default, while shell/script tools that opt in now have a path to raise the timeout. Resolving the value once in `SkillShellTool::new()` and using that same value for both the `tokio::time::timeout(...)` call and the timeout message keeps the behavior simple and auditable.

The zero clamp is also the right defensive edge case. A literal `timeout_secs = 0` should not create an instant-kill tool, and `.max(1)` gives that value a predictable minimum.

### 🟢 What looks good — the tests cover the core contract

The new tests cover the important contract points for this slice: unset manifests keep the default, explicit manifest values override it, zero clamps to one second, and serde round-trips the new field. The mechanical updates to the in-tree `SkillTool` literals also look complete in the current diff.

### 🟡 Warning — this restores timeout duration, not process-tree cleanup

The original #6054 eventually included process-tree cleanup after timeout. This PR intentionally restores the configurable timeout duration but not that cleanup. When `tokio::time::timeout(...)` elapses, the command future is still dropped through the existing `cmd.output()` path, without process-group cleanup.

I am not treating that as a blocker for this slice because the PR body calls the boundary out directly and the current change still fixes the broken `timeout_secs` duration behavior. I do think the timeout cleanup gap should stay visible as follow-up work if we want the full old #6054 behavior restored, especially because the timeout error still says the command "was killed" even though background work may survive the dropped future.

### 🔵 Suggestion — refresh before merge confidence

The GitHub Quality Gate is green on this head, but `master` has moved since the PR's base and there is same-file drift in `crates/zeroclaw-channels/src/orchestrator/mod.rs` and `crates/zeroclaw-runtime/src/agent/agent.rs`. I do not see a code blocker from that drift, but merge confidence should come from a fresh branch/CI refresh, an equivalent merge-result smoke, or an explicit stale-risk call before landing.

RyanHoldren:
> The original https://github.com/zeroclaw-labs/zeroclaw/pull/6054 eventually included process-tree cleanup after timeout. This PR intentionally restores the configurable timeout duration but not that cleanup. When tokio::time::timeout(...) elapses, the command future is still dropped through the existing cmd.output() path, without process-group cleanup.

Yes, I didn't close that gap, @Audacity88. I am happy to follow this up with another PR to restore that functionality, but I think it's appropriate to decouple the issues, since the process-tree **not** being cleaned up is a pre-existing bug in `HEAD`, regardless of the timeout?

RyanHoldren:
I rebased onto `HEAD`, as per @Audacity88's suggestion; no conflicts.

singlerider (approved):
Reviewed at head `8d6532f2` against current `master`, verifying the timeout is actually consumed. Approving.

### 🟢 The restored `timeout_secs` is wired through correctly and backward-compatible

`SkillTool.timeout_secs: Option<u64>` with `#[serde(default)]` keeps existing manifests deserializing to `None` and the unchanged 60s default. `SkillShellTool::new` resolves it once — `tool.timeout_secs.unwrap_or(SKILL_SHELL_TIMEOUT_SECS).max(1)` — so a manifest value overrides the default and a literal `timeout_secs = 0` clamps to 1s rather than creating an instant-kill tool. The resolved value feeds both the `tokio::time::timeout(Duration::from_secs(self.timeout_secs), cmd.output())` call and the timeout message, so the two can't drift. That restores the manifest plumbing lost in the schema-V3 rewrite for #6054/#6398.

### 🟢 Tests cover the contract and the literals are updated

The four tests pin the meaningful cases (unset → default, explicit override, zero → clamp-to-1, serde round-trip), and all 17 in-tree `SkillTool` struct literals across orchestrator/agent/prompt are updated to add `timeout_secs: None` — complete mechanical follow-through. CI is green and the branch is `MERGEABLE` / `CLEAN`.

### 🟡 Scoped boundary (per both reviewers)

This restores configurable timeout *duration*, not the process-tree cleanup the full #6054 eventually included — on timeout the command future is dropped through `cmd.output()` without process-group cleanup, so background work can survive even though the message says "killed." The PR body calls this out and it's reasonable as a follow-up; worth keeping visible.

### 🔵 Merge-ordering (per both reviewers)

Same-file drift with #7549 (removes lines in `skills/mod.rs`) and #7399 (adds a `SkillTool` literal without `timeout_secs`) — whichever lands second needs a trivial rebase/field-add. Not actionable now.

@Audacity88 approved at the prior head; the change is unchanged. Approving — minimal, correct, backward-compatible, well-tested.

PR #7457, opened 2026-06-10
fix(gateway): return JSON for unknown API fallback

Audacity88:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Makes the web dashboard SPA fallback path-aware so unknown `/api` and `/api/*` GETs return a JSON 404 instead of `index.html`.
  - Preserves the existing SPA fallback for dashboard routes such as `/config`.
  - Normalizes `gateway.path_prefix` before classifying API fallback paths, so prefixed deployments return the same JSON error shape for `{prefix}/api/*` misses.
  - Adds focused gateway regressions for `/api`, `/api/agents`, prefixed `/gw/api/agents`, normal `/config`, and the `/apiary` false-positive case.
- **Scope boundary:** This PR fixes the gateway fallback boundary only. It does not change source-install web asset copying, frontend JSON parsing/error copy, auth behavior, or add new API endpoints.
- **Blast radius:** Gateway HTTP fallback behavior for unknown GET paths. Known API routes continue to use their existing handlers; non-API dashboard routes continue to receive the SPA shell.
- **Linked issue(s):** Closes #6862. Closes #7253.
- **Labels:** `bug`, `risk: high`, `size: S`, `gateway`, `web`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
cargo fmt --all -- --check
Result: passed

cargo test -p zeroclaw-gateway spa_fallback --lib
Result: passed

running 5 tests
test tests::spa_fallback_returns_json_not_html_for_unknown_api_path ... ok
test tests::spa_fallback_returns_json_for_path_prefixed_api_miss ... ok
test tests::spa_fallback_returns_json_for_api_root_path ... ok
test tests::spa_fallback_does_not_treat_api_like_spa_paths_as_api ... ok
test tests::spa_fallback_still_serves_dashboard_routes ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 226 filtered out; finished in 0.07s
```

The focused test build also emitted an unrelated pre-existing warning in `zeroclaw-channels` for an unused `PacedChannel` import.

- **Beyond CI — what did you manually verify?** I verified the diff is limited to `crates/zeroclaw-gateway/src/static_files.rs` and gateway unit tests in `crates/zeroclaw-gateway/src/lib.rs`. The focused tests cover the reported `/api/*` HTML fallback failure, path-prefix handling, normal SPA fallback behavior, and a similarly named non-API route.
[...]

PR #7529, opened 2026-06-12
fix(gateway): only print dashboard URL when web_dist_dir is available

chengzhichao-xydt:
## Summary

When the Web dashboard is not found (no `web/dist`), the gateway still printed "🌐 Web Dashboard: http://..." at startup, which misleads users into expecting a working dashboard.

## Fix

- Only print the dashboard URL when `web_dist_dir` is available
- Print the supported installer guidance when the dashboard is not available

## Before
```
🦀 ZeroClaw Gateway listening on http://127.0.0.1:42617
  🌐 Web Dashboard: http://127.0.0.1:42617/    # <-- always shown!
```

## After (dashboard found)
```
🦀 ZeroClaw Gateway listening on http://127.0.0.1:42617
  🌐 Web Dashboard: http://127.0.0.1:42617/
```

## After (dashboard not found)
```
🦀 ZeroClaw Gateway listening on http://127.0.0.1:42617
  ⚠️  Web Dashboard: not available — reinstall with the supported installer (`./install.sh --source` on Linux/macOS, `setup.bat` on Windows) to build it
```

Fixes #7523

## Validation Evidence

```bash
cargo fmt --all -- --check
# (no output — formatting is clean)

cargo clippy --all-targets -- -D warnings
# Finished `dev` profile in 8.34s (no warnings)

cargo test
# all tests passing — CI green (Quality Gate)
```

- **Beyond CI — what did you manually verify?**
  - Verified the three scenarios: dashboard present, absent, and CLI output format
  - Confirmed no regression in the startup banner by checking both branches of the `if web_dist_dir.is_some()` guard
- **If any command was intentionally skipped, why:** None

## Security & Privacy Impact

- New permissions, capabilities, or file system access scope? **No**
- New external network calls? **No**
- Secrets / tokens / credentials handling changed? **No**
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No**
- If any `Yes`, describe the risk and mitigation: N/A

## Compatibility

- Backward compatible? **Yes** — no change to API, config, or CLI surface
- Config / env / CLI surface changed? **No**
- If `No` or `Yes` to either: exact upgrade steps for existing users: N/A

## Rollback

Gateway banner-output fix: `git revert <sha>` is the plan.

PR #7554, opened 2026-06-13
fix(install): install web dashboard on source installs and point users at the installer

### LINKED ISSUE #7553: [Bug]: source install never copies web/dist to the data dir, so the systemd daemon serves no dashboard

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Source installs built `web/dist` but never copied it into the gateway data directory, so a service-launched daemon resolved `web_dist_dir` to `None` and served no dashboard. This makes the supported source-install path (`./install.sh --source`) actually produce a working dashboard.
  - The prebuilt path already copied but resolved its target with an inline `case` block that could drift from the source path. Extracted `resolve_web_data_dir()` as the single source of truth used by both paths so they cannot diverge (the exact drift risk flagged on #7302).
  - Added Windows parity in `setup.bat`: source installs now `cargo web build` and copy `web/dist` to `%LOCALAPPDATA%\zeroclaw\web\dist`.
  - Stopped telling end users to run raw `cargo web build`. The gateway dashboard-unavailable 503 and the two startup log notes now point at the supported installer per platform (`./install.sh --source` on Linux/macOS, `setup.bat` on Windows).
- **Scope boundary:** Does not change the gateway's auto-detect logic, the `embedded-web` compile-time assert (developer-facing, keeps its cargo hint), or developer docs under `docs/book/src/developing/web.md` (correct audience for cargo invocations).
- **Blast radius:** Installer scripts (`install.sh`, `setup.bat`) and three user-facing gateway strings. No runtime behavior change beyond message wording; the install paths only add a copy step.
- **Linked issue(s):** Closes #7553. Supersedes #7302. Related #7457 (runtime JSON-fallback fix; complementary, addresses the symptom this PR removes the cause of).
- **Labels:** `bug`, `gateway`, `web`, `risk: medium`, `size: S`

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings
cargo test -p zeroclaw-gateway
bash -n install.sh && sh -n install.sh
```

  - `cargo fmt --all -- --check` — clean, no diff.
  - `cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings` — `Finished`, 0 warnings.
  - `cargo test -p zeroclaw-gateway` — `test result: ok. 250 passed; 0 failed; 0 ignored`.
  - `bash -n install.sh && sh -n install.sh` — both OK (POSIX-clean).
  - Unit-tested the extracted shell helpers in throwaway temp dirs: Linux default path (`$HOME/.local/share/zeroclaw/web/dist`, the #7253/#7553 repro), `XDG_DATA_HOME` honored, custom `PREFIX`, dry-run writes nothing and prints the would-install line, real recursive copy preserves nested assets, missing `index.html` is a no-op. 6/6 pass against the final `install.sh`.
  - Ran `install.sh --source --dry-run` and `install.sh --prebuilt --dry-run` with a temp `HOME`: confirmed no data-dir writes and correct messaging on both.
  - Verified the prebuilt path now resolves through the shared helper to the same `.local/share` location.
  - Did NOT run a full live source build/install end to end on a clean VM, nor execute `setup.bat` on Windows (no Windows host); the batch logic mirrors the verified `install.sh` copy step.
- **If any command was intentionally skipped, why:** Ran the gateway-scoped battery rather than the full workspace `cargo test` for turnaround; the diff only touches the gateway crate and the two shell installers.

- New permissions, capabilities, or file system access scope? **No** (the installer already writes to the data dir on the prebuilt path; this extends the same write to the source path).

## Compatibility (required)

- Config / env / CLI surface changed? **No** (installer flags and gateway config unchanged; only adds a copy step and rewords messages).
- If `No` or `Yes` to either: no upgrade steps required. Existing installs are unaffected; re-running the installer now also populates the dashboard.

- **Fast rollback command/path:** `git revert af4e5925d`
- **Observable failure symptoms:** Installer errors during the new copy step (grep installer output for `Web dashboard installed to`), or a source install that still serves no dashboard (gateway logs `Web dashboard: not available`). The copy is guarded on `index.html` existence and is a no-op when absent, so it cannot break an install that lacks a built bundle.

- Superseded PRs + authors:
  - #7302 by @hanZeng-08
- Scope materially carried forward: the source-install `web/dist` copy approach and its root-cause diagnosis originated in #7302. This PR carries that forward, corrects the path-resolution bug (the `$HOME/share` vs `$HOME/.local/share` expansion flagged in review), de-duplicates the directory resolution across both install paths, and extends the fix to Windows and the gateway messaging.
- `Co-authored-by` trailers added in commit messages for incorporated contributors? **No**
- If `No`, why: the final implementation is a distinct rewrite (shared helper, dry-run guard, Windows parity, message changes); #7302's diff was not taken verbatim. Attribution is credited here and on the superseded PR rather than via a commit trailer.

### DIFF
--- a/crates/zeroclaw-gateway/src/lib.rs
+++ b/crates/zeroclaw-gateway/src/lib.rs
@@ -1299,15 +1299,17 @@ pub async fn run_gateway(
             INFO,
             ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note),
             "Web dashboard: not available — configured gateway.web_dist_dir is missing on \
-             this machine and no fallback location was found. Build with `cargo web build` \
-             and point gateway.web_dist_dir at the resulting web/dist directory."
+             this machine and no fallback location was found. Reinstall with the supported \
+             installer (`./install.sh --source` on Linux/macOS, `setup.bat` on Windows) to \
+             build and place the dashboard where the gateway looks for it."
         );
     } else {
         ::zeroclaw_log::record!(
             INFO,
             ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note),
-            "Web dashboard: not available — no web/dist found. Build with `cargo web build` \
-             and point gateway.web_dist_dir at the resulting web/dist directory."
+            "Web dashboard: not available — no web/dist found. Reinstall with the supported \
+             installer (`./install.sh --source` on Linux/macOS, `setup.bat` on Windows) to \
+             build and place the dashboard where the gateway looks for it."
         );
     }
 
--- a/crates/zeroclaw-gateway/src/static_files.rs
+++ b/crates/zeroclaw-gateway/src/static_files.rs
@@ -50,11 +50,11 @@ pub async fn handle_spa_fallback(State(state): State<AppState>, uri: Uri) -> Res
     let Some(bytes) = load_index_html_bytes(state.web_dist_dir.as_ref()).await else {
         return (
             StatusCode::SERVICE_UNAVAILABLE,
-            "Web dashboard not available. Build the frontend with `cargo web build` \
-             (the supported entry point — it generates the TS API client and runs \
-             the Vite production build) and point gateway.web_dist_dir at the \
-             resulting web/dist. The daemon's API endpoints remain reachable \
-             independently of the dashboard.",
+            "Web dashboard not available. Reinstall with the supported installer \
+             so the dashboard is built and placed where the gateway looks for it: \
+             `./install.sh --source` on Linux/macOS, or `setup.bat` on Windows. \
+             The daemon's API endpoints remain reachable independently of the \
+             dashboard.",
         )
             .into_response();
     };
--- a/install.sh
+++ b/install.sh
@@ -276,17 +276,7 @@ install_prebuilt() {
   echo
 
   # Resolve platform-correct web data directory to match gateway auto-detect
-  case "$(uname -s)" in
-  Darwin)
-    web_data_dir="${HOME}/Library/Application Support/zeroclaw/web/dist"
-    ;;
-  MINGW* | CYGWIN* | MSYS*)
-    web_data_dir="${LOCALAPPDATA}/zeroclaw/web/dist"
-    ;;
-  *)
-    web_data_dir="${XDG_DATA_HOME:-${PREFIX}/.local/share}/zeroclaw/web/dist"
-    ;;
-  esac
+  web_data_dir=$(resolve_web_data_dir)
 
   if [ "$DRY_RUN" = true ]; then
     info "[dry-run] Would download $asset_url"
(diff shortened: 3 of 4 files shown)


singlerider:
Attribution note for the merger: this PR supersedes #7302, whose author's root-cause diagnosis and `install.sh` web/dist copy approach this work is built on. The commit carries a `Co-authored-by` trailer for that contributor — please preserve it in the squash-merge commit body so the co-authorship is retained on master.

WareWolf-MoonWall (approved):
This is the right fix, done correctly. The root-cause diagnosis from #7302 was sound; this PR carries it forward, corrects the path-resolution divergence, and extends the coverage to Windows and the gateway messaging.

### 🟢 What looks good — `resolve_web_data_dir()` is the correct SOT

Extracting the platform-dispatch block into `resolve_web_data_dir()` and having both the prebuilt and source install paths resolve through it eliminates the exact drift risk that was called out in #7302's review. The old inline `case` block in `install_prebuilt` was a copy of logic that now lives in one place. `install_web_dist()` is a clean helper that respects `DRY_RUN` and guards on `index.html` existence — it cannot break an install that lacks a built bundle.

### 🟢 What looks good — Gateway messaging consistently points at the installer

All three user-facing dashboard-unavailable surfaces (the two `record!` log events in `lib.rs` and the 503 response in `static_files.rs`) now give the same actionable instruction: run the supported installer. Telling users to run raw `cargo web build` was a leaky abstraction; these strings correctly use the public interface.

### 🟢 What looks good — Validation evidence is thorough

Format, gateway-scoped clippy, 250 gateway tests, and both `bash -n` / `sh -n` POSIX checks all clean. The unit tests on the shell helpers (Linux default path, `XDG_DATA_HOME`, custom `PREFIX`, dry-run guard, copy fidelity, missing `index.html` no-op) cover exactly the cases that would break silently in a CI-only environment.

### 🟡 Warning — `setup.bat` not live-tested on Windows

The PR body correctly discloses this. The batch logic is a faithful port of the verified shell steps, and the copy is guarded on `index.html` existence, so it cannot corrupt an install without a built bundle. Low actual risk, but worth a live validation on the next Windows CI run or a volunteer's machine before treating Windows source installs as fully supported.

### 🔵 Note — Relationship with #7529

PR #7529 (open, CHANGES_REQUESTED from me for template completeness) guards the startup `println!` on `web_dist_dir.is_some()` and adds a `⚠️ Web Dashboard: not available — build with \`cargo web build\`` message. If this PR lands first, #7529's warning message would become inconsistent with the installer-first guidance introduced here. If #7529 does land, it should adopt the same installer-pointing language that this PR establishes.

Issue #5947, opened 2026-04-20
[Feature]: schema v3 — batch breaking field migrations


singlerider:
> **Merge blocker.** All items in the checklist below must be complete before any PR in this batch is merged to master. No partial landings.

## Summary

Bump `CURRENT_SCHEMA_VERSION` to 3 and ship all pending breaking config field changes in a single coordinated migration, minimizing disruption to existing `zeroclaw.toml` files.

## Problem statement

Schema v2 has accumulated breaking field changes that cannot ship incrementally without forcing users through multiple migration passes. Two categories of debt exist:

1. **Type and shape changes** — fields whose type or key name changed in a way that a v2 config will silently misparse or fail to deserialize under the new code. These cannot be gated behind `#[serde(default)]` without lying about the field's semantics.

2. **Structural aliasing** — RFC #5890 establishes aliasing as a schema-wide pattern (`[channels.<type>.<alias>]`, `[providers.<type>.<alias>]`, etc.), converting every singleton block into a keyed table. This is a shape change to every existing config that uses channels or providers.

Shipping these piecemeal means users absorb migration friction repeatedly and the migration system accumulates one-off raw TOML patches rather than typed transformation paths (see #5889).

## Proposed solution

Hold all breaking field changes in-branch until this tracker is closed, then ship v2→v3 as a single `migration.rs` step. The migration must be:
- Loss-free: no config values are dropped
- Atomic: applied in one pass, not sequentially per-PR
- Typed: compound field transformations use a typed path, not raw TOML string manipulation (#5889)

**Breaking changes in scope (CHANGED/REMOVED fields only — no new additive fields):**

Mattermost channel (#5902):
- `bot_token: String → Option<String>` — was required, now optional
- `channel_id: String` removed; replaced by `channel_ids: Vec<String>` — migrate scalar value to single-element list

Multi-agent aliasing (RFC #5890, once accepted):
- `[channels.<type>]` → `[channels.<type>.default]`
- `[providers.<type>]` → `[providers.<type>.default]`
- `[mcp.servers]` list → `[mcp.servers.<alias>]` keyed table (alias from existing `name` field)
- `[storage]` / `[tunnels]` similarly

## Non-goals / out of scope

- New additive fields with `#[serde(default)]` — those ship independently, no tracker needed
- Runtime behavior changes — this tracker is config shape only
- v3 does not block on RFC #5890 being fully implemented; aliasing migration ships when the RFC is accepted and implementation PRs land

[...]

PR #7556, opened 2026-06-13
feat(config): declarative section grouping for the Config menu

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Section grouping is now a declarative cell in the one `sections!` table in `zeroclaw-config`. Each curated section names its display group (`Foundation`, `Tools`, …) in a new `group:` row field; a `SectionGroup` enum + `SECTION_GROUPS` order const + `section_group_for_key()` live alongside it. This is the single source of truth the gateway dashboard, the RPC `config/sections` handler, and the zerocode TUI all read.
  - **Moved grouping out of the gateway.** The gateway's hand-curated `section_group()` match (flagged in-code as *"hand-curated until v3 / #5947 lands a schema attribute that encodes grouping declaratively"*) is now a one-line delegate to `zeroclaw_config::sections::section_group_for_key()`.
  - **Filled the gap in the RPC path.** `handle_config_sections` (the path zerocode uses) previously sent `group: ""` for every section, so the TUI had no grouping at all. It now fills the real group label.
  - **Grouped rendering in the zerocode Config pane.** The section list is stably group-sorted at load (canonical dependency order preserved within each group) and renders dim bold group headers between buckets. Filtering and daemons that send no group label fall back to the flat all-sections list unchanged. Mouse clicks resolve through a draw-time display-row map so group headers are dead zones, not off-by-N selections, and respect the list scroll offset.
  - **Fixed five mis-grouped sections.** `providers.tts`, `providers.transcription`, `mcp.servers`, `mcp_bundles`, and `knowledge_bundles` previously fell into the gateway's `Other` catch-all; they now resolve to `Foundation` / `Tools` like their siblings. This is the one observable behavior change for the existing web dashboard.
  - **Related to #5947: schema-backed section grouping (no hand-list left).** The long tail of top-level config roots no longer lives in any hand-maintained match. The `Configurable` derive now harvests a `#[group = "..."]` field attribute (at all four nested-field shapes) into a generated `Config::nested_section_group(name)`, mirroring how `nested_section_help` is harvested from field docstrings. The 45 long-tail roots (`observability`, `gateway`, `cost`, the tool sections, vendor integrations, …) declare their group inline on the field; `section_group_for_key` resolves curated rows via the `sections!` table and everything else via the harvested attribute, then `Other`. The 85-key match is deleted. A `every_surfaced_root_has_a_group` drift test enumerates real roots from `prop_fields()` and fails (naming the root) if a non-allowlisted one lands in `Other` — so a future section added without a group is a test failure, not a silent catch-all.
- **Scope boundary:** No new sections, fields, tools, channels, or agent capabilities. Grouping is display-only metadata; the set of configurable sections and their values is unchanged. Group assignments replicate the prior hand-list exactly (the migration restored 8 roots — `claude_code`/`claude_code_runner`/`codex_cli`/`gemini_cli`/`opencode_cli` → Integrations, `sop`/`verifiable_intent` → Agent, `shell_tool` → Tools — that a partial sweep would have dropped; the no-regression test pins them).
- **Blast radius:** `zeroclaw-macros` (the `#[group]` harvester), `zeroclaw-config` (`sections!` table + enum/helpers + field annotations on `Config`), `zeroclaw-gateway` (delegates), `zeroclaw-runtime` (RPC handler fills the field), `zerocode` (TUI rendering + wire mirror). Group **labels** are byte-identical to what the gateway already emitted, so the web dashboard's `GROUP_ORDER` (`web/src/pages/Config.tsx`) needs no change. The macro change is additive (new generated method + helper attr); all existing `Configurable` structs compile unchanged.
- **Linked issue(s):** Related to #5947 (declarative section grouping via a schema attribute follow-up; #5947 was already closed by #6398 in v0.8.0-beta-1). Grouping is now declarative end-to-end: curated rows in the `sections!` table, the long tail on `#[group]` field attributes, zero hand-maintained mapping.
- **Known follow-up (intentionally out of scope):** `microsoft365` and the `file_upload`/`file_upload_bundle`/`file_download` tool sections are obvious grouping candidates (Integrations / Tools) but were left ungrouped (allowlisted) to keep this a faithful refactor — one-line `#[group]` each when someone wants them curated. Also ungrouped today: `escalation`, `locale`, `model_routes`, `embedding_routes`, `wss`.
- **Labels:** `enhancement`, `risk: high`, `size: L`, `config`, `gateway`, `runtime`, `provider:kilo`, `zerocode` (live labels at review time).

```bash
cargo fmt --all -- --check
cargo test
```

  - `cargo fmt --all -- --check` → clean (exit 0)
  - `cargo clippy --all-targets -- -D warnings` → **full workspace, 0 warnings / 0 errors** (`Finished` in 1m49s)
  - `cargo test` → **could not complete: `error: No space left on device (os error 28)`** while compiling test binaries — the build disk (`/opt/cargo-build`, 196G) hit 100%. Environment limit, not a code failure. Scoped suites for every touched crate pass on the same tree:
    - `cargo test -p zeroclaw-config --lib` → `ok. 834 passed; 0 failed` (incl. 12 `sections::` tests)
    - `cargo test -p zeroclaw-gateway api_sections` → `ok. 18 passed; 0 failed`
    - `cargo test -p zerocode` → `ok. 339 passed; 0 failed`
    - `cargo check -p zeroclawlabs --bin zeroclaw` → compiles (macro re-expands over every `Configurable` struct)
- **New tests (zeroclaw-config `sections::`):** `section_groups_const_is_exhaustive_unique_and_other_last`; `group_labels_are_pinned_for_ui_compat` (labels pinned as a wire contract vs the web `GROUP_ORDER`/TUI); `curated_sections_never_fall_into_other`; `every_surfaced_root_has_a_group` (drift guard — enumerates real roots from `prop_fields()`, fails naming any non-allowlisted root that resolves to `Other`); `migrated_hand_list_roots_keep_their_groups` (no-regression pin); plus kebab/snake resolution cases.
- **Beyond CI — manual verification:** Built `zeroclaw` + `zerocode` and ran `zerocode --config-dir <throwaway>` (auto-spawns the sibling daemon). Config pane shows Foundation first, then Agent/Tools/Integrations/Network/Storage/Operations with dim headers; `/` filter collapses to the flat list and restores on clear; clicking a header is a no-op while clicking a section selects exactly that row.
- **If any command was intentionally skipped, why:** the full-workspace `cargo test` aborted on `No space left on device` (build-disk full), not a test/compile error. `fmt` and `clippy --all-targets -D warnings` complete clean over the whole workspace; the per-crate suites above cover all touched code. CI will run the full battery on clean infra.

- Secrets / tokens / credentials handling changed? `No` — secret masking in `ConfigFieldEntry::from_prop_field` is untouched.
- If any `Yes`, describe the risk and mitigation: _(n/a)_

## Compatibility (required)

- Backward compatible? `Yes` — wire-compatible in both directions. The new `group` field has `#[serde(default)]`: a new zerocode against an old daemon receives `""` and renders the flat list; an old client against a new daemon ignores the extra field.
- Config / env / CLI surface changed? `No` — no new config keys, no on-disk format change. Grouping is derived metadata.
- If `No` or `Yes` to either: exact upgrade steps for existing users: None. To see grouping in the TUI, the daemon must be rebuilt and restarted (SIGTERM) so it emits `group`; until then the TUI shows the unchanged flat list.

## Rollback

- **Fast rollback command/path:** `git revert` (behavior is display/metadata-only, but the PR is labeled `risk: high` because it touches gateway/runtime/macro/config paths).
- **Feature flags or config toggles:** `None`. Grouping auto-disables (flat list) whenever the daemon sends no group labels, so a daemon-only rollback degrades the client gracefully.
- **Observable failure symptoms:** Worst case is cosmetic — sections render flat/ungrouped, or a section lands in the wrong group bucket. No effect on config values, picker dispatch, or editing.

## Design notes for reviewers

- **"Default all-sections view":** inline group headers over the full flat list (the default view still shows every section), not a drill-in group menu. The data layer (`SECTION_GROUPS`, per-entry `group`) supports a drill-in variant if preferred — contained follow-up in `draw_sections_pane`.
- **Two copies of the group order, by design:** `zeroclaw-config::sections::SECTION_GROUPS` is canonical. The web (`GROUP_ORDER` in `Config.tsx`) and the TUI (`group_rank` in `config_manager.rs`) each carry their own copy because they're wire clients of a remote daemon, not linkers of the crate — matching the existing dashboard pattern. `group_labels_are_pinned_for_ui_compat` guards drift.
- **Why `#[group]` carries a string, not a `SectionGroup`:** the derive macro lives in `zeroclaw-macros` and can't name `SectionGroup` (defined in `zeroclaw-config`, which depends on the macro — naming it would invert the dependency). So `nested_section_group` returns `Option<&'static str>` (the label, like `nested_section_help`) and `SectionGroup::from_label` validates the string on the way back in. A typo'd label resolves to `Other` and is caught by the drift test rather than at compile time — the one trade-off versus a compile-checked enum.

## Supersede Attribution

_(not applicable)_


singlerider (approved):
Read the whole change against current `master` source, walked the macro harvest path, and ground-truthed the grouping parity claims rather than taking the PR body's word for them. This is a clean refactor that actually closes #5947 the right way: grouping becomes declarative end to end (curated `sections!` rows carry a `group:` cell, the long tail declares `#[group = "..."]` on the field, the `Configurable` derive harvests it into `nested_section_group`), and the gateway's hand-curated 85-key `match` is deleted in favor of a one-line delegate. The single-source-of-truth direction is real, not cosmetic. The section explorer now walks the schema instead of shadowing it.

Verified the load-bearing claims:

- **No regression for live roots.** The old gateway `section_group()` match listed `"identity"` and `"agent"` as roots, but neither is a top-level `Config` field today (the `identity` field at schema.rs:3267 is `agents.<alias>.identity` inside `AliasedAgentConfig`). Both were dead arms; dropping them changes nothing. Every root `prop_fields()` actually surfaces resolves to the same group it did before, or to a real group it was previously missing.
- **The one behavior change is strictly an improvement.** `providers.tts`, `providers.transcription`, `mcp.servers`, `mcp_bundles`, and `knowledge_bundles` appear nowhere in the old gateway match, so they fell into `Other`. They now resolve to Foundation / Tools via their `sections!` rows. Confirmed against both trees.
- **Wire-label compatibility holds.** `SectionGroup::label()` emits byte-identical strings to the web `GROUP_ORDER` (web/src/pages/Config.tsx:56-66) and the TUI `group_rank` order, so the dashboard needs no change. `group_labels_are_pinned_for_ui_compat` pins this as a contract.
- **The macro harvest is structurally identical to the existing `nested_section_help` path.** Same four nested-field shapes (HashMap / Option / Vec / plain), each a mutually-exclusive branch, so no duplicate match arms. `extract_string_attr` parses `#[group = "..."]` correctly. The `group` helper attr is registered.

### 🟢 The drift guard earns its place

`every_surfaced_root_has_a_group` is the right shape: it enumerates real roots from `prop_fields()` and fails *naming the root* if a non-allowlisted one lands in `Other`, with a `HIDDEN` (system roots) / `UNGROUPED` (intentionally-uncurated) split that the test itself keeps honest: each `UNGROUPED` entry must still be a surfaced root and still ungrouped, so the allowlist can't rot silently. That makes "add a section without a group" a test failure instead of a silent catch-all, which is exactly the property #5947 was asking for.

### 🟢 Self-reviewed the click-path bug correctly

The dedicated `last_section_list_offset` is the correct fix, not over-engineering: `last_list_offset` is written at eight draw sites (config_manager.rs), only one of which is the sections pane, and draws aren't ordered, so a right-pane draw clobbers the shared offset before a section click is hit-tested. The section-pane handler now reads an offset written only by the section-pane draw. The `row_map: Vec<Option<usize>>` (None = header) makes headers dead zones and resolves both cursor and mouse through the map instead of assuming `row == section index`. Grouped rendering is gated on `filter.is_none() && any non-empty group`, so filtering and pre-group daemons keep the flat path unchanged.

The two intentional copies of the group order (web `GROUP_ORDER`, TUI `group_rank`) are the established wire-client pattern, not a registry shadow. They're remote clients of the daemon, not linkers of the crate, and `group_labels_are_pinned_for_ui_compat` guards the drift. The honest disclosure that `nested_section_group` returns `Option<&'static str>` (a label string the macro can't type-check, validated by `from_label` on the way back) rather than a compile-checked enum is the one accepted trade-off, and the drift test backstops the typo case.

CI 16/16 green, MERGEABLE, no open threads, authorship clean. The `cargo test` disk-full disclosure in the validation section is fine: `fmt` and `clippy --all-targets -D warnings` ran clean over the full workspace and the per-crate suites cover every touched crate; CI re-ran the full battery green.

Approving.

Audacity88 (approved):
I checked current head `6eabe9225d8fc64045deaff433c4d9ad72051e26`, the PR body, linked #5947 context, SingleRider's approval, the touched config/macro/gateway/runtime/zerocode paths, and the visible green CI. I focused on whether grouping now has one source of truth, whether old/new daemon and client pairs degrade safely, and whether grouped TUI rows preserve click and scroll selection.

### 🟢 What looks good — Grouping is config-owned now

The single-source direction holds. Curated rows get their group from the `sections!` table, the long tail declares `#[group = "..."]` on schema fields, the derive macro harvests that into `nested_section_group`, and both the gateway and RPC `config/sections` path delegate back to `zeroclaw-config`. That removes the gateway hand-list instead of moving it somewhere else, and the drift tests cover the important failure mode: a surfaced root that silently falls into `Other`.

### 🟢 What looks good — Compatibility and TUI behavior are covered

The wire change is additive: `zerocode` defaults missing `group` values to `""`, so an old daemon still renders the flat list, while a new daemon fills the group label for current clients. The TUI grouping path is careful about selection: the grouped display builds a row map, headers resolve to `None`, clicks use the section-list offset from the section pane itself, and filtered mode keeps the flat list. That matches the manual verification described in the PR body and the visible CI is green.

Approving.

WareWolf-MoonWall (approved):
Reviewed at 6eabe922. @singlerider and @Audacity88 both approved at this head and no active CHANGES_REQUESTED are open. I've gone through the full diff — 8 commits, +593/-70 across `zeroclaw-config`, `zeroclaw-macros`, `zeroclaw-gateway`, `zeroclaw-runtime`, and `apps/zerocode` — and I agree with the verdict. This is a well-executed SOT consolidation with two drift guards and backward-compatible wire serialization. Adding my approval.

### 🟢 What looks good — the gateway's 85-key match-arm is gone, and both paths now delegate to zeroclaw-config

The before state had group information in two places: the gateway had its own static match on every config key, and the macro had no visibility into groups at all. The after state has one location (`crates/zeroclaw-config/src/sections.rs`) that defines `SectionGroup`, `SECTION_GROUPS`, and `section_group_for_key`, and every surface — gateway, runtime RPC dispatch, and the macro harvester — resolves from there. Deleting 85 match arms from `api_sections.rs` and having the `#[group]` attribute harvested at derive time are both consequences of getting the SOT right. That's the hard part of this PR and it's done cleanly.

### 🟢 What looks good — `every_surfaced_root_has_a_group` is the right enforcement mechanism

Drift tests that assert a structural invariant at test time are more durable than doc comments or review reminders. If someone adds a new config section without a `#[group]` annotation, this test fails loudly instead of silently falling back to `Other` and making it to production. The `group_labels_are_pinned_for_ui_compat` test is the complementary guard: it ties the wire contract to a specific set of string values so a rename of a `SectionGroup` variant doesn't silently break clients that are already deployed. Both tests are doing real enforcement work.

### 🔵 Suggestion — the `group_rank` helper in `config_manager.rs` is intentionally a local copy and that's fine

The PR description explains the reasoning and I think it's correct: `zerocode` is a wire client of a remote daemon, not a linker of the `zeroclaw-config` crate. Its draw-time sort order is presentation logic that belongs to the client, the same way a web dashboard would own its own column-sort logic rather than importing the daemon's domain type. The `group_labels_are_pinned_for_ui_compat` test is the contract that ensures `group_rank` stays consistent with what the daemon actually emits — if the daemon's group label changes, the test breaks before the presentation diverges silently. This isn't a DRY violation; it's a client holding a stable contract against a tested wire value. No change needed, just noting it explicitly for anyone reading the diff who might flag it as a duplicate.

Approving. This is ready to land.

tidux:
@Nillth I'd like to see some screenshots in the main body for a UI change PR.

tidux (approved):
Reviewed at 6eabe9225d8fc64045deaff433c4d9ad72051e26. @singlerider, @Audacity88, and @WareWolf-MoonWall have all approved at this head; no active CHANGES_REQUESTED. I walked the full diff (8 commits, +593/-70 across `zeroclaw-config`, `zeroclaw-macros`, `zeroclaw-gateway`, `zeroclaw-runtime`, `apps/zerocode`), cross-checked the macro harvest path, ran the per-crate test batteries locally on this head against current `upstream/master`, and ground-truthed the load-bearing parity claims rather than just inheriting prior verdicts. Adding my approval.

The thing that makes this PR land cleanly is that it actually consolidates rather than just relocates: the gateway's 85-key hand `match` is deleted, the curated `sections!` rows grow a `group:` cell, the long tail of top-level config roots declare their group inline via `#[group = "..."]` on the schema field, and the `Configurable` derive harvests that into `Config::nested_section_group(name)` mirroring the existing `nested_section_help` pattern. After the change, there is exactly one place that answers "what group does this section render under" — `zeroclaw_config::sections::section_group_for_key` — and the gateway, the RPC `config/sections` handler, and the zerocode TUI all route through it. That closes #5947 the way #5947 was actually asking to be closed.

### 🟢 What looks good — the macro harvest plugs grouping into the existing nested-field pipeline

`zeroclaw_macros::derive_configurable` already harvests docstrings into `nested_section_help` at four mutually-exclusive `#[nested]` field shapes (`HashMap`, `Option`, `Vec`, plain). The new `#[group]` harvest sits exactly parallel at each of those four branches (lines 444, 1167, 1358, 1807) and emits one match arm per annotated field into the generated `nested_section_group(name)` lookup. The `group` helper attr is registered alongside `tab` at the `#[proc_macro_derive(... attributes(...))]` declaration, so existing `Configurable` structs compile unchanged whether they carry `#[group]` or not. Structurally identical to the help path means the same compile-time guarantees on shape; the runtime cost is one extra `match` arm per annotated field.

### 🟢 What looks good — `every_surfaced_root_has_a_group` is the right drift guard

The test enumerates real top-level roots from `prop_fields()` (not a hand-list), splits them into `HIDDEN` (system/bookkeeping that the explorer never shows) and `UNGROUPED` (intentionally-uncurated allowlist), and fails *naming the offending root* if any non-allowlisted root resolves to `SectionGroup::Other`. The reverse cross-check at the bottom — every `UNGROUPED` entry must still be a real surfaced root AND must still resolve to `Other` — keeps the allowlist from rotting silently as schema shifts. This is what makes grouping declarative-by-construction: adding a top-level `#[nested]` field without a `#[group]` annotation is a test failure, not a silent catch-all. Paired with `migrated_hand_list_roots_keep_their_groups` (which pins every formerly-hand-listed root to the same group it had in the gateway match) the regression surface is genuinely closed.

### 🟢 What looks good — the wire change is additive and degrades gracefully both ways

`ConfigSectionEntry.group` is `#[serde(default)]`, so a new zerocode against an old daemon receives `""` for every entry, `self.sections.iter().any(|s| !s.group.is_empty())` returns false, `grouped` stays false in `draw_sections_pane`, and the section list renders flat exactly as before. An old client against a new daemon ignores the extra field. The PR exercises both halves of that without needing a feature flag, which is the right shape for a metadata-only display change. Verified the rollback property by inspection: nothing in the right pane, picker dispatch, or section-value editing reads `group`; the worst rollback symptom is cosmetic.

### 🟢 What looks good — the click-path fix is correctly scoped to the section pane

The dedicated `last_section_list_offset` (separate from the shared `last_list_offset`) is the right fix, not over-engineering. `last_list_offset` is written by every list-style draw site in `config_manager.rs`; draws aren't ordered, so a right-pane draw clobbers the shared offset before a section-pane click is hit-tested. Reading from a field written only by `draw_sections_pane` makes the hit-test correct regardless of which screen drew last. `row_map: Vec<Option<usize>>` where `None` marks a header row means the cursor and mouse-click handlers both resolve through `row_map.get(pos)`, so headers are dead zones rather than off-by-N selection bugs and inserting headers between groups doesn't shift section indices anywhere else. Grouping is also gated on `filter.is_none() && any non-empty group`, which keeps the flat code path live for filtered mode and pre-group daemons unchanged — the new path and the old path don't fight each other.

### 🟢 What looks good — wire-label compatibility is pinned as a contract, not a convention

`group_labels_are_pinned_for_ui_compat` asserts that `SectionGroup::label()` produces exactly the strings the dashboard's `GROUP_ORDER` (web/src/pages/Config.tsx:56-66) and the zerocode `group_rank` array expect. Confirmed by inspection that the web `GROUP_ORDER` is byte-identical: `["Foundation", "Agent", "Multi-agent", "Tools", "Integrations", "Network", "Storage", "Operations", "Other"]`. The two intentional copies of the order (web `GROUP_ORDER` and zerocode `group_rank`) are the established wire-client pattern — both consume a remote daemon over JSON, neither links the crate — and the pinning test catches a rename before clients diverge silently. The honest disclosure in the design-notes section about `nested_section_group` returning `Option<&'static str>` (a label the macro can't type-check, validated by `from_label`) is the one accepted trade-off and the drift test backstops it.

### 🟢 What looks good — the observable behavior change is strictly an improvement

`providers.tts`, `providers.transcription`, `mcp.servers`, `mcp_bundles`, and `knowledge_bundles` were in `Other` under the pre-PR gateway match (they appear nowhere in those match arms, confirmed by grepping the pre-PR `api_sections.rs`). They now resolve to `Foundation` / `Tools` via their curated `sections!` rows. That is the dashboard regaining grouping it should always have had, and the new tests pin it. `Section::from_key("identity")` returns `None` and there is no top-level `identity` field on `Config` (the `identity` field at schema.rs:3267 lives inside `AliasedAgentConfig`, so its surfaced key is `agents.<alias>.identity`), confirming `singlerider`'s observation that the pre-PR gateway match's `"identity" => Storage` arm was dead code; dropping it has zero effect.

### 🔵 Suggestion — the four `UNGROUPED` candidates already named in the test

`microsoft365`, `file_upload`, `file_upload_bundle`, and `file_download` are called out in both the PR body and the inline comment on `UNGROUPED` as obvious grouping candidates (`Integrations` / `Tools` respectively). Since the harvest path is now in place, each is a one-line `#[group = "..."]` on the corresponding schema field, plus removing it from `UNGROUPED`. The drift test would catch a stale `UNGROUPED` entry on its way back through. Not blocking this PR — keeping the refactor faithful (no behavior change for the no-curation set) is the right call — but the follow-up is small and self-contained.

### 🔵 Suggestion — `escalation`, `locale`, `model_routes`, `embedding_routes`, `wss` deserve a parallel pass

Same pattern as the four above. `escalation` reads as `Operations`, `locale` as `Foundation`, `model_routes`/`embedding_routes` as `Agent`, and `wss` as `Network` from inspection of their schema docstrings, but the right call is to land them in a separate `feat(config): annotate remaining roots with #[group]` follow-up once someone confirms placement with @WareWolf-MoonWall or @singlerider. The infrastructure for getting them out of `Other` is now there and the cost-per-root is one line.

### Local verification at 6eabe9225

- `cargo test -p zeroclaw-config --lib -- sections::` → **12 passed; 0 failed** (including `section_groups_const_is_exhaustive_unique_and_other_last`, `group_labels_are_pinned_for_ui_compat`, `curated_sections_never_fall_into_other`, `section_group_for_key_resolves_curated_tail_and_unknown`, `every_surfaced_root_has_a_group`, `migrated_hand_list_roots_keep_their_groups`)
- `cargo test -p zeroclaw-gateway api_sections` → **18 passed; 0 failed**
- `cargo clippy -p zeroclaw-config -p zeroclaw-macros -p zeroclaw-gateway -p zeroclaw-runtime -p zerocode --all-targets -- -D warnings` → clean (Finished in 1m 08s)
- `git merge-tree --write-tree upstream/master pr-7556` → clean tree, zero conflict markers (PR is genuinely CLEAN against `03eb43733`)

CI shows 16/16 Quality Gate jobs green at this head; `mergeable=MERGEABLE` / `mergeStateStatus=CLEAN` / `reviewDecision=APPROVED`. Labels (`enhancement`, `size: L`, `risk: high`, `config`, `gateway`, `runtime`, `provider:kilo`, `zerocode`) match the diff scope. The PR body fills every required template section, including the `risk: high` rollback block, and the disk-full `cargo test` disclosure is honest — `fmt` and `clippy --all-targets -D warnings` ran clean over the full workspace and the per-crate batteries cover every touched crate, with CI re-running the full battery green.

Approving. Ready to land when @singlerider is ready.

Issue #7069, opened 2026-06-02
Twitter/X channel not available in pre-built binary despite channel-twitter feature existing


theonlyhennygod:
### Affected component

channel

### Severity

S3 — missing functionality

### Current behavior

The `channel-twitter` feature and `crates/zeroclaw-channels/src/twitter.rs` implementation exist in the source tree, and the docs at `docs/book/src/channels/social.md` describe Twitter/X as a supported channel. However, the pre-built binary (installed via the official install script) does not include it:

```
$ zeroclaw channel list
  ✅ Telegram
  ❌ Discord
  ❌ Slack
  ❌ Signal
  ❌ Reddit
  ❌ Bluesky
  # Twitter/X is not listed at all
```

```
$ zeroclaw config list | grep twitter
# (no results)
```

### Expected behavior

Twitter/X should appear in `zeroclaw channel list` and `zeroclaw config list` output, and be configurable via `[channels.twitter]` in `config.toml`, since the implementation exists and the docs describe it as enabled by default.

### Steps to reproduce

1. Install zeroclaw v0.8.0-beta-1 via the official install script on Ubuntu 26.04
2. Run `zeroclaw channel list`
3. Observe Twitter/X is absent

### Possible cause

`channel-twitter` may not be included in the `default` feature set in `Cargo.toml`, or the release CI pipeline compiles without it.

### ZeroClaw version

```
zeroclaw 0.8.0-beta-1
```

### Operating system

Ubuntu 26.04 LTS (x86_64)

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, personal paths, and personal data from all submitted content.

### Duplicate search

No exact duplicate found. Issue #6158 references `twitter` as a feature-gated module but that issue is about `--no-default-features` compilation, not the pre-built binary missing the channel.

Audacity88:
## Triage Summary

**Type:** Bug / release-bundle expectation mismatch  
**Priority:** Medium  
**Area:** `channel`, `channel:twitter`, `docs`  
**Risk:** High

**Summary:** This report is useful, but I would separate the implementation state from the prebuilt-binary expectation. The Twitter/X channel implementation and `channel-twitter` feature exist, but current docs also say prebuilt binaries use the lean default channel bundle rather than every bundled channel.

[...]

Issue #6895, opened 2026-05-24
feat(channels): define a lean default channel set


Audacity88:
# feat(channels): define a lean default channel set

## Summary

Define and measure a lean default channel feature set so ZeroClaw does not keep adding long-tail channel integrations to the core default build by inertia.

## Problem statement

#6165 asks whether ZeroClaw should stay lean by moving long-tail integrations out of core and toward skills, MCP servers, CLI-backed integrations, or optional/plugin surfaces. That broader issue is already RFC-shaped. This issue is the first concrete follow-up for the channel side.

#6680 is a useful example. The WeCom AI Bot long-connection channel is a distinct protocol and a reasonable addition, but it also increases the default channel surface when it lands as a default feature. The current default runtime channel set is already broad: email, Telegram, Lark, Discord, Slack, Signal, Mattermost, IRC, iMessage, DingTalk, QQ, Bluesky, Twitter/X, Reddit, Notion, Linq, WATI, Nextcloud Talk, MoChat, WeCom webhook, ClawdTalk, webhook, ACP server, WhatsApp Cloud, voice call, and, with PR #6680, WeCom WS. Current local source measurement shows about 36.3k Rust lines in the counted default channel implementation files; with PR #6680, that count is about 39.9k Rust lines. That is not the same as exact binary-size, compile-time, or dependency-size impact, but it is enough source/build-graph and maintenance surface to justify a focused follow-up.

#6770 and #6866 are adjacent work, but separate from this issue. They make selective channel builds and `zeroclaw channel list` truthful without changing the default channel set. This issue is about deciding and measuring whether the default set itself should become leaner.

There is also a concrete build-graph trap: `zeroclaw-gateway` currently depends on `zeroclaw-channels` with `default-features = true` while also naming only a few channel features. If we only narrow the root `agent-runtime` defaults, gateway builds may still pull the broad channel defaults.

## Proposed solution

Create the first #6165 slice for channels and channel integrations. Start with a draft PR or measurement note that proves the build graph and proposes the smallest safe default-set change:

- Measure default-build impact before and after changing channel defaults, including at least source/build graph, compile time where practical, and release binary size if feasible.
- Define a small default channel set for core/runtime builds. Candidate defaults should be channels needed for local operation, onboarding, gateway/API use, or common first-run behavior, not every bundled platform integration.
- Move long-tail platform channels to opt-in Cargo features while preserving documented install/build paths for users who need the fuller integration set.
[...]

PR #7558, opened 2026-06-13
feat(xtask): canonical install spec + cargo generate installers (drift elimination)

### LINKED ISSUE #3642: [Feature]: Provide a "full" docker image

### LINKED ISSUE #5995: Web dashboard missing from Docker release image (v0.7.3, :latest)
## Summary

The published `ghcr.io/zeroclaw-labs/zeroclaw:v0.7.3` and `ghcr.io/zeroclaw-labs/zeroclaw:latest` images serve **HTTP 503** on `GET /` instead of the web dashboard. The dashboard is missing from the image entirely — `/zeroclaw-data/web/dist` does not exist inside the container.

This is a regression from the dashboard decoupling in #5665. The companion distribution fix #5675 covered binary releases, AUR, and `cargo install`, but missed the Docker path.

## Reproduce

```sh
docker run --rm -d --name zc-repro -p 42617:42617 ghcr.io/zeroclaw-labs/zeroclaw:v0.7.3
sleep 4
curl -sf -o /dev/null -w "GET /        → HTTP %{http_code}\n" http://127.0.0.1:42617/
docker logs zc-repro 2>&1 | grep -i dashboard
docker cp zc-repro:/zeroclaw-data/web /tmp/zc-web 2>&1 || true
docker stop zc-repro >/dev/null
```

Output:
```
GET /        → HTTP 503
INFO zeroclaw_gateway: Web dashboard: not available (set gateway.web_dist_dir or ZEROCLAW_WEB_DIST_DIR)
Error response from daemon: Could not find the file /zeroclaw-data/web in container zc-repro
```

`GET /` returns:
```
Web dashboard not available. Build it with: cd web && npm ci && npm run build
```

## Expected

`GET /` → HTTP 200 with the dashboard HTML, matching the behaviour documented in the v0.7.3 release notes:

> **Web dashboard decoupled from the binary** — The dashboard is now built separately and embedded at release time. `cargo install` and AUR/Homebrew packages include it.

## Root cause

The release tarball for each arch **does** include the web bundle:

```sh
tar tzf zeroclaw-x86_64-unknown-linux-gnu.tar.gz | head
# zeroclaw
# web/dist/
# web/dist/assets/
# web/dist/assets/index-0moaj04v.js
# web/dist/index.html
# web/dist/logo.png
```

CI extracts it into `docker-ctx/bin/${TARGETARCH}/` (see [`.github/workflows/release-stable-manual.yml`](https://github.com/zeroclaw-labs/zeroclaw/blob/master/.github/workflows/release-stable-manual.yml#L517-L521)), so `docker-ctx/bin/amd64/web/dist/` is present in the build context. But [`Dockerfile.ci`](https://github.com/zeroclaw-labs/zeroclaw/blob/master/Dockerfile.ci#L11) only copies the binary:

```dockerfile
COPY bin/${TARGETARCH}/zeroclaw /usr/local/bin/zeroclaw
```

`web/dist` is silently discarded. The same applies to `Dockerfile.debian.ci`.

Additionally, the baked `config.toml` written by the docker job (workflow [lines 524-536](https://github.com/zeroclaw-labs/zeroclaw/blob/master/.github/workflows/release-stable-manual.yml#L524-L536)) does not set `gateway.web_dist_dir`, unlike the local `Dockerfile` which sets it to `/zeroclaw-data/web/dist` on [line 89](https://github.com/zeroclaw-labs/zeroclaw/blob/master/Dockerfile#L89).

Note this is distinct from the companion fragility fix #5924 ("ci: fix Dockerfile COPY fragility"), which landed after v0.7.3 but touched the multi-stage `Dockerfile`, not the CI images.

## Proposed fix

Opened as PR: add `COPY bin/${TARGETARCH}/web/dist /zeroclaw-data/web/dist` to both `Dockerfile.ci` and `Dockerfile.debian.ci`, and add `web_dist_dir = "/zeroclaw-data/web/dist"` to the baked config in the workflow. Validated locally by reproducing the CI docker context with the v0.7.3 release tarball and rebuilding with the patch — `GET /` → 200, `/_app/assets/index-*.js` → 200 (1.07 MB), `/logo.png` → 200, `Web dashboard: serving from /zeroclaw-data/web/dist` in logs.

## Environment

- Affected images: `ghcr.io/zeroclaw-labs/zeroclaw:v0.7.3`, `ghcr.io/zeroclaw-labs/zeroclaw:latest`
- Image digest (v0.7.3, checked today): `sha256:c364ffdf6d41e467bc22ef5a298c7395862ea6a563c5c396a57bba69dbfb7973`
- Host: Docker Desktop 4.x on macOS (Darwin 25.4.0), linux/amd64

### LINKED ISSUE #6836: [Bug]: setup.bat --minimal produces ~26 MB build instead of ~6 MB on Windows
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

I followed the official documentation for a minimal Windows build using `setup.bat --minimal`. However, the resulting build artifacts are approximately 26 MB, not the expected ~6 MB as stated in the docs.

It appears the `--minimal` flag might not be correctly filtering out components or reducing the binary size as documented. Could you please look into this discrepancy?

### Expected behavior

~6 MB.

### Steps to reproduce

```bash
1. Run `setup.bat --minimal` on Windows.
2. Check the size of the build output.
```

### Impact

_No response_

### Logs / stack traces

```text

```

### ZeroClaw version

0.8.0

### Rust version

rustc 1.93.0

### Operating system

Windows 10

### Regression?

No, first-time setup

### Pre-flight checks

### LINKED ISSUE #6906: [Feature]: improve Nix flake
### Summary

Nix flake should expose package and module

### Problem statement

Hello, I had a look at the project’s `flake.nix` but what I found isn’t what I was expecting.

The flake has the following outputs:
- `packages`: this is expected but it’s outputting a toolchain, not the `zeroclaw` package. Why?
- `devShells`: at a first glance looks reasonable.
- `nixosConfigurations`. I didn’t expect to find this. Do you use it to spin up test machines?

At the same time the flake is not exposing `nixosModules.default`. Why it is not exposed given that https://github.com/zeroclaw-labs/zeroclaw/tree/3f7c4f88cc605210c854e4c4b2cc300a1e0a9187/nix actually contains a NixOS module?

### Proposed solution

I would definitely change `packages.default` to output the compiled `zeroclaw`. I would also add `nixosModules.default`.

If `nixosConfigurations` are used I would document what they are for. If not I would remove them.

### Non-goals / out of scope

_No response_

### Alternatives considered

I see the package is already in NixOS and there is a PR for a module. You could decide that you don’t want to have them in a flake and remove them from your repo.

### Acceptance criteria

I can import the flake and use it to run a NixOS module for the latest version of `zeroclaw`.

### Architecture impact

_No response_

### Risk and rollback

_No response_

### Breaking change?

No

### Data hygiene checks

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Establishes a single canonical install spec and `cargo generate installers`, eliminating the per-surface hardcoded-feature drift. Nine surfaces now derive their feature sets and version from `Cargo.toml` (via `cargo_metadata`) plus a small `[package.metadata.zeroclaw]` registry, instead of each hand-maintaining its own copy.
  - Fixes the concrete bug that every distribution surface shipped a different arbitrary feature subset: prebuilt release `channel-matrix,channel-lark`, Dockerfile `channel-lark,whatsapp-web`, Dockerfile.debian `rag-pdf`, PKGBUILD `channel-matrix,channel-lark`, Containerfile two different hardcoded lists, scoop a stale `0.5.9`, PKGBUILD a stale `pkgver=0.6.9`. All now resolve canonically.
  - Provides a full docker image (`:all-features`) and an all-channels image (`:dist`) via a generated tag matrix and a docker-publish workflow.
  - Makes the Nix flake build the real zeroclaw and zerocode with overridable features, and exposes `nixosModules.default`.
  - Adds a CI drift gate so a hand-edit to any generated surface fails the build.
- **Scope boundary:** Does NOT rewrite `install.sh` wholesale; it stays the behavioral reference. Its source-build cargo-install step is generated from the canonical spec via a sentinel zone (byte-identical to HEAD, `--dry-run` output unchanged), and the rest of install.sh stays hand-authored.
- **Blast radius:** A new `cargo generate` xtask bin plus `[package.metadata.zeroclaw]` in the root `Cargo.toml` (ignored by cargo manifest consumers; `zeroclaw-config` tests confirm parsing is unaffected). The nine generated files carry sentinel-delimited regions; hand-written glue outside the sentinels is untouched. One new required CI job (Installer Drift) and one new publish workflow (Docker Publish) that consumes the generated tag matrix.
- **Linked issue(s):** Closes #3642 (provide a "full" docker image: the `:all-features` tag builds every feature including WhatsApp, and the docker-publish workflow pushes it to GHCR; `:dist` ships all channels minus heavyweight). Closes #6906 (Nix flake: `packages.default` now builds the real zeroclaw, `packages.zerocode` builds the TUI, both with overridable features; `nixosModules.default` exposes the `services.zeroclaw` module; the `nixosConfigurations` are documented as dev test systems). Related #7069 (Twitter/X missing from the pre-built binary despite the feature existing) and Related #3933 (Feishu/Lark missing from precompiled binaries). These are the same recurring failure: a channel exists in source but is silently absent from the distributed binary, fixed pointwise each time. This PR is the structural fix for the whole class. Because the dist feature set is now derived from `Cargo.toml` to include every channel, a feature can no longer exist in source and miss the distributed binary, so these stop recurring. Related #6895 (define a lean default channel set): the spec's `non_row_features` / `Dist` / `Full` distinction is the canonical encoding of the default-vs-full channel set that issue established. Related #6822 (zerocode in the release matrix and package managers; this PR now owns the canonical version sync for PKGBUILD and scoop, and the flake builds zerocode). Related #7553, #7554 (the `web/dist` install-surface drift that motivated this tooling).
- **Labels:** `enhancement`, `risk: medium`, `size: L`, `ci`

## Supersedes

- **#7493** (`build(container): add web-build, check, and fat-image stages to the StageX pipeline`, by @ConYel). That PR hand-edited `Containerfile` to add `config-gen`, `web-build`, and `check` stages, full `build`/`build-fat` binaries, and nonroot/HEALTHCHECK runtime hardening, but its `build`/`build-fat` feature lists were hardcoded meta-features (`default-channels` / `channels-full`), which would fail this PR's Installer Drift gate. ConYel's intent is preserved here verbatim (co-authored), with the two feature lists sourced from the generated `container-standard` (Dist) and `container-fat` (All) sentinel zones so the StageX pipeline composes with the canonical spec instead of conflicting with it. #7493 should close in favor of this.

## StageX container pipeline (integrated from #7493, made buildable)

#7493's Containerfile was never built by any CI job, so five latent defects in its pure-static musl pipeline had gone undetected. All are fixed here; every StageX stage now builds and the binaries run.

- **proc-macro / build-script static-link failure.** The musl target defaults `crt-static` on, and in StageX host == target == musl, so the static flag was applied to host build-scripts and proc-macros (which cannot link statically): `ld: attempted static link of dynamic object /lib/libc.so`. The `check` and `web-build` stages now export `RUSTFLAGS=-C target-feature=-crt-static`; `build` and `build-fat` carry the static flags via per-target `CARGO_TARGET_<triple>_RUSTFLAGS` (target-scoped, so they do not bleed into host artifacts) while host build-scripts link dynamically.
- **web-build SIGSEGV.** A `.so`-removal loop deleted shared libraries the already-built proc-macro `.so`s need at load time, crashing `rustc`. The loop is removed; the `-crt-static` opt-out makes it unnecessary.
- **check stage running tests that need files `.dockerignore` excludes.** The xtask doc-gen gates read `docs/` and `.github/` paths absent from the build context; xtask is excluded from the StageX test run (those gates run in the standard CI `Test` job against the full tree).
- **check stage running tests that need absent system tooling.** `zeroclaw-tools` content_search/git_operations shell out to GNU `rg`/`grep`/`git` (busybox grep lacks the GNU flags); the crate is excluded from the StageX test run. Benches were also pulled in by `--all-targets` and failed the musl static link (criterion's `alloca` C dep, `file format not recognized`); the test selector is now `--lib --bins --tests`, excluding both benches and host doctests.
- **all-features build failure on ALSA.** `Selection::All` pulled in `voice-wake` → `cpal` → `alsa-sys`, whose build script needs system ALSA (`pkg-config: Package alsa was not found`) absent from the minimal image. A new canonical `container_excluded_features` registry set is subtracted from `Selection::All`; `voice-wake` stays available via source `--features` or a nix override.

Also made the `zeroclaw-memory` lucid test fixtures POSIX (`#!/usr/bin/env bash` to `#!/bin/sh`) so they run in minimal environments, not just hosts with bash.

## Canonical model

- Spec resolves features and version through `cargo_metadata` (Cargo's own resolver). No awk, no hand-parsing.
- The bits that are genuinely not derivable from the feature graph live in the registry, read by the tool, never shadowed: `non_row_features` (aggregate/meta features), `heavyweight_features` (excluded from the default binary dist), and `container_excluded_features` (need system libs/tooling absent from the minimal static container image; excluded from `All`).
- Selections: `Minimal`, `Full` (Cargo default), `Dist` (all channels minus heavyweight, the default binary dist), `All` (kitchen sink), `Features` (explicit, validated).
- The install plan is a tree (diverge prebuilt|source, reconverge at a shared tail). Each step pairs a dry-run narration with its real action, so the dry-run line and the command cannot drift.

## Generated surfaces (9)

| Surface | Generated | Selection |
|---|---|---|
| install.sh | source-build cargo-install step (dry-run/execute dispatcher) | n/a |
| setup.bat | build menu, routing, presets | minimal/dist/default/all |
| Containerfile | both feature lists | standard=Dist, fat=All |
| Dockerfile / .debian | feature ARG default (overridable) | Dist |
| dist/aur/PKGBUILD | pkgver + build features | Dist |
| dist/scoop/zeroclaw.json | version | n/a |
| flake.nix | parameterized buildRustPackage (zeroclaw + zerocode, overridable) | Dist default |
| dev/ci/docker-tags.toml | Alpine-style pinned + floating tag matrix | all selections |

The docker-publish workflow consumes `docker-tags.toml` via `yq`, expands it into the Actions build matrix, and pushes each entry's pinned and floating tags to GHCR with the resolved features, so published images cannot drift from the spec.

setup.bat reaches dry-run parity with install.sh: a `--dry-run` flag narrates "[dry-run] Would ..." across the source, prebuilt, and rustup paths and skips all mutations. Its build modes are aligned to the generated menu (`--minimal`/`--dist`/`--default`/`--all`), replacing the stale `--standard`/`--full`. `bump-version.sh` runs `cargo generate installers` after the version bump (and now also bumps `apps/` path-dep pins, which it previously missed), so a release keeps every surface canonical; CI fails the PR if a surface drifts.

```bash
cargo fmt --all -- --check
cargo clippy -p xtask --all-targets -- -D warnings
cargo test -p xtask --lib generate
cargo generate installers --check
bash -n install.sh && sh -n install.sh
```

  - `cargo fmt --all -- --check`: clean, no diff.
  - `cargo clippy -p xtask --all-targets -- -D warnings`: `Finished`, 0 warnings (including the repo's disallowed-`anyhow!` policy).
  - `cargo test -p xtask --lib generate`: `test result: ok. 35 passed; 0 failed`.
  - `cargo test -p zeroclaw-config --lib`: `test result: ok. 828 passed; 0 failed` (confirms the Cargo.toml metadata addition does not disturb config parsing).
  - `cargo generate installers --check`: `ok:` for all nine surfaces, exit 0.
  - `bash -n install.sh && sh -n install.sh`: clean.
- **Beyond CI, what was manually verified:**
  - install.sh `--source --dry-run` emits `[dry-run] Would run: cargo install --path . --locked --force` from the generated zone, identical to HEAD.
  - Drift gate: an injected hand-edit to a generated zone makes `--check` exit nonzero with `DRIFT`; restored, it exits 0. Idempotent on regenerate.
  - Both new workflows were executed on a real Actions runner via `act`: the Installer Drift job ran `cargo generate installers --check` (all nine surfaces in sync, job succeeded), and the Docker Publish matrix job ran the real `yq` step and emitted a valid JSON build matrix for all four tags (minimal, default-features, dist, all-features).
  - A real `bump-version.sh 0.9.0` was run in-place to confirm the apps-pin fix: `zerocode` bumped, `cargo generate installers` succeeded, all surfaces updated; the test bump was then fully reverted (branch stays at 0.8.0).
  - dist vs all: dist ships every channel and excludes heavyweight (hardware, browser-native, rag-pdf, voice-wake); all includes the heavyweight set except the container-excluded `voice-wake` (needs system ALSA).
  - **Real StageX container build of every target via `docker build` (not just generation):** `check` (fmt + clippy `--features ci-all` + test suite, exit 0), `web-build` (`dist/index.html` produced, no SIGSEGV, exit 0), `package` (minimal runtime, 75.9MB, exit 0), `package-fat` (all-features runtime, 115MB, exit 0). Both runtime images run the static binary: `docker run zc-stagex:pkg --version` and `zc-stagex:pkgfat --version` both print `zeroclaw 0.8.0`; `zerocode --version` prints `zerocode 0.8.0`. The Dockerfile build-arg routing was verified per branch: `minimal` -> `--no-default-features`, `default-features` -> Cargo default, `dist` -> `--no-default-features --features ...`.
  - setup.bat has no broken `goto` targets; the StageX Containerfile build heredocs pass `sh -n` after the feature injection was moved to a `ZEROCLAW_FEATURES` assignment (a mid-continuation comment would have broken the `--frozen` build). The generated setup.bat `--minimal` preset resolves to `--no-default-features` (confirms no regression of the closed #6836 minimal-size bug), and the Dockerfiles' web/dist copy logic is untouched (only feature args changed), preserving the closed #5995/#5959/#6400/#6024 web-dist fixes.
- **Intentionally not run locally (no host capability; these run in CI):** the Docker Publish `publish` job (build and push to GHCR, needs registry credentials; validated via `act` dry-run), a live Windows `setup.bat`, and `nix build` / `nix flake check`. The generated outputs are correct by construction but not evaluated on their native toolchains here.

- New permissions, capabilities, or file system access scope? **No** (the xtask writes generated files within the repo; CI runs `--check` read-only; the publish workflow uses the scoped `GITHUB_TOKEN`).
- New external network calls? **No** in the spec/tooling; the publish workflow pushes to GHCR as a release action.
- Secrets / tokens / credentials handling changed? **No** beyond using the standard `GITHUB_TOKEN` for the GHCR push.

## Compatibility (required)

- Backward compatible? **Yes**. Generated surfaces produce equivalent or corrected behavior; installer flags and gateway config are unchanged except the setup.bat mode rename (`--standard`/`--full` to `--dist`/`--default`/`--all`), which aligns Windows with the canonical mode names.
- Config / env / CLI surface changed? **Yes (additive)**: a new `cargo generate` xtask alias and `[package.metadata.zeroclaw]` in Cargo.toml. No user-facing runtime CLI or config change.
- Upgrade steps: none. Maintainers regenerate surfaces with `cargo generate installers` after changing features in Cargo.toml; CI enforces it.

- **Fast rollback command/path:** revert the branch's commits; the generated files return to their prior content and the CI job is removed.
- **Observable failure symptoms:** the Installer Drift CI job failing on master (a feature changed in Cargo.toml without regenerating), or a generated surface shipping an unexpected feature set (inspect `cargo generate installers --check` output).


Audacity88 (dismissed):
Reviewed `47e13ca` against the live PR state, PR body, linked issues #3642 and #6906, current green CI, the full diff, and the public source paths for the generated Docker matrix and Dockerfile build branches. There are no prior formal reviews or inline threads to reconcile. The shared installer spec direction looks right, but I found one release-matrix bug that I would block before merge.

### 🟢 What looks good — The single-spec shape is the right direction

The overall move away from per-surface hand-maintained feature lists is strong. The generator registry, Cargo metadata-backed feature resolution, sentinel-spliced surfaces, and Installer Drift CI job are exactly the kind of structure that should keep install and release surfaces from quietly diverging again once the Docker minimal/default distinction is fixed.

### 🔴 Blocking — The generated `minimal` Docker tag still builds default features

`dev/ci/docker-tags.toml` generates the `minimal` entry with `features = ""`, and `docker-publish.yml` passes that through as `ZEROCLAW_CARGO_FEATURES=${{ matrix.features }}`. In both `Dockerfile` and `Dockerfile.debian`, an empty `ZEROCLAW_CARGO_FEATURES` takes the `else` branch and runs `cargo build --release --locked -p zeroclawlabs -p zerocode` without explicitly passing `--no-default-features`, so the root package's default feature set still applies.

That means the published `minimal` tag is not actually the `Selection::Minimal` / `--no-default-features` build the canonical spec promises. The drift gate will still pass because the generated files are internally in sync; the problem is that the generated Docker matrix cannot currently represent the “no default features” case through the Dockerfile build arg contract.

Please make the Docker publish path carry enough information to distinguish “minimal” from “default features” — for example by generating a full cargo flag string or a separate no-default/default toggle — and add a regression that would fail if the `minimal` matrix entry can again route to a default-feature Docker build.

Audacity88:
@singlerider: milestone alignment needed. This feature PR does not clearly fit the scope boundary of the currently open milestones, and it is still in the review lane because of the requested Docker minimal/default-feature fix. Please advise whether it should be placed in an open milestone or left unmilestoned until the review path is clear.

WareWolf-MoonWall (commented):
Reviewed at 47e13ca4. The direction here — a canonical install spec as the single source of truth with generated installer surfaces and a drift CI gate — is the right architectural call, and the xtask coverage is serious work. I want to register that clearly. But @Audacity88 holds an active CHANGES_REQUESTED at this same head, and I won't approve over it; the blocker is real and needs to be resolved first.

### 🟢 What looks good — SOT architecture and test coverage are genuinely strong

The spec-as-authority approach (install spec drives generated Dockerfile, package manifests, and any other installer surface) is exactly the pattern that prevents the class of drift bug this PR was filed to fix. The 35-test xtask suite and the `act` validation giving local CI parity are the right enforcement layer — they mean the drift gate isn't just a promise, it's checked on every push. That's the work doing what it says on the tin. Once the Docker publish bug is addressed this will be in good shape to land.

@singlerider — the one item to resolve before requesting re-review is @Audacity88's blocker on the `minimal` Docker tag: an empty `ZEROCLAW_CARGO_FEATURES` in the Dockerfile currently doesn't translate to `--no-default-features` being passed to cargo, so the published `minimal` tag is actually building default features. The spec says `minimal` means something specific; the published image needs to match that. Once the Docker publish path distinguishes `minimal` from default-features and the xtask tests cover that distinction, @Audacity88's CHANGES_REQUESTED is the one to resolve or dismiss before this can move forward.

Nillth (approved):
## #7558 — canonical install spec + `cargo generate installers` — reviewed, clean

Traced the spec/resolver and the emitted surfaces against the head commit rather than the summary. The central claim — *nine install surfaces derive their feature sets and version from `Cargo.toml`, so per-surface drift becomes structurally impossible* — holds. Detail below.

### Resolver / spec (the load-bearing part)
- **Single source of truth, verified.** `resolve()` reads via `cargo_metadata` (Cargo's own resolver, so the feature graph matches real builds), and the three judgment sets — `non_row_features`, `heavyweight_features`, `container_excluded_features` — all funnel through one `read_registry_list(pkg, key)` that reads `[package.metadata.zeroclaw]`. No surface and no resolver path shadows these with a literal list. The "not derivable from the graph" rationale in the Cargo.toml comments is legitimate (e.g. `embedded-web` vs `whatsapp-web` are structurally identical single cross-crate refs but semantically meta-vs-row).
- **Selection math is closed and tested.** `Dist = channels-full ∪ default − heavyweight`; `All = all − non_row − pure-alias − container_excluded`; `Full` expands `default` to leaves (drift-checkable, not implicit); `Minimal = --no-default-features`; `Features(..)` validates every name against `[features]` and errors on unknown. `all_is_superset_of_dist`, `dist_has_all_channels_minus_heavyweight`, and `explicit_features_validated` pin these.
- **Aggregate expansion is correct.** Both `expand_default` and `FeatureCtx::expand` walk aggregates, skip `dep:` / cross-crate `/` refs, and dedup — the typed twin of install.sh's `expand_default_features`.

### Splice / sentinel mechanism
- Both `splice()` impls (install_sh + container) are fail-closed: missing BEGIN **or** END sentinel is a hard error, never a silent no-op. Interior-only replacement preserves all hand-authored glue. `render_file_is_idempotent_against_real_install_sh` guards against double-application.
- **install.sh stays the behavioral reference.** The one generated zone interpolates cargo flags at runtime via `$CARGO_FLAGS` rather than baking them, so the zone is selection-stable — the "byte-identical to HEAD / `--dry-run` unchanged" claim is structurally credible.
- The container generator's `ZEROCLAW_FEATURES="..."` assignment (vs injecting `--features` mid backslash-continuation) is a genuinely thoughtful fix for the StageX `--frozen` shell-parse problem.

### CI / publish surface
- **Drift gate** is `cargo generate installers --check` — read-only (compares, writes nothing), wired into the required `gate` job's `needs`, so a hand-edit to any generated surface fails branch protection. Exactly the enforcement claimed.
- **docker-publish.yml**: least-privilege `permissions: {contents: read, packages: write}`, scoped `secrets.GITHUB_TOKEN` only (no new secrets), all actions pinned to commit SHAs, triggers limited to `v*` tags + manual dispatch, matrix sourced from the generated `docker-tags.toml`. No untrusted input reaches the build args. Clean.

### Out-of-pattern file
- `crates/zeroclaw-memory/src/lucid.rs` (16/16) is **test-fixture-only**: converts fake-lucid shell scripts from bash-isms (`[[ ]]`, `==`, `#!/usr/bin/env bash`) to POSIX `sh`. Consistent with the pure-`sh` musl/StageX environment this PR introduces; no production path touched.

### CI
17/17 green, including the new **Installer Drift** and **Nix Module Eval** gates.

### Non-blocking notes
1. `non_row_features(meta, pkg)` takes `meta` but only uses `pkg` (delegates straight to `read_registry_list(pkg, ..)`). Dead param — harmless, drop on a future pass.
2. The `:all-features`/StageX paths (Docker Publish, `nix build`, live `setup.bat`) are correct-by-construction but run only in CI, not on the author's host — flagging that the native-toolchain evaluation of those three is deferred to CI, as the PR states.

**Verdict: approve.** Genuine structural fix for a recurring drift class (#7069/#3933/#3642/#6906 et al.), fail-closed tooling, real enforcement, strong test coverage.

WareWolf-MoonWall (approved):
Re-review at a7946861 — converting my prior `--comment` to an approval. @Audacity88's CHANGES_REQUESTED has been dismissed (they addressed the blocker themselves), and @Nillth has posted a thorough APPROVED review at this head confirming the structural correctness of the spec/resolver.

### ✅ Resolved — Docker publish matrix now properly distinguishes minimal from default features

The prior block was that `ZEROCLAW_CARGO_FLAGS` was empty for the `minimal` tag, meaning the published image built with default features despite the name claiming otherwise. The new `docker-publish.yml` reads `flags` from the generated `docker-tags.toml` matrix and passes them as `ZEROCLAW_CARGO_FLAGS` build-arg — so `minimal` now provably passes `--no-default-features` at build time, sourced from the same spec that drives all other surfaces. The published image matches what the spec says.

### 🟢 What looks good — installer-drift gate is wired into branch protection

The `installer-drift` job (`cargo generate installers --check`) is now in the `needs` array of the required `gate` job. This means a hand-edit to any generated installer surface — Dockerfile, install.sh zone, docker-tags.toml — fails branch protection automatically. That's the enforcement claim made real, not just a promise.

### 🟢 What looks good — CI security posture on docker-publish.yml

Least-privilege `permissions: {contents: read, packages: write}`, all actions pinned to commit SHAs, no new secrets beyond `GITHUB_TOKEN`, trigger limited to `v*` tags and `workflow_dispatch`. The matrix expansion passes through `yq` from the generated TOML rather than reconstructing it in the workflow — consistent SSOT from spec to published tag.

---

The non-blocking notes from @Nillth's review stand (dead `meta` param in `non_row_features`, native-toolchain paths evaluated only in CI). Neither is a blocker. @singlerider: milestone alignment comment was left in a prior round — please advise if this should land in v0.8.1 or the next open milestone once review is clear.

Approving. The spec-as-authority architecture is correct, the enforcement is real, and the Docker publish blocker is resolved.

Audacity88 (approved):
Reviewed current head a7946861 against the live PR state, prior reviews, linked issues #3642 and #6906, the dismissed Audacity88 blocker from 47e13ca, the Docker publish path, generated tag matrix, generated Dockerfile args, installer-drift CI wiring, and the current green check set.

### ✅ Resolved — Minimal Docker tag now carries the no-default-features build intent

The prior blocker is resolved. The generated `dev/ci/docker-tags.toml` now carries an authoritative `flags` field, with `minimal` set to `--no-default-features` and `default-features` left empty for Cargo defaults. `docker-publish.yml` passes that value through as `ZEROCLAW_CARGO_FLAGS`, and both Dockerfiles consume that full flag string instead of trying to infer behavior from an empty feature list. That closes the gap where the published `minimal` tag could still build with default features while the spec said it was `Selection::Minimal`.

### 🟢 What looks good — The enforcement path is now real

The installer-drift job is wired into the required gate, and the generator has a focused regression for the exact distinction that broke before: `minimal` must emit `--no-default-features`, `default-features` must not, and `dist` must carry explicit `--no-default-features --features ...` flags. That makes the generated surfaces, the CI drift check, and the publish workflow agree on the same source of truth instead of relying on a reader to remember how empty strings should behave.

### 🟡 Warning — Fix PR-body issue-closing references

This is not a code-review blocker, but the public record should be cleaned up before merge. Two prose references to already-closed issues are being interpreted by GitHub as closing references: #5995 and #6836. Rewording those as “previously fixed” or “prior” references would keep the issue-closing metadata clean without changing the implementation.

Approving current head. This clears my prior Docker minimal/default-feature blocker; the remaining PR-body cleanup is maintainer-fixable metadata, not something I would hold the code approval on.

Audacity88 (approved):
Reviewed current head a7946861 against the live PR state, prior reviews, linked issues #3642 and #6906, the dismissed Audacity88 blocker from 47e13ca, the Docker publish path, generated tag matrix, generated Dockerfile args, installer-drift CI wiring, and the current green check set.

### ✅ Resolved — Minimal Docker tag now carries the no-default-features build intent

The prior blocker is resolved. The generated `dev/ci/docker-tags.toml` now carries an authoritative `flags` field, with `minimal` set to `--no-default-features` and `default-features` left empty for Cargo defaults. `docker-publish.yml` passes that value through as `ZEROCLAW_CARGO_FLAGS`, and both Dockerfiles consume that full flag string instead of trying to infer behavior from an empty feature list. That closes the gap where the published `minimal` tag could still build with default features while the spec said it was `Selection::Minimal`.

### 🟢 What looks good — The enforcement path is now real

The installer-drift job is wired into the required gate, and the generator has a focused regression for the exact distinction that broke before: `minimal` must emit `--no-default-features`, `default-features` must not, and `dist` must carry explicit `--no-default-features --features ...` flags. That makes the generated surfaces, the CI drift check, and the publish workflow agree on the same source of truth instead of relying on a reader to remember how empty strings should behave.

### 🟡 Warning — Fix PR-body issue-closing references

This is not a code-review blocker, but the public record should be cleaned up before merge. Two prose references to already-closed issues are being interpreted by GitHub as closing references: #5995 and #6836. Rewording those as “previously fixed” or “prior” references would keep the issue-closing metadata clean without changing the implementation.

Approving current head. This clears my prior Docker minimal/default-feature blocker; the remaining PR-body cleanup is maintainer-fixable metadata, not something I would hold the code approval on.

PR #7568, opened 2026-06-13
fix(docs): add .nojekyll so Pages serves _shared/ chrome

singlerider:
## Summary

- **What changed and why:** The published docs site lost all chrome (sidebar, theme palette, layout, custom overrides, favicons) because every page references absolute-root `/_shared/` assets that 404 on the live site. The `_shared/` tree is correctly present on `gh-pages` — the break is GitHub Pages itself: Pages runs Jekyll by default, and Jekyll excludes every top-level underscore-prefixed directory from the served output, so `_shared/` is silently dropped while non-underscore paths (`master/en/...`) serve fine. This adds `touch .nojekyll` on every `gh-pages` push, alongside the existing `CNAME` re-assert, on both the initial sync and the push-retry path. A tracked `.nojekyll` at the root disables Jekyll and serves the tree verbatim; re-asserting each run means a tree rewrite or orphan rebuild can never silently re-enable Jekyll and regress the chrome again.
- **Scope boundary:** Does not touch the chrome-extraction logic, `_shared` reference-depth math, theme generation, or any xtask code. Workflow-only, one file.
- **Blast radius:** `docs-deploy.yml` only. No runtime, channel, provider, library, or build-tool code changed.
- **Linked issue(s):** none
- **Labels:** `docs`, `type: docs`, `risk: low`, `size: S`

Workflow-only change. Validated the live fix end-to-end and the YAML.

```bash
# applied .nojekyll to live gh-pages, then verified every _shared asset:
for f in css/chrome.css theme/pc-themes.css theme/custom.css \
         theme/version-selector.js theme/pc-enhance.js theme/lang-switcher.js \
         favicon.svg favicon.png; do
  curl -s -o /dev/null -w "$f -> %{http_code}\n" "https://docs.zeroclawlabs.ai/_shared/$f"
done
python3 -c "import yaml; yaml.safe_load(open('.github/workflows/docs-deploy.yml'))"
```

  - Before fix: `_shared/css/chrome.css -> 404` (all eight `_shared/*` assets 404; `master/en/css/general-*.css -> 200`)
  - After pushing `.nojekyll` to `gh-pages`: all eight `_shared/*` assets `-> 200`; `master/en/sop/example.html -> 200`
  - `python3 ... yaml.safe_load(...)` → `YAML OK`
- **Beyond CI — what did you manually verify?** Cloned `gh-pages` and confirmed `_shared/` was fully populated on the branch (chrome.css, pc-themes.css, custom.css, all theme JS, favicons) while every `_shared/` URL 404d on the live site and non-underscore paths served 200 — isolating the cause to Jekyll dropping the underscore dir, not a missing/empty tree. Applied the one-line `.nojekyll` fix to live `gh-pages` and re-verified all assets serve 200. Did NOT run a full multi-locale `mdbook build` (requires the mdbook toolchain in CI); this change is workflow shell only and does not alter the build.
- **If any command was intentionally skipped, why:** The Rust battery (`fmt`/`clippy`/`test`) is N/A — no Rust changed. The diff is a 2-line shell addition to one workflow file.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: none.

Low-risk: `git revert <sha>`. Reverting only re-enables Jekyll on the next deploy; the live `gh-pages` already carries a tracked `.nojekyll` independent of this workflow change.

## Incident attribution — what broke the live docs

Two distinct regressions surfaced together, both originating from the versioned-docs deploy system introduced in **#7023** (`feat(docs): implement versioned documentation deployment and version selector`, commit `40be7738f3`, merged 2026-06-03).

1. **Chrome 404 (this PR fixes it).** Every page references absolute-root `/_shared/` assets. The `_shared/` tree is published to `gh-pages` correctly, but GitHub Pages runs Jekyll by default and Jekyll excludes top-level underscore-prefixed directories from the served site, so `_shared/` is silently dropped. #7023 introduced the `_shared/` layout without a `.nojekyll`. Fixed here by re-asserting `.nojekyll` on every deploy.

2. **Pinned version dirs wiped (separate latent bug, not fixed by this PR).** docs-deploy run **27461126807** (push of `fix(install)...` to master, 2026-06-13 08:14:07 UTC) took the orphan-branch path in `docs-deploy.yml` (lines 162-174: `git switch --orphan gh-pages` + `git rm -rf .`) even though `gh-pages` already existed. The run committed only the current `master/` build from an empty tree (`8758 files changed, 1264001 insertions(+)`) and force-replaced the branch as unrelated history (`* [new branch] gh-pages -> gh-pages`). This destroyed `stable/` and every `v0.x.x/` version dir. The current `gh-pages` root commit `115f181e3b` is parentless, confirming the wipe. The 12:20 UTC run found master unchanged and made no further change (`nothing to commit`), so the branch has carried only `master/` since 08:14 UTC.

The orphan-path trigger in #7023's deploy script is tracked separately and is **out of scope for this PR** — this PR ships only the `.nojekyll` chrome fix. The wiped version dirs need recovery from the orphaned `gh-pages` history or a rebuild from release tags.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -206,6 +206,13 @@ jobs:
           # or GitHub Pages will remove it (re-introduced fix from #6142).
           echo "docs.zeroclawlabs.ai" > CNAME
 
+          # Disable Jekyll. Pages runs Jekyll by default, which drops every
+          # top-level underscore-prefixed dir (`_shared/`) from the served
+          # site — so the shared chrome (CSS/JS/favicons) 404s on every page
+          # while non-underscore paths serve fine. Re-assert each deploy, same
+          # as CNAME, so a tree rewrite can never silently re-enable Jekyll.
+          touch .nojekyll
+
           SCRIPTS="${GITHUB_WORKSPACE}/.workflow-scripts/.github/scripts"
 
           # Shared chrome (_shared/) is owned by master only. Every deployed
@@ -290,6 +297,7 @@ jobs:
             fi
 
             echo "docs.zeroclawlabs.ai" > CNAME
+            touch .nojekyll
             cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- gen-versions > versions.json
             cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- prune-root
             cargo run --manifest-path "${GITHUB_WORKSPACE}/.workflow-scripts/Cargo.toml" -p xtask --bin mdbook -- retrofit-selector


Nillth (approved):
I read the full diff, the PR body's validation evidence, and cross-checked the change against the live `.github/workflows/docs-deploy.yml` and the `prune-root` / `retrofit-selector` / `gen-versions` xtask code on master. This is the first review on this PR (head 850071fd). No other reviewer holds an active block.

**My understanding of the change.** GitHub Pages serves the `gh-pages` branch through Jekyll by default (the workflow targets the classic "deploy from a branch" source — see the file header). Jekyll strips every top-level underscore-prefixed directory from the published output, so `_shared/` — the chrome (CSS/JS/favicons) that every page references at `/_shared/` — 404s site-wide, while the non-underscore version dirs (`master/en/...`) serve fine. The `_shared/` tree was present and fully populated on the branch the whole time; the break was Jekyll silently dropping it. A tracked root `.nojekyll` disables Jekyll and serves the tree verbatim. The fix `touch`es it on both the initial-sync path and the push-retry path, re-asserting it every deploy exactly like the existing `CNAME` guard, so an orphan rebuild or a by-tree rewrite can never silently re-enable Jekyll and regress the chrome again.

### 🟢 What looks good — both deploy paths covered, modeled on the CNAME guard

The re-assert lands in both the initial sync (right after `CNAME`) and inside the push-retry loop (after `git reset --hard FETCH_HEAD` + `git clean -fd`). That second placement is the one that's easy to miss and you got it right: on a rejected first push, the reset+clean wipes the not-yet-on-remote `.nojekyll`, and the retry `touch` re-creates it before `git add --all` stages it. Reusing the exact shape of the `CNAME` re-assert — same two sites, same "persist on every push or Pages drops it" reasoning — keeps the two invariants symmetric and easy to reason about together.

### 🟢 What looks good — composes cleanly with prune-root's root-file guarantee

`touch .nojekyll` runs *before* `prune-root` in both paths, so I checked whether the finalizer could delete it. It can't: `prune_root()` (`xtask/src/cmd/mdbook/versions.rs:84`) only removes *directories* — it `continue`s on any non-dir entry — and its doc comment states root files are deliberately never touched, precisely so a future deploy can add root files like `404.html` / `robots.txt` / `sitemap.xml`. `.nojekyll` lands squarely inside that contract. `retrofit-selector` and `gen-versions` likewise only walk version dirs, so the root file survives the full finalizer chain unscathed. It's a nice property that this fix slots into an invariant the code already documents rather than working around it.

### 🟢 What looks good — the validation isolated the real cause

The before/after `curl` matrix (all eight `_shared/*` assets 404 → 200, non-underscore paths already 200) is exactly the evidence that distinguishes "Jekyll dropped the underscore dir" from "tree missing or empty," and confirming `_shared/` was fully populated on `gh-pages` nailed it. Applying `.nojekyll` to live `gh-pages` already un-broke the published site, so this PR's job is to make that durable across redeploys — and it does. One note for posterity: the manual live push didn't exercise `prune-root`, but as above I confirmed prune-root leaves root files alone, so the workflow path is covered too.

### 🔵 Suggestion — minor label / body drift (non-blocking)

The PR body lists labels `docs, type: docs, risk: low, size: S`, but the applied set is `size: S, type: docs, ci, risk: low` — the path-labeler added `ci` (correct: this touches `.github/workflows/`) and there's no plain `docs` label. The current labels are accurate as-is, so there's nothing to change; just flagging the body↔actual drift in case you'd like the body to match.

---

This is a correct, minimal, well-reasoned fix with strong end-to-end validation, and it composes with the existing pruning invariants rather than fighting them. Approving. The only thing I'd want green before merge is the in-flight CI (Build / Test / Lint / Check were still running at 850071fd) — no Rust changed so I expect it to pass, but the approval is on the understanding that the required checks finish clean.

Issue #7470, opened 2026-06-11
[Bug]: delegate agentic mode rejects empty risk_profile.allowed_tools and same-profile gating blocks stricter delegated targets


vrurg:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

Two coupled delegate-path behaviors block practical multi-agent reviewer/research setups:

1. Delegating from one agent to an **agentic** target fails when the target's `risk_profile.allowed_tools` is empty, with an error like:

```text
Agent 'reviewer' is agentic but risk_profile 'yolo' has no allowed_tools
```

2. The delegate target-policy gate rejects targets whose `risk_profile` name differs from the caller, even though nearby comments describe subset / no-escalation validation.

Together, those behaviors make it impossible to cleanly model a common multi-agent topology:

- a broader operator agent
- stricter read-only `reviewer` / `deep-research` agents
- delegation from the broader caller to the stricter targets

### Expected behavior

- Agentic delegated targets should either:
  - honor the documented empty-`allowed_tools` semantics (`inherit / no authorization constraint`), or
  - the schema/docs/comments should be updated to consistently state that agentic delegate targets require an explicit allowlist.
- Delegate target authorization should accept a target whose effective policy is a non-escalating subset / equivalent policy, rather than hard-requiring exact `risk_profile` name equality.
- Another acceptable design would be to make delegation possible when the target agent, or its policy/profile, is explicitly allowed by the caller's security profile.
- In practice, a broader caller should be able to delegate to stricter read-only reviewer/research agents without weakening their policy just to satisfy routing.

### Steps to reproduce

Minimal config shape:

```toml
[runtime_profiles.unbounded]
agentic = true

[risk_profiles.yolo]
allowed_commands = ["*"]
forbidden_paths = []
block_high_risk_commands = false
level = "full"
require_approval_for_medium_risk = false
sandbox_enabled = false
workspace_only = false

[risk_profiles.yolo.delegation_policy]
mode = "allow"

[agents.main]
runtime_profile = "unbounded"
risk_profile = "yolo"

[agents.reviewer]
runtime_profile = "unbounded"
risk_profile = "yolo"

[agents.deep-research]
runtime_profile = "unbounded"
risk_profile = "yolo"
```

Observed runtime repro:

```bash
# Start the current build/service on latest master.
# From an agent session running as `main`, invoke the delegate tool with:
[...]

PR #7590, opened 2026-06-14
feat(agents): per-agent delegate roster with cross-profile reach

### LINKED ISSUE #7514: [Feature]: change `delegate` tool semantics to allow subagents with different risk profiles
### Summary

Delegation to subagents with different risk profiles must be possible for more effective separation of concerns support. 

### Problem statement

This is a follow-up to #7470. 

The current ZeroClaw design doesn't permit an agent to delegate to another agent if their risk profiles differ. I would like to propose a change that would keep the current behavior as default but given explicit allow lists would permit delegating to select risk profiles or specific agents.

# The reasoning and benefits.

Specializing an agent often requires limiting its abilities even in seemingly harmless areas. Here are some real-life examples of the designs I'm using.

## Task Orchestrator

This class of agents is narrowly specialized in preventing context drift by being focused on meticulous task planning and tracking. These agents are deprived of even being capable of web search and fetch. Any inputs they get are coming from user, research, and review agents. The only operations they're permitted are modifying the tracking media (files, specialized tools, whatever) and delegating to specialists.

The reason behind this is that given more capabilities than necessary, any LLM tends to bend the rules and do everything by itself.

The problem of the current delegation model is apparent: there is a risk that a TO agent won't delegate at all.

## Research/review agents

This is somewhat the opposite side of the problem. These agents need wider permissions than TO but still remain non-mutating ones.

A typical scenario of their use can be demoed with, say, a sysadmin agent which is a pure YOLO. Safety measures require it to do careful initial research of a problem first and reviewing the planned actions before actually acting. Yet, this is not possible because it can't delegate even to restricted agents. 

### Proposed solution

Introduce `allowed_risk_profiles` and `allowed_agents` keys on `delegation_policy`. If present, `delegate` will check if the target is an allowed agent or has an allowed risk profile:

```toml
[risk_profiles.foo.delegation_policy]
mode = "allow"
allowed_risk_profiles = ["bar", "baz"]
allowed_agents=["reviewer", "deep-research"]
```

The `allowed_` keys would map into `Option<Vec<String>>` fields on `DelegationPolicy` struct.

Given green light, I can try to implement this myself.

### Non-goals / out of scope

_No response_

### Alternatives considered

_No response_

### Acceptance criteria

_No response_

### Architecture impact

_No response_

### Risk and rollback

_No response_

### Breaking change?

No

### Data hygiene checks

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** Delegation reach was hardwired to "target must share the caller's risk profile." That forced every agent needing a distinct tool surface onto its own risk profile, which then made those agents mutually unreachable: you could not have a front agent delegate to both an MCP-only tool agent and a separate RAG agent. This replaces the single same-profile gate with a resolved, per-agent reachable set so an agent keeps a shared authorization tier while naming the specific peers it may hand off to.
- **What changed and why:** Two new per-agent fields on `AliasedAgentConfig`: `delegate_same_profile` (default `true`, auto-allow same-profile peers; set `false` to opt out) and `delegates` (explicit roster, possibly empty, may name cross-profile agents).
- **What changed and why:** `Config::reachable_delegate_targets` is the single source of truth for reach: `(same-profile peers when delegate_same_profile) ∪ delegates`, minus self, minus disabled agents, deduped and sorted. Both the delegate tool's advertised roster and its per-target authorization gate resolve through it, so the two cannot drift.
- **What changed and why:** `policy_for_target` refuses out-of-roster targets, keeps same-profile session-cwd inheritance + shared tracker, and routes explicit cross-profile delegates through `ensure_no_escalation_beyond` so a listed delegate can never widen privilege beyond the caller. `model_routing_config.upsert_agent` reads/writes/surfaces both fields. `validate()` fails loud on a dangling or self-referential `delegates` entry.
- **Scope boundary:** Does NOT change the `delegation_policy.mode` gate (stays on the risk profile), the depth mechanism (`runtime_profile.max_delegation_depth`, set to `1` for a single hop), the `spawn_subagent` path, or any migration code. No per-authenticated-user delegate restriction (that is a future v0.9.0 concern; the resolver is shaped as the single chokepoint to absorb it later without a rewrite).
- **Blast radius:** `delegate` tool reach/authorization; config schema (additive, `#[serde(default)]`); the `model_routing_config` tool surface. No channel, gateway, or provider behavior changes.
- **Linked issue(s):** Closes #7514 (cross-profile delegation via explicit allow-list). Related #7470 (this PR fixes the same-profile gate half so a broader caller can delegate to stricter targets). This PR also changes the empty-`allowed_tools` agentic case from reject to inherit: an agentic target with `allowed_tools = []` now inherits the caller's already-filtered tool registry instead of being rejected. The upsert-agent guard that bailed on `agentic=true` + empty `allowed_tools` is removed and the runtime path uses `inherit_all = allowed_tools.is_empty()`. This is safe because the caller's filtered registry is the ceiling: `sub_tools` is built from the caller's `parent_tools`, so an inheriting target can never receive a tool the caller lacks. `delegation.md` documents the new semantics.
- **Placement note vs #7514:** the issue proposed `allowed_risk_profiles` / `allowed_agents` on `delegation_policy` (risk-profile-scoped). This PR instead puts the roster per-agent (`[agents.<alias>].delegates` + `delegate_same_profile`), because reach is intrinsically per-agent while the profile stays the shared gate. Same capability, different placement; open to moving it if the maintainer prefers the issue's exact shape.

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-config -p zeroclaw-tools -p zeroclaw-runtime
```

`cargo fmt --all -- --check`: clean, exit 0.

`cargo clippy --all-targets -- -D warnings` (full workspace):
```
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 38s
```
Zero warnings, zero errors across the whole workspace.

`cargo test` (every crate the diff touches):
```
test result: ok. 838 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 90 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 2116 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out
test result: ok. 1334 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

- **Beyond CI, what did you manually verify?** New tests cover: resolver auto-allow / opt-out / explicit-union / self-exclusion / unknown-caller / disabled-peer / disabled-explicit-delegate; `validate()` dangling + self-ref rejection; serde-default round-trip for legacy configs; delegate tool roster advertising (same-profile peer, explicit cross-profile, opt-out); `policy_for_target` rejection of out-of-roster and escalating cross-profile targets, and acceptance of a narrowed cross-profile delegate; `model_routing_config` upsert/readback of both fields. Confirmed the caller's already-policy-filtered tool registry remains the ceiling for an agentic cross-profile target (so `ensure_no_escalation_beyond`, which does not diff tool allowlists, is sufficient).
- **What I did NOT verify:** A full-workspace `cargo test` to completion: the box was under heavy concurrent build contention. Full-workspace clippy passing clean already proves every downstream crate compiles; the touched crates' tests all pass.
- **If any command was intentionally skipped, why:** None skipped; the workspace-wide test run was substituted with full-workspace clippy + targeted tests on all touched crates due to machine contention.

- New permissions, capabilities, or file system access scope? **No.** Cross-profile delegation is gated by `ensure_no_escalation_beyond` and the caller's tool registry ceiling; a delegate can never exceed the caller.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No.** Test agents use neutral aliases (`aaa`, `aaatools`, `aaalore`, `caller`, `target`).

## Compatibility (required)

- Backward compatible? **Yes.** Both new fields are `#[serde(default)]`; existing configs parse unchanged with `delegate_same_profile = true` (historical reach preserved) and empty `delegates`.
- Config / env / CLI surface changed? **Yes** (additive). New optional per-agent keys `delegate_same_profile` and `delegates` under `[agents.<alias>]`, and two new optional args on the `model_routing_config` `upsert_agent` action. No upgrade steps required; omitting them reproduces prior behavior.

- **Fast rollback command/path:** `git revert <merge sha>`. The change is self-contained (one config schema addition + the delegate tool reach logic); reverting restores the same-profile-only gate.
- **Feature flags or config toggles:** None. Behavior is inert for existing configs (defaults preserve prior reach).
- **Observable failure symptoms:** Delegations refused unexpectedly would log `delegate refused: target not in caller's reachable set` or `cross-profile target would escalate beyond caller` (grep `delegate refused` in the runtime trace). A regression in reach resolution would surface as the `agent` enum in the delegate tool schema listing the wrong targets.


singlerider:
Reviewers, please gate on correct log span attribution.

Nillth (approved):
## Review: approve

Strong PR. The "reach is per-agent, the gate stays on the risk profile" framing is the right decomposition, and — more importantly — the privilege-boundary claims in the summary actually hold when traced through the code, not just asserted. I read the head state of the three security-critical pieces end to end:

### Reachable-set resolver (`Config::reachable_delegate_targets`) ✅
- Unknown caller → empty (fail-closed).
- Same-profile branch guards on a **non-empty** caller profile, so two agents with blank `risk_profile` don't accidentally become mutually reachable.
- Both branches require `agent.enabled`; explicit `delegates` additionally skip empty/self entries.
- `BTreeSet` gives dedupe + stable sort, so the tool's advertised `agent` enum and the per-target gate resolve through the identical set — they cannot drift, as claimed.

### Authorization gate (`policy_for_target`) — ordering is correct ✅
1. `delegation_policy.permits()` (caller may delegate at all)
2. roster membership (`reachable_delegate_targets`)
3. same-profile → inherit caller `workspace_dir` (historical behavior preserved)
4. cross-profile → `ensure_no_escalation_beyond(&self.security)`, refusing any widening of autonomy/roots/budgets
5. `target_policy.tracker = self.security.tracker.clone()` **before** the branch, so a cross-profile hop shares the caller's action/cost budget and can't escape the rate limit.

### The tool-ceiling invariant — verified, this is the load-bearing one
The summary leans on "`ensure_no_escalation_beyond` doesn't diff tool allowlists, but that's fine because the caller's registry is the ceiling." That's true in the code: `sub_tools` is constructed from `self.parent_tools` (the caller's **already-policy-filtered** registry), filtered by `tool_admitted` and with `delegate` stripped out. Consequences:
- empty target `allowed_tools` (`inherit_all`) → target gets a **subset** of the caller's tools, never a superset;
- non-empty → caller-registry ∩ target-allowlist;
- a cross-profile target whose risk profile names a tool the caller never had → that tool isn't in `parent_tools`, so it can't materialize;
- `delegate` is stripped → no privilege laundering by chaining a second hop.

So the escalation guard not inspecting tool allowlists is *sound*, because registry construction enforces the tool ceiling independently. Good call documenting this explicitly in `delegation.md` — it's exactly the invariant a future reader needs.

### Other
- Backward compat is real: both fields `#[serde(default)]`, `delegate_same_risk_profile` defaults `true`, `delegates` empty → existing configs reproduce prior reach.
- `validate()` failing loud on dangling/self-referential `delegates` is the right place to catch operator typos.
- Test coverage hits the cases that matter: auto-allow / opt-out / explicit-union / self-exclusion / disabled-peer / disabled-explicit-delegate / out-of-roster rejection / escalating-cross-profile rejection / narrowing-cross-profile acceptance / serde round-trip / `model_routing_config` readback.
- CI green across all 16 checks.

### Non-blocking notes
1. **Placement question is the maintainer's call, not mine.** You flagged that #7514 proposed `allowed_risk_profiles`/`allowed_agents` on `delegation_policy`, and you put the roster per-agent instead. I think per-agent is the better model (reach *is* intrinsically per-agent), but that's a design-owner decision — calling it out so it's not silently diverging from the issue.
2. **Field rename mid-PR** (`delegate_same_profile` → `delegate_same_risk_profile`) is fine and clearer, just make sure no docs/example config still reference the old name. The book pages in the diff look updated; worth a final `grep -r delegate_same_profile` to be sure nothing stale lingers.

Nothing here blocks. This is the kind of change where the security story is the whole point, and the security story checks out. LGTM.

WareWolf-MoonWall (dismissed):
This PR replaces the hardwired "same-profile-only" delegation gate with a per-agent resolved reachable set, letting an agent name specific cross-profile peers while keeping the risk-profile as the shared authorization tier. The core mechanic is `Config::reachable_delegate_targets` — a single resolver that feeds both the `delegate` tool's advertised schema enum and `policy_for_target`'s roster gate so the two cannot drift.

**Scope comprehension.** `AliasedAgentConfig` gains two new `#[serde(default)]` fields: `delegate_same_risk_profile` (default `true`, auto-includes same-profile peers) and `delegates` (explicit cross-profile roster). `Config::reachable_delegate_targets` is the canonical resolver: (same-profile peers when flag is set) ∪ explicit roster, minus self, minus disabled agents, deduped via `BTreeSet`. `policy_for_target` is reordered: delegation-policy gate → roster membership → same-profile (inherit workspace boundary, historical behavior) / cross-profile (`ensure_no_escalation_beyond` + shared tracker). `model_routing_config.upsert_agent` surfaces both new fields. `Config::validate()` fails loud on dangling or self-referential `delegates` entries.

---

### 🔴 Blocking — PR summary claims `allowed_tools = []` agentic check is "out of scope here and left unchanged" — the diff proves otherwise

The Summary section reads: _"the separate 'agentic target with empty allowed_tools' rejection in #7470 is out of scope here and left unchanged."_ The diff contradicts this on two counts:

1. The upsert-agent guard in `model_routing_config.rs` that bailed with `"Agent '{name}' has agentic=true but allowed_tools is empty"` is removed.
2. The runtime path in `delegate.rs` that previously rejected on an empty `allowed_tools` set is replaced with `let inherit_all = allowed_tools.is_empty()` — an explicit semantic change from "reject" to "inherit the caller's tool registry."

A comment in the diff at the revised site references `#7470` directly and calls out the new "inherit instead of reject" behavior. `delegation.md` also documents the new semantics. The change appears correct and safety-preserving (the caller's filtered registry is still the ceiling), but it is a _real behavioral change_ that the PR template explicitly disclaimed.

This matters beyond cosmetics: the Summary is what future readers, bisect-runners, and revert-preparers rely on to understand blast radius. If the template says "unchanged" and the diff says "changed," the audit trail is broken. Please either confirm the removal is intentional, update the Summary to document it, and explain why the `inherit_all` semantics are safe given the existing `allowed_tools = []` check in #7470 — or revert the upsert-agent guard if the removal is accidental.

### 🔴 Blocking — `caller_alias` is missing from three of four delegate refusal log events (author's explicit gate request)

@singlerider asked reviewers to "gate on correct log span attribution." After reading the four new log events in `policy_for_target`, I'm blocking on this.

`caller_alias` is present exactly once — in the "target not in caller's reachable set" event. The other three are missing it:

| Log message | Has `caller_alias`? |
|---|---|
| `delegate refused: caller delegation_policy forbids delegation` | ❌ — only `caller_risk_profile` |
| `delegate refused: target not in caller's reachable set` | ✅ |
| `delegate: could not resolve target's security policy` | ❌ — only `target_agent`, `error` |
| `delegate refused: cross-profile target would escalate beyond caller` | ❌ — only `caller_risk_profile`, `target_risk_profile`, `violation` |

In a multi-agent runtime the `caller_alias` is the primary correlation field: without it, an operator seeing a delegation refusal in the trace cannot identify which agent triggered it without reconstructing context from surrounding spans. Every delegate refusal event should carry `caller_alias` as a structured attribute alongside whatever other context it already has. The second event shows the right pattern; the other three need the same treatment.

### 🟡 Warning — Full-workspace `cargo test` was not run to completion for a `risk: high` PR

The PR body acknowledges the workspace test was substituted with full-workspace clippy + targeted per-crate tests due to "heavy concurrent build contention." Full-workspace clippy passing is strong evidence of compilation correctness, and the four per-crate runs all pass. But for a security-boundary change on a `risk: high` PR I'd like to see `cargo test --workspace` pass before merge. CI is green across all 16 checks, which helps, but please confirm a full workspace test run has cleared since the submission — or explain why the CI run is a sufficient substitute.

### 🔵 Informational — `delegates` field uses `Vec<String>` rather than `Vec<AgentAlias>`

`AgentAlias` is the typed newtype for agent aliases. Once #7594 lands it will carry `PropKind::AliasRef` + `AliasSource::Agents`, making any field of that type automatically picker-eligible in zerocode and the web config surface. The `delegates` field using `Vec<String>` means it won't inherit that treatment without a follow-up change. This is not a correctness issue and I don't think it should block this PR, but it's worth tracking as a follow-up so the delegates roster gets a picker rather than a freeform text entry.

---

### ✅ SSOT is respected — `Config::reachable_delegate_targets` is the single resolver

Both the tool's `parameters_schema()` (advertised `agent` enum) and `policy_for_target` (per-target authorization gate) resolve through `Config::reachable_delegate_targets`. The `BTreeSet` dedup guarantees the two surfaces can never diverge. Neither field on `AliasedAgentConfig` caches a pre-filtered list; both new fields are raw config state that the resolver reads on demand. This is exactly the right pattern.

### ✅ Authorization chain ordering is correct

As Nillth verified: `delegation_policy.permits()` gates first, then roster membership, then same-profile / cross-profile branching. The tracker is assigned _before_ the branch so both paths share the caller's action/cost budget. The tool-ceiling invariant holds: `sub_tools` is constructed from the caller's already-filtered `parent_tools`, so a cross-profile target with `inherit_all` or a non-empty allowlist can never receive a tool the caller didn't already have.

### 🟢 Rejection-path test coverage is thorough

Out-of-roster rejection, escalating-cross-profile rejection, dangling `delegates` entry, self-referential entry, disabled peer, disabled explicit delegate — all have explicit tests. The acceptance case (narrowed cross-profile delegate) is also covered. The `serde` round-trip test confirms that legacy configs without the new fields parse correctly with defaults.

Audacity88 (commented):
Reviewed current head 8f94b08ab80653d21506b0965a7978cec6ab7a00 against the live PR body, linked issues #7514 and #7470, @Nillth's approval, @WareWolf-MoonWall's active CHANGES_REQUESTED review, @singlerider's request to gate on log span attribution, the empty inline review thread list, current green GitHub checks, and the public diff/file patches for `schema.rs`, `delegate.rs`, `model_routing_config.rs`, and the delegation docs.

I'm leaving this as a comment review rather than an approval because @WareWolf-MoonWall's current-head CHANGES_REQUESTED is still active and its two required changes are supported by the diff.

### 🟡 Warning — Active requested changes are still supported

I am not opening a separate CHANGES_REQUESTED review because @WareWolf-MoonWall already owns the current block and I did not find a distinct issue beyond that review. I do agree those active requested changes remain valid.

**PR-body accuracy:** the PR body still says `delegate_same_profile`, while the code and docs now expose `delegate_same_risk_profile`, so that field name needs a body refresh. More importantly, the body says the #7470 "agentic target with empty `allowed_tools`" rejection is out of scope and unchanged, but the runtime path now treats empty `allowed_tools` as "inherit" with `inherit_all = allowed_tools.is_empty()`. The `model_routing_config` guard that rejected agentic agents with empty `allowed_tools` is also removed, which follows from that runtime semantic change.

I agree that the new behavior looks intentional and security-bounded: the target still receives only tools from the caller's filtered registry, with `delegate` stripped, and a non-empty target allowlist only narrows that set. The public record still needs to name that behavior change directly and use the actual config field name so future readers do not believe #7470 was left untouched.

**Log attribution:** in `policy_for_target`, the "target not in caller's reachable set" event includes `caller_alias`, but the delegation-policy refusal, target-policy resolution failure, and cross-profile escalation refusal do not. These are authorization-path failures in a multi-agent runtime, so risk profile alone is not enough to diagnose which configured agent triggered the refusal. The caller alias should be present on each delegate refusal/error event.

I also see @WareWolf-MoonWall's validation warning. The current GitHub Test check is green, so I am not adding a separate validation ask here, but the PR body still says full-workspace `cargo test` did not complete locally. Before merge, the author/maintainer path should either record a completed full workspace test or explain why the green CI Test check is the accepted substitute for this high-risk delegation/security change.

### 🟢 What looks good — The reach resolver and tool ceiling are the right shape

The core structure is solid. `Config::reachable_delegate_targets` is the single resolver for the advertised delegate roster and the authorization gate, so those surfaces do not drift. Disabled agents and self-targets are excluded, explicit delegates are validated, and the cross-profile path still runs through `ensure_no_escalation_beyond` while sharing the caller's tracker.

I also agree with @Nillth that the tool-ceiling invariant is the load-bearing safety story: agentic delegation builds sub-tools from the caller's already-policy-filtered parent registry, then intersects with the target allowlist when present. Once the PR body and the missing `caller_alias` log fields are fixed, I expect this to be close to approval if the current green checks hold.

singlerider:
Re the risk:high full-workspace test ask: branch is now merged with master (4a155e2) and CI ran the full suite on the merged tree. Test job green (6m49s), Lint green (7m0s), 0 failures across all checks. caller_alias is now on all four refusal events, and the Summary documents the empty-allowed_tools reject->inherit change. Ready for re-review.

Issue #6689, opened 2026-05-15
Production SOP audit is silently no-op: documented sop_run_*/sop_step_* Memory keys are never written


JordanTheJet:
## Summary

[`docs/book/src/sop/observability.md`](https://github.com/zeroclaw-labs/zeroclaw/blob/master/docs/book/src/sop/observability.md#L7-L14) promises that SOP runs persist audit entries to the configured Memory backend under category `sop`:

> SOP audit entries are persisted via `SopAuditLogger` into the configured Memory backend, category `sop`.
>
> Common key patterns:
> - `sop_run_{run_id}`: run snapshot (start + completion updates)
> - `sop_step_{run_id}_{step_number}`: per-step result
> - `sop_approval_{run_id}_{step_number}`: operator approval record
> - `sop_timeout_approve_{run_id}_{step_number}`: timeout auto-approval record

In production, **none of these keys are ever written.** Both production SOP execution paths bypass audit:

### Path 1 — MQTT listener uses `NoneMemory`

[src/main.rs:1820](https://github.com/zeroclaw-labs/zeroclaw/blob/master/src/main.rs#L1820):

```rust
let audit = Arc::new(SopAuditLogger::new(Arc::new(NoneMemory)));
```

`NoneMemory` is a no-op Memory implementation. Every `audit.log_run_start(...)`, `log_step_result(...)`, `log_approval(...)` call inside `dispatch_sop_event` and the engine writes to `/dev/null`.

### Path 2 — Agent tools never wire an audit logger at all

The SOP tool structs accept an optional audit logger via `with_audit(...)`:

```
$ grep -n "with_audit\|self.audit" crates/zeroclaw-runtime/src/tools/sop_execute.rs crates/zeroclaw-runtime/src/tools/sop_approve.rs crates/zeroclaw-runtime/src/tools/sop_advance.rs
sop_execute.rs:25:    pub fn with_audit(mut self, audit: Arc<SopAuditLogger>) -> Self { ... }
sop_execute.rs:94:        if let Some(ref audit) = self.audit
sop_approve.rs:27:    pub fn with_audit(mut self, audit: Arc<SopAuditLogger>) -> Self { ... }
sop_approve.rs:84:        if let Some(ref audit) = self.audit
sop_advance.rs:27:    pub fn with_audit(mut self, audit: Arc<SopAuditLogger>) -> Self { ... }
sop_advance.rs:139:        if let Some(ref audit) = self.audit {
```

[...]

PR #7593, opened 2026-06-14
fix(sop): wire audit logger into registered SOP tools

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `SopExecuteTool` / `SopAdvanceTool` / `SopApproveTool` accept an audit logger via `with_audit(...)`, but `all_tools_with_runtime` constructed them with bare `::new(...)`, leaving `self.audit` permanently `None`. Every audit branch (`log_run_start`, `log_step_result`, `log_run_complete`, `log_approval`) was dead code on the agent path, so the `sop_run_*` / `sop_step_*` / `sop_approval_*` Memory keys documented under category `sop` were never written.
  - This is the agent-driven path real deployments use: an inbound event (e.g. AMQP) enters the agent loop and fires the SOP via the `sop_execute` tool, not via the MQTT listener. Without the audit trail the run is not reconstructable and `SopMetricsCollector::rebuild_from_memory` returns empty after a restart.
  - Construct a single `SopAuditLogger` over the agent's real Memory backend and pass it to the execute/advance/approve tools at registration.
  - Add a registration-level test that fires `sop_execute` through `all_tools` and asserts a `sop_run_*` entry is persisted. The per-tool unit tests already cover audit behavior; none proved the wiring.
- **Scope boundary:** Only the agent tool-registration path. Does NOT change the MQTT listener's `NoneMemory` use (`src/main.rs`), which is the other half of #6689 and is entangled with engine unification (#6687). Does NOT change the audit logger, key format, or memory backend.
- **Blast radius:** SOP tool registration only. When `[sop].sops_dir` is set, the execute/advance/approve tools now write audit entries to the agent's existing Memory backend. No new tools, no schema change. Agents with no SOPs configured are unaffected (the block is gated on `sops_dir.is_some()`).
- **Linked issue(s):** Related #6689 (resolves the agent-path half; the MQTT/NoneMemory half remains, see #6687)
- **Labels:** `type: bug`, `risk: low`, `size: S`

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-runtime --lib sop
```

```
$ cargo fmt --all -- --check
FMT_EXIT=0   (no output, clean)

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.00s

$ cargo test -p zeroclaw-runtime --lib sop
test tools::sop_advance::tests::advance_success_writes_step_audit ... ok
test tools::sop_approve::tests::approve_writes_audit ... ok
test tools::tests::registered_sop_tools_persist_audit_trail ... ok
test result: ok. 183 passed; 0 failed; 0 ignored; 0 measured; 1930 filtered out; finished in 0.39s
```

- **Beyond CI — what I manually verified:** New test `registered_sop_tools_persist_audit_trail` fails on the pre-fix tree (tools registered with `::new`, no audit) and passes after wiring. Confirmed the channel orchestrator and agent-loop call sites pass the real per-agent Memory backend (`Arc::clone(&mem)`), so the persisted entries land in production memory, not a no-op. Verified the audit key (`sop_run_<id>`, category `Custom("sop")`) matches `docs/book/src/sop/observability.md`.
- **Beyond CI — what I did NOT verify:** Full workspace `cargo test` was not run; scoped to the touched crate. The MQTT listener path is out of scope and unverified here.
- **If any command was intentionally skipped, why:** Ran crate-scoped clippy/test rather than the full workspace battery because the change is a one-file, single-crate wiring fix with no cross-crate surface.

## Compatibility (required)

- Backward compatible? `Yes`
- Existing agents with `[sop].sops_dir` set begin persisting the documented audit entries automatically; no action required.

### DIFF
--- a/crates/zeroclaw-runtime/src/tools/mod.rs
+++ b/crates/zeroclaw-runtime/src/tools/mod.rs
@@ -1124,10 +1124,17 @@ pub fn all_tools_with_runtime(
         let mut engine = crate::sop::SopEngine::new(root_config.sop.clone());
         engine.reload(workspace_dir);
         let sop_engine = Arc::new(std::sync::Mutex::new(engine));
+        let sop_audit = Arc::new(crate::sop::SopAuditLogger::new(memory.clone()));
         tool_arcs.push(Arc::new(SopListTool::new(Arc::clone(&sop_engine))));
-        tool_arcs.push(Arc::new(SopExecuteTool::new(Arc::clone(&sop_engine))));
-        tool_arcs.push(Arc::new(SopAdvanceTool::new(Arc::clone(&sop_engine))));
-        tool_arcs.push(Arc::new(SopApproveTool::new(Arc::clone(&sop_engine))));
+        tool_arcs.push(Arc::new(
+            SopExecuteTool::new(Arc::clone(&sop_engine)).with_audit(Arc::clone(&sop_audit)),
+        ));
+        tool_arcs.push(Arc::new(
+            SopAdvanceTool::new(Arc::clone(&sop_engine)).with_audit(Arc::clone(&sop_audit)),
+        ));
+        tool_arcs.push(Arc::new(
+            SopApproveTool::new(Arc::clone(&sop_engine)).with_audit(Arc::clone(&sop_audit)),
+        ));
         tool_arcs.push(Arc::new(SopStatusTool::new(Arc::clone(&sop_engine))));
     }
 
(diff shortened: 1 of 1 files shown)


singlerider:
## End-to-end dry-run evidence (before/after a real binary)

Beyond the unit test, I drove the full agent path through a real deployment (the StageX auto-update bot) and confirmed the before/after directly. Same SOP, same synthetic trigger, two binaries:

**Release binary (0.8.0, pre-fix):** the SOP run completed, but the documented audit keys were never written.

```
$ sqlite3 -readonly <install>/data/memory/brain.db \
    "SELECT key FROM memories WHERE key LIKE 'sop_%';"
(no rows)
```

**Fixed binary (this branch, built with channel-amqp,channel-matrix):** the same run persists the full trail.

```
$ sqlite3 -readonly <install>/data/memory/brain.db \
    "SELECT key, category FROM memories WHERE key LIKE 'sop_%' ORDER BY key;"
sop_run_det-1781411965802-0001|sop
sop_step_det-1781411965802-0001_1|sop
sop_step_det-1781411965802-0001_2|sop
sop_step_det-1781411965802-0001_3|sop
sop_step_det-1781411965802-0001_4|sop
sop_step_det-1781411965802-0001_5|sop
sop_step_det-1781411965802-0001_6|sop
sop_step_det-1781411965802-0001_7|sop
sop_step_det-1781411965802-0001_8|sop

$ sqlite3 -readonly <install>/data/memory/brain.db \
    "SELECT substr(content,1,200) FROM memories WHERE key LIKE 'sop_run_%';"
{
  "run_id": "det-1781411965802-0001",
  "sop_name": "stagex-update",
  "trigger_event": { "source": "manual", ... },
  "status": "completed",
  "current_step": 8,
  "total_steps": 8,
  ...
```

The run snapshot (`sop_run_<id>`) and one `sop_step_<id>_N` per advanced step land under category `sop`, exactly as `docs/book/src/sop/observability.md` documents. On the release binary the identical run persists nothing.

Reproduced a second time from a clean from-scratch install (fresh clone, empty config dir, daemon CWD at the install root): SOP fired via `sop_execute`, `sop_run_*` row present. The fix is what makes `SopMetricsCollector::rebuild_from_memory` non-empty after a restart on the agent path.

Nillth (approved):
I read the full diff, the SOP audit/tool sources, the documented audit-key contract, and both tool-registration call sites, and cross-checked the change against the code where it lands. This is a tight, correct wiring fix: `all_tools_with_runtime` constructed `SopExecuteTool` / `SopAdvanceTool` / `SopApproveTool` with bare `::new(...)`, so `audit` stayed `None` and every `if let Some(ref audit) = self.audit` branch was dead on the agent-driven path — the path real deployments use when an inbound event fires `sop_execute`. Constructing one `SopAuditLogger` over the agent's Memory backend and threading it into the three audit-capable tools at registration is exactly the right shape. Verdict: approve.

### 🟢 What looks good — root cause is named precisely, not just patched

The diagnosis is correct down to the mechanism: `SopExecuteTool::new` sets `audit: None`, `with_audit` is the only thing that flips it, and `execute()` gates every persistence call on `self.audit` being `Some`. Bare registration therefore made `log_run_start` / `log_step_result` / `log_run_complete` / `log_approval` silent no-ops on the agent path, so the documented `sop_run_*` / `sop_step_*` / `sop_approval_*` keys were never written and `SopMetricsCollector::rebuild_from_memory` came back empty after a restart. The PR body explains the *why* (run not reconstructable), not just the *what*.

### 🟢 What looks good — the new test guards the wiring, not the unit

This is the gap that mattered. The per-tool tests already proved `with_audit` persists; none proved registration actually calls it, which is precisely the seam that broke. `registered_sop_tools_persist_audit_trail` routes through the real `all_tools` → `all_tools_with_runtime` path, fires `sop_execute`, and asserts a `sop_run_*` entry lands in the backing Memory — so it fails on the pre-fix tree and passes after. A test that fails before and passes after is the right bar for a wiring fix.

### 🟢 What looks good — resolves through the canonical backend, and the scope boundary is honest

`SopAuditLogger::new(memory.clone())` shares the same `Arc<dyn Memory>` already used to build `MemoryStoreTool`/`MemoryRecallTool`, so the audit entries land in the agent's real per-agent memory — an Arc handle to canonical state, not a cached duplicate, consistent with the single-source-of-truth rule in AGENTS.md. The scope note is also accurate: the MQTT-listener half wires its logger over `NoneMemory` (src/main.rs), which is a different root cause (no-op backend, not missing wiring) and is correctly left to #6687 / #6689 rather than smuggled in here. One concern per PR, declared boundary, confirmed in the source.

### 🟢 What looks good — restores the documented SOP audit contract on its primary path

This isn't only a bug fix, it's a design-conformance fix. The SOP docs make auditability a first-class property: the index calls SOPs "auditable run state," and `how-it-works.md` states unconditionally that "SOP runs are started by MQTT event fan-in or by the in-agent tool `sop_execute`" and that "SOP audit records are persisted in the configured Memory backend under category `sop`." The event-flow diagram dispatches the `sop_execute` tool into the same Engine→Run machinery as MQTT — the tool path is a co-equal designed entry point, not a corner case. It's also the path the reference deployment uses: the StageX worked example in `example.md` is driven over AMQP (a transport, not a native trigger), where "the agent fires the SOP with the `sop_execute` tool," and that page promises every audit key "maps to a concrete definition in the codebase." Pre-fix, those keys were never written on this path, so the flagship example described an audit trail that didn't exist at runtime. This change makes the implementation honor the documented contract, and the keys/category in `audit.rs` match `observability.md` one-for-one — no schema drift. The remaining gap (the MQTT listener still logging over `NoneMemory`) contradicts the same contract but is a different root cause and is correctly deferred to #6687 / #6689 rather than folded in here.

### 🔵 Suggestion — apply the labels the body already declares

The PR body lists `type: bug`, `risk: low`, `size: S`, but only the auto-scope `runtime` label is currently applied. Worth landing those so the metadata matches the description. On the risk label: AGENTS.md flags `crates/zeroclaw-runtime/src/**` as high-risk territory, though it qualifies that with "especially `src/security/`"; since this change is additive observability with no control-flow, boundary, or security impact and is gated behind `[sop].sops_dir`, `risk: low` is defensible — `risk: medium` would be the more conservative call. Not a blocker either way; flagging so the label is a deliberate choice.

Nice fix — small surface, real bug, and the test pins the exact regression so this can't silently revert.

PR #7594, opened 2026-06-14
feat(config): type-driven alias-ref pickers and self-declaring config enums

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** Config field UX behavior now follows from the field's Rust type, eliminating hardcoded per-path special-casing and central registration lists. This is an internal representation change only; on-disk TOML and runtime behavior are unchanged.
  - Added `PropKind::AliasRef` + `AliasSource` with a single `section_path()` / `is_two_tier()` mapping. Ref newtypes (`ModelProviderRef`, `TtsProviderRef`, `TranscriptionProviderRef`, `ChannelRef`, plus new `RiskProfileRef` / `RuntimeProfileRef`) carry their `ALIAS_SOURCE`. `agents.<alias>.{risk_profile,runtime_profile}` and `peer_groups.<name>.channel` become alias-selectable purely via their types.
  - Added `Config::resolve_alias_source` as the single resolver every surface uses; collapsed the gateway `build_agent_options` onto it.
  - Replaced the `impl_enum_prop_kind!` central list with a `ConfigEnum` derive so each enum self-declares `HasPropKind` at its definition site.
  - Plumbed `alias_source` through `PropFieldInfo`/`ConfigFieldEntry`, the macro, the RPC (`config/resolve-alias-source`), and a REST route. zerocode and the web FieldForm render alias-ref fields with a generic resolver-driven picker; deleted the hardcoded path branches.
  - Represented the observability/runtime value sets as typed `ConfigEnum`s (`ObservabilityBackend`, `LogPersistence`, `LogToolIo`, `RuntimeKind`) so the factories match the enum directly. Deserialization stays lenient: legacy aliases (`noop`, `opentelemetry`, `otlp`) map to their canonical variant, and any unrecognized value falls back to the field default exactly as the old string fields did.
- **Scope boundary:** No behavioral change to delegation, providers, channel logic, or config-loading tolerance. No new config fields; on-disk TOML shapes are byte-identical (newtypes are `#[serde(transparent)]`).
- **Blast radius:** Config schema, the macro crate, gateway config RPC/REST, zerocode + web config surfaces, and call sites constructing these config structs (mostly tests).
- **Linked issue(s):** None.
- **Labels:** `type: feature`, `risk: medium`

```bash
cargo fmt --all -- --check
cargo test --workspace
```

  - `cargo fmt --all -- --check` -> exit 0, no diff.
  - `cargo clippy --all-targets -- -D warnings` -> exit 0, no warnings.
  - `cargo test --workspace` -> exit 0: 8524 passed, 0 failed.
- **Beyond CI — what did you manually verify?** Confirmed configs load unchanged: `backend = "noop"` -> `None`, `backend = "otlp"`/`"opentelemetry"` -> `Otel`, unknown `backend`/`runtime.kind` -> field default (load never fails), and the legacy `runtime_trace_mode` alias key still maps to `log_persistence`. Did NOT manually exercise the zerocode/web picker UI in a live daemon.

## Compatibility (required)

- Backward compatible? Yes
- If `No` or `Yes` to either: No upgrade steps. On-disk config is unchanged. The observability/runtime fields accept exactly the same strings they did before (including the legacy `noop`/`opentelemetry`/`otlp` aliases), and unrecognized values still fall back to the field default at load rather than failing.

Low-risk: `git revert <squash-sha>` is the plan.


WareWolf-MoonWall (dismissed):
This PR makes config field UX behavior fall out of the field's Rust type rather than from hardcoded per-path branch logic and a central registration list. On-disk TOML and runtime behavior are unchanged; this is a representation refactor with one extension: new typed ref newtypes (`RiskProfileRef`, `RuntimeProfileRef`) and new `ConfigEnum` typed values (`ObservabilityBackend`, `LogPersistence`, `LogToolIo`, `RuntimeKind`) replace the corresponding raw strings.

**Scope comprehension.** `PropKind::AliasRef` + `AliasSource` are added to `zeroclaw-config/src/traits.rs`. Ref newtypes (`ModelProviderRef`, `TtsProviderRef`, `TranscriptionProviderRef`, `ChannelRef`, new `RiskProfileRef`, `RuntimeProfileRef`) declare their source via `const ALIAS_SOURCE: Option<AliasSource>`. `Config::resolve_alias_source` becomes the single resolver all surfaces use. The gateway's `build_agent_options` is collapsed onto it, deleting the old `dotted_aliases` helper. `impl_enum_prop_kind!` central registration list is replaced by a `#[derive(zeroclaw_macros::ConfigEnum)]` that lets each enum self-declare `HasPropKind` at its definition site. `PropFieldInfo` / `ConfigFieldEntry` gain `alias_source: Option<AliasSource>`. The `Configurable` derive macro threads `alias_source` into field emission. zerocode and the web `FieldForm` use a generic resolver-driven picker for all `AliasRef` fields; the hardcoded `agents.<alias>.risk_profile` / `runtime_profile` branches are deleted.

---

### 🔴 Blocking — `AliasSource::all()` is a manually-maintained parallel variant list (AGENTS.md SSOT rule)

`crates/zeroclaw-config/src/traits.rs` adds:

```rust
pub const fn all() -> &'static [AliasSource] {
    &[
        Self::ModelProviders,
        Self::TtsProviders,
        ...
        Self::McpBundles,
    ]
}
```

This is exactly the pattern AGENTS.md explicitly prohibits: _"A schema enum variant list duplicated across an enum and a `const &[Variant]` table that aren't generated from the same macro."_ Adding a new `AliasSource` variant now requires updating both the enum definition _and_ the `all()` body — the classic drift source AGENTS.md is designed to prevent.

Compounding this: a search of the diff shows `AliasSource::all()` is never called in the new code. It is dead code. AGENTS.md is equally direct about that: _"Do not suppress unused production code... delete it, wire it into behavior, or track a follow-up issue."_

Either remove `all()` (if it has no caller), or — if it is genuinely needed — drive both the enum and the list from a single macro invocation so they cannot diverge. The `ConfigEnum` derive you're already adding would be the natural place to emit the list if it's needed.

### 🟡 Warning — `zerocode/src/wire.rs::AliasSource` manually mirrors `zeroclaw_config::traits::AliasSource`

The comment says explicitly: _"Mirrors `zeroclaw_config::traits::AliasSource`."_ This follows the existing convention in `wire.rs` where `ConfigTab` and `PropKind` also have mirrored copies, so it's not a new pattern introduced by this PR. But I want to flag it anyway because `AliasSource` is now a load-bearing type: if someone adds a new variant to the canonical enum in `traits.rs` and omits it from `wire.rs`, zerocode silently fails to deserialize `alias_source` for fields of that new type. The mirrored types in this file drift silently — there's no compile-time enforcement that the two stay in sync.

If `apps/zerocode` can declare a `[dev-dependencies]` or `[dependencies]` path on `zeroclaw-config` (or if one already exists), replacing the local `AliasSource` with a re-export from the config crate would eliminate the risk. If the dependency boundary genuinely forbids this, it's worth a comment explaining why the mirror is necessary rather than just that it mirrors.

### 🟡 Warning — PR body's self-declared labels contradict actual labels

The Summary section says _"Labels: type: feature, risk: low."_ The actual labels on this PR include `risk: medium`. Minor template hygiene issue — please update or remove the label line in the Summary so future readers aren't confused.

---

### ✅ `ConfigEnum` derive correctly eliminates the central `impl_enum_prop_kind!` list

Each enum now self-declares `HasPropKind` at its definition site via `#[derive(zeroclaw_macros::ConfigEnum)]`. Adding a new config enum no longer requires remembering to register it in a separate list — the derive is the single registration point, and forgetting it is a compile error (not a silent miss). The macro generates a minimal `impl HasPropKind` with `PROP_KIND = PropKind::Enum`; the `ALIAS_SOURCE` default of `None` on the trait means enum types continue to work correctly through the existing path.

### ✅ `Config::resolve_alias_source` is the single resolver

The gateway's `build_agent_options` previously implemented its own `dotted_aliases` helper that walked `prop_fields()` and did its own prefix-stripping; that is now deleted and replaced with a direct call to `Config::resolve_alias_source`. zerocode's hardcoded `agents.<alias>.risk_profile` / `runtime_profile` branch is deleted and replaced with the generic `alias_source`-driven path. Both surfaces now resolve through the same code path, which cannot drift.

### ✅ Backward compatibility holds

Newtypes are `#[serde(transparent)]`; on-disk TOML is byte-identical. The observability/runtime string fields accept the same values as before, including legacy aliases (`noop`, `opentelemetry`, `otlp`). Unknown values fall back to the field default. The `runtime.kind` typed match is exhaustive over the `RuntimeKind` enum — any future variant addition is a compile error rather than a silent fallthrough.

### 🟢 Full workspace test run is the gold standard

`cargo test --workspace` at 8524 passed / 0 failed is exactly what I want to see on a config-schema PR of this blast radius. Thank you for running it.

WareWolf-MoonWall (approved):
Re-review at ba32d0d6. My Round 8 CHANGES_REQUESTED was dismissed; reviewing fresh at the current head.

### ✅ Resolved — `AliasSource::all()` SSOT violation is gone

My prior blocker was that `AliasSource::all()` was a manually-maintained parallel variant list that duplicated the enum definition and was never called. At the current head, `AliasSource` derives `zeroclaw_macros::ConfigEnum` — as do every other config enum in this diff (`ObservabilityBackend`, `LogPersistence`, `LogToolIo`, `RuntimeKind`, and the full set of profile/bundle enums). The derive generates the picker-eligible surface from the single enum definition, which is the correct SSOT pattern. I confirmed `all()` is not present in the diff.

### 🟢 What looks good — the per-path special-casing in `config_manager.rs` is gone

The `if field_path.starts_with("agents.")` block that duplicated path-based picker logic is replaced by `if let Some(source) = self.fields[idx].alias_source`. Any field type that carries `PropKind::AliasRef` gets picker treatment automatically, without a new code path per field. This is the structural fix the PR promised and the right way to close that class of drift.

### 🟢 What looks good — full workspace validation

`cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test --workspace` (8524 passed, 0 failed) — all clean. No required local step skipped. The full workspace test run is particularly important for a change that touches the macro crate, config schema, gateway RPC, and zerocode client surface.

### 🟡 Warning — rollback SHA is still a placeholder for a `risk: high` PR

The Rollback section reads: "Low-risk: `git revert <squash-sha>` is the plan." The GitHub labels include `risk: high`, for which the template requires a fast rollback command with an actual SHA. The PR correctly characterizes the change as additive-only with no behavior change (on-disk TOML unchanged, newtypes are `#[serde(transparent)]`), so the practical rollback risk is low — but the form should match the label. Please replace `<squash-sha>` with the actual merge commit SHA once this lands, or file a follow-up note.

Approving. The SSOT violation is resolved, the per-path special-casing is gone, and the full test suite confirms the change is behaviorally transparent.

PR #7596, opened 2026-06-14
fix(channels/amqp): make health_check probe the broker

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `channel doctor` builds each channel and calls `health_check()` cold, without ever starting a listener. The AMQP `health_check` returned `self.consuming`, an `AtomicBool` set true only inside the running consume loop, so under `channel doctor` it was structurally always `false`. AMQP reported `❌ unhealthy (auth/config/network)` on the one surface meant to diagnose connectivity, even when the broker connects fine.
  - Replace the flag read with a real broker probe (mirrors the IRC channel): attempt the actual connection (amqps + the configured client-cert auth) and report whether the handshake succeeds, then drop the short-lived connection. This validates auth/TLS/network, which is exactly what the doctor claims to check.
  - The `consuming` flag is now unused (runtime liveness is signalled separately via `mark_component_ok("amqp")`), so it and its stores are removed. The stale test `health_check_false_before_consuming` is renamed to `health_check_false_when_broker_unreachable` to describe the actual assertion.
- **Scope boundary:** Only `crates/zeroclaw-channels/src/amqp.rs`. Does NOT change the doctor harness, its per-channel 10s timeout, the runtime liveness marker, or any other channel's `health_check`.
- **Blast radius:** AMQP `channel doctor` result only. The live `channel start` consume path is unchanged (it never read `consuming` either; only wrote it). No config or schema change.
- **Linked issue(s):** none filed; found while validating the AMQP channel against Fedora's broker.
- **Labels:** `type: bug`, `risk: low`, `size: XS`

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-channels --features channel-amqp --all-targets -- -D warnings
cargo test -p zeroclaw-channels --features channel-amqp --lib
```

```
$ cargo fmt --all -- --check
(clean, exit 0)

$ cargo clippy -p zeroclaw-channels --features channel-amqp --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 25s

$ cargo test -p zeroclaw-channels --features channel-amqp --lib
test amqp::tests::health_check_false_when_broker_unreachable ... ok
test amqp::tests::name_is_amqp ... ok
test amqp::tests::self_handle_returns_sender_label ... ok
test result: ok. 755 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 5.63s
```

- **Beyond CI — live before/after against the real Fedora Messaging broker** (`rabbitmq.fedoraproject.org:5671`, public `public_pubsub` mutual-TLS certs):

```
# BEFORE (consuming-flag health_check): AMQP could never be healthy under doctor
  ❌ AMQP      unhealthy (auth/config/network)

# AFTER (connect-probe health_check), 3 consecutive runs:
  ✅ AMQP      healthy
Summary: 1 healthy, 0 unhealthy, 0 timed out
```

  The probe performs the same amqps + SASL EXTERNAL handshake the live channel uses; the broker accepts the client cert and the doctor reports healthy.

- **Edge observed:** Fedora's broker TLS handshake is slow (`openssl s_client` measured ~11s cold). The doctor's fixed 10s per-channel timeout can therefore classify a COLD first probe as `⏱️ timed out (>10s)`; warm probes complete well under 10s and report healthy (shown above). Timeout is an honest result ("could not confirm in 10s") and is still a strict improvement over the previous guaranteed false `unhealthy`. Bumping the doctor's timeout is out of scope for this fix; noted for follow-up if slow brokers prove common.

- New external network calls? `Yes` — `channel doctor` now opens a short-lived broker connection (amqps + SASL EXTERNAL) where the old flag-read opened none. It is the same connect the live channel makes; closed immediately with reply code 200.

## Compatibility (required)

- Backward compatible? `Yes`
- `channel doctor` now reports AMQP accurately; no operator action required.

Low-risk: `git revert 411f0b9676`.


Nillth (commented):
# Review draft — PR #7596 (singlerider)

I read this against PR head `e88aaf0`, cross-checked against the IRC channel on current `master`. CI is fully green — Format, Lint, Test, all build/check matrices, Required Gate all SUCCESS. The diagnosis is correct and the fix is the right idea. One design question worth a reply before approve, plus two small notes.

## Verdict: comment / approve-with-nit (not blocking)

The core change is sound: `health_check` reading `self.consuming` (an `AtomicBool` only ever set inside the running consume loop) was structurally always `false` under `channel doctor`, which never starts a listener. Replacing it with a real broker probe is the correct fix, and removing the now-dead flag + its stores is clean. Backward compatible, no config/schema surface, blast radius is doctor-only.

## 🟡 Design question — the probe does much more than the IRC one it cites

The body says the probe "mirrors the IRC channel." It doesn't, quite. IRC's `health_check` is deliberately minimal — `connect()` (TLS handshake) + `QUIT`:

```rust
match self.connect().await {
    Ok(tls) => { /* split, send QUIT */ true }
    Err(_) => false,
}
```

AMQP's new probe reuses the full `establish_consumer()`: `connect` → `create_channel` → `queue_declare` → `queue_bind` (per routing key) → `basic_consume`. That's the entire listener setup, not just a connectivity handshake. Two consequences:

1. **It's heavier and slower than it needs to be.** The doctor only wants to know "can I reach + auth the broker." The handshake (`connect`) already answers auth/TLS/network — which is exactly what the body claims the probe validates. Declaring/binding/consuming a queue is extra round-trips that push the cold probe closer to (and, per your own edge note, over) the doctor's 10s timeout. A `connect()` + `close()` probe would be the true IRC mirror and materially faster.

2. **It has real broker-side side effects.** Under the auto-queue path (`self.queue.is_none()`), each health check declares an `exclusive`, `auto_delete` queue and binds it — server-side state created and torn down on every `channel doctor` run. With a configured durable queue it's less ephemeral but still issues a redundant declare/bind/consume against the operator's real queue on every probe. `connect()`-only avoids all of that.

Not a correctness bug — it works — but "validate auth/TLS/network" is fully served by `connect()`, and that version is faster, side-effect-free, and actually mirrors IRC. Is there a reason the probe needs the full consumer (e.g. you wanted to confirm queue perms specifically)? If so, say so in the body; if not, I'd trim it to `connect` + `close`.

## 🟢 Notes (non-blocking)

- The cold-probe-vs-10s-timeout edge is honestly documented in the body, and "timed out" is a strict improvement over the old guaranteed-false. Trimming the probe to `connect`-only (above) would also shrink the cold-probe window, partially addressing the edge without touching the timeout — nice two-birds outcome.
- Test rename `health_check_false_before_consuming` → `health_check_false_when_broker_unreachable` matches the new behavior. Good. It only exercises the `Err` path (unreachable broker → false); the `Ok` path is necessarily live-broker-only, which the body covers with the manual Fedora run. Fine for this surface — no ask.

Happy to approve as-is if the full-consumer probe is intentional; otherwise the `connect`-only trim is a small, clearly-better change.

WareWolf-MoonWall (commented):
I read this against PR head `e88aaf0`, diff in `crates/zeroclaw-channels/src/amqp.rs` only.
Nillth has already left a COMMENTED review (not CHANGES_REQUESTED) raising a design question; I'm adding my independent read before either of us approves.

---

### ✅ The diagnosis is correct and the dead-state removal is clean

`self.consuming` was an `Arc<AtomicBool>` that was only ever written *inside* the running consume loop — which `channel doctor` never enters, because it calls `health_check()` cold without starting a listener. Under `channel doctor` the flag was structurally always `false`, making AMQP permanently report `❌ unhealthy` on the one surface designed to diagnose broker connectivity. Removing the flag and its two stores is a SSOT improvement: the runtime liveness signal now lives in one place (`mark_component_ok("amqp")`), which is exactly where it belonged.

### ✅ No `unwrap()` / `expect()` in the health-check path

`establish_consumer` propagates every error with `?`. `health_check` matches `Ok`/`Err` explicitly and returns `false` on failure — no panic path. Clean.

### ✅ The probe actually tests broker connectivity

The new `health_check` calls `self.connect()` (amqps + SASL EXTERNAL handshake), which means auth, TLS, and network reachability are all exercised. A failure in any of those layers returns `Err` → `false`. The probe is not just checking whether a local struct field is set.

### ✅ No shared-state mutation risk on the health-check path

`establish_consumer` calls `self.connect()` to open a *new* connection — it does not touch or share the connection used by a concurrently running `listen` loop. The probe connection is created, used, and closed (or dropped) entirely within the `health_check` call. No race with concurrent delivery.

### ✅ Failure is surfaced to the caller, not swallowed silently

`health_check` returns `bool` to the doctor harness, which acts on `false` by marking the channel unhealthy. The error is not panicked or ignored at the boundary the orchestrator/watchdog observes.

---

### 🟡 The probe runs the full consumer setup — I share Nillth's ask for a reply before approving

Nillth's design question is the right one and I want to add the angle that concerns me most operationally. `establish_consumer` runs: `connect` → `create_channel` → `queue_declare` → `queue_bind` (per routing key) → `basic_consume`. Under the auto-queue path (`self.queue.is_none()`), each `channel doctor` run declares an `exclusive` + `auto_delete` queue on the broker and binds it to the exchange before tearing it down. That is real, observable broker-side state created and destroyed on every doctor invocation. With a configured durable queue it's less ephemeral but still issues a redundant declare/bind/consume cycle against the operator's live queue on every probe — not a no-op from the broker's perspective.

The PR body says the probe "validates auth/TLS/network." `connect()` alone answers all three of those questions — the handshake performs the SASL EXTERNAL exchange, which is the auth step. Everything after `connect()` is beyond what the stated goal requires and adds latency and broker-side effects.

A `connect()` + `close()` probe (the true IRC mirror) would:
1. Validate auth/TLS/network — same as the current probe for the stated goal
2. Produce no broker-side state (no queues declared, no bindings, no consumers)
3. Complete materially faster, shrinking the cold-probe window that you noted can push past the doctor's 10-second timeout

If the intent was specifically to validate queue permissions (declare + bind rights), that's a legitimate reason to go deeper — but it should be documented in the PR body, since the current body only claims to check auth/TLS/network. Could you confirm which scenario applies? Happy to approve once I understand whether the full consumer path is intentional or incidental.

### 🔵 The `Err(_)` in `health_check` swallows diagnostic information

```rust
Err(_) => false,
```

The doctor harness knows the channel is unhealthy but has no idea *why* — network timeout, TLS handshake failure, auth rejection, and queue permission denial all look the same to the operator. A `tracing::warn!` (or `tracing::debug!`) on the error before returning `false` would be strictly better for incident diagnosis:

```rust
Err(e) => {
    tracing::debug!(error = %e, channel = "amqp", "health_check probe failed");
    false
}
```

This stays in English (tracing messages are not Fluent-translated per project policy), adds zero overhead on the happy path, and gives operators running `RUST_LOG=debug` something actionable when `channel doctor` reports unhealthy. Not blocking — the current behaviour is a strict improvement over the old guaranteed-false — but this would be a low-effort, high-value follow-up.

---

**Summary:** The fix is correct — the old flag-read was broken by design and removing it is the right call. My one open question before approving is whether the full `establish_consumer()` path in `health_check` is intentional (queue-permission validation) or incidental (a simpler `connect` + `close` probe was the goal). Once Nillth's design question has a reply from @singlerider I'll be ready to approve or request the trim.

Audacity88 (dismissed):
I reviewed PR #7596 at head e88aaf0. I checked the one-file diff in `crates/zeroclaw-channels/src/amqp.rs`, the AMQP config semantics around `durable_ack`, the `channel doctor` health-check path, the PR body and validation, and the two existing comment reviews from @Nillth and @WareWolf-MoonWall.

I agree with the diagnosis that the old flag-based health check was structurally broken, but the current implementation introduces one merge-blocking edge in the diagnostic path when `durable_ack = false`.

### 🟢 What looks good — the old health signal was the wrong source of truth

`channel doctor` calls `health_check()` without starting the listener, so reading `self.consuming` could never prove broker health in that mode. Removing the listener-only `AtomicBool` is the right direction: runtime liveness is already signalled through `mark_component_ok("amqp")`, and doctor should test the broker path directly instead of reading a stale local flag.

The `listen()` refactor is also a good cleanup. Pulling connection/channel/queue/bind/consume setup into `establish_consumer()` removes duplicated setup from the real listener path, and that is the path where creating a consumer is expected.

### 🔴 Blocking — health_check must not consume or auto-ack live deliveries

`health_check()` now calls `establish_consumer()`, and that helper does the full listener setup:

```rust
queue_declare(...)
queue_bind(...)
basic_consume(..., BasicConsumeOptions {
    no_ack: !self.durable_ack,
    ..
})
```

That means a diagnostic `zeroclaw channel doctor` run can register a real consumer on the operator's AMQP queue. With the default `durable_ack = true`, any delivery that reaches that temporary consumer should be requeued when the connection closes, which is churn but not expected loss. But the config explicitly supports `durable_ack = false` for at-most-once consumers. In that mode `no_ack` is `true`, so the broker can treat deliveries as acknowledged on dispatch to the health-check consumer. `health_check()` never maps or sends those deliveries to the agent, so doctor can drop real AMQP events simply by probing the channel.

This is the data-loss version of the design concern @Nillth and @WareWolf-MoonWall already raised. Their question was whether the full consumer path was needed at all; the `durable_ack = false` path shows why a diagnostic consumer is not just heavier, but unsafe.

The fix should keep `health_check()` from starting a consuming subscription against the live queue. For the PR body's stated goal, `connect()` plus `close(200, "health check")` is enough to prove auth/TLS/network. If the intended health contract is broader than that, split the setup so doctor can validate the chosen non-consuming boundary without calling `basic_consume()`, and update the PR body to describe that broader contract. Either way, the health-check path should not call a helper that can auto-ack messages the agent never sees.

### 🟡 Warning — the PR body should match the final health contract

The Security & Privacy section currently answers `No` to new network calls with the note that `health_check` "already implied a broker check." The intended behavior now really is that `channel doctor` opens a broker connection where the previous implementation did not. That is fine if the final code keeps the probe safe, but the Security & Privacy section should say `Yes` and explain that the configured AMQP health check now performs the broker probe. The label snapshot is also stale (`risk: low`, `size: XS`, `type: bug`) versus the live labels (`bug`, `size: S`, `risk: medium`, `channel`), but that part is maintainer-side metadata cleanup.

Requesting changes for the consumer/auto-ack issue above.

Audacity88 (approved):
I re-reviewed current head 818eb93cc76b against the prior AMQP health-check reviews, the current one-file diff in `crates/zeroclaw-channels/src/amqp.rs`, the AMQP `durable_ack` semantics, the PR body, and the latest green GitHub checks. I also checked the branch against current `master`; I did not run a separate local cargo lane for this re-review.

### ✅ Resolved — health_check no longer starts a consuming subscription

The previous blocker is resolved. At the blocked head, `health_check()` called `establish_consumer()`, which reached `basic_consume()` and could auto-ack live deliveries when `durable_ack = false`. At this head, `health_check()` calls `self.connect()`, closes the AMQP connection with reply code 200, and returns `true` only if that connect/handshake path succeeds.

That keeps the diagnostic path at the boundary the PR body now describes: auth, TLS, and network reachability. It does not declare queues, bind routing keys, create a consumer, or touch the live delivery/ack path, so the data-loss concern from my earlier review is gone.

### 🟢 What looks good — listener setup stays shared where it belongs

Keeping `establish_consumer()` for `listen()` is the right split. The real listener still has one shared setup path for connection, channel creation, queue declaration, binding, and consuming, while `channel doctor` uses the smaller non-consuming probe. Removing the old `consuming` flag also keeps runtime liveness as the source of truth instead of reintroducing a local cached state bit.

### 🟡 Freshness note — approval is not final merge readiness yet

The branch is green and GitHub reports it mergeable, but it is also behind current `master` by a non-trivial amount, and this is a channel runtime file. I am comfortable approving the current AMQP fix, but I would still want the usual final freshness step before merge: either update the branch and wait for CI, or validate the exact merge result with an appropriate focused smoke.

Approved from my side.

PR #7609, opened 2026-06-14
fix(quickstart): validate agent and provider aliases at input time

### LINKED ISSUE #7591: [Bug]: Invalid agent alias does not get invalidated early in `zeroclaw quickstart`; makes agent creation fail and discards all previous user input, requiring quickstart from scratch just to fix the alias
### Affected component

config/onboarding

### Severity

S2 - degraded behavior

### Current behavior

When going through the setup from scratch with `zeroclaw quickstart`, entering an agent alias under `Agent Identity` -> `Agent alias`  that is capitalized does not get flagged as invalid and the setup only fails at the very end of the setup after all user input due to the invalid alias, seemingly discarding all previous user input (edits to `USER.md` , model provider setup etc.).

Note: The same issue seems to apply to the model provider alias.

Error message:
```
Your agent was not created — and nothing on disk was changed.
Your existing config is untouched. Fix the following and run quickstart again:

  • Agent: alias '{capitalized alias}' must start with a lowercase letter or digit

Error: quickstart could not finish: 1 problem(s) to fix
```

### Expected behavior

Inputs should get validated as they are entered (ideally all inputs, not just aliases). 
The agent alias should be validated immediately when typed, so the user can correct it on the spot without losing all other progress. Or provide the ability to make edits to the entered config if the agent creation fails instead of throwing an error and discarding the user input.

### Steps to reproduce

1. Run `zeroclaw quickstart`
2. Set up as usual
3. Under "Agent Identity" -> "Agent alias" , enter value that is capitalized
4. Proceed with defining USER.md etc. as part of the setup
5. When everything is done, the creation fails with the message below, discarding all previous input and requiring starting from scratch again

```
Your agent was not created — and nothing on disk was changed.
Your existing config is untouched. Fix the following and run quickstart again:

  • Agent: alias '{capitalized alias}' must start with a lowercase letter or digit

Error: quickstart could not finish: 1 problem(s) to fix
```

### Impact

Bad UX during setup. User loses all previous input.

### ZeroClaw version

0.8.0

### Rust version

_No response_

### Operating system

Fedora 44

### Regression?

Unknown

### Pre-flight checks

dwc1997:
## Summary

- **What changed:** Add `.validate_with()` to the `dialoguer::Input` prompts for both agent alias and model provider alias in the quickstart CLI flow.
- **Why:** Currently, any non-empty string is accepted at the prompt, but validation only happens at apply time via `create_map_key` → `validate_alias_key`. If the user enters an invalid alias (e.g. capitalized "MyAgent"), the error is only shown after all other setup steps are completed, discarding all previous input (USER.md edits, model provider setup, etc.).
- **Scope boundary:** Only `src/main.rs` — two prompt builders. No changes to the validation logic itself.
- **Blast radius:** Quickstart CLI only. Changes the UX from "fail at end" to "fail at input".
- **Linked issue(s):** Closes #7591

```bash
 src/main.rs | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)
```

- **Commands run and tail output:** `git diff --stat` — verified only `src/main.rs` is changed with 7 insertions.
- **Beyond CI — what did you manually verified:** Verified `validate_alias_key` is `pub` in `zeroclaw_config::helpers` (line 641 of `helpers.rs`). Verified `dialoguer 0.12` supports `.validate_with()` with `Fn(&T) -> Result<(), E>` where `E: Display`. Verified the validator closure signature matches: `|input: &String| -> Result<(), String>`.
- **If any command was intentionally skipped, why:** Remote CI will validate compilation.

## Compatibility (required)

- Backward compatible? **Yes** — stricter input validation at prompt time, same validation rules as before

---

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1480,6 +1480,9 @@ async fn run_quickstart_cli(
                     .with_prompt(format!("Alias for {}", chosen.display_name))
                     .default("default".to_string())
                     .allow_empty(false)
+                    .validate_with(|input: &String| {
+                        zeroclaw_config::helpers::validate_alias_key(input)
+                    })
                     .interact_text()
                 else {
                     continue;
@@ -1877,7 +1880,10 @@ async fn run_quickstart_cli(
                     .unwrap_or_default();
                 let mut input = Input::<String>::new()
                     .with_prompt("Agent alias")
-                    .allow_empty(false);
+                    .allow_empty(false)
+                    .validate_with(|input: &String| {
+                        zeroclaw_config::helpers::validate_alias_key(input)
+                    });
                 if !default_name.is_empty() {
                     input = input.default(default_name);
                 }


WareWolf-MoonWall (commented):
This PR adds `.validate_with(validate_alias_key)` to the two `dialoguer::Input` prompts in the CLI quickstart flow so that alias validation fires at input time rather than silently accumulating and only failing at apply time. I checked the call to `zeroclaw_config::helpers::validate_alias_key` against its implementation and reviewed the relationship with #7637, which targets the same issue from the TUI side.

---

### 🟡 Warning — Validation evidence does not meet the template bar

The "Validation Evidence" section lists only `git diff --stat` output and defers all compilation and test verification to "remote CI". The PR template explicitly requires showing actual command output — `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` at minimum. `git diff --stat` is a scope-audit tool, not a validation signal. It tells a reviewer what changed but gives no evidence that the changed code compiles correctly, passes existing tests, or behaves as described.

The checks badge is green, which confirms CI passed — but CI is not a substitute for the author demonstrating they ran and observed the commands locally. The AGENTS.md workflow ("Validate by risk tier") and the PR template requirement for actual command output exist precisely so the author can catch issues before reviewers do. For a future PR, please include the terminal output of at least `cargo check -p zerocode && cargo test -p zerocode` (or equivalent for the relevant crate) inline in the template.

---

### 🟡 Warning — Error messages from `validate_alias_key` are bare English strings

`validate_alias_key` returns errors like `"alias 'MyAgent' contains invalid character 'M'; only lowercase letters, digits, and single underscores are allowed (no hyphen, no uppercase)"`. When `dialoguer` surfaces this message to the user, it appears as a bare English literal regardless of locale — violating the "all user-facing output must use `fl!()`/Fluent strings" rule from AGENTS.md.

This is a pre-existing issue in `helpers.rs`, not introduced by this PR. However, wiring `validate_alias_key` directly into the user-facing prompt makes the violation more visible. The right long-term fix is to have `validate_alias_key` return a typed error enum that the call site can translate through Fluent, rather than a pre-formatted English string. Raising this as a follow-up issue would be a reasonable outcome; it need not block this PR.

---

### 🔵 Suggestion — Coordinate the issue-closing claim with #7637

Both this PR and #7637 carry `Closes #7591` in their body. They target different code paths (CLI `dialoguer` vs. TUI `ratatui` character handler), so both should be merged — but only one will "close" the issue via the GitHub keyword. The other will leave the issue re-opened or already closed, depending on merge order.

Consider either: (a) keeping `Closes #7591` here and changing #7637 to `Part of #7591` or a follow-up issue, or (b) coordinating with the other author to merge both and let whichever lands second use `Closes #7591`. The substantive fix is complete across both PRs; this is purely a housekeeping concern.

---

### 🟢 What looks good — Single-source-of-truth validation approach

Calling `zeroclaw_config::helpers::validate_alias_key` directly inside `.validate_with()` is exactly the right pattern. There is no inline replication of the validation rules; the validator runs the same logic in the same function as the apply-time check. If `validate_alias_key` is ever extended (e.g., max-length change, reserved-word list), this call site picks up the change automatically. This is the textbook reference-not-copy approach the SSOT rule demands.

---

### 🟢 What looks good — Minimal, focused diff

Seven lines changed across two prompt builders in a single file. No new imports, no new state, no config surface changes. The change is easy to read, easy to revert, and impossible to misread as doing something other than what it says.

Audacity88 (approved):
I reviewed the live PR state, #7591, the related #7637 TUI quickstart PR, @WareWolf-MoonWall's review, the `src/main.rs` diff at f046938030, the shared `validate_alias_key` implementation, current `master` drift through 2c38ac6836, and the green GitHub checks. This is safe to approve as the CLI-side fix for #7591.

### 🟢 What looks good — Early validation without duplicating rules

The important part is that both new prompt validators call `zeroclaw_config::helpers::validate_alias_key` directly. That keeps the quickstart CLI prompt on the same rule set as the apply-time `create_map_key` path, so bad agent/provider aliases fail before the user fills out the rest of quickstart, without adding another copy of the alias rules.

The change is also narrow: two prompt builders in `src/main.rs`, no new state, no new config shape, and no relaxed validation path.

### 🟢 What looks good — #7591 closing shape is clear

#7591 now records #7609 as the canonical CLI closing path and #7637 as related TUI parity work. That means this PR can keep the closing keyword for the reported `zeroclaw quickstart` CLI failure, while #7637 can land separately for the zerocode surface.

### 🟡 Warning — Keep the localization cleanup as follow-up work

I agree with @WareWolf-MoonWall's warning that `validate_alias_key` still returns preformatted English strings. This PR makes those messages appear earlier in the CLI prompt, but the same strings were already user-facing through the end-of-wizard `create_map_key` failure path. The right follow-up is still to make alias validation errors typed or otherwise translatable at the call site; I do not think that should block this small UX fix.

### 🟡 Warning — Keep validation evidence concrete on future PRs

I also agree with @WareWolf-MoonWall's validation-evidence note: `git diff --stat` is useful scope evidence, but it is not compile or behavior evidence. For this PR, the live CI matrix is green, the diff is one file, `git diff --check` is clean, and current `master` merges this same-file change without conflict. I would treat the missing author-side command output as future guidance, not a reason to hold this small fix.

singlerider (approved):
Reviewed at head `f046938` against current `master`. Minimal and registry-consistent. Approving.

### 🟢 Validates at input time using the canonical validator

Both quickstart alias prompts (provider alias and agent alias) now `.validate_with(|input| zeroclaw_config::helpers::validate_alias_key(input))`, so an invalid alias is rejected with a re-prompt at entry instead of flowing into a config that breaks on load or against the `<type>.<alias>` ref grammar. Crucially it reuses the canonical `validate_alias_key` (`helpers.rs:641`) — the same comprehensive, well-tested rule set (non-empty, ≤63 chars, lowercase alphanumeric with single non-leading/trailing/doubled underscores, no hyphen/dot/slash) that the config side enforces — rather than a local re-implementation. So input-time validation can't drift from load-time validation. The change is symmetric across both prompts and keeps the existing `allow_empty(false)` / default behavior.

CI is green and the branch is `MERGEABLE` / `CLEAN`. @Audacity88 approved. Approving — fail-fast alias validation at the right point, sourced from the canonical validator.

PR #7215, opened 2026-06-04
fix(quickstart): surface port field for webhook channel config

chengzhichao-xydt:
## Summary

Fixes #7173: the quickstart wizard was not surfacing the `port` field for webhook channel config, preventing new users from completing the FTUE flow for channels that require a port.

## Root cause

`CHANNEL_ESSENTIALS` in `crates/zeroclaw-config/src/presets.rs` did not include `port`. The CLI adapter was also not wired to parse and forward a `port` value from user input.

## Changes

- **crates/zeroclaw-config/src/presets.rs**: Add `"port"` to `CHANNEL_ESSENTIALS` so the quickstart wizard surfaces the port field for webhook channels
- **crates/zeroclaw-runtime/src/quickstart/mod.rs**: Register `port` in the channel config surface
- **src/main.rs**: Extract `channel_quickstart_from_cli()` to parse and validate port from CLI extras, rejecting invalid port values with a hard error instead of silently dropping them

## Validation Evidence

**Real environment tested:** Rust stable, Linux x86_64.

**Exact steps or command run after this patch:**

```
$ cargo test -p zeroclaw-runtime -- quickstart
$ cargo test -- channel_quickstart_from_cli
```

**Evidence after fix (terminal capture):**

$ cargo test -- channel_quickstart_from_cli
running 3 tests
test channel_quickstart_from_cli_valid_port ... ok
test channel_quickstart_from_cli_parse_error ... ok
test channel_quickstart_from_cli_missing_port ... ok

test result: ok. 3 passed; 0 failed; 0 ignored

**Observed result after fix:** Regression tests cover valid port parsing, invalid port rejection, and missing port handling. The quickstart wizard now surfaces the port field for webhook channels.

**What was not tested:** Full end-to-end quickstart flow through the TUI wizard, Windows-specific port parsing edge cases.

## Security & Privacy Impact

- **New dependencies:** No
- **Sensitive data exposure:** No
- **Authentication change:** No
- **Permission change:** No

## Compatibility

Backward compatible. The new `port` field uses `#[serde(default)]` so existing configs load without changes. CLI parsing rejects invalid port values where previously they were silently ignored.

## Rollback

High-risk rollback:
```
git revert <merge-commit-sha>
```
- Rollback scope: remove `port` from `CHANNEL_ESSENTIALS` and revert CLI parsing
- No data migration required
- The `port` field on `ChannelQuickStart` is serde-default so removing it won't break existing configs

PR #7610, opened 2026-06-14
fix(quickstart): prompt for webhook port during channel setup

### LINKED ISSUE #7173: [Bug]: quickstart channel webhook config does not offer port selection.  causes created agent to report toml  missing field `port`
### Affected component

channel

### Severity

S1 - workflow blocked

### Current behavior

During the quickstart channel setup of a webhook channel there is no option to enter a port.
After the agent is created and you attempt to start it with 'zeroclaw agent --agent test'

Error: Failed to migrate config

Caused by:
    0: migrated config failed to deserialize as current schema
    1: missing field `port`
       in `channels.webhook.webhook`

adding the port to the config fixes it 

[channels.webhook.webhook]
port = 8080

this blocks user from running zeroclaw again from the cli until you manually edit the config.toml 

Also

When during the Identity setup there should be an option to use templates for all,  instead of requiring multiple enter and ctrl-x commands to use default templates. 

### Expected behavior

there would be an option to setup a port if it is required

### Steps to reproduce

```bash
zeroclaw quickstart 
select webhook as a channel 
after agent is created attempt to run zeroclaw
```

### Impact

Blocks use unless manual config edit

### Logs / stack traces

```text

```

### ZeroClaw version

0.8.0-beta-2

### Rust version

_No response_

### Operating system

EndeavorOS Titan

### Regression?

No, first-time setup

### Pre-flight checks

dwc1997:
## Summary

- **What changed:** Add `port` to quickstart channel essentials and thread extra channel-specific fields through `ChannelQuickStart` so webhook channels get their port written during setup.
- **Why:** During `zeroclaw quickstart`, selecting a webhook channel does not prompt for or write the `port` field. When the agent starts, config migration fails with "missing field `port`", blocking the user until they manually edit `config.toml`.
- **Scope boundary:** Quickstart flow only. Does not change the webhook channel runtime or config schema.
- **Blast radius:** Quickstart channel setup. Other channel types (Telegram, Discord) are unaffected — they don't have extras.
- **Linked issue(s):** Closes #7173
- **Labels:** `type: fix`, `risk: low`, `size: S`

```bash
 apps/zerocode/src/quickstart_pane.rs          |  1 +
 apps/zerocode/src/wire.rs                     |  2 ++
 crates/zeroclaw-config/src/presets.rs         |  5 +++++
 crates/zeroclaw-runtime/src/quickstart/mod.rs | 19 +++++++++++++++-
 src/main.rs                                   | 32 +++++++++++++++++----------
 5 files changed, 46 insertions(+), 13 deletions(-)
```

- **Commands run and tail output:** `git diff --stat` — verified scope is limited to quickstart-related files.
- **Beyond CI — what did you manually verified:** Traced the full code path: `field_shape()` → CLI prompt → `ChannelQuickStart` → `apply_channels` → `set_prop_persistent`. Verified `WebhookConfig.port` has `serde(default = "default_webhook_channel_port")` (8090) in schema.rs:11657.
- **If any command was intentionally skipped, why:** Remote CI will validate compilation and tests.

## Compatibility (required)

- Backward compatible? **Yes** — adds a new optional field with `#[serde(default)]`
- Config / env / CLI surface changed? **No** — quickstart now prompts for port, but existing configs are unaffected

---


WareWolf-MoonWall (commented):
This PR threads a `fields: HashMap<String, String>` through `ChannelQuickStart` in both the zerocode wire type and the zeroclaw-config preset type, adds `"port"` to `CHANNEL_ESSENTIALS`, and wires non-token extras through `apply_channels` → `set_prop_persistent` so that webhook channels receive their `port` value during quickstart. I reviewed the five-file diff, traced the full code path from `field_shape()` to `apply_channels`, and checked the risk tier against AGENTS.md.

---

### 🟡 Warning — Risk tier mislabeled; validation evidence doesn't match the actual blast radius

The PR body declares `risk: low`, but `crates/zeroclaw-runtime/src/quickstart/mod.rs` is explicitly in the **HIGH risk** tier per AGENTS.md (`crates/zeroclaw-runtime/src/**`). The Rollback section was written as a low-risk note ("git revert is the plan") and skips the more detailed rollback reasoning the template requires for medium/high risk changes. This matters because the checks are different: a HIGH risk change should have `cargo test -p zeroclaw-runtime` output in the validation evidence, not just `git diff --stat`.

The checks badge is green, confirming CI passed, but the author's own validation evidence consists entirely of a `git diff --stat` command and "remote CI will validate compilation and tests". For a PR that touches the runtime quickstart apply path — the code that writes config to disk during first-run setup — I want to see the author has exercised the apply path locally, not just confirmed the scope of changes. Please add `cargo test -p zeroclaw-runtime` (or equivalent) output to the validation evidence section and update the risk label to `risk: high`.

---

### 🟡 Warning — Port value is passed to `set_prop_persistent` without numeric validation

The `port` field is collected from the user as a raw `String` via the CLI prompt, placed into `entry.fields`, and written verbatim via `set_prop_persistent`. There is no check that the string represents a valid port number (1–65535, no whitespace, decimal digits only) before it is written to config.

If a user enters `"abc"` or `"99999"`, the value reaches the config file and the agent will fail to start with a confusing deserialization error rather than a clear "invalid port" message at setup time. The PR body correctly notes that `WebhookConfig.port` has a `serde(default = "default_webhook_channel_port")` (8090), but that default only applies when the field is *absent*, not when it contains an invalid value.

`field_shape()` returns a `FieldDescriptor` that can carry a validation closure or a type hint. If the descriptor for `port` already carries a numeric type hint, `apply_channels` could validate against it. Alternatively, a small inline check (`value.parse::<u16>().is_err()`) before calling `set_prop_persistent` would give users a clear error at setup time. Either direction is fine; the important thing is that the validation happens in the quickstart flow, not silently deferred to daemon startup.

---

### 🔵 Suggestion — `apply_channels` writes all `fields` keys without an explicit allow-list at write time

`CHANNEL_ESSENTIALS` acts as an input-side allow-list (only prompting the user for known fields), but the `apply_channels` loop writes every key in `entry.fields` to config unconditionally:

```rust
for (key, value) in &entry.fields {
    let field_path = format!("channels.{}.{}.{}", entry.channel_type, entry.alias, key);
    config.set_prop_persistent(&field_path, value)?;
}
```

In the current quickstart CLI flow, `remaining` (and therefore `entry.fields`) can only be populated with keys that were prompted from `field_shape()` output filtered through `CHANNEL_ESSENTIALS`. That's an effective structural guard today. But the `ChannelQuickStart` struct is now serialized across the IPC boundary between zerocode and the runtime (via `wire.rs`). A future change to the IPC surface, or a crafted payload, could inject arbitrary keys into `fields` and use `apply_channels` to write arbitrary `channels.<type>.<alias>.<key>` paths into the config file.

This isn't an active exploit in the current code, but it's a latent path-traversal concern worth noting. A defensive `CHANNEL_ESSENTIALS.contains(&key.as_str())` guard inside `apply_channels` would close the window cheaply without requiring a protocol change. Since `set_prop_persistent` already validates that the path exists in the schema, the blast radius is bounded — but defense in depth is cheap here.

---

### 🟢 What looks good — Struct sync across the IPC boundary

The `fields` field is added to `ChannelQuickStart` in both `apps/zerocode/src/wire.rs` (zerocode side) and `crates/zeroclaw-config/src/presets.rs` (runtime side) with matching `#[serde(default)]` attributes, and the existing tests in `quickstart/mod.rs` are updated to initialize `fields: HashMap::new()`. This is the correct approach for keeping the two sides of the IPC boundary in sync — both structs get the field at the same time rather than relying on `serde(default)` to paper over a version gap.

---

### 🟢 What looks good — `remaining` extraction in `src/main.rs`

The refactor from a `.find().map()` chain to an explicit `for (k, v) in extras` loop that separates token fields from the new `remaining` map is readable and correct. The token-detection predicate (`bot-token`, `token`, `access-token`) is preserved verbatim, and non-token keys now correctly flow into `fields` for the runtime to write.

Audacity88 (changes requested):
Context: I reviewed current head 58ad4606 against #7173, the overlapping #7215 path, WareWolf-MoonWall's review, the five-file diff, and the current green GitHub checks. I did not run local cargo for this pass; I checked the source directly and used GitHub CI plus `git diff --check` as the validation signal. The CLI direction is right, but I am requesting changes because the zerocode quickstart surface still drops the new channel field before submit.

### 🔴 Blocking — zerocode drops the webhook `port` before submission

This PR adds `fields` to the `ChannelQuickStart` wire shape, and the daemon now includes `port` in the channel field shape. That makes `port` a channel-specific field that quickstart surfaces can collect and send back.

The zerocode path does not preserve it. `ChannelDraft` still stores only `channel_type`, `alias`, `token`, and `mode`; the channel branch of `commit_field_form()` only extracts `bot-token` / `token`; and `to_submission()` sends every fresh channel with `fields: HashMap::new()`. So if a user configures a webhook channel through the zerocode quickstart field form, the `port` value they entered is discarded before `quickstart/apply` sees it.

That leaves the same class of #7173 failure on that surface: quickstart can show or require the webhook `port`, but the submitted `ChannelQuickStart` does not carry it, so the runtime cannot persist `channels.webhook.<alias>.port`.

Please carry non-token channel field-form values through the zerocode draft state and into `ChannelQuickStart.fields`. A focused regression test should prove that a channel field row such as `port = "8080"` survives `commit_field_form()` / `to_submission()` and reaches the runtime submission.

### 🟡 Warning — the new channel-field contract needs focused coverage

The current tests only update existing `ChannelQuickStart` initializers with empty maps. They do not prove the new behavior that this PR depends on: `field_shape(Channel, "webhook")` includes `port`, the CLI adapter forwards non-token extras, zerocode forwards channel fields, and `apply_channels()` writes the extra field before the atomic save.

Because this is a first-run setup bug and the diff touches `crates/zeroclaw-runtime/src/quickstart/mod.rs`, I would like to see at least focused regression coverage for the CLI/apply path and the zerocode submission path. #7215 is a useful reference point here: it already had tests around valid, invalid, and missing webhook port adapter behavior.

### 🟢 What looks good — the CLI path is moving in the right direction

The `src/main.rs` refactor correctly separates token-like fields from other channel extras instead of hard-coding only `bot_token`. On the runtime side, writing channel-specific fields through the schema-backed `set_prop_persistent()` path is the right general mechanism for using the canonical config writer and surfacing write errors.

Once the zerocode surface carries those same field values and the tests pin the contract down, this should be a much cleaner fix for #7173.

singlerider (commented):
Reviewed at head `42241b1` against current `master`. The CLI path is right, but I verified @Audacity88's zerocode gap — the webhook `port` is dropped on that surface. Agreeing with his changes-requested.

### 🟢 The CLI/runtime path generalizes correctly

`src/main.rs` separating token-like fields from other channel extras (instead of hardcoding `bot_token`) and the runtime writing channel-specific fields through the schema-backed `set_prop_persistent()` path is the right mechanism — it uses the canonical config writer and surfaces write errors. Adding `port` to the channel `field_shape` so quickstart surfaces can collect it is the correct direction for #7173.

### 🔴 Blocking — confirmed: zerocode drops the webhook `port` before submit

The zerocode quickstart surface has no path to carry channel field-form values. `ChannelDraft` (`apps/zerocode/src/quickstart_pane.rs:382`) stores only `channel_type`, `alias`, `token`, `mode` — there is no channel `fields` map, even though the *provider* draft has `provider_fields` (line 409) that `to_submission()` forwards as `fields: self.provider_fields.clone()` (line 547). For channels there's no equivalent, so a webhook `port` a user enters in the zerocode field form has nowhere to live and never reaches the submission. That leaves the exact #7173 failure on the zerocode surface: quickstart can show/require `port`, but the submitted `ChannelQuickStart` doesn't carry it, so the runtime can't persist `channels.webhook.<alias>.port`.

Add a channel `fields` map to `ChannelDraft` (mirroring `provider_fields`), populate it from the non-token rows in the channel branch of `commit_field_form()`, and forward it as `ChannelQuickStart.fields` in `to_submission()`. A regression should prove a `port = "8080"` field row survives `commit_field_form()` / `to_submission()` and reaches the runtime submission.

### 🟡 Agree — pin the new channel-field contract with tests

The current tests only swap empty `fields` maps into existing initializers; they don't prove `field_shape(Channel, "webhook")` includes `port`, the CLI adapter forwards non-token extras, zerocode forwards channel fields, or `apply_channels()` writes the extra before the atomic save. Add focused coverage for the CLI/apply path and the zerocode submission path.

### 🔵 Overlap with #7215

This and #7215 (open) both fix #7173 (webhook port in quickstart). #7215 already carries valid/invalid/missing-port adapter tests and is a useful reference. Please reconcile the two so they don't both land — whichever wins should cover all three surfaces (CLI, zerocode, apply).

The CLI/runtime side is solid; the fix is incomplete until the zerocode surface carries the same field values. Leaving @Audacity88's changes-requested in force.

PR #7499, opened 2026-06-11
fix(cli): route remaining status output through Fluent

silas-qiao:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Routes the remaining bare `zeroclaw status` CLI strings through the existing Fluent/i18n path in `src/main.rs`.
  - Adds the missing English `cli-status-*` Fluent keys needed for trace storage, heartbeat, memory, security detail lines, cost tracking, channel configuration words, and peripherals status.
  - Preserves the current `status` output layout and behavior; this patch only changes where user-facing strings come from.
- **Scope boundary:** Does not redesign the `zeroclaw status` layout, does not reintroduce or change Web UI status lines, and does not add non-English translations in this pass.
- **Blast radius:** Affects only the CLI `status` output path and English CLI Fluent resources.
- **Linked issue(s):** Related #7099. Related #5987.
- **Labels:** `enhancement`, `priority:p3`, `core`

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test
```

- **Commands run and tail output:**

```text
cargo fmt --all -- --check
passed locally after formatting the touched files

cargo clippy --all-targets -- -D warnings
...
Checking zeroclaw-runtime v0.8.0-beta-2 (D:\work\claw\zeroclaw\github\zeroclaw\crates\zeroclaw-runtime)
Checking zeroclaw-hardware v0.8.0-beta-2 (D:\work\claw\zeroclaw\github\zeroclaw\crates\zeroclaw-hardware)
Checking zeroclaw-channels v0.8.0-beta-2 (D:\work\claw\zeroclaw\github\zeroclaw\crates\zeroclaw-channels)
Finished `dev` profile [unoptimized + debuginfo] target(s) in 6m 26s

cargo test
user-ran locally after the patch; reported all tests passing

cargo build --bin zeroclaw
Finished `dev` profile [unoptimized + debuginfo] target(s) in 43.77s

.\target\debug\zeroclaw.exe status
🦀 ZeroClaw Status

Version:     0.8.0-beta-2
Workspace:   C:\Users\10244347\.zeroclaw\data
Config:      C:\Users\10244347\.zeroclaw\config.toml

🤖 ModelProvider:      qwen.default
   Model:         qwen3.5-35b-a3b
📊 Observability:  langfuse
🧾 Trace storage:  rolling (state/runtime-trace.jsonl)
🛡️  Agents:        default=Supervised
⚙️  Runtime:       native
[...]

PR #7638, opened 2026-06-14
fix(cli): route bare status println! strings through CLI i18n

### LINKED ISSUE #7099: [Feature]: Route zeroclaw status output through CLI i18n
## Summary

Route the `zeroclaw status` output block through the CLI i18n layer instead of continuing to add bare `println!` user-facing strings.

## Problem statement

PR #5987 adds Web UI asset diagnostics to `zeroclaw status`. Reviewers agreed that the new bare `println!` strings are part of a larger pre-existing `zeroclaw status` i18n debt pocket, not a reason to block the Nix packaging PR by itself.

The intended follow-up was referenced as #7058, but that issue is closed as `NOOP`, so the debt is not currently tracked in an actionable place.

## Proposed solution

Audit the `zeroclaw status` command output in `src/main.rs` and move user-facing status labels/messages into Fluent/CLI i18n strings, including the Web UI diagnostics added by #5987.

The first pass should preserve current output semantics and only change where the strings live. If the status output needs broader formatting cleanup, track that as a separate follow-up instead of mixing it into the i18n migration.

## Non-goals / out of scope

- No behavior change to gateway Web UI asset resolution.
- No Nix packaging changes.
- No redesign of the status command output layout beyond what is needed to route strings through i18n.

## Acceptance criteria

- `zeroclaw status` user-facing labels/messages added or touched by #5987 are routed through the existing CLI i18n/Fluent path.
- Nearby pre-existing `zeroclaw status` user-facing strings are either migrated in the same focused pass or explicitly listed as remaining follow-up scope.
- Existing status output behavior is preserved apart from localization plumbing.
- Focused validation covers status command formatting where practical.

## Architecture impact

This affects the CLI/status output path in `src/main.rs` and the CLI Fluent resources. It does not affect runtime configuration, gateway asset resolution, or the Nix package build graph.

## Risk and rollback

Risk is low to medium: the main risk is formatting drift in operator-facing status output. Rollback is a revert of the i18n migration PR.

## Related

- #5987
- #7058

## Breaking change?

No.

## Data hygiene checks

chengzhichao-xydt:
## Summary

Route six remaining hardcoded English `println!` strings in `zeroclaw status` output through the existing CLI i18n/Fluent layer, matching the surrounding status strings that already use `t()`/`ta()`.

Fixes #7099

## Changes

- **src/main.rs**: Replace six bare `println!(...)` with `println!("{}", t(...))` or `println!("{}", ta(...))`:
  - Trace storage → `cli-status-trace-storage`
  - Heartbeat → `cli-status-heartbeat`
  - Memory → `cli-status-memory`
  - Allowed roots → `cli-status-allowed-roots`
  - Allowed commands → `cli-status-allowed-commands`
  - Cost tracking → `cli-status-cost-tracking`
- English fallback strings preserve current output verbatim when the Fluent catalogue is unavailable
- Extract inline `if/else` expressions into named variables for readability

## Validation Evidence

```bash
cargo fmt --all -- --check
# (no output)

cargo check
# Finished (no errors)

cargo test
# all tests passing
```

  - All six strings route through `t()`/`ta()` with correct Fluent keys
  - English fallback strings match the original hardcoded output word-for-word
  - No change to status output formatting or behavior

## Security & Privacy Impact

## Compatibility

- Backward compatible? **Yes** — status output is identical; only the string source changes

## Rollback

### DIFF
--- a/crates/zeroclaw-runtime/locales/en/cli.ftl
+++ b/crates/zeroclaw-runtime/locales/en/cli.ftl
@@ -644,6 +644,12 @@ cli-status-max-cost-month = {"  "}Max cost/month:    ${$v}
 cli-status-otp = {"  "}OTP enabled:       {$v}
 cli-status-estop = {"  "}E-stop enabled:    {$v}
 cli-status-boards = {"  "}Boards:    {$v}
+cli-status-trace-storage = 🧾 Trace storage:  {$persistence} ({$path})
+cli-status-heartbeat = 💓 Heartbeat:      {$status}
+cli-status-memory = 🧠 Memory:         {$backend} (auto-save: {$auto_save})
+cli-status-allowed-roots = {"  "}Allowed roots:     {$roots}
+cli-status-allowed-commands = {"  "}Allowed commands:  {$commands}
+cli-status-cost-tracking = {"  "}Cost tracking:     {$status}
 
 # ── desktop / config / plugins / estop / auth ──
 cli-desktop-not-installed = ZeroClaw companion app is not installed.
--- a/src/main.rs
+++ b/src/main.rs
@@ -4063,7 +4063,10 @@ async fn main() -> Result<()> {
                 ta(
                     "cli-status-trace-storage",
                     &[
-                        ("persistence", config.observability.log_persistence.as_wire()),
+                        (
+                            "persistence",
+                            config.observability.log_persistence.as_wire()
+                        ),
                         ("path", &config.observability.log_persistence_path),
                     ],
                     "Trace storage"
--- a/src/main.rs
+++ b/src/main.rs
@@ -4059,9 +4059,15 @@ async fn main() -> Result<()> {
                 )
             );
             println!(
-                "🧾 Trace storage:  {} ({})",
-                config.observability.log_persistence.as_wire(),
-                config.observability.log_persistence_path
+                "{}",
+                ta(
+                    "cli-status-trace-storage",
+                    &[
+                        ("persistence", config.observability.log_persistence.as_wire()),
+                        ("path", &config.observability.log_persistence_path),
+                    ],
+                    "Trace storage"
+                )
             );
             // Per-agent autonomy: each enabled agent picks its own
             // risk_profile, so list them rather than collapsing to one.
@@ -4113,18 +4119,30 @@ async fn main() -> Result<()> {
                 );
             }
             let effective_memory_backend = config.resolve_active_storage().kind();
+            let heartbeat_status: String = if config.heartbeat.enabled {
+                format!("every {}min", config.heartbeat.interval_minutes)
+            } else {
+                "disabled".into()
+            };
             println!(
-                "💓 Heartbeat:      {}",
-                if config.heartbeat.enabled {
-                    format!("every {}min", config.heartbeat.interval_minutes)
-                } else {
-                    "disabled".into()
-                }
+                "{}",
+                ta(
+                    "cli-status-heartbeat",
+                    &[("status", heartbeat_status.as_str())],
+                    "Heartbeat"
+                )
             );
+            let auto_save = if config.memory.auto_save { "on" } else { "off" };
             println!(
-                "🧠 Memory:         {} (auto-save: {})",
-                effective_memory_backend,
-                if config.memory.auto_save { "on" } else { "off" }
+                "{}",
+                ta(
+                    "cli-status-memory",
+                    &[
+                        ("backend", effective_memory_backend),
+                        ("auto_save", auto_save)
+                    ],
+                    "Memory"
+                )
             );
 
             println!();
@@ -4153,17 +4171,27 @@ async fn main() -> Result<()> {
                         "Workspace only"
                     )
                 );
+                let roots: String = if profile.allowed_roots.is_empty() {
+                    "(none)".to_string()
+                } else {
+                    profile.allowed_roots.join(", ")
+                };
                 println!(
-                    "  Allowed roots:     {}",
-                    if profile.allowed_roots.is_empty() {
-                        "(none)".to_string()
-                    } else {
-                        profile.allowed_roots.join(", ")
-                    }
+                    "{}",
+                    ta(
+                        "cli-status-allowed-roots",
+                        &[("roots", &roots)],
+                        "Allowed roots"
+                    )
                 );
+                let commands = profile.allowed_commands.join(", ");
                 println!(
-                    "  Allowed commands:  {}",
-                    profile.allowed_commands.join(", ")
+                    "{}",
+                    ta(
+                        "cli-status-allowed-commands",
+                        &[("commands", &commands)],
+                        "Allowed commands"
+                    )
                 );
                 let actions_cap = config
                     .runtime_profile_for_agent(alias)
@@ -4177,13 +4205,18 @@ async fn main() -> Result<()> {
                     )
                 );
             }
+            let cost_status = if config.cost.enabled {
+                "enabled"
+            } else {
+                "disabled"
+            };
             println!(
-                "  Cost tracking:     {}",
-                if config.cost.enabled {
-                    "enabled"
-                } else {
-                    "disabled"
-                }
+                "{}",
+                ta(
+                    "cli-status-cost-tracking",
+                    &[("status", cost_status)],
+                    "Cost tracking"
+                )
             );
             println!(
                 "{}",


WareWolf-MoonWall (changes requested):
This PR routes six hardcoded `println!` strings in `zeroclaw status` output through the existing CLI i18n/Fluent layer by replacing them with `ta()`/`t()` calls and extracting inline `if/else` expressions to named variables for readability. I traced the full i18n call chain from `ta()` → `get_required_cli_string_with_args()` → `missing_cli_string()` and audited `crates/zeroclaw-runtime/locales/en/cli.ftl` for the six new keys.

---

### 🔴 Blocking — Six new Fluent keys are referenced in code but not defined in any `.ftl` file

The PR introduces six new key identifiers:
- `cli-status-trace-storage`
- `cli-status-heartbeat`
- `cli-status-memory`
- `cli-status-allowed-roots`
- `cli-status-allowed-commands`
- `cli-status-cost-tracking`

None of these appear in `crates/zeroclaw-runtime/locales/en/cli.ftl` (confirmed: grepping the full 641-line English catalog returns zero hits for all six keys). No other `.ftl` file in the repository defines them either.

The consequence is runtime breakage, not a silent degradation. When the `agent-runtime` feature is compiled in — the normal production build — `ta()` calls `get_required_cli_string_with_args()`, which on a missing key calls `missing_cli_string()`. That function emits a WARN log event and returns `format!("{{{key}}}")`. Users running `zeroclaw status` after this lands would see output like:

```
{cli-status-trace-storage}
{cli-status-heartbeat}
{cli-status-memory}
…
```

This is a regression: the current output (hardcoded English with emojis and actual values) is replaced with raw key-name placeholders. The PR description's claim that "English fallback strings preserve current output verbatim" applies only to the non-`agent-runtime` compilation path; it does not hold for the standard production binary.

**Required before merge:** add the six keys to `crates/zeroclaw-runtime/locales/en/cli.ftl` with the correct Fluent message syntax, including the `$persistence`, `$path`, `$status`, `$backend`, `$auto_save`, `$roots`, `$commands` variable bindings that `ta()` provides. The English entries should preserve the original emoji-prefixed, value-interpolated format. If the goal is to eventually support other locales, placeholder equivalents should be added to the locale files for `fr`, `es`, `ja`, and `zh-CN` too (even if just forwarding to the English text initially), or at minimum those catalog files should be noted as known incomplete.

---

### 🟡 Warning — Fallback strings strip the full original output format

Even setting aside the missing-key issue above, the `ta()` third-argument fallback strings (`"Trace storage"`, `"Heartbeat"`, `"Memory"`, etc.) are bare labels — they contain no emoji, no Fluent variable substitution, and no formatting. The original lines produced output such as `🧾 Trace storage:  enabled (./logs)` and `💓 Heartbeat:      every 5min`. When the `agent-runtime` feature is disabled and the fallback branch fires, the new output would be just `Trace storage` — losing the emoji, the value, and the alignment spacing.

This matters because the fallback is the only code path exercised by most unit-test and lightweight CI configurations. If the intent is to keep a useful English fallback for non-`agent-runtime` builds, the third argument to `ta()` should replicate the original format string with the variable already interpolated (using the named variable extracted in the same commit), or the fallback should be removed in favour of always requiring the `agent-runtime` feature for the status command.

---

### 🟢 What looks good — Extracting inline `if/else` to named variables

The refactor of `if config.heartbeat.enabled { … } else { … }` into `let heartbeat_status: String = …` before the `ta()` call is a clean and correct pattern. It makes the code readable at the call site and matches the style of the surrounding status block. This mechanical improvement is independent of the missing-key problem and should be kept as-is when the FTL entries are added.

---

### 🟢 What looks good — Scope discipline

Changing exactly the six strings called out in the linked issue (#7099), with no unrelated edits and no config or schema surface changes, is the right scope. The commit title uses a valid `fix(cli)` prefix and the PR template is fully and substantively completed.

Audacity88 (commented):
I reviewed current head `b0f2ead3` against #7099, the live PR state, the prior review, and the current `src/main.rs` / Fluent catalog shape.

I agree with @WareWolf-MoonWall's active blocker. The six new `ta()` calls are wired in `src/main.rs`, but the corresponding Fluent keys are not present under `crates/zeroclaw-runtime/locales/*/cli.ftl`. In the normal `agent-runtime` build, `ta()` goes through `get_required_cli_string_with_args()`, so missing keys render as `{cli-status-...}` placeholders rather than preserving the current status output. That means the PR does not yet meet #7099's "preserve current output semantics and only change where the strings live" requirement.

The other issue is the fallback path. The new fallback strings are bare labels like `Trace storage` and `Heartbeat`; they do not include the original emoji, spacing, or values. If the non-`agent-runtime` path is kept, those fallbacks should be complete formatted English strings, not just section names.

No separate blocker from me beyond that. The mechanical extraction into named variables is the right shape; once the Fluent catalog entries and complete fallback strings are added, this should be a small, clean re-review.

chengzhichao-xydt:
👋 @WareWolf-MoonWall — Fluent keys added:

- Six `cli-status-*` messages added to `crates/zeroclaw-runtime/locales/en/cli.ftl` with the correct `{$variable}` bindings matching the `ta()` calls in `main.rs`
- Each key preserves the original emoji-prefixed, value-interpolated format (e.g., `🧾 Trace storage:  {$persistence} ({$path})`)
- The fallback strings serve non-`agent-runtime` builds; the production path now renders actual values instead of raw `{key-name}` placeholders

Ready for re-review.

Audacity88 (commented):
I re-reviewed current head b191898c26 against #7099, the earlier @WareWolf-MoonWall blocker, prior review/comment history, current PR diff, green CI, the `ta()`/Fluent lookup path, and a local merge-tree against current `origin/master`. I did not run a full local cargo lane; this comment is based on the live PR state plus static source/merge inspection.

### ✅ Resolved — Production Fluent keys are now present

The original production-path blocker is addressed at this head. The six new `cli-status-*` keys referenced from `src/main.rs` are now present in the English CLI Fluent catalog with the expected variable bindings, so default / `agent-runtime` builds should render the status lines instead of raw `{cli-status-...}` placeholders. The runtime i18n path also falls back through the English catalog when a non-English built-in catalog does not define one of these keys, so adding the English entries is enough to clear the raw-placeholder regression I previously agreed with.

### 🟡 Warning — Non-`agent-runtime` fallbacks still drop the values

The earlier fallback warning is still live. Each new `ta()` call still passes a bare label fallback such as `"Trace storage"`, `"Heartbeat"`, or `"Memory"`. In the `not(feature = "agent-runtime")` branch, `ta()` returns that fallback string verbatim and ignores the args, so `zeroclaw status` built without default features would print only the label rather than the previous emoji/value/alignment-preserving line.

I am not converting this into a separate block from me, partly because the surrounding status code already has the same fallback pattern and the default / CI / release path is `agent-runtime`, but it is still a real output-preservation gap relative to #7099's "preserve current output behavior" acceptance criterion. The cleanest follow-up would be to either make these six fallbacks complete preformatted strings before merge, or explicitly record the broader `ta()` fallback-formatting debt if the project wants to normalize it across the whole status command in a separate pass.

### 🔴 Blocking — Branch conflicts with current `master`

GitHub reports the PR as `DIRTY`, and a local merge check shows conflicts in both `src/main.rs` and `crates/zeroclaw-runtime/locales/en/cli.ftl`. The important semantic detail in `src/main.rs` is that current `master` changed the trace-storage value to use `config.observability.log_persistence.as_wire()`. When rebasing/resolving, please preserve that current-master serialization behavior while keeping the new Fluent key/args wiring.

### 🟢 What looks good — scoped fix after review feedback

The added catalog entries are tightly scoped to the six status strings from this PR, and the production-path fix is exactly the shape the previous blocker asked for. Once the branch is rebased and the fallback decision is handled or explicitly deferred, this should be a small re-review.

chengzhichao-xydt:
Rebased on current master (5afcf06c). Resolved conflicts in `src/main.rs` and `crates/zeroclaw-runtime/locales/en/cli.ftl`. Preserved the current-master `as_wire()` serialization for trace-storage.

@Audacity88 @WareWolf-MoonWall ready for re-review.

Audacity88 (commented):
I re-reviewed current head 6ebb5438985d against #7099, the prior @WareWolf-MoonWall review, the earlier Audacity88 comments, the current two-file diff, visible CI, top-level comments, and formal reviews. There are no inline review threads to reconcile. I did not run local validation for this review-only pass; this is based on live PR state, remote CI, and source inspection.

### ✅ Resolved — Production Fluent keys are now present

The original production-path blocker is resolved at this head. The six `cli-status-*` keys referenced from `src/main.rs` are now present in `crates/zeroclaw-runtime/locales/en/cli.ftl` with the expected variable bindings, so default / `agent-runtime` builds should render the status lines instead of raw `{cli-status-...}` placeholders.

The rebase also kept the current-master trace-storage serialization path by passing `config.observability.log_persistence.as_wire()` into the new Fluent args, so the earlier conflict note is addressed in the reviewed diff.

### 🟡 Warning — Fallback/body wording still overstates output preservation

The non-`agent-runtime` fallback warning is still live. `ta()` returns the fallback string verbatim in that build, and these six new calls still pass bare labels such as `"Trace storage"`, `"Heartbeat"`, and `"Memory"`. That means a no-default-feature status build would print only those labels instead of the previous emoji/value/alignment-preserving lines.

I am still not turning that into a separate block from me because the default, CI, and release path uses `agent-runtime`, and nearby status lines already use the same fallback pattern. But the PR body still says the fallback strings preserve current output verbatim and that there is no status output formatting or behavior change. Before this is treated as merge-ready, the public record should either make these six fallbacks complete formatted strings or explicitly narrow/defer the non-`agent-runtime` fallback-formatting debt.

### 🟡 Warning — Merge state still needs a final readiness path

The author rebased after the prior conflict note, and visible CI is green on the PR head. Live GitHub mergeability now reports `DIRTY`, though, so I would keep this out of merge-ready territory until the branch is re-verified against the latest `master`. The safe paths are to refresh the branch and wait for official CI, run an exact merge-result smoke for this CLI/i18n surface, or make an explicit stale-risk acceptance if maintainers decide the scoped change is low enough risk.

### 🟢 What looks good — scoped fix after the rebase

The current diff stays focused on the six status strings and the English Fluent catalog entries. The added keys match the existing status catalogue style, and the named-variable extraction keeps the `ta()` calls readable without widening the PR into unrelated status-output cleanup.

Because @WareWolf-MoonWall's `CHANGES_REQUESTED` review is still active, I am posting this as a re-review comment rather than an approval or merge-readiness statement.

singlerider (changes requested):
I pulled the real PR head (`6ebb5438`) and compared it against current `master`. This PR is superseded by already-merged work, so it should be closed rather than landed.

### 🔴 Blocking — this change has already landed via #7499

PR #7499 ("fix(cli): route remaining status output through Fluent", merge commit `77b536f8`) **merged on 2026-06-17** and already routes the same six status lines through the CLI Fluent layer on master. The six keys this PR adds are already defined in `crates/zeroclaw-runtime/locales/en/cli.ftl` on master, and `src/main.rs` on master already calls `ta()` for all of them:

```
$ git show upstream/master:src/main.rs | grep -c 'cli-status-(trace-storage|heartbeat|memory|allowed-roots|allowed-commands|cost-tracking)'
7
```

Because the work is already in master, this PR now collides rather than applies. The two implementations even disagree on variable names — this PR uses `{$persistence}` / `{$status}`, master's #7499 uses `{$mode}` / `{$v}` for the same keys:

```
# this PR
cli-status-trace-storage = 🧾 Trace storage:  {$persistence} ({$path})
cli-status-heartbeat = 💓 Heartbeat:      {$status}
# master (#7499)
cli-status-trace-storage = 🧾 Trace storage:  {$mode} ({$path})
cli-status-heartbeat = 💓 Heartbeat:      {$v}
```

Landing this on top of master would duplicate the six key definitions and re-touch the same status block in `main.rs`, which is why the branch no longer applies cleanly.

@chengzhichao-xydt thanks for taking this on — the underlying goal (issue #7099, getting bare status `println!` strings off hardcoded English) was the right call, and it is now fully covered on master by #7499. Unless there is a status line #7499 missed that this PR uniquely fixes, the cleanest path is to close this as superseded. If you do find a gap #7499 left, please rebase onto current master and scope this down to only that delta, reusing the existing `cli-status-*` keys and their master variable names rather than redefining them.

WareWolf's original missing-key blocker is moot under this finding: the keys exist on master, but via #7499, not this branch.

Issue #5686, opened 2026-04-13
[Feature]: 添加QQ消息命令，文档未提及，如图


7568168:
### Summary

0

### Problem statement

<img width="733" height="694" alt="Image" src="https://github.com/user-attachments/assets/accf4cfc-aaf8-4ce8-a0ae-c9e62653974d" />

### Proposed solution

0

### Breaking change?

No

### Data hygiene checks

- [ ] I removed personal/sensitive data from examples, payloads, and logs.
- [ ] I used neutral, project-focused wording and placeholders.

Audacity88:
## Triage Summary

**Type:** Enhancement / docs request
**Priority:** Medium
**Area:** channel:qq
**Risk:** Low

**Summary:** This appears to request adding or documenting QQ message commands, but the issue body currently contains only a screenshot and placeholder text.

**Duplicates/Related:** Related to QQ/channel command work, but there is not enough detail to determine whether this duplicates an existing QQ command issue.

**Actionability:** Needs author action. Please add the exact command names, expected behavior, current behavior, ZeroClaw version, and whether the request is for a code change, documentation, or both.

**Repro quality:** Insufficient. The screenshot alone is not enough to triage or implement safely.

---

中文补充：这个问题目前只有截图和占位内容。请补充具体的 QQ 命令名称、期望行为、当前行为、ZeroClaw 版本，以及这是功能请求、文档请求，还是两者都需要。

Audacity88:
## Author-action deadline

This is still blocked on the missing details from the April 30 triage comment.

Please add the exact QQ command names, expected behavior, current behavior, ZeroClaw version, and whether this needs a code change, docs change, or both by Tuesday, June 16, 2026. If those details are not added by then, this issue should be closed as not actionable. It can be reopened later if the missing details are supplied.

中文补充：这个 issue 仍然缺少 4 月 30 日 triage comment 里列出的信息。请在 2026 年 6 月 16 日（星期二）前补充具体 QQ 命令名称、期望行为、当前行为、ZeroClaw 版本，以及这是代码修改、文档修改还是两者都需要。到期后如果仍然没有这些信息，这个 issue 应该作为暂不可执行关闭；之后补充信息可以再重新打开。

Audacity88:
## Status Update

This is still missing the details requested in the April 30 triage comment and the June 8 author-action deadline.

[...]

PR #7648, opened 2026-06-14
fix(channels): dedupe QQ voice redeliveries

### LINKED ISSUE #5662: QQ channel: voice messages are processed multiple times causing duplicate entries in brain.db
## Problem

When a user sends a single voice message in a QQ channel, ZeroClaw processes it **multiple times** (20+ times in our case), creating dozens of duplicate entries in `brain.db`. Each duplicate has a different `msg_id` but the same voice transcription content.

## Environment

- ZeroClaw version: v0.6.9 (pinned in memora-console)
- Platform: macOS (darwin)
- Channel: QQ (Private Channel)

## Evidence

**brain.db data** (from the affected workspace):
- Total `conversation` records: 112
- Unique content variants: 3
- Same voice transcription ("哈喽哈喽哈喽，你在干嘛？") stored **38 times** with different msg_ids

```sql
-- Records grouped by content (voice transcription):
-- Content: "<VOICE_TRANSCRIPTION>哈喽哈喽哈喽，你在干嘛？</VOICE_TRANSCRIPTION>"  → 38 records
-- Content: "<VOICE_TRANSCRIPTION>哈喽哈喽哈喽哈喽，你在干嘛？你在干嘛？</VOICE_TRANSCRIPTION>" → 38 records  
-- Content: "<VOICE_TRANSCRIPTION>你在干嘛呀？你在干嘛？</VOICE_TRANSCRIPTION>" → 36 records
```

**daemon_state.json** shows QQ channel restarted 5 times during the incident:
```json
"channel:qq": {
  "restart_count": 5,
  "status": "ok"
}
```

## Root Cause Analysis

In `src/channels/qq.rs`, the `last_sequence` is only persisted **after the WebSocket loop exits**, not in real-time per message:

```rust
// Inside the message loop - sequence is updated locally only
if let Some(s) = event.get("s").and_then(serde_json::Value::as_i64) {
    sequence = s;  // Local variable, not persisted
}

// Only persisted AFTER the loop terminates
*self.last_sequence.write().await = if sequence >= 0 { Some(sequence) } else { None };
```

When the QQ WebSocket connection drops and reconnects, `last_sequence` may point to an **older sequence number**. The subsequent RESUME (op 6) then re-delivers messages that were already processed, but with **new msg_ids** assigned by QQ during re-delivery. This causes `is_duplicate()` to fail since it only checks msg_id:

```rust
async fn is_duplicate(&self, msg_id: &str) -> bool {
    // Only checks msg_id - if QQ re-delivers with a new msg_id, this passes through
    let mut dedup = self.dedup.write().await;
    if dedup.contains(msg_id) { return true; }
    // ...
}
```

Additionally, QQ voice messages may arrive as multiple fragments/events during reconnection, each with a different `msg_id` but the same audio content.

### Fix 1 — Persist sequence number in real-time (most important)

Move the `last_sequence` persistence inside the message loop so every processed message updates it immediately:

```rust
// Inside the loop, after processing each dispatch event:
if let Some(s) = event.get("s").and_then(serde_json::Value::as_i64) {
    sequence = s;
    *self.last_sequence.write().await = Some(sequence);  // Persist immediately
}
```

### Fix 2 — Content-based deduplication for voice messages

For voice messages, also deduplicate based on the actual transcription content (excluding the file path which differs between re-deliveries). Consider adding a hash of the `<VOICE_TRANSCRIPTION>...` block to the dedup set alongside the msg_id.

### Fix 3 — Increase debounce_ms default for QQ

The `channels_config.debounce_ms` of 1000ms may be insufficient during QQ reconnect storms. Consider increasing the default or making it configurable per-channel.

## Impact

Each duplicate triggers a separate AI agent run, causing:
1. Flood of duplicate entries in brain.db
2. Multiple redundant AI responses
3. Rapid exhaustion of rate limits (QQ has a 4-per-msg_id-per-hour passive reply limit)

## Reproduction

1. Send a voice message in QQ channel
2. Intentionally trigger a reconnection (or wait for QQ server-initiated reconnect)
3. Observe multiple `C2C_MESSAGE_CREATE` events for the same voice with different msg_ids
4. brain.db accumulates dozens of duplicate conversation records

---

**Session JSONL** (234 lines, shows 116 user messages from a single voice):
- File: `workspace/sessions/qq_user_*.jsonl`
- Contains repeated "哈喽哈喽哈喽，你在干嘛？" with different timestamps and wav file paths

Audacity88:
## Summary

- **What changed and why:**
  - Persist QQ gateway dispatch sequence values as frames arrive so resume state advances beyond the handshake path.
  - Keep the existing QQ message-ID dedup path, but add a second short-window voice redelivery key for QQ voice attachments that arrive again with a new message ID.
  - Build that voice redelivery key from structured QQ attachment data: event type, reply target, sender, normalized voice source, and normalized transcript.
  - Drop duplicate platform-ASR voice redeliveries before attachment download or transcription side effects, while keeping same-transcript messages from distinct audio sources separate.
- **Scope boundary:** This PR only changes QQ receive-path deduplication and sequence persistence. It does not change outbound QQ sending, the shared media pipeline, or the separate QQ command/docs issue in #5686.
- **Blast radius:** QQ inbound message handling, QQ gateway resume behavior, and QQ voice transcription dispatch. Other channels are not touched.
- **Linked issue(s):** Closes #5662. Related #5686.
- **Labels:** `bug`, `risk: medium`, `size: M`, `channel`, `channel:qq`.

  ```text
  git diff --check
  Result: passed.

  cargo fmt --all -- --check
  Result: passed.

  cargo test -p zeroclaw-channels --features channel-qq voice_redelivery --lib
  Result: passed. 2 passed, 0 failed, 793 filtered.

  cargo test -p zeroclaw-channels --features channel-qq gateway_sequence_is_persisted_as_dispatch_frames_arrive --lib
  Result: passed. 1 passed, 0 failed, 794 filtered.

  cargo clippy -p zeroclaw-channels --features channel-qq --lib -- -D warnings
  Result: passed.
  ```
- **Beyond CI — what did you manually verify?** Confirmed the public PR diff currently changes one file, `crates/zeroclaw-channels/src/qq.rs`.
- **If any command was intentionally skipped, why:** Full-workspace clippy and CI-equivalent nextest were not run locally for this submission; GitHub CI should cover those broader gates before reviewer sign-off. Focused `zeroclaw-channels` clippy was run locally with the QQ feature enabled to cover the lint surface from the initial GitHub Lint failure.

- If any `Yes`, describe the risk and mitigation: No new security or privacy-sensitive data paths are introduced. The change reduces repeated processing of the same QQ voice payload and does not persist new user data beyond the existing in-memory dedup set.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: Existing QQ configurations continue to work without migration. The only behavior change is that repeated delivery of the same QQ voice attachment/transcript within the short dedup window is dropped instead of dispatched again.

- **Fast rollback command/path:** `git revert <merge commit for this PR>`
- **Observable failure symptoms:** QQ voice messages fail to dispatch after reconnect, legitimate repeated voice notes from distinct audio sources are dropped, or gateway reconnect logs show repeated resume/replay behavior without expected message delivery.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR is used.


singlerider (approved):
I pulled the branch, checked the diff against local source, and read #5662 end to end. This fix is well-targeted at both root causes the issue documents.

### 🟢 Real-time sequence persistence closes the redelivery window

`record_gateway_sequence` now writes `last_sequence` per dispatch frame (`s >= 0`) instead of only after the WebSocket loop exits. That directly addresses the #5662 root cause where a drop/RESUME replayed already-processed frames with fresh msg_ids. The old post-loop write at the end of the loop is still present as a backstop, so a clean exit path is unchanged. `gateway_sequence_is_persisted_as_dispatch_frames_arrive` locks the new behavior.

### 🟢 Content-hash voice dedup is correctly scoped and disjoint

The redelivery dedup hashes `event_type \0 reply_target \0 sender` plus each `(source_id, transcript)` part, bucketed on a 5-minute window. Two things I checked specifically:

- The two-phase check (platform `asr_refer_text` parts, then post-transcription fallback parts filtered by `!platform.contains(part)`) cannot double-record a single message: the part-sets are disjoint, so the set-sensitive SHA-256 produces distinct keys, and an empty fallback set returns early via the `parts.is_empty()` guard. No false self-dedup.
- `normalized_voice_source_id` strips query/fragment before hashing, so the same audio redelivered with a rotated `?sig=` is recognized as identical. That matches the issue evidence exactly (same transcript, different signed URLs).

### 🟢 Eviction refactor is behavior-preserving

`evict_dedup_if_full` is extracted and shared by both `is_duplicate` and the new `record_dedup_key`; the half-capacity eviction at `DEDUP_CAPACITY` is identical to the prior inline logic. No semantic change to the existing msg_id path.

### 🟢 Validation

`cargo test -p zeroclaw-channels --lib` is green on the branch (744 passed, 0 failed) after a clean build. Diff reads cleanly against current master; no drift in the surrounding receive path beyond the compose-message rename (`compose_message_content` → `compose_qq_message` returning `QQComposedMessage`), which is mechanical and consistent at both call sites.

Approving.

singlerider:

Issue #5831, opened 2026-04-17
[Bug]: show_tool_calls = true not effective


databillm:
### Affected component

runtime/daemon

### Current behavior

`show_tool_calls = true` under `[channel]` is not effective due to hard coded instruction in system prompt to hide the tools call details. This is contradictory and confuses the model (as could be seen in the thinking traces).

### Expected behavior

Instructions to remove tool call details from the final response should be removed from System prompt and should depend on the setting `show_tool_calls = true`

### Steps to reproduce

```bash
1. update SOUL.md and tell ZC to share tool call details
2. set show_tool_calls = true
3. now try a prompt that would trigger a tool call
4. if it's a thinking model, you could see its confusion within thinking tags
```

### Logs / stack traces

```text

```

### ZeroClaw version

0.7.0

### Operating system

24.04

### Regression?

Unknown

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

singlerider:
This should probably by rearchitected to bypass the system prompt. I expect to try to get this to be appended at the end of a message like my HMAC tool receipts PR. LLMs don't touch it there.

PR #7580, opened 2026-06-13
fix(config): warn on extra-nested provider alias tables

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Detect extra child tables under known model-provider aliases when the nested table contains provider-shaped fields.
  - Emit a warning that names the ignored section path and tells the operator to move fields back under `[providers.models.<family>.<alias>]`.
  - Preserve valid nested table-valued provider fields such as `extra_headers`, `provider_extra`, `chat_template_kwargs`, and `pricing`.
  - Add regression coverage for Zai, Azure OpenAI, Ollama, dotted aliases, unknown families, and invalid TOML shapes.
- **Scope boundary:** This PR adds a diagnostic for a malformed TOML shape. It does not reject the config, rewrite config files, change valid provider syntax, or alter provider factory/runtime selection.
- **Blast radius:** Config loading diagnostics for model-provider tables. Valid provider configs should continue to load the same way.
- **Linked issue(s):** Closes #7577. Related #7216. Related #7488.
- **Labels:** `bug`, `risk: medium`, `size: S`, `config`, `provider`, `provider:kilo`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
cargo fmt -p zeroclaw-config -- --check
result: passed

cargo test -p zeroclaw-config extra_nested_model_provider_tables
result: passed
lib tests: 1 passed, 828 filtered
migration tests: 90 filtered

git diff --check
result: passed

git cherry -v origin/master HEAD
+ bd7191691b3652098365ec5c5a1c042f1818316e fix(config): warn on extra-nested provider alias tables

git diff --stat origin/master...HEAD
crates/zeroclaw-config/src/schema.rs | 245 ++++++++++++++++++++++++++++++++++-
1 file changed, 244 insertions(+), 1 deletion(-)
```

- **Beyond CI — what did you manually verify?** Verified the previously silent extra-nested Zai shape is detected while `Config::validate()` still demonstrates the old acceptance path. Verified a valid direct provider alias table is not flagged, valid table-valued provider fields are not flagged, family-specific nested fields are flagged, dotted aliases are handled without splitting the alias incorrectly, unknown families stay owned by the existing unknown-family detector, and invalid/hostile raw TOML shapes are ignored by the diagnostic helper.
[...]

PR #7651, opened 2026-06-14
fix(config): expose global channel settings in dashboard

### LINKED ISSUE #6856: [Bug]: show_tool_calls is missing from [channel]
### Affected component

channel

### Severity

S2 - degraded behavior

### Current behavior

Tools call details are not shown in the response to channel in schema v3

### Expected behavior

There should be an option in schema v3 just like there used to be in v2, like following

`
[channels]
show_tool_calls = true
`

### Steps to reproduce

```bash
1. configure a channel
2. ask a question that will invoke a tool
3. agent response will not show the little tool call icon and tool call details like it used to show up until 0.7.5
```

### Impact

less transparency 

### Logs / stack traces

```text

```

### ZeroClaw version

0.8.0-beta1

### Rust version

_No response_

### Operating system

ubuntu 26.04

### Regression?

Unknown

### Pre-flight checks

Audacity88:
## Summary

- **What changed and why:**
  - Adds a `Global settings` tab to `/config/channels` so direct `[channels]` fields such as `show_tool_calls` are discoverable beside the existing channel-type picker.
  - Filters that tab to direct `channels.*` properties, keeping per-alias fields such as `channels.discord.default.*` out of the global editor.
  - Adds an opt-in `FieldForm` action scope so the global tab saves, counts, and clears only the fields it renders; existing tabbed forms keep the old full-prefix save behavior.
  - Updates the channel section help text and adds regression coverage for direct `[channels]` prop discovery.
- **Scope boundary:** This does not add per-channel `show_tool_calls` overrides, change runtime tool-call rendering, or address the older runtime/prompt behavior tracked separately by #5831.
- **Blast radius:** The changed surface is the config dashboard channel settings page, the shared `FieldForm` opt-in path, and `zeroclaw-config` schema metadata/tests. Existing `FieldForm` callers retain the default full-prefix action behavior unless they opt in.
- **Linked issue(s):** Closes #6856.
- **Labels:** `bug`, `risk: medium`, `size: S`, `channel`, `config`, `web`.

```text
cargo fmt --all -- --check
Result: passed.

cargo test -p zeroclaw-config channels_root_settings_stay_on_direct_prop_surface
Result: passed.
Test result: 1 passed; 0 failed; 828 filtered out.

npm run build
Result: passed.
Note: Vite emitted the existing large chunk-size warnings.

git diff --check
Result: passed.

Browser smoke: /config/channels global settings tab
Result: passed.
Verified the real Vite/React route with a gateway-shaped mock: `Global settings` activates `/config/channels?tab=global`, `Global channel settings` renders, `channels.show_tool_calls` is visible, alias-scoped `channels.matrix.default.enabled` is filtered out, and no browser console errors or error boundary appeared.
```

- **Beyond CI - what did you manually verify?** Issue #6856 was refreshed before drafting; the reporter confirmed manual `[channels] show_tool_calls = true` works, so this PR keeps the canonical global setting and fixes the discoverability path. I also reviewed the hidden-save boundary in `FieldForm` and kept the narrowed save/dirty/discard behavior opt-in for the new global tab only.
- **If any command was intentionally skipped, why:** Workspace clippy and CI-shaped nextest were not run locally. The local checks cover the changed config schema regression, the web production build, and a browser smoke of the changed `/config/channels` dashboard path; GitHub CI should provide workspace-wide clippy/test coverage before merge.

## Compatibility (required)

- If `No` or `Yes` to either: This exposes an existing config key in the web config surface; it does not change TOML, env, or CLI semantics. Existing users can keep using:

```toml
[channels]
show_tool_calls = true
```

- **Feature flags or config toggles:** None. Operators can still edit `[channels] show_tool_calls` manually if the dashboard path regresses.
- **Observable failure symptoms:** `/config/channels` fails to load the global settings tab, direct `[channels]` fields are missing from the tab, or saving from the global tab unexpectedly affects per-channel alias drafts.

- `Co-authored-by` trailers added in commit messages for incorporated contributors? (`Yes/No`): N/A

### DIFF
--- a/crates/zeroclaw-config/src/schema.rs
+++ b/crates/zeroclaw-config/src/schema.rs
@@ -27550,6 +27550,53 @@ allowed_users = []
         }
     }
 
+    /// The dashboard's `/config/channels` global-settings tab filters the
+    /// canonical `prop_fields()` list down to direct `[channels]` fields.
+    /// Keep root settings such as `show_tool_calls` discoverable there without
+    /// mixing per-alias fields into the same editor.
+    #[test]
+    async fn channels_root_settings_stay_on_direct_prop_surface() {
+        let mut config = Config::default();
+        config.init_defaults(None);
+        config
+            .channels
+            .matrix
+            .insert("default".into(), MatrixConfig::default());
+
+        let paths: Vec<_> = config
+            .prop_fields()
+            .into_iter()
+            .map(|field| field.name)
+            .collect();
+        let direct_channel_paths: Vec<_> = paths
+            .iter()
+            .filter_map(|path| {
+                path.strip_prefix("channels.")
+                    .filter(|rest| !rest.contains('.'))
+                    .map(|_| path.as_str())
+            })
+            .collect();
+
+        assert!(
+            direct_channel_paths.contains(&"channels.show_tool_calls"),
+            "root [channels] settings should include show_tool_calls: {direct_channel_paths:?}"
+        );
+        assert!(
+            direct_channel_paths.contains(&"channels.ack_reactions"),
+            "root [channels] settings should include other global channel controls"
+        );
+        assert!(
+            paths
+                .iter()
+                .any(|path| path == "channels.matrix.default.enabled"),
+            "fixture should include a nested channel alias field"
+        );
+        assert!(
+            !direct_channel_paths.contains(&"channels.matrix.default.enabled"),
+            "global channel settings must not include per-alias fields"
+        );
+    }
+
     /// Audit gate for RFC #6971 Phase 0: any credential-shaped property path
     /// that reaches the CLI/gateway/TUI property surface must have an explicit
     /// classification. This catches future config additions whose names imply
--- a/crates/zeroclaw-config/src/sections.rs
+++ b/crates/zeroclaw-config/src/sections.rs
@@ -449,8 +449,9 @@ sections! {
         key:   "channels",
         shape: TypedFamilyMap,
         group: Foundation,
-        help:  "Pick which chat platforms ZeroClaw should listen on. You can \
-                configure multiple — each channel gets its own alias.",
+        help:  "Pick which chat platforms ZeroClaw should listen on. Global \
+                channel settings live on [channels]; each configured platform \
+                still gets its own alias.",
     },
     Hardware => {
         key:   "hardware",
--- a/web/src/components/sections/FieldForm.tsx
+++ b/web/src/components/sections/FieldForm.tsx
@@ -141,11 +141,15 @@ interface FieldFormProps {
   drift?: DriftEntry[];
   /** Filter for which entries this form renders. Returning false hides
    *  the entry. Used to partition a section's fields across tabs (e.g.
-   *  Model providers: Connection / Model / Advanced). The form still
-   *  fetches every entry under `prefix`; the predicate only gates
-   *  rendering, so saves still validate against the full server-side
-   *  config. */
+   *  Model providers: Connection / Model / Advanced). By default, the
+   *  form still fetches every entry under `prefix`; the predicate only
+   *  gates rendering, so saves still validate against the full
+   *  server-side config. */
   includePath?: (path: string) => boolean;
+  /** When true, `includePath` also limits save, dirty-count, and
+   *  successful-save discard scope. Use this when hidden fields belong to a
+   *  different editor, not just another tab of the same editor. */
+  scopeActionsToIncludedPaths?: boolean;
   /** Render the save bar as a normal inline element instead of
    *  `sticky bottom-0`. Set when the FieldForm is embedded inside a
    *  taller composite editor (e.g. an expandable rate-sheet row) where
@@ -695,6 +699,7 @@ const FieldForm = forwardRef<FieldFormHandle, FieldFormProps>(
       title,
       drift,
       includePath,
+      scopeActionsToIncludedPaths = false,
       inlineSaveBar = false,
     },
     ref,
@@ -762,6 +767,16 @@ const FieldForm = forwardRef<FieldFormHandle, FieldFormProps>(
       // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [prefix]);
 
+    const actionableEntries = useMemo(() => {
+      if (!scopeActionsToIncludedPaths || !includePath) return entries;
+      return entries.filter((e) => includePath(e.path));
+    }, [entries, includePath, scopeActionsToIncludedPaths]);
+
+    const actionablePaths = useMemo(
+      () => actionableEntries.map((e) => e.path),
+      [actionableEntries],
+    );
+
     // Returns true when nothing was dirty or the save succeeded; false on
     // any error so callers (e.g. the wizard's Next button) can refuse to
     // advance past a broken state.
@@ -790,7 +805,7 @@ const FieldForm = forwardRef<FieldFormHandle, FieldFormProps>(
         }
         return value;
       };
-      for (const e of entries) {
+      for (const e of actionableEntries) {
         if (configDraft.tombstones.has(e.path)) {
           ops.push({ op: "remove", path: e.path });
           continue;
@@ -835,7 +850,11 @@ const FieldForm = forwardRef<FieldFormHandle, FieldFormProps>(
       try {
         const resp = await patchConfig(ops);
         setSavedAt(`${t("fieldform.saved_prefix")}${resp.results.length}${t("fieldform.saved_suffix")}`);
-        configDraft.discardSection(prefix);
+        if (scopeActionsToIncludedPaths && includePath) {
+          configDraft.discardPaths(actionablePaths);
+        } else {
+          configDraft.discardSection(prefix);
+        }
         await reload();
         onSaved?.();
         return true;
@@ -925,7 +944,7 @@ const FieldForm = forwardRef<FieldFormHandle, FieldFormProps>(
     // across the loading / loaded transition (React error #310).
     const unsavedCount = useMemo(() => {
       let n = 0;
-      for (const e of entries) {
+      for (const e of actionableEntries) {
         if (configDraft.tombstones.has(e.path)) {
           n += 1;
           continue;
@@ -938,7 +957,7 @@ const FieldForm = forwardRef<FieldFormHandle, FieldFormProps>(
         if (valueChanged || commentChanged) n += 1;
       }
       return n;
-    }, [entries, draft, comments, configDraft.tombstones]);
+    }, [actionableEntries, draft, comments, configDraft.tombstones]);
 
     // Warn user before navigating away with unsaved changes.
     useEffect(() => {
--- a/web/src/lib/draftStore.tsx
+++ b/web/src/lib/draftStore.tsx
@@ -41,6 +41,8 @@ type ConfigDraftCtx = ConfigDraftState & {
   clearComment: (path: string) => void;
   stageTombstone: (path: string) => void;
   unstageTombstone: (path: string) => void;
+  /** Drop drafts / comments / tombstones for exactly these paths. */
+  discardPaths: (paths: readonly string[]) => void;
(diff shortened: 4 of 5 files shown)


WareWolf-MoonWall (commented):
The feature scope is right — exposing global `[channels]` settings through the Config dashboard is a legitimate gap fix, and the SSOT analysis of `handle_resolve_alias_source` and the tab rendering is clean. However, the branch can't be reviewed in its current state.

### 🔴 Blocking — Branch is dirty (conflicted with master)

`mergeable: false / dirty`. The branch can't land as-is and must be rebased before merge.

### 🔴 Blocking — 63-commit merge history makes the diff unreviewable

The branch contains ~62 master carry-in commits on top of the one feature commit. `gh pr diff` returns HTTP 406 (diff exceeds 20,000 lines). The declared `size: S` label is incorrect — the raw stat is +17,409 / -3,698 due to the carry-ins. Reviewing this diff is not feasible because the feature delta is indistinguishable from already-landed master changes without per-commit attribution.

The fix: rebase onto the current master HEAD (resolve any actual conflicts in the feature commit itself), so the diff reflects only the ~200 lines of actual feature work. That will make the PR reviewable, fix the size label, and resolve the dirty state in one operation.

### 🟢 What looks good — SSOT pattern in `handle_resolve_alias_source`

```rust
let cfg = state.config.read().clone();
let values = cfg.resolve_alias_source(q.source);
```

Reading from `Arc<RwLock<Config>>` and materializing a local owned value per-request is the correct pattern — not a snapshot cache, not a stored duplicate. The resolved values are used and dropped within the request handler. ✅

### 🟢 What looks good — `AliasSource` enum design

Single definition in `traits.rs`; `api_config.rs` derives the wire string from it. One source, no duplication.

### 🔵 Suggestion — `Config::clone()` on every alias-resolve call

`handle_resolve_alias_source` clones the full `Config` struct to call one method. Depending on `Config`'s size, calling `resolve_alias_source` directly on the read guard (without cloning) would be cheaper. Not a SSOT issue, but worth considering if the config struct is large.

singlerider (approved):
Re-review on the rebased head. @WareWolf-MoonWall's two blockers were the dirty/conflicted branch and the 63-commit carry-in history that made the diff unreviewable. Both are gone: the PR is now a single commit, mergeable, +168/-17 across 5 files, and the `size: S` label is now accurate.

### ✅ Resolved - branch dirty / conflicted

`mergeable: MERGEABLE` now. The earlier `dirty` state is cleared.

### ✅ Resolved - unreviewable 63-commit history

Down to one feature commit. The diff now reflects only the actual feature delta, and the `Config::clone()` carry-in that prompted @WareWolf-MoonWall's 🔵 suggestion is no longer in the diff at all, so that point is moot on this head.

### 🟢 What looks good - opt-in action scoping is correctly default-off

`scopeActionsToIncludedPaths` defaults to `false`, so every existing `FieldForm` caller keeps full-prefix save/dirty/discard behavior. Only the new global tab opts in. The save loop, `unsavedCount`, and the post-save discard all switch on the same flag, so render scope and action scope stay consistent. No silent behavior change for current callers.

### 🟢 What looks good - `discardPaths` is a clean primitive

`discardPaths` mirrors `discardSection` exactly (drafts, comments, tombstones) but keys on an explicit set, and each setter returns the previous reference when nothing changed so it does not churn React state. The empty-set early return is the right guard.

### 🟢 What looks good - no hardcoded field list

The global tab filters the canonical `prop_fields()` through `isDirectChannelSetting` rather than enumerating channel keys, and the new regression test asserts both inclusion (`channels.show_tool_calls`) and exclusion (`channels.matrix.default.enabled`). That keeps the registry as the single source.

tidux (commented):
Walked the PR-as-displayed against current upstream/master and then walked the underlying feature commit (7d087179a) against the local tree separately so the actual change can still be reviewed on the merits. The two are not the same view.

### 🔴 Blocking — 62 master carry-in commits make the displayed diff unreviewable

@WareWolf-MoonWall already raised this; I want to ground-truth the shape so the rebase target is precise.

The PR has 63 commits. Exactly one of them is the feature (7d087179a, "fix(config): expose global channel settings in dashboard", 5 files, +168/-17). The other 62 are two `Merge origin/master into ...` commits (008122356 on 2026-06-15, 1acf504604 on 2026-06-16) plus everything master landed since the branch was cut. The displayed +17,409/-3,698 across 170 files is dominated by those master carry-ins:

- `crates/zeroclaw-channels/src/discord.rs` +2462/-128 — from #7525, #7489, #7490, #7526
- `crates/zeroclaw-config/src/schema.rs` +1260/-240 — from #7556, #7594, #7580, #7526, #5797, others
- `xtask/src/generate/spec.rs` +788/-0 — new file from #7558
- `crates/zeroclaw-channels/src/slack.rs` +629/-12 — from #6428
- `Cargo.lock` +590/-76 — incidental
- `web/package-lock.json` +543/-1149 — from #7569
- `web/src/pages/Skills.tsx` +182/-0 — new file from #6700
- ...and 90+ more files from already-merged PRs

The declared `size: S` label refers to the feature commit (which would correctly be S), but the displayed diff is well above `size: XXL`. CODEOWNERS routing fires on the displayed-diff scope rather than the feature scope, which is why this PR is requesting a half-dozen reviewers across channels, providers, runtime, gateway, ci, and web — none of which the feature actually touches.

The fix is mechanical: `git rebase upstream/master` on the feature commit alone (it's the only commit you own on the branch). I dry-ran the rebase via `git merge-tree --merge-base=7d087179a~1 upstream/master 7d087179a` and got two real content conflicts, both tractable:

- `crates/zeroclaw-config/src/sections.rs` — your `help:` text edit for the `Channels` section overlaps with the recent grouping work from #7556. The conflict is in the same help string; you keep your new wording and add the `#[group]` annotation the section now needs.
- `web/src/components/sections/FieldForm.tsx` — your new `scopeActionsToIncludedPaths` prop + the `actionableEntries`/`actionablePaths` plumbing overlaps with the Operator Console redesign in #7665, which rewrote 637 lines of this file. The behaviour you're adding is additive and orthogonal to that redesign, so the conflict is structural (where in the rewritten file the new prop and the new useMemos land) rather than logical.

`crates/zeroclaw-config/src/schema.rs`, `web/src/lib/draftStore.tsx`, and `web/src/pages/Config.tsx` auto-merge cleanly per the same dry-run.

### 🟡 Warning — @WareWolf-MoonWall's commendations cite code that isn't this PR's contribution

Calling this out for the record, not as a fault on Wolf's part — it's exactly the failure mode the unreviewable diff creates.

Wolf's review praises `handle_resolve_alias_source`, the `Config::clone()` cost question on that function, and the `AliasSource` enum design. None of those are in 7d087179a. They live in `crates/zeroclaw-gateway/src/api_config.rs:952`, `crates/zeroclaw-config/src/traits.rs:20`, and `crates/zeroclaw-config/src/schema.rs:3451`, and they all entered this branch via the merge of master commit d36cce3ea (#7594, `feat(config): type-driven alias-ref pickers and self-declaring config enums`). That work was already reviewed and approved in its own PR.

Two consequences: those 🟢s and the 🔵 should not be treated as carrying weight for #7651's verdict (they're praising someone else's already-merged work), and the same hazard waits for every reviewer who tries to walk the displayed diff without first isolating the feature commit. The rebase is the fix.

### 🔴 Blocking — Branch is dirty against current master

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY` at 1acf504604. The same rebase that closes the carry-in problem closes this. After the rebase, the only content conflicts to resolve are the two I sketched above; everything else in the displayed diff disappears.

### 🟢 What looks good — Feature commit is tight and the SSOT shape is right

Walking 7d087179a on its own (not the displayed diff): the change is exactly the five files the description says it is, and the shape is correct.

`includePath` was already the filter the form used to gate rendering; adding `scopeActionsToIncludedPaths` as an opt-in second meaning ("also gate save/dirty-count/discard scope") keeps the default behaviour byte-identical for every existing caller. The new `discardPaths` on `ConfigDraftCtx` is a clean parallel to `discardSection` — same shape, different selector — and it correctly bumps the three relevant pieces of state (`drafts`, `comments`, `tombstones`) only when something actually changes (`changed` flag pattern on each), so unrelated draft state stays stable.

`isDirectChannelSetting` is the right kind of helper — single-call, defined where it's used, exact semantic ("strip the `channels.` prefix, refuse anything still containing a dot"). Good fit for the intent.

The drift indicator and `onSaved` plumbing both flow through the new tab structure, so unsaved-change and post-save reload behaviour stay consistent with the `Channel types` tab.

### 🟡 Warning — Regression test couples the feature to an unrelated concurrent master change

`channels_root_settings_stay_on_direct_prop_surface` asserts that both `channels.show_tool_calls` and `channels.ack_reactions` appear in `direct_channel_paths`. The first is the load-bearing assertion for #6856 — that field is the whole reason the PR exists. The second is a coupling to `pub ack_reactions: bool` on `[channels]`, which is itself brand-new (it landed in master via 14d4c1b21 / #7526, only days before this PR was authored, and arrived in this branch via the same merge that produced the unreviewable diff).

If a future PR removes, renames, or promotes `ack_reactions` (e.g. into a typed enum or a sub-section), this test breaks even though nothing about the #6856 contract has changed. The assertion is also weaker than it looks: the test doesn't pin "every direct field is rendered", only that two specific ones are, so a future regression that drops `show_tool_calls` from the direct surface would still be caught by the first `assert!`. The `ack_reactions` line is dead weight from a coverage standpoint and active liability from a maintenance standpoint.

Suggest dropping the `ack_reactions` assertion. If you want to backstop the partition shape ("global tab really doesn't include alias fields"), the existing `!direct_channel_paths.contains(&"channels.matrix.default.enabled")` assertion already does that — it's the structural assertion that survives any field rename.

### 🔵 Suggestion — `FieldForm prefix="channels"` fetches every alias field too

When the global tab mounts, the form requests every entry under `channels.` and then filters render + actions via `isDirectChannelSetting`. So the wire fetch carries `channels.discord.default.token`, `channels.matrix.default.encryption.allowed_rooms`, every WhatsApp option, etc., all of which the global tab will never render or touch.

Today the cost is "a chunky GET on a tab the operator visits rarely" — fine, not a blocker. If/when there's a server-side `depth=1` (or equivalent shape-filter) on the prefix listing, this is the natural caller to switch over because the predicate boundary is already a single sharp helper (`isDirectChannelSetting`).

Pure follow-up. Don't take this in this PR.

### 🟢 What looks good — Help-text edit lands the discoverability in the right place

The `sections.rs` Channels `help:` change ("Global channel settings live on `[channels]`; each configured platform still gets its own alias") is the right two-sentence reframe of what the section now contains. It's also the kind of edit a TUI/dashboard reader actually reads at the moment they're looking for `show_tool_calls`, so it shortens the discovery path the issue describes. Small but well-placed.

---

Validation on the PR head 1acf504604 (the as-displayed merge, not the rebase target): `cargo fmt --all -- --check` clean; `cargo check -p zeroclaw-config` finishes; `cargo test -p zeroclaw-config channels_root_settings_stay_on_direct_prop_surface` 1 passed (the `#[test] async fn ...` shape resolves to `tokio::test` via the in-module `use tokio::test;` shadow at schema.rs:18735 — confirmed, this is the existing pattern in the test module); `npm run build` under `web/` passes with the pre-existing Vite chunk-size warnings the PR body already calls out.

I did *not* attempt the rebase locally and re-validate against current master — that is the author's pass to run. The two conflict sketches above are what the dry-run produced; the actual conflict resolution belongs in commits on the feature branch.

Marking `--comment`. The 🔴s are the same blocks @WareWolf-MoonWall named; my contribution is ground-truthing the conflict shape, isolating the actual feature commit so its merits can stand separately from the carry-in noise, and flagging the misattribution risk created by the displayed diff. Once the branch is rebased to a single feature commit on top of current master and the regression test drops the `ack_reactions` coupling, I'm comfortable converting `--comment` → `--approve` on the next pass. @JordanTheJet, @Nillth, and @singlerider are also still pending review.

PR #7660, opened 2026-06-14
fix(runtime): redact unquoted base64 credentials

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Expands the runtime credential scrubber's unquoted key/value matcher so base64-style credential values containing `/`, `+`, and `=` are redacted as one credential value.
  - Adds regression coverage for unquoted base64-style credential values with adjacent public output, so the scrubber redacts the secret while preserving the next public key/value pair.
  - Adds quoted base64-style coverage to keep the existing quoted-value behavior pinned.
  - Updates the background skill-review fork's `run_tool_call_loop` call to the current signature so `zeroclaw-runtime` tests compile on current `master`.
- **Scope boundary:** This PR does not change credential storage, provider auth resolution, log retention, or broad high-entropy secret detection. It only tightens the existing key/value credential scrubber and updates one current-signature call site.
- **Blast radius:** This changes the runtime `scrub_credentials()` helper in `crates/zeroclaw-runtime/src/agent/turn/redact.rs`. Any text passed through that helper now redacts unquoted base64-style credential values more completely. The skill-review fork change is limited to passing default/inert loop arguments required by the current function signature.
- **Linked issue(s):** None.
- **Labels:** `bug`, `risk: high`, `size: XS`, `runtime`, `agent`, `skills`

```text
cargo test -p zeroclaw-runtime scrub_credentials_redacts --lib -- --nocapture
test result: ok. 4 passed; 0 failed; 2166 filtered out.

cargo test -p zeroclaw-runtime skill_review_fork --lib -- --nocapture
test result: ok. 2 passed; 0 failed; 2168 filtered out.

cargo fmt --all -- --check
result: passed with no output.

cargo clippy -p zeroclaw-runtime --lib --tests -- -D warnings
result: passed with no warnings.

result: passed with no warnings.

Summary [  51.940s] 8605 tests run: 8605 passed, 10 skipped.

git diff --check
result: passed.
```

- **Beyond CI — what did you manually verify?** The new unquoted regression uses `token=QWxh+GRpbjpvcGVu/IHNlc2FtZQ== next=public` and verifies that the base64-style credential is redacted while `next=public` remains visible. The quoted regression covers the same value shape inside quotes.

- If any `Yes`, describe the risk and mitigation: The credential scrubber now redacts a broader unquoted credential value shape. The risk is accidental over-redaction, so the regression test includes adjacent public output and verifies that non-secret text after the credential remains visible.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No user upgrade steps are required.

- **Fast rollback command/path:** Before merge, close or abandon this PR. After merge, revert the squash merge commit with `git revert <merge-sha>`.
- **Observable failure symptoms:** Credential-shaped key/value output with base64 `/`, `+`, or `=` should redact more completely. Unexpected over-redaction would show non-secret text adjacent to credential-shaped key/value output missing from scrubbed output.

Not applicable.


WareWolf-MoonWall (approved):
Correct and targeted fix. Adding `/`, `+`, and `=` to the unquoted character class closes the gap where standard base64 credential values (RFC 4648 §4) weren't being redacted, while URL-safe base64 (`-` and `_`) was already covered. The regression tests confirm the precise boundaries. Approving.

### 🟢 What looks good — Base64 coverage analysis

The before/after coverage is explicit:
- Standard base64 (`A-Za-z0-9+/=`) was missing `+`, `/`, `=` — now covered.
- URL-safe base64 (`A-Za-z0-9-_`) was already covered by the old `\\-` and `_` entries.
- MIME base64 with line folds is intentionally out of scope (documented in PR body).

The `scrub_credentials_redacts_unquoted_base64_credential_values` test directly pins the character boundary: `token=QWxh+GRpbjpvcGVu/IHNlc2FtZQ==` becomes `token=QWxh*[REDACTED]`, with the trailing `next=public` preserved — confirming the greedy match stops at the space boundary.

### 🟢 What looks good — Over-redaction risk is bounded

The keyword gate (`token|api_key|secret|password|...`) and the 8-character minimum together limit false positives. The PR body explicitly acknowledges this tradeoff.

### 🟡 Warning — `/` addition may over-redact path-shaped values

Adding `/` to the unquoted character class means a string like `token=/home/user/some/path/longer` would now be redacted if the path is 8+ characters. The keyword gate provides the first defense (only `token=` or similar keywords trigger the pattern), but path-valued fields in configs that happen to use a credential keyword name would now be redacted where they weren't before. This is an accepted expansion of the redaction surface — worth noting in release notes so operators whose configs use path-valued credential-named fields know what to expect.

### 🔵 Note — `codex/` branch prefix

The branch name `codex/coderabbit-redaction-base64-credentials` suggests this was authored with AI tooling assistance. There's no `Co-Authored-By:` trailer in the commit message, so the commit history is clean per project policy. Noting it for the audit trail.

PR #6297, opened 2026-05-03
feat(channels): expose poll-vote / interactive-reply events; add Channel::send_choice

bglusman:
Maintainer note: I refreshed this PR body after re-review so the public record matches the final Signal `sendPollCreate` implementation and the live Signal/WhatsApp channel-boundary evidence posted in the comments. The implementation remains Brian's contribution.

## Summary

- **Base branch:** `master`
- **What changed and why:** Three additive patches that expose inbound interactive events the deserializers were silently dropping, and one trait extension that gives every channel a uniform discrete-choice API:
  - **Signal**: extend `DataMessage` to deserialize `pollAnswer` and `pollVote` shapes that were previously dropped; surface poll votes via `process_envelope` as synthetic `[choice-index]N` content for real `pollVote` payloads, with `[choice]<title>` still accepted for alternate `pollAnswer` title payloads so consumers driving `Channel::listen` get the user's selection. Add `SignalChannel::send_poll(recipient, question, options, multiple_choice)` outbound helper that uses signal-cli daemon JSON-RPC `sendPollCreate`.
  - **WhatsApp Cloud**: extend `parse_webhook_payload` to recognize `type=interactive` messages with `button_reply` / `list_reply` sub-types — previously these were silently dropped as non-text. Surface as `[choice]<id>` so consumers can correlate to the original option. Add outbound helpers `send_interactive_buttons` (up to 3 buttons per Meta's limit) and `send_interactive_list` (up to 10 sections x 10 rows). Adds `InteractiveListSection` / `InteractiveListRow` types.
  - **Channel trait**: add `async fn send_choice(&self, recipient, prompt, &[(callback_id, label)])` as a default-text-fallback method so every existing impl gets numbered `(reply with name or number)` rendering for free; `SignalChannel` and `WhatsAppChannel` override to render natively.
- **Scope boundary:** No changes to other channels' impls. No changes to `zeroclaw-runtime`, `zeroclaw-gateway`, WhatsApp Web, or any non-channel surface. The default trait method is additive — existing `Channel` implementors compile unchanged.
- **Blast radius:** `Channel` is in `zeroclaw-api` (Experimental tier per AGENTS.md, formal milestone at v1.0.0). Adding a default trait method is source-compatible. The Signal/WhatsApp inbound deserializer changes are pure field additions with `#[serde(default)]` — older payloads still parse.
- **Linked issue(s):** None.

## Validation Evidence

```text
$ cargo fmt --all -- --check
(no output — clean)

[...]

PR #4504, opened 2026-03-24
feat(channels): add room creation and user invite API to Channel trait

theredspoon:
## Summary

Add `create_room()` and `invite_user()` methods to the `Channel` trait with default no-op implementations. Implement both for `MatrixChannel` using the Matrix SDK.

### Changes

- **Modified:** `src/channels/traits.rs` — added `create_room()` and `invite_user()` trait methods with default implementations
- **Modified:** `src/channels/matrix.rs` — `MatrixChannel` implementation using `matrix_sdk` room creation and invite APIs
- **Modified:** `tests/integration/channel_matrix.rs` — new enum variants, test channel implementations, lifecycle tests, updated minimal channel contract test

### Parameters

`create_room(name, topic, invites, visibility, encryption)`:
- `name`: optional room name
- `topic`: optional room description
- `invites`: list of user IDs to invite
- `visibility`: `"private"` or `"public"`
- `encryption`: optional flag for E2EE (Megolm v1)

`invite_user(room_id, user_id)`:
- `room_id`: platform-scoped room identifier
- `user_id`: platform-scoped user identifier

### Test coverage

- Trait default tests (DummyChannel returns empty/Ok)
- MatrixTestChannel lifecycle tests (create_room, invite_user)
- Minimal channel contract test (Box<dyn Channel> dispatch)
- 8 trait tests + 53 integration tests passing

### Risk

**Medium** — channel trait boundary modification. All existing implementations continue to compile without changes (default no-op implementations).

### Breaking changes

None. New trait methods have default implementations.

### Non-goals

- No tool exposure (deferred to future PR)
- No Discord/Telegram/Slack implementations (use trait defaults)
- No advanced room settings (power levels, aliases, federation)
- No bulk invites or room membership management

PR #7661, opened 2026-06-14
feat(channels): recover Matrix room management

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds additive `Channel::create_room` and `Channel::invite_user` methods so room-management capable channels have a shared API again.
  - Implements Matrix room creation with name, topic, invite, visibility, and optional encryption request fields.
  - Implements Matrix user invites with the direct room invite endpoint so valid room IDs do not need to be present in the local SDK room cache first.
  - For unsupported channels, the default methods return explicit unsupported errors instead of silent success.
  - For paced outbound channels, forwards room-management calls to the inner channel so enabling reply pacing does not hide the new capability.
- **Scope boundary:** This PR does not add user-facing CLI tools or autonomous workflows that call these methods. It does not change #6297's separate `send_choice` trait work, and it does not implement room management for Signal, WhatsApp, or other channel backends.
- **Blast radius:** The changed surface is the experimental channel trait API, the Matrix channel implementation, and the `PacedChannel` wrapper. Existing channels remain source-compatible through default trait methods.
- **Linked issue(s):** Related #6074. Supersedes #4504.
- **Labels:** `bug`, `channel`, `channel:matrix`, `risk: medium`, `size: M`

```text
cargo fmt --all -- --check
result: passed

git diff --check
result: passed

cargo test -p zeroclaw-api default_room_management_methods_report_unsupported
result: passed; 1 passed, 0 failed, 51 filtered out

cargo test -p zeroclaw-channels room_management --features channel-matrix --lib
result: passed; 7 passed, 0 failed, 883 filtered out
```

- **Beyond CI — what did you manually verify?** Compared the PR branch against the latest fetched `origin/master`: one surviving commit touches only `crates/zeroclaw-api/src/channel.rs`, `crates/zeroclaw-channels/src/matrix.rs`, and `crates/zeroclaw-channels/src/paced_channel.rs`. Also checked live overlap with #6297; it touches the `Channel` trait for `send_choice`, but does not implement room creation or invites.
- **If any command was intentionally skipped, why:** Full workspace clippy and full local CI-style tests were not run locally. This is a medium-risk, narrow experimental channel/API recovery slice with focused API and Matrix feature validation; the visible GitHub Quality Gate has passed, including Lint, Test, platform builds, Benchmarks Compile, and CI Required Gate.

- New permissions, capabilities, or file system access scope? `Yes` — this restores a channel capability surface for room creation and user invites. It does not grant new credentials or bypass platform authorization; Matrix still enforces the authenticated user's server-side permissions.
- New external network calls? `Yes` — Matrix `create_room` and `invite_user` send Matrix client-server API requests when a caller explicitly invokes these new methods.
- If any `Yes`, describe the risk and mitigation: The main risk is exposing a shared API that a future caller could invoke on a channel that does not support it. Unsupported channels now fail closed with explicit errors, and Matrix request builders validate room and user IDs before sending.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No operator upgrade steps are required. The trait methods are additive and have default implementations; there is no new config, env var, or CLI flag.

- **Observable failure symptoms:** Matrix callers of the new room APIs may receive Matrix API errors for room creation or invites. Unsupported channels should return errors containing `channel does not support room creation` or `channel does not support room invites`.

  - #4504 by @theredspoon
- Scope materially carried forward: The shared channel room-management API shape and Matrix room creation / invite capability are carried forward, rebuilt against the current channel trait, Matrix SDK surface, and pacing wrapper.
- If `No`, why (inspiration-only, no direct code/design carry-over): N/A. A human `Co-authored-by` trailer is included for @theredspoon because the shared room-management API shape and Matrix capability from #4504 are materially carried forward, rebuilt against the current crate split and Matrix SDK surface.

### DIFF
--- a/crates/zeroclaw-api/src/channel.rs
+++ b/crates/zeroclaw-api/src/channel.rs
@@ -378,6 +378,25 @@ pub trait Channel: Send + Sync + crate::attribution::Attributable {
         Ok(())
     }
 
+    /// Create a room, channel, or conversation on the platform.
+    ///
+    /// Channels without a room-creation API report unsupported explicitly.
+    async fn create_room(
+        &self,
+        _name: Option<&str>,
+        _topic: Option<&str>,
+        _invites: &[String],
+        _visibility: Option<&str>,
+        _encryption: Option<bool>,
+    ) -> anyhow::Result<String> {
+        anyhow::bail!("channel does not support room creation")
+    }
+
+    /// Invite a user to an existing room, channel, or conversation.
+    async fn invite_user(&self, _room_id: &str, _user_id: &str) -> anyhow::Result<()> {
+        anyhow::bail!("channel does not support room invites")
+    }
+
     /// Request interactive tool-call approval from the channel operator.
     ///
     /// Returns `Ok(Some(response))` when the operator answers within the
@@ -584,6 +603,37 @@ mod tests {
         let back: ChannelApprovalResponse = serde_json::from_str(&json).unwrap();
         assert!(
             matches!(back, ChannelApprovalResponse::DenyWithEdit { replacement } if replacement == "new content")
+        );
+    }
+
+    #[tokio::test]
+    async fn default_room_management_methods_report_unsupported() {
+        let channel = StubChannel { handle: None };
+
+        let room_err = channel
+            .create_room(
+                Some("Project Room"),
+                Some("Project discussion"),
+                &["@user:example.com".to_string()],
+                Some("private"),
+                Some(true),
+            )
+            .await
+            .unwrap_err();
+
+        assert!(
+            room_err
+                .to_string()
+                .contains("does not support room creation")
+        );
+        let invite_err = channel
+            .invite_user("!room:example.com", "@other:example.com")
+            .await
+            .unwrap_err();
+        assert!(
+            invite_err
+                .to_string()
+                .contains("does not support room invites")
         );
     }
 }
--- a/crates/zeroclaw-channels/src/matrix.rs
+++ b/crates/zeroclaw-channels/src/matrix.rs
@@ -1,4 +1,4 @@
-//! Matrix channel using matrix-rust-sdk 0.16.
+//! Matrix channel using matrix-rust-sdk 0.17.
 //!
 //! Organisation (single file, internal `mod` blocks):
 //! - `markers`: parse `[image:...] [voice:...]` etc. from outbound text
@@ -3461,6 +3461,75 @@ impl MatrixChannel {
     }
 }
 
+fn build_create_room_request(
+    name: Option<&str>,
+    topic: Option<&str>,
+    invites: &[String],
+    visibility: Option<&str>,
+    encryption: Option<bool>,
+) -> Result<matrix_sdk::ruma::api::client::room::create_room::v3::Request> {
+    use matrix_sdk::ruma::{
+        OwnedUserId,
+        api::client::room::{Visibility, create_room::v3::Request as CreateRoomRequest},
+        events::{InitialStateEvent, room::encryption::RoomEncryptionEventContent},
+    };
+
+    let mut request = CreateRoomRequest::new();
+    request.name = name.map(str::to_owned);
+    request.topic = topic.map(str::to_owned);
+    request.invite = invites
+        .iter()
+        .map(|user_id| {
+            user_id
+                .parse::<OwnedUserId>()
+                .with_context(|| format!("Invalid Matrix user ID: {user_id}"))
+        })
+        .collect::<Result<Vec<_>>>()?;
+
+    if let Some(visibility) = visibility {
+        request.visibility = match visibility {
+            "private" => Visibility::Private,
+            "public" => Visibility::Public,
+            other => bail!("Invalid visibility: expected `private` or `public`, got `{other}`"),
+        };
+    }
+
+    if encryption.unwrap_or(false) {
+        request.initial_state.push(
+            InitialStateEvent::with_empty_state_key(
+                RoomEncryptionEventContent::with_recommended_defaults(),
+            )
+            .to_raw_any(),
+        );
+    }
+
+    Ok(request)
+}
+
+fn build_invite_user_request(
+    room_id: &str,
+    user_id: &str,
+) -> Result<matrix_sdk::ruma::api::client::membership::invite_user::v3::Request> {
+    use matrix_sdk::ruma::{
+        OwnedRoomId, OwnedUserId,
+        api::client::membership::invite_user::v3::{
+            InvitationRecipient, InviteUserId, Request as InviteUserRequest,
+        },
+    };
+
+    let room_id = room_id
+        .parse::<OwnedRoomId>()
+        .with_context(|| format!("Invalid Matrix room ID: {room_id}"))?;
+    let user_id = user_id
+        .parse::<OwnedUserId>()
+        .with_context(|| format!("Invalid Matrix user ID: {user_id}"))?;
+
+    Ok(InviteUserRequest::new(
+        room_id,
+        InvitationRecipient::UserId(InviteUserId::new(user_id)),
+    ))
+}
+
 impl ::zeroclaw_api::attribution::Attributable for MatrixChannel {
     fn role(&self) -> ::zeroclaw_api::attribution::Role {
         ::zeroclaw_api::attribution::Role::Channel(::zeroclaw_api::attribution::ChannelKind::Matrix)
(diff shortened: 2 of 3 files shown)


Nillth (commented):
Really nice recovery — I read the whole change against current `master` and the vendored matrix-sdk 0.17 / ruma source, and the mechanics check out: `create_room`/`invite_user` are extracted into pure request-builder functions (so the six unit tests cover field mapping and every validation path without a live homeserver), each SDK call matches the vendored API, using the direct `invite_user::v3` request instead of `Room::invite_user` is the right call given a freshly-known room ID may not be in the local store yet, CI is fully green, and #4504 / @theredspoon is credited per the supersede convention. No blocking concerns from me.

One thing before I approve: this lands `create_room`/`invite_user` as currently-unwired API — no production caller invokes them yet — which brushes the "no speculative abstractions" guidance. It's reasonable here as deliberate recovery of the previously-merged, issue-tracked surface (#6074), but could you open or link a tracking issue for wiring these to a real consumer, so the recovered surface doesn't sit dead indefinitely? With that tracked I'm happy to flip to approve.

Two minor, non-blocking notes:
- The SDK-version comment at `matrix.rs:1540` still says `matrix-sdk 0.16` while the file header was updated to 0.17 — worth syncing both so they don't drift.
- A one-line doc on why the two new trait defaults `bail!` (vs the silent `Ok(())` the other optional `Channel` methods use) would save the next contributor a double-take; the intent reads as deliberate, just worth stating.

singlerider:
@Audacity88 Matrix is my baby and I'd like to keep it as a good "reference" channel. Has this been tested by you, first-hand?

Audacity88:
> @Audacity88 Matrix is my baby and I'd like to keep it as a good "reference" channel. Has this been tested by you, first-hand?

@singlerider no, it's a recovery PR, I don't have Matrix. Are you able to test it?

tidux (commented):
I read this against the current PR head and current `upstream/master`, including @Nillth's review, @singlerider's Matrix smoke-test question, #6074, and the original #4504 shape. I'm posting this as a comment rather than an approval for now: I don't see a code blocker in the diff, but the consumer-tracking question and Matrix owner/manual-test question are still open enough that I don't want my review to imply the PR is merge-ready yet.

### 🟡 Warning — keep the recovered API tied to a real consumer

I agree with @Nillth's framing here. Recovering `Channel::create_room` / `Channel::invite_user` is defensible because this is explicitly carrying forward previously merged work from #4504 as part of the #6074 recovery sweep, not inventing a random new abstraction. The remaining process risk is that the surface is currently unwired: no tool, RPC path, CLI path, or autonomous workflow calls these methods yet.

Before merge, please either link an existing issue that specifically owns the consumer/tool wiring or open a small follow-up issue for it. #6074 tracks the lost-commit recovery itself, but it does not by itself close the "who will exercise this API in production?" loop. A concrete follow-up keeps this on the right side of the AGENTS.md no-speculative-abstractions rule without bloating this PR.

### 🟡 Warning — Matrix room management still needs live homeserver evidence or owner sign-off

The request-builder tests are the right local shape, and I reproduced the focused validation locally. They prove the validation and field mapping paths without needing a homeserver. What they cannot prove is the end-to-end Matrix behavior of `client.create_room(request)` and the direct `invite_user::v3` send against a real homeserver.

Given @singlerider explicitly asked whether this had been tested first-hand, and the answer was "no, I don't have Matrix," I would like either a short live smoke result from someone with Matrix access or an explicit Matrix-owner decision that the typed SDK/API cross-check plus unit coverage is enough for this recovery slice. I am not making this a code block, but it is the reason I'm not converting to approval in this pass.

### 🟢 What looks good — the Matrix implementation avoids the old cache trap

The direct invite request is the right improvement over the original #4504 approach. Building `membership::invite_user::v3::Request` and sending it through the client means a freshly-created or externally-known room ID does not have to be present in the SDK's local room cache before invite works. That is exactly the edge case this API would otherwise trip over.

The local validation boundary is also clean: room IDs and user IDs are parsed into ruma-owned types before the network call, visibility is fail-closed to `private`/`public`, and the encryption path uses typed `RoomEncryptionEventContent::with_recommended_defaults()` rather than hand-rolled event JSON. That gives future callers useful errors before they hit the homeserver.

### 🟢 What looks good — unsupported channels now fail closed, and pacing does not hide the capability

Changing the trait defaults to explicit unsupported errors is the right contract for future caller code. A no-op `invite_user` would be dangerous because a caller could believe an invite happened when the channel silently ignored it. Returning `channel does not support room creation` / `channel does not support room invites` forces the future tool/RPC layer to handle unsupported channels honestly.

The `PacedChannel` forwarding is also the right wrapper behavior. Room creation and membership invites are management operations, not paced outbound replies, so forwarding straight to the inner channel preserves capability without routing them through the send/finalize queue.

### 🟢 What looks good — current-master integration is mechanical

The PR base is behind current `master`, so I checked the current-master auto-merge shape rather than only the PR head. `git merge-tree --write-tree HEAD upstream/master` produced b1038a53c2ea6c4a57b7675e14108396cad762d8 with no conflicts. On that merged tree archive I ran the focused format/tests/clippy pass:

- `cargo fmt -p zeroclaw-api -p zeroclaw-channels -- --check`
- `cargo test -p zeroclaw-api default_room_management_methods_report_unsupported` — 1 passed
- `cargo test -p zeroclaw-channels room_management --features channel-matrix --lib` — 7 passed
- `cargo test -p zeroclaw-channels room_management_forwards_to_inner_channel --lib` — 1 passed
- `cargo clippy -p zeroclaw-api -p zeroclaw-channels --features channel-matrix --lib -- -D warnings`

All of those completed cleanly here. I also confirmed the public CI on the PR head is green.

### 🔵 Suggestion — fold in @Nillth's two small doc nits while touching the file

Not duplicating these as separate findings, but I confirmed both are still worth doing if the branch gets another push: the internal comment in `matrix.rs` still mentions `matrix-sdk 0.16` even though the crate is on 0.17, and the two new public trait methods would benefit from a slightly fuller error-contract note explaining why they `bail!` while older optional channel methods no-op. Neither changes the merge decision for me; they just make the recovered API easier for the next implementor to read correctly.

singlerider:
I'd want this tested on a Matrix room before merging.

Audacity88:
@singlerider, I opened #7769 to track the consumer/tool wiring that @Nillth and @tidux asked for.

That keeps #7661 scoped to the recovered channel API and Matrix implementation. The follow-up covers the first real caller for `create_room` / `invite_user`.

The remaining merge question is Matrix live evidence. I applaud your request for evidence, but I unfortunately do not use Matrix myself, so this still needs a short homeserver smoke test from someone with Matrix access. If you are able to do so, that would be great. I will also look for someone else who might be able to, and I opened Discussion #7770 to ask whether someone with Matrix access can run that smoke test.

singlerider (changes requested):
Re-review on the current head. Taking stock of what is already on this PR: @Nillth and @tidux both raised the unwired-API / consumer-tracking concern, and @Audacity88 resolved it by opening #7769 to own the first real caller. That closes the speculative-abstraction process question for me. The code itself is a faithful recovery: request building is extracted into pure `build_create_room_request` / `build_invite_user_request` functions, the six unit tests exercise field mapping and every validation path without a homeserver, and the direct `invite_user::v3` send is the correct choice for a freshly-known room ID not yet in the local store.

Two things hold this back.

### 🔴 Blocking - string-literal match on visibility

`build_create_room_request` decides visibility with a match on string literals:

```rust
request.visibility = match visibility {
    "private" => Visibility::Private,
    "public" => Visibility::Public,
    other => bail!("Invalid visibility: ..."),
};
```

We do not match on string literals in this codebase; identity and option-sets live in enums, not in stringly-typed branch arms. ruma already gives you `Visibility`, so the cleanest form is to parse into it (or a small local enum with a `FromStr`/`TryFrom<&str>`) and reject the unknown case there, rather than branching on the raw `&str` here. Same outcome, but the valid set stops being a literal list embedded in a function body.

### 🔴 Blocking - no live Matrix evidence, and I asked for it

I asked earlier whether this was tested first-hand against a real homeserver, and the answer was no. Matrix is the reference channel and I want `create_room` plus the direct `invite_user::v3` send exercised on a live room before this lands: request builders prove field mapping and validation, they cannot prove the SDK call actually creates the room and the invite is delivered. I will run this live test myself against my homeserver and report the evidence back here. Until that pass is posted this stays blocked on my gate, independent of the code nit above.

### 🟢 What looks good - pure request builders make the surface testable

Splitting the SDK calls from request construction is the right structure. The validation paths (invalid user ID, invalid room ID, invalid visibility) are all unit-covered without a homeserver, and the encryption-on path maps to `RoomEncryptionEventContent::with_recommended_defaults()` cleanly.

### 🟢 What looks good - PacedChannel forwards instead of swallowing

`PacedChannel::create_room` / `invite_user` forward to the inner channel rather than inheriting the unsupported default, so enabling reply pacing does not silently hide the new capability. The default-trait `bail!` (instead of silent `Ok(())`) is also the right call for a capability that is genuinely absent.

singlerider:
Closing this as refused. Verified live against matrix.org with a throwaway daemon, real bot account, and a real homeserver.

The two channel methods (`create_room`, `invite_user`) function in isolation, but the PR ships only the channel-side plumbing. There is no agent-facing tool that reaches a live `MatrixChannel` instance, so there is no path to create a room or invite a user through the running daemon. Messaging the bot to "create a room and invite me" does nothing because nothing is wired from the tool layer to these methods. Confirmed by running the daemon and chatting the bot directly: it has no room-management capability.

For this to land, the work needs to introduce a real channel-management tool (the wiring tracked in #7769): an agent-facing tool that holds a handle to the live channel and exposes room creation and invites end to end, with the path exercised through the daemon rather than the channel methods in isolation.

Separate code-quality blocker that also stands: the string-literal `match` on room visibility must become an enum.

This is a refusal on scope and completeness, not a judgment on the author. The direction is right; the deliverable needs the tool layer to be usable. Reopen with the tool wired and a demonstrated daemon-driven flow and I will re-review.

singlerider:
I tested this and it does not work as-wired. It needs complete rescoping.

PR #7665, opened 2026-06-15
feat(web): Operator Console redesign of the gateway web UI (schema-driven config, multi-agent chat, ported fixes)

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** A ground-up redesign of the gateway `web/` console into a coherent "Operator Console" design system, because the prior UI was a thin shell over the config API with navigation dead-ends and no visual language.
  - **Design system + shell:** new token set + UI primitives (`Card`/`Button`/`Badge`/`PageHeader`/`StatCard`/`ConfirmDialog`/`ComboBox`), slim icon rail + mobile drawer, ⌘K command palette.
  - **Config (the bulk):** schema-driven master-detail navigator (searchable section→entity tree, `+Add` for empty/previously-unreachable sections), inline validation, reference/alias pickers, humanized labels — all preserving the existing schema-driven JSON-Patch / drift / `?tab=` contract (#6175), no hardcoded section/field names.
  - **Dashboard / Chat / Tools / Integrations / Cron / Doctor / Pairing:** real-metric dashboard with list+drawer; multi-agent chat workspace (tabs + split, panes stay WS-connected) with per-agent model switch; Tools page with a per-risk-profile allow/block matrix; Integrations as a launcher into config; Doctor fix-in-place modal + deep-links; pairing list with unpaired-vs-zero distinction.
  - **Four upstream fixes ported** onto the redesign (see *Incorporated contributions*).
  - **i18n / a11y / perf:** all user-facing strings routed through `t()` (+531 en keys); focus traps, keyboard nav, 44px targets, `ConfirmDialog` replacing native `confirm/prompt`; polling pauses on hidden tab; `apiFetch` coalesces concurrent identical GETs.
- **Scope boundary:** **Pure web — zero `crates/` changes.** No gateway/runtime/channel/provider/tool/schema changes. The daemon API contract is consumed as-is.
- **Blast radius:** the gateway web dashboard only. No change to the WS/HTTP protocol, endpoints, or config schema.
- **Linked issue(s):**
  - `Fixes #7157` — chat-bubble server-timestamp leak (fully resolved web-side).
  - `Related #6891` — cron edit 422 is fixed web-side; the issue also asks for more PATCH-editable fields, which needs a `CronPatchBody` (backend) change.
  - `Related #6488` — integration category labels now render human-readably in the UI; the issue's `category_label` API field + Rust test remain a backend item (see PR #6490).
[...]

PR #7666, opened 2026-06-15
feat(gateway/cron): accept enabled on CronPatchBody for pause/resume; scope the agent check to shell-command patches

### LINKED ISSUE #7356: [Feature]: Add a pause/disable button to Scheduled Tasks

Nillth:
> Maintainer note (2026-06-17): Audacity88 refreshed the label snapshot and validation evidence after d0cd25809e. The implementation summary remains the author's.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **Pause/resume over HTTP:** add `enabled: Option<bool>` to `CronPatchBody` and thread it into `CronJobPatch`. The runtime cron store/scheduler already honor `enabled` (disabled jobs don't run), but the gateway PATCH handler dropped the field, so a job could never be paused/resumed via the API. This wires it through. `None` leaves the state unchanged.
  - **Agent requirement scoped to shell-command patches:** `agent` becomes `#[serde(default)]`, and the agent-existence check now runs **only** when a shell `command` is actually being set — i.e. a non-agent (shell) job whose patch carries `command` **or** `prompt` (the `prompt` field is folded into the shell command downstream). Schedule/name/timezone changes and enable/disable toggles no longer need an `agent` field, fixing the `422 "missing field agent"` they previously returned. Agent-type jobs (where `prompt` is an LLM prompt, not a shell command) are correctly not agent-gated.
  - **400 not 500:** because the existence check now also covers the `prompt`-as-shell-command path, an unknown agent on that path returns a clean `400` instead of falling through to a `500` downstream.
- **Scope boundary:** the gateway cron PATCH handler + `CronPatchBody` only. No change to the runtime cron store/scheduler (they already support `enabled`), the WS protocol, or any other endpoint.
- **Blast radius:** `PATCH /api/cron/{id}` request shape and validation. Existing callers that always sent `agent` keep working (it's still accepted); callers can now omit it for non-command patches and supply `enabled` to toggle.
- **Linked issue(s):**
  - `Closes #7356` — "Add a pause/disable button to Scheduled Tasks." This is the gateway half; the web button shipped in the redesign PR (#7665, merged). The two work together to make pause/resume work end-to-end.
  - `Supersedes #7384` — replaces the earlier pause/resume PR (see Supersede Attribution).
- **Labels:** `enhancement`, `gateway`, `risk: high`, `size: S`.

Rust change — focused validation reported on current head d0cd25809e:

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings
cargo test -p zeroclaw-gateway
```

  - `cargo fmt --all -- --check` → **exit 0** (clean).
  - `cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings` → **exit 0** (clean).
  - `cargo test -p zeroclaw-gateway` → `test result: ok. **262 passed; 0 failed**`, exit 0.
- **Beyond CI — what did you manually verify?** Companion PR #7665 shipped the web pause/resume control. This PR wires the gateway endpoint that control needs, and the new gateway handler regression tests cover `enabled` without `agent`, name/schedule without `agent`, shell-command patches without a known `agent`, shell-job `prompt` with unknown `agent` returning 400 instead of 500, and agent-job `prompt` without `agent`.
- **If any command was intentionally skipped, why:** full-workspace local `cargo test` was not rerun; this PR touches `crates/zeroclaw-gateway/src/api.rs`, and the focused gateway crate tests plus the visible GitHub checks are green on current head d0cd25809e. No fresh web E2E smoke is included in this PR-body update.

- **Risk-gating note:** the `agent` field gates the risk profile for **shell-command** changes. This PR *preserves* that gate — and in fact *widens* its coverage to the `prompt`-as-shell-command path — while correctly skipping it only for patches that do not modify the executed command (schedule/name/enable). No risk-profile bypass is introduced.

## Compatibility (required)

- Backward compatible? **Yes** — `enabled` is optional (`None` = unchanged); `agent` moving to `#[serde(default)]` only *relaxes* a previously-required field, so existing clients that send it are unaffected.
- Upgrade steps: none.

- **Fast rollback command/path:** revert this PR's eventual squash merge — single file, no state/schema migration.
- **Observable failure symptoms:** cron PATCH returns 422/400 on valid toggles, or pause/resume no longer takes effect; jobs still run/dispatch normally otherwise.

## Supersede Attribution (required — `Supersedes #7384` is used)

- **Superseded PRs + authors:** `#7384 by @Nillth`
- **Scope materially carried forward:** the pause/resume capability (`enabled` on `CronPatchBody` + threading into `CronJobPatch`) from #7384, refined and combined here with the agent-requirement scoping and the 400-not-500 fix so non-command patches and the `prompt`-as-command path are handled correctly.
- **`Co-authored-by` trailers added for incorporated contributors?** N/A — same author (@Nillth); no external contributor code incorporated.

### DIFF
--- a/crates/zeroclaw-gateway/src/api.rs
+++ b/crates/zeroclaw-gateway/src/api.rs
@@ -3259,6 +3259,280 @@ mod tests {
         );
     }
 
+    // --- PATCH agent-requirement boundary (#7666) ---------------------------
+    // These pin the relaxed `agent` rule: it is required only when a patch sets a
+    // *shell command* (the risk-profile gate), and not for enable/disable,
+    // metadata-only patches, or agent-type jobs. Regression coverage for the
+    // `#[serde(default)] agent` + `setting_shell_command` scoping so a future
+    // refactor can't silently reopen the gate or re-require agent where it isn't.
+
+    #[tokio::test]
+    async fn cron_api_patch_enabled_without_agent() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let config = zeroclaw_config::schema::Config {
+            data_dir: tmp.path().join("data"),
+            config_path: tmp.path().join("config.toml"),
+            ..zeroclaw_config::schema::Config::default()
+        };
+        std::fs::create_dir_all(&config.data_dir).unwrap();
+        let state = test_state(with_test_agent(config));
+        let job = zeroclaw_runtime::cron::add_shell_job_with_approval(
+            &state.config.read().clone(),
+            "test-agent",
+            Some("toggle-job".to_string()),
+            zeroclaw_runtime::cron::Schedule::Cron {
+                expr: "*/5 * * * *".to_string(),
+                tz: None,
+            },
+            "echo hello",
+            None,
+            true,
+        )
+        .expect("job added");
+
+        // No `agent` field at all — pause/resume must not require one.
+        let response = handle_api_cron_patch(
+            State(state.clone()),
+            HeaderMap::new(),
+            Path(job.id.clone()),
+            Json(
+                serde_json::from_value::<CronPatchBody>(serde_json::json!({ "enabled": false }))
+                    .expect("body should deserialize"),
+            ),
+        )
+        .await
+        .into_response();
+
+        assert_eq!(
+            response.status(),
+            StatusCode::OK,
+            "enable/disable toggle must not require an agent"
+        );
+        let updated = zeroclaw_runtime::cron::get_job(&state.config.read().clone(), &job.id)
+            .expect("updated job");
+        assert!(!updated.enabled, "job should be disabled after the patch");
+    }
+
+    #[tokio::test]
+    async fn cron_api_patch_name_and_schedule_without_agent() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let config = zeroclaw_config::schema::Config {
+            data_dir: tmp.path().join("data"),
+            config_path: tmp.path().join("config.toml"),
+            ..zeroclaw_config::schema::Config::default()
+        };
+        std::fs::create_dir_all(&config.data_dir).unwrap();
+        let state = test_state(with_test_agent(config));
+        let job = zeroclaw_runtime::cron::add_shell_job_with_approval(
+            &state.config.read().clone(),
+            "test-agent",
+            Some("old-name".to_string()),
+            zeroclaw_runtime::cron::Schedule::Cron {
+                expr: "*/5 * * * *".to_string(),
+                tz: None,
+            },
+            "echo hello",
+            None,
+            true,
+        )
+        .expect("job added");
+
+        // Metadata-only patch (no command/prompt) — agent must be optional.
+        let response = handle_api_cron_patch(
+            State(state.clone()),
+            HeaderMap::new(),
+            Path(job.id.clone()),
+            Json(
+                serde_json::from_value::<CronPatchBody>(serde_json::json!({
+                    "name": "new-name",
+                    "schedule": "30 9 * * *"
+                }))
+                .expect("body should deserialize"),
+            ),
+        )
+        .await
+        .into_response();
+
+        assert_eq!(
+            response.status(),
+            StatusCode::OK,
+            "name/schedule patch must not require an agent"
+        );
+        let updated = zeroclaw_runtime::cron::get_job(&state.config.read().clone(), &job.id)
+            .expect("updated job");
+        assert_eq!(updated.name.as_deref(), Some("new-name"));
+        assert_eq!(
+            updated.schedule,
+            zeroclaw_runtime::cron::Schedule::Cron {
+                expr: "30 9 * * *".to_string(),
+                tz: None,
+            }
+        );
+    }
+
+    #[tokio::test]
+    async fn cron_api_patch_shell_command_requires_known_agent() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let config = zeroclaw_config::schema::Config {
+            data_dir: tmp.path().join("data"),
+            config_path: tmp.path().join("config.toml"),
+            ..zeroclaw_config::schema::Config::default()
+        };
+        std::fs::create_dir_all(&config.data_dir).unwrap();
+        let state = test_state(with_test_agent(config));
+        let job = zeroclaw_runtime::cron::add_shell_job_with_approval(
+            &state.config.read().clone(),
+            "test-agent",
+            Some("shell-job".to_string()),
+            zeroclaw_runtime::cron::Schedule::Cron {
+                expr: "*/5 * * * *".to_string(),
+                tz: None,
+            },
+            "echo hello",
+            None,
+            true,
+        )
+        .expect("job added");
+
+        // Setting a shell `command` still hits the risk gate: a missing agent
+        // must be a clean 400, not a fall-through.
+        let response = handle_api_cron_patch(
+            State(state.clone()),
+            HeaderMap::new(),
+            Path(job.id.clone()),
+            Json(
+                serde_json::from_value::<CronPatchBody>(
+                    serde_json::json!({ "command": "echo bye" }),
+                )
+                .expect("body should deserialize"),
+            ),
+        )
+        .await
+        .into_response();
+
+        assert_eq!(
+            response.status(),
+            StatusCode::BAD_REQUEST,
+            "shell command patch with no agent must be rejected at the risk gate"
+        );
+        let json = response_json(response).await;
+        assert!(
+            json["error"]
+                .as_str()
+                .unwrap_or_default()
+                .contains("Unknown agent"),
+            "error should name the unknown agent"
+        );
+    }
+
+    #[tokio::test]
+    async fn cron_api_patch_shell_prompt_unknown_agent_is_bad_request_not_500() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let config = zeroclaw_config::schema::Config {
+            data_dir: tmp.path().join("data"),
+            config_path: tmp.path().join("config.toml"),
+            ..zeroclaw_config::schema::Config::default()
+        };
+        std::fs::create_dir_all(&config.data_dir).unwrap();
+        let state = test_state(with_test_agent(config));
+        let job = zeroclaw_runtime::cron::add_shell_job_with_approval(
+            &state.config.read().clone(),
+            "test-agent",
+            Some("shell-job".to_string()),
+            zeroclaw_runtime::cron::Schedule::Cron {
+                expr: "*/5 * * * *".to_string(),
+                tz: None,
+            },
+            "echo hello",
+            None,
+            true,
+        )
+        .expect("job added");
+
+        // For a shell job a new command can arrive via `prompt`; it still routes
+        // through the command-risk gate, so an unknown agent is a 400 — not the
+        // 500 that an unguarded path would surface.
+        let response = handle_api_cron_patch(
+            State(state.clone()),
+            HeaderMap::new(),
+            Path(job.id.clone()),
+            Json(
+                serde_json::from_value::<CronPatchBody>(serde_json::json!({
+                    "agent": "ghost",
+                    "prompt": "echo bye"
+                }))
+                .expect("body should deserialize"),
+            ),
+        )
+        .await
+        .into_response();
+
+        assert_eq!(
+            response.status(),
+            StatusCode::BAD_REQUEST,
+            "shell-job prompt with unknown agent must be 400, not 500"
+        );
+    }
+
+    #[tokio::test]
+    async fn cron_api_patch_agent_prompt_without_agent() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let config = zeroclaw_config::schema::Config {
+            data_dir: tmp.path().join("data"),
+            config_path: tmp.path().join("config.toml"),
+            ..zeroclaw_config::schema::Config::default()
+        };
+        std::fs::create_dir_all(&config.data_dir).unwrap();
+        let state = test_state(with_test_agent(config));
+
+        let add_response = handle_api_cron_add(
+            State(state.clone()),
+            HeaderMap::new(),
+            Json(
+                serde_json::from_value::<CronAddBody>(serde_json::json!({
+                    "name": "agent-job",
+                    "agent": "test-agent",
+                    "schedule": "*/5 * * * *",
+                    "job_type": "agent",
+                    "command": "ignored",
+                    "prompt": "old prompt"
+                }))
+                .expect("body should deserialize"),
+            ),
+        )
+        .await
+        .into_response();
+        assert_eq!(add_response.status(), StatusCode::OK);
+        let id = zeroclaw_runtime::cron::list_jobs(&state.config.read().clone()).unwrap()[0]
+            .id
+            .clone();
+
+        // For an agent-type job `prompt` is an LLM prompt, not a shell command,
+        // so it is not agent-gated and may omit `agent`.
+        let response = handle_api_cron_patch(
+            State(state.clone()),
+            HeaderMap::new(),
+            Path(id.clone()),
+            Json(
+                serde_json::from_value::<CronPatchBody>(
+                    serde_json::json!({ "prompt": "new prompt" }),
+                )
+                .expect("body should deserialize"),
+            ),
+        )
+        .await
+        .into_response();
+
+        assert_eq!(
+            response.status(),
+            StatusCode::OK,
+            "agent-type prompt patch must not require an agent"
+        );
+        let updated = zeroclaw_runtime::cron::get_job(&state.config.read().clone(), &id)
+            .expect("updated job");
+        assert_eq!(updated.prompt.as_deref(), Some("new prompt"));
+    }
+
     #[tokio::test]
     async fn cron_api_rejects_announce_delivery_without_target() {
         let tmp = tempfile::TempDir::new().unwrap();
--- a/crates/zeroclaw-gateway/src/api.rs
+++ b/crates/zeroclaw-gateway/src/api.rs
@@ -624,19 +624,6 @@ pub async fn handle_api_cron_patch(
     }
 
     let config = state.config.read().clone();
-    // The agent only gates a shell `command` change (risk profile). Skip the
-    // existence check for patches that don't touch the command — schedule, name,
-    // and enable/disable toggles don't need an agent supplied.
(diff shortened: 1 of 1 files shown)


Audacity88:
Thanks, this gateway-only split looks like the right shape. The implementation direction makes sense to me: `enabled` is just a PATCH input threaded into the existing cron state, and moving the agent check after loading the existing job is the right way to distinguish shell-job `prompt` from agent-job `prompt`.

Before approval, please refresh this branch over current `master`. The PR still merges cleanly, but current `master` has same-file churn in `crates/zeroclaw-gateway/src/api.rs`, so the visible green checks did not test the current merge result.

Please also add focused gateway handler regression tests for the exact PATCH boundary this changes. Runtime/store tests already cover `enabled`, but this PR changes request deserialization and gateway validation:

- `enabled` can be patched without `agent`.
- name/schedule/timezone-only patches can omit `agent`.
- shell command patches still reject a missing or unknown `agent`.
- shell-job `prompt` patches follow the same command-risk gate.
- agent-job `prompt` patches do not require `agent`.

Separate maintainer cleanup: the live labels/body snapshot also need alignment. Since this touches `crates/zeroclaw-gateway/src/api.rs`, I expect this to be `risk: high`; with 41 changed lines, the size label should be `size: XS`.

JordanTheJet (changes requested):
I read the PR body, @Audacity88's comment, the full diff, and cross-checked it against the cron handler and runtime store on `master`. I'm the first formal reviewer here. The change is small, correct, and well-scoped — I traced the risk-gate path end-to-end and it holds, so this is not a correctness objection. I'm requesting changes on one thing: a `crates/zeroclaw-gateway/src/**` change that alters request deserialization and validation branching needs tests pinning the new boundary before it merges, and @Audacity88 already asked for exactly those. Everything else below is supporting context (a rebase and a label fix).

### 🟢 What looks good — the risk-gate scoping is exactly right

This is the load-bearing part of the change, so I verified it carefully. The actual gate is `validate_shell_command`, and `update_shell_job_with_approval` only invokes it when `patch.command.is_some()` (crates/zeroclaw-runtime/src/cron/mod.rs:145-147). In the handler, `patch_command` is `Some` precisely when `!is_agent && (command.is_some() || prompt.is_some())` (api.rs:679-683) — which is *identical* to the new `setting_shell_command` guard. So the early existence check now fires on exactly the set of patches that will hit the risk gate, and the `#[serde(default)]` empty-alias default is never consumed on the skipped paths (non-command patches and agent-type jobs never reach `validate_shell_command`). No bypass is introduced, and the early `400` is strictly a better error than letting an empty/unknown alias fall through to a `500`. The PR body's "preserves the gate, widens its coverage" claim is accurate.

### 🟢 What looks good — `enabled` threading closes a real gateway gap

Confirmed the runtime side already honors it: `CronJobPatch.enabled: Option<bool>` (crates/zeroclaw-runtime/src/cron/types.rs:201) is applied at crates/zeroclaw-runtime/src/cron/store.rs:326-327 (`if let Some(enabled) = patch.enabled { job.enabled = enabled; }`), and `None` leaves the row untouched. So the gateway was the only layer dropping the field, and wiring it through is the minimal correct fix for the pause/resume half of #7356.

### 🔴 Blocking — no tests for the new PATCH boundary

The diff touches request deserialization (`agent` → `#[serde(default)]`) and the gateway validation branch, but adds no tests. The 250 passing tests cited in the validation evidence are pre-existing and don't exercise the new behavior. Per FND-005, missing coverage for new behavior is a blocking item, and on a `crates/zeroclaw-gateway/src/**` path — where the field being relaxed is the risk-profile gate — I want this pinned before merge rather than after. This is the one item I'm formally requesting changes on. Reinforcing @Audacity88's earlier ask — the focused regression cases that pin this exact boundary:

- `enabled` can be patched without `agent`.
- name/schedule/timezone-only patches can omit `agent`.
- shell `command` patches still reject a missing or unknown `agent` with a clean `400`.
- shell-job `prompt` patches follow the same command-risk gate (and `400`, not `500`, on unknown agent).
- agent-type `prompt` patches do not require `agent`.

These are the precise lines the diff changed, so they're cheap to add and they're what protects this behavior from a future refactor silently reopening the bypass.

### 🟡 Warning — refresh over current `master` before merge

The required checks are green, but they ran on a head that predates current `master`'s churn in crates/zeroclaw-gateway/src/api.rs, so they did not test the merge result. A `git rebase upstream/master` (or merge) plus a fresh green run is worth doing before squash. Endorsing @Audacity88's point here.

### 🔵 Suggestion — label alignment

The PR currently carries only `enhancement`, `cron`, `gateway` — no `risk:` or `size:`. Per `AGENTS.md`, `crates/zeroclaw-gateway/src/**` is a high-risk path, so this should be `risk: high` (the body's `risk: medium` undersells it). The diff is 41 changed lines, which lands at `size: S` on the repo's thresholds. Labels are maintainer metadata, not a contributor blocker — I'm happy to apply `risk: high` + `size: S` myself so it's not on you.

Net: the implementation is right and I'm satisfied with the security reasoning — the only thing holding the approval is the missing boundary tests (🔴). Add those, refresh over `master` for a clean green run, and I'll flip to approve. I'll take the label fix off your plate.

Nillth:
@JordanTheJet thanks — done, pushed `d0cd25809e`.

Added the 5 boundary tests you and @Audacity88 asked for (in `crates/zeroclaw-gateway/src/api.rs`), pinning exactly the lines the diff changed:
- `enabled` patched **without** `agent` → 200
- name/schedule-only patch without `agent` → 200
- shell `command` patch with missing `agent` → 400 (`Unknown agent`)
- shell-job `prompt` patch with unknown `agent` → 400, **not** 500
- agent-type `prompt` patch without `agent` → 200

Also merged `upstream/master` into the branch (`b33b37e15f`) so CI runs on the merge result rather than the stale head you flagged. `cargo test -p zeroclaw-gateway` → 262 passed; `cargo fmt --all -- --check` clean; `cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings` clean.

Audacity88 (commented):
Thanks for the update. I re-read current head d0cd25809e against the live PR state, the prior review/comment history, the `crates/zeroclaw-gateway/src/api.rs` diff, the runtime cron update path, and the linked #7356 / #7384 / #7665 context. I did not run local Cargo for this review-only pass; the visible GitHub checks are green on the PR head, and I used source inspection plus the remote CI results as the evidence path.

### ✅ Resolved — PATCH boundary tests now cover the requested cases

The new gateway tests cover the boundary my earlier comment and Jordan's formal review called out: `enabled` can be patched without `agent`, name/schedule-only patches can omit `agent`, shell `command` patches without a known agent return 400, shell-job `prompt` patches with an unknown agent return 400 instead of falling through to 500, and agent-job `prompt` patches can omit `agent`. Those are the load-bearing cases for the `#[serde(default)] agent` change, so the missing-test blocker is resolved from my side.

### 🟢 What looks good — the shell-command risk gate still holds

The handler now fetches the existing job before deciding whether the patch is setting a shell command. That matters because `prompt` means two different things: for shell jobs it is folded into the command path, while for agent jobs it is an LLM prompt. The new `setting_shell_command` check matches the runtime behavior: shell-job `command` or `prompt` updates still require a configured agent before `update_shell_job_with_approval` reaches the command validation path, while schedule/name/enable toggles and agent-job prompt edits no longer need an unrelated `agent` field.

### 🟡 Warning — branch freshness still needs a final refresh before merge

This head is now behind current `master` again, and `master` has changed `crates/zeroclaw-gateway/src/api.rs` since the PR's merge base. The same-file change I saw is in the gateway test-state setup for channel alias routing, not in the cron PATCH handler, and the merge shape looks clean from source inspection. Still, because this is a high-risk gateway API path, I would not treat the PR as merge-ready until the branch is updated and CI runs on the updated head, or maintainers explicitly accept that stale-risk.

Jordan's formal `CHANGES_REQUESTED` review also remains active, so this comment is not an approval and is not trying to override that review. From my side, the implementation and test updates address the earlier cron PATCH boundary ask; the remaining public blocker is review/freshness state rather than a new code finding from this pass.

Audacity88 (commented):
Reviewed current head d0cd25809e against the live PR state, #7356, #7665, #7384, the earlier Audacity88 comment, @JordanTheJet's formal review, the maintainer-refreshed PR body, current labels, visible GitHub checks, the one-file gateway diff, and the new gateway tests. I did not run local cargo; this status is based on the author's current focused validation plus green visible checks. This is a status comment, not an approval review, because @JordanTheJet's formal `CHANGES_REQUESTED` review remains active.

### ✅ Resolved — main gateway PATCH boundary coverage is in place

The new tests pin the main command/non-command cases from the earlier review pass: `enabled` without `agent`, name/schedule without `agent`, shell `command` without a known `agent`, shell-job `prompt` with an unknown `agent` returning 400 instead of 500, and agent-job `prompt` without `agent`. Earlier feedback also mentioned timezone-only patches; that is not a separate new test here, but this is enough for me to treat the gateway command-risk boundary as covered for re-review.

### 🟢 What looks good — the command-risk gate remains scoped correctly

The gateway handler now checks for an agent only when a shell command is being set, including the shell-job `prompt` path. Non-command patches and agent-job prompts skip that early check, and the diff still routes shell-job `command` / `prompt` patches into the shell-command patch path while keeping agent-job prompts as prompts. That preserves the command-risk boundary while removing the unnecessary `agent` requirement for pause/resume and metadata-only patches.

The branch was refreshed over current `master`, visible GitHub checks are green, and the PR body now reflects the live `risk: high` label plus current focused gateway validation on d0cd25809e.

From my earlier asks, this is ready for @JordanTheJet's re-review. The PR is still not merge-ready while @JordanTheJet's `CHANGES_REQUESTED` review remains active.

singlerider (commented):
I re-reviewed at d0cd2580 on a fresh checkout of `feat/cron-patch-gateway` (MERGEABLE over current master). Built the gateway crate, ran the new boundary tests, and ran clippy. Reviewing the head commit on its current merits.

### 🟢 The command-risk gate is scoped correctly and the boundary is now pinned

The unconditional agent-existence check is gone, replaced by `setting_shell_command = !is_agent && (command.is_some() || prompt.is_some())`. This is the right shape: for a shell job, `patch_command = command.or(prompt)`, so a new shell command can arrive through either field, and the gate covers both. For an agent job, `prompt` is an LLM prompt and is correctly not agent-gated. Pause/resume, name, schedule, and timezone patches no longer require an agent they have no reason to supply.

I verified the risk boundary holds end to end: `cron_api_patch_shell_prompt_unknown_agent_is_bad_request_not_500` proves a shell command delivered via `prompt` with an unknown agent returns 400, not an ungated mutation and not a 500. The `enabled: Option<bool>` field already existed on `CronJobPatch`; this PR only wires the gateway body into it, and the store applies it.

Local validation on d0cd2580:

```
running 5 tests
test api::tests::cron_api_patch_shell_command_requires_known_agent ... ok
test api::tests::cron_api_patch_shell_prompt_unknown_agent_is_bad_request_not_500 ... ok
test api::tests::cron_api_patch_agent_prompt_without_agent ... ok
test api::tests::cron_api_patch_enabled_without_agent ... ok
test api::tests::cron_api_patch_name_and_schedule_without_agent ... ok

test result: ok. 5 passed; 0 failed
```

`cargo clippy -p zeroclaw-gateway --all-targets` is clean.

From my side this is ready. The PR remains blocked only by @JordanTheJet's active `CHANGES_REQUESTED` from 2026-06-16; the test-pin commit landed 2026-06-17 and covers the boundary that review asked for. @JordanTheJet — this looks resolved to me; please convert or dismiss when you get a chance.

Nillth:
@JordanTheJet — thanks again for flagging the missing boundary coverage on the cron PATCH handler. That landed in the test-pin commit ("test(gateway/cron): pin the PATCH agent-requirement boundary"), which adds five focused tests around the relaxed `agent` rule (agent is required only when a patch sets a *shell command* — the risk gate — and not for enable/disable, metadata-only patches, or agent-type jobs):

- `cron_api_patch_shell_command_requires_known_agent` — setting a shell `command` still hits the risk gate; a missing/unknown agent is a clean 400, not a fall-through.
- `cron_api_patch_shell_prompt_unknown_agent_is_bad_request_not_500` — for a shell job, a new command arriving via `prompt` routes through the same gate, so an unknown agent is a 400, not the 500 an unguarded path would have surfaced (the bug the fix commit addresses).
- `cron_api_patch_agent_prompt_without_agent` — for an agent-type job, `prompt` is an LLM prompt (not a shell command), so it is not agent-gated and may omit `agent`.
- `cron_api_patch_enabled_without_agent` — pause/resume (`enabled`) must not require an agent.
- `cron_api_patch_name_and_schedule_without_agent` — metadata-only patches (name/schedule) must not require an agent.

Audacity88 and singlerider have since confirmed the boundary is covered (singlerider noted it looks resolved and suggested converting or dismissing the review).

Since a high-risk gateway path deserves a fresh check, I've also refreshed the branch over current `master` (clean rebase, no conflicts; no code change required) and re-ran the suite locally on the new head:

```
test api::tests::cron_api_patch_shell_command_requires_known_agent ... ok
test api::tests::cron_api_patch_shell_prompt_unknown_agent_is_bad_request_not_500 ... ok
test api::tests::cron_api_patch_agent_prompt_without_agent ... ok
test api::tests::cron_api_patch_enabled_without_agent ... ok
test api::tests::cron_api_patch_name_and_schedule_without_agent ... ok
test result: ok. 270 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

(`cargo +1.93.0 fmt --all` is a no-op and `cargo +1.93.0 clippy -p zeroclaw-gateway --all-targets` is clean.)

Whenever you have a moment, could you re-review and convert or dismiss the CHANGES_REQUESTED? Happy to adjust anything if there's a case you'd still like pinned. Thanks!

JordanTheJet (approved):
Re-reviewed at head fed8b698 (clean linear rebase of `feat/cron-patch-gateway` over current `master`). I read the new commit history, @Audacity88's and @singlerider's re-reviews, @Nillth's two follow-up comments, and the full diff at this head, and I cross-checked the five new tests against the handler change. My prior `CHANGES_REQUESTED` rested on a single blocker — no coverage for the new PATCH boundary — plus a rebase warning and a label nit. All three are now addressed, the implementation I verified last pass is intact at the new head, and CI is green on the merge result. Flipping to approve.

### ✅ Resolved — the PATCH boundary is now pinned by tests

The blocker was that this `crates/zeroclaw-gateway/src/api.rs` change relaxes request deserialization (`agent` → `#[serde(default)]`) and re-scopes the validation branch, with nothing exercising the new behavior. The test-pin commit adds exactly the five cases I called out, and they pin the right boundary:

- `cron_api_patch_enabled_without_agent` — pause/resume with no `agent` field → 200, and asserts the job is actually disabled afterward.
- `cron_api_patch_name_and_schedule_without_agent` — metadata-only patch omits `agent` → 200, with name and schedule verified applied.
- `cron_api_patch_shell_command_requires_known_agent` — a shell `command` patch with no agent still trips the risk gate → 400 naming the unknown agent.
- `cron_api_patch_shell_prompt_unknown_agent_is_bad_request_not_500` — a shell-job command arriving via `prompt` routes through the same gate → 400, not the 500 an unguarded path would surface.
- `cron_api_patch_agent_prompt_without_agent` — an agent-type job's `prompt` is an LLM prompt, not a shell command, so it is not agent-gated → 200, with the new prompt verified persisted.

These are behavior tests, not implementation snapshots — they assert the HTTP status and the resulting persisted job state, and they deserialize bodies that omit `agent`, so they also exercise the `#[serde(default)]` path itself. That is the kind of coverage that protects this boundary from a future refactor silently reopening the bypass, which is exactly what I wanted pinned before merge. The one case from my list that doesn't get its own test is timezone-only, but it travels the identical non-command path the name/schedule test already pins (`setting_shell_command` is false for any patch that sets no command/prompt), so I'm comfortable leaving it — no action needed.

### ✅ Resolved — branch refreshed over current `master`

My 🟡 was that the green checks predated `master`'s churn in `api.rs` and so hadn't tested the merge result. The branch is now a clean rebase onto current `master` (linear, no conflicts, no code change required), and the full Quality Gate is green on this head — Build x86_64, Test, Lint, Format, Security, Check (all features / no default / 32-bit / aarch64-apple-darwin / windows-msvc), Benchmarks Compile, Docs Style, Installer Drift, Nix Module Eval, Zerocode RPC Boundary, and the CI Required Gate. This is a real green run on the merge shape, not a stale head.

### ✅ Resolved — labels aligned

The PR now carries `risk: high` and `size: S` alongside `enhancement` and `gateway`, matching the gateway high-risk path per `AGENTS.md`. As promised, this was maintainer cleanup, not on the author.

### 🟢 What still looks good — the risk-gate scoping holds at the new head

I re-confirmed the load-bearing property survives the rebase. The agent-existence check now fires precisely when `setting_shell_command = !is_agent && (command.is_some() || prompt.is_some())`, which is the exact set of patches whose `patch_command` becomes `Some` and therefore reaches `validate_shell_command` in `update_shell_job_with_approval` (cron/mod.rs:145-147). Agent-type jobs route `prompt` to the LLM-prompt path and never hit that gate; non-command patches never set a command. So no bypass is introduced, and the early 400 remains strictly better than letting an empty/unknown alias fall through to a 500 — now with a test (`..._is_bad_request_not_500`) standing guard over that exact difference. Moving the `is_agent` binding up so the gate can consult `existing.job_type` is the clean way to express "this depends on the fetched job," and it reads well.

### 🟢 What still looks good — `enabled` threading is the minimal correct fix

`enabled: Option<bool>` is wired straight from the gateway body into `CronJobPatch.enabled`, which the store already applies (store.rs:326-327), with `None` leaving state unchanged. The gateway was the only layer dropping the field; this closes the pause/resume half of #7356 without duplicating any state.

Net: the one blocker is resolved with precise, behavior-level coverage; the rebase and label items are done; the security reasoning I traced last pass is intact and now test-guarded. Thanks for the thorough follow-through, @Nillth — and to @Audacity88 and @singlerider for the corroborating re-reviews. Approving.

PR #7668, opened 2026-06-15
fix(memory): drop postgres client on OS thread to avoid nested runtime panic

dvgamerr:
## Problem

When `zeroclaw agent -a <alias>` exits after a conversation turn, the Tokio runtime drops `PostgresMemory` as part of cleanup. The `postgres` crate's `Client::drop` implementation sends a clean-shutdown message to the server by calling `Runtime::block_on` internally. Because `Drop` runs inside the existing Tokio runtime context, this triggers:

```
thread 'main' panicked at postgres-0.19.13/src/connection.rs:66:
Cannot start a runtime from within a runtime. This happens because a function
(like `block_on`) attempted to block the current thread while the thread is
being used to drive asynchronous tasks.
```

The full backtrace confirmed the panic originates in `<postgres::client::Client as Drop>::drop` → `Client::close_inner` → `Connection::poll_block_on`, reached via:

```
zeroclaw_runtime::agent::loop_::run   (end-of-turn cleanup)
  → drop(Arc<dyn Memory>)
    → drop(AgentScopedMemory)
      → drop(PostgresMemory)
        → drop(Arc<Mutex<postgres::Client>>)
          → postgres::Client::drop  ← panic here
```

The existing `run_on_os_thread` helper correctly handles *async* operations by executing every SQL call on a plain OS thread (avoiding the same nested-runtime issue for queries). The `Drop` path was not covered.

## Fix

Introduce a `DropOnThread<T>` wrapper whose `Drop` implementation moves the inner value to a freshly spawned OS thread before releasing it:

```rust
struct DropOnThread<T: Send + 'static>(Option<T>);

impl<T: Send + 'static> Drop for DropOnThread<T> {
    fn drop(&mut self) {
        if let Some(value) = self.0.take() {
            std::thread::Builder::new()
                .name("postgres-client-drop".to_string())
                .spawn(move || drop(value))
                .ok();
        }
    }
}
```

`PostgresMemory.client` is changed from `Arc<Mutex<Client>>` to `DropOnThread<Arc<Mutex<Client>>>`. All async methods that previously called `self.client.clone()` now call `self.client.get().clone()` to reach through the wrapper.

**Why this works:** A plain OS thread spawned with `std::thread::Builder` has no Tokio runtime context. `postgres::Client::drop` can call `Runtime::block_on` freely on such a thread. By the time `PostgresMemory` drops, all `run_on_os_thread` futures have completed (they are awaited inline), so the Arc's strong count is 1. The background thread becomes the sole owner and the client closes cleanly.

## Test plan

- [x] `cargo build --features memory-postgres` compiles without errors
- [x] `zeroclaw agent -a <alias> -m "hello"` returns a response and exits cleanly — no panic
- [x] Existing unit tests in `zeroclaw-memory` pass (`cargo test -p zeroclaw-memory --features memory-postgres`)

### DIFF
--- a/crates/zeroclaw-memory/src/postgres.rs
+++ b/crates/zeroclaw-memory/src/postgres.rs
@@ -43,11 +43,26 @@ impl<T: Send + 'static> DropOnThread<T> {
 
 impl<T: Send + 'static> Drop for DropOnThread<T> {
     fn drop(&mut self) {
-        if let Some(value) = self.0.take() {
-            std::thread::Builder::new()
-                .name("postgres-client-drop".to_string())
-                .spawn(move || drop(value))
-                .ok();
+        let Some(value) = self.0.take() else { return };
+        // Wrap in ManuallyDrop so the value is NOT dropped on the current
+        // thread if spawn fails — ManuallyDrop's own Drop is a no-op.
+        let slot = std::mem::ManuallyDrop::new(value);
+        if std::thread::Builder::new()
+            .name("postgres-client-drop".to_string())
+            .spawn(move || drop(std::mem::ManuallyDrop::into_inner(slot)))
+            .is_err()
+        {
+            // The OS refused to spawn a thread. Intentionally leak the value
+            // rather than drop it here: postgres::Client::drop calls
+            // Runtime::block_on, which panics on a Tokio runtime thread.
+            // A controlled leak is preferable to an unrecoverable panic.
+            ::zeroclaw_log::record!(
+                WARN,
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                    .with_outcome(::zeroclaw_log::EventOutcome::Unknown),
+                "postgres-client-drop thread spawn failed; leaking client to avoid nested-runtime panic"
+            );
+            // `slot` is ManuallyDrop — T is intentionally not dropped.
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Audacity88 (changes requested):
Context: I reviewed current head 5f8745d against the PR body, visible GitHub checks, the one-file `crates/zeroclaw-memory/src/postgres.rs` diff, the existing Postgres memory tests, and the `run_on_os_thread` pattern this code extends. I did not run local Postgres validation; the review below is based on the public diff, existing source, and the validation claimed in the PR body.

The direction is right: the sync `postgres` client should not be dropped on a Tokio runtime thread if its own `Drop` path can call `Runtime::block_on`. I do not think this is ready to merge yet, though, because the PR does not add a regression for the shutdown path it is fixing, and the new helper leaves its spawn-failure behavior implicit.

### 🟢 What looks good — the fix targets the real runtime boundary

Moving the final `Arc<Mutex<Client>>` release out of `PostgresMemory::drop` is the right shape for this specific panic. The existing SQL operations already use plain OS threads through `run_on_os_thread`; this patch applies the same boundary idea to the destructor path that was previously missed. The rest of the memory methods still borrow the same client through the wrapper, so the patch stays narrow and does not change the query/update semantics.

### 🔴 Blocking — add a regression test for the nested-runtime drop path

The diff changes production shutdown behavior but adds no test. The existing `new_does_not_panic_inside_tokio_runtime` test only covers constructor failure against an unreachable endpoint; it does not create a successful `PostgresMemory`, drop it inside a Tokio runtime, or exercise the new `DropOnThread` wrapper at all. That means the exact bug can regress later while the current unit tests still pass.

Please add a focused regression that fails on the old behavior and passes with this patch. A live Postgres integration test for `PostgresMemory` drop would be ideal if the test harness can support it. If not, a small unit test around `DropOnThread` with a test-only drop guard is enough: construct the wrapper inside a Tokio runtime, drop it, and assert the guard's `Drop` ran on a plain OS thread rather than a thread with an active Tokio runtime. That pins the contract this PR relies on without needing a real database in CI.

### 🟡 Warning — make the spawn-failure path explicit

`DropOnThread::drop` currently calls `Builder::spawn(...).ok()` and discards the result. The safety contract in the comment says the final drop always happens on a plain OS thread, but the code does not define what happens if the OS thread cannot be spawned. Because this helper exists specifically to keep `postgres::Client::drop` away from the runtime thread, the failure path should be deliberate rather than silent.

Please handle that case explicitly. For example, log the spawn failure and choose a fail-closed behavior that does not accidentally drop the client on the current Tokio runtime thread. If the intended fallback is to leak the client rather than risk the nested-runtime panic, make that visible in code and in a test or comment. If you believe `Builder::spawn` cannot drop the captured value on the caller thread when spawning fails, please document that invariant with the source you are relying on.

Audacity88 (approved):
Context: I re-reviewed current head 48a890d against the updated PR body, visible GitHub checks, the one-file `crates/zeroclaw-memory/src/postgres.rs` diff, the prior Audacity88 review, and the existing `run_on_os_thread` pattern this code extends. I did not run local Postgres validation; this review is based on the public diff, green CI, and the validation claimed in the PR body.

Thanks for the update. The follow-up commit addresses the block from my previous review, and this is ready from my side.

### ✅ Resolved — Drop path regression is now pinned

The new `drop_on_thread_drops_value_on_plain_os_thread` test drops a guard through `DropOnThread` from inside a current-thread Tokio runtime, then waits for the guard to report that its destructor ran without an active Tokio handle. That covers the helper contract I asked for without requiring a live PostgreSQL instance in CI.

### ✅ Resolved — Spawn failure now has an explicit fail-closed path

`DropOnThread::drop` now moves the value into `ManuallyDrop` before spawning the cleanup thread, logs a warning if thread creation fails, and deliberately leaks instead of falling back to dropping the PostgreSQL client on the runtime thread. That makes the rare failure mode explicit and preserves the safety property this helper exists to enforce.

### 🟢 What looks good — the patch stays on the runtime boundary

The production change keeps ordinary SQL operations going through the existing `run_on_os_thread` boundary and only wraps the final client release. It does not add duplicate client state or change the query/update semantics, so the fix stays focused on the destructor path that was panicking.

One small public-record cleanup remains: the PR body still has the generated-tool footer, and its `DropOnThread` sample still shows the earlier `.spawn(...).ok()` version rather than the current explicit leak-on-spawn-failure path. Please refresh those before merge. That does not change my code verdict.

PR #7670, opened 2026-06-15
test(runtime): cover Windows shell code page decoding

### LINKED ISSUE #6859: [Feature]: Add behavioral test coverage for Windows shell code-page decoding
### Summary

Add behavioral test coverage for Windows shell output decoding so non-UTF-8 console code pages are tested, not only compiled.

### Problem statement

PR #6772 fixes #6704 by decoding Windows shell stdout and stderr from the active console output code page before returning UTF-8 strings. Current CI builds the Windows target, but the normal test job runs on Linux. That means the new Windows-only behavior is compiled and reviewed, but not exercised by CI.

Without behavioral coverage, future changes to the shell tool, decoder mapping, or Windows dependency features could regress CJK/non-UTF-8 shell output without a test failure.

### Proposed solution

Add a focused test path for the decoder behavior. Either approach is acceptable:

- add a Windows CI test slice that runs the shell decoding tests on a Windows runner; or
- split the decoder so the code-page-to-decoding path can be tested with an explicit code page on every platform, while keeping the runtime `GetConsoleOutputCP()` / `GetACP()` lookup Windows-only.

The important outcome is that representative non-UTF-8 bytes, such as GBK/CP936 output for Simplified Chinese text, are asserted to become the expected UTF-8 string.

### Non-goals / out of scope

This follow-up should not redesign the shell tool, change command construction, add new user-facing configuration, or broaden Windows process handling beyond output decoding.

### Alternatives considered

Relying on the Windows build job alone verifies imports and feature flags, but it does not prove the decoder returns the expected text. Manual review is useful, but this class of locale bug should have a regression test.

### Acceptance criteria

- CI exercises at least one non-UTF-8 Windows shell decoding case behaviorally.
- Coverage includes a representative CJK code page, preferably CP936/GBK.
- The test proves invalid or non-UTF-8 bytes are decoded to the intended UTF-8 string instead of becoming replacement characters.
- The production runtime path still uses the active Windows console output code page.
- No new user-facing configuration or API surface is introduced solely for testing.

### Architecture impact

Affected areas: `crates/zeroclaw-runtime/src/tools/shell.rs`, Windows CI/test configuration, and possibly a small internal decoder helper boundary.

### Risk and rollback

Risk is low if the change only adds tests or extracts a small internal helper. Rollback is reverting the test/helper extraction; runtime shell behavior should remain unchanged.

### Breaking change?

No

### Data hygiene checks

Alix-007:
## Summary
- Adds a private test-compilable helper for shell output decoding by explicit Windows code page.
- Runs CP936/GBK bytes through the same decoder path so Linux CI behaviorally verifies CJK output becomes UTF-8 instead of replacement characters.
- Keeps the production Windows path using `GetConsoleOutputCP()` with `GetACP()` fallback.

Fixes #6859

## Test plan
- [x] `cargo fmt --all -- --check` — passed
- [x] `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` — passed
- [x] `cargo test -p zeroclaw-runtime decode_output --lib` — passed


Audacity88:
This is the right shape for #6859: the runtime Windows path still gets the active console code page at runtime, while the explicit-code-page helper gives CI a behavioral CP936/GBK regression test.

Before approval, please refresh this branch over current `master`. This PR is still mergeable, but it is behind current `master` on a runtime shell surface, so the visible checks did not test the current merge result.

Please also clean the public validation text. The current body includes machine-specific details (`root@xx.xx.xx.xx`, `/media/...`) and “Independent reviewer PASS”. Please replace those with repo-level command/result evidence, for example `cargo test -p zeroclaw-runtime decode_output --lib` and the relevant pass result.

Separate maintainer cleanup: labels need alignment. Given the touched path, I expect `risk: high`; with 22 changed lines, this should be `size: XS`.

Alix-007:
Thanks — good calls on both. Refreshed over current master so the checks reflect the real merge result, and swapped the machine-specific build evidence for repo-level commands:
- `cargo fmt --all -- --check` — passed
- `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` — passed
- `cargo test -p zeroclaw-runtime decode_output --lib` — passed

Keeping build-host/internal details out of the PR body is a good habit to carry forward — appreciate the careful review.

singlerider (approved):
Reviewed at current head (singlerider). Test-only PR that makes the Windows shell code-page decoding path testable on non-Windows CI. CI is 18/18 green, MERGEABLE, and @Audacity88's three asks are all addressed (refreshed over master, body cleaned of machine-specific text, labels aligned to risk: high / size: XS). Approving.

### 🟢 What looks good: the helper split makes CP936 testable without faking the console

Splitting `decode_output_with_code_page(bytes, cp)` out of `decode_output` is the right seam: `decode_output` keeps the Windows-only `GetConsoleOutputCP`/`GetACP` runtime probe, while the pure transcoding helper takes an explicit code page so a test can pass `936` directly. That lets the GBK regression assert the real production path (`decode_output_with_code_page(gbk_bytes, 936) == "你好"` and no U+FFFD) instead of the prior test that only exercised `encoding_rs::GBK.decode` directly and so never touched the code under test. Gating the helper and the mapping fn with `#[cfg(any(target_os = "windows", test))]` is the correct way to compile them into test builds on every platform without shipping them in non-Windows release binaries.

### 🟢 What looks good: the dev-dependency is necessary, not redundant

`encoding_rs` is declared only under `[target.'cfg(windows)'.dependencies]`, so it is absent from non-Windows builds. Since the whole point is to run this transcoding test on Linux/macOS CI, adding `encoding_rs` to `[dev-dependencies]` is exactly what makes the helper and its test compile and link off-Windows. It does not become a production dependency on those targets. The four mapping tests (CJK code pages, UTF-8 variants, unknown-falls-back-to-UTF-8) lose their `#[cfg(target_os = "windows")]` gate and now run everywhere, which is pure coverage gain for a `risk: high` shell surface.

Audacity88 (approved):
I checked the current PR metadata/body, public diff, prior PR comments mentioning #6859, visible checks, and public source excerpts. I did not use local validation logs or private source context for this approval pass.

### 🟢 What looks good — CP936 coverage now exercises the decoder path

This now covers the CP936/GBK shell-output regression path described in the PR body and prior PR discussion without changing the runtime contract. The production Windows path still reads the active console output code page at runtime, while the new test-gated helper lets CI exercise the same mapping/decoding path with explicit CP936 bytes.

The focused GBK assertion checks both the expected UTF-8 text and the absence of replacement characters, which is the behavior #6859 needed to protect. The earlier public validation/body cleanup is also addressed, and the current head is green in CI.

Approved for this head.

PR #7672, opened 2026-06-15
fix(cli): preserve utf8 erase handling

### LINKED ISSUE #6995: [Bug]: Backspace in `zeroclaw agent` CLI deletes byte-by-byte instead of character-by-character (UTF-8 CJK chars need 3 backspaces)
### Affected component

unknown

### Severity

S2 - degraded behavior

## Bug Description
When using `zeroclaw agent` interactive CLI mode, pressing backspace deletes one byte at a time instead of one Unicode grapheme cluster. For CJK characters (3 bytes in UTF-8), 3 backspaces are needed to delete one character.

## Environment
- OS: Linux (Device0)
- Terminal: xterm-256color
- Locale: en_HK.UTF-8
- ZeroClaw version: [run `zeroclaw --version` to fill]

## Actual Behavior
- Chinese character "测" (3 bytes in UTF-8) requires 3 backspaces to delete
- The terminal shows `?` or corrupted bytes after first backspace

## Root Cause (suspected)
The line editor in `zeroclaw agent` likely processes bytes instead of grapheme clusters when handling backspace. Rust's default string handling is byte-indexed; proper Unicode handling requires using `unicode-segmentation` crate or similar.

## Suggested Fix
Use `unicode-segmentation` crate to split input into grapheme clusters, and delete one cluster per backspace press.

## Expected Behavior
Pressing backspace once should delete one character (one grapheme cluster), regardless of byte length.

### Steps to reproduce

```bash
## Reproduction Steps
1. Run `zeroclaw agent`
2. Type a Chinese character: 测
3. Press backspace once → only 1/3 of the character is deleted (shows as corrupted byte)
4. Press backspace 2 more times → character fully deleted
```

### Impact

_No response_

### Logs / stack traces

```text

```

### ZeroClaw version

zeroclaw 0.7.5

### Rust version

_No response_

### Operating system

Linux device0 6.12.90+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.90-1 (2026-05-22) x86_64 GNU/Linux

### Regression?

Unknown

### Pre-flight checks

Alix-007:
## Summary
- Enable the terminal `IUTF8` input flag while the CLI channel listens on Linux/Android
- Restore terminal attributes after the listener exits and no-op safely for non-TTY stdin
- Add a pseudoterminal regression test for setting and restoring `IUTF8`

Fixes #6995

## Test plan
- Worker implementation completed in `.worktrees/alix-6995-agent-backspace-fresh`
- Independent reviewer PASS
- Remote build machine `root@10.54.156.81`:
  - `cargo test -p zeroclaw-runtime cli_input --lib`


Audacity88 (approved):
I reviewed current head `0c40eb8e08ca` against #6995, the two-file CLI/runtime diff, the terminal-attribute guard, the pty regression test, and the latest green GitHub checks. I also checked the current `master` freshness state: the PR is behind by 35 commits, has no overlapping files with base changes, and applies cleanly by merge-tree analysis. I did not complete a local interactive CJK terminal smoke or a full exact-merge test run.

### 🟢 What looks good — IUTF8 fixes the cooked-terminal erase path

This is the right layer for the reported bug. `CliChannel::listen()` reads from the terminal in canonical/cooked mode through Tokio line reading, so Linux terminal erase behavior is decided by the line discipline before ZeroClaw receives the final line. Enabling `IUTF8` while the CLI listener is active matches the reported CJK backspace failure without adding a user-space line editor or a new Unicode dependency.

### 🟢 What looks good — terminal state is restored and non-interactive stdin stays safe

The guard stores the original termios only after a successful `tcsetattr`, restores it on drop, and no-ops when `tcgetattr` fails. That keeps piped/non-TTY stdin safe, avoids mutating platforms that do not expose this Linux/Android flag, and limits the terminal setting change to the lifetime of the CLI listener.

### 🟢 What looks good — regression coverage hits the OS boundary

The pty test clears `IUTF8`, calls the helper against the slave fd, verifies the bit is set, then verifies drop restores the original flag. That is the right level for this fix: it tests the actual termios boundary rather than only an internal boolean.

Approved from my side. Because the branch is stale and touches CLI/runtime surface, final merge confidence should still come from a branch refresh, merge-queue validation, or explicit stale-risk acceptance rather than treating old-head CI alone as the merge result.

singlerider (approved):
Read current head against `master`: the `ensure_terminal_utf8_erase` guard in `crates/zeroclaw-runtime/src/cli_input.rs`, the `CliChannel::listen` call site, and the pty regression test. @Audacity88's approval already covers the freshness/CI state; I am confirming the guard mechanics.

### 🟢 What looks good: RAII guard restores cleanly and fails closed

The `TerminalUtf8EraseGuard` is the right shape: it captures `original` only when `tcsetattr` actually succeeds, so a non-TTY stdin (where `tcgetattr` returns non-zero) yields `original: None` and the `Drop` is a no-op rather than restoring garbage. The early return when `IUTF8` is already set avoids clobbering a terminal that is already correct and means an unchanged terminal is never "restored" at drop. `Drop` swallows the `tcsetattr` result with a comment that explains *why* (drop must not panic), which is the correct call for best-effort line-discipline cleanup. Non-Linux/Android builds compile to an empty guard. The pty test exercises the real path end to end: clears `IUTF8` on a slave fd, asserts the guard sets it, then asserts restoration after the guard drops. Tight, platform-gated, and reversible via the documented revert.

PR #6986, opened 2026-05-28
fix(gateway/acp): wire shared CanvasStore into WS chat and ACP sessions

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `Agent::from_config_with_session_cwd_and_mcp_backchannel` was hard-coding `None` as `canvas_store` when calling `all_tools_with_runtime`, causing every web-UI WebSocket chat session and every ACP session to instantiate a private `CanvasStore::default()`. Frames pushed by the agent through the `canvas` tool were silently dropped because that private store is isolated from `AppState::canvas_store` -- the store that `/ws/canvas/:id` WebSocket subscribers read from.
  - #6221 (`61db5c8`) wired up the daemon-supervisor path (gateway <-> channels orchestrator) but left two paths unpatched: the WebSocket `/ws/chat` handler (`ws.rs`) and the ACP server (`acp_server.rs`).
  - `agent.rs`: added `canvas_store: Option<CanvasStore>` to `from_config_with_session_cwd_and_mcp_backchannel` (and the private `_approval_mode` variant); threads it through to `all_tools_with_runtime` instead of the hardcoded `None`.
  - `ws.rs`: passes `Some(state.canvas_store.clone())` so the WS chat agent shares the gateway `AppState` store.
  - `acp_server.rs`: adds `canvas_store: Option<CanvasStore>` field + `with_canvas_store(CanvasStore) -> Self` builder method; existing constructors default to `None` (no API break); all three `session/new` call sites pass `self.canvas_store.clone()`.
  - `acp.rs`: calls `.with_canvas_store(state.canvas_store.clone())` when building the ACP server for the gateway WebSocket `/acp` endpoint.
- **Scope boundary:** Does not change standalone `zeroclaw acp` behavior (still gets `None`/its own store, no gateway to subscribe to). Does not touch channels orchestrator wiring (already correct after #6221).
- **Blast radius:** High-risk runtime/gateway/channel-session wiring. The behavior change is intentionally limited to sharing the existing gateway `CanvasStore` with web UI chat and gateway-hosted ACP sessions; standalone ACP keeps the prior isolated-store behavior.
- **Linked issue(s):** Closes #6965
- **Labels:** `bug`, `risk: high`, `size: XS`, `agent`, `channel`, `gateway`, `runtime`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
$ cargo fmt --all -- --check
(no output -- clean)

[...]

PR #7678, opened 2026-06-15
fix(runtime): thread shared CanvasStore into WS chat and ACP agent sessions

### LINKED ISSUE #7563: [Bug]: canvas-store regression in WS chat/ACP sessions breaks /canvas after #6986
Title: [Bug]: canvas-store regression in WS chat/ACP sessions breaks /canvas after #6986

### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

The Web UI `/canvas` page stays empty after a normal WebSocket chat session (`/ws/chat`) uses the `canvas` tool, even though the tool call itself reports success.

This is a regression of the bug previously fixed by:
- issue `#6965`
- PR `#6986`
- commit `35624dbf95e9a0d1d4a5eb4fb6cf4a015799aa18`

I updated a local checkout from upstream `master`, redeployed, and the problem still persists.

Current upstream `master` as of **2026-06-13** at commit:
- `0eb90eaaf640ef9e556ce2107f22f71cd02c2a17`

still appears to route WS chat sessions (and ACP session agent construction) through a code path that does **not** pass the shared gateway `CanvasStore` into agent tool construction. As a result, the `canvas` tool writes to a private default store while `/canvas` reads from `AppState.canvas_store`.

### Expected behavior

Frames pushed by agents from Web UI chat sessions should appear on `/canvas` in real time, using the shared gateway `CanvasStore`, as intended by `#6986`.

ACP sessions should also not regress relative to the same shared-store wiring where applicable.

### Steps to reproduce

```bash
1. Update a local checkout from upstream master and redeploy.
2. Start ZeroClaw in daemon/gateway mode.
3. Open the Web UI `/canvas` page and confirm it is connected and waiting on canvas `default`.
4. Open a normal Web UI chat session (`/ws/chat`) and ask the agent to use the `canvas` tool to render simple HTML to the default canvas.
5. Observe:
   - tool call reports success, e.g. `Rendered html content to canvas 'default' (frame: <uuid>)`
   - `/canvas` remains empty / waiting for content

Comparison check:
6. POST directly to `/api/canvas/default` with the same bearer token.
7. Observe `/canvas` updates immediately.
```

### Impact

- Affected users: anyone using the `canvas` tool from Web UI chat sessions
- Frequency: reproducible
- Consequence: Live Canvas appears broken from chat even though tool calls claim success

### Logs / stack traces

Observed successful tool result from chat path:

```text
Rendered html content to canvas 'default' (frame: <uuid>)
```

No obvious runtime error is emitted; frames are effectively written to the wrong store.

Redacted `zeroclaw doctor` output from the same environment:

```text
🩺 ZeroClaw Doctor (enhanced)

  [config]
    ✅ config file loaded
    ✅ model_provider configured
    ✅ gateway port configured
  [workspace]
    ✅ directory exists and is writable
  [daemon]
    ✅ heartbeat fresh
    ✅ scheduler healthy
  [environment]
    ✅ git present
    ✅ shell present
    ✅ home directory env set
    ✅ curl present
  [cli-tools]
    ✅ cargo present
    ✅ rustc present

  Summary: no environment-specific failure that explains the canvas behavior.
```

### ZeroClaw version

zeroclaw 0.8.0

### Rust version

### Operating system

Linux 6.17.0-35-generic x86_64 GNU/Linux

### Regression?

Yes, it worked before

### Pre-flight checks

### Notes for maintainer triage

This looks like a regression relative to `#6986`, not a new unrelated bug.

Relevant upstream history:
- Original fix commit: `35624dbf95e9a0d1d4a5eb4fb6cf4a015799aa18`
- Current `master` as of 2026-06-13 still contains the older broken call shape in `ws.rs` and `agent.rs`
- `git blame` on current upstream `master` points to commit `2c9147814f` (`feat(integration): introduce zerocode TUI, RPC socket transport, DenyWithEdit approval, and beta-2 integration (#6848)`) as the commit that appears to have reintroduced the broken `None` / missing shared-store path in these lines

Current upstream source evidence:

- WS chat still calls `from_config_with_session_cwd_and_mcp_backchannel(..., true, false)` with no shared `CanvasStore` argument:
  - <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-gateway/src/ws.rs#L416-L422>
- `from_config_with_session_cwd_and_mcp_backchannel` still has no `canvas_store` parameter and forwards `None` into the lower constructor path:
  - <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-runtime/src/agent/agent.rs#L1060-L1075>
- Agent tool construction still passes `None` into `all_tools_with_runtime(..., None, ...)`:
  - <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-runtime/src/agent/agent.rs#L1203-L1220>
- Meanwhile, ACP/gateway still retains the later shared-canvas plumbing surfaces, which makes the partial regression especially confusing:
  - gateway ACP builder still has `with_canvas_store(...)`:
    <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-gateway/src/acp.rs#L79-L88>
  - ACP server still has `canvas_store` field and builder:
    <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-channels/src/orchestrator/acp_server.rs#L39-L41>
    <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-channels/src/orchestrator/acp_server.rs#L110-L110>
    <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-channels/src/orchestrator/acp_server.rs#L165-L169>
- But ACP session agent creation calls currently also appear to use the old two-bool signature rather than passing the shared store:
  - <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-channels/src/orchestrator/acp_server.rs#L558-L564>
  - <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-channels/src/orchestrator/acp_server.rs#L768-L774>
  - <https://github.com/zeroclaw-labs/zeroclaw/blob/0eb90eaaf640ef9e556ce2107f22f71cd02c2a17/crates/zeroclaw-channels/src/orchestrator/acp_server.rs#L967-L973>

This suggests the original fix was later partially lost during integration refactoring, producing a real regression on current `master`.

singlerider:
## Summary

WS chat (`/ws/chat`) and ACP session agents were built through `from_config_with_session_cwd_and_mcp_backchannel`, which forwarded `None` for the `canvas_store` argument to `all_tools_with_runtime`. The `canvas` tool therefore wrote frames to a private default store while `/canvas` reads from the gateway's shared `AppState.canvas_store` — so canvas renders from chat reported success but never appeared on `/canvas`.

This regressed the fix from #6986: later integration refactoring (#6848) dropped the shared-store wiring on these construction paths while the ACP server retained its own `canvas_store` field and builder, producing the confusing partial regression described in #7563.

Fix threads `Option<CanvasStore>` through the backchannel constructor down to the tool builder:

- `crates/zeroclaw-gateway/src/ws.rs` passes `AppState.canvas_store` (Arc-backed clone), so frames land in the same store `/canvas` subscribes to.
- `crates/zeroclaw-channels/src/orchestrator/acp_server.rs` (3 session-construction sites) passes its existing `self.canvas_store`, which the gateway ACP builder already populates via `with_canvas_store` (`crates/zeroclaw-gateway/src/acp.rs`).
- The non-backchannel constructors forward `None` unchanged.

Closes #7563

`cargo fmt --all -- --check`
```
(clean, exit 0)
```

`cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway -p zeroclaw-channels -p zeroclaw-tools --all-targets -- -D warnings`
```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 27s
(0 warnings, 0 errors)
```

`cargo test -p zeroclaw-tools canvas`
```
test canvas::tests::canvas_tool_renders_into_shared_store ... ok
test result: ok. 21 passed; 0 failed; 0 ignored; 0 measured; 1314 filtered out
```

New regression test `canvas_tool_renders_into_shared_store` asserts a `CanvasTool` built from a cloned shared store renders frames visible through the original handle — the exact contract the WS/ACP wiring depends on.

None. Change is internal tool wiring: it routes canvas frames to the already-authenticated gateway store the `/canvas` endpoint serves. No new surface, no new data captured. Canvas content was already user-authored render payloads.

## Compatibility (required)

Backward compatible. Public constructor `from_config_with_session_cwd_and_mcp_backchannel` gains a trailing `canvas_store: Option<CanvasStore>` parameter; all in-tree callers updated. Non-backchannel constructors are unchanged in behavior (forward `None`). No config or schema changes.

`git revert <sha>`. Single self-contained commit; no migrations or persisted state involved.


singlerider (commented):
Reviewed as the author (self-review, noted for the audit trail). Traced the full construction chain against local source on master rather than trusting the diff in isolation.

### 🟢 Root cause and fix are correct and verified end to end

I confirmed every claim in the description against the tree:

- All four callers of `from_config_with_session_cwd_and_mcp_backchannel` are updated: the three ACP sites in `acp_server.rs` (561, 772, 972) and the WS site in `ws.rs:419`. The two runtime convenience wrappers (`agent.rs:947`, `agent.rs:999`) correctly forward `None`.
- The ACP path is genuinely wired: `acp.rs:82-88` populates `AcpServer.canvas_store` via `.with_canvas_store(state.canvas_store.clone())` on both branches, so `self.canvas_store` at the three session sites is the shared gateway store, not the default.
- WS passes `state.canvas_store.clone()`, the same store `/canvas` subscribes to.

The regression test asserts the actual contract (a clone of the shared store observes frames pushed through the tool), which is exactly the invariant the wiring depends on.

### 🟢 Scope boundary is correct

The standalone `zeroclaw acp` CLI path (`src/main.rs:3329`) constructs `AcpServer` without `with_canvas_store`, so its sites pass `None`. That is correct: stdio ACP has no `/canvas` HTTP reader, so there is nothing for frames to reach. The fix is scoped precisely to the gateway WS and gateway-ACP paths that #7563 describes, and does not over-reach.

### 🟢 Risk labeling and rollback are honest

`risk: high` is appropriate for a public constructor signature change threaded through four call sites across three crates. Rollback is a clean single-commit revert with no persisted state. fmt/clippy/test evidence matches a local re-run.

Approving. Mark Ready and merge when the milestone is being assembled.

Audacity88:
Thanks for the tight fix here. I checked the current head against live `master`, and GitHub is right that this branch is now conflicted.

`master` has since added SOP engine/audit handles through the same WS/ACP agent-construction path, so the refresh needs to preserve that wiring while adding the shared `CanvasStore` argument into the same calls. The conflicts are in:

- `crates/zeroclaw-runtime/src/agent/agent.rs`
- `crates/zeroclaw-gateway/src/ws.rs`
- `crates/zeroclaw-channels/src/orchestrator/acp_server.rs`

Please refresh the branch against current `master` and rerun the existing validation on the refreshed head. Once the conflict is resolved, I can do the technical approval pass on the reconciled diff.

WareWolf-MoonWall (commented):
The fix is architecturally correct — the `CanvasStore` created at gateway startup should flow through to every agent session that runs within that gateway context, and the previous `None` default was producing sessions that couldn't see the shared canvas. However, this PR is currently blocked from merge and needs attention before a technical review is meaningful.

### 🔴 Blocking — Branch is dirty (conflicted with master)

The branch conflicts with master in `agent.rs`, `ws.rs`, and `acp_server.rs`. Audacity88 noted the root cause: the SOP engine and audit-handle wiring that landed in master after this branch was cut touches the same agent construction path. The conflict must be resolved cleanly — the rebase must preserve both the SOP/audit wiring from master and the new `canvas_store` parameter threading from this branch.

This is a `priority:p1` fix and I want it to land quickly. Please rebase, validate the three conflict sites explicitly, and re-run `cargo test -p zeroclaw-runtime -p zeroclaw-channels -p zeroclaw-gateway` against the fresh head.

### 🟡 Warning — Self-approval doesn't substitute for independent review

@singlerider's review here is a self-approval. By project governance it doesn't count toward the independent reviewer requirement. Audacity88 has explicitly deferred their technical pass until after the rebase. Once the branch is clean and CI is green on the rebased head, please re-request review from Audacity88.

### 🟢 What looks good — Correct shared-ownership pattern

Passing `state.canvas_store.clone()` through to agent construction is the right move — it passes a cloned `Arc` handle, not a deep copy of state. Both the tool and the gateway reader hold a reference to the same underlying allocation. No data race concern, no new `Mutex`/`RwLock` introduced.

### 🟢 What looks good — SSOT fix, not a SSOT violation

The previous code was creating a *private* default `CanvasStore` per session, which meant gateway canvas API responses and the tool's internal store were separate objects. This PR consolidates to the canonical `AppState.canvas_store` — that's removing duplicate state, not introducing it.

### 🟢 What looks good — Regression test

`canvas_tool_renders_into_shared_store` exercises the `Arc` clone contract: tool writes through one handle, reader observes through the other. That directly covers the class of bug this PR fixes.

tidux (commented):
Tracing the construction chain from `AppState.canvas_store` down to `CanvasTool::new` against current master and the four documented call sites. Read everything on the PR before writing.

### 🟢 What looks good — Root cause and fix shape are correct

The diagnosis matches the tree. `crates/zeroclaw-gateway/src/acp.rs:79-88` populates `AcpServer.canvas_store` via `.with_canvas_store(state.canvas_store.clone())` on both `new_with_writer_and_store` and `new_with_writer` paths, so at the three ACP session-construction sites in `acp_server.rs` (558, 769, 969 on the PR head) `self.canvas_store` is genuinely the shared gateway store — not a private default. The WS site at `ws.rs:416` passes `state.canvas_store.clone()` directly. The previous `None` argument really was the source of the partial regression in #7563, and threading `Option<CanvasStore>` through `from_config_with_session_cwd_and_mcp_backchannel` → `..._approval_mode` → `all_tools_with_runtime` is the right scoped fix.

### 🟢 What looks good — Scope discipline

Confirmed `src/main.rs:3328-3335` constructs `AcpServer::new_with_store` / `AcpServer::new` *without* `with_canvas_store`, so the standalone `zeroclaw acp` CLI keeps `self.canvas_store = None` and the three ACP sites pass `None` through. Correct — stdio ACP has no `/canvas` HTTP reader to feed. The two non-backchannel convenience constructors (`from_config_with_session_cwd_and_mcp` and `from_config_with_tui_env`) also forward `None`, which is right for the zerocode/TUI path that doesn't run inside a gateway with a shared store. The fix touches exactly the WS + gateway-ACP entry points the issue describes.

### 🔴 Blocking — Branch is conflicted with current master

@WareWolf-MoonWall already named this and @Audacity88 explicitly deferred their technical pass until after the rebase. I want to ground-truth it for the record so the author can see both the conflict shape and what's load-bearing on the rebase:

`git merge-tree --write-tree pr-7678 upstream/master` produces three content conflicts:

- `crates/zeroclaw-runtime/src/agent/agent.rs`
- `crates/zeroclaw-gateway/src/ws.rs`
- `crates/zeroclaw-channels/src/orchestrator/acp_server.rs`

The conflicting master change is 475238513 (`fix(runtime): unify SopEngine construction — single instance per daemon` / #7546). That PR threaded `sop_engine: Option<Arc<std::sync::Mutex<SopEngine>>>` and `sop_audit: Option<Arc<SopAuditLogger>>` into the same `from_config_with_session_cwd_and_mcp_backchannel`, `from_config_with_tui_env`, and `from_config_with_session_cwd_and_mcp_approval_mode` signatures this PR is editing, and into the same WS / ACP call sites. Both changes are pure trailing-arg additions in non-overlapping logical territory; the resolution is a union — keep SOP/audit and add `canvas_store` — but the merge driver can't compose those automatically because the textual conflict regions sit on top of each other.

What the rebased head should look like at each site:

- `from_config_with_session_cwd_and_mcp_backchannel` signature: `(config, agent_alias, session_cwd, initialize_mcp, exclude_memory, sop_engine, sop_audit, canvas_store)` — 8 args.
- `from_config_with_tui_env` signature: `(config, agent_alias, session_cwd, initialize_mcp, exclude_memory, tui_env, sop_engine, sop_audit)` — 8 args. (Canvas not needed here; the existing `agent.rs` non-backchannel wrappers already pass `None` for canvas through the inner call.)
- `from_config_with_session_cwd_and_mcp_approval_mode` signature: `(config, agent_alias, session_cwd, initialize_mcp, approval_backchannel, exclude_memory, tui_env, sop_engine, sop_audit, canvas_store)` — 10 args.
- `ws.rs:416` call: pass `state.sop_engine.clone()`, `state.sop_audit.clone()`, `Some(state.canvas_store.clone())` as the trailing three.
- All three `acp_server.rs` call sites: pass `self.sop_engine.clone()`, `self.sop_audit.clone()`, `self.canvas_store.clone()` as the trailing three.
- The two callers of `from_config_with_tui_env` in `crates/zeroclaw-runtime/src/rpc/dispatch.rs:753` and `1150` already pass the SOP handles in master; no edit needed there.
- The non-backchannel `from_config_with_session_cwd_and_mcp` wrapper that forwards into `..._approval_mode` needs the trailing `None` count bumped to match the new 10-arg shape (sop_engine None + sop_audit None + canvas_store None — three trailing `None`s after `tui_env`).

After the rebase please re-run the validation battery on the fresh head: `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway -p zeroclaw-channels -p zeroclaw-tools --all-targets -- -D warnings`, `cargo test -p zeroclaw-tools canvas`, and (per Wolf's request) `cargo test -p zeroclaw-runtime -p zeroclaw-channels -p zeroclaw-gateway` so the SOP-engine wiring stays covered alongside the canvas threading.

### 🟡 Warning — Regression test covers the contract but not the regression path

`canvas_tool_renders_into_shared_store` correctly pins the Arc-handle invariant that the wiring depends on: a clone given to the tool and a separately-held handle observe the same frames. That's the right invariant. What it does not pin is the construction-path bug this PR is actually fixing — i.e. that `from_config_with_session_cwd_and_mcp_backchannel` (or whichever entry point downstream callers go through) actually forwards a `Some(store)` into `all_tools_with_runtime` and not `None`. The bug class #7563 caught is "the constructor receives a real store but threads `None` to the tool builder anyway", and a future refactor that re-introduces a hardcoded `None` mid-chain would not fail this test.

A construction-level test on the runtime crate that builds an `Agent` via `from_config_with_session_cwd_and_mcp_backchannel(..., Some(shared_store))`, finds the `CanvasTool` in the resulting dispatcher, drives a render through it, and asserts the original `shared_store.snapshot(...)` sees the frame would close the gap. It's still a unit-level assertion — no gateway/HTTP — but it covers the seam that actually regressed.

Non-blocking. If you'd rather not add it in this PR, an issue tagged "follow-up" is fine; the wiring is straightforward enough that I'm comfortable trusting the call-site audit on its own here.

### 🟢 What looks good — Arc-handle threading and SSOT consolidation

`Option<CanvasStore>` flowing through as a cheap Arc handle (`#[derive(Clone)]` on `CanvasStore` wrapping `Arc<RwLock<HashMap<...>>>`) is the right shape — one allocation, every clone shares it, no `Mutex`/`RwLock` introduced in this PR. The SSOT framing in Wolf's review is correct: pre-PR the WS/ACP-gateway sessions were silently constructing a *private* default store per agent on the backchannel path, so the apparent "canvas already exists" state was actually two separate objects under the same name. This PR collapses them back to the canonical `AppState.canvas_store` that #6986 first wired and #6848 inadvertently bypassed for this path.

### 🟢 What looks good — Backward-compat shape on the non-backchannel paths

The non-backchannel wrappers (`from_config_with_session_cwd_and_mcp`, `from_config_with_tui_env`) forward `None` for `canvas_store`, and the inner approval-mode constructor passes whatever it receives straight to `all_tools_with_runtime`. So the only externally-visible behaviour change is "the WS chat and gateway-ACP paths now see frames on `/canvas`" — every other caller behaves exactly as before, with the same private default store that already worked everywhere it was reachable. The trailing-arg signature growth is the right ergonomic call here — a builder pattern would be cleaner but is out of scope for a `risk: high` p1 fix.

---

Validation on the PR head as authored (e671ac78, prior to the rebase): `cargo fmt --all -- --check` clean; `cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway -p zeroclaw-channels -p zeroclaw-tools --all-targets -- -D warnings` clean; `cargo test -p zeroclaw-tools canvas` 21 passed including the new `canvas_tool_renders_into_shared_store`. Validation evidence in the PR body reproduces. I did *not* attempt to rebase locally and re-validate — that's the author's pass to make, and a reviewer claim of "I verified the rebased head" against code that doesn't exist on the PR would be making up evidence.

Marking `--comment`. The 🔴 is the same rebase block @WareWolf-MoonWall named; I do not have anything new to block on of my own. Once the branch is back on top of current master with the SOP/audit + canvas wiring composed and the cargo battery rerun on the fresh head, ping me and I'm happy to convert to `--approve` on the next pass. @Audacity88 is also waiting on the rebase for the same reason.

Nillth (commented):
I took stock of the existing discussion before writing: @singlerider's author self-review (a careful self-audit, noted for the trail but not counting toward the independent-reviewer requirement), @WareWolf-MoonWall's review naming the rebase block and the self-approval caveat, and @tidux's thorough construction-chain trace with the rebase ground-truthed via `git merge-tree` plus a substantive note on the regression test's coverage. @Audacity88 has explicitly deferred their technical pass until after the rebase. I cross-checked the diff against the PR head e671ac78026616ad35215e797912ce6019c2dd33 (not my local clone, which is well behind master), tracing the full chain from `AppState.canvas_store` down to `CanvasTool::new`. My read agrees with the three reviewers: the fix is substantively correct, and the only thing standing between it and merge is the conflict with master. I'm posting this as a comment — I have nothing new to block on, and the active block is the rebase everyone else already named.

### 🟢 What looks good — Single-source-of-truth is satisfied; this consolidates, it does not duplicate

This is the heart of the review for a `risk: high` shared-state seam, so I verified it end to end against the head rather than trusting the summary. `CanvasStore` is `#[derive(Clone)]` over `inner: Arc<RwLock<HashMap<String, CanvasEntry>>>` (crates/zeroclaw-tools/src/canvas.rs:53-56). Every clone shares one allocation — there is exactly one canonical store. That store is constructed once per gateway as `AppState.canvas_store` (crates/zeroclaw-gateway/src/lib.rs:499, built from `canvas_store.unwrap_or_default()` at lib.rs:753 and threaded into `AppState` at lib.rs:1469), and the `/canvas` REST/SSE handlers read, write, and subscribe exclusively through `state.canvas_store` (crates/zeroclaw-gateway/src/canvas.rs:37,51,75,125,219,232). What this PR threads into the WS and gateway-ACP agent sessions is a clone of *that* handle — so the tool and the `/canvas` reader observe the same frames. Per AGENTS.md "ABSOLUTE RULE — SINGLE SOURCE OF TRUTH," this is squarely on the right side of the line: it is the resolver/handle-threading pattern ("`&Config` / `Arc` handles threaded through call sites"), not a cached duplicate. As @WareWolf-MoonWall and @tidux both noted, the *pre-PR* state was the SSOT violation — the backchannel path silently built a private default store per session via the hardcoded `None`, so the canvas the agent wrote to and the canvas `/canvas` served were two different objects under the same name. This PR collapses them back to canonical. That's exactly the drift-bug class AGENTS.md describes, and removing it is the right call.

### 🟢 What looks good — The construction chain is intact and the fix is surgical

I traced the threading the PR claims, against the head. The trailing `Option<CanvasStore>` flows `from_config_with_session_cwd_and_mcp_backchannel` → `from_config_with_session_cwd_and_mcp_approval_mode` → `all_tools_with_runtime` (crates/zeroclaw-runtime/src/tools/mod.rs:493) → `CanvasTool::new(canvas_store.unwrap_or_default())` (mod.rs:594). The single line that actually fixed #7563 is the diff's `-None` / `+canvas_store` at crates/zeroclaw-runtime/src/agent/agent.rs:1121 — the prior hardcoded `None` at that call site is precisely what forced the gateway paths into a private default store. The `unwrap_or_default()` at the leaf is the correct mechanism, not a smell: when `Some(shared)` arrives (WS at ws.rs and the three ACP sites), the shared handle is kept; when `None` arrives (TUI/zerocode, standalone `zeroclaw acp`), a fresh default is built — which is correct, because those paths have no `/canvas` reader to feed. I confirmed the gateway-ACP wiring is genuine: crates/zeroclaw-gateway/src/acp.rs:79-88 calls `.with_canvas_store(state.canvas_store.clone())` on *both* `AcpServer` builder branches, and `AcpServer.canvas_store` (crates/zeroclaw-channels/src/orchestrator/acp_server.rs:110, default `None` at 161) is then passed at the three session sites (564, 775, 975). The `.unwrap_or_default()` calls elsewhere in acp_server.rs are on unrelated fields — no private canvas store is re-created mid-chain. The scope discipline @singlerider and @tidux praised holds up: standalone `zeroclaw acp` constructs `AcpServer` without `with_canvas_store`, so it correctly threads `None`.

### 🟢 What looks good — Architecturally aligned with the gateway-as-shared-host model

The change respects FND-001 §4.5 (the gateway separation) and §4.2 (dependencies flow inward). The gateway owns the canvas store as part of its `AppState`; the runtime takes it as an opaque `Option<CanvasStore>` parameter and does not reach back into gateway types — the runtime still knows nothing about the gateway. Passing the store *down* as an argument is the correct direction across that seam. Same reasoning the description gives for why this is internal-wiring-only with no new surface: I agree the Security & Privacy "None" is accurate — frames already landed in an authenticated, already-served store; this just stops a subset of sessions from writing to the wrong one.

### 🟡 Warning — Regression test pins the Arc contract but not the construction-path bug (reinforcing @tidux)

I want to second @tidux's point here rather than re-raise it fresh, because it's the one genuinely-open substantive item and it's worth the author seeing it from a second reviewer. `canvas_tool_renders_into_shared_store` (crates/zeroclaw-tools/src/canvas.rs) asserts the right *invariant* — a clone handed to the tool and a separately-held handle observe the same frame — but it builds the `CanvasTool` directly. The bug #7563 actually caught was not "Arc clones don't share state"; it was "the constructor receives a real store but threads `None` to the tool builder anyway" at the `all_tools_with_runtime` call site. A future refactor that re-introduces a hardcoded `None` mid-chain (exactly what #6848 did) would still pass this test green. Per AGENTS.md "Anti-Patterns" and FND-005 §4 (review discipline — "does this test verify behaviour, or that the code does what it currently does?"), the test that would actually guard this seam is a runtime-crate construction test: build an `Agent` via `from_config_with_session_cwd_and_mcp_backchannel(..., Some(shared))`, find the `CanvasTool` in the resulting dispatcher, drive a render, and assert `shared.snapshot(...)` sees it. @tidux already offered to accept a follow-up issue instead, and I agree that's a reasonable disposition given the call-site audit three of us have now independently done — but if it's deferred, FND-005 §5 (🟡 Conditional) is clear that it needs a tracked issue with a named owner, not just an intention. Not blocking on my end.

### 🔵 Suggestion — The trailing-arg growth is acceptable here, but note the collision pressure

`all_tools_with_runtime` already carries `#[allow(clippy::too_many_arguments)]`, and this PR plus the SOP PR are both adding trailing `Option<...>` parameters to the same constructor family. That's the right ergonomic call for a `risk: high` p1 fix — a builder refactor would be out of scope and higher risk — and I'm not asking for it here. But the fact that two in-flight PRs are growing the same signature and colliding on it (see below) is a mild signal that this constructor is becoming a maintenance hotspot. Worth a future builder-pattern issue for `from_config_with_session_cwd_and_mcp_approval_mode`; explicitly out of scope for this PR.

### 🔴 Blocking — Branch conflicts with master; rebase required before merge (already named by @WareWolf-MoonWall and @tidux)

This is the same block @WareWolf-MoonWall and @tidux already raised, and I'm not adding a new one — I'm confirming it's still live on the head I reviewed. `mergeable` is CONFLICTING and `mergeStateStatus` is DIRTY on e671ac78026616ad35215e797912ce6019c2dd33. @tidux ground-truthed the conflict to 475238513 (the SOP-engine unification, #7546), which threaded `sop_engine` and `sop_audit` into the same `from_config_with_session_cwd_and_mcp_backchannel` / `..._approval_mode` / `from_config_with_tui_env` signatures and the same WS/ACP call sites this PR edits. Both are pure trailing-arg additions in non-overlapping logical territory, so the resolution is a union — keep SOP/audit, add `canvas_store` — but the textual conflict regions overlap, so the merge driver can't compose them. @tidux's site-by-site map of the expected post-rebase signatures is excellent and I'd follow it verbatim. Per FND-005 §4 and AGENTS.md (validate by risk tier), after rebasing please re-run the full battery on the *fresh* head — `cargo fmt --all -- --check`, `cargo clippy --all-targets -- -D warnings` (the `--all-targets` form, so test-code lints are covered), `cargo test -p zeroclaw-tools canvas`, and `cargo test -p zeroclaw-runtime -p zeroclaw-channels -p zeroclaw-gateway` so the SOP wiring stays covered alongside the canvas threading — because all the green CI on this PR ran against the pre-conflict tree and tells us nothing about the merged result. I did not attempt the rebase locally; claiming I'd validated a head that doesn't exist yet would be inventing evidence.

---

**Verdict: a comment, not a block.** The fix is correct, surgical, and on the right side of the single-source-of-truth rule — it consolidates the canvas store back to the canonical `AppState.canvas_store` rather than duplicating it, and I verified that end to end against the head. I have nothing of my own to block on. But the branch is CONFLICTING/DIRTY against master (the trailing-arg collision with the SOP-engine PR #7546), and CI's green status reflects the pre-conflict tree only, so this cannot merge as-is regardless of substance — the rebase @WareWolf-MoonWall and @tidux named is required first, with the cargo battery re-run on the rebased head. Once the branch is back on top of current master with SOP/audit and `canvas_store` composed and the battery is green on the fresh head, I'm a ready `--approve` on the next pass — and the only carry-forward I'd ask the author to settle is the 🟡 above: either add the runtime-crate construction test or open a tracked, owned follow-up issue for it. The milestone (v0.8.1, due 2026-06-19) looks right for an integrations-era canvas-wiring fix and I'm not proposing any change to it.

WareWolf-MoonWall:
Carry-forward from rounds 11 and 12. I verified the current head against local master this
round and the situation is unchanged: the branch is still at `e671ac78`, `mergeable_state:
dirty`, and GitHub reports `mergeable: false`. The rebase has not landed.

### 🔴 Blocking — Branch is still conflicted with master (carry-forward)

This is the same block I named in round 11 and round 12, now independently ground-truthed by
both @Audacity88 and @tidux against the current master tree.

**What the conflict is:** master merged #7546 (`fix(runtime): unify SopEngine construction`)
after this branch was cut. That PR added `sop_engine: Option<Arc<Mutex<SopEngine>>>` and
`sop_audit: Option<Arc<SopAuditLogger>>` as trailing parameters to the same three function
signatures this PR edits: `from_config_with_session_cwd_and_mcp_backchannel`,
`from_config_with_session_cwd_and_mcp_approval_mode`, and `from_config_with_tui_env`. The PR
adds `canvas_store: Option<CanvasStore>` after those same positions, but the textual diff was
authored before the SOP params existed, so the merge driver cannot compose them automatically.
I confirmed this against local master: `ws.rs:416-425` on master already passes
`state.sop_engine.clone(), state.sop_audit.clone()` and does not yet have `canvas_store`.

**What the rebased head needs to look like.** @tidux mapped this precisely (4 h ago) and I
agree with every step:

- `from_config_with_session_cwd_and_mcp_backchannel` → 8 args, trailing:
  `sop_engine, sop_audit, canvas_store`
- `from_config_with_session_cwd_and_mcp_approval_mode` → 10 args, trailing:
  `sop_engine, sop_audit, canvas_store`
- `from_config_with_tui_env` → 8 args, trailing: `sop_engine, sop_audit` (no `canvas_store`
  needed here; the non-backchannel wrappers pass `None` for canvas through the inner call)
- `ws.rs` call site: pass `state.sop_engine.clone(), state.sop_audit.clone(),
  Some(state.canvas_store.clone())` as the trailing three
- Three `acp_server.rs` call sites: pass `self.sop_engine.clone(),
  self.sop_audit.clone(), self.canvas_store.clone()` as the trailing three
- The non-backchannel `from_config_with_session_cwd_and_mcp` wrapper: three trailing `None`s
  (`sop_engine None`, `sop_audit None`, `canvas_store None`) to match the 10-arg inner shape

After rebase, please re-run the full battery on the fresh head:

```
cargo fmt --all -- --check
cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway -p zeroclaw-channels -p zeroclaw-tools \
    --all-targets -- -D warnings
cargo test -p zeroclaw-tools canvas
cargo test -p zeroclaw-runtime -p zeroclaw-channels -p zeroclaw-gateway
```

The last command (per my round 11 request) keeps the SOP-engine wiring covered alongside
canvas threading. Then ping @Audacity88 to resume their technical pass.

### 🟡 Warning — Self-approval carries forward

@singlerider's review is still a self-approval (state: COMMENTED in the reviews API, not a
formal independent approval). It does not satisfy the independent reviewer gate. @Audacity88
has explicitly deferred until the rebase. Once the branch is clean and CI is green, please
re-request @Audacity88's review.

### 🟢 What looks good — Fix is architecturally correct

Nothing about the substance of the fix has changed since my prior rounds. The diagnosis is
right: `ws.rs` and the three ACP session-construction sites were passing `None` to
`all_tools_with_runtime`, causing each session to build a private default `CanvasStore`
instead of sharing the gateway's `AppState.canvas_store`. Threading `Option<CanvasStore>` as
a cheap `Arc` clone through `from_config_with_session_cwd_and_mcp_backchannel` is the correct
scoped fix. This is removing duplicate state (private per-session stores), not introducing it.

### 🟢 What looks good — Regression test covers the right invariant

`canvas_tool_renders_into_shared_store` pins the Arc-handle contract: a clone given to the
tool and a separately-held handle observe the same frames. That's the invariant the wiring
depends on, and @tidux's non-blocking note about a construction-path-level test is a good
follow-up to track. The current test is sufficient to let this land; the gap is worth an
issue tagged `quality: test`.

### 🟢 What looks good — Scope discipline verified

I re-verified against local master: `src/main.rs:3328-3335` constructs `AcpServer` without
`with_canvas_store`, so standalone `zeroclaw acp` correctly passes `None` through all three
session sites. The non-backchannel wrappers (`from_config_with_session_cwd_and_mcp`,
`from_config_with_tui_env`) forward `None` for canvas and are unchanged in behavior. The fix
touches exactly the WS + gateway-ACP paths described in #7563.

---

**Verdict: `--comment`.** My blocking finding stands. No new blockers of my own beyond the
rebase. @Audacity88 holds no active `CHANGES_REQUESTED` (review not yet submitted); they are
waiting on the same gate. Once the branch is clean, CI is green on the rebased head, and
@Audacity88 has completed their technical pass, I can convert to `--approve` — the fix is
correct and I have no other concerns.

Audacity88 (approved):
Thanks for refreshing this.

I re-reviewed current head c47909d after the branch refresh. I checked the live PR state, current GitHub checks, prior comments/reviews, the final PR diff, the refreshed constructor/call-site source, and an exact merge-tree against current `origin/master`. I did not run local Cargo validation for this review-only pass.

### ✅ Resolved — The rebase composes SOP/audit and canvas-store wiring correctly

The branch now keeps the SOP handles from current master and adds the shared canvas store on the same construction path. The refreshed `from_config_with_session_cwd_and_mcp_backchannel` signature carries `sop_engine`, `sop_audit`, and `canvas_store` together, and the inner approval-mode constructor passes all three through to `all_tools_with_runtime`.

The WS chat path now passes `state.sop_engine.clone()`, `state.sop_audit.clone()`, and `Some(state.canvas_store.clone())`. The three ACP session construction sites likewise pass `self.sop_engine.clone()`, `self.sop_audit.clone()`, and `self.canvas_store.clone()`. That is the union resolution the earlier reviews asked for.

### ✅ Resolved — The canvas store fix preserves the shared-store invariant

The fix shape is still right. `CanvasStore` is an Arc-backed handle, so cloning it into the WS and gateway-ACP agent sessions keeps the canvas tool writing into the same store that `/canvas` reads from. The non-backchannel constructors still pass `None`, which preserves the standalone/TUI behavior where there is no shared gateway canvas reader.

The new `canvas_tool_renders_into_shared_store` regression test covers the important Arc-handle invariant directly. A construction-path test would still be a useful follow-up, as tidux called out, but I do not think it needs to block this PR now that the refreshed call-site wiring is small and auditable.

### 🟡 Warning — This approval is not merge-readiness

The earlier hard conflict is resolved in the merge-tree, and current GitHub CI is green. Current `master` has still moved past this head again: my freshness check shows the PR is 28 commits behind current `origin/master`, with same-file base churn in `crates/zeroclaw-runtime/src/agent/agent.rs` and risky touched surfaces (`runtime`, `gateway`, `channels`, `tools`).

So I am approving the technical fix and clearing my review hold, but I am not calling this merge-ready. Before landing, please either update the branch and wait for refreshed CI, run/record an exact merge-result smoke, or get explicit maintainer acceptance of that stale-branch risk.

WareWolf-MoonWall (approved):
# Re-Review — PR #7678 · fix(runtime): thread shared CanvasStore into WS chat and ACP agent sessions

**Head SHA:** 0422d0036128e935bee794c8a8ca69677ace76e0
**Verdict:** APPROVED (`--approve`)
**Prior verdict:** COMMENTED at e671ac78 (🔴 dirty branch — held for rebase)
**Active CHANGES_REQUESTED from other reviewers:** None

---

Re-reviewed against the rebased head 0422d0036. My Round-1 comment held this PR on the dirty branch — the rebase was the only gate I named. The branch is now clean (`mergeable: true`), Audacity88 has done the independent technical pass I asked for and approved, and the rebase correctly preserved the SOP/audit wiring from master alongside the new `canvas_store` threading. Clearing my hold and approving. One 🟡 for maintainers before landing.

---

### ✅ Resolved — Branch is now clean and rebased correctly

My Round-1 🔴 was a straight merge-blocker: `agent.rs`, `ws.rs`, and `acp_server.rs` all conflicted with the SOP engine/audit wiring that landed in master via #7546 while this branch was open. The rebase applied the union resolution that the whole review thread had sketched:

- `from_config_with_session_cwd_and_mcp_backchannel` signature now carries `sop_engine`, `sop_audit`, and `canvas_store` together.
- `ws.rs:422` passes `state.sop_engine.clone()`, `state.sop_audit.clone()`, `Some(state.canvas_store.clone())` as the trailing three. ✓
- All three `acp_server.rs` session-construction sites (581, 793, 994) pass `self.sop_engine.clone()`, `self.sop_audit.clone()`, `self.canvas_store.clone()` as the trailing three. ✓
- The two non-backchannel convenience constructors (`from_config_with_tui_env`, `from_config_with_session_cwd_and_mcp`) still forward `None` for `canvas_store`, which is correct — those paths have no gateway canvas reader. ✓

The API reports `mergeable: true` at this head.

---

### ✅ Resolved — Independent technical review completed

My Round-1 🟡 noted @singlerider's self-approval didn't count toward the independent reviewer requirement and asked for Audacity88's technical pass after the rebase. Audacity88 has approved at this head and confirmed:

- The SOP/audit + canvas_store union on all construction paths.
- The `CanvasStore` Arc-clone invariant is preserved: the WS and gateway-ACP sessions share the same underlying allocation that `/canvas` subscribes to.
- The regression test `canvas_tool_renders_into_shared_store` covers the Arc-handle contract directly.

Independent reviewer requirement is satisfied.

---

### ✅ Resolved — SSOT consolidation: shared store threaded, not duplicated

The root bug was that `from_config_with_session_cwd_and_mcp_backchannel` forwarded `None` for `canvas_store` to `all_tools_with_runtime`, so WS chat and ACP session agents silently created private default stores per session while `/canvas` read from `AppState.canvas_store`. This PR consolidates both paths to the shared gateway store.

The pattern is Arc-clone threading, not duplication — every clone shares the same underlying allocation. The standalone `zeroclaw acp` CLI path (`src/main.rs`) still constructs `AcpServer` without `with_canvas_store`, so its session sites pass `None` correctly (no gateway reader to feed).

---

### 🟡 Warning — SHA discrepancy in Audacity88's review body; maintainers should confirm

Audacity88's review text references "current head c47909d" but the reviews API records their approval at commit 0422d0036. These are different short SHAs. On a `risk: high` change, maintainers should confirm before landing whether this reflects a force-push that happened between Audacity88's review pass and the recorded commit ID, and if so that the post-review delta was trivial (e.g., an amended commit message only). If the diff was unchanged and this is a typo in the review body, no action needed — just worth noting explicitly given the risk tier.

---

### 🟡 Warning — Branch is behind current master; confirm merge-result smoke before landing

Audacity88 noted approximately 28 commits between this head and current `origin/master`, with `agent.rs` among the still-changing files. `mergeable: true` means GitHub's 3-way merge algorithm sees no textual conflict, but for a change touching `runtime`, `gateway`, `channels`, and `tools` across four call sites, a quick `git merge-tree` or post-rebase CI run is worth the 10 minutes before the squash-merge. This is a freshness gate, not a merits gate.

---

### 🟢 What looks good — Arc-clone threading is the correct shared-ownership pattern

`Option<CanvasStore>` flowing through as `Arc<RwLock<HashMap<...>>>` under `#[derive(Clone)]` means every `.clone()` call is a pointer increment, not a heap allocation. No new `Mutex` or `RwLock` is introduced by this PR. The tool and the gateway reader hold separate handles into the same object. There is no data race concern here.

---

### 🟢 What looks good — `canvas_tool_renders_into_shared_store` regression test

The test directly asserts the Arc-handle invariant: a `CanvasTool` built from a cloned `gateway_store` renders a frame visible through the original handle. This is the exact contract the WS/ACP wiring depends on, and it is the right unit-level assertion for the bug class #7563 describes.

tidux's 🟡 about adding a construction-path test covering the `from_config_with_session_cwd_and_mcp_backchannel` → `all_tools_with_runtime` seam is still a valid follow-up. The call-site wiring is small and auditable enough that I'm comfortable approving without it — but if a follow-up issue gets filed and tagged, that would close the regression-path gap cleanly.

PR #6667, opened 2026-05-14
feat(skills): background review fork + skill_manage tool (agentskills.io SKILL.md)

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Closes the integration gap from #4619 — `SkillImprover` shipped without a caller, so `skills.skill-improvement.enabled` had no runtime effect. This wires it up via a **post-turn background review fork** with a restricted toolset (`skills_list`, `skill_view`, `skill_manage`), retargeted to the [agentskills.io](https://agentskills.io) `SKILL.md` YAML front-matter standard.
  - **Opt-in only:** `skills.skill-improvement.enabled` defaults to `false`. The fork runs after the user-visible response is printed, so it never delays output.
  - **Cost accounting (addresses prior blocker):** the post-turn fork now runs under the **same** `TOOL_LOOP_COST_TRACKING_CONTEXT` scope as the parent turn, so its provider calls are recorded against — and bounded by — the parent's cost tracker and budget.
  - **Safety boundary preserved:** `safe_skill_dir` canonical-root anchoring, symlink rejection for mutating ops, post-mutation audit rollback under the active `allow_scripts` policy, and the `should_improve_skill` cooldown gate on the `patch` path.
- **Scope boundary:** Does not change the main agent's tool registry (these tools are built on demand only inside the fork), does not touch the interactive REPL path (single-shot only, same scope as `SkillCreator`), and adds no memory tools to the fork.
- **Blast radius:** `crates/zeroclaw-runtime` skills/tools + the single-shot branch of `agent::loop_::run`; `crates/zeroclaw-config` schema (`[skills.skill_improvement]`). Inert unless a user explicitly opts in.
- **Linked issue(s):** Supersedes #6594. Related #3683, #4619.
- **Labels:** `enhancement`, `size: XL`, `risk: high`, `docs`, `agent`, `config`, `runtime`, `skills`, `needs-author-action`

## Rebase / blocker resolution (this revision)

Rebased onto current `master` (was DIRTY/CONFLICTING) and integrated with the multi-agent V3 runtime that landed since the last push:

- Review fork uses `ModelProvider` (renamed from `Provider`) and `zeroclaw_log::record!` (the workspace's sole logging surface — `tracing` is not a runtime dep) instead of `tracing::` macros.
- Fork is wired through `config.agent_workspace_dir(agent_alias)` so it operates on the **same** skills root the agent loads skills from.
[...]

PR #7679, opened 2026-06-15
test(skills): cover skill_manage patch cooldown gate

### LINKED ISSUE #6683: skill_manage `patch` ignores cooldown — unbounded patches per skill possible
## Summary

The `SkillImprover::should_improve_skill` predicate and its `is_on_disk_cooldown` helper (both introduced in #6667) exist and are unit-tested, but **no production code path calls them**. As a result, `skill_manage` `action=patch` writes unconditionally — there's no system-level rate limit on patches to the same skill.

The fork's prompt-level soft gates (review prompt's "Nothing to save." encouragement, the iteration-budget trigger) are real, but they're *advisory*, not enforced. Once the LLM decides to patch, the patch happens.

## Why this matters

Concrete scenario:

1. User runs `zeroclaw agent -m "<task>"` → fork fires, picks `git-assistant`, patches it.
2. User runs another agent command 30 seconds later, generating ≥`nudge_interval_iterations` tool calls.
3. Fork fires again. LLM sees the just-touched `git-assistant` skill, decides it still has signal worth saving, patches *again*.
4. Same skill rewritten N times in N successive runs. Audit trail grows unbounded, LLM tokens spent on a skill that just changed, drift accumulates.

This is more visible with stronger models (`gpt-4o`, Claude) that act on signals more aggressively than `gpt-4o-mini`. With `gpt-4o-mini`'s conservative "Nothing to save." bias we observed during #6667 smoke testing, the gap is masked; with better models it surfaces.

## Where the gap is

`SkillImprover::improve_skill` is documented as caller-gated:

```rust
/// **Caller-gated:** this does NOT check `should_improve_skill` — callers
/// must check that themselves before invoking, so they can also skip the
/// (expensive) LLM call that produces `improved_content`.
```

The natural caller is `SkillManageTool::patch` ([tools/skill_manage.rs](https://github.com/zeroclaw-labs/zeroclaw/blob/master/crates/zeroclaw-runtime/src/tools/skill_manage.rs)). Currently it constructs a `SkillImprover` with `cooldown_secs: 0` (effectively disabling cooldown) and skips the `should_improve_skill` check entirely.

## Fix shape

About 30 lines including a test. Two changes:

1. Plumb `SkillImprovementConfig` (or just `cooldown_secs`) through `SkillManageTool::new` so the tool has the real cooldown value from runtime config — currently the tool is constructed with `cooldown_secs: 0`.
2. In `SkillManageTool::patch`, call `improver.should_improve_skill(slug)` before delegating to `improve_skill`. If false, return a structured `ToolResult { success: false, error: Some(<cooldown reason>) }`. The fork's LLM sees the cooldown error in its tool history and can pick a different skill or emit `Nothing to save.`

## What this protects against

- Same-skill rewrite storms within a single user session (in-memory cooldown — fast path)
- Cross-session rewrite storms when the user runs the agent repeatedly within `cooldown_secs` (on-disk cooldown via the YAML `updated_at:` timestamp — durable across process restarts)

## What this does NOT do

- Doesn't gate the fork itself from running. The fork still fires every `nudge_interval_iterations` and the LLM still surveys the conversation. The cost is borne whether or not there's anything to save.
- Doesn't gate `skill_manage` `write_file` (adding `references/...` files is allowed regardless of cooldown — that's a different operation) or `archive` (archival is rare and should always be allowed).

## Tests to add

- `skill_manage_patch_blocks_when_skill_is_on_cooldown` — write SKILL.md with recent `updated_at`, call patch, assert `success: false` and that the error message names the cooldown.
- `skill_manage_patch_proceeds_when_skill_is_stale` — write SKILL.md with `updated_at` older than cooldown_secs, call patch, assert it writes.
- `skill_manage_patch_proceeds_when_no_updated_at` — fresh SKILL.md with no `updated_at`, call patch, assert it writes.

## Related

- Introduced (dormant) in #6667
- Builds on the on-disk cooldown infrastructure added in commit 1/2 of #6667

singlerider:
## Summary

The cooldown gate for `skill_manage action=patch` shipped in #6667: `SkillManageTool::patch` calls `SkillImprover::should_improve_skill` before delegating to `improve_skill` and returns a structured "on cooldown" error when the skill was patched within `cooldown_secs` (`crates/zeroclaw-runtime/src/tools/skill_manage.rs:500-506`). But every existing test built the tool with `cooldown_secs = 0` (`cfg_no_cooldown`), so the gate had **zero test coverage** and could silently regress.

This adds the three regression cases originally requested in #6683, all exercising a non-zero cooldown:

- `skill_manage_patch_blocks_when_skill_is_on_cooldown` — a fresh `updated_at` in front-matter is refused with a cooldown error and the file is left untouched.
- `skill_manage_patch_proceeds_when_skill_is_stale` — `updated_at` older than `cooldown_secs` patches normally.
- `skill_manage_patch_proceeds_when_no_updated_at` — a skill with no prior timestamp is not on cooldown and the first patch proceeds.

Adds a `cfg_with_cooldown(secs)` test helper alongside `cfg_no_cooldown`. No production code change — this locks in behavior that already shipped.

Closes #6683

`cargo fmt --all -- --check`
```
(clean, exit 0)
```

`cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings`
```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 16s
(0 warnings, 0 errors)
```

`cargo test -p zeroclaw-runtime skill_manage_patch`
```
running 8 tests
test tools::skill_manage::tests::skill_manage_patch_blocks_when_skill_is_on_cooldown ... ok
test tools::skill_manage::tests::skill_manage_patch_proceeds_when_no_updated_at ... ok
test tools::skill_manage::tests::skill_manage_patch_proceeds_when_skill_is_stale ... ok
test tools::skill_manage::tests::skill_manage_patch_atomically_updates_md ... ok
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 2163 filtered out
```

None. Test-only change.

## Compatibility (required)

Backward compatible. Tests-only; no public API, config, or schema change.

`risk: low` — test-only. `git revert <sha>` if needed.

### DIFF
--- a/crates/zeroclaw-runtime/src/tools/skill_manage.rs
+++ b/crates/zeroclaw-runtime/src/tools/skill_manage.rs
@@ -819,6 +819,14 @@ mod tests {
         }
     }
 
+    fn cfg_with_cooldown(secs: u64) -> zeroclaw_config::schema::SkillImprovementConfig {
+        zeroclaw_config::schema::SkillImprovementConfig {
+            enabled: true,
+            cooldown_secs: secs,
+            ..Default::default()
+        }
+    }
+
     async fn write_skill(workspace: &Path, slug: &str, md: &str) {
         let dir = workspace.join("skills").join(slug);
         tokio::fs::create_dir_all(&dir).await.unwrap();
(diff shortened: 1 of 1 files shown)


singlerider (commented):
I built this PR, so this is a self-review pass against the project's review protocol rather than a formal approval. I read the full diff against `upstream/master`, ran the suite, and cross-checked the test assertions against the cooldown gate in `skill_manage.rs`.

Context: #6683 reported that `skill_manage` patch ignored the configured cooldown. The production gate shipped in #6667 (the `should_improve_skill` check at the top of the patch path), but every existing test used `cooldown_secs: 0`, so the gate had zero coverage and could regress silently. This PR adds that coverage. It is test-only, no production change.

### 🟢 All three cooldown branches are covered
`skill_manage_patch_blocks_when_skill_is_on_cooldown`, `_proceeds_when_skill_is_stale`, and `_proceeds_when_no_updated_at` exercise the fresh-`updated_at`, stale-`updated_at`, and absent-`updated_at` paths through the public `execute` entry, which is the real dispatch the agent uses. Each asserts both the `ToolResult` and the on-disk SKILL.md (presence or absence of the patched marker), so a regression that returned success but skipped the write, or vice versa, would fail.

### 🟢 Blocking-case assertion is specific
The on-cooldown test asserts `error` contains `cooldown`, which matches the literal at `skill_manage.rs:504`, and confirms the file was not mutated. That is the exact failure mode #6683 described.

### 🟢 Validation
8/8 `skill_manage_patch` tests pass on the branch, clippy clean (0 warnings) on `zeroclaw-runtime`. `risk: low` is accurate: no production code path changed.

No blocking findings. Nothing to fix.

Audacity88 (approved):
Reviewed current head `a8f85d71caf28212d40749399665d2ad18a8d8da` against the live PR state, linked issue #6683, @singlerider's self-review, the full diff, and the current `SkillManageTool::patch` / `SkillImprover` cooldown source on `origin/master`.

I also checked the green CI run on this head and the validation evidence in the PR body. I did not run additional local commands for this test-only PR. I do not see an author-facing blocker.

### 🟢 What looks good — the tests exercise the real patch boundary

The new cases go through `SkillManageTool::execute` with `action = "patch"` and a non-zero `cooldown_secs`, so they cover the same dispatch boundary the runtime tool uses. That matters here because `SkillImprover::improve_skill` is explicitly caller-gated; a helper-only test would not prove that `SkillManageTool::patch` actually calls `should_improve_skill` before writing.

### 🟢 What looks good — the file-system assertions prove the cooldown outcome

The fresh-`updated_at` test checks both sides of the failure mode from #6683: the tool returns a cooldown error, and the on-disk `SKILL.md` does not gain the improved content. The stale and missing-`updated_at` tests then prove the allowed branches still write that content. Together, those assertions would catch both a false success that skipped the write and a false failure that still mutated the file.

Approved.

PR #7680, opened 2026-06-15
fix(daemon): allow any configured channel as heartbeat.target

### LINKED ISSUE #6433: [Bug]: Heartbeat not working in combination with Matrix channel
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

The deamon only accepts 4 channels for its heartbeat target: [mod.rs#L989](https://github.com/zeroclaw-labs/zeroclaw/blob/master/crates/zeroclaw-runtime/src/daemon/mod.rs#L989)
ERROR zeroclaw_runtime::daemon: Daemon component 'heartbeat' failed: unsupported heartbeat.target channel: matrix

But i am using Matrix and that doesnt work. 

### Expected behavior

I expect all configurable channels to be available for the heartbeat.

### Steps to reproduce

```bash
[heartbeat]
enabled = true
# ...
target = "matrix"
to = "room_id"
```

### Impact

Can't use heartbeat

### Logs / stack traces

```text

```

### ZeroClaw version

v0.7.4

### Rust version

_No response_

### Operating system

docker

### Regression?

No, first-time setup

### Pre-flight checks

singlerider:
## Summary

`validate_heartbeat_channel_config` hardcoded a four-arm string match (telegram/discord/slack/mattermost) and bailed on everything else, so `heartbeat.target = "matrix"` failed at daemon start with `unsupported heartbeat.target channel: matrix` — even though the actual delivery path (`deliver_announcement` → the registered `DELIVERY_FN`) already handles every channel generically. The validation gate was stricter than delivery for no reason.

Fix drives the check off the channel registry instead of a hardcoded subset:

- `ChannelsConfig::channel_presence()` returns one `(name, configured)` row per channel; `is_known_channel` / `is_channel_configured` iterate it. No surface restates a slice of the channel list.
- The daemon validation now rejects only genuinely unknown channels and otherwise requires the named channel to be configured.

This unblocks matrix and the other ~30 channels (email, signal, irc, ...) as heartbeat targets.

Repro (before): set `[heartbeat] enabled = true`, `target = "matrix"`, `to = "<room_id>"` with a configured `[channels.matrix.*]`; daemon logs `Daemon component 'heartbeat' failed: unsupported heartbeat.target channel: matrix`.

Closes #6433

`cargo fmt --all -- --check`
```
(clean, exit 0)
```

`cargo clippy -p zeroclaw-config -p zeroclaw-runtime --all-targets -- -D warnings`
```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 17s
(0 warnings, 0 errors)
```

`cargo test -p zeroclaw-runtime resolve_delivery`
```
running 7 tests
test daemon::tests::resolve_delivery_accepts_matrix_target ... ok
test daemon::tests::resolve_delivery_rejects_unsupported_channel ... ok
test daemon::tests::resolve_delivery_requires_channel_configuration ... ok
test daemon::tests::resolve_delivery_accepts_telegram_configuration ... ok
test result: ok. 7 passed; 0 failed
```

None. The validation gate only widened to match what the delivery layer already permits; no new delivery capability, no new surface. Targets still must name a configured channel.

## Compatibility (required)

Backward compatible. Previously-accepted targets behave identically. Previously-rejected-but-configured channels (matrix, email, etc.) now validate. New `ChannelsConfig` methods are additive.

`git revert <sha>`. Self-contained; no migrations or persisted state.


Audacity88 (dismissed):
Checked the PR head diff, linked issue #6433, current checks/labels, PR review/comment state, and the relevant heartbeat validation and delivery paths. CI is green and there are no prior PR reviews or PR comments.

Thanks for chasing this.

### 🟢 What looks good — Matrix heartbeat coverage moves in the right direction

Moving the heartbeat check away from the original four-channel allowlist is the right direction, and the Matrix regression coverage is useful.

### 🔴 Blocking — `heartbeat.target` now accepts configured channels that cannot deliver heartbeat messages

`channel_presence()` treats every configured channel table as a valid name-addressed heartbeat target, including listener-oriented channels such as `mqtt`. `resolve_heartbeat_delivery()` then returns the accepted `heartbeat.target` string unchanged, but heartbeat delivery ultimately needs a target that `deliver_announcement()` can actually send through. `mqtt` is not in that outbound delivery surface.

That means a config with `[heartbeat] target = "mqtt"` plus a configured `[channels.mqtt.<alias>]` now passes startup validation even though there is no supported heartbeat send path for that target. This preserves the same class of bug the PR is trying to remove: the config validator says a target is usable even though the delivery surface cannot send through it.

Please make the heartbeat check derive from the actual sendable delivery surface, or otherwise keep a single canonical list of heartbeat-deliverable channel names instead of all configured channel tables. A regression test for a configured-but-not-deliverable channel such as `mqtt`, alongside the Matrix acceptance test, would cover the edge.

singlerider:
🔴 **Fixed.** You're right: `channel_presence()` treated every configured channel table as a deliverable heartbeat target, including input-only transports. Confirmed the no-op surface: `amqp::send` returns `Ok(())` without publishing ("consumed as an inbound trigger source"), `voice_wake::send` is "input-only; outbound messages are not supported", and `mqtt` is a pure SOP fan-in listener with no `Channel` impl at all. So `target = "mqtt"` / `"amqp"` previously passed validation while the delivery layer either silently dropped the heartbeat (amqp) or bailed at send time (mqtt), exactly the validator/delivery mismatch this PR set out to remove.

Kept it on the single registry rather than a hardcoded subset: extended `channel_presence()` to a `(name, configured, deliverable)` row per channel, with `deliverable = false` for `mqtt`, `amqp`, `voice_wake`. Added `is_channel_deliverable()` reading that column. `validate_heartbeat_channel_config` now rejects a configured-but-undeliverable target with an input-only message, in addition to the unknown and not-configured cases. Registry-driven channels (matrix, mattermost, qq, irc, twitch, etc.) stay deliverable via the live `CRON_CHANNEL_REGISTRY` send path, so the matrix fix from #6433 is unaffected.

Added `resolve_delivery_rejects_configured_but_undeliverable_channel` (configured `mqtt` rejected) alongside the existing matrix acceptance test, per your suggestion.

fmt clean, `clippy -p zeroclaw-config -p zeroclaw-runtime --all-targets -D warnings` 0 warnings, 30/30 daemon lib tests pass (incl. the new one), 20/20 config channel tests pass.

Nillth (approved):
I reviewed this against PR head d163d26098134c783dc552879e8146b1b65bc10d and took stock of the full history first. The earlier review from @Audacity88 (submitted on 5a81c67, now **dismissed**) raised one substantive blocker: that driving `heartbeat.target` off every configured channel table would accept listener-oriented channels such as `mqtt` whose delivery path is a no-op, recreating the same validator-vs-delivery mismatch the PR set out to remove. I traced issue #6433, the delivery surface (`deliver_announcement` → `CRON_CHANNEL_REGISTRY` / the fallback `match`), and each implicated `Channel::send`, then checked whether the two commits that landed after that review actually close the concern. They do — 2d9a979 added the deliverability column and the input-only rejection, and d163d26 added a drift guard. There are no inline threads or top-level comments, no other reviews, and the dismissed review is not an active block. CI is green across all 18 checks.

### 🟢 What looks good — validation now derives from the channel registry, not a hand-cut subset

The core of #6433 is fixed correctly. `validate_heartbeat_channel_config` (crates/zeroclaw-runtime/src/daemon/mod.rs:1364) no longer carries its own four-arm telegram/discord/slack/mattermost allowlist; it asks `ChannelsConfig` three questions — known, configured, deliverable — and the daemon restates no slice of the channel list. That is exactly the single-source-of-truth posture AGENTS.md asks for under "patterns that are NOT duplicate state": the daemon threads `&Config` and resolves channel facts from the canonical registry on demand rather than caching a parallel list. matrix, email, signal, irc and the rest now validate, which is the behavior the issue reporter @niekp expected.

### ✅ Resolved — configured-but-undeliverable targets (the @Audacity88 mqtt blocker)

The dismissed review's blocker is addressed. `channel_presence()` (crates/zeroclaw-config/src/schema.rs:11546) now carries a third `deliverable` column, set to `false` for exactly the input-only transports — `mqtt`, `amqp`, `voice_wake` — and `is_channel_deliverable()` (schema.rs:11608) reads it. The daemon's validation adds a third gate: a configured-but-undeliverable target is rejected with a clear, actionable message ("…is an input-only channel that cannot deliver outbound messages"). I confirmed against PR-head source that the three flagged channels genuinely have no-op sends: amqp (crates/zeroclaw-channels/src/amqp.rs:250, "consumed as an inbound trigger source") and voice_wake (voice_wake.rs:90, "input-only; outbound messages are not supported") both `Ok(())` without transmitting, while voice_call's send does real work (TTS to the active call), so marking it deliverable is right. This matters because the registry path in `deliver_announcement` calls `ch.send()` directly when a channel is live — so for mqtt that would have *silently dropped* the heartbeat rather than erroring, which is precisely the silent-failure class this PR fights. Catching it at startup validation is the correct fix. Thanks for chasing the edge rather than papering over it. 2d9a979

### 🟢 What looks good — test coverage spans both the newly-allowed and the still-rejected cases

The acceptance and rejection sides are both covered, which is what makes this safe to widen: `resolve_delivery_accepts_matrix_target` (daemon/mod.rs:1689) for the newly-unblocked path, `resolve_delivery_rejects_unsupported_channel` (1677, now using `carrier_pigeon` so it tests a genuinely-unknown name rather than the stale `email` case), `resolve_delivery_rejects_configured_but_undeliverable_channel` (1706) for the mqtt edge, and `resolve_delivery_requires_channel_configuration` (1727) for the known-but-unconfigured path. The error precedence in the validator is also correct — unknown is checked before configured, configured before deliverable — so each rejection surfaces the most specific actionable message.

### 🟡 Warning — the `deliverable` column mirrors `Channel::send`, and the drift guard locks membership but not correctness

This is the one place worth a second look before merge, and it sits squarely on the AGENTS.md single-source-of-truth rule. The *true* source of "can this channel deliver" is each `Channel::send` implementation — whether it transmits or is a documented no-op. `channel_presence()` re-encodes that fact as 34 hand-written booleans. The PR mitigates the drift risk with `channel_presence_names_are_unique_and_undeliverable_set_is_fixed` (schema.rs:28385), which is a genuinely good instinct — but note what it actually pins: it asserts the *undeliverable set equals exactly `["amqp","mqtt","voice_wake"]`*. It does not assert those booleans agree with the underlying sends. So the guard catches an accidental flip of an existing row, but if a future input-only channel is added (say a `kafka` fan-in listener) and its row is written `deliverable: true`, the test passes and the silent-drop bug returns for that channel — the same failure mode #6433 describes, one channel over. I would not block merge on this: the current 34 rows are correct as written, the guard meaningfully narrows the drift window, and a compiler-enforced link from `send` to this column is a larger change than this fix should carry. But it is worth a tracking note (a follow-up that derives deliverability from the channel registry/trait, or at minimum a comment on each no-op `send` pointing back to this column so the next person edits both) so the mitigation does not quietly decay. Flagging per the SSOT section's "materialized views must be built from canonical state" framing.

### 🔵 Suggestion — `voice_duplex` deliverability could not be fully verified from the diff alone

`voice_duplex` is marked `deliverable: true`. The author's curation reads correctly — a duplex (bidirectional) voice channel should be able to send — and I have no reason to doubt it. I simply want to be transparent that I could not confirm its `Channel::send` body against PR-head source the way I did for amqp/voice_wake/voice_call (the duplex channel postdates the snapshot I cross-checked the no-op sends in). If it is convenient, a one-line confirmation that `voice_duplex::send` actually transmits would fully close the loop; otherwise I am comfortable trusting the curation given the rest of the column checks out.

---

Verdict: this cleanly fixes #6433, the previously-blocking concern from @Audacity88 is fully resolved with a test that exercises the exact edge, and the change keeps validation derived from the canonical channel registry rather than a duplicated list. CI is green and the PR is mergeable. My one substantive note (the deliverability column mirroring `Channel::send` with a membership-only guard) is a real but non-blocking maintainability concern best handled as a follow-up, and the voice_duplex point is a minor verification courtesy. None of this rises to a block, and there is no active `CHANGES_REQUESTED` standing against the PR. Approving.

WareWolf-MoonWall (commented):
# Review — PR #7680 · Round 1

**Reviewed at**: d163d26098134c783dc552879e8146b1b65bc10d
**First review of this PR**
**Verdict**: `--comment`

---

## Context

Audacity88 posted a CHANGES_REQUESTED in Round 1 (blocking: `channel_presence()` accepted
configured input-only channels that cannot deliver). That review was dismissed after
singlerider pushed the deliverability fix (fourth commit adds `deliverable` column and
`is_channel_deliverable()`). There are no active CHANGES_REQUESTED from other reviewers.

My findings are warnings and a suggestion. No new blockers.

---

## Findings

### 🟢 What looks good — Registry-driven validation replaces hardcoded match correctly

Replacing the four-arm `match channel { "telegram" | "discord" | "slack" | "mattermost" }` with
three sequential checks against `ChannelsConfig::channel_presence()` is the right approach.
No hardcoded subset; `validate_heartbeat_channel_config` now accepts the same channel space
that `deliver_announcement` can reach.

Verifying the SSOT property: `validate_heartbeat_channel_config` receives a `&Config` reference
(not a snapshot or clone). `is_known_channel` / `is_channel_configured` / `is_channel_deliverable`
all call `self.channel_presence()` which reads directly from `self` (the live `ChannelsConfig`
fields). No intermediate cache. SSOT satisfied.

### 🟢 What looks good — `deliverable` column correctly captures input-only transports

The PR body correctly identifies all three non-deliverable channels and explains why each one
cannot deliver: `amqp::send` returns `Ok(())` without publishing, `voice_wake::send` is
input-only, `mqtt` has no `Channel` impl at all. The registry correctly marks all three as
`deliverable: false` and the test `channel_presence_names_are_unique_and_undeliverable_set_is_fixed`
encodes the invariant so a future PR that silently adds a fourth input-only transport would
fail the test rather than pass validation.

### 🟢 What looks good — Three-gate validation with regression tests for each path

`resolve_delivery_rejects_unsupported_channel`, `resolve_delivery_requires_channel_configuration`,
`resolve_delivery_rejects_configured_but_undeliverable_channel` (mqtt), and
`resolve_delivery_accepts_matrix_target` together cover the full decision tree.
The test that previously used `"email"` (a valid channel) as the "unsupported" test case was
updated to `"carrier_pigeon"` — a meaningless name that actually tests the unknown-channel path.
That was a latent test-correctness bug and it is good that it was caught.

### 🟡 Warning — Full validation suite absent after the deliverability fix

The PR body's validation section was written before Audacity88's blocking concern was
raised. It shows:

```
cargo test -p zeroclaw-runtime resolve_delivery
running 7 tests  ← targeted, not full suite
test result: ok. 7 passed
```

Singlerider's fix comment adds "30/30 daemon lib tests pass, 20/20 config channel tests pass"
but these are still targeted (`-p zeroclaw-config -p zeroclaw-runtime`). The PR template
requires:

```
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings   ← workspace-wide, not per-crate
cargo test                                   ← full suite, literal output
```

Please run the full battery post-fix and paste the output as a PR comment or amend the
PR body. The targeted evidence is reassuring, but the template requirement is
workspace-wide for a reason: `--all-targets` may catch warnings in crates that
import `ChannelsConfig` downstream.

### 🟡 Warning — Rollback section incomplete for `risk: medium`

The PR provides the fast rollback command (`git revert <sha>`) but the template requires
three items for medium-risk PRs:

- **Fast rollback command/path**: `git revert <sha>` ✓
- **Feature flags or config toggles**: not stated (should be "None")
- **Observable failure symptoms**: not stated (what to grep in daemon logs; which metric or
  alert would fire if heartbeat fails to start after a bad deploy)

Please add the two missing items to the Rollback section. For this PR they are likely
both simple ("None" and "daemon log line `Daemon component 'heartbeat' failed:` at startup"),
but they belong in the body for the risk tier.

### 🔵 Suggestion — Add a completeness guard for `channel_presence()`

`channel_presence_names_are_unique_and_undeliverable_set_is_fixed` verifies that channel
names in the registry are unique and the undeliverable set is exactly the three expected
channels. It does not verify that **every** field in `ChannelsConfig` has a row in
`channel_presence()`.

If a new channel (`foobar`) is added to `ChannelsConfig` and its `channel_presence()` entry
is forgotten, heartbeat validation will reject it as "unsupported" with no signal that the
registry is out of date. Consider adding one of:

1. An assertion that `channel_presence().len() == EXPECTED_CHANNEL_COUNT` with a comment
   pointing back to `ChannelsConfig` field count (requires manual update but is explicit).
2. A comment on `channel_presence()` that reads "if you add a channel field to
   `ChannelsConfig`, add a row here and update the count in the test."

The current test is good. This would make the maintenance contract explicit.

---

## Summary

The fix is correct, the SSOT property is satisfied, and Audacity88's blocking concern was
properly resolved. Two template-compliance warnings (validation evidence, rollback section)
and one suggestion (completeness guard). No blockers from me.

Once the full validation evidence is posted and the rollback section is completed, this
should be ready for approval.

Audacity88 (dismissed):
I re-reviewed this at d163d26098134c783dc552879e8146b1b65bc10d after the deliverability follow-up. I checked the current diff, CI/check state, prior reviews/comments, issue #6433, and the PR-head source for the heartbeat validator, `channel_presence()`, the channel orchestrator registry, `deliver_announcement()`, and the voice channel implementations.

The original blocker I raised for configured input-only transports is mostly resolved: `mqtt`, `amqp`, and `voice_wake` are now marked non-deliverable, and the new `resolve_delivery_rejects_configured_but_undeliverable_channel` test covers the exact `mqtt` edge.

There is still one channel in the new registry that appears to preserve the same validator-vs-delivery mismatch.

### 🔴 Blocking — `voice_duplex` is accepted as a heartbeat target but is not an outbound delivery channel

`ChannelsConfig::channel_presence()` marks `voice_duplex` as `(configured, deliverable = true)`, so a config with `[heartbeat] target = "voice_duplex"` plus a `[channels.voice_duplex.<alias>]` table now passes `validate_heartbeat_channel_config()`.

But I do not see a matching outbound delivery path for that target. The PR-head `crates/zeroclaw-gateway/src/voice_duplex.rs` file defines the WebSocket `VoiceEvent` protocol and `handle_voice_event()`; it does not implement `zeroclaw_api::channel::Channel` or provide a `send()` transport. `crates/zeroclaw-channels/src/lib.rs` has no `voice_duplex` module/export, and `crates/zeroclaw-channels/src/orchestrator/mod.rs` builds configured `VoiceWakeChannel` and `VoiceCallChannel` entries but has no corresponding `VoiceDuplex`/`voice_duplex` `ConfiguredChannel` construction. I also checked the `CRON_CHANNEL_REGISTRY` population path: it is built from those configured `Channel` instances, so there is no `voice_duplex` entry for heartbeat delivery to reuse.

That matters because `deliver_announcement()` first looks up `channel.to_ascii_lowercase()` in `CRON_CHANNEL_REGISTRY`; if there is no registry hit, it falls back to parsing a dotted `<type>.<alias>` ref. A bare `heartbeat.target = "voice_duplex"` that passed validation therefore falls through to the dotted-ref error path instead of being rejected at startup as non-deliverable.

That is the same class of bug this PR is trying to remove: validation says a configured heartbeat target is usable, but the delivery surface cannot actually send a heartbeat through it. Please either mark `voice_duplex` non-deliverable in `channel_presence()` or wire a real outbound delivery implementation and add a regression test for the chosen behavior. Given the current source, I think it should join the non-deliverable set alongside `mqtt`, `amqp`, and `voice_wake`.

Once that row and the drift test are updated, the core Matrix fix still looks right to me.

PR #7681, opened 2026-06-15
fix(runtime): detect no-progress loops across interleaved tool calls

### LINKED ISSUE #7143: Agent repeatedly runs near-duplicate shell discovery commands until max_tool_iterations is exhausted
## Summary

First, thank you for the project. It is great to see a Rust-based agent runtime that is much lighter on resources than many other agent systems we have tried.

We are seeing a ZeroClaw Slack-connected coding agent hit `max_tool_iterations` because it repeatedly calls `shell` with near-duplicate repo-discovery commands instead of moving on to edit files.

This does not look like a Slack delivery issue or a filesystem permission issue. The agent can access the repo and read files successfully, but it gets stuck re-running variants of `git status` + `find` until the tool iteration cap is reached.

## Environment

- ZeroClaw running as a systemd daemon
- Channel: Slack
- Agent use case: coding assistant for a private QA/dev repo
- Model: OpenAI GPT-5 via Codex-style coding agent setup
- Workspace path: `/home/agent-qa`
- Target repo path: `/home/agent-qa/repos/private-project`

## Relevant config, sanitized

```toml
[agents.assistant]
enabled = true
channels = ["slack.default"]
model_provider = "openai.default"
risk_profile = "assistant"
runtime_profile = "assistant"
compact_context = true
context_aware_tools = false
keep_tool_context_turns = 4
max_context_tokens = 20000
max_history_messages = 20
max_tool_iterations = 50
max_tool_result_chars = 100000
parallel_tools = true
tool_dispatcher = "auto"

[agents.assistant.workspace]
path = "/home/agent-qa/"
unrestricted_filesystem = true

[risk_profiles.assistant]
level = "full"
allowed_commands = ["*"]
allowed_roots = ["/home/agent-qa"]
auto_approve = ["*"]
workspace_only = false

[runtime_profiles.assistant]
agentic = true
max_tool_iterations = 50
max_actions_per_hour = 2000
shell_timeout_secs = 1800
agentic_timeout_secs = 3600

[channels.slack.default]
enabled = true
mention_only = true
thread_replies = false
interrupt_on_new_message = false
stream_drafts = true
cancel_reaction = "x"
```

## What we observed

A Slack task asked the agent to add a Playwright regression test in an existing repo.

The agent successfully found:

```text
private-web/playwright.config.mjs
private-web/playwright.next.config.mjs
private-web/e2e/map.spec.js
private-web/e2e/listing-draft.spec.js
private-web/e2e/diagnostics.spec.js
private-web/e2e-next/listing-detail.spec.js
```

It also successfully read the relevant files.

But after the initial discovery, it kept running variants of commands like:

```bash
cd /home/agent-qa/repos/private-project &&
git status --short --branch &&
find . -maxdepth 3 -type f \( -name 'playwright.config.*' -o -name 'package.json' -o -name '*.spec.*' \) |
sort |
head -200
```

This continued until iteration 50. There were no write/edit tool calls before the limit was reached.

Tool-call summary from the affected turn:

```text
shell:          43 calls
file_read:       7 calls
content_search:  2 calls
git_operations:  1 call
total:          53 tool calls across 50 iterations
```

## Expected behavior

After discovering and reading the relevant test files, the agent should either:

- proceed to edit/write the regression test, or
- stop and ask for clarification, or
- produce a final response explaining what is blocking it.

It should not spend dozens of iterations repeating near-identical discovery commands.

## Actual behavior

ZeroClaw eventually stops the turn because `max_tool_iterations = 50` is exhausted. The agent then reports that it did not complete the task before the execution limit.

## Questions

Is there a recommended config setting to prevent this kind of near-duplicate shell loop?

Would you recommend simply increasing `max_tool_iterations` for coding-agent use cases, or would that mostly mask the underlying problem and allow longer loops?

I saw docs for `tool_call_dedup_exempt` and `[pacing]` loop detection, but this case may not be caught by exact duplicate detection because the shell commands differ slightly while doing the same thing semantically.

Should this be handled via:

- stronger prompting/skills for coding agents,
- stricter shell-loop detection in ZeroClaw,
- semantic/normalized deduplication of shell commands,
- higher `max_tool_iterations`,
- lower `max_tool_iterations` plus better finalization behavior,
- or some other recommended configuration?

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** `detect_no_progress` used `take_while` over the reversed window, so it only counted a strictly consecutive run of same-tool same-result records. A single interleaved call of any other tool reset the streak to zero, letting an agent loop indefinitely by alternating a productive-looking call with the stuck one. It now counts matching records across the whole window instead of just the consecutive tail, matching the semantics the ping-pong and exact-repeat detectors already use.
- **Scope boundary:** Heuristic logic only inside `loop_detector.rs`. No change to how the agent loop reacts to a detection, to the escalation thresholds, or to the other two detectors' firing conditions.
- **Blast radius:** Agent runtime loop detector only (`crates/zeroclaw-runtime/src/agent/loop_detector.rs`). This is a runtime heuristic that can terminate agent turns early, so it is `risk: high`: a false positive ends a legitimate multi-step tool sequence one iteration sooner than before. It cannot loosen detection or affect any path outside the loop detector.
- **Linked issue(s):** Closes #7143
- **Labels:** `bug`, `risk: high`, `agent`, `runtime`, `tests`, `agent:loop`, `priority:p2`

## Root cause

This is the exact case in #7143: an agent ran 43 `shell` calls of near-duplicate repo-discovery commands (`git status` + `find` variants) against an unchanged tree, never editing, until `max_tool_iterations` was hit. The commands differed in text but returned identical output because the repo did not change, and they were interleaved with the occasional `file_read` / `content_search`, so the old consecutive-only detector never fired.

## What this catches

`detect_no_progress` now fires when, within the window: same tool, identical result, and at least two distinct argument hashes. That is precisely the #7143 shape: same `shell` tool, byte-identical output (no new information), different command phrasings. The `unique_args >= 2` guard distinguishes this from exact-repeat (handled separately) and targets the near-duplicate-but-not-identical commands the reporter described. Interleaved unrelated calls no longer reset the count.

## Changes

- `detect_no_progress`: `take_while` becomes `filter` over the window so non-consecutive repeats are caught
- Doc comments on both affected paths updated to state the cross-window behavior
- New regression test `no_progress_triggered_when_interleaved_with_other_calls`
- Corrected `exact_repeat_resets_on_different_call`, which encoded the old buggy assumption that any interleaving escapes detection

Run on this branch with latest `master` merged in. Full battery, literal tails.

**`cargo fmt --all -- --check`** — exit 0
```
(clean)
```

**`cargo clippy --all-targets -- -D warnings`** (entire workspace) — exit 0
```
    Checking zeroclaw-runtime v0.8.0
    Checking zeroclaw-hardware v0.8.0
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 44s
(0 warnings, 0 errors)
```

**`cargo test -p zeroclaw-runtime --lib loop_detector`** — exit 0
```
running 19 tests
test agent::loop_detector::tests::disabled_detector_always_returns_ok ... ok
test agent::loop_detector::tests::hash_value_is_key_order_independent ... ok
test agent::loop_detector::tests::hash_value_nested_key_order_independent ... ok
test agent::loop_detector::tests::exact_repeat_takes_priority_over_no_progress ... ok
test agent::loop_detector::tests::exact_repeat_warning_at_threshold ... ok
test agent::loop_detector::tests::exact_repeat_break_at_threshold_plus_two ... ok
test agent::loop_detector::tests::exact_repeat_resets_on_different_call ... ok
test agent::loop_detector::tests::no_progress_not_triggered_when_results_differ ... ok
test agent::loop_detector::tests::ping_pong_not_triggered_for_same_tool ... ok
test agent::loop_detector::tests::exact_repeat_block_at_threshold_plus_one ... ok
test agent::loop_detector::tests::ping_pong_warning_at_four_cycles ... ok
test agent::loop_detector::tests::no_progress_escalates_to_block_and_break ... ok
test agent::loop_detector::tests::ping_pong_detects_alternation_with_varying_args ... ok
test agent::loop_detector::tests::window_eviction_prevents_stale_pattern_detection ... ok
test agent::loop_detector::tests::window_size_limits_memory ... ok
test agent::loop_detector::tests::no_progress_warning_at_five_different_args_same_result ... ok
test agent::loop_detector::tests::no_progress_not_triggered_when_all_args_identical ... ok
test agent::loop_detector::tests::no_progress_triggered_when_interleaved_with_other_calls ... ok
test agent::loop_detector::tests::ping_pong_escalates_with_more_cycles ... ok
test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 2180 filtered out; finished in 0.00s
```

**`cargo test -p zeroclaw-runtime --lib`** (full runtime crate) — exit 0
```
test result: ok. 2198 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 5.89s
```

**`cargo test --workspace`** (entire workspace) — exit 0
```
TOTAL: 6383 passed; 0 failed; 20 ignored across 41 test binaries
```

- **Commands run and tail output:** all five above; fmt clean, workspace clippy 0 warnings, targeted 19/19, runtime crate 2198/0, whole workspace 6383/0.
- **Beyond CI — what did you manually verify?** Confirmed the new regression test reproduces the #7143 shape (same tool, identical result, interleaved unrelated calls, >= 2 distinct arg hashes) and that the corrected `exact_repeat_resets_on_different_call` no longer encodes the consecutive-only escape. Did not run a live multi-hour agent session against the original reporter's repo.

- If any `Yes`, describe the risk and mitigation: n/a. Internal loop-detection heuristic; no new surface, no data captured.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: none. Pure heuristic change inside the agent loop detector; no public API, config, or schema change.

- **Fast rollback command/path:** `git revert <sha>`; single self-contained commit touching only `loop_detector.rs`. No migration, no schema, no state to unwind.
- **Feature flags or config toggles:** `[pacing] loop_detection_enabled = false` disables the entire loop detector at runtime without a revert (the `record` path short-circuits to `Ok` when disabled, see `loop_detector.rs::record`). `loop_detection_window_size` and `loop_detection_max_repeats` tune sensitivity without disabling. This change adds no new flag; it rides the existing pacing toggles.
- **Observable failure symptoms:** A false-positive no-progress detection surfaces as an agent turn ending earlier than expected with a `loop_detector` no-progress message in the turn output (`"different arguments but identical results"` / `"no progress"`). Operators can grep turn traces / logs for `no progress` or `identical results` rising in frequency, or for agents that begin bailing out of legitimate multi-step tool sequences (especially repeated read-only discovery against an unchanged tree) sooner than before this change. If observed, set `loop_detection_enabled = false` to confirm the detector is the cause, then revert.


singlerider (commented):
I reviewed #7681 against the loop detector source on `upstream/master`, traced the three detectors and their escalation ordering, and ran the suite plus clippy locally.

### 🟢 Correct root-cause fix

The bug was real: `detect_no_progress` used `.rev().take_while(...)`, so a single interleaved unrelated call reset the same-tool-same-result streak to zero. An agent alternating `search` with any other tool could loop forever without tripping no-progress. Switching to `.filter(...)` counts matches across the whole window, which is the right semantics for "no progress" (the order of unrelated calls is irrelevant to whether this tool keeps returning identical results).

### 🟢 Exact-repeat boundary preserved

The `unique_args.len() < 2` guard still cleanly separates no-progress from exact-repeat, so the `filter` change does not cannibalize pattern 1. `detect_exact_repeat` remains consecutive-only (correct, since exact repeat is about hammering the identical call). Detection ordering (exact -> ping-pong -> no-progress) is unchanged, so the most-severe-first contract holds.

### 🟢 Window cap is safe

window_size is 20; the interleaved regression test produces 10 records, well within the cap, so no early probe is evicted before the 5th match. Escalation via count (`no_progress_escalates_to_block_and_break`) still reaches Block/Break correctly under the new counting.

### 🟢 Tests lock the behavior

`no_progress_triggered_when_interleaved_with_other_calls` directly encodes #7143 and fails against the old `take_while`. The pre-existing `exact_repeat_resets_on_different_call` was correctly tightened: it previously baked in the buggy assumption that interleaving escapes all detection, and now uses a distinct result hash so it isolates exact-repeat reset without tripping the now-correct no-progress path. 19/19 loop_detector tests pass, clippy `-D warnings` clean, fmt applied.

No blocking findings. Ready for review.

Audacity88:
@singlerider this looks like a useful narrow guardrail for the interleaving half of #7143. Counting same-tool / same-result records across the whole window should catch cases where unrelated tool calls previously reset the no-progress streak.

Before approval, please reconcile the closing claim with the issue evidence. The #7143 triage comment says the old detector missed the report partly because repo-discovery output can vary, so the identical-result hash may not hold. This patch still requires identical `result_hash`; it catches non-consecutive identical results, but not normalized or semantic near-duplicates with drifting output. Please either change the PR from `Closes #7143` to a narrower related/partial fix, or add evidence/regression coverage showing the reported #7143 records had identical result hashes and this detector would have fired.

Please also refresh the branch over current `master` before approval. There is no same-file churn, but this is 32 commits behind on runtime agent-loop behavior, so the final merge path needs fresh CI on the updated head or an explicit merge-result/stale-risk decision.

Separate maintainer cleanup I’d do after that: set labels to `risk: high` and `size: XS`, and update the PR body rollback line so it no longer says `risk: low`.

WareWolf-MoonWall (dismissed):
The core idea is correct — using `filter` instead of `rev().take_while()` means non-consecutive repeated calls across interleaved tool results are now detected. The new test `no_progress_triggered_when_interleaved_with_other_calls` validates the scenario that motivated the fix. Two items need to be addressed.

### 🔴 Blocking — Rollback section doesn't match the `risk: high` label

The PR is labeled `risk: high` (correctly — this is a runtime heuristic that can cause agent turns to terminate early). The Rollback section reads `git revert <sha>` with no observable failure symptoms and no feature-flag toggle — that's the low-risk template.

Per the PR template, `risk: medium` and `risk: high` PRs must fill:
- **Fast rollback command/path:** (present)
- **Feature flags or config toggles:** (missing — or explicitly `None` if there are none)
- **Observable failure symptoms:** (missing — what should an operator grep for if the detector fires incorrectly? Which metric or log line would indicate false-positive loop detection?)

Please complete the rollback section. For a no-progress loop detector, observable symptoms might include `loop_detector` spans in traces or log entries at `warn` level, or agents that begin bailing out of multi-step tool sequences sooner than before.

### 🟡 Warning — `Closes #7143` may overclaim the fix

Audacity88 raised this: the fix catches repeated calls with identical `result_hash`. If the originally-reported loop in #7143 involved tool output that varied slightly across iterations (timestamps, changing file state, stochastic output), the hash would differ and the detector still wouldn't fire. The PR body mentions "1738 identical tool calls with same args + same result" as the observed data — that's strong evidence for the hash-identity case, but it would be worth either:
- Changing to `Related to #7143` if the fix is a partial improvement, or
- Adding the 1738-record observation as a code comment near the fix, so the intent is preserved in context

### 🟢 What looks good — `filter` over `take_while`

Switching from `.rev().take_while(...)` to `.filter(...)` across the window correctly counts non-consecutive records without requiring them to be the most recent N. The semantics are intentional and the doc-comment update explains them. The `exact_repeat_resets_on_different_call` test fix (changing `r1` → `r_distinct`) is the right correction — the old test was accidentally passing for the wrong reason.

singlerider:
Addressed both items.

🔴 **Rollback section now matches `risk: high`.** Confirmed the label is `risk: high` (the body's old `risk: low` claim was wrong — corrected both the Labels line and the Blast radius framing). The Rollback section now fills all three required fields:
- Fast rollback: `git revert <sha>`, single self-contained commit.
- Feature flags/toggles: `[pacing] loop_detection_enabled = false` disables the detector at runtime without a revert (`record` short-circuits to `Ok` when disabled); `loop_detection_window_size` / `loop_detection_max_repeats` tune sensitivity. This change rides existing pacing toggles, adds none.
- Observable failure symptoms: turns ending early with `no progress` / `identical results` in the loop-detector message; grep traces for those strings rising, or agents bailing out of legitimate multi-step discovery sooner. Mitigation: flip `loop_detection_enabled = false` to confirm, then revert.

🟡 **`Closes #7143` kept, observation preserved in code.** The reporter's data ("1738 identical tool calls with same args + same result", and the 43 near-duplicate `shell` calls returning byte-identical output) is the hash-identity case this fix targets, so the closing link is accurate, not an overclaim. Per your second option, I added the observation as a terse why-comment at the `filter` site (commit `f24e49cd11`) so the intent survives in context.

fmt clean, `clippy -p zeroclaw-runtime --all-targets -D warnings` 0 warnings, 19/19 loop_detector tests pass.

perlowja:
Babysitting Shane's PRs while he's offline — status check on this one:

- **Mergeable**, all CI checks green.
- No unresolved review threads. Shane already replied "Addressed both items" to the earlier review note, so this looks ready pending **re-review** (`REVIEW_REQUIRED`) — no code changes needed.

(Of his three open PRs, only #7684 needed work — it had a real semantic conflict with master's #7775; resolution posted there.)

WareWolf-MoonWall (approved):
Re-review at latest head. Both items from my 2026-06-15 CHANGES_REQUESTED are resolved (my review was auto-dismissed when the remediation commits landed). Approving.

### ✅ Resolved — Rollback section complete for `risk: high`

The PR body now includes: `[pacing] loop_detection_enabled = false` runtime toggle; observable symptoms (agent turns ending early with "no progress" / "identical results" output; grep patterns); explicit confirm-and-revert procedure. The full risk:high rollback template is satisfied.

### ✅ Resolved — `Closes #7143` scope documented

The PR body honestly states the scope: the fix matches the #7143 shape (1738 identical tool calls with same args and same result) and includes a regression test that reproduces the pattern by name. The limitation (hash-identity detection may miss output that drifts slightly) is acknowledged.

### 🟢 What looks good — `filter` over `take_while`

The semantics are correct: `filter` counts non-consecutive matching records across the detection window, which is exactly what interleaved-tool-call loops require. `singlerider`'s positive assessment confirms the logic.

### 🟢 What looks good — Test corrections

`exact_repeat_resets_on_different_call` now uses `r_distinct` for the final result hash, which correctly isolates the exact-repeat test from the interleaved-detection path. The old test was passing for the wrong reason.

Audacity88 (approved):
@singlerider thanks for the update here.

I re-reviewed the refreshed head bd4d00f against the live PR state, #7143, the current one-file `loop_detector.rs` diff, the prior review comments and formal reviews, the PR body, the fresh CI run on the updated branch, and current `master` freshness. The earlier scope/risk/freshness concerns are resolved from my side.

### ✅ Resolved — Scope, rollback, and branch freshness are aligned

The PR body now matches the `risk: high` label, includes the runtime toggle and observable failure symptoms for rollback, and keeps the #7143 closing claim tied to the byte-identical-result shape this patch actually covers. The added why-comment at the `filter` site also preserves the important limitation: this is not semantic shell-command normalization, it is the non-consecutive identical-result case from the report.

The branch was refreshed over current `master`, and the updated head is current with base. The fresh CI run is green, including Lint, Test, Benchmarks Compile, and CI Required Gate.

### 🟢 What looks good — The detector now matches the interleaved-loop shape

Changing `detect_no_progress` from a consecutive tail scan to a whole-window filter fixes the blind spot that let unrelated tool calls reset the streak. The existing `unique_args` guard still keeps exact-repeat separate, so this does not collapse the three detector categories into each other.

### 🟢 What looks good — The regression test proves the old escape

`no_progress_triggered_when_interleaved_with_other_calls` exercises the important shape directly: same tool, identical result, different arguments, and unrelated calls interleaved between probes. The correction to `exact_repeat_resets_on_different_call` is also the right cleanup because it avoids accidentally tripping the newly-correct no-progress detector while testing exact-repeat behavior.

No blocking findings from me. Approving.

PR #7684, opened 2026-06-15
fix(acp): surface history-pruner and turn-cancel as visible events

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The history pruner collapsed old tool turns into a plain `assistant` message; the render surface printed it verbatim as bot output, so users could not tell the pruner emitted it. It now replays as a styled system event instead of bot prose.
  - Client-side cancels (`session/cancel`, `cancel_cause: client_rpc`) ended a turn with no final message and no explanation, indistinguishable from a hang. The turn now emits a visible `turn-cancelled` event and names the actual cause channel instead of hardcoding "turn cancelled by user".
  - Renamed the collapse marker to `[history-pruner: collapsed N tool call(s); results dropped from context]` and exposed `HISTORY_PRUNER_MARKER_PREFIX` as a shared const so the pruner's pairing detector and the render surface key off one value with no drifting literal.
  - New i18n key `turn-cancelled-client-rpc`; `turn-interrupted-by-user` is left untouched because `safety_net` pattern matching depends on its exact text.
- **Scope boundary:** Display/observability only. Does not change pruning policy, cancellation control flow, or when cancels fire. Does not touch the `turn-interrupted-by-user` string or `safety_net` matching.
- **Blast radius:** ACP Code-pane render surface (`acp_server.rs`), runtime history pruner, RPC dispatch/session, and the `en` CLI locale. No change to channel transports or agent reasoning.
- **Linked issue(s):** none
- **Labels:** `agent`, `channel`, `runtime`

## Root cause

The ACP session store recorded 1738 cancels this session, every one with payload `{"cancel_cause":"client_rpc"}`, with no adjacent user-input event. `client_rpc` names the transport channel a `session/cancel` arrived on, not the actor. `cancel_session` is hit by any client-side cancel: a new prompt superseding an in-flight turn, reconnect, client-side timeout, or pane teardown. The previous "by user" wording invented an actor the events never prove. The fix surfaces the cancellation by its actual cause channel and stops claiming a human pressed stop.

## What changed

- Renamed the collapse marker to `[history-pruner: collapsed N tool call(s); results dropped from context]` and exposed `HISTORY_PRUNER_MARKER_PREFIX` as a shared const so the pruner's own pairing detector and the render surface key off the same value with no drifting literal.
- The render surface replays the pruner marker as a styled `tool_call` update (`kind: think`, name `history-pruner`) instead of an `agent_message_chunk`. It is visibly a system event, not bot prose.
- On client cancellation, a styled `tool_call` (`turn-cancelled`) is emitted before the `cancelled` stopReason so the absence of a final response is explained instead of silent.
- `dispatch.rs`: cancel message now names the cause channel (`turn cancelled via client_rpc ...`) instead of asserting a user.
- `session.rs`: `CancelCause::ClientRpc` doc states the actor is unverified and lists the programmatic paths that hit it.
- New i18n key `turn-cancelled-client-rpc`. The shared `turn-interrupted-by-user` string is left untouched because `safety_net` pattern matching depends on its exact text.

No new dependencies. zerocode stays a thin wrapper; all logic lives in the runtime and channels crates that already have these deps. No string-literal `match`.

```bash
$ cargo fmt --all -- --check
    (clean)

    Finished (no warnings)

$ cargo test
test result: ok. 2171 passed; 0 failed
```

- **Commands run and tail output:** see above; fmt clean, clippy no warnings, full test suite 2171 passed / 0 failed.
- **Beyond CI — what did you manually verify?** Confirmed in a live Code pane that the pruner marker renders as a `history-pruner` system event (not bot prose), and that a client `session/cancel` now surfaces a visible `turn-cancelled` event naming the `client_rpc` cause. Did not exercise every non-`client_rpc` cancel cause path.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: none required. The collapse marker text changed, but it is internal render/pairing state, not a stable API; no config, env, or CLI surface is affected.

Low-risk display-only change: `git revert <sha>` is the plan.


Audacity88:
@singlerider this direction looks reasonable, but current master has moved across the same ACP/RPC surface this PR touches: `acp_server.rs`, `cli.ftl`, and `dispatch.rs`.

Could you refresh the branch over current `master` and let CI rerun? The merge looks textually clean, so this should be a straightforward freshness pass before the substantive review.

WareWolf-MoonWall (commented):
The root cause analysis is solid — quantifying 1738 observed spurious cancels and tracing them to silent `CancelCause::ClientRpc` is thorough work. The fix is narrow and the new Fluent key is properly introduced. A few things to address before merge.

### 🟡 Warning — Marker rename silently breaks existing persisted histories

The old pruner marker was `[Tool exchange: N tool call(s) — results collapsed]`; the new one is `[history-pruner: collapsed N tool call(s); results dropped from context]`. The updated `is_tool_exchange_summary` only matches the new prefix. Any conversation history persisted before this upgrade containing the old format will no longer be recognized as a pruner summary — the render surface will emit those old entries as `agent_message_chunk` (bot prose) instead of a collapse event.

The PR body acknowledges this is "internal render/pairing state, not a stable API," which is true — but users upgrading from a build with old histories will see previously-collapsed tool exchanges re-appear as bot prose in their history. This is worth explicitly documenting in the PR body as an accepted trade-off, and ideally noting in release notes. If the team would prefer a migration shim (checking both old and new prefix variants), that's also straightforward to add.

### 🟡 Warning — Commit 1 body contains a stale claim

Commit c902071851 says "Reuses the existing `turn-interrupted-by-user` i18n string" — but the final diff introduces the new `turn-cancelled-client-rpc` key. This stale claim should be cleaned up in the squash commit body at merge time. It's not a code defect but it would mislead future `git bisect` and `git log` readers.

### 🟡 Warning — Missing `risk:` label

The files changed — `acp_server.rs`, `dispatch.rs`, `session.rs` — fall in the medium-risk tier per AGENTS.md ("High risk: `crates/zeroclaw-runtime/src/**`, `crates/zeroclaw-channels/src/orchestrator/`"). The PR currently has no `risk:` label applied. A maintainer should set at least `risk: medium` before merge.

### 🟡 Warning — Branch behind master

Audacity88 has already flagged this — please rebase before requesting re-review. The changes in master to `acp_server.rs` and `cli.ftl` will need to be reconciled with the changes here.

### 🟢 What looks good — SSOT for the pruner marker

`HISTORY_PRUNER_MARKER_PREFIX` is the single const; both `is_tool_exchange_summary` (the pruner's own de-dup guard) and `history_notifications_for_message` (the render surface) key off it. That's the right factoring — no way for the two to drift against each other.

### 🟢 What looks good — Fluent key introduction

`turn-cancelled-client-rpc` is correctly introduced via `fl!()` and `get_required_cli_string()`. The deliberate non-translation of `turn-interrupted-by-user` (safety_net pattern dependency) is documented. Both decisions are correct.

perlowja:
Shane asked me to babysit his open PRs while he's offline, so I took a pass at this one's conflict. **Nothing pushed** — the branch is on his fork and I only have read access here, so this is a proposed resolution for him to review/apply in the morning.

### This wasn't a clean refresh
While #7684 was open, master merged **#7775** ("keep pruned tool markers out of live context"), which reworked the *same* pruned-history marker on the same ACP surface:

| | marker string | detection |
|---|---|---|
| **master (#7775)** | `[Tool exchange: N tool call(s) — results collapsed]` | `ChatMessage::is_pruned_tool_exchange_summary()` |
| **this PR** | `[history-pruner: collapsed N tool call(s); results dropped from context]` | `HISTORY_PRUNER_MARKER_PREFIX` in `acp_server.rs` |

git only flags 2 tiny textual hunks, so a naive "take master" merge **compiles and the unit test still passes** (the test builds its own marker string). But at runtime the pruner then emits master's marker while `acp_server.rs` still looks for this PR's marker — so the "replay history-pruner as a visible tool_call" feature **silently stops firing**. Green-looking, but a real regression.

### Proposed resolution
Unify on master's marker API (now referenced in 4+ places vs this PR's 2):
1. `history_pruner.rs` — keep master's `MAX_SYNTHETIC_TOOL_SUMMARIES`; drop the orphaned `HISTORY_PRUNER_MARKER_PREFIX` + `is_tool_exchange_summary` (master rewrote its callers to the `is_pruned_tool_exchange_summary()` method) and the now-unused inline `let summary`.
2. `acp_server.rs` — detect via `chat.is_pruned_tool_exchange_summary()`; test builds `ChatMessage::pruned_tool_exchange_summary(3)`; remove the dead import.

Verified locally: `cargo check` ✅ · `cargo clippy -- -D warnings` ✅ · `cargo test -p zeroclaw-channels history_pruner_marker_replays` ✅ (1 passed).

<details><summary>Conflict-resolution diff (<code>git showremerge-diff</code> — only the hand-resolved lines)</summary>

```diff
diff --git a/crates/zeroclaw-channels/src/orchestrator/acp_server.rs b/crates/zeroclaw-channels/src/orchestrator/acp_server.rs
@@ use
 use zeroclaw_runtime::agent::agent::{Agent, TurnEvent};
-use zeroclaw_runtime::agent::history_pruner::HISTORY_PRUNER_MARKER_PREFIX;
 use zeroclaw_runtime::tools::CanvasStore;
@@ fn history_notifications_for_message(
     match msg {
         ConversationMessage::Chat(chat) => {
-            if chat.role == "assistant" && chat.content.starts_with(HISTORY_PRUNER_MARKER_PREFIX) {
+            if chat.is_pruned_tool_exchange_summary() {
@@ mod tests / fn history_pruner_marker_replays_as_tool_call_not_agent_message
         use zeroclaw_api::model_provider::{ChatMessage, ConversationMessage};
-        let marker =
-            format!("{HISTORY_PRUNER_MARKER_PREFIX}3 tool call(s); results dropped from context]");
+        let marker = ChatMessage::pruned_tool_exchange_summary(3);
         let msg = ConversationMessage::Chat(ChatMessage::assistant(&marker));

diff --git a/crates/zeroclaw-runtime/src/agent/history_pruner.rs b/crates/zeroclaw-runtime/src/agent/history_pruner.rs
@@ struct PrunedOrphans
-<<<<<<< (this PR)
-pub const HISTORY_PRUNER_MARKER_PREFIX: &str = "[history-pruner: collapsed ";
-
-fn is_tool_exchange_summary(content: &str) -> bool {
-    content.starts_with(HISTORY_PRUNER_MARKER_PREFIX)
-        && content.contains("results dropped from context]")
-}
-=======
 const MAX_SYNTHETIC_TOOL_SUMMARIES: usize = 3;
->>>>>>> 883164ee (#7775)
@@ prune_history
                 if tool_count > 0 && !any_tool_protected {
-<<<<<<< (this PR)
-                    let summary = format!(
-                        "[history-pruner: collapsed {tool_count} tool call(s); results dropped from context]"
-                    );
-=======
->>>>>>> 883164ee (#7775)
                     messages[i] = ChatMessage {
                         role: "assistant".to_string(),
                         content: ChatMessage::pruned_tool_exchange_summary(tool_count),
```
</details>

### One decision for Shane
I unified on **master's** wording. This PR's thesis was the *clarity* of the `[history-pruner: collapsed …]` wording, and master overrode it. If you'd rather keep yours, don't take this as-is — change the two constants in `crates/zeroclaw-api/src/model_provider.rs` (`PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` / `_SUFFIX`) and keep everything pointed at the single `*_pruned_tool_exchange_summary` API. Don't reintroduce a second parallel marker — that's what caused the silent break.

### Apply it
Resolved merge commit `8f15709` (parents: `f42e55d5` + master `883164ee`). Full writeup, the `--remerge-diff`, and a `git bundle` (base64) are here: https://gist.github.com/perlowja/00112a11fa883e5b9deee76af38232ee

```bash
git fetch origin master
# decode the bundle from the gist, then:
git fetch ./pr7684-refresh.bundle pr7684-merged
git checkout fix/history-pruner-marker-clarity
git reset --hard pr7684-merged    # then: git commit --amend --reset-author
git push origin fix/history-pruner-marker-clarity
```

Nillth (dismissed):
I took stock of everything on this PR before writing: @Audacity88's freshness request (2026-06-15), @WareWolf-MoonWall's COMMENTED review (2026-06-15, no inline threads attached), and @perlowja's detailed babysitting writeup (2026-06-16) proposing a conflict resolution on @singlerider's behalf. I also re-checked the diff against the actual PR-HEAD source at c03170123813122f756a6d281fc7473cf2f7a6b5 and against current master, since master has moved 49 commits ahead and the two have diverged. The root-cause work here is genuinely good — quantifying 1738 `client_rpc` cancels with no adjacent user-input event and refusing to invent an actor the events never prove is exactly the kind of evidence-first reasoning we want. The display/observability framing is right, the i18n discipline is correct, and the legacy-marker follow-up commit c03170123813 closed @WareWolf-MoonWall's first warning. The problem is entirely in how this lands on top of #7775, which merged into master while this PR was open and reworked the very same marker on the same surface. That collision is now the gate.

### 🔴 Blocking — Duplicate pruner-marker SSOT collides with #7775; rebase must unify, not re-fork

While this PR was open, #7775 ("keep pruned tool markers out of live context") landed on master and established the single source of truth for the pruned-tool-exchange marker in `crates/zeroclaw-api/src/model_provider.rs:27-28,60-77`: `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` / `PRUNED_TOOL_EXCHANGE_SUMMARY_SUFFIX`, the `ChatMessage::pruned_tool_exchange_summary()` constructor, and the `ChatMessage::is_pruned_tool_exchange_summary()` predicate. master's `crates/zeroclaw-runtime/src/agent/history_pruner.rs` now routes every emit and every detection through that API (the collapse write at line ~297 calls `ChatMessage::pruned_tool_exchange_summary(tool_count)`, and detection at lines 154/375/381/398-399/420/426 calls `is_pruned_tool_exchange_summary()`).

This PR's HEAD (c03170123813) instead introduces a *second*, parallel marker authority in `crates/zeroclaw-runtime/src/agent/history_pruner.rs:69`: `pub const HISTORY_PRUNER_MARKER_PREFIX: &str = "[history-pruner: collapsed "` plus a hand-typed legacy literal `const LEGACY_TOOL_EXCHANGE_PREFIX: &str = "[Tool exchange:"`, with its own `is_tool_exchange_summary` / `is_history_pruner_marker` pair. I confirmed the PR HEAD does not reference `model_provider`'s `PRUNED_TOOL_EXCHANGE_*` API at all. Merging as-is leaves the same fact — what string the pruner stamps and how it is recognized — defined in two crates that can drift independently. That is the exact pattern AGENTS.md opens with under "ABSOLUTE RULE — SINGLE SOURCE OF TRUTH (NO DRY VIOLATIONS)" and FND-001 §4 warns "will drift apart silently over time." It is not deferrable to a follow-up.

@perlowja's analysis is the right read and worth honoring: the resolution should unify on master's `*_pruned_tool_exchange_summary` API, drop `HISTORY_PRUNER_MARKER_PREFIX` and the duplicate detector, and point `acp_server.rs` at `chat.is_pruned_tool_exchange_summary()`. One correction to the "silently stops firing" claim in that writeup, now that c03170123813 exists: this PR's legacy branch matches `starts_with("[Tool exchange:") && contains("results collapsed]")`, and #7775's `pruned_tool_exchange_summary` emits `[Tool exchange: N tool call(s) — results collapsed]`, which *does* satisfy that predicate — so post-merge the render feature would actually still fire, via the legacy path, rather than die. That makes the failure mode quieter, not safer: the feature would survive by accidental coincidence of two hardcoded literals across crate boundaries, which is precisely the drift trap. So the blocker stands on the duplication itself, independent of whether the feature happens to keep working.

### 🟡 Warning — The PR's headline clarity wording becomes dead on master

The stated thesis of this change is the *clarity* of `[history-pruner: collapsed N tool call(s); results dropped from context]` over the terse old marker. But master's pruner (post-#7775) emits `[Tool exchange: N tool call(s) — results collapsed]`, and #7775 is now the SSOT. If the rebase unifies on master's API (the only DRY-clean option), the new wording never reaches a live session — the pruner stamps master's string and `acp_server.rs` renders that. This is a Team decision @singlerider should make explicitly during the rebase, and it deserves a line in the PR body: either (a) accept master's wording and drop the rename from this PR's scope (the observability fix — replaying the marker as a styled `tool_call` instead of `agent_message_chunk` — stands on its own and is the real value here), or (b) if the team prefers this PR's wording, change the two constants in `crates/zeroclaw-api/src/model_provider.rs` and keep everything pointed at the single API. As @perlowja put it: do not reintroduce a second parallel marker.

### 🟡 Warning — Branch is conflicting with master and must be rebased before it can merge

mergeable is CONFLICTING / mergeStateStatus is DIRTY. @Audacity88 flagged the freshness need on 2026-06-15 and the in-flight merge f42e55d5 predates the relevant master movement, so the conflict has reopened. Beyond the marker collision above, I confirmed three more textual conflict points the rebase must reconcile: `crates/zeroclaw-runtime/locales/en/cli.ftl` (both this PR and master append right after `turn-interrupted-by-user`, master line 752); `crates/zeroclaw-runtime/src/rpc/dispatch.rs:1563-1570` (master still carries the `"turn cancelled by user in RPC_SESSION"` arm this PR rewrites — this PR's change is the desired one, just needs re-applying over master); and `crates/zeroclaw-runtime/src/rpc/session.rs:21` (master still has the terse `ClientRpc` doc this PR expands). None of these three are substantive — they resolve in the PR's favor — but they must be cleared. This conflict is why I cannot recommend approval regardless of substance.

### 🔵 Suggestion — Tighten the legacy-marker comment to name #7775

`history_pruner.rs:71` says "Pre-#7684 marker." Once you rebase onto #7775, that framing is slightly off: the `[Tool exchange:` form is not merely "pre-#7684 persisted state" — it is the *current* on-master marker that #7775 owns. If you keep any legacy-detection logic at all after unifying on master's API, anchor the comment to #7775 so a future reader does not mistake the live marker for a deprecated one.

### 🟢 What looks good — Evidence-first cancel attribution

The `CancelCause::ClientRpc` doc rewrite in `session.rs` and the `dispatch.rs` cancel-message change are exactly right: `client_rpc` names the transport channel a `session/cancel` arrived on, not the actor, and "a new prompt superseding an in-flight turn, reconnect, client-side timeout, pane teardown all land on this path" is the correct enumeration. Refusing to stamp "by user" on events that never prove a user is the kind of correctness-under-uncertainty that prevents a whole class of misleading UX. The 1738-cancel empirical basis in the commit body is the right way to justify a wording change.

### 🟢 What looks good — The observability fix is the right boundary and the right shape

Routing both the pruner marker and the cancel through the ACP render boundary as `tool_call` updates (`kind: think`, `status: completed`) instead of `agent_message_chunk` is the correct call — a system event should not masquerade as bot prose, and this keeps presentation concerns in `acp_server.rs` (the channel/render surface) rather than leaking them into the runtime pruner. The unit coverage is real, not theatrical: `turn_cancelled_notification_is_styled_tool_call` and `history_pruner_marker_replays_as_tool_call_not_agent_message` both assert the actual `sessionUpdate` shape and that a plain reply still renders as `agent_message_chunk`, and the legacy-marker assertion guards the persisted-history path. The Fluent discipline is also correct — `turn-cancelled-client-rpc` introduced via `fl!()`/`get_required_cli_string()`, and the deliberate non-translation of `turn-interrupted-by-user` (safety_net pattern dependency) is documented rather than accidental.

### ✅ Resolved — Legacy persisted-history markers no longer leak as bot prose

@WareWolf-MoonWall's first warning — that the marker rename would make pre-upgrade persisted histories re-appear as bot prose — was addressed by commit c03170123813 (2026-06-17), which added `is_history_pruner_marker` with backward-compatible detection of the `[Tool exchange:` form plus coverage for legacy-marker detection and legacy-marker replay. Within this PR's own world the concern is handled. (The caveat is the blocker above: that legacy literal duplicates #7775's SSOT, so the right end-state folds this detection into `is_pruned_tool_exchange_summary()` rather than keeping a separate predicate.)

### ✅ Resolved — risk: medium and scope labels are present

@WareWolf-MoonWall's "Missing risk: label" warning is now stale — the PR carries `risk: medium`, `size: S`, `agent`, `channel`, and `runtime`. One note for maintainers, not a blocker and not something I'm changing: AGENTS.md "Risk Tiers" classifies `crates/zeroclaw-runtime/src/**` as **High risk**, so the auto-applied `risk: medium` is arguably one tier light. The display-only blast radius is a fair mitigating factor, but if a maintainer wants to honor "when uncertain, classify as higher risk," `risk: high` would be defensible. Either way the rollback plan (`git revert`) is appropriate for the change's actual reach.

---

**Verdict: request changes.** The substance is strong and the observability fix is genuinely worth landing — but this PR introduces a duplicate marker SSOT that collides head-on with #7775 now on master, which is a hard AGENTS.md / FND-001 violation, not a stylistic nit. The branch is also CONFLICTING/DIRTY and must be rebased onto master regardless. The path forward is clean and already mapped by @perlowja: rebase onto master, unify on `model_provider`'s `*_pruned_tool_exchange_summary` API (one marker authority, not two), make the explicit team call on whether the new clarity wording survives or master's wording wins, and keep the `acp_server.rs` styled-`tool_call` replay plus the cancel-attribution fix — those are the real value and they stand on their own. I'd happily re-review and approve once the rebase lands on a single marker source.

WareWolf-MoonWall (dismissed):
# Review — PR #7684 · Round 12

**Reviewed at**: c03170123813122f756a6d281fc7473cf2f7a6b5
**Prior review commit**: f42e55d56ff165ec1aa7ab559ba24f52ea99722f (2026-06-15T23:31:44Z, state: COMMENTED)
**Verdict**: `--request-changes`

---

## Status check

Two commits have landed since Round 11 (`f42e55d5`): a merge-master commit and a new
fourth commit "detect legacy pruner marker on persisted sessions". The merge conflict is **still
present**: GitHub API reports `mergeable: false`, `mergeable_state: dirty`.

perlowja's comment from ~6 hours ago gives a precise diagnosis and a proposed resolution —
read it carefully before touching anything. Summary of what I see below.

---

## Carry-forward from Round 11

### ✅ Resolved — `risk: medium` label applied

The label is now on the PR. Requirement met.

### ✅ Resolved — Legacy persisted-history backward compat added

`LEGACY_TOOL_EXCHANGE_PREFIX = "[Tool exchange:"` and dual-format detection in
`is_tool_exchange_summary` mean sessions persisted before this upgrade are not
re-emitted as raw assistant text. Good work addressing Round 11 concern #1.

---

## New / remaining findings

### 🔴 Blocking — Branch is unmerged and the marker-API conflict is unresolved

The GitHub API returns `mergeable: false / dirty` for this branch. The merge commit
(commit 3, "Merge branch 'master' into fix/history-pruner-marker-clarity") brought master
in but **did not reconcile the competing marker APIs**:

| | marker emitted by `prune_history` | detection |
|---|---|---|
| master (#7775, already landed) | `ChatMessage::pruned_tool_exchange_summary(n)` → `[Tool exchange: N tool call(s) — results collapsed]` | `ChatMessage::is_pruned_tool_exchange_summary()` |
| this PR (post-merge) | `format!("[history-pruner: collapsed …]")` via `HISTORY_PRUNER_MARKER_PREFIX` | `is_history_pruner_marker()` |

Local master's `history_pruner.rs` is already on the `ChatMessage::pruned_tool_exchange_summary()`
call (line ~297). The PR diff removes the pre-#7775 inline `format!("[Tool exchange: …]")` and
adds a new inline `format!("[history-pruner: …]")`. These target different merge bases and
will produce a textual conflict.

The fourth commit ("detect legacy pruner marker") adds `LEGACY_TOOL_EXCHANGE_PREFIX` so that
`[Tool exchange: …]` is still detected and replayed as a `tool_call`. That's a reasonable
symptom-fix, but the underlying conflict is not resolved: the PR still adds
`HISTORY_PRUNER_MARKER_PREFIX` to `history_pruner.rs` alongside master's
`PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` in `model_provider.rs`, leaving two independent
marker-prefix constants for the same concept. See the SSOT finding below.

**Required action**: Rebase the branch onto current master, then choose one of the paths
perlowja spelled out:

- **Path A (recommended — less churn)**: Keep master's `ChatMessage::pruned_tool_exchange_summary()`
  for emission. In `acp_server.rs` use `chat.is_pruned_tool_exchange_summary()` for detection
  instead of `is_history_pruner_marker`. In the test, build the marker via
  `ChatMessage::pruned_tool_exchange_summary(3)`. Drop `HISTORY_PRUNER_MARKER_PREFIX` and the
  now-redundant `is_history_pruner_marker` wrapper. The legacy detection (`LEGACY_TOOL_EXCHANGE_PREFIX`)
  becomes unnecessary too since master's API already targets that same prefix.

- **Path B (keeps the new wording)**: Change the two constants in `model_provider.rs`
  (`PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` / `_SUFFIX`) to the new `[history-pruner: collapsed …]`
  wording. Update `is_pruned_tool_exchange_summary()` to match. Add legacy detection there (not
  in `history_pruner.rs`). Use `chat.is_pruned_tool_exchange_summary()` in `acp_server.rs`. Drop
  `HISTORY_PRUNER_MARKER_PREFIX`.

Either path produces a single canonical marker API. Do NOT ship with both
`HISTORY_PRUNER_MARKER_PREFIX` and `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` live.

### 🔴 Blocking — `HISTORY_PRUNER_MARKER_PREFIX` duplicates `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` in `model_provider.rs` (SSOT violation)

After the merge, the repository would contain:
- `crates/zeroclaw-runtime/src/agent/history_pruner.rs`: `pub const HISTORY_PRUNER_MARKER_PREFIX`
- `crates/zeroclaw-api/src/model_provider.rs`: `pub const PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX`

Both constants represent "what string a pruner marker starts with." Two constants = two places
to update if the format ever changes again = the exact drift pattern AGENTS.md forbids.
The resolution paths above eliminate one of them. Pre-edit ritual: the source of truth for the
marker string is `model_provider.rs` (where `pruned_tool_exchange_summary()` and
`is_pruned_tool_exchange_summary()` already live). `HISTORY_PRUNER_MARKER_PREFIX` in
`history_pruner.rs` would be a duplicate and must not be introduced.

### 🟡 Warning — `turn-cancelled-client-rpc` Fluent key is missing from all non-English locales

`en/cli.ftl` has the new key. The other four locale files (`es`, `fr`, `ja`, `zh-CN`) do
not. AGENTS.md governance rule #4 is explicit: new user-facing Fluent keys must be present in
all five locales. Stubs using the English text are acceptable as placeholders until translators
update them. Please add:

```ftl
# es/cli.ftl, fr/cli.ftl, ja/cli.ftl, zh-CN/cli.ftl
turn-cancelled-client-rpc = [turn cancelled via client]
```

(Note: Round 11 praised the Fluent introduction but did not catch the missing non-en files.
This is a newly identified gap, not a carry-forward.)

### 🟡 Warning — Stale claim in commit `c902071851` body (Round 11 carry-forward)

The commit body says "Reuses the existing `turn-interrupted-by-user` i18n string" but the final
diff introduces the new `turn-cancelled-client-rpc` key. Clean this up in the squash commit
message at merge time to avoid misleading `git bisect` and `git log` readers.

---

## What is working well

### 🟢 What looks good — `turn_cancelled_notification` shape is correct

The notification correctly emits `sessionUpdate: "tool_call"` with `kind: "think"` and a
non-empty Fluent-sourced text. The test covers the shape (not just the existence of the
notification) and verifies the Fluent lookup is wired. Good test.

### 🟢 What looks good — `CancelCause::ClientRpc` doc update is precise

Stating that the actor is unverified and enumerating the programmatic paths (new prompt
superseding an in-flight turn, reconnect, client-side timeout, pane teardown) is exactly
the right level of detail. This will save the next engineer who reads this enum from
repeating the debugging session that motivated the PR.

---

## Summary of actions required before re-review

1. Rebase onto current master (resolve the `mergeable: dirty` state).
2. Reconcile with master's `ChatMessage::pruned_tool_exchange_summary()` / `is_pruned_tool_exchange_summary()` API using Path A or Path B above. Do NOT leave both `HISTORY_PRUNER_MARKER_PREFIX` and `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` live.
3. Add `turn-cancelled-client-rpc` stubs to all four non-English locale files.
4. Clean up commit `c902071851`'s stale body claim at squash time.

Once those four items are addressed, this is ready for re-review.

tidux (approved):
Walked the current head (52395d5a) against current master, confirmed both prior CHANGES_REQUESTED concerns (the duplicate marker SSOT, the CONFLICTING/DIRTY state) are resolved on the merits, ran the new regression tests locally, and took stock of every prior reviewer's findings against what the squashed result will actually ship.

### ✅ Resolved — Duplicate pruner-marker SSOT (Path A landed)

@Nillth's and @WareWolf-MoonWall's Round-12 🔴 was the headline block: `HISTORY_PRUNER_MARKER_PREFIX` in `history_pruner.rs` duplicated #7775's `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` in `model_provider.rs`. On the current head that duplication is gone.

Direct source check on the post-merge tree:

- `crates/zeroclaw-api/src/model_provider.rs:27-28,60-77` remains the single marker authority — `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX/_SUFFIX`, `ChatMessage::pruned_tool_exchange_summary()`, `ChatMessage::is_pruned_tool_exchange_summary()`.
- `crates/zeroclaw-runtime/src/agent/history_pruner.rs` calls `ChatMessage::pruned_tool_exchange_summary(tool_count)` at the collapse-write site (line 297) and `is_pruned_tool_exchange_summary()` at every detection site (154, 375, 381, 398-399, 420, 426).
- `crates/zeroclaw-channels/src/orchestrator/acp_server.rs:2031` detects via `chat.is_pruned_tool_exchange_summary()` and the regression test at 3392-3398 builds its marker via `ChatMessage::pruned_tool_exchange_summary(3)`.
- `grep -r HISTORY_PRUNER_MARKER_PREFIX crates/` returns nothing.

That is Path A from @perlowja's writeup, exactly as @Nillth and @WareWolf-MoonWall asked for. The branch history still contains intermediate commits (c902071851, c031701238) that introduced the parallel constant, but the squashed result will be the displayed +129/-10 diff which has no parallel marker authority at all. The Squash-merge is the right shape for this PR.

### ✅ Resolved — Branch is mergeable

The CONFLICTING/DIRTY state both Round-12 reviews flagged is gone. Current head reports MERGEABLE with mergeStateStatus BLOCKED only on review-required. All 17 Quality Gate checks are green.

### ✅ Resolved — `risk: medium` label applied

Carried forward from @WareWolf-MoonWall's Round-11 🟡, already cleared at Round-12, kept ✅ here for the audit trail. @Nillth's note that AGENTS.md tier-classifies `crates/zeroclaw-runtime/src/**` as high and the auto-applied `risk: medium` is arguably one tier light is a fair maintainer call. The display-only blast radius (no policy, no control flow, no transport change) is a defensible mitigating factor and the `git revert <sha>` rollback plan is appropriate for the change's real reach. Not pushing for a bump from `medium` → `high` myself.

### ✅ Resolved — Legacy persisted-history compatibility

@WareWolf-MoonWall Round-11 🟡 about pre-upgrade persisted histories leaking as bot prose is naturally closed by Path A: master's `is_pruned_tool_exchange_summary()` predicate matches `[Tool exchange: …]`, and since master's `pruned_tool_exchange_summary()` constructor also emits `[Tool exchange: …]`, there is no rename and no legacy gap. The `legacy_tool_exchange_marker_still_detected` test in `history_pruner.rs:837` pins both shapes (constructor-built marker and a hand-typed legacy literal) under the same predicate, so the contract is asserted rather than implied.

### 🟢 What looks good — Evidence-first cancel attribution

The `CancelCause::ClientRpc` doc rewrite at `session.rs:21-25` is the right level of detail — it enumerates the programmatic paths that land on the `client_rpc` channel (new prompt superseding an in-flight turn, reconnect, client-side timeout, pane teardown) and explicitly says the actor is not verified here. The 1738-cancel empirical basis behind the change is the right way to justify a wording change about user-facing attribution.

The `dispatch.rs` cancel-message rewrite at 1561-1576 actually does one better than the PR summary describes: it generalizes the "name the cause channel" treatment from `ClientRpc` to all `CancelCause` variants. Pre-PR the message was bifurcated between "by user in RPC_SESSION ..." (ClientRpc) and "by daemon: ..." (everything else). Post-PR every cause uses the same shape — `"turn cancelled via {cause} in RPC_SESSION ..."` — which preserves the cause information that the old "by daemon" prefix flattened and drops the fictional actor that the old "by user" prefix invented. The unattributed `None` branch becomes `"turn cancelled (cause unattributed) in RPC_SESSION ..."`, which is more honest than the old `"turn cancelled by daemon: unattributed"` (the daemon mediates either way; the unattributed status is what the message should say).

### 🟢 What looks good — The observability fix is the right boundary and the right shape

Routing both the pruner marker and the cancel through the ACP render boundary as `tool_call` updates (`kind: think`, `status: completed`) instead of `agent_message_chunk` keeps presentation concerns in `acp_server.rs` (the channel/render surface) rather than leaking them into the runtime pruner or `dispatch.rs`. That's the correct architectural placement — a system event should not masquerade as bot prose, and the runtime should not need to know how the ACP renders pruner artifacts.

The unit coverage is real, not theatrical. `turn_cancelled_notification_is_styled_tool_call` asserts the actual `sessionUpdate` shape — `tool_call` + `name: "turn-cancelled"` + `status: completed` + non-empty Fluent-sourced text — and `history_pruner_marker_replays_as_tool_call_not_agent_message` asserts (1) the marker replays as `tool_call`, (2) a plain reply still renders as `agent_message_chunk`, and (3) the legacy `[Tool exchange:` literal also replays as `tool_call`. The Fluent discipline is intact — `turn-cancelled-client-rpc` enters via `get_required_cli_string("turn-cancelled-client-rpc")`, and the deliberate non-translation of `turn-interrupted-by-user` (safety_net pattern dependency) is left untouched as the PR body documents.

### 🟡 Warning — Stale claims in PR body, summary, and commit body now that Path A landed

@WareWolf-MoonWall flagged this twice already and it is still relevant; I want to extend it to include the PR description and summary since Path A moved the goalposts.

Three pieces of text on this PR no longer match what the squashed result actually ships, and a future `git bisect` or `git log` reader will be misled if they go in unchanged at merge time:

- Commit c902071851 body says "Reuses the existing `turn-interrupted-by-user` i18n string" but the final diff introduces `turn-cancelled-client-rpc`. (Wolf's Round-11/Round-12 carry-forward.)
- The PR summary's "What changed and why" still reads "Renamed the collapse marker to `[history-pruner: collapsed N tool call(s); results dropped from context]`". That rename did not survive Path A — `crates/zeroclaw-api/src/model_provider.rs:60-63` builds `[Tool exchange: N tool call(s) — results collapsed]`, which is master's wording, not this PR's wording.
- The PR title itself ("surface history-pruner and turn-cancel as visible events") is still accurate as a description of the observability change, but the body's headline-clarity claim about the new wording does not match the tree.

None of these are code defects — they are squash-merge body hygiene. The clean version of the squash message is something like "Replays the pruned-tool-exchange marker (master's `[Tool exchange: …]` wording from #7775, unchanged) as a styled `tool_call` instead of bot prose; replays the client-RPC cancel as a styled `tool_call` and stops attributing programmatic cancels to a user; introduces `turn-cancelled-client-rpc` Fluent key for the user-visible cancel marker."

### 🔵 Suggestion — `turn-cancelled-client-rpc` locale-parity is a wash with codebase precedent

@WareWolf-MoonWall Round-12 🟡 asked for `turn-cancelled-client-rpc` stubs in `es/cli.ftl`, `fr/cli.ftl`, `ja/cli.ftl`, `zh-CN/cli.ftl`. I went to apply that finding mechanically and stopped when I checked sibling state:

```sh
$ for loc in en es fr ja zh-CN; do
    echo "=== $loc ==="
    grep -E "^turn-" "crates/zeroclaw-runtime/locales/$loc/cli.ftl"
  done
=== en ===
turn-interrupted-by-user = [interrupted by user]
turn-cancelled-client-rpc = [turn cancelled via client]
turn-stream-interrupted = [stream interrupted]
turn-tool-interrupted-before-result = [interrupted by user before this tool produced a result]
=== es ===
=== fr ===
=== ja ===
=== zh-CN ===
```

The entire `turn-*` family is en-only by existing codebase precedent — none of `turn-interrupted-by-user`, `turn-stream-interrupted`, `turn-tool-interrupted-before-result` appears in any non-en `cli.ftl`. Adding `turn-cancelled-client-rpc` only in `en/` is byte-consistent with the sibling rule the PR explicitly does not touch. There is also a runtime caveat in `docs/book/src/maintainers/docs-and-translations.md:44` that "Only `en` and `zh-CN` are wired into the runtime as built-ins" — even the existing `zh-CN/cli.ftl` does not carry the `turn-*` family today.

Two things both look right to me here: (1) keep the en-only landing on this PR so it stays consistent with the `turn-*` siblings, and (2) clean up the en-only `turn-*` family as a separate `cargo fluent fill` follow-up that picks up every `turn-*` at once rather than just `turn-cancelled-client-rpc`. If the team would rather lock parity at the per-key level instead, the `cargo fluent fill --locale es,fr,ja,zh-CN` pass for this one key is straightforward — but I would not block on it here.

Marking as 🔵 rather than carrying Wolf's 🟡 forward because the parity reading is at odds with how the existing siblings actually land.

### 🟢 What looks good — Approachable resolution of two competing reviews

The Path-A resolution honored @perlowja's writeup, addressed @Nillth's and @WareWolf-MoonWall's Round-12 🔴s on the merits rather than via a feature concession, and kept the actual value of this PR (cancel-attribution honesty + the observability fix) intact. The substantive work — the empirical 1738-cancel basis, the doc clarification, the `tool_call` replay routing — was always the load-bearing part of the PR; what got dropped along the way (the marker rename) was the bit that collided head-on with #7775. The end state is cleaner than either path the PR started with.

---

Validation on head 52395d5a: `cargo fmt --all -- --check` clean; `cargo test -p zeroclaw-channels --lib -- history_pruner_marker_replays turn_cancelled_notification_is_styled` 2 passed (both new ACP regression tests); `cargo test -p zeroclaw-runtime --lib -- legacy_tool_exchange_marker_still_detected` 1 passed; full Quality Gate is already green at 17/17. PR body's full `cargo test` 2171-pass claim reproduces on the kind of slice I checked.

Marking `--approve`. Both prior CHANGES_REQUESTED reviews from @Nillth and @WareWolf-MoonWall were dismissed by @singlerider with a dismissal message citing Path A as the resolution; I ground-truthed that claim directly via source inspection (the SSOT duplicate constant is gone from the tree, every detection site goes through master's API, the new regression tests assert the wire contract) before approving. The substantive merits side is good, the SSOT block is resolved on the merits (not papered over), and the remaining 🟡 is squash-message hygiene that the maintainer running the merge button can pick up at merge time. The 🔵 on locale parity is author-or-maintainer choice. Happy to see this land.

Audacity88 (dismissed):
I checked the current public PR head, diff/files API, prior public reviews/comments, green checks, and the ACP/source excerpts in the public evidence bundle; no private/local validation evidence is part of this review.

PR #7684 cleans up the history-pruner replay path and the RPC cancel attribution, but I still see one ACP terminal surface carrying the old user-attributed marker.

### 🔴 Blocking — ACP cancel result still returns the user-attributed marker

`session/cancel` now emits a styled `turn-cancelled` update with `[turn cancelled via client]`, but `cancelled_prompt_result()` still appends `turn-interrupted-by-user` to the terminal `session/prompt` result content. The ACP channel documents that this `content` field is still included for existing clients, and the cancel example returns accumulated text plus that marker. Even if preserving the old prompt-result shape is intentional compatibility, it is still a visible cancellation surface for those clients, so it should not keep the invented user actor without an explicit compatibility caveat. Please make the cancelled ACP result content use the same client-channel attribution, or otherwise make the compatibility behavior explicit and non-misleading, and update the preserved-shape test/docs so both visible cancellation surfaces agree.

### 🟢 What looks good — History-pruner replay uses the shared marker API

The history-pruner side now reuses `ChatMessage::pruned_tool_exchange_summary()` / `is_pruned_tool_exchange_summary()` instead of adding a second marker source, and the replay test covers both current and legacy collapsed summaries.

singlerider:
@Audacity88 thanks — fixed in ce2a094b.

You were right that the two visible cancellation surfaces disagreed on the actor. `session/cancel` emitted the styled `turn-cancelled` update with `[turn cancelled via client]`, but `cancelled_prompt_result()` still appended `turn-interrupted-by-user` (`[interrupted by user]`) to the terminal `session/prompt` result content, asserting a user action the cancel path does not verify.

`cancelled_prompt_result()` now routes through `turn-cancelled-client-rpc`, so both surfaces report the same client-channel attribution. Updated `cancelled_prompt_result_preserves_content_string_shape` and the `acp.md` cancel example (`content: "partial...\n\n[turn cancelled via client]"`) so the preserved-shape contract and the docs agree.

The runtime interrupt path (`ws.rs` / `safety_net.rs` / `agent.rs` / `turn`) deliberately keeps `turn-interrupted-by-user` — those are genuine user-initiated interrupts (stop), where the actor really is the user. The marker split is now: client-RPC cancel → `[turn cancelled via client]`, user interrupt → `[interrupted by user]`.

`cargo test -p zeroclaw-channels --features channel-acp-server cancel` green (11/11), fmt clean.

WareWolf-MoonWall (changes requested):
# Re-Review — PR #7684 · fix(acp): surface history-pruner and turn-cancel as visible events

**Head SHA:** 52395d5a6c1a050a1a231ac470d3baf4b2205700
**Verdict:** CHANGES_REQUESTED (`--request-changes`)
**Prior verdict:** COMMENTED at f42e55d5 (four 🟡 warnings); prior review dismissed at c0317012
**Active CHANGES_REQUESTED from other reviewers:** Audacity88 (active at 52395d5a)

---

Re-reviewed against the rebased head 52395d5a. All four of my Round-1 🟡 warnings are resolved, the SSOT violation that Nillth and I both identified in the c0317012 pass has been eliminated by adopting master's `is_pruned_tool_exchange_summary()` API, and the branch is now clean. However, I independently agree with Audacity88's 🔴: `cancelled_prompt_result()` still carries the user-attributed marker on the ACP prompt-result path, and the PR leaves that surface half-fixed. That is a concrete blocker from my side as well.

---

### ✅ Resolved — Branch is now clean (mergeable: true)

My Round-1 🟡 and the subsequent merged-pass concern were both about the branch being behind master with non-trivial surface overlap in `acp_server.rs`, `dispatch.rs`, and `cli.ftl`. The API reports `mergeable: true` at 52395d5a. ✓

---

### ✅ Resolved — SSOT violation eliminated; PR uses master's pruner-marker API

My prior-pass DISMISSED review (and Nillth's concurrent finding) identified a duplicate-state violation: this PR's `HISTORY_PRUNER_MARKER_PREFIX` in `history_pruner.rs` existed independently alongside master's `PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX` / `PRUNED_TOOL_EXCHANGE_SUMMARY_SUFFIX` constants and `ChatMessage::is_pruned_tool_exchange_summary()` in `model_provider.rs` — two parallel constants for the same concept on the same surface.

At 52395d5a, `HISTORY_PRUNER_MARKER_PREFIX` is gone. `acp_server.rs::history_notifications_for_message` now detects via `chat.is_pruned_tool_exchange_summary()`, and the test builds the marker via `ChatMessage::pruned_tool_exchange_summary(3)`. No second constant, no drift path. ✓

---

### ✅ Resolved — Legacy pruner marker is detected and replays correctly

My Round-1 🟡 about persisted histories silently regressing: the legacy format `[Tool exchange: N tool call(s) — results collapsed]` is now covered. `is_pruned_tool_exchange_summary()` checks `starts_with(PRUNED_TOOL_EXCHANGE_SUMMARY_PREFIX)` (i.e. `"[Tool exchange:"`) and `contains(PRUNED_TOOL_EXCHANGE_SUMMARY_SUFFIX)` (i.e. `"results collapsed]"`), so both the current API-generated format and the pre-upgrade persisted format match. The `history_pruner_marker_replays_as_tool_call_not_agent_message` test in `acp_server.rs` explicitly covers the legacy string case. ✓

---

### ✅ Resolved — `risk: medium` label applied

My Round-1 🟡 about the missing risk label is resolved. `risk: medium` is now set, appropriate for the `acp_server.rs` / `dispatch.rs` / `session.rs` surfaces touched. ✓

---

### 🔴 Blocking — `cancelled_prompt_result()` still carries the user-attributed marker on the ACP prompt-result path

Audacity88 has already blocked on this and I agree the finding is correct, so I'm also raising it explicitly rather than simply deferring.

The PR adds `write_notification(&Self::turn_cancelled_notification(&session_id))` immediately before `return Ok(Self::cancelled_prompt_result(session_id, &accumulated_text))`. This means:

1. A styled `turn-cancelled` `session/update` notification goes out with `[turn cancelled via client]` — the new, accurate attribution. ✓
2. The `session/prompt` result returned immediately after still calls `get_required_cli_string("turn-interrupted-by-user")` to build its content, appending `[interrupted by user]` to whatever the client receives as the terminal result for that prompt.

Any client that processes the `session/prompt` result (the canonical RPC response to `session/prompt` requests) still sees "interrupted by user" — an actor the events never prove. The `session/update` notification only reaches clients that subscribe to it; the prompt result is authoritative for all clients that issued the original prompt. So the PR's stated fix ("stops claiming a human pressed stop") is incomplete: the fix reaches one surface but leaves the primary response surface unchanged.

The PR body notes that `turn-interrupted-by-user` is preserved because `safety_net` pattern-matching depends on its exact text. That constraint is real, but it applies to the model-facing safety net detection, not necessarily to the ACP client-facing prompt result content. These are two different consumers of the same string. If the safety net match is inside the runtime's model-response handler (checking what the model returns), and `cancelled_prompt_result()` builds a client-facing JSON response, then decoupling them is feasible: use a separate internal constant for the safety net match and update `cancelled_prompt_result()` to use `turn-cancelled-client-rpc` or a new dedicated key.

If changing `cancelled_prompt_result()` content is genuinely incompatible (i.e. some existing ACP client parses the exact text), the PR should document that explicitly as a compatibility caveat in the PR body AND surface it as an accepted known limitation in the ACP protocol notes — not leave it as an unexplained inconsistency between two cancellation surfaces.

Either fix the prompt-result content, or explicitly document and acknowledge the inconsistency with a compatibility note. The current state has the PR claiming it "stops claiming a human pressed stop" while still claiming exactly that on the primary response path.

---

### 🟡 Warning — `dispatch.rs` change collapses the `AdminKill` path into the unified format without a note

The old code had three arms:
- `Some(ClientRpc)` → "turn cancelled by user in RPC_SESSION X"
- `Some(cause)` → "turn cancelled by daemon: cause"
- `None` → "turn cancelled by daemon: unattributed"

The new code has two arms:
- `Some(cause)` → "turn cancelled via cause in RPC_SESSION X" (all causes, including `AdminKill`)
- `None` → "turn cancelled (cause unattributed) in RPC_SESSION X"

The `AdminKill` path now emits "turn cancelled via admin_kill in RPC_SESSION X" rather than "turn cancelled by daemon: admin_kill". This is a behavioral change in internal log messages — the session ID is now always included, and "by daemon" is gone. It is probably the right unified format, but any monitoring rules or log-parsing tooling that keyed on "by daemon:" will need to be updated. Worth a one-liner in the PR body or a squash commit note flagging the log-format change so ops is aware.

---

### 🟡 Warning — Stale commit message may still be present

My Round-1 🟡 flagged that commit c902071851 said "Reuses the existing `turn-interrupted-by-user` i18n string" — a stale claim since the PR introduces `turn-cancelled-client-rpc`. If those commit messages survived the rebase to 52395d5a unedited, please clean the squash commit body before landing so that `git log` and `git bisect` readers see an accurate description.

---

### 🟢 What looks good — `turn_cancelled_notification` and the history-pruner replay are well-structured

`turn_cancelled_notification` builds a `session/update` notification with `kind: "think"` and `status: "completed"`, mirroring the shape of other system events. `history_notifications_for_message` using `chat.is_pruned_tool_exchange_summary()` as the dispatch condition is correct and zero-drift. The associated tests in `acp_server.rs` cover both the current and legacy marker shapes directly.

---

### 🟢 What looks good — `dispatch.rs` and `session.rs` CancelCause doc are accurate

The `CancelCause::ClientRpc` doc at `session.rs` now correctly states the actor is unverified and enumerates the programmatic paths that produce it. The `dispatch.rs` log message now names the cause channel rather than inventing a user. These are the right corrections on the paths this PR directly touches.

PR #7697, opened 2026-06-15
fix(telegram): validate API base URL default

### LINKED ISSUE #6807: [Feature]: Telegram custom webapi endpoint
### Summary

Add ability to specify custom api endpoint for telegram.

### Problem statement

I have issues with official Telegram API, and the issue number is 400 - it doesn't work for me. But I installed telegram official web api service and I can use it instead of official api url, I just need to provide custom API url to the bot.

### Proposed solution

Add configuration option custom_api_url to configure custom endpoint when needed. By default use official API.

### Non-goals / out of scope

First it would be enough to properly parse config and initialize bot (to validate it's working).

### Alternatives considered

I tried proxies, but we are talking about https here, I'd have to forge certificate to zeroclaw.

### Acceptance criteria

_No response_

### Architecture impact

channels/

### Risk and rollback

_No response_

### Breaking change?

No

### Data hygiene checks

Alix-007:
## Summary

Fixes #6807 by giving Telegram Bot API configuration a usable official default and validating configured base URLs before runtime channel construction.

## Changes

- Add `channels.telegram.<alias>.api_base_url` with the official Telegram Bot API endpoint as the default source of truth.
- Validate Telegram API base URLs as non-empty HTTP(S) URLs with a host.
- Thread the configured base URL into Telegram channel construction and file downloads.
- Add regression coverage for default construction, missing TOML defaults, custom endpoints, empty values, and malformed URLs.

## Tests

- `cargo fmt --all -- --check` - passed
- `cargo test -p zeroclaw-config telegram_api_base_url --lib` - passed
- `cargo test -p zeroclaw-channels telegram_api_url --lib` - passed
- `cargo test -p zeroclaw-channels telegram --lib` - passed
- `cargo clippy -p zeroclaw-config -p zeroclaw-channels --all-targets -- -D warnings` - passed

## Risk

Low. The change is scoped to Telegram channel configuration defaults, validation, and propagation. Existing configs that omit `api_base_url` now use the official Telegram endpoint; invalid explicit values fail validation instead of reaching runtime.


singlerider (approved):
Read against current `master`: the `TelegramConfig` schema change and `validate_http_base_url` in `zeroclaw-config/src/schema.rs`, `with_api_base` / `download_file` in `telegram.rs`, and the three construction sites in `orchestrator/mod.rs`. The change is coherent end to end and fixes a real gap. No active blocks.

### 🟢 What looks good: the base URL is now actually honored everywhere

The latent bug is `download_file`: it hardcoded `https://api.telegram.org/file/bot...` while `api_url` already routed through `self.api_base`, so a self-hosted Bot API server worked for everything except file downloads. Switching it to `{}/file/bot{}` off `self.api_base` closes that, and wiring `with_api_base(tg.api_base_url.clone())` into all three constructors - `build_channel_by_id`, `collect_configured_channels`, and `deliver_announcement` - means the override is honored on the listen path and the announcement path alike, not just one of them.

### 🟢 What looks good: default contract held without a derive footgun

Adding a non-empty-default field forced replacing `#[derive(Default)]` with a hand-written `Default` impl; doing that (rather than leaving the derive to produce an empty `api_base_url`) is the correct call, and `#[serde(default = "default_telegram_api_base_url")]` keeps existing TOML without the key parsing to the official endpoint. `TELEGRAM_OFFICIAL_API_BASE_URL` lives in the config schema crate as the single source rather than as scattered string literals. `validate_http_base_url` fails closed on empty/malformed/non-http(s)/hostless input at config-validation time, and `normalize_telegram_api_base` strips the trailing slash so `.../` and `...` resolve identically. Coverage is thorough across schema default, missing-key default, custom parse, trailing-slash normalization, both validation rejections, and the constructed download/api URLs.

Audacity88 (approved):
Reviewed the live PR state at 51d3d2d9d21d8e2a41e70fbeb86d1151f1b383b4, the linked #6807 request, the prior approval from @singlerider, the full diff across config/channel/runtime wiring, and the current green GitHub checks. I did not rerun local Cargo validation; for this review I am relying on the author’s focused config/channel commands plus the current CI results. I do not see an active blocker.

### 🟢 What looks good — The default and override contract is wired through

`TelegramConfig` now has a single official default for `api_base_url`, existing TOML that omits the key still deserializes to the official endpoint, and explicit empty/malformed/non-http(s)/hostless values fail during config validation before channel construction. The configured value is then passed into the Telegram channel in the one-shot builder, configured-channel collection, and announcement delivery path, so the setting affects the runtime surfaces that can create Telegram channels rather than only the schema.

### 🟢 What looks good — Telegram API and file URLs now share the same base

`TelegramChannel` keeps the official endpoint as its constructor default, normalizes a trailing slash when an override is provided, and uses the stored base for both normal Bot API calls and `file/bot...` downloads. That closes the self-hosted Bot API gap from #6807 without changing the default behavior for existing configs, and the added coverage exercises the default, missing-key deserialization, custom endpoints, trailing-slash normalization, invalid config rejection, and constructed API URLs.

Approving; I do not see remaining technical blockers for this scoped Telegram base URL change.

PR #7698, opened 2026-06-15
test(providers): cover Gemini current projectId fallback

Alix-007:
## Summary

Cover the Gemini OAuth `loadCodeAssist` fallback shape where the current project is returned as an object with `projectId`.

## Changes

- Adds a regression assertion for `currentCloudaicompanionProject.projectId`.
- Keeps the existing production parser unchanged.

## Tests

- `cargo test -p zeroclaw-providers load_code_assist --lib`

## Risk

Low. Test-only change for an already supported response shape.


singlerider (approved):
Read against the PR head: the `CodeAssistProject` deserializer (`#[serde(alias = "projectId")]` on the field, `resolve_project_id` handling bare-string / `{id}` / `{projectId}`) and the `currentCloudaicompanionProject` fallback path. The new assertion exercises a path the suite genuinely left uncovered. Pure test, no production change.

### 🟢 What looks good: closes the fallback-shape matrix

`resolve_project_id` accepts three shapes for both the primary (`cloudaicompanionProject`) and the fallback (`currentCloudaicompanionProject`) keys: bare string, `{id}`, and `{projectId}` via the serde alias. The primary key already had all three asserted and the fallback had bare-string and `{id}`, but the fallback `{projectId}` shape was the one untested cell - exactly the combination most likely to silently regress if someone reworks the `currentCloudaicompanionProject` branch without re-checking the alias. The added case (`current-alias`) pins it. Real deserializable path, correct expectation.

singlerider:

PR #7699, opened 2026-06-15
test(providers): cover native OpenAI timeout forwarding

Alix-007:
## Summary
Add a narrow regression test for native OpenAI provider timeout forwarding through the provider factory.

## Changes
- Build a native OpenAI provider via `OpenAIModelProviderConfig::create_provider` with `provider_timeout_secs = 1`.
- Exercise it against a local slow chat-completions server and assert the configured timeout is observed before the server responds.

## Tests
- `cargo test -p zeroclaw-providers openai_factory_forwards_timeout_to_native_provider --lib -- --nocapture`

## Risk
Low. Test-only change; no provider runtime behavior changes.


singlerider (approved):
Read against current `master`: the test, `apply_compat_options` (the single funnel where `provider_timeout_secs` becomes `with_timeout_secs`), and the OpenAI native `create_provider` path that routes through it. Pure test, no production change. No active blocks.

### 🟢 What looks good: guards the real forwarding chain, not a mock

The test stands up an actual axum server that sleeps 3s, builds the native provider through the production factory with `provider_timeout_secs: Some(1)`, and asserts both `result.is_err()` and `elapsed < 3s`. That pins the property that matters: the configured timeout has to be threaded from `ModelProviderRuntimeOptions` through `apply_compat_options` into the provider's HTTP client. Drop the `with_timeout_secs` call in that funnel and this fails by blocking the full 3s rather than passing silently. The 1s-configured / 3s-server margin is wide enough that it should not flake on a loaded runner, and `server.abort()` runs before the asserts so a failing assertion does not leak the task. Good end-to-end coverage of a previously untested wiring path.

singlerider:

PR #7703, opened 2026-06-15
fix(providers): normalize Azure OpenAI credentials

Alix-007:
## Summary
Normalize Azure OpenAI API key input at provider construction time so blank credentials behave like missing credentials.

## Changes
- Trim the optional Azure OpenAI credential before storing it.
- Treat whitespace-only credentials as `None` so calls return the existing missing-key error instead of sending a blank `api-key` header.
- Add regression tests for blank and padded credential input.

## Tests
- `cargo test -p zeroclaw-providers credential --lib -- --nocapture`

## Risk
Low. The change only normalizes invalid credential input before requests are made.


singlerider (approved):
Read against the PR head `AzureOpenAiModelProvider::new`. The change is small and aligns Azure with the credential-handling convention used elsewhere in the crate. No active blocks.

### 🟢 What looks good: blank credentials fail closed, consistently

`credential.map(str::trim).filter(|v| !v.is_empty()).map(ToString::to_string)` is the same trim-then-treat-blank-as-missing shape the rest of the provider credential resolution uses, so a whitespace-only credential now resolves to `None` instead of being stored and later sent as a malformed auth header. Trimming before storage also means a credential pasted with a trailing newline (`"...credential \n"`) is normalized once at construction rather than silently breaking auth at request time. Both new branches are pinned: blank-is-missing and trimmed-storage. The 6-arg constructor calls in the tests line up with the head signature. Tidy fix.

singlerider:

PR #7704, opened 2026-06-15
fix(memory): disable hot response cache at zero capacity

Alix-007:
## Summary

Ensure `max_entries = 0` disables both warm and hot response-cache hits.

## Changes

- Skips in-memory hot cache promotion when the configured cache capacity is zero.
- Adds a regression test that calls `get()` after `put()` to catch hot-cache hits.

## Tests

- `cargo test -p zeroclaw-memory max_entries_zero --lib`
- `cargo test -p zeroclaw-memory response_cache --lib`
- `cargo clippy -p zeroclaw-memory --all-targets -- -D warnings`

## Risk

Low. The change only affects the explicit zero-capacity cache path; nonzero capacity behavior is unchanged.


Audacity88 (approved):
Reviewed current head efe89438de7fcb662cb6dbbbbce38641f0f98a39. I checked the live PR state, current labels, visible green CI, the one-file diff in `crates/zeroclaw-memory/src/response_cache.rs`, surrounding `ResponseCache` hot/warm-cache behavior, existing zero-capacity tests, and branch freshness against current `master`. This looks good as a code fix: `max_entries = 0` now disables both the SQLite warm cache and the in-memory hot cache path.

### 🟢 What looks good — zero capacity now disables both cache tiers

The fix is in the right place. `put()` still writes through the normal path, but `promote_to_hot()` now returns immediately when `max_entries == 0`, so a zero-capacity cache cannot serve a response from memory after the SQLite row has been evicted. The new `max_entries_zero_does_not_return_hot_hit` regression covers the important behavior boundary: `put()` followed by `get()` must still miss when the configured capacity is zero. That complements the existing stats/count test, which only proved the warm SQLite tier was emptied.

### 🟡 Warning — refresh the branch and public risk wording before merge

This approval is for the patch itself, not a merge-ready statement. The branch is 17 commits behind current `master`, and the changed file is a memory/cache source file, so before merging this should get either a branch update with fresh CI, an exact merge-result smoke for `zeroclaw-memory`, or explicit stale-risk acceptance. Also, the PR body still says the risk is low while the live label is `risk: medium`; please align that public record before merge.

Alix-007:
Thanks for merging, and good catch on the risk wording — you're right that the body should track the live `risk: medium` label rather than what I wrote at open time. I'll keep the public record aligned up front next time. Also appreciated the careful read on the hot/warm tier boundary; the zero-capacity `promote_to_hot()` early-return was exactly the case I wanted covered.

Issue #7700, opened 2026-06-15
[Bug]: Matrix channel loses thread conversation history after first agent reply


singlerider:
### Affected component

channel

### Severity

S2 - degraded behavior

### Current behavior

When an agent receives a message in a Matrix thread and replies, subsequent messages in that same thread arrive in what appears to be a fresh/empty session — the agent has no memory of the prior exchange within the thread. The session key is correctly scoped to the thread root event ID (e.g. `matrix_<agent>__<room_id>__<thread_root_event_id>`), but `sessions_history` returns no messages for that session on follow-up turns.

The agent must rely on external memory notes to reconstruct context rather than reading it from conversation history.

### Expected behavior

All messages in a Matrix thread reply chain should be appended to the same session and visible via `sessions_history`. The agent should have full thread context on every subsequent reply within the same thread.

### Steps to reproduce

1. Configure a Matrix channel with `reply_in_thread = true` and `mention_only = true`
2. Send a message mentioning the agent in a Matrix room (thread root)
3. Agent replies — a thread is created
4. Reply to the agent's message within the thread (a threaded reply)
5. Observe: agent responds with no context of the prior turn; `sessions_history` for the thread-scoped session returns empty

### Impact

Affected users: All Matrix users with threaded reply mode enabled (`reply_in_thread = true`)
Frequency: Reproducible on every thread after the first reply
Consequence: Agent appears stateless within threads; users must re-provide full context on every message; memory tools are used as a workaround

### ZeroClaw version

master @ 75195f2 (2026-06-15)

### Operating system

linux

### Regression?

Unknown

### Pre-flight checks

- [X] I reproduced this on the latest master branch or latest release.
- [X] I redacted secrets, tokens, and personal data from all submitted content.

Audacity88:
## Status Update

#7705 merged on 2026-06-17 and carries the targeted Matrix conversation-history fix for this report. It changes the Matrix `conversation_history_key` path so the root message and threaded follow-ups share the same room+sender session key, while non-Matrix threaded channels keep their per-thread isolation.

Closing this as completed. If Matrix threaded replies still lose context on current `master`, comment here with the room/thread shape and we can split a fresh regression.

PR #7705, opened 2026-06-15
fix(channel/matrix): persist thread history after first reply

singlerider:
## Summary

- **Base branch:** `master`
- **Closes:** #7700
- **What changed and why:**
  - A Matrix thread follow-up resolved to a different conversation-history session than the self-anchored root that opened the thread, so the agent saw an empty session on every turn after its first reply. `sessions_history` returned nothing for the thread-scoped session even though the session key was correctly scoped, exactly as reported.
  - Root cause is in `conversation_history_key` (`crates/zeroclaw-channels/src/orchestrator/mod.rs`). It stripped the thread anchor from the **root** turn (where `thread_ts == msg.id`) but kept it for **follow-ups** (where `thread_ts` is the root and `msg.id` is the new event). The result: the root keyed to `matrix_<room>_<sender>` while the follow-up keyed to `matrix_<room>_<root>_<sender>`. The two never matched, so the follow-up read an empty history.
  - On Matrix, `thread_ts` is always a delivery anchor (a self-anchored root, or the root a follow-up replies into), never a user-chosen topic boundary the way a Slack/forum thread is. The fix collapses the thread anchor for every Matrix message, so the root and all of its thread follow-ups share one `sender+room` session.
- **Scope boundary:** Only the Matrix branch of `conversation_history_key`. Non-Matrix forum channels keep per-thread (`thread_ts`) isolation unchanged. No changes to gating, outbound routing, streaming, or session-append plumbing.
- **Blast radius:** Matrix conversation-history session key derivation only.

## Validation Evidence

- `cargo clippy -p zeroclaw-channels --features channel-matrix --all-targets -- -D warnings` — clean
- `cargo test -p zeroclaw-channels --features channel-matrix matrix` — 133 passed, 0 failed
- New regression test `matrix_thread_follow_up_shares_root_session_key` asserts the root turn and a thread follow-up compute the **same** session key. The prior test that locked in the buggy behavior (`matrix_thread_conversation_history_key_uses_thread_root`) was replaced, since it encoded the divergence that caused this bug.

## Risk (required)

- New `unsafe`? No
- Tool/agent capability or access scope? No

## Compatibility (required)

- Backward compatible? Behaviorally yes for new conversations. Note: in-flight Matrix threads created before this change stored history under the old thread-scoped key; after upgrade those follow-ups resolve to the `sender+room` session and will not see that orphaned pre-upgrade history. New threads are unaffected.

Low-risk PR: `git revert e1cd4932d0`.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -543,11 +543,9 @@ pub fn conversation_history_key(msg: &zeroclaw_api::channel::ChannelMessage) -> 
     // after a restart; otherwise hydration loads sessions under the on-disk
     // (sanitized) name while lookup keeps producing the un-sanitized form.
     let thread_scope = match msg.thread_ts.as_deref() {
-        // Matrix root events can be self-anchored when `reply_in_thread`
-        // is enabled so outbound replies open a thread. That anchor is a
-        // delivery detail, not a conversation-history boundary; otherwise
-        // every top-level Matrix message becomes a fresh session.
-        Some(tid) if is_matrix_channel_name(&msg.channel) && tid == msg.id => None,
+        // Matrix thread_ts is a delivery anchor, not a topic boundary: root
+        // and follow-ups must share one sender+room session. See #7700.
+        Some(_) if is_matrix_channel_name(&msg.channel) => None,
         other => other,
     };
     let raw = match thread_scope {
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
Reviewed the captured live PR state, #7700, the current diff, adjacent public Matrix/thread-history source context, empty captured prior reviews/comments, and the public check state, which is green. This is narrowly scoped to Matrix conversation-history key derivation: Matrix roots and follow-ups now share the sender+room history key, while non-Matrix channel threads still keep their `thread_ts` isolation.

### 🟢 What looks good — Matrix follow-ups now resolve to the same history session

The root cause in #7700 was the root and follow-up taking different session-key paths: the self-anchored root dropped the Matrix event anchor, but the follow-up kept the root event id. Collapsing Matrix `thread_ts` for both sides matches the Matrix channel code's existing treatment of those anchors as delivery/threading details, and the new `matrix_thread_follow_up_shares_root_session_key` regression checks the root/follow-up pair directly.

The scope is also contained. The non-Matrix branch still includes `thread_ts`, so forum-style channels keep per-topic history isolation.

### 🟡 Warning — Rollback text should match the current medium-risk PR

The live labels mark this PR as `risk: medium`, but the rollback section still says "Low-risk PR" and points at `e1cd4932d0` instead of the current head commit, 14dcdde805f1f4ce72336ba3f6aef12752398862. This is not a code blocker, but the body should use the medium-risk rollback format and a current rollback target before merge.

PR #7707, opened 2026-06-15
fix(tools): default zero-minute polls

Alix-007:
## Summary

Prevent polls from being created with an effective zero-minute duration.

## Changes

- Treats `duration_minutes = 0` as the default poll duration.
- Advertises `duration_minutes` as a minimum-1 integer in the tool schema.
- Covers both the sent poll text and the returned tool output.

## Tests

- `cargo test -p zeroclaw-tools duration_minutes_zero_uses_default --lib`
- `cargo test -p zeroclaw-tools execute_zero_duration_uses_default_in_sent_text_and_output --lib`
- `cargo test -p zeroclaw-tools poll --lib`

## Risk

Low. The change only affects zero or invalid duration inputs; positive durations continue to pass through unchanged.

### DIFF
--- a/crates/zeroclaw-tools/src/poll.rs
+++ b/crates/zeroclaw-tools/src/poll.rs
@@ -98,6 +98,13 @@ fn validate_options(args: &serde_json::Value) -> Result<Vec<String>, String> {
     Ok(options)
 }
 
+fn duration_minutes_or_default(args: &serde_json::Value) -> u64 {
+    args.get("duration_minutes")
+        .and_then(|v| v.as_u64())
+        .filter(|duration| *duration > 0)
+        .unwrap_or(DEFAULT_DURATION_MINUTES)
+}
+
 /// Returns true for channel names that support native polls (Telegram, Discord).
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
I checked the live PR state, current diff, comments and reviews, labels, current GitHub checks, and the `crates/zeroclaw-tools/src/poll.rs` source path. There are no existing review comments to resolve.

### 🟢 What looks good — zero-minute polls now use the default duration

The patch moves duration parsing into a small helper that filters out `0` before applying the existing 60-minute default, while positive values still pass through unchanged. The schema update also advertises `duration_minutes` as a minimum-1 integer, so the tool-call contract now matches the runtime behavior.

The regression coverage reaches the actual `execute()` path, not just the helper. It checks both the returned tool output and the text sent to the channel, including the important negative assertion that the message no longer says `0 min`.

Approved for the code change. This is not a merge-readiness verdict: the branch is behind current `master`, so refresh/update and CI should be checked before merge under the usual branch-protection flow.

PR #7708, opened 2026-06-15
docs(readme): sync installer options

Alix-007:
## Summary
Sync the README installer flag examples with the current `install.sh --help` output.

## Changes
- Add concise examples for preset, apps, TUI, gateway, prefix, dry-run, and uninstall options.
- Keep the install section focused and leave the rest of the README unchanged.

## Tests
- `bash -n install.sh`
- `bash install.sh --help` with checks for the documented options

## Risk
Docs-only. Low risk; the examples mirror existing installer options.


singlerider (approved):
First review (singlerider). Docs-only README sync, single file, CI 17/17 green, no prior reviews. I cross-checked every flag this PR adds against `install.sh` on current master and they all resolve. Approving.

### 🟢 What looks good: every added flag is real

`--preset minimal` / `--preset full`, the `--minimal` alias, `--apps`, `--without-tui`, `--with-gateway` / `--without-gateway`, `--prefix`, `--dry-run`, and `--uninstall` all exist in `install.sh`, and the inline descriptions match their actual behavior (preset is source-only kernel-only, `--minimal` aliases `--preset minimal`). Clarifying that `--minimal` is an alias rather than its own preset is the right call since the script treats it that way. This closes the drift between the README installer block and the script.

### 🟢 Note for a future drive-by: the pre-existing `--skip-quickstart` line

Not introduced by this PR (it is unchanged context), so it is out of scope here, but `install.sh` has no `--skip-quickstart` flag on current master. Worth a follow-up to either restore the flag or drop the line. Flagging only so it is on record, not as anything blocking this sync.

singlerider:

PR #7709, opened 2026-06-15
test(channels): reject non-web WhatsApp cron delivery

Alix-007:
## Summary
- Add regression coverage for WhatsApp cron delivery when the configured WhatsApp channel is Cloud API mode rather than Web mode.

## Changes
- Keep the cron delivery branch on the WhatsApp arm for configured non-Web channels.
- Align the rejection message with the accepted Web selectors instead of only naming `session_path`.
- Assert the error does not fall through to unsupported-channel or not-configured paths.

## Tests
- `cargo test -p zeroclaw-channels deliver_announcement_rejects_whatsapp_non_web_config_clearly --features whatsapp-web -- --nocapture`
- `cargo clippy -p zeroclaw-channels --all-targets --features whatsapp-web -- -D warnings`

## Risk
- Low. Test-focused change with one clearer error string for an existing rejection path.

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -9465,7 +9465,9 @@ pub async fn deliver_announcement(
                 .get(alias)
                 .ok_or_else(not_configured)?;
             if !wa.is_web_config() {
-                anyhow::bail!("WhatsApp channel send requires Web mode (session_path must be set)");
+                anyhow::bail!(
+                    "WhatsApp channel send requires Web mode (set session_path, pair_phone, or mode = personal)"
+                );
             }
             let peers = config.channel_external_peers("whatsapp", alias);
             let peer_resolver: Arc<dyn Fn() -> Vec<String> + Send + Sync> =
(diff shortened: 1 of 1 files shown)


singlerider (approved):
Read against the PR head: the `deliver_announcement` WhatsApp branch and `is_web_config` / `has_web_selector` in the config schema. The improved message is accurate and the test pins the right behavior. No active blocks.

### 🟢 What looks good: message names selectors that actually satisfy the gate

The old bail text only mentioned `session_path`, but the gate is `is_web_config()` → `has_web_selector()`, which accepts `session_path`, `pair_phone`, `pair_code`, `ws_url`, or `mode == Personal`. So the previous message under-described the fix: a user could legitimately enable Web mode via `pair_phone` or `mode = personal` and the old text never told them. The new message naming `session_path`, `pair_phone`, and `mode = personal` is a correct (representative) subset of the real selectors, so it points the user at paths that genuinely clear the gate rather than a single one. The test asserts the configured Cloud-mode config is rejected with the Web-mode message *and* that it does not fall through to "unsupported delivery channel" or a not-configured error, which is the distinction that matters: the channel is configured, just in the wrong mode for cron delivery. Correctly feature-gated under `whatsapp-web`.

Audacity88 (approved):
I checked the live PR state, current diff/source, existing reviews/comments, visible CI, and current `master` overlap around `crates/zeroclaw-channels/src/orchestrator/mod.rs`. @singlerider's approval is current, and I do not see active review threads to resolve.

### 🟢 What looks good — cron delivery rejects the configured wrong mode

This is the right narrow test for the cron-delivery bug class. The cron `deliver_announcement` path has its own WhatsApp branch, so it needed its own coverage that a configured Cloud API channel is rejected as the wrong WhatsApp mode rather than falling through as an unsupported or missing channel.

The new test builds `channels.whatsapp.default` as a configured Cloud API channel, calls `deliver_announcement("whatsapp.default", ...)`, and asserts the error stays on the Web-mode rejection path. That pins the behavior boundary the PR claims. The updated message also points users at selectors that actually satisfy the Web-mode gate, and it aligns this cron path with the other WhatsApp send path already present on current `master`.

Visible CI is green. Approved for the code change.

GitHub currently reports the PR as clean and visible CI is green. This approval is for the code change and does not replace the normal branch-protection checks before merge.

PR #7710, opened 2026-06-15
fix(channels): require IRC mention boundaries

Alix-007:
## Summary

Avoid routing ordinary IRC messages as mentions when the bot nick appears inside another word.

## Changes

- Replaces substring mention detection with token-boundary matching.
- Preserves common address forms such as `bot:`, `@bot`, and `hey bot,`.
- Adds regression coverage for `robotics`, `bothering`, and nick-suffix false positives.

## Tests

- `cargo test -p zeroclaw-channels --features channel-irc mention_only_requires_token_boundary --lib`
- `cargo test -p zeroclaw-channels --features channel-irc mention_only_case_insensitive --lib`
- `cargo test -p zeroclaw-channels --features channel-irc irc::tests --lib`
- `cargo clippy -p zeroclaw-channels --features channel-irc --all-targets -- -D warnings`

## Risk

Low to medium. The change only narrows IRC `mention_only` matching so embedded substrings no longer trigger the bot.

### DIFF
--- a/crates/zeroclaw-channels/src/irc.rs
+++ b/crates/zeroclaw-channels/src/irc.rs
@@ -43,6 +43,10 @@ pub struct IrcChannel {
 
 type WriteHalf = tokio::io::WriteHalf<tokio_rustls::client::TlsStream<tokio::net::TcpStream>>;
 
+fn is_irc_nick_char(ch: char) -> bool {
+    ch.is_ascii_alphanumeric() || ch == '_'
+}
+
 /// Style instruction prepended to every IRC message before it reaches the LLM.
 /// IRC clients render plain text only — no markdown, no HTML, no XML.
 const IRC_STYLE_PREFIX: &str = "\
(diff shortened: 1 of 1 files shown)


singlerider (approved):
Read against current `master`: `is_mentioned` and the new `is_irc_nick_char` boundary helper in `crates/zeroclaw-channels/src/irc.rs`. The fix is correct for the cases that matter and strictly better than what it replaces. No active blocks.

### 🟢 What looks good: kills the substring false-positives

The old `is_mentioned` was a raw `contains`, so nick `bot` lit up on `robotics`, `bothering`, and `bot_away` - every one of those would wrongly trip `mention_only` delivery. Walking `match_indices` and requiring the char immediately before and after each hit to be a non-nick char (`[A-Za-z0-9_]`) gives proper token boundaries, and the empty-nick early return avoids a degenerate match that would otherwise treat every message as a mention. The `is_none_or` boundary checks handle start-of-string and end-of-string cleanly. Tests pin both directions: the three substring negatives and the `bot:` / `@bot` / `hey bot,` positives.

One residual I looked at and am comfortable with: IRC nicks can technically include `-[]\`^{}|`, which `is_irc_nick_char` treats as boundaries, so `bot` could still match inside something like `bot-foo`. That is an over-match, not an under-match, and for a `mention_only` gate erring toward responding is the safe direction - it does not regress anything this PR set out to fix. Good, focused change.

Audacity88 (approved):
Reviewed the current diff, empty comment threads, prior approval, visible check state, and PR body.

This is a narrow IRC `mention_only` fix: it stops raw substring matches while preserving normal direct-address forms.

### 🟢 What looks good — IRC mention matching now avoids embedded nick substrings

The old path lowercased the message and accepted any `contains` hit, so a bot nick like `bot` could trigger on `robotics`, `bothering`, or `bot_away`. The new scan requires both sides of each nick match to be either absent or outside the accepted nick-token characters, and the added regression covers the false positives plus `bot:`, `@bot`, and `hey bot,` positives. That matches the intended behavior change without broadening the IRC receive path.

### 🟡 Warning — Medium-risk rollback/body details are still thin

The live labels mark this as `risk: medium`, but the body only says “Low to medium” and does not include the medium-risk rollback fields from the current PR template. Not a code blocker for this approval, but before merge the body should record the rollback path and observable failure symptom for this IRC mention gate.

PR #7711, opened 2026-06-15
docs(book): sync generated reference command

Alix-007:
## Summary
Sync the generated-reference table with the current `cargo mdbook refs` wrapper command.

## Changes
- Update the CLI reference row to use `cargo mdbook refs`.
- Update the config reference row to use `cargo mdbook refs`.

## Tests
- `cargo clippy -p xtask --all-targets -- -D warnings`
- `cargo test -p xtask mdbook -- --nocapture`

## Risk
Docs-only. Low risk; the wrapper still invokes the underlying reference generators internally.


singlerider (approved):
Verified against the toolchain: `cargo mdbook` is aliased in `.cargo/config.toml` to the xtask `mdbook` bin, and `xtask/src/cmd/mdbook/refs.rs` is the `refs` subcommand that internally runs both `markdown-help` (→ `reference/cli.md`) and `markdown-schema` (→ `reference/config.md`). No active blocks.

### 🟢 What looks good: table now matches the real generator

Both reference rows previously listed the raw underlying commands (`cargo run -- markdown-help` / `cargo run -- markdown-schema`), but the actual current entry point is `cargo mdbook refs`, which wraps both. Pointing the "Generated by" column at `cargo mdbook refs` matches what a contributor actually runs and keeps the doc in step with the xtask subcommand. Correct two-line sync.

Audacity88 (approved):
I reviewed head 68c44f9 against the live PR state, the one-file docs snippet diff, singlerider's approval, the current green CI run, and current `master` freshness. There are no inline threads or active blockers.

### 🟢 What looks good — The generated-reference rows now match the wrapper

The snippet already tells contributors to use the `cargo mdbook ...` wrapper instead of running lower-level tools directly. Updating both generated reference rows to `cargo mdbook refs` matches the actual wrapper entry point: `.cargo/config.toml` aliases `cargo mdbook` to the xtask mdbook binary, and the `refs` subcommand runs both `markdown-help` and `markdown-schema` internally.

No blocking findings from me. Approving this docs change.

One merge-readiness caveat: this head is behind current `master`, so before merge it should be refreshed and have CI rerun, or otherwise follow the normal branch-protection freshness path. I do not see an overlapping-file issue in this docs snippet, but this approval is not a merge-ready statement.

PR #7712, opened 2026-06-15
fix(providers): preserve structured Bedrock tool results

Alix-007:
## Summary
- Preserve structured Bedrock tool result content instead of dropping non-string JSON values.

## Changes
- Keep string content values unchanged when converting tool messages into Bedrock toolResult blocks.
- Serialize object and array content values to compact JSON text for Bedrock text-only tool result payloads.
- Add regression coverage for object, array, and existing string content behavior.

## Tests
- cargo fmt --all -- --check
- cargo test -p zeroclaw-providers parse_tool_result -- --nocapture
- cargo clippy -p zeroclaw-providers --all-targets -- -D warnings

## Risk
- Low. Change is limited to Bedrock tool result message conversion and preserves existing string and missing-content behavior.


singlerider (approved):
Read against current `master`: `parse_tool_result_message` and the new `tool_result_content_text` helper in `crates/zeroclaw-providers/src/bedrock.rs`. The dropped-content bug is real and the fix is the minimal correct one. No active blocks.

### 🟢 What looks good: structured tool output no longer silently dropped

The old path was `content.get("content").and_then(as_str).unwrap_or("")`: any tool that returned a JSON object or array (rather than a bare string) fell through `as_str()` to `None` and was flattened to an empty string before reaching Bedrock's text-only `toolResult` block. Routing through `tool_result_content_text` - string passthrough, else `value.to_string()` for compact JSON - preserves the payload instead of discarding it. The one-line comment earns its place by explaining the *why* (Bedrock toolResult is text-only, so structured output has to be encoded rather than dropped). Tests pin all three shapes: string `"ok"`, object, and array. The object assertion expecting `{"count":2,"ok":true}` from `{"ok":true,"count":2}` correctly reflects the BTreeMap-sorted serialization, so it is deterministic rather than flaky. Tight, well-scoped fix.

singlerider:

PR #7713, opened 2026-06-15
fix(zerocode): localize config status messages

Alix-007:
## Summary
Localize zerocode config manager status messages through Fluent.

## Changes
- Replace config status `format!` messages with `t_args` calls for alias, personality, skills, model-fetch, and field set/reset flows.
- Add matching English Fluent keys while preserving the existing rendered text.

## Tests
- `cargo clippy -p zerocode --all-targets -- -D warnings`
- `cargo test -p zerocode config -- --nocapture`

## Risk
Low risk. This is a user-facing text-source change only; RPC calls and editor flow are unchanged.


singlerider (approved):
Read against current `master`: the `t_args` helper signature in `apps/zerocode/src/i18n.rs`, every converted call site in `config_manager.rs`, and the new keys in `locales/en/zerocode.ftl`. I cross-checked all 24 referenced keys and their placeholder names. No active blocks.

### 🟢 What looks good: every key resolves and every placeholder lines up

This is a clean mechanical localization pass: hardcoded `format!("Saved {name}")`-style status strings become `i18n::t_args` lookups against Fluent keys, with no control-flow change. I verified the two failure modes that would actually bite - missing keys and mismatched placeholders - and both are clean:

- All 24 keys referenced from code exist: 20 are added here, and the other 4 (`alias-empty`, `fetching-templates`, `loading-personality`, `loading-skills`) already exist in the bundle.
- Every FTL `$placeholder` matches the arg key passed at the call site one-for-one: `$alias`→`("alias",..)`, the `*-failed` family →`("err",..)`, `field-reset`/`field-set`→`("prop",..)`, the personality/template keys →`("filename",..)`, the skill keys →`("name",..)`, and `fetching-models`→`("family",..)`.

Routing through the i18n bundle instead of inline literals is the right move for these surfaces. Adding the keys to the `en` source bundle only (translations landing separately) is the expected pattern. No logic risk.

singlerider:

Audacity88 (approved):
Reviewed current head `ae178e0fcf7f3444bca6d76543ef1cc10fc7de57` against the live PR state/checks after the `master` refresh, the current diff, the existing approval, the converted `config_manager.rs` call sites, and the English Fluent bundle. I did not run local Cargo; Format and title validation have passed, and the remaining refreshed CI jobs are still pending, so this approval is for the patch rather than merge readiness.

### 🟢 What looks good — status messages now use Fluent without changing behavior

This is a clean localization pass. The changed status strings keep the same rendered English text while moving the alias, personality, skill, model-fetch, and field set/reset flows through `i18n::t_args`.

I checked the failure mode that matters most here: every new `zc-config-status-*` key used from `config_manager.rs` exists in `apps/zerocode/locales/en/zerocode.ftl`, and the placeholder names line up with the arguments passed at each call site (`alias`, `err`, `prop`, `filename`, `name`, and `family`). The converted paths still set `status_msg` in the same success/error branches, so this does not change the editor flow or RPC behavior.

No code blocker found. Before merge, wait for the remaining refreshed CI jobs from the branch update to finish green.

PR #7714, opened 2026-06-15
fix(config): trim legacy typed scalar strings

Alix-007:
## Summary

Trim whitespace around legacy string values before coercing them into typed scalar config values.

## Changes

- Trim string inputs for bool, integer, and float typed config fields before validation and serialization.
- Preserve whitespace for string, enum, and unknown-kind string coercion.
- Add regression coverage for typed scalar trimming and string preservation.

## Tests

- `cargo test -p zeroclaw-config legacy_typed_scalar_strings_trim_whitespace --lib`
- `cargo test -p zeroclaw-config string_fields_preserve_whitespace --lib`
- `cargo test -p zeroclaw-config typed_value --lib`
- `cargo clippy -p zeroclaw-config --all-targets -- -D warnings`

## Risk

Low. The change is limited to legacy string coercion for typed scalar config values and preserves existing string passthrough behavior.


singlerider (approved):
Read against current `master`: `coerce_for_set_prop` and `json_type_name` in `crates/zeroclaw-config/src/typed_value.rs`. The change tightens, not loosens, validation while fixing a real whitespace papercut. No active blocks.

### 🟢 What looks good: trims typed scalars, preserves strings, sharpens the error

The previous guards (`s.parse::<i64>().is_ok()`, the `true`/`false` check) failed on a value like `" 42 "` or `" True "` and fell through to the generic `(Some(PropKind::Integer), other)` arm, so a hand-edited TOML or copy-pasted value with stray whitespace produced a confusing type-mismatch even though the intent was unambiguous. Trimming inside the Bool/Integer/Float arms before parsing accepts those, and the new `else` branches reject genuinely non-conforming strings with a *specific* message (`integer field requires a whole number; got ...`) instead of the old silent fall-through. That is a net tightening: malformed input still errors, just more clearly.

Critically, the trim is scoped to the three numeric/bool typed arms only - `String`, `Enum`, and the untyped `None` path are untouched, so whitespace-significant values survive intact. The `string_fields_preserve_whitespace` test pins that (`"  value  "` round-trips across `String`, `Enum`, and `None`), and `legacy_typed_scalar_strings_trim_whitespace` pins the trim side. Right scope, right safety direction.

Audacity88 (approved):
I reviewed the live PR state, current diff for `crates/zeroclaw-config/src/typed_value.rs`, PR body, visible CI, and singlerider's prior approval. There are no open comments or review threads to reconcile. I did not run local validation for this review-only pass; the visible CI is green, and the author's focused `zeroclaw-config` tests plus clippy/fmt evidence match the changed helper surface.

### 🟢 What looks good — trim is scoped to typed scalar coercion

The change trims only the legacy string inputs for `Bool`, `Integer`, and `Float` before parsing and serialization. It does not touch the `String`, `Enum`, or untyped match arms, so those values still preserve whitespace exactly. The new explicit failures for invalid typed-scalar strings also keep malformed values rejected with clearer errors.

### 🟢 What looks good — regression coverage pins both sides

`legacy_typed_scalar_strings_trim_whitespace` covers the intended bool/integer/float normalization, and `string_fields_preserve_whitespace` protects the whitespace-significant cases. That is the right pair of tests for this change because the risk is not just accepting `" 42 "`, it is making sure `"  value  "` still round-trips unchanged where strings are the actual data.

This is a focused compatibility fix with the right guard tests.

singlerider:

PR #7715, opened 2026-06-15
docs(setup): mention Linux uninstall script

Alix-007:
## Summary

Document the Linux bootstrap uninstall command in the setup guide.

## Changes

- Add `./install.sh --uninstall` to the Linux uninstall section.
- Keep the existing manual service, binary, and workspace cleanup steps for non-bootstrap or manual cleanup paths.

## Tests

- `./scripts/ci/docs_quality_gate.sh`
- `git diff --check origin/master...HEAD`
- `bash -n install.sh`
- Manual changed-line link scan: no added links detected

`./scripts/ci/docs_links_gate.sh` could not run on the validation host because its default `python3` is 3.6.8 and the repository script requires Python 3.7+ syntax. This change adds no links, so the link-integrity path is empty.

## Risk

Low. Documentation-only change that points users to an existing installer flag.

### DIFF
--- a/docs/book/src/setup/linux.md
+++ b/docs/book/src/setup/linux.md
@@ -176,6 +176,18 @@ zeroclaw service restart
 
 ## Uninstall
 
+If you installed with the bootstrap script, use the same script to uninstall:
+
+<div class="os-tabs-src">
+
+#### sh
+
+```sh
+./install.sh --uninstall
+```
+
+</div>
+
 Stop and remove the service:
 
 <div class="os-tabs-src">


singlerider (approved):
Verified against `install.sh`: `--uninstall` is a real flag (help text, arg parse at the `--uninstall) UNINSTALL=true` case, and the `do_uninstall` routine that runs `service uninstall` and removes the binary). No active blocks.

### 🟢 What looks good: documents an actual supported flag

The added `./install.sh --uninstall` is accurate - the bootstrap script genuinely implements that path, including stopping the service and removing the binary, so a reader following this gets a working teardown rather than a phantom command. The `os-tabs-src` / `#### sh` block matches the markup already used by the "Stop and remove the service" section right below it, so it renders consistently. Small, correct, and it closes a real gap (install was documented, uninstall via the same script was not).

Audacity88 (approved):
Reviewed the live PR state at ad0b5ba9b4e5829b0851a6022aa1411d9b112619, the one-file docs diff, the prior approval from @singlerider, the current `install.sh` implementation, and the green GitHub checks. I did not rerun local validation; for this docs-only review I am relying on the author’s docs checks plus current CI. I do not see an active blocker.

### 🟢 What looks good — The documented uninstall path exists

`./install.sh --uninstall` is a supported bootstrap path: the installer exposes the flag in help text, parses it into the uninstall mode, and routes to `do_uninstall`, which stops/uninstalls the service when the binary exists and removes the installed binary. Adding it to the Linux uninstall section gives bootstrap users the matching teardown command instead of making them infer the lower-level service and binary cleanup steps.

### 🟢 What looks good — Manual cleanup paths remain available

The new block sits above the existing service, binary, Homebrew, and optional config/workspace removal steps, so it improves the bootstrap path without deleting the manual cleanup guidance for non-bootstrap installs or extra cleanup cases. The `os-tabs-src` / `#### sh` / fenced-shell markup matches the surrounding setup guide style.

Approving; this is a small, accurate docs improvement.

PR #7716, opened 2026-06-15
fix(tools): reject excessive round precision

Alix-007:
## Summary
Reject excessive `round` precision in the calculator tool instead of letting huge `decimals` values overflow the internal multiplier.

## Changes
- Adds a local maximum for `round` decimal places before computing the multiplier.
- Returns a parameter error when `decimals` is above that maximum.
- Keeps the existing normal `round` behavior covered by its current test.

## Tests
- `cargo test -p zeroclaw-tools calculator::tests::test_round -- --nocapture`

## Risk
Low. The change is limited to calculator input validation and adds a regression test for the rejected boundary.

### DIFF
--- a/crates/zeroclaw-tools/src/calculator.rs
+++ b/crates/zeroclaw-tools/src/calculator.rs
@@ -2,6 +2,8 @@ use async_trait::async_trait;
 use serde_json::json;
 use zeroclaw_api::tool::{Tool, ToolResult};
 
+const MAX_ROUND_DECIMALS: i64 = 15;
+
 pub struct CalculatorTool;
 
 impl CalculatorTool {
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
Review context: I checked PR #7716 at head c53e58d against the current diff, current calculator implementation, visible GitHub checks, and branch freshness state. I did not run local Cargo validation for this review-only pass; I relied on source inspection, the author's listed validation, and green GitHub checks.

@Alix-007 thanks for tightening this edge case. The fix is small and in the right place: `round` now rejects excessive `decimals` before computing the `10^decimals` multiplier, while the existing negative-decimal rejection and normal rounding behavior stay intact.

### 🟢 What looks good — excessive precision now fails as a tool error

The new upper bound turns the huge-precision path into the same kind of parameter error the calculator already uses for invalid inputs, instead of letting the multiplier path overflow or produce unusable numeric output. The regression test covers the abusive value, and the existing `test_round` still covers the normal behavior.

One merge-readiness note: this approval is only for the reviewed head c53e58d. The branch is currently 17 commits behind `master`. The freshness snapshot did not find overlapping changed files, but the PR changes `crates/zeroclaw-tools/src/calculator.rs`, so the safer pre-merge path is to update the branch and wait for CI, run a merge-result smoke, or have a maintainer explicitly accept merging the stale head.

PR #7717, opened 2026-06-15
fix(tools): encode weather location path input

Alix-007:
## Summary
- Percent-encode wttr.in location path input before appending the fixed JSON query string.
- Preserve existing behavior for normal city names, airport codes, zip codes, and GPS coordinates.

## Changes
- Reuse the existing `urlencoding` dependency in `WeatherTool::build_url`.
- Keep spaces as `+` and GPS coordinate commas as literal commas for wttr.in compatibility.
- Add regression coverage for `A&B`, `Paris?format=3`, and a non-ASCII location.

## Tests
- `cargo test -p zeroclaw-tools build_url --lib`

## Risk
- Low. This is limited to URL construction for weather location input and does not add or duplicate state.

### DIFF
--- a/crates/zeroclaw-tools/src/weather_tool.rs
+++ b/crates/zeroclaw-tools/src/weather_tool.rs
@@ -140,8 +140,9 @@ impl WeatherTool {
 
     /// Build the wttr.in request URL for the given location.
     fn build_url(location: &str) -> String {
-        // Percent-encode spaces; wttr.in also accepts `+` but %20 is safer.
-        let encoded = location.trim().replace(' ', "+");
+        let encoded = urlencoding::encode(location.trim())
+            .replace("%20", "+")
+            .replace("%2C", ",");
         format!("{WTTR_BASE_URL}/{encoded}?format=j1")
     }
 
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
I checked the current #7717 diff at 389f483c6eccd3a10db6f45138bb3ec064206d9e, the PR metadata, reviews/comments, and the GitHub checks. The change is limited to `WeatherTool::build_url` and focused regression coverage.

### 🟢 What looks good — Location input is encoded before the fixed query

The patch encodes the user-provided location before appending wttr.in's fixed `?format=j1` query, so inputs such as `A&B` and `Paris?format=3` stay inside the path instead of changing the request query. It also preserves the existing city, GPS coordinate, airport-code, and zip-code expectations, and adds non-ASCII coverage for the path encoding case.

All visible checks are green on the current head.

Alix-007:
Thanks! You read the intent exactly — keeping inputs like `A&B` and `Paris?format=3` inside the path instead of leaking into the query was the whole point. Appreciate the careful review.

PR #7719, opened 2026-06-15
test(gateway): cover public event history filtering

Alix-007:
## Summary
Add narrow regression coverage for `/api/events/history` so session-scoped events stay out of the public history payload.

## Changes
- Reuses the existing history filtering path through a private payload helper.
- Adds a buffer regression test containing both a session-scoped event and public lifecycle/observability events.
- Asserts the history payload only includes the public events.

## Tests
- `cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings`
- `cargo test -p zeroclaw-gateway sse::tests::history_payload_returns_only_public_events -- --nocapture`

## Risk
High area, low change risk. This is limited to gateway SSE history coverage and does not change route registration, authentication, or response shape.


Audacity88 (approved):
Thanks @Alix-007. I reviewed the refreshed head 78be53ded4c56ebd6e42dd6f81bf5564a4a07a36 after the branch update, including the one-file gateway SSE diff in `crates/zeroclaw-gateway/src/sse.rs`, the PR body, current comments/reviews, and green CI. I did not find linked issues or prior review threads that change this review.

### 🟢 What looks good — History payload uses the same public filter

The route still performs auth in `handle_events_history`, and the new helper only returns the same filtered `{ "events": ... }` payload. That keeps the route behavior equivalent while making the payload easy to test. The regression covers the boundary I care about for this endpoint: a session-scoped message stays out of `/api/events/history`, while public observability/lifecycle events remain present.

Current CI is green on head 78be53ded4c56ebd6e42dd6f81bf5564a4a07a36.

Alix-007:
Thanks for re-reviewing the refreshed head after the branch update — good to know the session-scoped vs public-event boundary still held on the new diff. Appreciate the thorough pass.

Issue #6351, opened 2026-05-03
[Bug]: WhatsApp Web — self-chat-mode triggers on all fromMe messages, causes agent to reply to the operator's contacts


theonlyhennygod:
## Affected component

`channel` — specifically WhatsApp Web (`channel:whatsapp`, `mode = personal` or `business`).

## Severity

**S2 — degraded behavior.** Causes the agent to send unwanted replies *to the operator's own contacts* under the operator's identity. Not a crash, but actively damaging in a paired-personal-account context: the bot piggybacks on the user's own outbound messages with a follow-up reply, which the recipient sees as a second message from their friend.

## Current behavior

When `channels.whatsapp.self-chat-mode = true`, the channel triggers an agent reply on **any** inbound event with `fromMe: true` — not just messages in the dedicated self-chat thread (where `chat_jid == bot's own jid`).

Concretely: WhatsApp Web mirrors all the user's outbound messages to all linked devices, marked `fromMe: true`. zeroclaw's linked-device session sees these mirrored events. With `self-chat-mode = true`, the channel treats every such event as eligible for an agent reply, regardless of which chat thread it belongs to. If the sender (the bot's own number) is in `allowed-numbers`, the agent generates a reply and sends it **into the same chat thread**, which is the conversation with whoever the user was originally messaging.

Result: when the user types "hey what's up?" to a friend on their phone, the agent receives the mirrored event, decides to reply, and sends a follow-up message ("yo, just chillin" or whatever) **to that friend** under the user's identity. The friend sees a double-text from the user.

## Expected behavior

`self-chat-mode = true` should enable replies **only in the actual self-chat thread**, where `chat_jid == bot's own jid` (WhatsApp's "Message Yourself" / Note-to-Self thread). It should NOT trigger on `fromMe: true` events in arbitrary chat threads.

The current scope is implied by the field name (`self-chat-mode`) and by the docstring in `src/channels/whatsapp_web.rs`. The implementation gates on `fromMe`, not on the more restrictive `fromMe AND chat_jid == my_jid`.

## Steps to reproduce

```bash
# 1. Pair zeroclaw to your own WhatsApp in Web mode (mode = personal)
zeroclaw config set channels.whatsapp.mode personal
zeroclaw config set channels.whatsapp.session-path /var/lib/zeroclaw/whatsapp.session
zeroclaw config set channels.whatsapp.pair-phone 1XXXXXXXXXX
[...]

PR #7720, opened 2026-06-15
feat(channels/whatsapp): add allowed_groups per-JID group allowlist for Web mode

### LINKED ISSUE #6371: [Feature]: WhatsApp Web: per-JID group allowlist (allowed_groups)
### Summary

Add an `allowed_groups: Vec<String>` field to `WhatsAppConfig` so WhatsApp Web operators can restrict the bot to a named set of group chats.

### Problem statement

WhatsApp Web operators have no way to scope the bot to specific group chats. The existing gates filter by chat type (`dm_policy` and `group_policy`) or by sender (`allowed_numbers`), but not by group identity. An operator running WhatsApp Web across personal and business contexts can admit the bot to a class of chats (DMs, all groups), but cannot enumerate which specific groups it operates in.

Issue #6351 (open, 2026-05-03) is the most recent report of this gap: in self-chat-mode the bot replies under the operator's identity across every chat the operator originated, with no JID-level control to limit which chats are affected.

The Matrix channel solved the same problem with `allowed_rooms` (PR #4260, merged 2026-03-22), and the same shape applies to WhatsApp Web.

### Proposed solution

Add `allowed_groups: Vec<String>` to `WhatsAppConfig`, modeled on Matrix's `allowed_rooms` (PR #4260, merged 2026-03-22). The same shape (a sibling `Vec<String>` field with `#[serde(default)]`, not a new variant on an existing enum) fits next to the existing `allowed_numbers`. PR #4131 (merged 2026-03-21) followed the same pattern in `WhatsAppConfig`: chat-type filtering arrived as new sibling fields (`dm_policy`, `group_policy`, `mode`, `self_chat_mode`), not as variants of an existing enum.

Each entry is either a full group JID (`123456789012345@g.us`) or a JID prefix (`123456789012345`).

When `allowed_groups` is non-empty, the WhatsApp Web event handler keeps only those group messages whose chat JID matches an entry; everything else is dropped silently. The check runs between the existing `is_number_allowed_for_list` sender filter and the personal-mode policy block.

`#[serde(default)]` makes the empty list the default, so existing configurations need no migration and behave exactly as today.

Sketch:

```rust
// crates/zeroclaw-config/src/schema.rs
pub struct WhatsAppConfig {
    // ... existing fields ...

    /// Allowed group chats by JID. An empty list permits all groups
    /// (current default). Each entry matches as a full JID
    /// (`123@g.us`) or a JID prefix (`123`). Direct messages bypass
    /// this filter.
    #[serde(default)]
    pub allowed_groups: Vec<String>,
}

// crates/zeroclaw-channels/src/whatsapp_web.rs (event handler)
if !allowed_groups.is_empty() && is_group && !is_chat_allowed(&chat, &allowed_groups) {
    tracing::debug!(
        target: "whatsapp_web",
        chat = %chat,
        "dropping group message: chat not in allowed_groups",
    );
    return;
}
```

### Non-goals / out of scope

- **Special tokens.** No `"*"` or `"dm"` shortcuts. The empty list means *allow all*; for a DM-only bot, set `dm_policy = Allowlist` and `group_policy = Ignore` on the existing fields.
- **Auto-leave.** This proposal does not reject or leave invites to non-allowed groups. Matrix's `allowed_rooms` does, but WhatsApp's invite model differs, and message-time filtering is enough.
- **Pattern matching.** No regex, no globs. Plain JID strings only.
- **Sender allowlisting inside groups.** Out of scope. `allowed_numbers` already covers sender gating.

### Alternatives considered

- **Extend the `WhatsAppChatPolicy` enum** with a new variant such as `Allowlist(Vec<String>)`. Rejected. First, the enum is flat (`#[serde(rename_all = "snake_case")]`); adding a tuple variant would break serde compatibility with every config that already uses `group_policy = "allowlist"` as a bare string. Second, `group_policy` applies uniformly to all groups; the match expression in `whatsapp_web.rs` has no per-chat-JID test in any arm. Distinguishing between specific groups requires a sibling field.

- **Reuse `allowed_numbers`** with group JIDs treated as quasi-phone-numbers. Rejected. `allowed_numbers` filters senders, not chats: `is_number_allowed_for_list` calls `normalize_phone_token`, which `crates/zeroclaw-config/src/schema.rs:245` documents as *"Accepts raw numbers, `+` numbers, and JIDs (uses the user part before `@`)"*. A group JID placed in `allowed_numbers` would normalize to a digit prefix and match a sender's phone digits only by coincidence. The deeper issue is semantic. `allowed_numbers` answers *"which senders may interact with the bot"*, not *"which chats the bot operates in"*. Collapsing the two would overload one field with two unrelated meanings.

### Acceptance criteria

- `WhatsAppConfig` gains an `allowed_groups: Vec<String>` field with `#[serde(default)]`.
- An empty list preserves current behavior: every group passes.
- A non-empty list drops every group message whose chat JID matches no entry.
- Direct messages (chats ending `@s.whatsapp.net`) bypass the filter regardless of list contents.
- An entry matches both the full JID (`123@g.us`) and the digit-only prefix (`123`).
- Unit tests cover: empty list, full-JID match, JID-prefix match, no-match drop, DM bypass.
- The WhatsApp Web channel page in `docs/book/src/channels/whatsapp-web.md` documents the field and gives one config example.

### Architecture impact

- **Config:** Add the field on `WhatsAppConfig` in `crates/zeroclaw-config/src/schema.rs`, alongside the existing `allowed_numbers`.
- **Channel:** Add a JID match helper in `crates/zeroclaw-channels/src/whatsapp_web.rs` and call it from the event handler before the personal-mode policy block.
- **Docs:** Document the field, its semantics, and one example in `docs/book/src/channels/whatsapp-web.md`.
- No `zeroclaw-api` trait surface changes. No schema migration. No new dependencies.

### Risk and rollback

Low risk. The `#[serde(default)]` empty list preserves current behavior for existing configurations; operators opt in by populating the list. Rollback is `git revert <sha>` of the merge commit, with no state migration. The unit tests cover the behavioral surface, so a regression would fail CI before merge.

### Breaking change?

No

### Data hygiene checks

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:** WhatsApp Web operators had no way to scope the bot to specific group chats — the existing gates filter by chat *type* (`dm_policy`/`group_policy`) or by *sender* (peer allowlist), but not by group *identity* (the gap reported in #6351). This adds an `allowed_groups: Vec<String>` field to `WhatsAppConfig` and a message-time filter in the Web event handler, modeled directly on the Matrix channel's `allowed_rooms` (PR #4260) and following the same sibling-field convention that `dm_policy`/`group_policy` arrived under (PR #4131).
- **What it does:** Empty list (the default) permits every group — no behavior change. A non-empty list drops every group message whose chat JID matches no entry. Each entry matches the full group JID (`123456789012345@g.us`) **or** the JID user part (the segment before `@`, `123456789012345`), compared **exactly** (not a string prefix). Direct messages always bypass the filter. The check runs before the personal-mode policy block, so it applies in both `business` and `personal` mode.
- **Scope boundary:** No special tokens (`*`/`dm`), no auto-leave of non-allowed groups, no regex/glob matching, no sender allowlisting inside groups, no `zeroclaw-api` trait changes, no schema migration, no new dependencies — exactly the non-goals in #6371.
- **Blast radius:** `zeroclaw-config` (`WhatsAppConfig` schema) and `zeroclaw-channels` (`whatsapp_web.rs` event handler). Behavior changes only for the WhatsApp Web channel, and only when an operator sets a non-empty `allowed_groups`. All other channels and the default WhatsApp config are unchanged.
- **Linked issue(s):** Closes #6371. Related #6351.
- **Labels:** `enhancement`, `channel`, `channel:whatsapp`, `config`, `docs`, `risk: high`, `size: S`.

Run on the PR head against `master` (build host: native arm64). The affected crates were validated in full; the whatsapp-web code path is gated behind the `whatsapp-web` feature, so the channel battery is run with that feature enabled. CI runs the full workspace matrix.

```
$ cargo fmt -p zeroclaw-config -p zeroclaw-channels -- --check
  (clean — no diff)

$ cargo clippy -p zeroclaw-channels --features whatsapp-web --all-targets -- -D warnings
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 16s     # 0 warnings

$ cargo test -p zeroclaw-config --lib whatsapp
  test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 847 filtered out

$ cargo test -p zeroclaw-channels --features whatsapp-web allowed_groups
  test whatsapp_web::tests::allowed_groups_empty_permits_all ... ok
  test whatsapp_web::tests::allowed_groups_full_jid_match ... ok
  test whatsapp_web::tests::allowed_groups_jid_prefix_match ... ok
  test whatsapp_web::tests::allowed_groups_no_match_drops ... ok
  test whatsapp_web::tests::allowed_groups_dm_bypasses_filter ... ok
  test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 894 filtered out
```

- **Beyond CI — what I manually verified:** an adversarial review pass flagged that an earlier `starts_with` implementation made prefix entries over-broad (`"123"` would have admitted `123999@g.us`). Fixed to exact full-JID-or-user-part matching, with a regression test (`allowed_groups_no_match_drops`) asserting `"123"` admits `123@g.us` but rejects `123999@g.us`. Tests cover all acceptance cases: empty list, full-JID match, user-part match, no-match drop, DM bypass.
- **If any command was intentionally skipped, why:** the whole-workspace `cargo test` / `clippy --all-targets` were left to CI; local validation was scoped to the two crates this PR touches (the change is additive and feature-gated).

- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — all JIDs in tests/docs are synthetic placeholders (`123…@g.us`).
- The feature is a restrictive allowlist: it can only *narrow* which groups the bot answers in, never widen the existing sender/chat-type gates.

## Compatibility (required)

- Backward compatible? **Yes** — `allowed_groups` is `#[serde(default)]`, so the empty-list default preserves current behavior and existing TOML configs need no migration.
- Config / env / CLI surface changed? **Yes (additive only)** — one new optional `[channels.whatsapp.<alias>]` key, `allowed_groups`. No existing key changes meaning. Upgrade steps for existing users: none; to use it, add `allowed_groups = ["<jid>", …]`.
- **On API stability (`WhatsAppConfig` field addition):** intentionally *not* marked `#[non_exhaustive]`. `zeroclaw-config` is a workspace-internal crate (`publish = false`) — there are no external literal-constructors — and no channel config struct in the crate uses `#[non_exhaustive]`. Adding a `#[serde(default)]` sibling field is the established, maintainer-blessed convention for these configs (the cited Matrix `allowed_rooms` in #4260 and the WhatsApp `dm_policy`/`group_policy` cluster in #4131 both did exactly this). All in-tree struct-literal constructors (5 test sites) are updated in this PR.

- **Feature flags or config toggles:** None compiled in; operationally, clearing `allowed_groups` (or leaving it empty) disables the filter without a code change.
- **Observable failure symptoms:** if reverted, operators relying on `allowed_groups` would see the bot resume answering in *all* groups again. While active, dropped group messages emit a `DEBUG` log: grep `whatsapp_web` logs for `dropping group message: chat not in allowed_groups`.


singlerider (approved):
Read against current `master`: the `is_group_chat_allowed` helper and its call site in `crates/zeroclaw-channels/src/whatsapp_web.rs`, the `WhatsAppConfig::allowed_groups` schema field, and the docs. This is a careful, defensive allowlist and the safety properties all hold. No active blocks.

### 🟢 What looks good: exact match, deny-by-default, DMs bypass, no behavior change on empty

The four properties that matter for a group gate are all correct and pinned:

- **Backward compatible**: empty list permits every group, so existing configs are untouched; every schema test fixture seeds `allowed_groups: vec![]` and `allowed_groups_empty_permits_all` asserts it.
- **Exact, not prefix**: matching is `entry == chat_jid || entry == chat_user` where `chat_user` is the segment before `@`. This is the subtle correctness point for an identity allowlist - `"123"` must admit `123@g.us` but never `123999@g.us` - and both the doc and `allowed_groups_no_match_drops` pin exactly that, including the `123999@g.us` rejection. A naive `starts_with` here would have been a real authorization hole; this avoids it.
- **DMs bypass**: the call site gates on `is_group &&` before calling the helper, so direct messages are never dropped by group rules; `allowed_groups_dm_bypasses_filter` documents that contract.
- **Fail direction**: a non-empty list denies unmatched groups (deny-by-default), and blank/whitespace-only entries are trimmed and never match, so a stray `""` can't silently widen the allowlist.

The gate runs before the chat-type policy block and applies in both business and personal mode, matching the doc. The drop is logged through `record!` wrapped in `Event::new(module_path!(), ...)`, so it stays attributed. The field doc explains *why* this exists distinct from `group_policy` (chat type) and the sender allowlist (sender) - it gates group identity, which neither of those do. Modeled sensibly on the Matrix `allowed_rooms` precedent.

perlowja:
Heads-up: pushed a trivial CI-hygiene commit (`8f7e93c17`) replacing em-dashes with hyphens in the added comments/docs to clear the prose em-dash gate on this branch too. Char-only, no behavior change — the approved logic is untouched.

Audacity88 (dismissed):
I checked the live PR metadata/body, public diff/source excerpts, @singlerider's prior approval, visible CI, `AGENTS.md`, and the WhatsApp/config docs. I did not run local Cargo or live WhatsApp validation.

### 🟢 What looks good — exact group identity matching is well shaped

The matching rule itself is the right one for this feature. Empty `allowed_groups` preserves existing behavior, and a populated list matches only the full group JID or the exact JID user part. The `123` versus `123999@g.us` regression is especially important because a prefix-style match would turn this into a real allowlist bypass.

The call-site ordering also points in the right direction: the check only runs for group messages, so direct messages bypass the group allowlist, and the filter runs before the personal-mode chat-type policy so the group identity gate applies in both business and personal Web mode.

### 🔴 Blocking — resolve `allowed_groups` from canonical config instead of caching it on the channel

The PR adds `WhatsAppConfig.allowed_groups` as the config source of truth, then clones it into `WhatsAppWebChannel.allowed_groups: Arc<Vec<String>>` and captures that clone in the event handler. That gives the channel handle its own copy of an authorization allowlist. This is the class of duplicate the repository's single-source-of-truth rule calls out directly: channel allowlist `Vec` fields cached inside handles while the truth lives in config TOML, with config reloads unable to refresh the channel copy. I’m raising this even though @singlerider already approved the matching behavior, because this is a separate source-of-truth/wiring concern rather than a match-semantics concern.

Please keep `allowed_groups` resolved from canonical config at use time instead of storing the raw list on the channel handle. The existing `peer_resolver` pattern in this same constructor shows the allowed shape: pass a resolver closure, or another approved on-demand config resolver, and call it before `is_group_chat_allowed(...)` in the message handler. The pure matching helper and its focused tests can stay; the wiring should avoid a second stored copy of the allowlist.

perlowja:
Thanks @Audacity88 — the single-source-of-truth blocker is addressed (commit `3834bb7a`).

## `allowed_groups` now resolved from canonical config at use time

- Removed the cached `WhatsAppWebChannel.allowed_groups: Arc<Vec<String>>` field; the channel handle no longer stores its own copy of the allowlist.
- Added an `allowed_groups_resolver` closure on the handle, following the existing `peer_resolver` pattern in the same constructor. Live construction resolves `[channels.whatsapp.<alias>].allowed_groups` from the canonical `Arc<RwLock<Config>>`, and the message handler calls the resolver **immediately before** `is_group_chat_allowed(...)` — so a config reload is honored instead of a stale cached authorization list.
- The pure `is_group_chat_allowed` matcher (exact full-JID / exact user-part, empty = allow-all) and its focused unit tests are unchanged.
- Constructor call sites / tests updated for the new resolver argument.

## Validation (head `3834bb7a`)

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | ✅ clean |
| `cargo clippy -p zeroclaw-channels -p zeroclaw-config --all-targets -- -D warnings` | ✅ 0 warnings |
| `cargo check -p zeroclaw-channels -p zeroclaw-config` | ✅ |
| `cargo test -p zeroclaw-channels` (whatsapp matcher tests) | ✅ |

Ready for another look.

Audacity88:
@perlowja thanks for the update here. I rechecked the current head after 3834bb7a and the source-of-truth concern from my prior review is addressed: `WhatsAppWebChannel` now keeps an `allowed_groups_resolver`, the normal channel construction paths resolve from canonical config, and the message handler calls the resolver immediately before the `is_group_chat_allowed(...)` check.

I am not posting a formal approval right now because GitHub reports this branch as `DIRTY` / `CONFLICTING` against `master`. Once the branch is refreshed and CI reruns on the conflict-resolved head, this should be ready for a final re-review.

perlowja:
Refreshed against latest `master` — the branch is no longer `DIRTY`/`CONFLICTING`, so this is ready for your final re-review.

The rebase had conflicts in `whatsapp_web.rs` and `orchestrator/mod.rs` from master's newer local-media trust-boundary work. I resolved them by **keeping both sides**: master's `workspace_dir` field/logic and #7720's `allowed_groups_resolver` coexist, every constructor initializes both fields, and the allowed-groups resolution still runs immediately before `is_group_chat_allowed(...)` (DMs bypass; empty list = allow all).

Validation on the rebased head: full gate at pinned 1.93.0 — `clippy --features ci-all`, `check --locked` (ci-all + no-default), both architecture guards, and `nextest` (9381 pass). Adversarial review (cross-family) — APPROVE. The green `cargo check --features ci-all` also confirms the merge is field-consistent (a missed init would be a compile error).

perlowja:
@singlerider this one's approved and green (18/18, mergeable-clean) after the master refresh — good to squash-merge whenever it's convenient. Happy to rebase again if it drifts.

Audacity88 (approved):
I rechecked refreshed head 280fbeff after updating the branch from `master`: live PR state, prior reviews/comments, linked #6371 and #6351, the WhatsApp/config/docs diff, the exact refreshed-head source for the construction and event-handler paths, and green GitHub CI on that refreshed head. I did not run local Cargo or a live WhatsApp Web smoke; this review is based on source, the regression coverage in the PR, and CI. @perlowja thanks for the refresh, and thanks for reworking the allowlist source-of-truth path.

### ✅ Resolved — `allowed_groups` now comes from canonical config

The prior blocker from my review is addressed. The normal inbound construction paths now pass an `allowed_groups_resolver` that reads `[channels.whatsapp.<alias>].allowed_groups` from shared config at message time, and the event handler calls it immediately before `is_group_chat_allowed(...)`. The allowlist is no longer cached on the channel handle, so I do not see the old cached-allowlist concern remaining.

### 🟢 What looks good — exact group identity gate stays narrow

I agree with @singlerider's earlier read on the matching behavior: an empty list preserves existing behavior, a non-empty list denies unmatched groups, full JID or exact user-part matching avoids prefix bypass, blank entries do not widen access, and direct messages bypass through the `is_group` guard. The docs and tests pin the important `123` versus `123999@g.us` case.

I also think it is the right tradeoff to keep this field separate from `group_policy` and the sender allowlist. This feature gates group identity; the existing fields answer different questions.

One caveat: I did not perform live WhatsApp Web validation here. I am still comfortable source-reviewing this change because the behavior is a narrow predicate over the inbound chat JID, and the PR now has focused regression coverage for the edge cases that would create an allowlist bypass.

PR #7721, opened 2026-06-15
fix(runtime): return no heartbeat runs for zero limit

Alix-007:
## Summary
- Make `list_runs(workspace_dir, 0)` return an empty result instead of one stored heartbeat run.
- Preserve heartbeat history write-side pruning behavior for `max_history = 0`.

## Changes
- Return `Ok(Vec::new())` before querying heartbeat history when the requested list limit is zero.
- Keep `record_run` pruning clamped with `max_history.max(1)`.
- Add regression coverage for zero list limits while confirming a zero max-history record is still persisted.

## Tests
- `cargo test -p zeroclaw-runtime list_runs_zero_limit_returns_empty --lib`
- `cargo test -p zeroclaw-runtime heartbeat::store::tests --lib`
- `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings`

## Risk
- Low. This is limited to heartbeat history listing semantics for an explicit zero limit and does not add or duplicate state.


Audacity88 (approved):
Reviewed current head 7ab608355e4eae6ca06c48f988194136efea4375. I checked the live PR state, labels, comments and reviews, visible green CI, the one-file diff in `crates/zeroclaw-runtime/src/heartbeat/store.rs`, the old and new `list_runs` behavior, adjacent heartbeat store tests, caller search, and freshness against current `master`. This looks good as a focused runtime bug fix: a zero list limit now returns no heartbeat runs instead of being clamped up to one row.

### 🟢 What looks good — zero list limit is handled at the right boundary

The read-side fix is exactly where it should be. Returning `Ok(Vec::new())` before opening/querying the SQLite store makes `list_runs(workspace_dir, 0)` match the caller's explicit limit, while keeping normal nonzero listing behavior unchanged. The regression test covers the old failure mode directly: after recording a run, `list_runs(..., 0)` must be empty, and `list_runs(..., 10)` still proves the write-side history record remains available. That keeps the read-limit fix separate from the existing `record_run` pruning semantics.

### 🟡 Warning — refresh branch freshness before merge

This approval is for the patch itself, not a merge-ready statement. The branch is 20 commits behind current `master`, and the touched file is high-risk runtime heartbeat storage code, so before merging this should get a branch update with fresh CI, an exact merge-result smoke for `zeroclaw-runtime`, or explicit stale-risk acceptance after naming the triggers.

PR #7723, opened 2026-06-15
fix(channels): bypass mention_only gate for replies to bot messages

### LINKED ISSUE #5866: [Bug]: in the Telegram group bot ignores replies on its messages when mention_only=true
### Affected component

channel

### Severity

S1 - workflow blocked

### Current behavior

### Expected behavior

### Steps to reproduce

```bash
1. setup mention_only=true
```

### Impact

_No response_

### Logs / stack traces

```text

```

### ZeroClaw version

0.7.3

### Rust version

_No response_

### Operating system

Ubuntu 24.04

### Regression?

Unknown

### Pre-flight checks

dwc1997:
## Summary

- **What changed:** Bypass `mention_only` gate in Telegram groups when the message is a reply to the bot's own message.
- **Why:** When `mention_only=true`, the bot ignores replies to its own messages because the reply text doesn't contain an @mention. Users expect the bot to respond when they reply directly to its message.
- **Scope boundary:** `crates/zeroclaw-channels/src/telegram.rs` — `parse_update_message()` only. No changes to other channels or the mention_only config.
- **Blast radius:** Telegram group behavior only. DMs and non-mention_only configs are unaffected.
- **Linked issue(s):** Closes #5866

```bash
 crates/zeroclaw-channels/src/telegram.rs | 22 +++++++++++++++++-----
 1 file changed, 17 insertions(+), 5 deletions(-)
```

- **Commands run and tail output:** `git diff --stat` — verified only `telegram.rs` is changed.
- **Beyond CI — what did you manually verified:** Traced the logic: `reply_to_message.from.is_bot` is `true` when the replied-to message was sent by the bot. This bypasses the `contains_bot_mention` check while preserving the gate for non-reply messages.
- **If any command was intentionally skipped, why:** Remote CI will validate.

## Compatibility (required)

- Backward compatible? **Yes** — only changes behavior for replies to bot messages in groups with mention_only=true

---


singlerider (changes requested):
Read against current `master`: the `mention_only` gate in `crates/zeroclaw-channels/src/telegram.rs` and the surrounding bot-username handling. The intent is right and the UX gap is real, but the check is broader than the stated scope.

### 🔴 Blocking: `is_bot` matches any bot, not this bot

The comment and PR title both say "replies to the bot's own messages," but the predicate only reads `reply_to_message.from.is_bot`, which is `true` for *every* bot. In a group that contains another bot, a user replying to that other bot's message would satisfy `is_reply_to_bot` and bypass the `mention_only` gate, so this bot would process a message that was never directed at it. That is exactly the kind of unintended trigger `mention_only` exists to prevent.

The tighter check is cheap and uses machinery already present: compare `reply_to_message.from.username` against the cached `self.bot_username` (case-insensitive, `trim_start_matches('@')`), the same matching this file already does at the username sites. Telegram populates `from.username` for bot senders, so `reply_to_message.from.username == bot_username` precisely captures "reply to *my own* message." Please gate on that rather than the generic `is_bot` flag. (If you want to be robust to a future username change, matching on the bot's numeric `from.id` is an alternative, but username matching is consistent with the rest of this file.)

### 🔴 Blocking: no test for the new branch

This adds a behavioral fork in the `mention_only` path with no coverage. Please add cases for: reply-to-this-bot bypasses the gate, reply-to-a-different-bot does **not** bypass, and a non-reply group message without an @mention is still dropped. Those three pin both the fix and the boundary the first finding is about.

dwc1997:
Updated the fix to match this bot's username specifically instead of any bot's is_bot flag. The check now compares reply_to_message.from.username against the cached bot_username, so only replies to THIS bot bypass the mention_only gate.

Addresses @singlerider's review feedback.

Audacity88 (changes requested):
Reviewed current head `bd6f0e065f6778f8b740429ed66ba313d849d8b9` against the live PR state, singlerider's review, the author's follow-up comment, the current `crates/zeroclaw-channels/src/telegram.rs` diff, and visible GitHub checks. I did not run local validation.

Thanks for the update. The username-based reply check addresses the first blocker: it now compares the replied-to sender username with the cached bot username instead of treating any Telegram bot as this bot.

### ✅ Resolved — Reply bypass is now scoped to this bot

The current diff no longer uses the broad `reply_to_message.from.is_bot` predicate. It reads `reply_to_message.from.username`, strips any leading `@`, and compares it case-insensitively to the cached bot username. That means a reply to a different bot should fail this comparison and fall through to the normal `contains_bot_mention` gate.

### 🔴 Blocking — Add regression coverage for the new mention_only branch

The second blocker is still open: this PR changes a user-visible Telegram group routing path, but no tests were added for the new `is_reply_to_this_bot` branch.

Please add the cases singlerider requested:

- a reply to this bot bypasses the `mention_only` gate without an explicit mention
- a reply to a different bot does not bypass the gate
- a normal group message without an `@bot` mention is still dropped

Those tests would pin both sides of the new condition and protect the regression this PR is intended to fix without needing live Telegram credentials.

singlerider (changes requested):
Re-review at head bd6f0e06 against current master and my prior CHANGES_REQUESTED. Checked out the head locally.

### RESOLVED ✅ Reply bypass is now scoped to this bot

My first blocker was that the predicate read `reply_to_message.from.is_bot`, which is true for every bot, so a reply to a different bot in a group would bypass the `mention_only` gate. Fixed exactly as proposed (commit 6e1f3389):

```rust
let is_reply_to_this_bot = {
    let bot_username = self.bot_username.lock();
    bot_username.as_ref().and_then(|bot_name| {
        ...reply_to_message.from.username, trim_start_matches('@'), case-insensitive...
    })
};
```

It now compares the replied-to sender username against the cached `bot_username` with the same normalization the rest of this file uses, so a reply to another bot falls through to the normal mention gate. Correct scoping.

### 🔴 Blocking — the regression tests are still missing

My second blocker is unaddressed, and @Audacity88 concurs. This PR adds a behavioral fork in the group `mention_only` routing path with no coverage. The diff adds no `#[test]`/`#[tokio::test]`; the only commit since the username fix was `fix: cargo fmt`. Please add the three cases:

- a reply to this bot bypasses the gate without an explicit `@mention`
- a reply to a different bot does not bypass the gate
- a normal group message without an `@bot` mention is still dropped

Those pin both sides of the new `is_reply_to_this_bot` condition, which is the exact boundary the first finding was about, and they protect the fix from silent regression without needing live Telegram.

Add the coverage and this is approve-quality. The fix logic itself is correct.

Audacity88 (approved):
Reviewed current head 5f9450375d0347cd5bbeed5b981983fa3a9c149b against the live PR state, linked issue #5866, @singlerider's reviews, my prior review, the current `crates/zeroclaw-channels/src/telegram.rs` diff and tests, the visible GitHub checks, and the merge result against current `master`. I did not run local Cargo validation.

Thanks @dwc1997 for the update. This clears my prior request for changes: the reply detection is now scoped to this bot, and the requested regression tests are present. I treated the older #5866 follow-up note about a reply that also contains an explicit `@bot` mention as non-blocking here because current `master` already preserves explicit mentions in `mention_only` message bodies; this PR is the no-explicit-mention reply fix for the reported Telegram group bug. This is not a merge-ready signal yet because the branch is still behind current `master`, the same Telegram file changed on `master` after this branch point, and @singlerider's prior `CHANGES_REQUESTED` review remains active until they clear it.

### ✅ Resolved — Reply bypass is scoped to this bot

The new branch no longer treats every Telegram bot as this bot. It reads the `reply_to_message.from.username`, strips a leading `@`, and compares it case-insensitively with the cached bot username. A reply to another bot now falls through to the normal `contains_bot_mention` gate instead of bypassing `mention_only`.

### ✅ Resolved — The mention_only branch has regression coverage

The latest commit adds the three cases requested in review: reply to this bot bypasses the gate without an explicit mention, reply to a different bot does not bypass the gate, and a normal group message without an `@bot` mention is still dropped. Those tests pin both sides of the new condition and match the behavior needed for #5866.

### 🟢 What looks good — Narrow channel fix

The patch stays focused on Telegram text-message parsing. It does not change the `mention_only` config shape or broaden behavior for DMs and non-`mention_only` paths, which keeps the blast radius appropriate for this bug fix.

vyahhi:
This appears to duplicate [#7958](https://github.com/zeroclaw-labs/zeroclaw/pull/7958) for #5866. Difference: this PR uses the cached bot username and changes only the text mention gate; #7958 uses a cached bot id and covers both text and media-caption gates.

crates/zeroclaw-channels/src/telegram.rs
```diff
@@ -2220,13 +2220,36 @@ Allowlist Telegram username (without '@') or numeric user ID.",
...
         let is_group = Self::is_group_message(message);
         if self.mention_only && is_group {
-            let bot_username = self.bot_username.lock();
-            if let Some(ref bot_username) = *bot_username {
-                if !Self::contains_bot_mention(text, bot_username) {
+            // Bypass mention_only gate for replies to THIS bot's own
+            // messages. Check the replied-to message's username against
+            // our cached bot username (not just is_bot, which would
+            // match any bot in the group).
+            let is_reply_to_this_bot = {
```
singlerider, line 2227: 🟢 Correctly scoped to THIS bot: username compared against cached `bot_username` case-insensitively with `@` stripped, so a reply to another bot in the group falls through to `contains_bot_mention`. Resolves my prior is_bot-matches-any-bot blocker. (Minor: the `bot_username` lock is taken here and re-taken below — harmless with parking_lot scoping, not worth a change.)

crates/zeroclaw-channels/src/telegram.rs
```diff
@@ -7641,4 +7664,107 @@ mod tests {
...
+        });
+
+        assert!(
+            ch.parse_update_message(&update).is_some(),
+            "reply to this bot should bypass mention_only gate"
+        );
+    }
+
+    #[test]
+    fn mention_only_reply_to_different_bot_is_dropped() {
```
singlerider, line 7705: 🟢 This is the test that matters — cached `mybot`, reply to `otherbot` — it actually exercises the fall-through rather than asserting a no-op. Pins the exact boundary my first finding was about.

singlerider (approved):
Re-reviewed at head `4979b231` (checked out locally, ran the suite). Both my blockers are resolved and the fix is approve-quality. One coordination flag below that is a maintainer decision, not a code defect.

### RESOLVED ✅ — username-scoped predicate

`is_reply_to_this_bot` compares the replied-to sender's `username` against the cached `bot_username` (case-insensitive, `@`-stripped), so a reply to a *different* bot in a group falls through to the normal `contains_bot_mention` gate. This was my first blocker; correctly scoped.

### RESOLVED ✅ — regression coverage added

The three cases I asked for are present with real assertions on `parse_update_message`, and all pass locally (`cargo test -p zeroclaw-channels --lib mention_only` → 5 passed):

- `mention_only_reply_to_this_bot_bypasses_gate`
- `mention_only_reply_to_different_bot_is_dropped` (cached `mybot`, reply to `otherbot` — exercises the fall-through, not a no-op)
- `mention_only_group_non_mention_message_is_dropped`

### 🔴 Coordination — duplicate of #7958, both open against `telegram.rs`

#7958 (mazhuima) solves the identical `mention_only` reply-to-bot bypass in the same file and is already approved. Two open PRs touching the same gate logic will conflict — whichever lands first forces the other to rework or be closed. This PR is the older one (created 06-15), closes #5866, and carries the fuller regression set, so my preference is to land this and close #7958 as superseded. That's a maintainer call; I'm flagging it so we don't merge both. Not a defect in this code.

Approving the code on its merits.

Audacity88:
@dwc1997 thanks for carrying this through the reviewer fixes. I am closing this PR as superseded by #7958 rather than keeping two competing Telegram `mention_only` implementations open for #5866.

#7958 covers the same text-reply bug and also covers the media-caption gate, uses the bot id cached from `getMe`, and has the current approvals/checks on that broader version. #5866 should stay open until that PR lands.

PR #7724, opened 2026-06-15
fix(channels): respect ack_reactions config in Lark/Feishu channel

### LINKED ISSUE #5558: Feishu channel: ack_reactions=false has no effect

dwc1997:
## Summary

- **What changed:** Add `ack_reactions` field to `LarkChannel` with `with_ack_reactions()` builder, gate the fast-ack reaction paths on it, and wire `config.channels.ack_reactions` through the orchestrator initialization.
- **Why:** The Lark channel had a fast-ack reaction path that sent 👀 on incoming messages without checking the `ack_reactions` config. Setting `ack_reactions = false` had no effect on Feishu/Lark channels.
- **Scope boundary:** `lark.rs` (add field + builder + gate) and `orchestrator/mod.rs` (wire config). No changes to other channels.
- **Blast radius:** Lark/Feishu channel only. Other channels unaffected.
- **Linked issue(s):** Closes #5558

```bash
 crates/zeroclaw-channels/src/lark.rs             | 20 ++++++++++++++++++++
 crates/zeroclaw-channels/src/orchestrator/mod.rs |  3 ++-
 2 files changed, 22 insertions(+), 1 deletion(-)
```

- **Commands run and tail output:** `git diff --stat` — verified scope is limited to Lark channel and orchestrator wiring.
- **Beyond CI — what did you manually verified:** Traced both fast-ack paths (WS at line ~1347 and HTTP at line ~3370) and confirmed both are now gated on `self.ack_reactions`. Verified the orchestrator passes `config.channels.ack_reactions` to the builder.
- **If any command was intentionally skipped, why:** Remote CI will validate.

## Compatibility (required)

- Backward compatible? **Yes** — default `ack_reactions: true` preserves existing behavior

---

### DIFF
--- a/crates/zeroclaw-channels/src/lark.rs
+++ b/crates/zeroclaw-channels/src/lark.rs
@@ -662,10 +662,6 @@ pub struct LarkChannel {
     /// orchestrator from `[channels.lark.<alias>].draft_update_interval_ms`
     /// via [`Self::with_streaming`].
     draft_update_interval_ms: u64,
-    /// Whether to send ack reactions (👀, ✅, ⚠️) on incoming messages.
-    /// Defaults to `true`. Set from `[channels].ack_reactions` or the
-    /// per-channel override via [`Self::with_ack_reactions`].
-    ack_reactions: bool,
     /// Per-`message_id` timestamp of the last successful PATCH. Reads /
     /// writes are guarded by an async mutex so concurrent token streams
     /// cooperate on the same draft without racing the rate-limit window.
--- a/crates/zeroclaw-channels/src/lark.rs
+++ b/crates/zeroclaw-channels/src/lark.rs
@@ -734,7 +734,6 @@ impl LarkChannel {
             reaction_ids: Arc::new(tokio::sync::Mutex::new(std::collections::HashMap::new())),
             stream_mode: StreamMode::Off,
             draft_update_interval_ms: 1000,
-            ack_reactions: true,
             last_draft_edit: Arc::new(tokio::sync::Mutex::new(HashMap::new())),
             #[cfg(test)]
             api_base_override: None,
--- a/crates/zeroclaw-channels/src/lark.rs
+++ b/crates/zeroclaw-channels/src/lark.rs
@@ -662,6 +662,10 @@ pub struct LarkChannel {
     /// orchestrator from `[channels.lark.<alias>].draft_update_interval_ms`
     /// via [`Self::with_streaming`].
     draft_update_interval_ms: u64,
+    /// Whether to send ack reactions (👀, ✅, ⚠️) on incoming messages.
+    /// Defaults to `true`. Set from `[channels].ack_reactions` or the
+    /// per-channel override via [`Self::with_ack_reactions`].
+    ack_reactions: bool,
     /// Per-`message_id` timestamp of the last successful PATCH. Reads /
     /// writes are guarded by an async mutex so concurrent token streams
     /// cooperate on the same draft without racing the rate-limit window.
@@ -734,6 +738,7 @@ impl LarkChannel {
             reaction_ids: Arc::new(tokio::sync::Mutex::new(std::collections::HashMap::new())),
             stream_mode: StreamMode::Off,
             draft_update_interval_ms: 1000,
+            ack_reactions: true,
             last_draft_edit: Arc::new(tokio::sync::Mutex::new(HashMap::new())),
             #[cfg(test)]
             api_base_override: None,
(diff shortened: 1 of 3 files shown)


singlerider (approved):
Read against current `master`: the `ack_reactions` field, `with_ack_reactions`, and both reaction gates in `crates/zeroclaw-channels/src/lark.rs`, plus the orchestrator wiring. CI Format/Lint/Test are green. No active blocks.

### 🟢 What looks good: both reaction paths gated, default preserves behavior

Lark/Feishu was the channel ignoring `ack_reactions` and always emitting the 👀/✅/⚠️ acks. The fix threads the config through `with_ack_reactions` and gates *both* emission paths - the fast-ack spawn in the inbound handler and the generic `add_reaction` in the event handler at the `parse_event_payload_async` site - so a disabled config genuinely suppresses all acks rather than only one path. The field defaults to `true` and the orchestrator passes `config.channels.ack_reactions`, so existing deployments are unchanged. The field and method doc comments name the reactions and the config source, which is the right level of explanation. Gating the generic path with `&& state.channel.ack_reactions` inside the existing `if let` chain keeps it a single condition.

One observation, not blocking: the new branch (`ack_reactions = false`) has no dedicated test, so the existing `lark_inbound_ack_lifecycle_*` regressions still exercise only the enabled path. A small case asserting no reaction is sent when disabled would close that, but for a config-gated skip of an already-tested path with green CI it is not a gate.

Audacity88 (changes requested):
Reviewed the live PR state, linked issue #5558, singlerider’s prior approval of the newly gated reaction emitters/default behavior, empty comment threads, visible check state, the PR body, and the current diff/source around `LarkConfig`, the Lark reaction paths, and orchestrator wiring.

### 🟢 What looks good — Global Lark ack gating is wired into both reaction paths

The patch fixes the channel-wide path from #5558: `config.channels.ack_reactions` is passed into `LarkChannel`, the field defaults to `true`, and both Lark/Feishu reaction emitters now check the resolved field before adding the 👀 fast-ack. That preserves existing behavior by default while making the global disabled setting suppress both reaction paths.

### 🔴 Blocking — #5558 still has an unfixed per-channel path

Issue #5558 reports trying both the channel-wide `[channels_config]` value and the Feishu/Lark per-channel override. This PR fixes the channel-wide wiring by passing the resolved `config.channels.ack_reactions` value into `with_ack_reactions`, but the per-channel override is still absent: `LarkConfig` still has no `ack_reactions: Option<bool>` field like Telegram and Matrix do. In current config terms, a Lark/Feishu alias such as `[channels.lark.<alias>]` with `ack_reactions = false` is still ignored.

Because the PR still uses `Closes #5558`, merging this as-is would close an accepted bug while one of the reported config paths remains unresolved. Please either add the Lark/Feishu per-channel override to the schema, resolve it in the orchestrator with the same fallback shape used by Telegram/Matrix, and add a regression that proves an alias-level `ack_reactions = false` suppresses the Lark reaction path; or change the body/commit linkage to a non-closing reference and record the remaining per-channel work separately.

### 🟡 Warning — PR body risk metadata is stale

The live labels include `risk: medium`, but the PR body still says `risk: low` and keeps the low-risk rollback note. Since this is a medium-risk channel behavior change, please align the label snapshot and fill the medium-risk rollback fields before merge: fast rollback command/path, feature flags or config toggles, and observable failure symptoms.

Audacity88:
Thanks for the update here. Since #7495 has now merged the full #5558 fix, including the Lark/Feishu per-channel `ack_reactions` override and the reaction-path gating, this PR is now superseded.

The remaining diff against current `master` is behavior-equivalent builder-chain cleanup around wiring that already exists on the base branch, so I am going to close this rather than ask you to rework a PR whose bug-fix scope has already landed.

PR #7730, opened 2026-06-15
fix(skills): parse skill review summary from tool history

### LINKED ISSUE #6644: Skill review fork: `💾` summary parser misses tool receipts
## Summary

The post-turn skill review fork landed in #6594 emits a one-line `💾 Skill review: …` summary after it takes action. The intent: parse `SkillManageTool` receipts like `"Patched skill 'file-lister'."` into compact action strings like `"patched file-lister"`. In practice, the receipts vector is empty at the parse site, so [`summarize_actions`](https://github.com/zeroclaw-labs/zeroclaw/blob/master/crates/zeroclaw-runtime/src/skills/review.rs) falls back to showing the first 80 chars of the LLM's final text. The patch itself still happens correctly — this is purely a display bug.

## Repro (verified locally during #6594 smoke testing)

```bash
# A SKILL.toml-format skill needs to exist; the fork's tool only handles that format (see #N for `manifest.toml` follow-up).
mkdir -p ~/.zeroclaw/workspace/skills/file-lister
cat > ~/.zeroclaw/workspace/skills/file-lister/SKILL.toml <<'TOML'
[skill]
name = "file-lister"
description = "Helper for listing files."
version = "0.1.0"
TOML

echo "y" | cargo run --bin zeroclaw -- agent --provider openai --model gpt-4o \
  -m 'use shell to list ~/.zeroclaw/workspace. then save this preference into the file-lister skill by calling skill_manage with action=patch: when asked for directory listings, give a flat bullet list. always.'
```

**Observed:**
```
💾 Skill review: Patched skill 'file-lister' to provide a flat bullet list by default, as per use
```

That's an 80-char prefix of the LLM's *text* response, not a parsed receipt — the giveaway is the trailing `"use"` (clipped from `"user preference"`).

**Expected:**
```
💾 Skill review: patched file-lister
```

The patch *did* happen — `SKILL.toml` has the new `updated_at`, `improvement_reason`, audit-trail comments — so this is purely the display layer.

## Root cause (hypothesis)

`run_tool_call_loop` accepts an `Option<&Mutex<Vec<String>>>` for `collected_receipts`. The review fork passes `Some(&receipts)` ([review.rs](https://github.com/zeroclaw-labs/zeroclaw/blob/master/crates/zeroclaw-runtime/src/skills/review.rs)), but on inspection during smoke testing the vector was empty after the fork returned. So either:

1. `run_tool_call_loop` only populates `collected_receipts` for a specific tool subset (e.g. only the official receipt-generating tools), and arbitrary `Tool::execute` results don't flow into it.
2. The receipt collection requires a `receipt_generator: Option<&ReceiptGenerator>` to be paired with `collected_receipts`, which the fork passes as `None`.

Either way, the fork either needs to provide a `ReceiptGenerator` or my `summarize_actions` needs a different signal source.

## Fix shape

Two paths:

**Option A — inspect the fork's history instead of receipts.** After `run_tool_call_loop` returns, scan `review_history` for the most recent tool-result messages whose content matches `"Patched skill 'X'."` / `"Wrote ... for skill 'X'."` / `"Archived skill 'X'..."`. This avoids depending on `run_tool_call_loop`'s receipt mechanism entirely.

**Option B — wire a `ReceiptGenerator` into the fork.** Requires understanding what generators expect; bigger surgery.

Option A is the smaller, lower-risk fix.

## Impact

Cosmetic only. The patch happens correctly; the user sees a slightly clipped LLM-prose line instead of a tidy action verb. Worth fixing but doesn't block #6594.

## Files touched (Option A)

- `crates/zeroclaw-runtime/src/skills/review.rs` — change `summarize_actions` to take `&[ChatMessage]` instead of (or in addition to) `&Mutex<Vec<String>>`. Update existing unit tests in the same file.

Related: #6594

Pick-cat:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Skill review `summarize_actions` now falls back to parsing `tool`-role messages in `review_history` when `collected_receipts` is empty.
  - Fixes the cosmetic display bug where users saw an 80-char clip of LLM prose instead of `patched <slug>`.
- **Scope boundary:** Does not change skill_manage execution or receipt-generator wiring.
- **Blast radius:** Skill review fork summary line only.
- **Linked issue(s):** Closes #6644
- **Labels:** `bug`, `runtime`, `risk: low`, `size: XS`

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-runtime summarize_actions
```

- **Commands run and tail output:** `cargo fmt --all -- --check` passed locally. Full `cargo test` skipped — no system linker (`cc`) in environment; CI will run the battery.
- **Beyond CI — what did you manually verify?** Added unit test `summarize_actions_reads_skill_manage_results_from_history` covering the empty-receipt path.
- **If any command was intentionally skipped, why:** Missing `build-essential` / `cc` linker locally.

## Compatibility (required)

### DIFF
--- a/crates/zeroclaw-runtime/src/skills/review.rs
+++ b/crates/zeroclaw-runtime/src/skills/review.rs
@@ -87,6 +87,7 @@ pub async fn maybe_run_skill_review(
     let review_input = build_review_input(&failed_slugs);
 
     let mut review_history = history;
+    let fork_start_len = review_history.len();
     review_history.push(ChatMessage::user(&review_input));
 
     let receipts: Mutex<Vec<String>> = Mutex::new(Vec::new());
@@ -135,7 +136,8 @@ pub async fn maybe_run_skill_review(
 
     match result {
         Ok(final_text) => {
-            let summary = summarize_actions(&receipts, &final_text);
+            let summary =
+                summarize_actions(&receipts, &review_history[fork_start_len..], &final_text);
             if !summary.is_empty() {
                 println!(
                     "{}",
@@ -199,7 +201,15 @@ fn count_tool_iterations(history: &[ChatMessage]) -> usize {
 
 /// Convert the review's tool receipts + final text into a one-line summary
 /// for the user. Returns "" if the fork did nothing notable.
-fn summarize_actions(receipts: &Mutex<Vec<String>>, final_text: &str) -> String {
+///
+/// `fork_history` must contain only messages produced by the review fork
+/// itself (not the parent turn), so that the fallback scan does not pick up
+/// tool results from the user's main turn.
+fn summarize_actions(
+    receipts: &Mutex<Vec<String>>,
+    fork_history: &[ChatMessage],
+    final_text: &str,
+) -> String {
     let receipts = receipts.lock().ok();
     let actions: Vec<String> = receipts
         .as_ref()
@@ -208,6 +218,18 @@ fn summarize_actions(receipts: &Mutex<Vec<String>>, final_text: &str) -> String 
 
     if !actions.is_empty() {
         return actions.join(" · ");
+    }
+
+    // The fork passes `collected_receipts` but `run_tool_call_loop` does not
+    // populate it without a receipt generator. Fall back to the tool-result
+    // messages produced by the review fork itself (not the parent turn).
+    let from_history: Vec<String> = fork_history
+        .iter()
+        .filter(|m| m.role == "tool")
+        .filter_map(|m| extract_action_summary(&m.content))
+        .collect();
+    if !from_history.is_empty() {
+        return from_history.join(" · ");
     }
 
     let trimmed = final_text.trim();
@@ -316,7 +338,7 @@ mod tests {
             "Patched skill 'deploy'.".to_string(),
             "Wrote references/staging.md for skill 'deploy'.".to_string(),
         ]);
-        let summary = summarize_actions(&receipts, "");
+        let summary = summarize_actions(&receipts, &[], "");
         assert!(summary.contains("patched deploy"));
         assert!(summary.contains("wrote file for deploy"));
     }
@@ -324,14 +346,14 @@ mod tests {
     #[test]
     fn summarize_actions_empty_for_nothing_to_save() {
         let receipts = Mutex::new(Vec::new());
-        let summary = summarize_actions(&receipts, "Nothing to save.");
+        let summary = summarize_actions(&receipts, &[], "Nothing to save.");
         assert_eq!(summary, "");
     }
 
     #[test]
     fn summarize_actions_empty_when_no_receipts_and_no_text() {
         let receipts = Mutex::new(Vec::new());
-        let summary = summarize_actions(&receipts, "");
+        let summary = summarize_actions(&receipts, &[], "");
         assert_eq!(summary, "");
     }
 
@@ -340,6 +362,7 @@ mod tests {
         let receipts = Mutex::new(Vec::new());
         let summary = summarize_actions(
             &receipts,
+            &[],
             "Noted that deploy needs DEPLOY_TOKEN.\nMore details below.",
         );
         assert!(summary.starts_with("Noted that deploy"));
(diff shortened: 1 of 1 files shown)


WareWolf-MoonWall (approved):
Clean targeted fix — the fallback from `collected_receipts` to tool-role history messages is the right behavior when the skill completes without generating machine-readable receipts. Approving.

### 🟢 What looks good — Correct fallback logic

Adding `history: &[ChatMessage]` as a parameter and falling back to `extract_action_summary()` on tool-role messages when receipts are empty is composable and doesn't change the existing code path for the non-empty receipt case. The new test `summarize_actions_reads_skill_manage_results_from_history` covers exactly the scenario being fixed.

### 🟢 What looks good — No new state, no SSOT concern

The function reads from a slice passed in — it doesn't store the history or cache any derived value. The source of truth for the history remains wherever `run_skill_review` holds it.

### 🟡 Warning — Missing `bug`, `risk: low`, `size: XS` labels

The PR body lists these labels but the GitHub label set only shows `runtime` and `skills`. The auto-labeler may not have applied them. A maintainer with label permission should add `bug`, `risk: low`, and `size: XS` before merge — not blocking but worth noting.

### 🔵 Suggestion — `m.role == "tool"` string literal

The `m.role == "tool"` comparison is consistent with how this codebase handles `ChatMessage` roles elsewhere, so it's not a concern for this PR. If a `const ROLE_TOOL: &str = "tool"` ever gets introduced for deduplication, this spot would be an obvious update target.

Audacity88 (changes requested):
Reviewed the current diff, linked issue #6644, the existing approval, the passing check snapshot, and public source context for how the review fork receives history and how native tool results are appended.

### 🟢 What looks good — Receipt precedence stays intact

The direction of the fix is right: when real collected receipts exist, the PR still returns those first, so the established summary path remains unchanged. Reusing the existing action-summary parser for the fallback is also a good narrow approach for #6644.

### 🔴 Blocking — Limit the fallback to messages produced by the review fork

The prior approval called the fallback composable, and I agree with that direction. The part I think needs one more change is the scan boundary: `review_history` starts as the parent turn history (`let mut review_history = history`) before the review prompt is appended, but the new fallback scans every `role == "tool"` message in that whole vector.

When `collected_receipts` is empty and the review fork ultimately saves nothing, an earlier native tool output from the user’s main turn with text matching the existing action parser, such as `Patched skill 'slug'.`, `Wrote ... for skill 'slug'.`, or `Archived skill 'slug' ...`, can produce a false `💾 Skill review: patched slug` summary. The post-turn hook passes `history.clone()` into the review fork, and native tool results are recorded as `role=tool`, so the fallback needs to distinguish parent-turn tool messages from fork-generated tool messages.

Please bound the fallback to messages added by the review fork itself, for example by recording the index after pushing the review prompt and passing only `&review_history[review_start_len..]` into `summarize_actions`, or by collecting the loop’s `new_messages_out`. A regression test should cover a pre-existing matching tool message plus a no-op review result so this cannot report a save the fork did not perform.

Pick-cat:
Addressed @Audacity88's blocking review in commit 2c4ffbbf:

### Scope fallback to fork-only messages

The `summarize_actions` fallback now receives only `&review_history[fork_start_len..]` instead of the full `review_history`. `fork_start_len` is captured before the review prompt is pushed, so only tool messages produced by the review fork itself are scanned — parent-turn tool results can no longer produce false summaries.

### Regression test added

`summarize_actions_ignores_parent_turn_tool_messages` covers the exact scenario: the fork produced no tool messages (`fork_history = vec![]`), and the final text is "Nothing to save." — the summary must be empty, even if the parent turn had matching tool output (which is excluded from the slice).

Audacity88 (approved):
@Pick-cat, I checked the live PR state, current head `cab7cb19`, linked issue #6644, the prior blocking review, the one-file runtime diff, the PR body metadata, live labels, and the current public check snapshot. I did not run local Cargo validation.

### ✅ Resolved — Fallback scan is now scoped to the review fork

The latest head records the parent-history length before adding the review prompt and passes only `&review_history[fork_start_len..]` into `summarize_actions()`. That excludes parent-turn tool messages while still letting fork-produced tool results feed the empty-receipt fallback, so the false `💾 Skill review: patched ...` summary I blocked on is addressed.

### 🟢 What looks good — Receipt precedence and display-only scope stay narrow

The fallback still only runs after collected receipts are empty, and the new history path reuses the existing action-summary parser. The added history fallback test covers the original empty-receipt display path, and the fork-only slice keeps no-op review results from being summarized from parent-turn tool output.

### 🟡 Warning — PR body metadata still needs cleanup

The live labels are `bug`, `size: XS`, `risk: high`, `runtime`, and `skills`, while the PR body lists `bug`, `runtime`, `risk: low`, and `size: XS`. Since this touches `zeroclaw-runtime`, the high-risk label is mechanically expected even though the blast radius is only the skill-review summary line. Please update the PR body label snapshot to include `skills` and `risk: high`, and replace the low-risk rollback line with the required medium/high-risk rollback fields before merge. This is not a code blocker.

This clears my prior block. This is not a merge-ready statement by itself.

PR #7732, opened 2026-06-15
fix(self-test): authenticate websocket handshake probe

### LINKED ISSUE #7038: zeroclaw check 11/11 websocket 401 despite valid auth profile
zeroclaw check fails 11/11 with 401 even when gateway auth profile is configured and websocket works manually with bearer token

Pick-cat:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `zeroclaw self-test` websocket check now sends `Authorization: Bearer` (and `?token=` when pairing is required) plus the required `?agent=` query param, matching how real clients connect.
  - Fixes false 401 failures when the gateway has pairing enabled.
  - When no agent alias is configured, the probe URL omits `?agent=` entirely (no `"default"` fallback that could 404 on instances without a default agent).
- **Scope boundary:** Does not change gateway auth policy or websocket handler behavior.
- **Blast radius:** Full self-test suite websocket check only.
- **Linked issue(s):** Closes #7038
- **Labels:** `bug`, `risk: high`, `size: S`, `gateway:ws`

## Validation Evidence

CI run [27922918229](https://github.com/zeroclaw-labs/zeroclaw/actions/runs/27922918229) on SHA `9df96dcf` — all visible checks green, including CI Required Gate.

```bash
cargo fmt --all -- --check   # passed (CI-validated)
cargo clippy --all-targets -- -D warnings   # passed (CI-validated)
cargo test   # passed (CI-validated, including resolve_gateway_bearer_token and build_websocket_probe_url unit tests)
```

- **Beyond CI — what did you manually verify?** Added unit tests for plaintext vs hashed `paired_tokens` resolution, plus `build_websocket_probe_url` tests covering agent alias + token, no-agent + token, no-pairing, and no-token cases. Reviewed that the probe URL no longer uses a `"default"` alias fallback and that the no-agent pairing path builds `?token=` rather than `&token=`.
- **If any command was intentionally skipped, why:** No local `cc` linker; CI run 27922918229 on SHA `9df96dcf` provides the compile + test signal.

## Security & Privacy Impact

- New external network calls? **No** (same probe as before, now with auth)
- Secrets / tokens / credentials handling changed? **No** (reads existing env/config tokens for local diagnostics only)

## Compatibility

- Config / env / CLI surface changed? **No** (accepts optional `ZEROCLAW_GATEWAY_TOKEN` / existing `ZEROCLAW_ACP_BRIDGE_TOKEN`)

## Rollback

- **Fast rollback command/path:** Revert the eventual squash merge commit for #7732.
- **Observable failure symptoms:** `zeroclaw self-test` websocket checks report authentication or handshake failures, or websocket probe URLs are formed incorrectly for pairing/no-agent cases.


WareWolf-MoonWall (changes requested):
The direction is right — the websocket self-test probe was sending an unauthenticated handshake and this adds token resolution + bearer header injection. A few things need attention before this is merge-ready.

### 🔴 Blocking — 2 CI checks are currently failing

`gh pr view 7732` reports 2 of 17 checks failing. I need those to be green before approving. Please check the Actions tab, resolve whatever is failing, and re-request review.

### 🟡 Warning — `"default"` alias fallback is not guaranteed to exist

In `check_websocket_handshake()` the agent alias for the probe URL falls back to `"default"` when no alias is configured. If no agent named `default` is registered in the running instance, the gateway will return a 404 or 400 on that `?agent=default` path even after auth succeeds. The probe would then report failure for the wrong reason. Consider either using the first available alias from config or accepting that the probe is a best-effort connection check (and document that `?agent=default` may 404 in single-instance setups that don't use the default alias name).

### 🟡 Warning — Validation evidence is partial

`cargo fmt` passed, but `cargo test` and a full compile were skipped due to missing `cc` linker in the authoring environment. CI is the fallback — once CI is green on a fresh head, that satisfies the compile/test requirement. Please note the CI run SHA explicitly in the validation evidence rather than leaving it open.

### 🟢 What looks good — Token resolution priority

`resolve_gateway_bearer_token()` has a clear, documented resolution order: env `ZEROCLAW_GATEWAY_TOKEN` → `ZEROCLAW_ACP_BRIDGE_TOKEN` → first `zc_*` paired token from config. The `hashed_token_pattern` filter that rejects `$2b$`-prefixed bcrypt hashes is the right safety net — the probe needs a plaintext token, not a stored hash.

### 🟢 What looks good — `.expect("writing to String cannot fail")`

Documenting the invariant inline (a `write!` to `String` cannot return an error) satisfies the AGENTS.md rule against bare `expect()`. Well done.

Pick-cat:
Thanks for the review! All three points addressed:

1. **CI checks** — re-triggered after a transient `apt-get` mirror sync failure in the 32-bit check runner. All 17 checks green on the latest run ([27585676211](https://github.com/zeroclaw-labs/zeroclaw/actions/runs/27585676211)).

2. **`"default"` alias fallback** — removed. When no agents are configured, the probe URL now omits `?agent=` entirely (`ws://{host}:{port}/ws/chat`), matching the original pre-PR behavior. This avoids a spurious 404 on instances that don't have an agent named "default".

3. **Validation evidence** — CI run 27585676211 (SHA `9aa67d34e`) confirms compile + test pass on all platforms. Latest commit `5eb85928f` adds the alias fallback fix.

Pick-cat:
All review feedback addressed: (1) CI green on latest run 27586497639 (SHA 5eb85928f), (2) `"default"` alias fallback removed — probe URL omits `?agent=` when no alias configured, (3) validation evidence updated with CI run SHA. @WareWolf-MoonWall could you re-review?

Pick-cat:
CI is now fully green (17/17 checks pass) and all three review points have been addressed in commit 5eb85928f. Ready for re-review. @WareWolf-MoonWall could you update the review decision?

WareWolf-MoonWall (approved):
Re-review at head 5eb85928. All three items from my 2026-06-15 CHANGES_REQUESTED are resolved. Approving.

### ✅ Resolved — CI checks are green

17/17 checks passing on CI run 27586497639 at 5eb85928. The two previously failing checks are confirmed fixed.

### ✅ Resolved — `"default"` alias fallback removed

The probe URL now uses `config.agents.keys().next()` to take the first configured alias, or omits `?agent=` entirely if no agents are configured. No hardcoded `"default"` fallback.

### ✅ Resolved — Validation evidence complete

Two new unit tests added: `resolve_gateway_bearer_token_reads_plaintext_paired_token` and `resolve_gateway_bearer_token_ignores_hashed_paired_tokens`. Token resolution is covered at the unit level and CI provides the compile/integration signal.

### 🟢 What looks good — Token resolution priority

`resolve_gateway_bearer_token()` resolution order (`ZEROCLAW_GATEWAY_TOKEN` → `ZEROCLAW_ACP_BRIDGE_TOKEN` → first `zc_*` paired token, hashed excluded) is documented and tested. The `$2b$`-prefixed bcrypt filter is the right safety net.

singlerider:

Audacity88 (changes requested):
@Pick-cat I re-reviewed current head 650b565 after the master refresh, the linked issue #7038, the prior review/comment thread, the full `src/commands/self_test.rs` diff, the gateway WebSocket auth path, and the current CI. All visible checks are green, and I did not run local Cargo validation. The prior CI, default-alias, and validation-evidence concerns are resolved, but the no-agent fallback still has one URL construction bug.

### ✅ Resolved — Prior websocket probe concerns

The current diff removes the hardcoded `default` alias fallback, adds bearer-token resolution for the self-test probe, and keeps the token source aligned with the gateway pairing model: env token first, then plaintext `zc_*` entries, while hashed `paired_tokens` are not treated as bearer secrets. That addresses the main 401 path from #7038 for configured-agent installs.

### 🔴 Blocking — Token query appends with `&` even when no query string exists

The fallback branch for "no configured agent alias" now builds `ws://{host}:{port}/ws/chat`, but the pairing branch always appends the token as `&token={token}`. In the no-agent + `require_pairing = true` case, that produces `ws://{host}:{port}/ws/chat&token=...` instead of `ws://{host}:{port}/ws/chat?token=...`. That is not a valid query string on the `/ws/chat` route, so the fallback branch can fail before it proves the intended WebSocket auth behavior.

Please either append `?token=` when the URL does not already contain a query string, or build the probe URL with query pairs so the delimiter cannot drift. A regression test for `require_pairing = true`, no configured agents, and a plaintext/env token would cover the fallback that this PR added to avoid `?agent=default`.

### 🟢 What looks good — Gateway auth model alignment

Using `Authorization: Bearer <token>` for the request, while also supporting the existing browser-compatible query-token path when pairing is required, matches the gateway WebSocket extractor. The `zc_*` plaintext fallback is also consistent with `PairingGuard::new`, which hashes plaintext tokens on load for backward compatibility.

Pick-cat:
@Audacity88 Thanks — your blocking review caught a real URL-construction bug. Fixed in commit `4ff69ce28`.

### Root cause

The pairing branch in `check_websocket_handshake()` always appended the token as `&token={token}`, regardless of whether the URL already contained a `?`. On the no-agent fallback path (`ws://{host}:{port}/ws/chat` with no `?agent=`), that produced `ws://.../ws/chat&token=...` — not a valid query string on the `/ws/chat` route, so the probe could fail for the wrong reason before exercising the intended WebSocket auth behavior.

### Fix

Extracted the URL construction into a pure helper `build_websocket_probe_url(probe_host, port, alias, require_pairing, token)` that picks the separator based on whether the URL already contains `?`:

```rust
let sep = if url.contains('?') { '&' } else { '?' };
url.push(sep);
url.push_str("token=");
url.push_str(token);
```

Behavior matrix:

| `alias` | `require_pairing` | `token` | Result |
|---|---|---|---|
| `Some("dev")` | `true` | `Some("zc_test")` | `ws://.../ws/chat?agent=dev&token=zc_test` |
| `None` | `true` | `Some("zc_test")` | `ws://.../ws/chat?token=zc_test` (the regression path) |
| `Some("dev")` | `false` | `Some("zc_test")` | `ws://.../ws/chat?agent=dev` |
| `None` | `true` | `None` | `ws://.../ws/chat` (caller still fails fast on missing token) |

The early-fail path when `require_pairing && token.is_none()` is preserved; the helper just builds the URL given whatever token (or lack thereof) the caller resolved. As a side benefit, `resolve_gateway_bearer_token` is now called once instead of three times.

### Regression tests

Four unit tests on `build_websocket_probe_url`, including the exact case from your review (`require_pairing = true`, no configured agents, plaintext/env token) — `build_websocket_probe_url_uses_question_mark_when_no_alias` asserts `ws://127.0.0.1:42617/ws/chat?token=zc_test`. The other three cover the `&` separator on the agent-alias branch, the no-pairing case, and the no-token case.

### Validation

No local `cc` linker (same constraint as the rest of this PR), so I'm relying on CI. The new head is `4ff69ce28`; once the `Test` and `CI Required Gate` checks rerun they should stay green — the change is test-only at the URL-construction layer plus a refactor of the surrounding call site.

WareWolf-MoonWall (commented):
Re-reviewing at head 7aa04ee9. To orient the thread: I APPROVED at 5eb85928 (commit 650b565ecdb5, after singlerider's master refresh). @Audacity88 then posted CHANGES_REQUESTED at that same commit with one new blocking finding. @Pick-cat pushed 7aa04ee9 in response. Audacity88's CR is still active; per the review protocol I cannot approve over an active `CHANGES_REQUESTED`, so this is a `--comment`.

Taking stock: Audacity88's 🔴 URL-construction bug is fixed in 7aa04ee9. All other items from this PR's history are settled. CI is green at the new head.

### ✅ Resolved — URL query string separator is correct in the no-alias fallback (Audacity88's 🔴)

`build_websocket_probe_url` (new pure helper) picks the separator with:

```rust
let sep = if url.contains('?') { '&' } else { '?' };
```

In the no-alias path the base URL is `ws://{host}:{port}/ws/chat` (no query string yet), so the token appends as `?token={token}` — not `&token={token}`. That produces a valid query pair on the `/ws/chat` route. The regression test `build_websocket_probe_url_uses_question_mark_when_no_alias` asserts the exact case from Audacity88's finding:

```
ws://127.0.0.1:42617/ws/chat?token=zc_test
```

The comment on that test cross-references PR #7732 and the prior broken output `ws://.../ws/chat&token=...`, which makes it easy to understand why the test exists.

### 🟢 What looks good — `resolve_gateway_bearer_token` is now called once

Extracting `build_websocket_probe_url` as a self-contained helper means `resolve_gateway_bearer_token` is called exactly once in the handshake path. The previous implementation called it up to three times in the same function body. This is an incidental cleanup that makes the auth-resolution flow unambiguous.

### 🟢 What looks good — behavior matrix is fully covered

The four tests on `build_websocket_probe_url` map cleanly to the four rows in the behavior matrix @Pick-cat posted in the response comment: agent-alias + pairing + token (`&` separator), no-alias + pairing + token (`?` separator, the regression case), agent-alias + no pairing (no token appended), no-alias + pairing + no token (no token appended). All four branches are tested at the unit level, CI is green at 7aa04ee9.

---

@Audacity88 — the URL construction bug is fixed and the four unit tests cover it directly, including the exact case from your blocking finding. When you have a moment, please re-review 7aa04ee9 and dismiss or convert your CR if everything looks right. CI is clean and the branch is otherwise ready.

Pick-cat:
@Audacity88 Friendly re-review ping — your blocking URL-construction finding is addressed at head `7aa04ee9`.

**Fix:** `build_websocket_probe_url` now picks `?` vs `&` based on whether the URL already contains a query string, so the no-alias fallback produces `ws://{host}:{port}/ws/chat?token=...` instead of `.../ws/chat&token=...`.

**Coverage:** `build_websocket_probe_url_uses_question_mark_when_no_alias` asserts the exact regression case from your review.

**CI:** 18/18 checks passing on the current head.

Thanks again for catching this — ready when you have a moment to re-review.

Audacity88 (approved):
@Pick-cat I re-reviewed head 7aa04ee9 after your update. I checked the live PR state, #7038, the prior review thread, the current one-file diff in `src/commands/self_test.rs`, the gateway `/ws/chat` token extractor, and the current GitHub checks. I also checked merge freshness. I did not run local Cargo validation.

This clears my prior request-changes review. It is not a merge-ready call yet: the branch is now 115 commits behind current `master`, current `master` has also changed `src/commands/self_test.rs`, and the passing checks on this head do not cover a merge result with current `master`. The branch still needs an update with fresh CI, an exact merge-result smoke, or an explicit maintainer stale-risk acceptance before landing.

### ✅ Resolved — No-agent pairing URL now uses a real query string

The blocker from my prior review is fixed. `build_websocket_probe_url()` now chooses `?` when the base URL has no query string and `&` when `?agent=` is already present. That means the no-agent + pairing-required path now builds `ws://127.0.0.1:42617/ws/chat?token=zc_test`, not `ws://.../ws/chat&token=...`.

The regression test `build_websocket_probe_url_uses_question_mark_when_no_alias` covers the exact fallback path that was broken before. The other helper tests cover the agent-alias `&token=` case, no-pairing case, and no-token helper output.

### ✅ Resolved — The probe still matches the gateway auth model

The gateway `/ws/chat` handler accepts bearer auth from the `Authorization` header first, then bearer subprotocol, then `?token=`. The self-test probe now resolves one plaintext diagnostic token, sends it as `Authorization: Bearer ...`, and also appends `?token=` when pairing is required so the browser-compatible query-token path is exercised too.

The early failure when pairing is required but no plaintext token is available is the right behavior for a diagnostic command: it tells the operator why the probe cannot authenticate instead of returning a misleading websocket failure.

### 🟢 What looks good — The fix stays scoped to the diagnostic client

This PR fixes the self-test client side of #7038 without weakening gateway auth policy or changing the websocket handler. The linked issue is a good fit: `zeroclaw check` was probing `/ws/chat` differently from a real authenticated client, and this patch makes the diagnostic path present the configured bearer token.

CI is green on head 7aa04ee9, including `Test` and `CI Required Gate`. The remaining work is branch freshness, not this review blocker.

Issue #6695, opened 2026-05-16
[Feature]: Add skills management to the gateway web UI


NiuBlibing:
### Summary

Add a "Skills" management section to the gateway web UI (dashboard) for listing, viewing, enabling/disabling, and editing skills.

### Problem statement

Skills are a first-class runtime concept in ZeroClaw (see `crates/zeroclaw-runtime/src/skills/`) and the gateway already exposes management surfaces for adjacent concerns — config (`api_config.rs`), plugins (`api_plugins.rs`), personality (`api_personality.rs`), onboarding, pairing — but there is no equivalent for skills. Operators currently have to edit files on disk or use the CLI to inspect or change skill state, which is friction-heavy and inconsistent with how other extensible subsystems are managed.

### Proposed solution

1. Add a skills HTTP API to `zeroclaw-gateway` (e.g. `crates/zeroclaw-gateway/src/api_skills.rs`) following the existing `api_plugins.rs` pattern, exposing endpoints such as:
   - `GET /api/skills` — list skills with metadata (name, description, source, enabled state, last-modified)
   - `GET /api/skills/{name}` — view a single skill's manifest and SKILL.md content
   - `POST /api/skills/{name}/enable` and `/disable` — toggle activation
   - `POST /api/skills/reload` — re-scan skills directory
2. Add a "Skills" page/tab in the dashboard web UI alongside existing tabs (channels, plugins, etc.) consuming those endpoints.
3. Reuse existing skills audit logic (`crates/zeroclaw-runtime/src/skills/audit.rs`) to surface validation status per skill in the UI.
4. Respect the same auth/path-policy guards as other admin endpoints.

### Non-goals / out of scope

- In-browser skill creation/editing (creator/improver flows can stay CLI-only in v1).
- Uploading skills via the UI.
- Multi-user RBAC for skills beyond the existing gateway auth model.

### Alternatives considered

- Keep skills CLI-only — leaves a gap relative to plugins/config which already have UI surfaces.
- Surface skills only via the existing config endpoint — couples unrelated concerns and doesn't fit the per-skill manifest/audit shape.

### Acceptance criteria

- New `/api/skills*` endpoints documented and covered by tests.
- Dashboard surfaces a Skills tab listing installed skills with their metadata and audit status.
- Enable/disable round-trips through runtime state correctly.
- Auth and path-policy behave consistently with `api_plugins.rs`.

### Architecture impact

[...]

PR #7734, opened 2026-06-15
feat(skills): expose frontmatter tags in the editor + slash-command toggle

Nillth:
## Summary

- **Base branch:** `master` — **stacked on #7490** (`Depends on #7490`); review the skill-tags editor layer above the dynamic-skill-command runtime.
- **What changed and why:**
  - **Tags API fix (the load-bearing bit):** `SkillFrontmatter` — the single source of truth for the skill editor (dashboard form, CLI, TUI) — had no `tags` field, even though the loader reads `tags:` from the same YAML. So the editor couldn't see tags *and silently stripped them on save* (`PUT /api/skills/.../{name}` rewrites the frontmatter from `SkillFrontmatter`). This adds `tags: Vec<String>` to the struct + `prop_fields` (`StringArray`), and round-trips it through `SkillDocument` parse/serialize (inline `tags: [a, b]` and block `- item`, matching the loader's parser). Net: **editing a skill no longer drops its tags** — which #7490 depends on, since a stripped `slash` tag would silently deregister the skill's command.
  - **Skills page UI:** the bundle's skill editor now shows tags as **badges** (removable, with an add input), excluding the two special tags, plus a **"Slash command" toggle**. Toggling it writes/removes the `slash` tag that #7490 turns into a Discord `/command`. `open-skills` (loader-managed, community-synced) is shown read-only and never user-editable. The full tag list is preserved on save.
- **Scope boundary:** the skill **frontmatter editing surface** (`SkillFrontmatter` + `SkillDocument` round-trip + the dashboard `SkillsBundleEditor`). No change to the slash-command *runtime* (#7490 owns that), the loader, or any other config. The CLI `skills add` scaffolds a **tagless** skill (a `--tags` flag is a reasonable follow-up; tag management is the editor's job here).
- **Blast radius:** `GET/PUT /api/skills/bundles/:alias/skills/:name` now carry `tags`; SKILL.md files written by the editor gain a `tags: [...]` line when tags are set (omitted when empty, so tagless skills are byte-identical). The dashboard/CLI/TUI gain the field via `prop_fields`.
- **Linked issue(s):** `Depends on #7490` (the consumer). `Related #6695` (skills management in the gateway web UI).
- **Labels:** `enhancement`, `skills`, `web`, `gateway`, `risk: low`, `size: M`.

Backend + web change. Local battery:

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-runtime -p zeroclawlabs --all-targets -- -D warnings
cargo test -p zeroclaw-runtime skills::
(cd web && npm ci && npm run build)
```

  - `cargo fmt --all -- --check` → **exit 0**.
  - `cargo clippy -p zeroclaw-runtime -p zeroclawlabs --all-targets -- -D warnings` → **exit 0**, no warnings (`Finished dev … in 1m 48s`).
  - `cargo test -p zeroclaw-runtime skills::` → `test result: ok. **170 passed; 0 failed**` (includes the new inline/block tag parse + the strip-on-save regression tests).
  - `cd web && npm ci && npm run build` → **exit 0** (vite 8 / rolldown).
- **Beyond CI — what did you manually verify?** New `document.rs` tests cover inline + block tag parsing and the strip-on-save regression (parse → serialize → parse preserves `[slash, open-skills]`). The drift test (`prop_fields_matches_struct`) bumped 6 → 7. Tag round-trip uses the same inline shape the loader's `parse_simple_frontmatter` reads, so #7490's `discord_slash_specs_from_skills` (filters `tags.contains("slash")`) sees tags written by the editor.
- **If any command was intentionally skipped, why:** full-workspace `cargo test` scoped to `zeroclaw-runtime skills::` (the touched module); CI runs the full battery.

- New permissions, capabilities, or file system access scope? **No** — uses the existing skill read/write service + auth.
- PII / real identities in diff? **No.**
- **Exposure note:** the `slash` toggle opts a skill into being a guild-visible Discord command — but the *authorization* of those commands is #7490's (same allowlists as inbound messages). This PR only writes the opt-in tag; the gate lives downstream. `open-skills` skills are deliberately not slash-toggleable here (and #7490 excludes them).

## Compatibility (required)

- Backward compatible? **Yes** — `tags` is optional/additive; the serializer omits empty tags so existing tagless SKILL.md files are byte-identical; the API field is `#[serde(default)]`.
- Config / env / CLI surface changed? **No** breaking change (CLI `skills add` scaffolds tagless as before).
- Upgrade steps: none.

## Rollback (required for `risk: low`)

`git revert` — single feature, no state/schema migration. Reverting restores the prior (tag-blind) editor; the strip-on-save behavior returns until re-applied.


WareWolf-MoonWall (approved):
The strip-on-save fix is correct and the round-trip regression test pins exactly the right invariant: `parse → serialize → parse` preserves `[slash, open-skills]`. The web UI additions (tag badges, slash-command toggle) are additive and gated behind the existing `slash_commands` config flag. Approving.

### 🟢 What looks good — SSOT for tags

`SkillFrontmatter.tags: Vec<String>` is the source of truth, created here. It doesn't duplicate anything in `SkillDocument` or the wider config — the field is parsed from and serialized back to the SKILL.md frontmatter on demand. The `#[serde(default)]` + empty-omit-on-serialize pattern means tagless files stay byte-identical across a round-trip.

### 🟢 What looks good — Backward compatibility

`#[serde(default)]` on `tags` and the `prop_fields_matches_struct` drift test being bumped from 6 to 7 are both correct hygiene. Existing skill files with no `tags:` line will deserialize cleanly, and the drift test would catch a future field addition that doesn't update the editor surface.

### 🟡 Warning — Labels reflect the pre-merge #7490 scope

The current GitHub labels (`channel`, `channel:discord`, `config`, `daemon`, `runtime`, `skills`) were set when the branch included the Discord slash-command work from #7490. Now that #7490 has merged (2026-06-15), a rebase against master will shrink the diff to roughly 2 commits and the auto-labeler will update accordingly. The PR body's stated target labels (`enhancement, skills, web, risk: low, size: M`) look correct for the true delta — expect them to settle after rebase.

### 🔵 Suggestion — Rebase against master before merge

After the #7490 merge the diff is smaller and the label churn resolves automatically. A fresh `git rebase upstream/master` here will make the review history cleaner and give CI a final green run on the true delta. Not blocking merge, but worth doing before squash.

WareWolf-MoonWall:
@singlerider — milestone alignment needed: this feat PR (skill frontmatter tags + slash-command toggle, approved by @WareWolf-MoonWall, pending @Audacity88/@JordanTheJet review) does not have a milestone set. Please advise on placement — likely v0.8.1.

JordanTheJet (commented):
I read the PR body, @WareWolf-MoonWall's approving review, and the full diff, and I cross-checked the skill-tags layer against the source. @WareWolf-MoonWall has already approved on this head and called the SSOT and backward-compat story correctly — I independently confirmed both and won't re-raise them. I'm adding a second set of eyes on the round-trip (it holds end-to-end, including the web layer) and sharpening one merge-readiness point: because the branch is conflicting, CI has not actually run the full battery on this delta yet.

### ✅ Resolved — round-trip and SSOT confirmed independently

@WareWolf-MoonWall's two 🟢 points hold up against the code. `SkillFrontmatter.tags: Vec<String>` is the single source of truth (crates/zeroclaw-runtime/src/skills/frontmatter.rs:33), with `#[serde(default, skip_serializing_if = "Vec::is_empty")]` so tagless files stay byte-identical, and the `prop_fields` drift test is correctly bumped 6 → 7. On the parse/serialize side (crates/zeroclaw-runtime/src/skills/document.rs), `write_tags` emits the inline `tags: [a, b]` shape (omitted when empty), and the parser handles both inline-flow and `- item` block lists. The three tests are the right ones — behavior-focused, and `editing_preserves_tags` pins the exact `parse → serialize → parse` invariant the strip-on-save bug violated. Nothing to re-litigate here; flagging it as registered.

### 🟢 What looks good — the strip-on-save fix holds in the UI layer too

I checked the editor because that's where tags could silently drop again. `TagsField` (web/src/components/sections/SkillsBundleEditor.tsx) gets it right: `setSlash`, `removeTag`, and `addTag` all operate on the *full* `tags` array, while `editableTags` is display-only and `open-skills` is rendered read-only. So `onTagsChange` always carries the complete list — the special `slash`/`open-skills` tags are preserved through every edit, not just on the backend round-trip. The fix is closed on both layers, which is what makes it actually safe to edit a skill that the loader manages.

### 🟡 Warning — not merge-ready yet: conflicting + CI hasn't validated this delta

The branch is `CONFLICTING` against `master`, and the only check that has run is `Apply path labels` — clippy, test, and build have *not* executed on this code in CI. @WareWolf-MoonWall flagged the rebase as a non-blocking 🔵; I'd put it one notch higher: it's a merge gate, not just hygiene, because right now nothing in CI has exercised the delta. Now that #7490 merged (2026-06-15), a `git rebase upstream/master` will both resolve the conflict and collapse the diff to the actual skill-tags surface — the large discord.rs / `schema.rs` `slash_commands` / test-struct hunks currently showing are #7490 residue and will disappear, which also lets the auto-labeler settle the labels to the body's stated set (`enhancement, skills, web, risk: low, size: M`). Please rebase and let the required checks go green before squash.

### 🔵 Suggestion — tag values with commas/brackets

Minor edge note, not a blocker: the inline serializer (`tags: [a, b]`) and the comma-splitting parser mean a tag containing a comma or bracket would not survive a round-trip cleanly. This matches the loader's `parse_simple_frontmatter` contract exactly, so the consistency is deliberate and correct — just worth a one-line doc note on `write_tags` so a future reader knows the simple-list shape is intentional rather than an oversight. The `addTag` lowercasing is good defensive behavior (keeps `slash`/`open-skills` matching downstream).

Net: the reviewable delta is correct on both the backend and the UI, and @WareWolf-MoonWall's approval stands. Treat the rebase + green required checks as the gate before merge rather than optional polish.

Audacity88 (commented):
Review context: I checked PR #7734 at current head 7fa49ee, including the live PR state, both prior reviews, the two actual skill-tags/editor commits, current `master`, the current `SkillsBundleEditor` surface, and visible GitHub checks. I also did a live skill-bundle editor UI pass. I am treating the UI notes below as non-blocking UX findings.

@Nillth this is a status comment rather than an approval from me: the feature slice looks sound, but the current head still needs a rebase and real CI before merge.

### ✅ Resolved — the tag round-trip uses the right source of truth

I agree with @WareWolf-MoonWall and @JordanTheJet on the core fix. `SkillFrontmatter.tags` is the right single source for the editor/RPC shape, and `SkillDocument` now round-trips both inline `tags: [slash, ops]` and block-list tags. The `editing_preserves_tags` regression test pins the important failure mode: parse, serialize, and parse again keeps `[slash, open-skills]` instead of silently stripping it.

### 🟢 What looks good — the UI preserves special tags while editing

The tag editor also preserves the right invariant. The slash toggle, add-tag path, and remove-tag path all update the full `tags` array, while `editableTags` is only the display subset. That means `open-skills` can stay loader-managed/read-only and still survive a normal edit/save cycle, and `slash` can be toggled without dropping unrelated tags.

### 🔵 Suggestion — a few skill-editor UX gaps from live testing

None of these are blockers for this PR's core tag-preservation claim, but the live path still has some rough edges worth tracking while this surface is being touched:

- Saving a skill does not give the same clear success feedback as the normal config field editor. The dirty state clears and the button darkens, but there is no green "saved" confirmation, so it is easy to wonder whether the write actually happened.
image

- The create-skill form shows `my-skill` and a sample description as if they are usable defaults, but they are placeholders: pressing Create without typing still errors with "Name is required." Either prefill real defaults or make the placeholder/example nature clearer.
image

- The dedicated Skills page is still view-only, even though it is the natural place users look for skill management. At minimum, each skill card should deep-link to the bundle editor/config surface; longer term, creation/editing may belong directly on the Skills page rather than only under Config.
image

### 🟡 Warning — current head needs rebase plus full CI before merge

The current branch is `DIRTY` against `master`, with merge conflicts present, and the only visible check on this head is `Apply path labels`; the clippy/test/build jobs have not run on the final delta. Since #7490 has already merged and current `master` has also moved in the same skills-editor surface, the branch needs to be rebased before the actual mergeable patch is knowable.

One related metadata note: the PR body currently lists `enhancement`, `skills`, `web`, `gateway`, `risk: low`, `size: M`, while GitHub currently applies `channel`, `config`, `daemon`, `runtime`, `skills`, `channel:discord`. That looks like the same #7490 residue; after rebase, please sync the body snapshot to the final labels.

Please rebase onto current `master`, keep the true skill-tags/editor changes, adapt the tag controls to the current localized `SkillsBundleEditor` surface, and let the required checks run. After that, the approval path should be straightforward from my side.

singlerider:
Maintainer merge-resolution pass (no code changes to the skill-tags surface).

Merged current `master` into the branch and resolved the single conflict in `crates/zeroclaw-channels/src/orchestrator/mod.rs`: the Discord builder now chains both `.with_intents_mask(dc.intents_mask)` and `.with_reaction_notifications(dc.reaction_notifications)` (master's #7524 reaction work, union'd with the slash-commands base), and the skill-derived slash-command resolver block is preserved intact.

With #7490 already merged (8517a1ec), the rebase collapsed the diff to the true skill-tags surface — `frontmatter.rs`, `document.rs`, `src/skills/mod.rs`, `SkillsBundleEditor.tsx`, `api.ts` — so the #7490 residue (discord.rs / orchestrator / schema.rs `slash_commands` / cli.ftl locales / daemon) is gone, exactly as @WareWolf-MoonWall, @JordanTheJet, and @Audacity88 anticipated. That clears the standing rebase warning all three raised.

Validation on the merged head d1fab703:

- `cargo clippy -p zeroclaw-channels -p zeroclaw-runtime -p zeroclaw-config --all-targets -- -D warnings` → no warnings
- `cargo test -p zeroclaw-runtime skills::` → 170 passed, 0 failed (incl. the `editing_preserves_tags` strip-on-save regression)
- `cd web && npm run build` → clean

Labels synced to the true delta: `enhancement, skills, web, risk: low, size: S`.

The branch is `MERGEABLE` and CI is rerunning the full battery on the real delta. @WareWolf-MoonWall's approval stands on unchanged code.

PR #7745, opened 2026-06-15
fix(providers): retry email OAuth refresh failures

### LINKED ISSUE #7739: bug(channels/email): retry transient OAuth refresh failures
## Affected component

channel

## Severity

S2 - degraded behavior

## Current behavior

The email OAuth2 refresh path calls `email_oauth2::refresh_access_token()` directly. Other provider refresh paths, such as OpenAI and Gemini, use retry/backoff wrappers for transient failures.

## Expected behavior

Email OAuth refresh should use the same retry/backoff discipline as the other token refresh paths when the error is transient. The fix should keep permanent auth failures explicit and should not hide invalid credentials.

## Steps to reproduce

1. Inspect the email OAuth refresh path on current `master`.
2. Compare it with the OpenAI and Gemini refresh paths.
3. Simulate a transient refresh failure.
4. Observe that email refresh does not use the same retry/backoff wrapper.

## Impact

Affected users: email channel users with OAuth-backed accounts.

Frequency: intermittent network/provider failures.

Consequence: transient refresh failures can interrupt email channel operation when other auth paths would retry.

## Logs / stack traces

None.

## ZeroClaw version

## Rust version

## Operating system

## Regression?

Unknown.

## Pre-flight checks

## Related

- Follow-up from the 2026-06-15 maintainer sweep.

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Route expired email OAuth2 access tokens through a retrying refresh helper instead of calling the refresh endpoint once.
  - Keep permanent OAuth credential failures explicit with an auth-local classifier for `invalid_grant`, `invalid_client`, and non-retryable 4xx responses.
  - Retry transient refresh failures such as provider/network errors, 408, 429, and 5xx responses before surfacing the final error.
  - Add regression coverage for transient retry, permanent failure fail-fast behavior, and OAuth refresh error classification.
- **Scope boundary:** This PR only changes the email OAuth2 refresh path. It does not change the email device-code login flow, email config schema, OpenAI/Gemini OAuth refresh behavior, or non-email channels.
- **Blast radius:** Email channel accounts that use OAuth-backed credentials and need to refresh an expired access token. The auth profile store update path is touched only when a refresh succeeds.
- **Linked issue(s):** Closes #7739.
- **Labels:** `bug`, `risk: medium`, `size: M`, `channel`, `provider`, `channel:email`.

```text
$ cargo test -p zeroclaw-providers email_oauth_refresh -- --nocapture
running 3 tests
test auth::tests::email_oauth_refresh_classifier_keeps_permanent_and_transient_errors_separate ... ok
test auth::tests::email_oauth_refresh_does_not_retry_permanent_failure ... ok
test auth::tests::email_oauth_refresh_retries_transient_failure ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 964 filtered out; finished in 2.16s

$ rustfmt --edition 2024 crates/zeroclaw-providers/src/auth/mod.rs --check
passed

$ git diff --check
passed
```

- **Beyond CI — what did you manually verify?** The new tests exercise the `get_valid_email_oauth2_token()` path against a test token endpoint, proving a transient refresh failure is retried and an `invalid_grant` response is not retried. The classifier test covers permanent OAuth errors, retryable rate-limit/temporary-unavailable responses, and transport-style failures.
- **If any command was intentionally skipped, why:** Workspace-shaped clippy and a CI-shaped full test command were not run locally. GitHub Actions is green for this draft, including Lint, Test, Security, platform builds, Benchmarks Compile, and CI Required Gate.

- New external network calls? `No`. The existing configured email OAuth token endpoint may receive retry attempts after transient refresh failures.
- Secrets / tokens / credentials handling changed? `Yes`. The refresh-token path now retries transient refresh failures, keeps invalid credentials non-retryable, preserves the prior refresh token when the provider omits a replacement, and only writes refreshed tokens after a successful response.
- If any `Yes`, describe the risk and mitigation: Retrying token refresh can repeat a request that carries a refresh token to the same configured OAuth token endpoint. The retry helper caps attempts, treats invalid credentials as non-retryable, and reuses the existing backoff guard for retryable failures.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No operator action required.

- **Observable failure symptoms:** Email OAuth-backed accounts log repeated `Email OAuth2 token refresh failed` warnings, fail refresh after the capped retry attempts, or return `Email OAuth2 token refresh is in backoff` while the short refresh backoff is active.


singlerider (approved):
Read against current `master`: the email refresh path in `crates/zeroclaw-providers/src/auth/mod.rs`, the new `refresh_email_access_token_with_retries` / `is_non_retryable_oauth_refresh_error`, and the shared backoff helpers it reuses. CI Test/Lint/Format are green. No active blocks.

### 🟢 What looks good: retry classification is the right shape

The risk in any refresh-retry is the retryable vs non-retryable split, and this one is correct:

- `temporarily_unavailable` and `server_error` are checked first and force retryable, so they win over any later 4xx/5xx text heuristic. That matches RFC 6749 §5.2 transient semantics.
- The permanent OAuth error codes (`invalid_grant`, `invalid_client`, `unauthorized_client`, `unsupported_grant_type`, `invalid_scope`, `access_denied`) are correctly non-retryable; retrying an `invalid_grant` just burns the provider's rate budget on a token that will never succeed.
- The reqwest status downcast treats 4xx as non-retryable except 429 and 408, which is exactly right (rate-limit and request-timeout are transient). The string-scan numeric fallback applies the same 429/408 exception and only runs after the structured status check, so it is a genuine last resort; a false positive there fails fast rather than hammering, which is the safe direction for a refresh.

The cross-call backoff is sound: `set_refresh_backoff` only on retryable exhaustion, `clear_refresh_backoff` on success, and an early `refresh_backoff_remaining` bail so a profile in backoff does not re-attempt tightly. Permanent failures skip the backoff and return immediately, which is correct since they would fail identically next time. The constants and backoff helpers are the same ones the Gemini/OpenAI refresh paths already use, so this is following proven precedent rather than inventing a new mechanism. Both `record!` sites are wrapped in `Event::new(module_path!(), ...)`, so the retry/exhaustion logs stay attributed. The tests exercise the real `AuthService` flow end to end: one 500 then success asserts recovery at exactly 2 attempts, and the permanent-failure case asserts no retry.

One cosmetic note, not blocking: `OPENAI_REFRESH_FAILURE_BACKOFF_SECS` reused on the email path reads oddly given the name, but it is a shared constant and the behavior is correct.

singlerider:

PR #7747, opened 2026-06-15
fix(runtime): wire mcp_bundles into agent loop — enforce per-agent MCP server scoping

### LINKED ISSUE #7733: [Bug]: mcp_bundles is parsed and shown in Config but never enforced at runtime — per-agent MCP scoping is a silent no-op
### Affected component

tools

### Severity

S2 - degraded behavior (silent no-op of a security-relevant isolation field; see Impact — maintainers may want to bump)

### Current behavior

`[mcp_bundles.<alias>]` and the per-agent `mcp_bundles = [...]` field parse cleanly and are surfaced in the Config menu, but the runtime never reads them. MCP servers are connected globally for **every** agent and tools are gated only by the tool-name policy (`allowed_tools`/`excluded_tools`), so selecting a bundle on an agent has zero effect.

- Config surface exists: `McpBundleConfig { servers, exclude }` at `crates/zeroclaw-config/src/schema.rs:10257`; top-level map at `schema.rs:449`; per-agent field `mcp_bundles: Vec<String>` at `schema.rs:3284`; Config menu via `crates/zeroclaw-config/src/sections.rs` (`Section::McpBundles`).
- Runtime never consumes it: `git grep mcp_bundles -- crates/zeroclaw-runtime/` returns **nothing**.
- All servers connect for every agent: `crates/zeroclaw-runtime/src/agent/loop_.rs:1032` (and `:2526`) call `McpRegistry::connect_all(&config.mcp.servers)`. Per-tool gating is only `ToolAccessPolicy` via `register_eager_mcp_tool_if_allowed` (`loop_.rs:232`) / `eager_mcp_tool_allowed` — no per-agent server/bundle filter.
- Tool-name policy is exact-match, no wildcards (`crates/zeroclaw-config/src/policy.rs:267`).

### Expected behavior

An agent's `mcp_bundles` should scope **which MCP servers' tools are registered for that agent** (include/exclude by server name), mirroring how `skill_bundles` scopes skills per agent (`skills::load_skills_for_agent`). Or, at minimum, the runtime should warn/validate that `mcp_bundles` is currently unenforced instead of silently ignoring it.

### Steps to reproduce

```bash
# config.toml: two servers, agent selects only one via a bundle
# [[mcp.servers]] name = "serverA"  ...
# [[mcp.servers]] name = "serverB"  ...
# [mcp_bundles.onlyA] servers = ["serverA"]
# [agents.scoped] mcp_bundles = ["onlyA"]   # expect: only serverA tools

zeroclaw agent -a scoped -m "ping" --verbose --log-level debug

# Observed: BOTH serverA and serverB connect and their tools are
# available to `scoped`; the mcp_bundles selection is ignored.
# Code-level proof: `git grep mcp_bundles -- crates/zeroclaw-runtime/` is empty.
```

### Impact

Operators who set `mcp_bundles` on an agent get **no isolation** — silently. This is security-relevant: one may believe a sensitive server is scoped away from a public-/group-facing agent when it is not. The only working per-agent isolation today is enumerating every server tool in `excluded_tools`/`allowed_tools` (exact match, no wildcard) — e.g. a 35-tool server becomes a 35-entry denylist duplicated per agent.

### Logs / stack traces

```text
Initializing MCP client — 2 server(s) configured   (loop_.rs:1024)
MCP server `serverA` connected — 4 tool(s) available
MCP server `serverB` connected — 35 tool(s) available
# both connect for the agent regardless of its mcp_bundles selection
```

### ZeroClaw version

0.8.0 (master 75195f2faf)

### Rust version

rustc 1.95.0

### Operating system

Windows 11

### Regression?

No, first-time setup (mcp_bundles has never been wired — schema-only since it was introduced)

### Pre-flight checks

### Relation

Distinct from #6699 (`tool_filter_groups` prefix bug — per-turn context shaping, not per-agent server scoping). Not covered by tracker #7320 (v0.8.3 MCP dashboard).

Yyukan:
## Summary

- **Base branch:** `master`
- **What changed and why:** `mcp_bundles` config has been schema-only since it was introduced — the runtime calls `McpRegistry::connect_all(&config.mcp.servers)` unconditionally at both agent loop call sites, giving every agent every MCP server. This PR wires `mcp_bundles` into the runtime so agents only connect to the servers their bundle config admits.
- **Scope boundary:** No changes to `McpBundleConfig` / `AgentConfig` schema, MCP connection mechanics, tool-name policy (`allowed_tools` / `excluded_tools`), or any channel/provider/gateway code.
- **Blast radius:** Only agents with `mcp_bundles` explicitly set are affected. Agents with empty `mcp_bundles` fall back to all servers — identical to current behaviour.
- **Linked issue(s):** Closes #7733
- **Labels:** `bug`, `size: S`, `risk: high`, `agent`, `runtime`, `domain:security`, `priority:p1`

  ```bash
  cargo fmt --check -p zeroclaw-runtime   # clean, no output
  cargo clippy -p zeroclaw-runtime -- -D warnings  # Finished dev profile, 0 errors
  ```

  ```
  test result: ok. 2209 passed; 0 failed; 1 ignored; finished in 5.72s
  test result: ok. 1 passed; 0 failed; 0 ignored; finished in 0.15s
  ```

  `cargo doc --no-deps --workspace --exclude zeroclaw-desktop` produced 3 warnings — all pre-existing (`MemoryRecall` in zeroclaw-api, `AliasSource` in zeroclaw-config, `<think>` tag in loop_.rs test fixtures). Zero warnings introduced by this PR.

- **Beyond CI — what did you manually verify?** Confirmed both `connect_all` call sites in `loop_.rs` pass the filtered `agent_mcp_servers` slice, not `&config.mcp.servers`. `git grep mcp_bundles -- crates/zeroclaw-runtime/` now returns the new helper and both wired call sites (was empty before this PR). Log line updated at both sites to report the filtered server count.

This PR *reduces* the attack surface — agents with `mcp_bundles` configured now receive only their scoped server set instead of silently getting all servers.

## Compatibility (required)

- Backward compatible? `Yes` — empty `mcp_bundles` returns all servers, matching current behaviour exactly. Schema unchanged, no migration needed.

- **Observable failure symptoms:** Agent tools from a non-bundled server become unavailable; log line "Initializing MCP client — N server(s) configured" reflects the filtered count rather than the total.


singlerider (approved):
First review (singlerider). Read the PR body, the full diff, and traced both production call sites. CI is 17/17 green, milestone v0.8.1 is set, no prior reviews. This closes a real privilege-scoping gap: `mcp_bundles` was schema-only and every agent got every MCP server. Approving.

### 🟢 What looks good: both call sites actually route through the new helper

The gap was that `loop_.rs` called `McpRegistry::connect_all(&config.mcp.servers)` unconditionally at both the main turn loop (`run`) and the persistent-message loop (`process_message`). The diff replaces both with `resolve_mcp_servers_for_agent(&config, agent_alias)` and updates the count in the log line at both sites, so the scoping cannot be true at one entry point and bypassed at the other. The in-file comment calling out that both sites MUST route through the helper (and that the unit tests would not catch a revert of either site) is exactly the right place to leave that warning, since the tests exercise the helper directly.

### 🟢 What looks good: the resolution semantics are conservative and fully covered

Union the `servers` of each named bundle, subtract every `exclude`, then intersect against the actually-configured servers. Exclude wins over include, a bundle server absent from `mcp.servers` is silently dropped rather than fabricated, and an unknown bundle alias logs an attributed WARN and contributes nothing. The two backward-compat escape hatches (unknown agent and empty `mcp_bundles` both return all servers) keep existing deployments byte-identical, which matches the stated blast radius. The 8 unit tests pin every one of these branches: unknown agent, empty bundles, unknown alias, single bundle, exclude-removes, multi-bundle union-minus-exclude, phantom server absent, and exclude-takes-priority.

### 🟢 What looks good: net attack-surface reduction, clean rollback

This only ever narrows what a configured-bundle agent can reach, never widens it, so the security direction is correct for a `domain:security` / `priority:p1` fix. The schema is untouched, no migration is needed, and `git revert` is a complete rollback. The helper mirrors the existing `load_skills_for_agent_from_config` shape (takes the alias, resolves the agent internally) so the call sites stay clean.

singlerider:

Audacity88 (changes requested):
I reviewed the live PR state, #7733, singlerider’s prior approval, current checks, the full three-file diff, public source excerpts from `agent.rs` and `ws.rs`, and the current MCP initialization paths. singlerider correctly verified that the two `loop_.rs` call sites now use the filtered server list. Those fixes remain sound. I’m requesting changes because a third production path in `Agent::from_config` still connects every configured MCP server.

### 🟢 What looks good — loop_.rs now uses the filtered server list

The CLI/main turn loop and persistent-message loop now pass the resolved server slice into `McpRegistry::connect_all`, and the log count reflects the filtered list. That is the right shape for the two paths this PR already touched.

### 🟢 What looks good — the shared resolver has the right contract

The resolver semantics themselves look conservative: empty `mcp_bundles` preserves the old all-server behavior, configured bundles union their server lists, excludes win over includes, unknown bundle aliases warn and contribute nothing, and phantom server names are not fabricated. Those are the right defaults for compatibility and for narrowing access.

### 🔴 Blocking — Agent::from_config still bypasses mcp_bundles

The new resolver is wired into `crates/zeroclaw-runtime/src/agent/loop_.rs`, but `crates/zeroclaw-runtime/src/agent/agent.rs` still checks `config.mcp.servers` and calls `McpRegistry::connect_all(&config.mcp.servers)` inside `from_config_with_session_cwd_and_mcp_approval_mode`. That constructor is not test-only: the dashboard WebSocket path builds agents through `Agent::from_config_with_session_cwd_and_mcp_backchannel(..., true, ...)`, and the default `Agent::from_config` path also initializes MCP by default.

That means an agent with `mcp_bundles` set can still get every MCP server when it enters through the dashboard/WS-style `Agent::from_config` path, so this does not fully close #7733 yet. Please route that constructor through `resolve_mcp_servers_for_agent`, update the log count to use the filtered list, and add a regression that constructs an agent through the `from_config`/dashboard path with a configured bundle and would fail if that path goes back to connecting `config.mcp.servers` directly.

Yyukan:
Thanks for catching this @Audacity88 — you're right, the `from_config` path was missed.

**What changed:**

- `agent.rs:from_config_with_session_cwd_and_mcp_approval_mode` now routes through `resolve_mcp_servers_for_agent` (same pattern as the two `loop_.rs` sites)
- Protective comment in `mcp_bundle.rs` updated to name all **three** mandatory call sites explicitly
- Added regression test `from_config_path_agent_with_bundle_excludes_sensitive_server` — constructs the same config shape `from_config` would pass, asserts the resolver correctly filters a `sensitive` server out of an agent scoped to a `safe_only` bundle

Audacity88 (changes requested):
I reviewed the live PR state, #7733, singlerider’s prior approval, my earlier request-changes review, the author’s follow-up comment, current checks, the current four-file diff, and the current MCP initialization paths at head 2ce924cfaf12ea170e3c4c257b42bc0d0663e4a9. The functional `Agent::from_config` bypass I blocked on is fixed now, but I’m keeping this at request-changes because the new test intended as the `from_config` regression still exercises only the helper, not the production constructor path that was missed.

### ✅ Resolved — Agent::from_config now uses the filtered server list

The third production path now calls `tools::resolve_mcp_servers_for_agent(config, agent_alias)`, checks the filtered list for emptiness, logs the filtered count, and passes that filtered slice into `McpRegistry::connect_all`. That addresses the functional bypass from my prior review: the default `Agent::from_config` path and the ACP/WS-style backchannel constructors now share the same scoped-server behavior as the two `loop_.rs` paths.

### 🟢 What looks good — the shared resolver keeps the security direction narrow

The helper keeps the backward-compatible empty-bundle behavior, unions configured bundle server names, lets `exclude` win over `servers`, drops phantom server names instead of fabricating config, and warns when an agent names an unknown bundle. That is the right shape for narrowing MCP access without changing the schema or surprising existing agents that have no bundle configured.

### 🔴 Blocking — the added from_config test would not catch a wiring revert

The new `from_config_path_agent_with_bundle_excludes_sensitive_server` test lives in `crates/zeroclaw-runtime/src/tools/mcp_bundle.rs` and calls `resolve_mcp_servers_for_agent` directly. The comment above it also says a revert of the `agent.rs` wiring would make the test “vacuously pass” because the production site would go back to calling `connect_all` with every configured server while the helper test stayed green.

That leaves this security fix without a regression for the exact path that was missed once already. Please add coverage that fails if `from_config_with_session_cwd_and_mcp_approval_mode` stops routing through the filtered server list. That could be a constructor-level test with a lightweight MCP/connect seam, or a narrow static/call-site guard that proves all three production MCP initialization blocks call `resolve_mcp_servers_for_agent` and never pass `&config.mcp.servers` directly. The important property is simple: if `agent.rs` reverts to the old full-server call, a test must turn red.

singlerider (approved):
Re-reviewed at head `71e4c79e56`. Pulled the branch, ran the `mcp_bundle` suite and clippy `-D warnings`, and checked Audacity88's outstanding blocker specifically.

### ✅ RESOLVED — the call-site regression that Audacity88 blocked on is now covered

His last blocker was that the `from_config` regression test exercised only the helper, so a revert of the `agent.rs` wiring back to `connect_all(&config.mcp.servers)` would pass vacuously. The new `all_call_sites_route_through_resolver_no_direct_config_mcp_servers` test is exactly the static call-site guard he offered as an acceptable option: it greps `loop_.rs` and `agent.rs` for the literal `connect_all(&config.mcp.servers)`, filters comment lines, and panics if any production occurrence exists. I confirmed the grep finds zero production occurrences today and the test passes; if any of the three init sites reverts to the unscoped call, this turns red. That gives the security fix the regression anchor for the exact path that was missed before.

### 🟢 Resolver shape and security direction (unchanged, still correct)

`resolve_mcp_servers_for_agent` keeps the backward-compatible empty-bundle behavior, unions configured bundle server names, lets `exclude` win over `servers`, drops phantom names rather than fabricating config, and warns on unknown bundles. All three production MCP init paths now route through it.

### 🟢 Validation

Local: `cargo test -p zeroclaw-runtime --lib mcp_bundle` 10/10 pass (including the new call-site guard), `clippy --all-targets -D warnings` clean.

Audacity88's blocker is addressed and the implementation he praised is unchanged. Approving.

Yyukan:
Closing as duplicate of #8120.

This PR implemented per-agent MCP server scoping via `tools::resolve_mcp_servers_for_agent()` to fix issue #7733. While this PR was open, #8120 was merged (2026-06-22) with an alternative implementation using `Config::mcp_servers_for_agent()` that solves the same problem.

Both approaches:
- Resolve MCP servers based on agent's `mcp_bundles` configuration
- Apply include/exclude logic from bundle definitions
- Maintain backward compatibility (empty bundles → all servers)
- Cover all three call sites (agent.rs, loop_.rs main + PM loops)

The functionality is now in master via #8120, making this PR redundant.

Yyukan:
Superseded by #8120

PR #7676, opened 2026-06-15
chore(docs): make gh-pages ephemeral to bound clone size

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - gh-pages accumulated ~2.9 GiB of generated mdBook artifacts across its commit history, dragging every default clone to ~163 MiB. Three layers compounded: dead commit history, redundant version copies in the tip, and per-locale search indexes.
  - Treat gh-pages as **ephemeral**: each deploy force-pushes a single orphan commit instead of appending, so no superseded artifact stays reachable through history.
  - Add `prune-versions` (`DOCS_KEEP_VERSIONS`, default 3): retain master and the newest N final releases; drop pre-releases and older finals every deploy.
  - Replace the duplicate `stable/` tree with a **pointer**. `bump-version.sh` writes `docs/book/stable-version.txt`; `gen-versions` and the new `gen-root-index` resolve "Stable (latest release)" and the root redirect to the real version dir. The release tag deploy is what builds and publishes a version's docs dir; landing a pointer change on master refreshes the stable metadata only (root redirect and selector entry resolve to that release's already-published version dir, failing loudly if it is absent), it does not rebuild or republish the release tag's docs.
  - Build full-text search **only for the primary locale** (English); translated locales build without a search index or search box.
  - Net: a default clone drops from ~163 MiB to ~35 MiB (-79%), kept there automatically with no manual git cleanup.
- **Scope boundary:** Does not rewrite master history (never touched). Does not change docs content, themes, or the translation pipeline. Does not delete or force-push any branch as part of this PR; the gh-pages rewrite happens via the workflow on the next deploy.
- **Blast radius:** `docs-deploy.yml` workflow, the `mdbook` xtask binary, and the release `bump-version.sh` step. No runtime, gateway, provider, or channel code touched.
- **Linked issue(s):** none.
- **Labels:** `ci`, `docs`, `documentation`, `type: docs`, `scripts`, `risk: high`, `size: M` (auto + manual).

## Validation Evidence (required)

\`\`\`bash
cargo fmt --all -- --check
cargo clippy -p xtask --all-targets -- -D warnings
cargo test -p xtask
\`\`\`

- **Commands run and tail output:**
  - \`cargo fmt --all -- --check\` -> exit 0, no diff.
  - \`cargo clippy -p xtask --all-targets -- -D warnings\` -> \`Finished dev profile ... in 1m 17s\`, exit 0, zero warnings.
[...]

PR #7754, opened 2026-06-16
perf(docs): publish rustdoc once, dedupe per-locale assets, drop print pages

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The gh-pages site copied the workspace rustdoc into every published version tree (`<tag>/api`), duplicated shared assets per locale, and emitted a whole-book `print.html` per locale. A full clone carried hundreds of MB many times over.
  - This publishes rustdoc once at the site root (`/api`), drops rustdoc's redundant `api/src/` source-view, dedupes `mermaid.min.js` into `_shared/`, and disables mdBook print output.
  - The shared-chrome hash detector mishandled compound extensions (`mermaid-193ae996.min.js`), so mermaid was never deduped; fixed and unit-tested via a new `strip_chrome_hash` helper.
- **Scope boundary:** Does not rewrite existing gh-pages or git history; does not change docs prose, the version selector UI, or full-text search. Old version trees already deployed keep their nested `api/` until rebuilt.
- **Blast radius:** docs-deploy workflow (gh-pages sync), `docs/book/book.toml`, the rustdoc bridge page (`docs/book/src/api.md` now links site-absolute `/api/...`), and the xtask mdbook build/refs/versions paths. No runtime, channel, provider, or gateway code touched.
- **Linked issue(s):** Related #7676
- **Labels:** `type: docs`, `risk: medium`, `size: M`, `docs`

```bash
cargo fmt --all -- --check
cargo clippy -p xtask --all-targets -- -D warnings
cargo test -p xtask
./target/release/mdbook --tag master build
```

  - `cargo fmt --all -- --check`: clean, exit 0.
  - `cargo clippy -p xtask --all-targets -- -D warnings`: `Finished dev profile`, exit 0, no warnings.
  - `cargo test -p xtask`: `test result: ok. 79 passed; 0 failed` plus the asset binary suite `5 passed`. Includes 4 new `strip_chrome_hash` tests (single/compound extension, unhashed, non-hex).
  - `mdbook --tag master build`: `==> Internal link check passed`, build exit 0.
- **Beyond CI, what I manually verified (real build of master):**
  - rustdoc lands once at `book/api` (212 MB); `book/master/api` no longer exists.
  - `print.html` count after build: 0.
  - Per-version tree dropped from 311 MB to 49 MB.
  - `book/master/en/api.html` links resolve to site-absolute `/api/zeroclaw/...`.
  - 0 per-locale `mermaid-*.min.js` copies remain; single `_shared/mermaid.min.js`, refs rewritten depth-aware to the shared copy.
  - 3-version site math: ~933 MB to ~360 MB (one shared 212 MB api + three 49 MB trees).
  - **Not verified locally:** the gh-pages force-push/prune path runs only on GitHub Actions; the workflow `apply_run_slice` api sync mirrors the existing `_shared` ownership rule and is exercised on first deploy.
- **If any command was intentionally skipped, why:** Full workspace `cargo test`/`clippy --all-targets` were scoped to `-p xtask`; the diff only touches xtask + docs + workflow, and the tmpfs build quota on this host cannot hold a full workspace debug + release build simultaneously. CI runs the full battery.

## Compatibility (required)

- Backward compatible? `Yes` for readers: both old (`<tag>/api`) and new (`/api`) link forms resolve during the transition. Print-to-PDF via the browser print page is removed.
- Upgrade steps: none. First master deploy after merge writes the root `/api`; previously deployed tags keep working and shed their nested `api/` only when rebuilt.

## Rollback (required for risk: medium)

- **Fast rollback command/path:** `git revert <sha>` and redeploy docs; the next master deploy restores per-version `api/` and print pages.
- **Feature flags or config toggles:** `[output.html.print] enable` in `docs/book/book.toml` re-enables print pages.
- **Observable failure symptoms:** Broken API reference links on the docs site (`/api/...` 404), or the version selector losing the API link. Check the docs-deploy workflow log for the `api/` rsync step and confirm `/api/index.html` exists at the gh-pages root.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -267,6 +267,17 @@ jobs:
             mkdir -p "$TAG"
             rsync -a --delete "$TAG_SRC/" "$TAG/"
 
+            # Rustdoc is shared site-wide at '/api', owned by master (same rule
+            # as _shared). master rewrites it each deploy; a tag only seeds it
+            # when bootstrapping the first deploy, never regressing newer docs.
+            API_SRC="${GITHUB_WORKSPACE}/docs/book/book/api"
+            if [ -d "$API_SRC" ]; then
+              if [ "$TAG" = "master" ] || [ ! -d "api" ]; then
+                mkdir -p api
+                rsync -a --delete "$API_SRC/" "api/"
+              fi
+            fi
+
             # No stable/ tree: Stable is a pointer to a real version dir, never a
             # duplicate copy. The committed pointer (docs/book/stable-version.txt)
             # is published to the gh-pages root as stable-version.txt; gen-versions,
--- a/docs/book/book.toml
+++ b/docs/book/book.toml
@@ -25,6 +25,9 @@ enable = true
 limit-results = 30
 use-boolean-and = true
 
+[output.html.print]
+enable = false
+
 [output.html.fold]
 enable = true
 level = 0
--- a/docs/book/src/api.md
+++ b/docs/book/src/api.md
@@ -2,7 +2,7 @@
 
 Full rustdoc for every public type in the workspace, auto-generated from the `///` comments on each type, function, and module. Use this when you need to know the exact shape of a struct, the methods on a trait, or what a function returns: anything the generated reference exposes better than prose can.
 
-**[Open the rustdoc →](../api/zeroclaw/index.html)**
+**[Open the rustdoc →](/api/zeroclaw/index.html)**
 
 ## How to navigate it
 
@@ -15,18 +15,18 @@ Full rustdoc for every public type in the workspace, auto-generated from the `//
 
 | Crate | What it exposes |
 |---|---|
-| [`zeroclaw`](../api/zeroclaw/index.html) | Top-level umbrella with re-exports |
-| [`zeroclaw-api`](../api/zeroclaw_api/index.html) | Public traits: `Provider`, `Channel`, `Tool`, `StreamEvent` |
-| [`zeroclaw-config`](../api/zeroclaw_config/index.html) | Config schema, autonomy types, secrets |
-| [`zeroclaw-runtime`](../api/zeroclaw_runtime/index.html) | Agent loop, security, SOP, onboarding |
-| [`zeroclaw-providers`](../api/zeroclaw_providers/index.html) | Every LLM-provider implementation |
-| [`zeroclaw-channels`](../api/zeroclaw_channels/index.html) | Messaging integrations |
-| [`zeroclaw-gateway`](../api/zeroclaw_gateway/index.html) | HTTP/WebSocket gateway |
-| [`zeroclaw-tools`](../api/zeroclaw_tools/index.html) | Agent-callable tools |
-| [`zeroclaw-memory`](../api/zeroclaw_memory/index.html) | Conversation memory, embeddings |
-| [`zeroclaw-plugins`](../api/zeroclaw_plugins/index.html) | WASM plugin host |
-| [`zeroclaw-hardware`](../api/zeroclaw_hardware/index.html) | GPIO / I2C / SPI / USB |
-| [`zeroclaw-infra`](../api/zeroclaw_infra/index.html) | Tracing, metrics |
+| [`zeroclaw`](/api/zeroclaw/index.html) | Top-level umbrella with re-exports |
+| [`zeroclaw-api`](/api/zeroclaw_api/index.html) | Public traits: `Provider`, `Channel`, `Tool`, `StreamEvent` |
+| [`zeroclaw-config`](/api/zeroclaw_config/index.html) | Config schema, autonomy types, secrets |
+| [`zeroclaw-runtime`](/api/zeroclaw_runtime/index.html) | Agent loop, security, SOP, onboarding |
+| [`zeroclaw-providers`](/api/zeroclaw_providers/index.html) | Every LLM-provider implementation |
+| [`zeroclaw-channels`](/api/zeroclaw_channels/index.html) | Messaging integrations |
+| [`zeroclaw-gateway`](/api/zeroclaw_gateway/index.html) | HTTP/WebSocket gateway |
+| [`zeroclaw-tools`](/api/zeroclaw_tools/index.html) | Agent-callable tools |
+| [`zeroclaw-memory`](/api/zeroclaw_memory/index.html) | Conversation memory, embeddings |
+| [`zeroclaw-plugins`](/api/zeroclaw_plugins/index.html) | WASM plugin host |
+| [`zeroclaw-hardware`](/api/zeroclaw_hardware/index.html) | GPIO / I2C / SPI / USB |
+| [`zeroclaw-infra`](/api/zeroclaw_infra/index.html) | Tracing, metrics |
 
 See [Architecture → Crates](./architecture/crates.md) for a plain-English description of how the crates fit together.
 
--- a/xtask/src/cmd/mdbook/build.rs
+++ b/xtask/src/cmd/mdbook/build.rs
@@ -134,10 +134,51 @@ pub fn assemble(root: &std::path::Path, tag: Option<&str>) -> anyhow::Result<()>
 
 pub fn prune_rustdoc_source_view(api_dir: &Path) -> anyhow::Result<()> {
     let src_view = api_dir.join("src");
-    if src_view.exists() {
-        std::fs::remove_dir_all(&src_view)?;
+    if !src_view.exists() {
+        return Ok(());
     }
+    let mut stack = vec![api_dir.to_path_buf()];
+    while let Some(dir) = stack.pop() {
+        if dir == src_view {
+            continue;
+        }
+        let Ok(entries) = std::fs::read_dir(&dir) else {
+            continue;
+        };
+        for entry in entries.flatten() {
+            let path = entry.path();
+            match entry.file_type() {
+                Ok(ty) if ty.is_dir() => stack.push(path),
+                Ok(_) if path.extension().is_some_and(|e| e == "html") => {
+                    if let Ok(content) = std::fs::read_to_string(&path) {
+                        let stripped = strip_source_anchors(&content);
+                        if stripped != content {
+                            std::fs::write(&path, stripped)?;
+                        }
+                    }
+                }
+                _ => {}
+            }
+        }
+    }
+    std::fs::remove_dir_all(&src_view)?;
     Ok(())
+}
+
+pub fn strip_source_anchors(html: &str) -> String {
+    let needle = "<a class=\"src\"";
+    let mut out = String::with_capacity(html.len());
+    let mut rest = html;
+    while let Some(start) = rest.find(needle) {
+        let after = &rest[start..];
+        let Some(end) = after.find("</a>") else {
+            break;
+        };
+        out.push_str(&rest[..start]);
+        rest = &after[end + "</a>".len()..];
+    }
+    out.push_str(rest);
+    out
 }
 
 pub fn strip_chrome_hash(file_name: &str) -> Option<String> {
@@ -272,7 +313,7 @@ pub fn extract_shared_chrome(version_dir: &Path, shared_dir: &Path) -> anyhow::R
 
 #[cfg(test)]
 mod tests {
-    use super::strip_chrome_hash;
+    use super::{strip_chrome_hash, strip_source_anchors};
 
     #[test]
     fn strips_single_extension_hash() {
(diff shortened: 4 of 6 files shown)


Audacity88 (dismissed):
I checked the captured PR state, current diff, prior comments/reviews, visible CI, the PR's validation notes, the docs-deploy workflow, and the mdBook/rustdoc xtask paths. There are no prior review threads on this PR. I did not independently run a local rustdoc/docs build.

### 🟢 What looks good — root-owned API docs and shared Mermaid assets

The deploy-side ownership model is the right shape: `master` owns the shared root `/api` tree, while tag deploys only seed it when it does not already exist. The Mermaid hash handling fix is also well scoped; extracting the `strip_chrome_hash()` helper and covering compound extensions directly protects the specific missed `mermaid-193ae996.min.js` case.

### 🔴 Blocking — remove or validate rustdoc source links before pruning

The PR intentionally drops rustdoc's `api/src/` source-view tree, but the implementation only deletes that target directory after copying rustdoc. I do not see a postprocess step that removes or rewrites any generated per-item source links from the HTML, or a validation guard proving those links are absent or non-breaking. If rustdoc generated `src/...` links in the HTML, deleting the target tree alone would leave the published API pages pointing at missing paths.

This is easy to miss because the current internal link checker intentionally skips site-absolute and `/api` links, and the PR's manual verification checks the top-level `/api/zeroclaw/...` crate links, not the pruned rustdoc source-view links. Please verify whether the generated rustdoc HTML contains these source links; if it does, either keep `api/src`, add the HTML rewrite that removes/retargets the links, or add a postprocess/test that verifies no generated rustdoc page still points at a missing `src/...` target before pruning.

Nillth (commented):
I took stock of the existing state before writing: there are no inline diff threads and no top-level conversation, and the only formal review is @Audacity88's `CHANGES_REQUESTED` at 212240b6f6e05138df2e5f56f8ebfd0adc51d3b6 (the current HEAD), which raises one blocker about pruned rustdoc source links. That block is active and unaddressed (no author reply committing to a fix). I cross-checked the diff against the PR-HEAD sources via the contents API rather than my local clone (which is well behind master), read the docs-deploy workflow's `apply_run_slice`/prune path, `xtask/src/cmd/mdbook/{build,refs,versions,linkcheck}.rs`, the docs quality/link gates, FND-002, and FND-005. I also reproduced rustdoc's real output to test the source-link claim directly. CI is green across the Quality Gate (Format, Lint, Build, Test, Security, all the Check matrix), and the PR is `MERGEABLE`; `mergeStateStatus` is `BLOCKED` solely on the outstanding review.

### 🔴 Blocking — pruning `api/src` orphans every rustdoc "Source" link

I concur with @Audacity88, and I can make the concern concrete rather than hypothetical. `build_api` in xtask/src/cmd/mdbook/refs.rs:72 runs `cargo doc --no-deps --workspace` with no `RUSTDOCFLAGS`, so rustdoc generates its source-view tree (`doc/src/...`) and, on **every** item page, a "Source" link into it. I verified the exact form against a real rustdoc build: item pages emit relative links like `href="../../src/<crate>/<file>.rs.html#L<n>"`, and the `src/` directory is standard rustdoc output. `prune_rustdoc_source_view` (xtask/src/cmd/mdbook/build.rs) deletes `api/src` but nothing removes or rewrites those baked-in hrefs, so on the published site every per-item Source link resolves to a now-missing `/api/src/...` and 404s. That is a reader-facing dead-link regression, and FND-002's documentation exit criteria are explicit that there should be "No dead links in `docs/`".

This is genuinely easy to miss, which is worth saying plainly so the fix targets the right layer: the internal link checker cannot catch it. `check_internal_links` (xtask/src/cmd/mdbook/linkcheck.rs:24) walks only `book/<tag>/<locale>` and `is_checkable` deliberately returns false for anything under `api/` (linkcheck.rs ~129) because the api tree is rustdoc's own output, copied in after the check runs. So the green "Internal link check passed" in the validation evidence says nothing about these links, and the manual check that `book/master/en/api.html` resolves to `/api/zeroclaw/...` only exercises the top-level crate landing links, not the per-item Source links.

Resolving it can go a few ways, in roughly increasing cost: keep `api/src` (it is a real size cost, but it is what the Source links point at); strip source generation at build time so the links are never emitted (e.g. a rustdoc flag / settings toggle that disables the `[src]` links, then prune); or post-process the generated HTML to remove/retarget the source links and add a guard that fails the build if any `src/...` href survives in the api tree. Whichever path, please add evidence that no published rustdoc page still points at a missing `src/...` target, since neither the existing link gate nor the current manual check covers that surface.

### 🟢 What looks good — the deploy ownership model and the prune guard

The site-wide `/api` ownership rule is the right design and is applied consistently. In .github/workflows/docs-deploy.yml the api rsync sits inside `apply_run_slice`, so it is re-seeded on both the first push and the concurrent-retry re-clone after the `git rm -rf .` wipe, exactly the discipline that bit a sibling docs PR. The `[ "$TAG" = "master" ] || [ ! -d "api" ]` condition makes master the single writer and lets a tag only bootstrap `/api` when it is missing, mirroring the existing `_shared` rule so an older tag deploy can never regress newer rustdoc. Crucially, adding `"api"` to `ROOT_KEEP_DIRS` in xtask/src/cmd/mdbook/versions.rs is not incidental: `prune_root` lists top-level `api/` as exactly the kind of pre-versioned orphan it deletes, so without that entry the new root `/api` would be pruned on the very next deploy. Catching that interaction is the kind of detail that is invisible until production, and it was handled.

### 🟢 What looks good — `strip_chrome_hash` extraction and the mermaid fix

Pulling the chrome-hash logic out into `strip_chrome_hash` (xtask/src/cmd/mdbook/build.rs) is a clean refactor with a real bug fix behind it: the old `rfind('.')` found the *last* dot, so a compound extension like `mermaid-193ae996.min.js` never matched the 8-hex-digit gate and mermaid was silently never deduped. Scanning forward from the first dot after the hyphen fixes the specific missed case, and the four unit tests (single/compound extension, unhashed, non-hex) pin the boundaries directly. Adding `"mermaid"` to the shared-chrome allowlist alongside that fix is the right pairing.

### 🟡 Warning — the link checker's `api/` blind spot is now load-bearing

This is not a request to change the checker in this PR, but a flag worth recording: skipping the `api/` subtree in `linkcheck.rs` was a reasonable simplification when rustdoc was self-consistent (its internal links all resolved within its own output). This PR is the first change that deliberately removes part of the rustdoc tree while leaving links into it, which turns that blind spot from harmless into the exact gap that lets the blocker above ship green. Even after the blocker is resolved, a follow-up that adds a narrow assertion ("no surviving `src/...` href under the published api tree") would convert this class of regression from "discovered by a reader" into "caught by CI". Worth a tracking note even if it lands separately.

I cannot approve this as it stands: there is an active `CHANGES_REQUESTED` from @Audacity88 that I independently confirmed is a real reader-facing dead-link regression, so approving over it would be wrong on both protocol and substance. The deploy-side ownership model, the prune-keep guard, and the mermaid hash fix are all solid and I would happily approve the rest once the source-link surface is either preserved, suppressed, or proven clean with a guard. I'm posting this as a comment that concurs with and reinforces the existing block — surfacing the rustdoc detail and a path to resolution rather than stacking a second formal rejection.

WareWolf-MoonWall (commented):
# Review draft — PR #7754
**perf(docs): publish rustdoc once, dedupe per-locale assets, drop print pages**
Author: singlerider · Head SHA: `212240b6f6e05138df2e5f56f8ebfd0adc51d3b6`
Reviewer: WareWolf-MoonWall · Verdict: `--comment` (warnings only, no blockers)

---

Reviewed the full diff, local source cross-check, CI results, and the PR body against the template. No blockers. Two items worth fixing before or shortly after merge, and several things done well.

---

### 🟡 Warning — `prune_root` doc comment still lists `api/` as an orphan example

`xtask/src/cmd/mdbook/versions.rs` lines 79–80:

```rust
/// Remove orphaned root *directories* left over from the pre-versioned docs
/// layout (e.g. top-level `en/`, `fr/`, `api/`). Keeps the shared chrome dir
```

After this PR adds `"api"` to `ROOT_KEEP_DIRS`, the `api/` directory **survives** pruning — it is no longer an orphan. The example in this doc comment still lists it as something that gets removed, which will mislead anyone reading the source to understand the pruning policy. The code is correct; only the comment is stale.

Suggested fix: drop `api/` from the example list, e.g.:

```
/// layout (e.g. top-level `en/`, `fr/`). Keeps the shared chrome dir,
/// the site-wide rustdoc dir (`api/`), and every recognized version dir.
```

---

### 🟡 Warning — PR body label snapshot does not match applied GitHub labels

The PR body `Labels:` line reads `type: docs, risk: medium, size: M, docs`. The actual applied GitHub labels are `ci, docs, enhancement, risk: high, size: S`.

The risk escalation to `risk: high` is correct — CI workflow changes are automatically tagged at that tier. The `ci` and `enhancement` labels are also missing from the snapshot. The PR template asks for the snapshot *after* labels are applied (sidebar → auto-labeler). The body should be updated to reflect what is actually on the PR. The Rollback section does cover the `risk: high` case as required, so this is a housekeeping gap only.

---

### 🟢 What looks good — `strip_chrome_hash` correctly handles compound extensions

The root cause of the mermaid dedup miss was subtle: the old inline code used `rfind('.')` (last dot), so for `mermaid-193ae996.min.js` it produced the candidate hash string `193ae996.min` (12 chars), which failed the 8-char check and silently skipped the file. The extracted `strip_chrome_hash` helper uses `file_name[pos+1..].find('.')` — the **first** dot after the last hyphen — which correctly isolates the 8-char hex hash even with compound extensions like `.min.js`. The four unit tests cover the exact failing case (`strips_compound_extension_hash`), single-extension happy path, unhashed names, and the wrong-length/non-hex guard. That test suite is precisely the right validation artifact for an extracted helper like this.

---

### 🟢 What looks good — `api` added to `ROOT_KEEP_DIRS` closes a would-be silent deletion

Without this addition, the first `prune-root` run following deployment would treat the newly written root-level `api/` directory as an orphan from the pre-versioned layout and silently remove it — the same hazard the comment at line 79 was written to describe. Putting the fix in `ROOT_KEEP_DIRS` rather than hardcoding an exception in `prune_root` is the correct approach, and the constant is already the SSOT for this policy.

---

### 🟢 What looks good — workflow `api/` ownership model mirrors `_shared/` cleanly

The new `apply_run_slice` block follows the established single-writer pattern exactly: master unconditionally rewrites `/api/`, a tag only seeds it when the directory does not yet exist (bootstrapping). The inline comment explains the invariant in the same style as the `_shared/` block above it. The parallel structure makes the ownership rule easy to audit side-by-side.

---

### 🟢 What looks good — validation evidence matches the risk tier

For a change scoped to xtask + docs + workflow, the author ran:
- `cargo test -p xtask` → 79 passed + 5 asset binary suite, covering 4 new `strip_chrome_hash` tests
- `cargo clippy -p xtask --all-targets -- -D warnings` → clean
- `mdbook --tag master build` → confirmed 0 `print.html` files, `book/master/api` gone, 3-version site shrinks from ~933 MB to ~360 MB

That is the right validation boundary. The one acknowledged gap — the gh-pages first-deploy path through `apply_run_slice` runs only on GitHub Actions and was not exercised locally — is clearly documented in the PR body, and the logic directly mirrors the `_shared/` path that is already proven in production. The CI gate covers the xtask correctness side; the deploy path itself is tested on first merge to master.

Audacity88 (approved):
I re-reviewed the current head, 3298ca3ce5d90c4e1f3c380c5606d7d11b358086, following the new fix commit for the rustdoc source-link concern. I checked the current diff, prior reviews/comments, CI, the docs-deploy workflow, and the mdBook/rustdoc xtask paths, including the earlier comment reviews from @Nillth and @WareWolf-MoonWall. I also ran a targeted rustdoc verification to confirm current output still uses the `<a class="src" ...>Source</a>` shape that the new stripping helper targets. I did not rerun the full local docs build.

### ✅ Resolved — my earlier blocker: rustdoc source links are removed before pruning `api/src`

The new `prune_rustdoc_source_view()` pass now walks the generated `api` HTML before deleting `api/src`, strips each rustdoc source anchor, and only then removes the source-view tree. That closes the earlier blocker: generated item pages should no longer expose links to the pruned `/api/src/...` paths. The unit coverage pins the exact source-anchor case, multiple anchors, and non-source anchors, and a current rustdoc item page still emits the source link with `class="src"`, so the helper is matching the markup shape it needs to handle.

### 🟢 What looks good — root-owned API docs and shared Mermaid assets

The deploy ownership model is still the right shape: `master` owns the shared root `/api` tree, while tag deploys only seed it when the directory is missing. That mirrors the existing `_shared` rule and prevents an older tag deploy from regressing newer root rustdoc. The Mermaid dedupe fix is also well scoped; `strip_chrome_hash()` handles compound extensions like `mermaid-193ae996.min.js`, and the tests cover both the previous miss and the guard cases.

### 🟢 What looks good — merge surface is isolated despite the stale branch

The branch is behind current `master`, but the files touched by this PR have not changed on `master` since the PR's merge base, and a merge-tree check shows the current diff applying cleanly. That does not make the PR merge-ready by itself, because the deploy workflow should still get refreshed CI on the final merge result, but it does mean this approval is not resting on a conflicting or stale same-file assumption.

PR #7760, opened 2026-06-16
fix(i18n): auto-detect system locale when none is configured

NiuBlibing:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - CLI locale detection fell straight back to `"en"` whenever `config.toml` had no `locale` set, ignoring the host's actual language — despite the schema doc claiming env-var-based auto-detection. Restored that capability via the cross-platform `sys-locale` crate (Linux/macOS/Windows), since the original ad-hoc `LANG`/`LC_ALL` parsing had been silently dropped in a prior i18n rewrite.
  - `sys-locale` already implements `LANGUAGE` > `LC_ALL` > `LC_MESSAGES` > `LANG` precedence on Unix and queries the OS directly on Windows/macOS, so no hand-rolled env parsing was added — just `config.toml` → `sys-locale` → `"en"`.
  - Fixed a robustness gap in our own glue code: only the single top-priority `sys-locale` candidate was checked, so `LC_ALL=C` (common in CI/containers to force deterministic tool output) while `LANG` carried the real locale (e.g. `zh_CN.UTF-8`) caused detection to give up instead of trying `LANG`. Now walks every candidate.
  - Dropped a transient `ZEROCLAW_LOCALE` env var override that was reintroduced mid-PR — `config.toml`'s `locale` field is already the explicit override, so a second mechanism was unneeded surface area.
- **Scope boundary:** Does not change the `locale` config field's semantics or schema — only what happens when it's absent/empty. Does not touch the FTL loading/fallback logic, only `detect_locale()`.
- **Blast radius:** `zeroclaw-runtime::i18n` and `zeroclaw-tools::i18n` (the two independent, intentionally-duplicated locale-detection implementations — see comment in `zeroclaw-tools/src/i18n.rs` explaining why they can't share code). Adds one new dependency (`sys-locale`) to both crates and the workspace root.
- **Linked issue(s):** None.
- **Labels:** To be set by the path/size auto-labeler; suggested manual review: `risk: low`, `size: S`.

  ```
  $ cargo fmt --all -- --check
  (no output — clean)

  ...
      Finished `dev` profile [unoptimized + debuginfo] target(s) in 11m 24s
  (zero warnings across the whole workspace)

  $ cargo test
  ...
  test result: ok. 159 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
  test result: ok. 0 passed; 0 failed; 7 ignored (live-credential tests, expected)
  test result: ok. 9 passed; 0 failed; 0 ignored
  (zeroclaw-runtime i18n: 15 passed; zeroclaw-tools i18n: 5 passed — all green in earlier targeted runs)

  Doc-tests zeroclaw
  error: Option 'default-theme' given more than once
  ```

  - Scratch integration test (since removed) confirmed `LANG=zh_CN.UTF-8` with no config → detects `zh-CN`.
  - Scratch integration test confirmed `LC_ALL=C` + `LANG=zh_CN.UTF-8` → detects `zh-CN` (not `en`), proving the CI/container edge case is handled.
  - Did **not** verify on an actual Windows or macOS machine — relying on `sys-locale`'s own platform-specific implementations and test suite for those backends.
- **If any command was intentionally skipped, why:** The doctest run fails with `error: Option 'default-theme' given more than once` — cargo invokes rustdoc with `--default-theme=ayu` twice in this sandbox (visible in the raw rustdoc invocation). This is unrelated to this diff: `.cargo/config.toml` (the source of that flag) is untouched by this PR (`git diff master...HEAD -- .cargo/config.toml` is empty), and the flag duplication is a local toolchain/cargo-version interaction, not something introduced here.

## Compatibility (required)

- Backward compatible? **Yes** — no schema/API changes; users who already set `locale` in `config.toml` see no change.
- Config / env / CLI surface changed? **No** new fields/flags. Behavioral default changes: users with no `locale` set who previously always got English now get their OS/locale-derived language (or English if none is detected) — this was the originally-documented and intended behavior.
- If `No` or `Yes` to either: exact upgrade steps for existing users: None required. Users who want to force English regardless of system locale should explicitly set `locale = "en"` in `config.toml`.

## Rollback

Low-risk: `git revert <merge-sha>` is the plan.

---
<a href="https://app.blacksmith.sh/zeroclaw-labs/codesmith/zeroclaw/pr/7760"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784188339&installation_id=140579446&pr_number=7760&repository=zeroclaw-labs%2Fzeroclaw&return_to=https%3A%2F%2Fgithub.com%2Fzeroclaw-labs%2Fzeroclaw%2Fpull%2F7760&signature=fac6a5dc08451fe7fce4fc1be3b141c19c18e3c92f2b7ea2a3aa2ca9af511527"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


singlerider (approved):
First review (singlerider). Read the PR body and the full diff. CI is 18/18 green, no prior reviews. This restores OS locale auto-detection that was lost when the env-var fallbacks were removed, and does it through a single well-known crate. Approving.

### 🟢 What looks good: the LC_ALL=C shadowing case is handled deliberately

The load-bearing subtlety is `pick_locale` walking every candidate `sys-locale` returns rather than trusting the first. `LC_ALL=C` is routinely set in CI and containers to force deterministic tool output, and it sorts ahead of `LANG` in the POSIX precedence `sys-locale` reports, so taking the top candidate would discard a perfectly usable `LANG=zh_CN.UTF-8` and silently fall back to English. Skipping the `""` / `C` / `POSIX` sentinels and continuing to the next candidate is the correct behavior, and `pick_locale_skips_posix_sentinel_to_find_a_usable_candidate` pins exactly that.

### 🟢 What looks good: the logic is split into pure, env-free testable units

`normalized_env_locale` and `pick_locale` take their input by value/iterator instead of reading the environment, so the whole policy is covered by unit tests that never touch real env vars (sentinel rejection, POSIX-to-BCP47 normalization, sentinel-skip, all-sentinel returns None). That keeps the tests deterministic and parallel-safe, which is the right discipline for locale code. `sys-locale` is a small, widely-used crate whose only dependency is `libc`, so the footprint cost is minimal for the cross-platform coverage it buys (it queries the OS directly on Windows/macOS instead of guessing from env vars).

### 🟢 Note for a future drive-by: the two i18n modules now share this helper verbatim

`pick_locale` + `normalized_env_locale` are identical in `zeroclaw-runtime/src/i18n.rs` and `zeroclaw-tools/src/i18n.rs`. That mirrors the pre-existing duplication of `normalize_locale` / `detect_locale` across those two crates, so it is consistent with what is already there rather than new drift, and not blocking. Worth a follow-up to lift the shared locale-resolution logic into one place if a suitable shared crate edge exists.

I also dropped the spurious `provider:kilo` label, which the auto-labeler attached even though no provider code is touched.

singlerider:

Audacity88 (approved):
Reviewed current head 740e8286082ebea37f66919b9e7f69717d1bff18. I checked the live PR state, PR body, current labels, prior review/comment history, the full public diff, and current check state. I did not run local Cargo validation for this review; the visible GitHub checks are green, and the author’s validation notes cover fmt, clippy, tests, and the `LC_ALL=C` fallback case. Approving the code path; I do not see an author-facing blocker in the locale detection change.

### 🟢 What looks good — config stays the explicit override

The final resolution order is clean: configured `locale` still wins, absent or empty config falls through to the host locale, and only then does ZeroClaw use English. Dropping the temporary `ZEROCLAW_LOCALE` override was the right call because `config.toml` is already the explicit operator-controlled source of truth.

### 🟢 What looks good — sentinel locale handling is covered directly

The load-bearing behavior is that `LC_ALL=C` does not shadow a usable lower-priority locale candidate such as `LANG=zh_CN.UTF-8`. Splitting the selection into `pick_locale` and `normalized_env_locale` makes that behavior deterministic to test without mutating real process environment, and the runtime and tools modules both get the same sentinel/normalization coverage.

Issue #3566, opened 2026-03-15
[Tracker]: A2A protocol interoperability


5queezer:
### Summary

Add native A2A protocol support so a ZeroClaw instance can communicate with external agents (other ZeroClaw instances, NanoClaw, OpenClaw, or any A2A-compliant agent) over HTTP using the open [Agent2Agent protocol](https://github.com/a2aproject/A2A) (v0.3.0+, Linux Foundation).

### Problem statement

ZeroClaw's multi-agent story (#218, #3502) is currently **intra-instance only** — sub-agents live inside the same runtime and share the same process. There is no standardized way for a ZeroClaw agent to:

1. **Be discovered** by external agents (no published capability manifest).
2. **Receive tasks** from agents running on different hosts or frameworks.
3. **Delegate tasks** to remote agents without routing through a messaging channel like Telegram as a makeshift bus.

The broader ecosystem is converging on A2A as the standard for inter-agent communication. OpenClaw already has three community plugins (win4r/openclaw-a2a-gateway, n00n0i/openclaw-a2a, marketclaw-tech/openclaw-a2a). ZeroClaw's trait-driven, pluggable architecture is well-suited for this, but the integration doesn't exist yet.

### Proposed solution

Implement A2A as two new components that fit into ZeroClaw's existing extension model.

### 1. `A2ATool` — outbound (client role)

A `Tool` trait implementation that lets the agent send tasks to remote A2A agents.

```toml
# config.toml
[tools.a2a]
enabled = true
peers = [
  { name = "nanoclaw", url = "http://192.168.1.50:18800", token = "secret" },
]
```

Tool interface exposed to the agent:

- `a2a_discover { url }` — fetch a remote Agent Card, return capabilities
- `a2a_send { peer, message }` — send a task to a named peer, return task ID
- `a2a_status { peer, task_id }` — poll task status
- `a2a_result { peer, task_id }` — retrieve completed artifact

### 2. `A2AServer` — inbound (server role)

A lightweight HTTP server (built on the existing gateway's HTTP stack) that exposes:

- `GET /.well-known/agent-card.json` — auto-generated from ZeroClaw's identity config (`SOUL.md`, skills, allowed tools)
- `POST /a2a` — JSON-RPC 2.0 endpoint for receiving tasks per A2A spec

Inbound tasks are routed to the agent loop as if they were channel messages, reusing the existing `agent_turn` pipeline.

```toml
# config.toml
[a2a.server]
[...]

Issue #7218, opened 2026-06-04
RFC: A2A agent discovery (.well-known/agent-card.json) for multi-agent installs


singlerider:
### Summary

Define how A2A agent discovery (`/.well-known/agent-card.json`) works for a ZeroClaw install that hosts multiple agents, where the protocol assumes one agent per origin. Groundwork for interoperability with other "claw"-likes and external agent systems.

### Problem statement

This is groundwork for interoperability. A2A (Agent2Agent, Linux Foundation) is the emerging open standard letting agents built on different frameworks, by different vendors, on separate hosts discover and collaborate as peers. Getting discovery right is the precondition for a ZeroClaw install to interoperate with other "claw"-likes (NanoClaw, OpenClaw, other ZeroClaw instances) and with any A2A-compliant external agent system — without routing through a chat channel as a makeshift bus. Discovery is first contact: if a multi-agent ZeroClaw install can't present a correct, spec-compliant discovery surface, nothing downstream interoperates. This RFC settles that surface so the rest of A2A support (#3566, related) can build on it.

A2A discovery centers on one IANA-registered well-known URI (A2A spec §14.3): `GET https://{origin}/.well-known/agent-card.json` returns one Agent Card describing one agent's identity, capabilities, skills, endpoint, and auth. The mechanism assumes **one agent per origin**.

ZeroClaw hosts N agents per install under `[agents.<alias>]`, each with its own:

- workspace (`<install>/agents/<alias>/workspace/`)
- tool surface (`skill_bundles` / `mcp_bundles`) and guardrails (`risk_profile` / `runtime_profile`) — there is no `allowed_tools` field
- memory backend
- peer-group membership (`[peer_groups.<name>]`, members by alias)

There is no install-level agent to occupy the origin-root well-known path. The discovery design must decide what that path returns when the origin is multi-agent, and reconcile A2A's "external agent" concept with the existing `external_peers` substrate so we don't ship a second, parallel peer model (per AGENTS.md single-source-of-truth).

This RFC scopes the discovery surface only — the part of A2A support (#3566) that interacts directly with the multi-agent model. Outbound client tooling, task lifecycle, and transport selection are separate slices of #3566.

### Proposed solution

[...]

PR #7763, opened 2026-06-16
feat(gateway): A2A agent discovery surface

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds the A2A agent discovery surface to the gateway. A catalog card enumerates published agents at the origin (deliberately non-conforming to the one-agent-per-origin spec assumption), and each published agent serves a spec-shaped per-alias A2A card at a deterministic endpoint. Exposure is triple opt-in and default-closed (`[a2a.server] enabled`, per-alias `published`, `exposed_skills` filter) so nothing is reachable until an operator opts in. Skills resolve on demand from the existing skills service; per-alias and catalog cards both carry populated tags from skill frontmatter, no empty fields are emitted. The `message/send` task endpoint is folded into the gateway OpenAPI doc with schemas derived from the structs.
- **Scope boundary:** Discovery and the inbound task surface only. No outbound A2A client tooling, no signing, no credential-to-`external_peers` enforcement beyond the discovery config binding. The catalog root is intentionally not a runnable A2A agent.
- **Blast radius:** `zeroclaw-gateway` (new routes mounted on the existing axum listener), `zeroclaw-config` (new optional config blocks), `zeroclaw-runtime` (daemon writes resolved host/port into config for advertised URLs). All A2A routes are inert unless `[a2a.server] enabled = true`.
- **Linked issue(s):** Relates to #3566. Related to #7218.

```
cargo fmt --all -- --check        # clean, no diff
cargo clippy --all-targets -- -D warnings   # clean on zeroclaw-config and zeroclaw-gateway
cargo test                        # a2a + daemon + config + migration suites pass
mdbook build (all locales)        # internal link check passes
```

Re-review verification (this round's changes):

```
# all four feature combinations compile (the schema-export + a2a-off case was the regression)
cargo build -p zeroclaw-gateway                                                 # default
cargo build -p zeroclaw-gateway --no-default-features --features schema-export  # a2a off
cargo build -p zeroclaw-gateway --no-default-features --features a2a            # schema-export off
cargo build -p zeroclaw-gateway --no-default-features                           # both off

cargo test -p zeroclaw-gateway --features schema-export,a2a a2a::      # 17 passed
cargo test -p zeroclaw-gateway --features schema-export,a2a openapi::  # 3 passed
cargo fmt --all -- --check                                            # clean
cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings         # clean
cargo clippy -p zeroclaw-gateway --no-default-features --features schema-export --all-targets -- -D warnings  # clean
```

- **Beyond CI, what did you manually verify?** Stood up two isolated throwaway daemons (separate `--config-dir`, non-prod ports) with an underscore-aliased agent (`agent_alpha`) and a second deployment (`agent_beta`). Verified live: catalog served on the gateway listener at three paths (root alias, `/a2a/` canonical, per-alias), advertised URLs derive from the gateway host/port (and honor an advertise-only override with no second listener bound), per-alias cards scope to `exposed_skills`, disabled agents are absent from the catalog and return 404 per-alias, a real `message/send` POST ran an LLM turn and returned a completed Task with the reply as an artifact, and a `jq` walk found zero empty arrays/objects/strings across all payloads. Negative matrix: unknown alias 404, wrong method -32601, malformed params -32602/400. For the re-review changes, confirmed the OpenAPI spec renders identically with both features on (the a2a path and schemas are present) and that the spec still builds with a2a off. Latest round: the `/a2a/{alias}` operation now declares operation-level `security: []` so the spec stops advertising bearer auth on a deliberately unauthenticated handler (asserted by `a2a_task_operation_opts_out_of_bearer_auth`), and `exposed_skills` is now narrowed by the owning bundle's include/exclude filter through a single canonical `SkillBundleConfig::admits_skill` predicate shared with the runtime effective-skill load (asserted by `exposed_skills_drop_when_bundle_filter_excludes_them` and `skill_bundle_admits_skill_honors_include_and_exclude`). Ran `cargo fmt --all --check`, `cargo clippy -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-gateway --features a2a,schema-export --all-targets -D warnings` (clean), and the gateway suite with `--features a2a schema-export` (279 passed).
- **What I did NOT verify:** outbound client flows (not in scope), transports beyond JSONRPC, signing. The synchronous-turn timeout move is verified structurally (the task route now sits on the long-running router with the cron trigger); I did not stage a turn that exceeds 30s to observe the old cutoff.
- **If any command was intentionally skipped, why:** the OpenAPI spec is generated to `target/openapi.json` (gitignored) and validated by the in-crate `openapi::tests`; there is no committed snapshot to diff, so no `gen-openapi --check` step exists.

- New permissions, capabilities, or file system access scope? Yes. Publishing an alias exposes `POST /a2a/{alias}`, which runs a full agent turn for that alias through `run_gateway_chat_with_tools` with the alias's entire configured toolset (shell, file, browser, etc.). The endpoint is not behind the gateway's bearer/pairing auth (auth is enforced per handler, and the A2A task handler gates only on server-enabled + alias-published, not on a token), so anyone who can reach the gateway listener can invoke a published agent with no credential. This is the deliberate cross-deployment interop posture, not an oversight, but it is a real exposure: it stays default-closed behind three independent opt-in gates (server enabled, alias published, exposed_skills filter), and the operator-facing doc now spells out the contract and the steps to scope it (bind posture, reverse-proxy auth, minimal exposed_skills, tool selection per published alias).
- New external network calls? No (serves inbound on the existing gateway listener; no outbound calls added).
- PII, real identities, or personal data in diff, tests, fixtures, or docs? No (fixtures use neutral aliases like `agent_alpha`).
- If any `Yes`, describe the risk and mitigation: see the first bullet. Mitigation is operator-side scoping documented in `docs/book/src/gateway/a2a-discovery.md` ("What publishing actually exposes"): bind to a private interface or auth-enforcing reverse proxy, publish only aliases whose toolset you accept being invoked, and narrow exposed_skills. Whether the endpoint should honor pairing when `require_pairing` is on is flagged as an open team decision in the review thread.

## Compatibility (required)

- Backward compatible? Yes
- Config / env / CLI surface changed? Yes (new optional `[a2a.server]` and per-alias `[agents.<alias>.a2a]` blocks; all default-closed)
- Upgrade steps for existing users: none required. The A2A surface is off by default; existing deployments are unaffected until an operator sets `[a2a.server] enabled = true` and publishes an agent.

- **Fast rollback command/path:** `git revert` the merge, or set `[a2a.server] enabled = false` to disable the surface at runtime with no redeploy.
- **Feature flags or config toggles:** `[a2a.server] enabled` (runtime, default false); `a2a` gateway Cargo feature (default-on at compile).
- **Observable failure symptoms:** A2A routes returning unexpected status on `/.well-known/agents-card.json` or `/a2a/<alias>`; gateway startup logs for the A2A router mount.


Nillth (dismissed):
I read the full diff, the PR body, the A2A discovery doc, and cross-checked the gateway wiring against the source at head 4358e2a6eb5b (not my working tree, which is behind master). I focused on the exposure posture, the single-source-of-truth rule, and whether the committed docs match what the code actually emits. This is careful, well-tested work — the discovery design is genuinely thoughtful — and I have one blocking item plus a handful of warnings to resolve before this lands in v0.8.2.

Taking stock first: no prior reviews, comments, or inline threads, so nothing is settled yet and I'm not re-raising anything. The title already says DO NOT MERGE / v0.8.2, and the milestone is set to v0.8.2 — that's correct, no milestone action needed from me.

### 🔴 Blocking — the required Security & Privacy section understates what "publish" actually exposes

The Security & Privacy Impact section answers "New permissions, capabilities, or file system access scope? No" and "New external network calls? No", and the docs' only flagged "sharp edge" is the dashboard GET fallback. But publishing an agent does something much larger than that section admits: `POST /a2a/{alias}` runs a full agent turn through `run_gateway_chat_with_tools` (crates/zeroclaw-gateway/src/a2a.rs, handle_alias_task) — the agent's *entire configured toolset* (shell, file, browser, whatever that alias carries). And it runs with no authentication: bearer/pairing auth in this gateway is enforced per-handler (api::require_auth / pairing.is_authenticated, e.g. lib.rs ~2392 and the admin-reload gate at ~3572), not as a router-wide layer, and the a2a handlers call none of it. So once an operator flips the two opt-in switches, anyone who can reach the gateway listener can execute that agent — tools and all — with no token, while /api/config right next to it requires a bearer token.

I want to be precise about what's blocking and what isn't. The *open* posture may well be the intended design — the cross-deployment section ("calls a company deployment's vetted agents without sharing logins") only works if the endpoint is reachable without a shared secret, and A2A's own model puts auth in the card's security schemes. I'm not blocking on "you must add auth." I'm blocking on the disclosure: for a high-risk gateway change, the required security section has to state the real contract — publishing exposes an unauthenticated, remotely-invokable agent-execution endpoint that runs the alias's full toolset — and the operator-facing doc needs that stated just as plainly as the three opt-in gates are. Per AGENTS.md ("Do not silently weaken security policy or access constraints") and the template's required Security section, an operator deciding whether to publish must be able to read the true blast radius. Resolution: correct the Yes/No answers and their explanations, and add a prominent exposure note to docs/book/src/gateway/a2a-discovery.md (the "Turning it on" section is the natural home) spelling out that a published agent is callable by anyone who can reach the listener, that it runs the agent's tools, and what an operator should do to scope that (bind posture, which skills/tools to expose, reverse-proxy auth).

### 🟡 Warning — the catalog card's committed "real output" can't be produced by this code

docs/book/src/gateway/a2a-discovery.md states "Every response on this page is real output from a running daemon. Nothing here is illustrative," then shows catalog skills like `"tags": ["github", "issues", "triage", "agent_beta"]`. But build_catalog_card (a2a.rs) does `skill.tags = vec![alias.clone()]` — it *overwrites* the frontmatter tags with just the alias, so the real emitted value is `["agent_beta"]`. Your own test agrees with the code (exposed_skills_resolve_tag_and_scope_to_bundle asserts `catalog.skills[0].tags == vec!["maker"]`), which means the doc is the outlier. Two clean ways out, and I'd lean toward the first because it matches the stated intent ("skills … each tagged with its owning alias" reads as *in addition to* its own tags) and makes the catalog as useful as the per-alias cards: change the line to push the alias onto the resolved tags (`skill.tags.push(alias.clone())`) instead of replacing them, and update the test. Or, if the catalog really should carry only the alias tag, fix the doc example so the committed "real output" is actually real.

### 🟡 Warning — synchronous message/send shares the 30s gateway timeout

handle_alias_task runs the agent turn inline and returns the completed Task. But a2a_routes() is merged into `inner` (lib.rs:1716), which sits under the 30s `TimeoutLayer` (lib.rs ~1770). The cron manual-trigger endpoint was deliberately split into its own router with the 10-minute long-running timeout (lib.rs ~1783, and the comment at ~156 says exactly why: "synchronous agent [runs]"). A2A message/send is the same shape of work and will get cut off at request_timeout_secs for any non-trivial turn — your "Reply with PONG" manual test is too fast to surface it. Either route the task endpoint through the long-running timeout the way cron does, or document that message/send is bounded by request_timeout_secs and size that knob accordingly.

### 🟡 Warning — openapi build breaks when schema-export is on but a2a is off

crates/zeroclaw-gateway/src/openapi.rs build_spec is gated only on `#[cfg(feature = "schema-export")]` (openapi.rs:62) and unconditionally does `use crate::a2a::{JsonRpcRequest, OutTask};` (openapi.rs:64), plus the two schema_value::<…>() entries and the /a2a/{alias} path block. But the a2a module is `#[cfg(feature = "a2a")]` (lib.rs:15). Default features carry both, so the default build and CI's gen-openapi --check are green — but you made `a2a` a real toggle ("default-on at compile"), and `--no-default-features --features schema-export` won't compile because crate::a2a doesn't exist. Gate the a2a-specific openapi additions on `feature = "a2a"` (the `use`, the two schema entries, the path block, and the two a2a asserts in the tests) so the feature is actually toggleable.

### 🟡 Warning — vendored proto is missing its upstream license header

crates/zeroclaw-gateway/tests/fixtures/a2a-v1.proto is a verbatim vendor of the Linux Foundation / Google A2A v1 proto (package lf.a2a.v1), and the committed file starts at the protoc comment with no license header. The upstream carries an Apache-2.0 header; vendoring it without the header and a source URL is an attribution/compliance gap. Re-add the original license header and a one-line provenance comment (source repo + revision) at the top of the fixture.

### 🔵 Team decision — should the task endpoint be authenticatable?

This is the design question under the blocking item, and it's a direction call rather than one for any single reviewer. A2A puts auth in the card (securitySchemes), which we don't model yet; meanwhile this gateway already has pairing/bearer. Options worth the team weighing: keep it open for frictionless cross-deployment interop (current behavior); honor pairing when require_pairing is on so a locked-down deployment doesn't get an open bypass; or model the card's security scheme and advertise it. Whatever we choose, the card should eventually tell a client the truth about what auth it needs. Flagging for the group, not blocking the slice.

### 🔵 Suggestion — agent_description ignores the per-alias identity we already have

agent_description (a2a.rs) hardcodes `format!("ZeroClaw agent '{alias}'.")`, and the comment notes a richer source "can supersede this later." That source already exists in this very diff — AliasedAgentConfig.identity (schema.rs). Pulling the card description from the alias's identity document (first line / a summary field) instead of a synthetic string would make the cards meaningfully more useful and avoids leaving a placeholder on a published surface. Single source of truth already in hand.

### 🔵 Suggestion — daemon now mutates config.gateway.host/port as a side effect

daemon::run (crates/zeroclaw-runtime/src/daemon/mod.rs) takes `mut config` and writes the runtime host/port back into config.gateway so advertised_base resolves correctly. It works, but it overwrites the operator's configured gateway host/port in the live Config to make a derived value (the advertised URL) come out right — and if the bind host is 0.0.0.0, the catalog will advertise `http://0.0.0.0:<port>` unless public_base_url is set. Threading the effective host/port into the card builder (or resolving the advertised base from the listener directly) keeps Config as the single source of the *configured* values rather than reusing it as a scratchpad for the *resolved* ones. Minor, but it's the cleaner seam.

### 🟢 What looks good — the catalog-vs-conforming-card split

Refusing to squat `/.well-known/agent-card.json` with a non-conforming catalog and instead serving the plural `agents-card.json` for the aggregate while reserving the singular spec path for the genuinely-conforming per-alias cards (a2a.rs, the CATALOG_CARD_PATH doc comment) is exactly right. A standards-aware client that probes the origin root won't be misled into parsing a catalog as an agent. That's the kind of protocol-boundary care that prevents interop bugs nobody can debug later.

### 🟢 What looks good — exposure stays default-closed and DRY

Three independent gates (server enabled / alias published / exposed_skills filter) all defaulting closed, enforced per-request so a config reload toggles them without a router rebuild, is a clean posture. And exposed_skills resolves through the same SkillsService the dashboard uses and filters by the alias's declared bundles (the intruder test proves a skill in an undeclared bundle is dropped) — no parallel skill registry, bundles stay canonical. That's the single-source-of-truth rule (AGENTS.md) honored exactly where it's tempting to cache.

### 🟢 What looks good — the docs teach, and they're honest about the warts

The discovery doc is two-requests-and-done, every field is explained from the reader's perspective, and it owns the bare-GET-returns-dashboard fallback instead of hiding it. That's the FND-005 "name what's good / explain the principle" standard applied to user docs. Fix the tags example and add the exposure note and this becomes a genuinely strong reference.

---

Requesting changes — the blocking item is the inaccurate required security disclosure on a high-risk surface, not a demand to re-architect; correcting the disclosure + the exposure doc resolves it, and the four warnings are all concrete and small. Everything here is addressable without touching the core design, which is sound. Happy to re-review quickly once the security section, the tags mismatch, and the feature-gate are sorted.

WareWolf-MoonWall:
I read the PR body, Nillth's formal CHANGES_REQUESTED (submitted 2026-06-17T03:10:49Z at head 4358e2a6eb5b), and spot-checked the gateway A2A wiring against the diff. Nillth's review is thorough and correct: the blocking item (inaccurate Security & Privacy disclosure + missing operator-facing exposure doc) is real and required for a high-risk gateway change, and the four warnings are all concrete and addressable. I have no additional findings beyond what Nillth raised.

### Verdict

Per the PR review protocol (§189: "Never approve over another reviewer's active CHANGES_REQUESTED"), I'm not stacking a second review here. Nillth's block is active and legitimate; the author should address it. Once Nillth clears or converts their review, I'm happy to do an independent cross-check if requested.

For now: author action needed — resolve Nillth's blocker (correct the Security section Yes/No answers + add the exposure note to the a2a-discovery.md doc), fix the tags overwrite vs push mismatch, gate the openapi a2a additions on `feature = "a2a"`, and re-add the Apache-2.0 license header to the vendored proto fixture. Then re-request Nillth for re-review.

WareWolf-MoonWall (dismissed):
Re-reviewing at head 1494a4ff. My prior post at 4358a2a6 was a bare comment deferring to Nillth's CR; Nillth's review is now dismissed and there are no other active blocks, so this is my first formal review.

Taking stock: all five items from Nillth's CHANGES_REQUESTED are addressed in the new head. One new finding — the CI Required Gate is failing — is the only thing left before this can advance to merge.

### 🔴 Blocking — CI Required Gate is failing at 1494a4ff (Docs Style check)

Two of 17 checks are failing at the current head: `Docs Style` and `CI Required Gate`. The gate failure rolls up from `Docs Style`. The PR validation evidence lists `mdbook build (all locales)` passing and `cargo fmt --all -- --check` clean, but the docs quality gate is a separate check (`docs_quality_gate.sh` or equivalent) and the re-review validation block does not show it being run or passing locally.

Please run the docs style gate against the updated `docs/book/src/gateway/a2a-discovery.md` and any other touched doc files, fix whatever the check catches, and push a clean head. The CI Required Gate must be green before this can move forward in the review lane.

### ✅ Resolved — Security disclosure now correctly states the exposure (was Nillth's 🔴)

The Security & Privacy Impact section now answers "Yes" and explains the contract plainly: `POST /a2a/{alias}` runs a full agent turn with the alias's entire configured toolset, the task handler does not gate on a bearer token, and anyone who can reach the gateway listener can invoke a published agent with no credential. The `docs/book/src/gateway/a2a-discovery.md` section "What publishing actually exposes" states the unauthenticated-by-design posture directly and gives operators three concrete scoping steps: bind posture, alias and toolset selection per published alias, and `exposed_skills` narrowing. This was the blocking item; it is resolved.

### ✅ Resolved — Catalog card now pushes alias onto frontmatter tags; doc matches (was Nillth's 🟡)

`build_catalog_card` in `a2a.rs` now calls `skill.tags.push(alias.clone())` instead of `skill.tags = vec![alias.clone()]`. The catalog card aggregates the full frontmatter tag set with the alias appended, which is what the doc example shows (`"tags": ["github", "issues", "triage", "agent_beta"]`). Code and docs are now consistent.

### ✅ Resolved — `message/send` task endpoint is on the long-running router (was Nillth's 🟡)

`a2a_task_route()` is now merged into `long_running_router` (gated on `#[cfg(feature = "a2a")]`), which carries the 10-minute timeout for synchronous agent runs — the same router the cron manual-trigger endpoint uses. The 30-second gateway-wide `TimeoutLayer` on `inner` no longer applies to the task endpoint. Agent turns via A2A will not be cut off at the short gateway timeout.

### ✅ Resolved — OpenAPI additions are gated on `feature = "a2a"` (was Nillth's 🟡)

`build_spec` is now gated on `#[cfg(all(feature = "schema-export", feature = "a2a"))]` and the `use crate::a2a::{...}` import is gated on `#[cfg(feature = "a2a")]` as well. The PR validation evidence confirms all four feature-flag combinations compile cleanly — including the previously broken `--no-default-features --features schema-export` (a2a off) case.

### ✅ Resolved — Vendored proto carries Apache-2.0 header and source attribution (was Nillth's 🟡)

`tests/fixtures/a2a-v1.proto` now opens with "Copyright 2025 Google LLC", the full Apache-2.0 license header, and a source URL pointing to the upstream A2A specification repository. Attribution is correct and the compliance gap is closed.

### 🟢 What looks good — catalog-vs-per-alias card split is protocol-correct

Using the plural `/.well-known/agents-card.json` for the non-conforming catalog while reserving the spec-singular `/.well-known/agent-card.json` for the genuinely conforming per-alias cards is exactly right. A standards-aware A2A client probing the origin root will not be misled into treating a catalog as a single-agent card, and interop with compliant agents is not compromised.

### 🟢 What looks good — `exposed_skills` resolves from bundles, no duplicate registry

The `exposed_skills` filter operates over results resolved through the same `SkillsService` the dashboard uses, narrowing by the alias's declared bundles. There is no secondary skill registry in the A2A surface: skill bundles remain the single source of truth (per AGENTS.md), and the intruder test (`exposed_skills_resolve_tag_and_scope_to_bundle`) confirms that a skill in an undeclared bundle is dropped rather than advertised. This is the pattern AGENTS.md requires.

### 🔵 Suggestion — `agent_description` could read from `AliasedAgentConfig.identity`

`agent_description` in `a2a.rs` still hardcodes `format!("ZeroClaw agent '{alias}'.")`. This was Nillth's original suggestion: the identity document (`AliasedAgentConfig.identity`) was added in this same diff and is the natural source for a richer, operator-set description. The PR scope note that this can be improved in a follow-up is reasonable — but filing a tracking issue before merge would make the deferral concrete rather than aspirational.

### 🔵 Team Decision — A2A task endpoint and `require_pairing` (carry-forward from Nillth)

Nillth flagged this as a design question for the team: when `require_pairing` is on, the A2A task handler does not gate on a token, giving an open bypass through a path that neighboring endpoints (`/api/config`, admin-reload) do not share. The options are documented in Nillth's prior review and the updated doc covers the practical consequence. This is not a block on this slice, but the team should record a decision in this thread before v0.8.2 ships on whether the task endpoint should honor pairing — so the history is clear for whoever inherits this surface.

---

Requesting changes on the CI failure only. Fix the Docs Style gate, and all of the prior-round items are resolved. Happy to do a final pass once the required checks are green.

WareWolf-MoonWall (approved):
Re-reviewing at head `25186ef408dda6d2bf35882e0d09d4b718d25a8e`. My previous formal review was at `1494a4ff` and held a single 🔴 — the CI Required Gate failing (Docs Style check). That gate is now green: all 17 checks pass at the current head, including Docs Style and CI Required Gate. All items from Nillth's original CHANGES_REQUESTED were resolved prior to my last review and remain resolved. No other reviewer holds an active CHANGES_REQUESTED. The two carry-forward non-blocking items from my last round are preserved below with their original status markers, and one new 🟡 is added.

### ✅ Resolved — CI Required Gate now green (was my Round 15 🔴)

All 17 checks pass at `25186ef`: Docs Style, Format, Lint, Test, Build x86_64, Check (all features), Check (no default features), Check (32-bit), Check aarch64-apple-darwin, Check x86_64-pc-windows-msvc, Benchmarks Compile, Security, Installer Drift, Nix Module Eval, Zerocode RPC Boundary, and CI Required Gate itself. The docs quality gate that was failing at `1494a4ff` is clean. The re-review validation block now includes the explicit four-feature-combination build matrix and the docs-style run, and CI confirms them.

### ✅ Resolved — Security disclosure now correctly states the exposure (was Nillth's 🔴)

The Security & Privacy Impact section answers "Yes" with a plain statement of the actual blast radius: `POST /a2a/{alias}` runs a full agent turn through `run_gateway_chat_with_tools` with the alias's entire configured toolset, the task handler does not gate on a bearer token, and anyone who can reach the gateway listener can invoke a published agent with no credential. The `docs/book/src/gateway/a2a-discovery.md` section "What publishing actually exposes" states the unauthenticated-by-design posture directly in bold and gives operators three concrete scoping steps. The doc is unambiguous.

### ✅ Resolved — Catalog card now pushes alias onto frontmatter tags; doc matches (was Nillth's 🟡)

`build_catalog_card` in `a2a.rs` calls `skill.tags.push(alias.clone())` rather than overwriting with `vec![alias.clone()]`. The catalog card aggregates the full frontmatter tag set with the alias appended, which is exactly what the doc's "real output" example shows (`"tags": ["github", "issues", "triage", "agent_beta"]`). Code, test, and doc are now consistent.

### ✅ Resolved — `message/send` task endpoint is on the long-running router (was Nillth's 🟡)

`a2a_task_route()` is merged into `long_running_router` (gated on `#[cfg(feature = "a2a")]`), which carries the 10-minute timeout used by the cron manual-trigger endpoint. The 30-second `TimeoutLayer` on `inner` no longer applies to A2A task turns.

### ✅ Resolved — OpenAPI additions are gated on `feature = "a2a"` (was Nillth's 🟡)

`build_spec` is gated on `#[cfg(all(feature = "schema-export", feature = "a2a"))]`. All four feature combinations compile cleanly, including the previously broken `--no-default-features --features schema-export` case.

### ✅ Resolved — Vendored proto carries Apache-2.0 header and source attribution (was Nillth's 🟡)

`tests/fixtures/a2a-v1.proto` opens with "Copyright 2025 Google LLC", the full Apache-2.0 license header, and a source URL pointing to the upstream A2A specification repository.

### 🟡 Warning — Unauthenticated A2A task endpoint carries no rate-limit layer (new observation)

`handle_alias_task` runs a full agent turn — potentially shell, file, and browser tool chains — with no per-IP or per-origin request limit beyond what an upstream reverse proxy supplies. The "What publishing actually exposes" section documents the unauthenticated posture thoroughly and recommends binding to a private interface or sitting behind an auth-enforcing proxy, but does not call out rate limiting as a separate mitigation. An operator who publishes an agent on a publicly reachable gateway has a practical DoS surface: a flood of crafted `message/send` requests each trigger a full LLM-plus-tools turn with no circuit break.

The doc already leads with the binding and proxy-auth mitigations. Adding a single bullet that explicitly recommends configuring per-IP rate limiting at the reverse proxy (nginx `limit_req`, Caddy rate-limit middleware, etc.) would close this gap without any code change. Not a blocker — the triple opt-in gates and the documented bind posture are the right operator-level controls — but the rate-limit point is worth a doc line before v0.8.2 ships.

### 🟢 What looks good — catalog-vs-per-alias card split is protocol-correct

Using the plural `/.well-known/agents-card.json` for the non-conforming catalog while reserving the spec-singular `/.well-known/agent-card.json` for genuinely conforming per-alias cards is exactly right. A standards-aware A2A client probing the origin root will not be misled.

### 🟢 What looks good — `exposed_skills` resolves from bundles, no duplicate registry

The `exposed_skills` filter resolves through the same `SkillsService` the dashboard uses and narrows by the alias's declared bundles. No secondary skill registry in the A2A surface; the intruder test (`exposed_skills_resolve_tag_and_scope_to_bundle`) confirms a skill in an undeclared bundle is dropped. AGENTS.md SSOT rule is honored exactly where it's tempting to cache.

### 🔵 Suggestion — `agent_description` could read from `AliasedAgentConfig.identity` (carry-forward)

`agent_description` in `a2a.rs` still hardcodes `format!("ZeroClaw agent '{alias}'.")`. The identity document (`AliasedAgentConfig.identity`) added in this diff is the natural source. The PR scope note deferring this to a follow-up is reasonable — a tracking issue filed before merge would make the deferral concrete. The doc was updated to state that the bio from an AIEOS identity document supplies the description when configured, which is forward-looking parity.

### 🔵 Team Decision — A2A task endpoint and `require_pairing` (carry-forward from Nillth)

When `require_pairing` is on, the A2A task handler does not gate on a token, giving an open bypass through a path that neighboring endpoints (`/api/config`, admin-reload) do not share. The updated doc covers the practical consequence. The team should record a decision before v0.8.2 ships on whether the task endpoint should honor pairing — so the history is clear for whoever inherits this surface.

---

The one block from my last round is resolved and CI is clean. Approving. The 🟡 rate-limit doc note and the two 🔵 carry-forwards are addressable in follow-on issues or the v0.8.2 polish pass without holding the lane.

Audacity88 (dismissed):
Reviewing current head 25186ef408dda6d2bf35882e0d09d4b718d25a8e for #7763. I read the PR body, linked issues #3566 and #7218, the prior formal reviews, the current checks, and the head diff/source excerpts for the gateway/config/docs changes. Taking stock: Nillth's original blocker and warnings were addressed, WareWolf-MoonWall's Docs Style blocker is resolved, and CI is green. I found one new blocker in the generated OpenAPI contract plus two smaller follow-ups.

### 🔴 Blocking — OpenAPI still advertises bearer auth for the unauthenticated A2A task endpoint

The generated spec has a top-level `security: [{"bearerAuth": []}]` in `crates/zeroclaw-gateway/src/openapi.rs`, and `augment_spec_with_a2a` inserts `/a2a/{alias}` without overriding that operation's security. OpenAPI inheritance means the new A2A task operation is documented as requiring the pairing-derived bearer token, but the handler intentionally does not call the gateway bearer/pairing auth path, and the docs/PR body correctly say anyone who can reach the listener can invoke a published agent.

That leaves the machine-readable API contract saying the opposite of the high-risk exposure disclosure we just fixed. Please add an explicit empty operation-level security requirement, e.g. `"security": []`, on the `/a2a/{alias}` `post` entry and cover it with a spec test so the unauthenticated-by-design posture stays visible in generated clients and docs.

### 🟡 Warning — `Closes #3566` would auto-close the umbrella issue while scoped work remains

The PR body says `Closes #3566`, but #3566 is the broader A2A umbrella: outbound client/tooling, task lifecycle/status/result flows, bearer/auth/peer binding, signing/transport follow-ups, and cross-deployment sequencing are still intentionally out of scope here. The issue thread says #7218 is the discovery/card slice and #3566 still owns the broader work. Per the PR template, closing keywords should be used only when the PR fully resolves the issue. Please change this to a non-closing reference and/or file concrete follow-ups for the remaining #3566 pieces before merge.

### 🟡 Warning — advertised skills bypass bundle include/exclude filters

`exposed_skills` calls `SkillsService::list_skills(None)` and then checks only whether the matched skill's bundle alias appears in `agent.skill_bundles`. That skips the `SkillBundleConfig.include` / `exclude` filtering applied by the runtime in `load_skills_for_agent`, so an operator can advertise a skill that lives in a declared bundle but is excluded from the agent's actual effective skill set. Since the A2A docs promise the card is a filter over what the alias actually carries, please either resolve through the same effective-skill path or apply the bundle include/exclude rules here before accepting a named skill.

### 🔵 Suggestion — path-prefixed gateways need advertised URLs with the prefix

When `gateway.path_prefix` is configured, the A2A routes are nested under that prefix with the rest of the gateway, but `advertised_base` appends `/.well-known/agents-card.json` and `/a2a/{alias}` directly to the host/port unless `public_base_url` is set. That means the fallback advertised URLs are wrong for a path-prefixed deployment. `public_base_url` is a workaround, so I would not block on this, but folding the normalized prefix into the fallback would make the generated cards match the actual listener shape.

### 🟢 What looks good — the previous exposure fixes held

The updated docs are much more honest about what publishing means, the task route now sits on the long-running timeout path, the A2A OpenAPI additions are feature-gated correctly, and the identity-derived card description follow-up landed cleanly. The catalog/per-alias split still looks like the right protocol shape.

Requesting changes for the OpenAPI security-contract mismatch. The other items are cleanup/follow-up caliber, but the generated API contract should tell the truth before this high-risk gateway surface merges.

WareWolf-MoonWall (approved):
Post-approval re-review of #7763. My Round 20 approval was at `25186ef408dda6d2bf35882e0d09d4b718d25a8e`. Two commits were pushed after that approval: `ae2abefad96f` (fix(gateway/a2a): unauth A2A op in spec, filter exposed skills by bundle) and `05d84e663774` (Merge branch 'master' into integration/a2a). Audacity88's CHANGES_REQUESTED review (posted at `25186ef4`, finding two new issues) is now DISMISSED. CI is 18/18 green at the current head.

This review covers the incremental changes only. All items from prior rounds (Nillth's CHANGES_REQUESTED, my Round 20 review) remain resolved as previously confirmed. Prior resolved items are preserved with ✅ below for reference.

---

### ✅ Resolved — CI Required Gate now green (was my Round 15 🔴)

Confirmed in Round 20. Unchanged.

### ✅ Resolved — Security disclosure now correctly states the exposure (was Nillth's 🔴)

Confirmed in Round 20. Unchanged. The PR body's Security & Privacy section and `docs/book/src/gateway/a2a-discovery.md` "What publishing actually exposes" remain accurate and explicit about the unauthenticated-by-design posture.

### ✅ Resolved — Catalog card pushes alias onto frontmatter tags; doc matches (was Nillth's 🟡)

Confirmed in Round 20. `skill.tags.push(alias.clone())` is unchanged.

### ✅ Resolved — `message/send` task endpoint is on the long-running router (was Nillth's 🟡)

Confirmed in Round 20. Unchanged.

### ✅ Resolved — OpenAPI additions gated on `feature = "a2a"` (was Nillth's 🟡)

Confirmed in Round 20. Unchanged.

### ✅ Resolved — Vendored proto carries Apache-2.0 header and source attribution (was Nillth's 🟡)

Confirmed in Round 20. Unchanged.

### ✅ Resolved — `agent_description` sources from identity document (was my Round 20 🔵, closed by `25186ef4`)

Confirmed at the prior approval head. Unchanged.

### ✅ Resolved — OpenAPI spec no longer advertises bearer auth on the A2A task operation (was Audacity88's 🔴)

Commit `ae2abefad96f` adds `"security": []` at operation level in `augment_spec_with_a2a` (`crates/zeroclaw-gateway/src/openapi.rs`), explicitly opting the `/a2a/{alias}` POST operation out of the top-level `bearerAuth` security scheme. This correctly makes the machine-readable API contract match the handler's deliberately unauthenticated posture.

Test `a2a_task_operation_opts_out_of_bearer_auth` asserts `spec.pointer("/paths/~1a2a~1{alias}/post/security")` is an empty array. Gated on `#[cfg(all(feature = "schema-export", feature = "a2a"))]`. The spec contract now tells the truth.

### ✅ Resolved — `exposed_skills` now applies bundle include/exclude filter (was Audacity88's 🟡)

Commit `ae2abefad96f` introduces `SkillBundleConfig::admits_skill(name: &str) -> bool` in `crates/zeroclaw-config/src/schema.rs`:

```rust
pub fn admits_skill(&self, name: &str) -> bool {
    if !self.include.is_empty() && !self.include.iter().any(|s| s == name) {
        return false;
    }
    !self.exclude.iter().any(|s| s == name)
}
```

`exposed_skills()` in `a2a.rs` now calls `bundle.admits_skill(s.r#ref.name())` as a third condition. `load_skills_for_agent()` in `zeroclaw-runtime/src/skills/mod.rs` was simultaneously refactored to use the same predicate, replacing its own inline include/exclude logic. This is an SSOT improvement: one canonical predicate governs both the runtime's effective-skill load and the A2A card's skill advertisement.

Tests confirm correctness:
- `skill_bundle_admits_skill_honors_include_and_exclude` (in `zeroclaw-config`) covers include-only, exclude-only, and combined filter logic.
- `exposed_skills_drop_when_bundle_filter_excludes_them` (in `zeroclaw-gateway`) exercises the gateway path with a `gadget` excluded from the bundle; the A2A card omits it while serving `widget`.

No new security-sensitive paths are touched. No new auth bypasses. No new bare string literals. No AI/bot attribution trailers in either commit. Commit signatures are valid PGP-signed.

### 🟡 Warning — Unauthenticated A2A task endpoint carries no rate-limit layer (carry-forward from Round 20)

Unchanged from my Round 20 note. The doc already recommends binding to a private interface or auth-enforcing reverse proxy; a follow-up bullet explicitly recommending per-IP rate limiting at the reverse proxy remains a doc-polish item for the v0.8.2 polish pass.

### 🔵 Suggestion — `agent_description` could read from `AliasedAgentConfig.identity` (carry-forward, now resolved)

Closed by `25186ef4` (confirmed in Round 20). Not applicable.

### 🔵 Team Decision — A2A task endpoint and `require_pairing` (carry-forward from Nillth)

No new code changes affect this decision. Remains open as a governance item for the team before v0.8.2 ships.

### 🔵 Suggestion — `Closes #3566` references umbrella issue (was Audacity88's 🟡)

The PR body still reads "Relates to #3566" (not `Closes`). Audacity88's concern was addressed when the author corrected the closing keyword to a non-closing reference. Confirmed in current PR body.

---

The two commits since my Round 20 approval are targeted fixes for Audacity88's blocking concern (OpenAPI contract) and a non-blocking concern (bundle filter SSOT improvement). Both are safe, additive, and governance-clean. CI is 18/18 green. No active CHANGES_REQUESTED from any reviewer. **Re-approving.**

Audacity88 (approved):
I re-reviewed the current head (`05d84e663774071e8b9a056768c9adb3d670b760`) against the PR body, the live linked issues, the prior review thread, and the current gateway/config/runtime/docs diff. I did not run a local live A2A smoke test; this is a source review against the green CI and the manual smoke evidence already recorded on the PR.

### ✅ Resolved — OpenAPI now matches the unauthenticated task endpoint

My previous blocker is resolved. The generated A2A `POST /a2a/{alias}` operation now explicitly opts out of bearer auth with an empty `security` array, and the PR carries a regression test for that shape. That matches the implementation and the docs/body language that publishing an alias exposes an unauthenticated agent invocation endpoint to any caller that can reach the listener.

### ✅ Resolved — advertised skills now respect the runtime bundle filters

The advertised skill set now shares the same include/exclude predicate as runtime skill loading via `SkillBundleConfig::admits_skill`, and the A2A tests cover a bundle-level exclusion being dropped from the card. That addresses my earlier warning about cards advertising skills the alias would not actually load.

### 🟡 Warning — task failures still echo raw agent errors to unauthenticated callers

One remaining follow-up I would file/track: `handle_alias_task` currently returns JSON-RPC `-32000` with `Agent task failed: {e}` on agent execution failure. The normal `/webhook` path sanitizes/logs provider errors and returns a generic caller-facing failure instead. Because the A2A task endpoint is intentionally unauthenticated, I think we should converge that error surface toward the existing sanitized/generic pattern so provider/config/internal error text is not handed back to arbitrary reachable callers.

I do not think that warning should keep this slice blocked given the explicit opt-in exposure model, the fixed OpenAPI contract, and the now-correct skill filtering. The still-open pairing/rate-limit decision remains a team/governance follow-up for the v0.8.2 A2A posture rather than a new blocker in this diff.

WareWolf-MoonWall (approved):
# Review — PR #7763 `feat(gateway): A2A agent discovery surface`

**Prior review:** Round 23 — APPROVED at `05d84e663774071e8b9a056768c9adb3d670b760`
**Current head:** `05d84e663774071e8b9a056768c9adb3d670b760` (same — no new commits)
**Verdict:** `--approve`

---

Current head is identical to the commit at which both WareWolf-MoonWall (2026-06-21T16:12:34Z)
and Audacity88 (2026-06-22T21:50:54Z) approved. No new commits have been pushed. No active
`CHANGES_REQUESTED` from any reviewer. The "more new commits" flag in the re-review queue
appears to have been a false positive — the PR is stable at `05d84e6`. This approval confirms
the prior review is still the active verdict.

---

### ✅ Confirmed — All Round 23 blockers remain resolved

From my Round 23 review:
- Security & Privacy section corrected: publishing an alias exposes an **unauthenticated**,
  remotely-invokable agent-execution endpoint running the alias's full toolset. The disclosure
  is explicit and accurate. ✅
- Operator docs (`docs/book/src/gateway/a2a-discovery.md`) include the prominent exposure note
  in the "What publishing actually exposes" section. ✅
- Feature gating: `cfg(feature = "a2a")` + `cfg(feature = "schema-export")` combinations all
  compile cleanly. ✅
- Proto fixture `tests/fixtures/a2a-v1.proto` includes the upstream Apache-2.0 header and
  source provenance comment. ✅
- Tags: `build_catalog_card` pushes the alias tag onto the resolved tags rather than overwriting;
  doc examples and tests are consistent. ✅
- `a2a_routes()` mounted on the long-running router (not the 30 s `TimeoutLayer`), matching
  the cron-trigger precedent. ✅

---

### ✅ Audacity88 independently approved at the same head

Audacity88's approval at `05d84e6` (2026-06-22T21:50:54Z) confirms:
- OpenAPI opts out of bearer auth on the A2A task operation (regression test present). ✅
- Advertised skills respect runtime bundle filters via `SkillBundleConfig::admits_skill`. ✅

Their 🟡 warning about task failures echoing raw agent errors to unauthenticated callers
is a valid follow-up item to track but is not a merge blocker given the explicit opt-in
exposure model and operator-facing doc. ✅

---

### 🟡 Carry-forward from Round 23

- **Error sanitization on task failure:** `handle_alias_task` returns `-32000` with
  `Agent task failed: {e}` verbatim, potentially leaking internal/provider error text to
  unauthenticated callers. The normal webhook path sanitizes to a generic caller-facing message.
  Track as a follow-up for the v0.8.2 A2A posture.
- **Pairing/rate-limit team decision:** Whether the task endpoint should honor
  `require_pairing` when set remains an open governance question for the team. Not a merge
  blocker per the explicit opt-in model.
- **`daemon::run` config mutation:** writes `host`/`port` back into `config.gateway` as a
  side-effect to resolve the advertised base URL. Cleaner to thread the effective values
  through the card builder rather than using `Config` as a scratchpad. Minor architectural
  follow-up.

---

14 commits — no AI attribution trailers in the commit history. ✅

### ✅ All CI checks passing

18/18 CI jobs pass at run `27908779876`. ✅

### ⚠️ Note: PR is flagged `DO NOT MERGE — INTENDED for v0.8.2`

The PR body has a DO NOT MERGE notice (milestone v0.8.2). This approval is a code-review
verdict only; merge timing is a maintainer decision per the v0.8.2 release schedule.

PR #7765, opened 2026-06-16
fix(zerocode): use grapheme clusters for backspace and cursor movement

Pick-cat:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `pop_input_char()` (backspace) now deletes one grapheme cluster instead of one code point, so CJK characters and multi-code-point graphemes (emoji with modifiers, flag sequences, combining marks) are removed in a single press.
  - `move_cursor_left()` and `move_cursor_right()` now move by one grapheme cluster instead of one code point, keeping cursor positions aligned with visible characters.
  - Added `unicode-segmentation = "1"` dependency to the zerocode app crate. It was already present in `Cargo.lock` as a transitive dependency of ratatui.
- **Scope boundary:** Only changes the zerocode TUI input bar (`apps/zerocode/src/input_bar.rs`). It does not touch search-box `.pop()` calls in other zerocode panes (logs, file_explorer, dashboard), and it does not change the separate `zeroclaw agent` CLI input path.
- **Linked issue(s):** Related to #6995. The `zeroclaw agent` CLI path was fixed by #7672; this PR covers the separate zerocode input bar.
- **Labels:** `bug`, `size: XS`, `risk: medium`, `dependencies`, `zerocode`

## Validation Evidence

```bash
cargo check -p zerocode
```

- **Commands run:** Local compile check passed.
- **Beyond CI — what did you manually verify?** Reviewed that `.graphemes(true)` correctly handles extended grapheme clusters (CJK single characters, emoji sequences, combining marks). Verified that `replace_range` correctly removes the entire cluster byte range.
- **If any command was intentionally skipped, why:** Full `cargo test` skipped locally (no system linker); CI will run the full battery.

## Security & Privacy Impact

## Compatibility

- Config / env / CLI surface changed? **No** (only internal zerocode cursor/deletion logic)

## Rollback

Revert this PR if zerocode input-bar editing regresses.


singlerider (approved):
First review (singlerider). Read the PR body and the full diff. CI is 17/17 green, no prior reviews. Correct grapheme-cluster fix for the zerocode input bar, closing #6995. Approving.

### 🟢 What looks good: all three edit operations move to graphemes consistently

`pop_input_char`, `move_cursor_left`, and `move_cursor_right` all switch from code-point stepping to `graphemes(true)` (extended grapheme clusters), so a CJK character, an emoji-with-modifier, a flag sequence, or a combining-mark cluster is deleted or stepped over as one visible unit instead of leaving a half-formed cluster behind. Fixing all three together is what keeps the cursor model coherent: deleting by grapheme but moving by code point would desync the cursor from what the user sees. The byte math is right (`prev_start = cursor - prev_grapheme.len()`, `replace_range(prev_start..cursor, "")`, and `cursor += next_grapheme.len()`), and the `unwrap_or("")` guards keep the empty-edge cases from panicking. `unicode-segmentation` is already in the tree as a ratatui transitive, so this adds no new external surface.

### 🟢 Worth a follow-up: a unit test for the grapheme edge cases

The substitution is simple enough to verify by reading, but a small `#[test]` over `pop_input_char` / `move_cursor_*` with a CJK string, an emoji-with-skin-tone, and a combining-mark cluster would pin the exact regression #6995 is about and guard against a future revert to code-point stepping. Not blocking given the size and clarity of the change, but it is the kind of thing that is cheap to add now and annoying to rediscover later.

(Minor: the body lists `channel:cli` / `risk: low` / `size: S` while the applied labels are `risk: medium` / `size: XS`. The applied labels are fine; just noting the body snapshot drifted.)

singlerider:

Audacity88 (approved):
Reviewed current head `e820ef20422df8541f1d788b6acb0e075bccf869`, including the current PR body, live PR state, linked #6995 and merged #7672, singlerider's prior approval, current checks, the `apps/zerocode/src/input_bar.rs` diff, and the separate `zeroclaw agent` CLI input path. I did not run my own local Cargo; the PR body reports `cargo check -p zerocode`, and the current remote CI/check suite is green.

### 🟢 What looks good — scope boundary is now clear

The body now makes the right distinction: #7672 fixed the separate `zeroclaw agent` CLI input path, while this PR covers zerocode's custom input bar. That keeps the public record clear for future readers following the linked issue and PRs.

### 🟢 What looks good — input editing now steps over grapheme clusters

`pop_input_char()`, `move_cursor_left()`, and `move_cursor_right()` now all use `graphemes(true)` and update the byte cursor by the selected cluster length. That keeps deletion and cursor motion on the same boundary, so combining-mark clusters and multi-code-point emoji sequences stay intact, while CJK characters continue to move and delete as one visible unit.

### 🔵 Suggestion — add a focused grapheme regression test in a follow-up

I agree with singlerider's non-blocking test note. A small unit test for CJK text, a combining-mark sequence, and an emoji modifier sequence would make this behavior harder to regress. The implementation is narrow and readable enough that I would not block the PR on that.

PR #7767, opened 2026-06-16
fix(channels/email): stabilize missing message ids

### LINKED ISSUE #7738: bug(channels/email): use stable UID fallback when Message-ID is missing
## Affected component

channel

## Severity

S2 - degraded behavior

## Current behavior

Email parsing still falls back to a random UUID when `Message-ID` is missing. For IMAP-backed messages, reconnecting or refetching the same mailbox item can therefore produce different local message IDs for the same stable UID.

## Expected behavior

When `Message-ID` is absent, the email channel should use a stable fallback derived from mailbox identity and IMAP UID or another existing stable source. The fix should preserve privacy by avoiding raw account identifiers in public IDs.

## Steps to reproduce

1. Inspect `crates/zeroclaw-channels/src/email_channel.rs` on current `master`.
2. Follow the missing-`Message-ID` fallback path.
3. Parse or refetch the same IMAP message without `Message-ID`.
4. Observe that the fallback can change because it is random rather than UID-stable.

## Impact

Affected users: email channel users with messages that lack `Message-ID`.

Frequency: depends on upstream email source quality.

Consequence: duplicate detection, message continuity, and reconnect/refetch behavior can drift for the same message.

## Logs / stack traces

None.

## ZeroClaw version

## Rust version

## Operating system

## Regression?

Unknown.

## Pre-flight checks

## Related

- Follow-up from the 2026-06-15 maintainer sweep.

Audacity88:
## Summary

- **What changed and why:**
  - Replace random missing-`Message-ID` email fallbacks with stable synthetic IDs so refetch/reconnect paths do not create a new message identity for the same mail.
  - Include IMAP `UIDVALIDITY` with UID-based fallback IDs so UID reuse across mailbox epochs does not collide.
  - Use a mailbox-scoped content hash fallback when UID/UIDVALIDITY is unavailable.
  - Suppress synthetic internal fallback IDs from outbound `In-Reply-To` headers.
- **Scope boundary:** This does not change IMAP search/fetch behavior, email channel configuration, storage schema, or behavior for emails that already provide a real `Message-ID`.
- **Blast radius:** Email channel message identity and reply-header generation for inbound messages missing `Message-ID`. Other channels/providers are not affected.
- **Linked issue(s):** Closes #7738
- **Labels:** `bug`, `risk: high`, `size: M`, `channel`, `channel:email`

  - `cargo fmt --all -- --check` - passed; no formatter output.
  - `git diff --check` - passed; no output.
  - `cargo test -p zeroclaw-channels build_parsed_email -- --nocapture` - `test result: ok. 5 passed; 0 failed; 757 filtered out`
  - `cargo test -p zeroclaw-channels synthetic_email_message_ids_are_not_reply_header_ids -- --nocapture` - `test result: ok. 1 passed; 0 failed`
  - `cargo clippy -p zeroclaw-channels --all-targets -- -D warnings` - passed.
  - `cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings` - `Finished dev profile ... in 29m24s`
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` - failed on unrelated `zeroclaw-runtime` test `tools::delegate::tests::background_agentic_delegate_rebinds_memory_tools_to_target_agent_scope`; summary was `5825 passed, 1 failed, 10 skipped` before cancellation.
  - `cargo test -p zeroclaw-runtime background_agentic_delegate_rebinds_memory_tools_to_target_agent_scope -- --nocapture` - `test result: ok. 1 passed; 0 failed; 2198 filtered out; finished in 2.57s`
- **Beyond CI - what did you manually verify?** Verified real `Message-ID` values are preserved, missing-ID fallbacks are stable for the same mailbox/UID/UIDVALIDITY, UIDVALIDITY changes produce different synthetic IDs, mailbox identity is hashed rather than embedded in generated IDs, content fallback remains stable when UIDVALIDITY is unavailable, and synthetic fallback IDs are not emitted as outbound reply headers. I did not perform a live IMAP/SMTP smoke test.
- **If any command was intentionally skipped, why:** Full local `cargo test` and `./dev/ci.sh all` were not run. I used focused email-channel tests, repo-wide fmt, workspace clippy, and a CI-shaped nextest run; the broad nextest run hit an unrelated runtime delegate timeout, and the failing test passed on focused rerun.

- If any `Yes`, describe the risk and mitigation: The runtime fallback hashes mailbox identity and parsed message content when deriving synthetic IDs; generated IDs do not embed raw account identifiers or message content. Tests use reserved/example addresses only.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No upgrade steps required. Emails with real `Message-ID` keep existing behavior; emails missing `Message-ID` now receive stable synthetic IDs instead of random UUIDs.

- **Fast rollback command/path:** `git revert e02e1c648727f7f9c18c88b117784a8823602e41`
- **Observable failure symptoms:** Duplicate or missing email-channel messages after reconnect/refetch for mail without `Message-ID`, unexpected grouping/threading for such messages, or outbound replies containing malformed/unexpected reply headers.

Not applicable; this PR does not use `Supersedes #`.

### DIFF
--- a/crates/zeroclaw-channels/src/email_channel.rs
+++ b/crates/zeroclaw-channels/src/email_channel.rs
@@ -21,6 +21,7 @@ use mail_parser::{MessageParser, MimeHeaders};
 use pulldown_cmark::{Options, Parser, html};
 use rustls::{ClientConfig, RootCertStore};
 use rustls_pki_types::DnsName;
+use sha2::{Digest, Sha256};
 use std::collections::HashSet;
 use std::sync::Arc;
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
@@ -28,7 +29,6 @@ use tokio::net::TcpStream;
 use tokio::sync::{Mutex, mpsc};
 use tokio::time::{sleep, timeout};
 use tokio_rustls::TlsConnector;
-use uuid::Uuid;
 
 use zeroclaw_api::channel::{Channel, ChannelMessage, SendMessage};
 use zeroclaw_tools::email_imap::{ImapSession, TlsStreamTolerant};
@@ -351,7 +351,12 @@ impl EmailChannel {
     /// Bounds peak memory when the mailbox has a large unseen backlog.
     const MAX_FETCH_BATCH: usize = 10;
 
-    fn build_parsed_email(&self, parsed: &mail_parser::Message, _uid: u32) -> ParsedEmail {
+    fn build_parsed_email(
+        &self,
+        parsed: &mail_parser::Message,
+        uid: u32,
+        uid_validity: Option<u32>,
+    ) -> ParsedEmail {
         let sender = Self::extract_sender(parsed);
         let subject = parsed.subject().unwrap_or("(no subject)").to_string();
         let body_text = Self::extract_text(parsed);
@@ -359,7 +364,9 @@ impl EmailChannel {
         let msg_id = parsed
             .message_id()
             .map(|s| s.to_string())
-            .unwrap_or_else(|| format!("gen-{}", Uuid::new_v4()));
+            .unwrap_or_else(|| {
+                self.stable_missing_message_id(parsed, uid, uid_validity, &sender, &body_text)
+            });
         #[allow(clippy::cast_sign_loss)]
         let timestamp = parsed
             .date()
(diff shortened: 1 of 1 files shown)


singlerider (approved):
No prior reviews. Single-file fix in `email_channel.rs`, risk:high because it changes inbound message identity, and the diff earns that label with the care it takes. The core decision is right: a missing `Message-ID` must not get a fresh `Uuid::new_v4()` per fetch, because refetch and reconnect then mint a new identity for the same mail. Deriving a stable synthetic ID from mailbox-scoped inputs (host, username, folder) plus UIDVALIDITY+UID, falling back to a content hash, makes the identity reproducible across sessions. Plumbing `uid_validity` through every fetch path (`fetch_unseen_active`, `fetch_new`, `process_new`, idle and poll loops) is complete and consistent.

One non-blocking consideration, not a formal finding: `is_synthetic_email_message_id` keys off the `email-imap-` / `email-fallback-` prefixes, so a real inbound `Message-ID` literally beginning with one of those strings would be wrongly suppressed from `In-Reply-To`. The collision is implausible in practice and the privacy upside of stripping synthetic IDs is the right tradeoff, so I would not hold merge on it, just flagging for awareness.

### 🟢 What looks good - synthetic IDs never leak mailbox PII

The fix hashes host, username, and folder into the digest but the emitted ID is only the hex of the truncated digest plus the UID. The tests prove this directly: `build_parsed_email_uses_stable_uid_fallback_without_message_id` and `build_parsed_email_missing_uid_fallback_is_stable_and_private` both assert the result does not contain the host, username, or folder. For a risk:high change to message identity, proving the non-leak property in the test rather than asserting it in prose is exactly right.

### 🟢 What looks good - UIDVALIDITY guards against UID reuse across epochs

Including `UIDVALIDITY` (and only when `uid != 0`) means a mailbox that resets its UID space does not produce colliding IDs for what are actually different messages. The content-hash fallback for when UID/UIDVALIDITY is unavailable keeps stability even on servers that do not surface those, and `build_parsed_email_missing_uid_validity_uses_content_fallback` covers that path.

### 🟢 What looks good - suppressing synthetic IDs from outbound headers

Stripping `email-imap-*` / `email-fallback-*` from outbound `In-Reply-To` is the correct correctness guard: a synthetic ID is internal-only and must never be sent on the wire as if it were a real RFC 5322 identifier. `synthetic_email_message_ids_are_not_reply_header_ids` pins the predicate, including the negative case for a real ID.

### 🟢 What looks good - scope and dependency hygiene

The change is one file, `sha2` and `hex` were already crate dependencies, and dropping the `use uuid::Uuid` import leaves no unused dep since other channels still use it. Privacy fixtures use `.invalid` addresses throughout.

PR #7768, opened 2026-06-16
feat(line): loading indicator, icon/nickname switch, and bind reply feedback

dvgamerr:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Added **Loading Indicator** — fires `POST /v2/bot/chat/loading/start` after both the group policy gate and DM allowlist/pairing gate pass, so the animated typing bubble appears in LINE while the agent processes an authorized request. Exception: the `/bind` pairing handshake fires the loading indicator before the agent turn, since the bind flow handles the response itself and never reaches the agent.
  - Added **Icon & Nickname Switch** — every outgoing message (Reply API and Push API), including the `/bind` outcome replies, now includes a `sender` object with `name` (configurable via `sender_name` in `LineConfig`, resolved from live config at send-time and falling back to `"AI"` when unset or empty) and `iconUrl` from the bot's profile photo fetched at startup from `GET /v2/bot/info`; the configured `sender.name` is normalized to LINE's 20-character limit at send-time so an overlong operator value cannot invalidate the outbound message
  - Added **Bind Reply Feedback** — `/bind` flow now sends an inline reply for each outcome (success / wrong code / rate-limited) instead of silently dropping; previously users had no indication whether pairing succeeded; strings route through the Fluent i18n catalogue (`channel-line-bind-*` keys in `cli.ftl`); these bind replies carry the same `sender` object as normal agent replies
  - Fixed `send_bind_reply` hardcoding `https://api.line.me` — it now uses the configured `api_base_url`, which was causing bind reply tests to escape to the real LINE API and broke proxy-override support
  - Threaded the send-time sender resolver and fetched icon handle onto `LineState` so the webhook-driven bind replies render with the configured sender identity rather than the default Official Account identity
- **Scope boundary:** changes are confined to `crates/zeroclaw-channels/src/line.rs`, `crates/zeroclaw-channels/src/orchestrator/mod.rs` (send-time `sender_name` resolver wiring), `crates/zeroclaw-config/src/schema.rs` (new `sender_name` field on `LineConfig`), and `crates/zeroclaw-runtime/locales/en/cli.ftl` (three new `channel-line-bind-*` keys)
- **Blast radius:** LINE channel only; Telegram, Lark, WeChat, and email channels are unaffected
- **Linked issue(s):** none
- **Labels:** `enhancement`, `channel`, `config`, `channel:core`, `risk:medium`, `channel:line`, `size:L`

---

### 1. Loading Indicator

> **LINE Docs:** https://developers.line.biz/en/docs/messaging-api/use-loading-indicator/

Shows an animated loading animation in the chat while the agent is processing. The indicator fires **only after both the group policy gate and the DM allowlist/pairing gate pass**, so the typing bubble is never shown to unauthorized senders. The sole pre-agent exception is the `/bind` pairing handshake: the bind path fires the loading indicator inside `LineDmPolicy::Pairing` handling before sending the inline bind-outcome reply and `continue`-ing without forwarding to the agent.

**Request sent to LINE API:**
```http
POST https://api.line.me/v2/bot/chat/loading/start
Authorization: Bearer {channel_access_token}
Content-Type: application/json

{
  "chatId": "U4af4980629..."   ← userId for DM, groupId for group chat
}
```

**Behavior matrix:**

| Source | `chatId` used |
|---|---|
| DM (`type: user`) | `userId` of the sender |
| Group (`type: group`) | `groupId` |
| Room (`type: room`) | `roomId` |

- `loadingSeconds` not specified → defaults to 5 s (LINE's default)
- Fire-and-forget: a non-2xx from LINE logs at `WARN` and does **not** drop the inbound message

**What the user sees in LINE:**

```
User  →  sends a message
Bot   →  ⋯  (animated dots appear within ~100 ms)
Bot   →  "Here is my answer…"
```

---

### 2. Icon & Nickname Switch

> **LINE Docs:** https://developers.line.biz/en/docs/messaging-api/icon-nickname-switch/

Every outgoing message now carries a `sender` object so LINE renders a custom display name and profile photo next to the bot's bubble — distinct from the Official Account's default identity.

**Bot info fetched at startup:**
```http
GET https://api.line.me/v2/bot/info
Authorization: Bearer {channel_access_token}

→ {
    "userId": "U1afe9a65...",
    "displayName": "Popcorn",
    "pictureUrl": "https://profile.line-scdn.net/..."
  }
```

**Message payload sent (Reply API):**
```http
POST https://api.line.me/v2/bot/message/reply
Authorization: Bearer {channel_access_token}
Content-Type: application/json

{
  "replyToken": "nHuyWiB7...",
  "messages": [
    {
      "type": "text",
      "text": "Hello! How can I help?",
      "sender": {
        "name": "AI",
        "iconUrl": "https://profile.line-scdn.net/..."
      }
    }
  ]
}
```

**Design decisions:**

| Decision | Reason |
|---|---|
| `sender.name` defaults to `"AI"` | LINE auto-appends `from 'Popcorn'` — repeating the bot name would show "AI Popcorn from 'Popcorn'" |
| `sender.name` configurable via `LineConfig.sender_name` | Operators who want "Bot", "Assistant", or a localized name can override without a code change |
| `sender.iconUrl` from `pictureUrl` | Bot's own profile photo; no extra asset needed |
| Empty or unset configured name falls back to `"AI"`; the low-level sender builder still omits `sender` for a direct empty input | LINE API rejects empty `sender.name`, so the user-facing send path keeps a valid fallback while the helper remains defensive |

**What the user sees in LINE:**

```
┌──────────────────────────────────────┐
│  🤖  AI  ·  from 'Popcorn'          │
│  Hello! How can I help?              │
└──────────────────────────────────────┘
```

---

### 3. Bind Reply Feedback

The `/bind <code>` pairing flow previously accepted or rejected silently. Users had to guess whether pairing worked. Now the bot replies inline using the event's `replyToken`. All three outcome strings are Fluent-keyed (`channel-line-bind-*` in `crates/zeroclaw-runtime/locales/en/cli.ftl`) — no bare English literals reach the LINE chat bubble:

| Outcome | Fluent key | Default English text |
|---|---|---|
| ✅ Valid code | `channel-line-bind-success` | `"✅ Paired! You can now chat."` |
| ❌ Wrong code | `channel-line-bind-invalid-code` | `"❌ Invalid code. Please try again."` |
| ⏱ Rate-limited | `channel-line-bind-rate-limited` | `"⏳ Too many attempts. Retry in 30s."` |

Reply is sent via `POST /v2/bot/message/reply` with the `replyToken` from the same webhook event — no Push API quota consumed. Non-2xx responses from LINE are now logged at `WARN` with the HTTP status code.

---

```
cargo fmt --all -- --check
→ (no output) fmt OK

cargo clippy --all-targets --features channel-line -p zeroclaw-channels -- -D warnings
→ Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.52s
  (no warnings, no errors)

cargo test --package zeroclaw-channels --features channel-line line::
→ test result: ok. 93 passed; 0 failed; 0 ignored; 0 measured; 899 filtered out; finished in 1.62s
```

**New test coverage:**

| Test | What it pins |
|---|---|
| `build_sender_obj_empty_name_returns_none` | Empty name → no sender (LINE rejects empty `sender.name`) |
| `build_sender_obj_name_only_no_icon` | `sender.name` present, `iconUrl` absent when no `pictureUrl` |
| `build_sender_obj_with_icon_url` | Both `name` and `iconUrl` in sender object |
| `build_sender_obj_truncates_overlong_name_to_line_limit` | Overlong configured name truncated to LINE's 20-char `sender.name` cap |
| `build_sender_obj_truncates_multibyte_name_on_char_boundary` | Truncation counts characters, never splits a multibyte codepoint |
| `send_reply_includes_sender_name_when_set` | Reply API body has `messages[0].sender.name = "AI"` |
| `send_reply_includes_sender_icon_when_set` | Reply API body has `messages[0].sender.iconUrl` |
| `send_reply_uses_ai_fallback_when_no_resolver` | Unset/empty resolver falls back to `sender.name = "AI"` |
| `send_push_includes_sender_name_when_set` | Push API body has sender name |
| `webhook_dm_sends_loading_indicator_with_user_id` | Loading `chatId` = `userId` for DM |
| `webhook_group_sends_loading_indicator_with_group_id` | Loading `chatId` = `groupId` for group |
| `webhook_loading_indicator_failure_does_not_break_message_delivery` | 500 from loading endpoint → message still delivered |
| `webhook_bind_success_sends_paired_reply` | Valid code → i18n `channel-line-bind-success` inline reply |
| `webhook_bind_invalid_code_sends_error_reply` | Wrong code → i18n `channel-line-bind-invalid-code` inline reply |
| `bot_info_deserializes_picture_url` | `pictureUrl` field parsed from `/v2/bot/info` response |
| `bot_info_picture_url_defaults_to_none` | Missing `pictureUrl` → `None` (no serde error) |
| `listen_sets_sender_icon_from_bot_info` | `listen()` stores fetched `pictureUrl` into `sender_icon` |

  - Deployed to live daemon (`zeroclaw service restart`); confirmed port 8443 listens and `connected as 'Popcorn'` appears in trace log
  - Sent DM in LINE app — loading indicator bubble appeared immediately before bot replied
  - Outgoing messages show **AI** `from 'Popcorn'` with bot profile photo (icon/nickname switch confirmed visually)
  - `/bind <wrong>` replied "❌ Invalid code. Please try again." in LINE directly
- **If any command was intentionally skipped, why:** Full `cargo test` suite skipped — `line::` module under `--features channel-line` is the scope of this change; full suite requires `memory-postgres` infra not available locally

- New external network calls? `Yes` — `POST /v2/bot/chat/loading/start` is a new outbound call to the LINE API. It fires only after both policy gates pass (group and DM). It is fire-and-forget; failure is logged at WARN and does not affect message delivery.
- Secrets / tokens / credentials handling changed? `No` — the existing `channel_access_token` bearer is reused; no new credential surface
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No` — test fixtures use synthetic IDs (`Ubot`, `Uuser1`, `Ggroup123`); the live `userId` captured during manual testing does not appear in the diff

## Compatibility (required)

- Backward compatible? `Yes` — `LineConfig.sender_name` is `Option<String>` with `#[serde(default)]`; existing configs without the field are unaffected and continue to use `"AI"` as the sender name
- Config / env / CLI surface changed? `Yes` — new optional `sender_name` field in `[channels.line.<alias>]` config block; fully backward-compatible

**Risk: medium** — three additive network calls and one new config field.

After squash-merge, revert the single squash commit: `git revert <squash-sha>`. No state migration needed; no data is written or deleted by this PR.

**Observable failure symptoms** (signals that a revert is needed):
- LINE users receive no inline confirmation after `/bind` attempts (silent again)
- Loading animation no longer appears for authorized senders in DM or group chats
- Outgoing messages lose the sender object overlay (`AI from 'BotName'` label disappears in LINE UI)
- No data loss; no config migration needed on rollback


singlerider (changes requested):
First review (singlerider). Read the PR body and the full single-file diff including the test module. CI is 19/19 green and the feature work is well-tested (loading DM/group/failure, sender object across all branches, bind feedback success/invalid, BotInfo serde). The icon/nickname switch, the bind-reply feedback, and the `api_base_url` fix for `send_bind_reply` are all good. One thing blocks: the loading indicator emits ungated outbound API traffic for unauthorized senders.

### 🔴 Blocking: the loading indicator fires before the policy gate, so unauthorized senders trigger bot API calls

`send_loading_indicator` is called inside the per-event loop after signature verification but before the group policy gate (the `is_group` match) and before the DM allowlist/pairing gate (`is_line_user_allowed`). The PR body states this is intentional ("before any policy gate"), but for a deny-by-default runtime it is the wrong order. The signature only proves the webhook came from LINE, not that the sender is authorized: under `LineGroupPolicy::Mention` any member of any group the bot sits in, and under DM `Pairing`/allowlist any unpaired or non-allowlisted user, will each cause an authenticated `POST /v2/bot/chat/loading/start` on every message they send. That is a side effect on an unauthorized path and an amplification vector (a spammer in a shared group makes the bot hammer the LINE API on their cadence), which is exactly what the policy gates exist to prevent.

Move the loading-indicator call to fire only once a message has passed the gate that decides it will actually be processed (after the group policy check for group/room, after `is_line_user_allowed` for DMs, with the bind path being the one deliberate exception since pairing is itself the authorization handshake). Showing the bubble a moment later for messages you are going to act on is the correct tradeoff against emitting API traffic for messages you are going to drop. `webhook_loading_indicator_failure_does_not_break_message_delivery` already proves it stays fire-and-forget wherever it lands.

### 🔴 Blocking: `listen_sets_sender_name_to_ai` asserts the opposite of its name

The test is named for verifying `listen()` sets `sender_name` to "AI", but it bypasses `listen()` and asserts the pre-listen defaults (`*sender_name.read() == ""`, icon `None`), with a comment conceding "the real assertion is covered by the integration path." As written it is a green test that does not exercise the contract it claims, so a regression that left `sender_name` empty after `listen()` would not be caught here. Either drive the real `listen()` path (you already mock `/v2/bot/info` returning `displayName`/`pictureUrl` in `listen_sets_sender_name_to_ai`'s sibling) and assert `sender_name == "AI"` and `sender_icon == Some(pictureUrl)` after it runs, or rename it to reflect that it only checks pre-init defaults. The `bot_info_deserializes_picture_url` test already covers serde, so the gap is specifically the "listen() applies it" link.

### 🟢 What looks good: the sender object and bind-reply plumbing

`build_sender_obj` returning `None` on an empty name (LINE rejects empty sender names) and only attaching `iconUrl` when present is the right shape, and it is applied uniformly to both the Reply and Push paths with matching tests. Threading `api_base_url` onto `LineState` and through `send_bind_reply` fixes the hardcoded `https://api.line.me` that was escaping test mocks to the real API, and the bind flow now gives the user explicit success/invalid/rate-limited feedback instead of silently dropping. Holding the sender identity on `LineChannel` behind `RwLock` and seeding it once in `listen()` is a reasonable place for it.

Audacity88 (commented):
Re-reviewed zeroclaw-labs/zeroclaw PR #7768 head `3486af4220169bb5cf3f323b4553da626ac076b1` against singlerider's active `CHANGES_REQUESTED` review. I checked the current diff, the updated LINE tests, public checks, the freshness snapshot, the PR body, and live labels.

### ✅ Resolved — Loading indicator now runs after the normal policy gates

`send_loading_indicator` now runs after the group policy gate and after the DM allowlist/pairing gate for normal messages. The only remaining pre-agent exception I see is `/bind`, which is the pairing handshake path rather than a dropped unauthorized chat path.

### ✅ Resolved — `listen_sets_sender_name_to_ai` now exercises `listen()`

The test now drives `listen()` against a mocked `/v2/bot/info` response and asserts `sender_name == "AI"` plus the fetched `pictureUrl`, so it covers the contract its name describes.

### 🟡 Warning — I am not approving or calling this merge-ready yet

singlerider's formal `CHANGES_REQUESTED` review is still active and should be resolved by that reviewer or maintainer judgment.

The branch is also stale against current `master` (`95` commits behind). The public freshness snapshot shows this is a risky channel file. I did not find overlapping files, and a local `git merge-tree --write-tree origin/master refs/tmp/pr-7768` produced a merge tree, but the standard path before approval is still branch refresh plus CI, an exact merge-result smoke, or explicit maintainer stale-risk acceptance.

### 🔵 Suggestion — Sync the PR body before merge

The body still says the loading indicator fires "before any policy gate", and its label snapshot should match the live labels: `enhancement`, `size: L`, `risk: medium`, `channel`, `channel:line`.

singlerider (approved):
Re-review (singlerider). Pulled the branch to head `ce4631cf52`, built `zeroclaw-channels` locally (clean), and ran the LINE suite (36 passed, 0 failed). Both of my prior blockers are resolved in the current code, so I am clearing my `CHANGES_REQUESTED`.

### ✅ Resolved — Loading indicator now fires only after the policy gates

`send_loading_indicator` on the normal message path (line 596) now runs after the group policy gate (447) and the DM allowlist/pairing gate (472), guarded by the comment "Show loading indicator only for messages that passed both policy gates." The sole pre-agent call left (495) is inside the `Pairing` bind branch, which is the authorization handshake itself, not a dropped unauthorized chat path. This is the correct ordering for a deny-by-default runtime: no authenticated `POST /v2/bot/chat/loading/start` for senders the bot is going to drop. The amplification vector I flagged is closed.

### ✅ Resolved — `listen_sets_sender_name_to_ai` now exercises the real contract

The test drives `listen()` against a mocked `/v2/bot/info` returning `displayName`/`pictureUrl` and asserts `sender_name == "AI"` plus the fetched icon after the call, so it now verifies the link its name claims rather than the pre-init defaults.

### 🟢 Sender object and bind-reply plumbing remain solid

`build_sender_obj` returning `None` on empty name, `iconUrl` only when present, applied uniformly to Reply and Push with matching tests; `api_base_url` threaded through `send_bind_reply` so the bind flow no longer escapes test mocks to the real API; explicit success/invalid/rate-limited user feedback on bind. All good.

### 🟢 Branch staleness is acceptable here

The branch is well behind master, but the touched surface is a single channel file with no overlapping edits on master, and a local merge tree resolves clean. Refresh and let CI run before merge as the final gate, but this is not a code blocker.

One non-blocking housekeeping note: the PR body still describes the loading indicator as firing "before any policy gate," which is now the opposite of the shipped behavior. Sync the body before merge so the description matches the code.

tidux (approved):
Re-review at `ce4631cf52a55597695f5864c631f3b0d76d54c6` (= prior head `3486af4220` + `master` merge `ce4631cf52` resolving the staleness Audacity88 flagged). Single file changed: `crates/zeroclaw-channels/src/line.rs`, +661/-3. I read the full diff, walked the policy-gate ordering against the LINE webhook handler, ran the focused test slice locally, and cross-checked the three commits against singlerider's two blockers and Audacity88's interim re-review.

The technical surface is in good shape — both `CHANGES_REQUESTED` blockers are genuinely resolved on this head, the work is well-tested, and singlerider's `APPROVED` on the same head matches what I see in the code. The one open item is the PR body sync that both prior reviewers asked for and the author hasn't picked up yet, which I'm seconding below as non-blocking but worth doing before merge so the commit record matches the shipped behavior.

### ✅ Resolved — loading indicator now fires only after the policy gates pass

`webhook_handler` at `line.rs:444-585` runs the group policy gate (`447-469`) and the DM policy gate (`471-585`) *before* the normal-message `send_loading_indicator` call at `596-602`. The `// Show loading indicator only for messages that passed both policy gates.` comment at 586 names the contract directly. The only remaining pre-agent loading call is at line 495 inside the `LineDmPolicy::Pairing` branch — guarded by `extract_bind_code` requiring the literal `/bind` prefix at `line.rs:894`, and the subsequent `PairingGuard::try_pair` is rate-limited (the `Err(wait_ms)` branch at 564 surfaces the bind-rate-limit feedback to the user). That's exactly the deny-by-default ordering for a webhook-authenticated channel: signature proves LINE sent it, but only authorized senders cause outbound API traffic. The amplification vector singlerider flagged on `c3cabcc20b` is closed, with `webhook_loading_indicator_failure_does_not_break_message_delivery` keeping the fire-and-forget guarantee pinned wherever the call lands.

### ✅ Resolved — `listen_sets_sender_name_to_ai` exercises the real `listen()` path

`line.rs:2646-2689` now stands up a `wiremock::MockServer` responding to `/v2/bot/info` with `displayName: "Popcorn"` and `pictureUrl: "https://profile.line-scdn.net/popcorn.png"`, clones the `sender_name`/`sender_icon` `Arc` handles before spawning, drives the real `listen()` task, and asserts `*sender_name.read() == "AI"` plus `*sender_icon.read() == Some("https://profile.line-scdn.net/popcorn.png")` after a 200 ms settle. That's the contract the name claims — a regression that left `sender_name` empty post-`listen()` would now fail this test. The 200 ms timer is the same shape other channel tests use; not ideal but not the bar to hold.

### ✅ Resolved — `send_bind_reply` uses `api_base_url`

`line.rs:144-170` constructs the URL from the `api_base_url` parameter and is threaded through `LineState.api_base_url` at every bind-reply call site (`531`, `552`, `573`). The hardcoded `https://api.line.me` that was escaping test mocks to the real API is gone, and the bind-feedback flow (success / invalid code / rate-limited) is uniformly observable in tests via the mock-server host. Quietly nice.

### 🔵 Non-blocking — PR body still describes the pre-gate ordering, label/risk snapshot is stale

Seconding what @singlerider's APPROVED review and @Audacity88's COMMENTED review have both already said: the body still claims the loading indicator "fires `POST /v2/bot/chat/loading/start` immediately on every incoming message before any policy gate" (Summary §2) and "Called immediately before any policy gate" (Feature Details §1). That is now the opposite of the shipped behavior on this head. The body also lists `Labels: type: feature, risk: low, size: M, channel: line` while the live labels on the PR are `enhancement, size: L, risk: medium, channel, channel:line`. Squash-merge takes the PR body as the commit message, so a body that says "before any policy gate" lands as the permanent record of what shipped, which makes future debugging harder than it needs to be. Specifically worth syncing:

- Summary §2: replace "before any policy gate" with the post-gate ordering (with the bind-handshake exception named)
- Feature Details §1 "Behavior matrix": same correction in the prose above the matrix
- Summary label snapshot: `enhancement`, `size: L`, `risk: medium` (the rollback section is also keyed to `risk: low` in its preamble, but the Rollback section is otherwise correct)

Calling this 🔵 rather than 🟡 because @singlerider explicitly tagged it as non-blocking housekeeping when he approved. I agree on the severity, just naming it here so the body sync stays visible across the request queue.

### 🔵 Non-blocking — fire-and-forget on transport error only, not on non-2xx

`send_loading_indicator` (`line.rs:122-143`) and `send_bind_reply` (`line.rs:144-170`) both log only on the transport error branch (`if let Err(e) = client...send().await`). A non-2xx response from LINE (e.g. revoked token, malformed `chatId`, rate-limit headers) returns `Ok(response)` from reqwest and is silently discarded. The PR body's Feature Details §1 claims "a non-2xx from LINE logs at `WARN`", which doesn't match the implementation. Functionally this is still correct fire-and-forget — message delivery is unaffected, and the existing `webhook_loading_indicator_failure_does_not_break_message_delivery` test exercises a 500 response and passes — but the operator-visible failure mode is weaker than what the description promises. If you're going to update the body anyway, either fix the prose to "transport-error failures are logged" or add a `resp.status()` check + `WARN` log inside the helper. The latter is two lines per helper and gives operators a chance to spot a stuck bearer token; either is fine.

### 🟢 What looks good — uniform `sender` plumbing across Reply and Push

`build_sender_obj` at `line.rs:964-973` returns `None` on empty name (which LINE's API rejects — pre-`listen()` defensive guard), only emits `iconUrl` when `Some`, and is applied byte-identically at the Reply path (`984`) and the Push path (`1023`) inside the same `.chunks(5)` batch shape. The `sender_name = "AI"` choice is the right call given LINE auto-appends `from 'BotName'` — the alternative would render "AI Popcorn from 'Popcorn'" which is the kind of nominal redundancy a careful operator would have to disable. `bot_info_picture_url_defaults_to_none` and `bot_info_deserializes_picture_url` pin the serde shape so a future LINE API change that drops `pictureUrl` doesn't 500 the listen path.

### 🟢 What looks good — bind reply feedback is the right shape

The pairing branch at `line.rs:490-583` consumes the same `replyToken` from the original webhook event for all three outcomes (success / invalid / rate-limited), so the inline feedback is free relative to the Push API quota. The rate-limited message includes the `Retry in {secs}s.` hint derived from `wait_ms / 1000`, which is more useful than a bare error. The pairing path also still calls `persist_line_paired_identity` *before* sending the success reply, so a paired-but-failed-persist state still emits an honest WARN log (`paired userId= but persist failed`) rather than confusing the user with a "Paired!" reply on an unpersisted record.

### 🟢 What looks good — `LineState` cleanup

Dropping the unused `sender_name` field from `LineState` (webhook handler is inbound-only; outbound sender identity lives on `LineChannel` behind `RwLock<String>`) is the right ownership split. The state struct that gets cloned per-event no longer carries a field it doesn't use.

---

Verification locally on `ce4631cf52`:

- `cargo clippy --all-targets --features channel-line -p zeroclaw-channels -- -D warnings` → clean.
- `cargo test --package zeroclaw-channels --features channel-line --lib line::` → `68 passed; 0 failed` (matches the literal output in the PR body's Validation Evidence section exactly).
- Individually re-ran `listen_sets_sender_name_to_ai` and `webhook_loading_indicator_failure_does_not_break_message_delivery` to spot-check the two prior-blocker regression tests — both pass in isolation.
- Walked the policy-gate ordering against the webhook handler at `line.rs:444-602`: confirmed normal-message `send_loading_indicator` runs *only* after both group and DM gates; confirmed bind-branch loading indicator is guarded by `extract_bind_code` + `PairingGuard::try_pair` rate limit.
- 18/18 CI checks pass on the head.

Approving. The technical work is solid and the prior blockers are real-resolved. Please pick up the body sync before squash-merge so the commit record matches the shipped policy-gate ordering and the live labels; that single edit closes the only thing two reviewers have flagged on this head.

WareWolf-MoonWall (changes requested):
# Review — PR #7768 `feat(line): loading indicator, icon/nickname switch, and bind reply feedback`

**Head reviewed**: `ce4631cf52a55597695f5864c631f3b0d76d54c6`  
**CI**: ✓ 18/18 gates pass (run ID `27716431382`)  
**Active blocking reviews**: None — singlerider cleared CHANGES_REQUESTED and APPROVED on this head; tidux APPROVED on this head; Audacity88 COMMENTED (non-blocking).  
**Verdict**: `--request-changes` — one new 🔴 blocker (Fluent/i18n) found on independent review; filing independently since no other active CR exists.

---

### 🔴 Blocking — Bind reply strings are bare English literals sent directly to LINE users

`send_bind_reply` is called three times with hardcoded English strings that appear as chat messages in users' LINE apps:

```rust
// line.rs — bind success path
send_bind_reply(..., "Paired! You can now chat.").await;

// bind invalid-code path
send_bind_reply(..., "Invalid code. Please try again.").await;

// bind rate-limited path
send_bind_reply(..., &format!("Too many attempts. Retry in {secs}s.")).await;
```

These are **user-facing output** — they render as LINE chat bubbles visible to the end user. Per AGENTS.md: *"All user-facing output (CLI messages, tool descriptions, onboarding prompts) must use `fl!()`/Fluent strings — never bare string literals."* The governance protocol confirms: bare literals are 🔴 Blocking.

**Note on codebase state**: `zeroclaw-channels` currently has no Fluent infrastructure (`crates/zeroclaw-channels/locales/` does not exist; no `fl!()` calls anywhere in the crate). Resolving this blocker requires:
1. Standing up a `zeroclaw-channels/locales/en/line.ftl` (and locale peers) with keys for the three bind-outcome messages.
2. Wiring `fl!()` into the crate (or delegating to a shared runtime Fluent handle if one is accessible).
3. Replacing the bare literals with the `fl!()` calls.

I acknowledge this is more than a line-level fix, and that similar bare-literal messages exist in `zeroclaw-gateway` (pre-existing technical debt). That does not justify adding new violations. The bind reply path is a natural seam to establish the `zeroclaw-channels` Fluent baseline.

---

### 🟡 Warning — PR body still describes pre-gate loading indicator ordering (seconding singlerider + tidux)

The body Summary and Feature Details §1 still read:

> "Added Loading Indicator — fires `POST /v2/bot/chat/loading/start` **immediately on every incoming message before any policy gate**"

The shipped behavior on this head is the opposite: `send_loading_indicator` fires **only after both the group policy gate and the DM allowlist/pairing gate pass** (line 596, guarded by the comment "Show loading indicator only for messages that passed both policy gates."). The one deliberate exception — the `/bind` pairing handshake at line 495 — is not mentioned in the body.

This is the squash-merge commit message. Landing a description that says "before any policy gate" when the code does the opposite makes future debugging harder and misrepresents the security rationale.

**Request**: update Summary §2 and Feature Details §1 prose to reflect post-gate ordering, with a note that the bind path is the one authorized pre-agent exception. Also update the label/risk snapshot in the body (`type: feature, risk: low, size: M` → `enhancement, risk: medium, size: L`).

---

### 🟡 Warning — Rollback section has two gaps for `risk: medium`

1. **Stale SHA**: the rollback reads `git revert c3cabcc20`, which is the first commit of the original PR branch head — not the current squashable range. After a squash-merge, the correct target would be the single squash commit on master. The rollback should either reference the squash SHA (post-merge) or note that the full revert is `git revert <squash-sha>` once landed.
2. **Missing observable failure symptoms**: for `risk: medium`, the rollback section must include a symptoms sub-bullet. Example: *"Observable: LINE users receive no inline confirmation after `/bind` attempts; loading animation no longer appears for authorized senders; outgoing messages lose the sender object overlay (`AI from 'BotName'` label disappears in LINE UI). No data loss; no config migration needed."*
3. **Risk mismatch in prose**: the rollback text says "Risk is low" but the live label is `risk: medium`.

---

### 🔵 Suggestion — Non-2xx responses from LINE silently discarded (seconding tidux)

`send_loading_indicator` (line 122–143) and `send_bind_reply` (line 144–170) only log on transport error:

```rust
if let Err(e) = client.post(...).send().await {
    // warn log
}
```

A LINE 403 (revoked token), 429 (rate-limited loading endpoint), or 500 returns `Ok(Response)` and is silently dropped. For `send_bind_reply` in particular, a silent LINE-side error means the user gets no feedback on their bind attempt — the opposite of what this feature intends.

Consider calling `.error_for_status()` or checking `response.status().is_success()` and logging a `WARN` with the status code. The fire-and-forget contract for loading indicator can stay; the bind reply path benefits from at least a logged non-success status.

---

### 🔵 Suggestion — `sender.name = "AI"` is hardcoded and not configurable

`*self.sender_name.write() = "AI".to_string()` in `listen()` (line 1145) fixes the sender display name to "AI" for all LINE deployments. Operators who want "Bot", "Assistant", or a localized equivalent cannot override this without a code change. Since this feature is specifically the "icon/nickname switch," a `sender_name` field in `LineConfig` (optional, defaulting to `"AI"`) would make this operator-configurable at no structural cost and avoids a future code change.

This is not a SSOT concern — the config schema currently has no `sender_name` field, so there is no duplicate state. This is purely a UX/extensibility suggestion.

---

### 🟢 Praise — Policy gate ordering is correct and well-tested

`webhook_handler` at lines 444–585 runs the group policy gate (447–469) and DM policy gate (471–585) before the post-gate `send_loading_indicator` call at 596–602. The comment names the contract. The sole pre-agent loading call (line 495) is inside `LineDmPolicy::Pairing` — the authorization handshake itself, not a dropped unauthorized path. The amplification vector singlerider flagged on `c3cabcc20` is closed. `webhook_loading_indicator_failure_does_not_break_message_delivery` keeps the fire-and-forget guarantee pinned.

---

### 🟢 Praise — Excellent test coverage (15 new tests)

All three new behaviors are thoroughly exercised:
- `build_sender_obj` edge cases (empty name → None, icon optional)
- `send_reply` / `send_push` sender object injection via wiremock
- Loading indicator DM (`userId`) vs group (`groupId`) routing
- Loading indicator transport failure does not drop message
- Bind feedback all three outcomes (success / invalid code / rate-limited)
- `BotInfo` serde (picture_url present and absent)
- `listen_sets_sender_name_to_ai` drives real `listen()` against mocked `/v2/bot/info`

---

### 🟢 Praise — `api_base_url` fix in `send_bind_reply`

`line.rs:144–170` constructs the URL from the `api_base_url` parameter, threaded through `LineState.api_base_url` at every bind-reply call site (531, 552, 573). The hardcoded `https://api.line.me` that was escaping test mocks is gone.

---

### ✅ SSOT — No config duplication

`sender_name: Arc<RwLock<String>>` and `sender_icon: Arc<RwLock<Option<String>>>` in `LineChannel` do NOT duplicate any `LineConfig` schema field. `LineConfig` in `crates/zeroclaw-config/src/schema.rs` has no `sender_name` or `icon_url` field. `sender_name = "AI"` is a hardcoded design constant (not from config); `sender_icon` is sourced from LINE's own `/v2/bot/info` API. No SSOT violation.

---

### ✅ Security — No token or message-content exposure through logs

`send_bind_reply` passes `channel_access_token` as a bearer header (not logged). The `replyToken` extracted from the webhook event is passed to LINE's Reply API (one-use, already consumed by LINE upon use). Neither appears in log messages. The bind outcome log events at lines 524–573 log only `user_id`, not the token values.

---

All three commits authored by humans: Kananek Thongkam (2 commits), Dan Gilles (merge commit). No `Co-Authored-By: <AI>` trailers.

---

## Summary of required actions before merge

| # | Item | Severity | Who |
|---|------|----------|-----|
| 1 | Replace three bare bind-reply English literals with `fl!()/Fluent` keys (requires establishing Fluent infrastructure in `zeroclaw-channels`) | 🔴 Blocking | dvgamerr |
| 2 | Sync PR body: loading indicator ordering, label/risk snapshot | 🟡 Warning | dvgamerr |
| 3 | Fix rollback: stale SHA, missing symptoms, risk mismatch | 🟡 Warning | dvgamerr |
| 4 | (Optional) Log non-2xx from `send_bind_reply` | 🔵 Suggestion | dvgamerr |
| 5 | (Optional) Make `sender.name` configurable in `LineConfig` | 🔵 Suggestion | dvgamerr |

WareWolf-MoonWall (approved):
# Review — PR #7768 `feat(line): loading indicator, icon/nickname switch, and bind reply feedback`

**Re-review (Round 25)** — prior CHANGES_REQUESTED posted at head `ce4631cf52a55597695f5864c631f3b0d76d54c6`  
**Head reviewed**: `e054a420f2cca9837fadc5c8319f8daca3e6bc92`  
**CI**: ✓ 18/18 gates pass  
**Active blocking reviews**: None — singlerider APPROVED at `ce4631cf52`; tidux APPROVED at `ce4631cf52`; Audacity88 COMMENTED (non-blocking); no other CHANGES_REQUESTED active.  
**Verdict**: `--approve` — all three prior blockers resolved; no new blockers found.

---

### ✅ Resolved — Bind reply strings now route through Fluent i18n

All three bare English literals have been replaced with `fl!()`/Fluent keys. The three keys are defined in the `channel-*` section of `crates/zeroclaw-runtime/locales/en/cli.ftl` alongside other channel-facing strings (`channel-discord-*`, `channel-whatsapp-*`):

```
+channel-line-bind-success = ✅ Paired! You can now chat.
+channel-line-bind-invalid-code = ❌ Invalid code. Please try again.
+channel-line-bind-rate-limited = ⏳ Too many attempts. Retry in { $secs }s.
```

All three `send_bind_reply` call sites now use `i18n::get_required_cli_string` / `i18n::get_required_cli_string_with_args` from `zeroclaw_runtime::i18n`. No bare English string reaches the LINE chat bubble. The approach (delegating to `zeroclaw-runtime`'s Fluent handle rather than standing up a separate `zeroclaw-channels/locales/`) is consistent with the existing cross-crate Fluent pattern. CI passes including Lint, confirming the dependency is clean.

---

### ✅ Resolved — PR body now correctly describes post-gate loading indicator ordering

The Summary bullet and Feature Details §1 have been updated. The body now reads "fires `POST /v2/bot/chat/loading/start` **after both the group policy gate and DM allowlist/pairing gate pass**" and explicitly calls out the `/bind` path as the one authorized pre-agent exception. The prior description ("before any policy gate") is gone.

---

### ✅ Resolved — Rollback section updated with correct form and observable failure symptoms

The rollback now reads `git revert <squash-sha>` (correct post-squash form — no stale branch SHA), states "Risk: medium" consistently, and enumerates three observable failure symptoms:
- LINE users receive no inline confirmation after `/bind` attempts
- Loading animation no longer appears for authorized senders
- Outgoing messages lose the sender object overlay

All three gaps from the prior review are closed.

---

### 🟢 Prior 🔵 suggestions both shipped

Both optional suggestions from Round 24 have been implemented in this round:

**`sender_name` is now operator-configurable**: `LineConfig.sender_name: Option<String>` added to `crates/zeroclaw-config/src/schema.rs` with `#[serde(default)]`. `from_config` seeds `LineChannel.sender_name` from it when non-empty; `listen()` falls back to `"AI"` when no name was configured. Backward-compatible — existing configs without the field continue to use `"AI"`.

**Non-2xx from `send_bind_reply` now logged**: The revised implementation checks `resp.status().is_success()` and logs `WARN` with `{"status": <code>}` on non-2xx. The fire-and-forget contract for loading indicator is unchanged.

---

### 🟢 Commit history clean; no AI trailers

All four commits are authored by Kananek Thongkam (dvgamerr) or are a merge commit. No `Co-Authored-By: <AI>` trailers in any commit message.

---

### 🔵 Note — singlerider/tidux approvals predate current head by one commit

singlerider and tidux both approved at `ce4631cf52`. The 4th commit (`e054a420`) — "fix(line): replace bare bind-reply literals with Fluent i18n keys" — was added after their approvals. GitHub shows both as still Approved; branch protection has not auto-dismissed them. The 4th commit is a non-behavioral-security change (i18n key wiring only), and both reviewers had already reviewed the same binding logic at `ce4631cf52`. No governance concern, but noting for the squash operator's awareness.

---

### Summary

| # | Prior item | Status |
|---|-----------|--------|
| 1 | 🔴 Bare bind-reply literals — must use `fl!()`/Fluent | ✅ Resolved |
| 2 | 🟡 PR body described pre-gate loading indicator ordering | ✅ Resolved |
| 3 | 🟡 Rollback stale SHA, missing symptoms, risk mismatch | ✅ Resolved |
| 4 | 🔵 Log non-2xx from `send_bind_reply` | ✅ Shipped |
| 5 | 🔵 Make `sender.name` configurable in `LineConfig` | ✅ Shipped |

Audacity88 (changes requested):
### 🔴 Blocking — `sender_name` now duplicates config state in the channel handle

The new `LineConfig.sender_name` field is copied into `LineChannel.sender_name` during construction (`from_config()` reads `config.sender_name` and writes the value into the channel's `RwLock`). After this PR makes the display name operator-configurable, that `RwLock` is no longer just transient LINE bot metadata; it becomes a second stored copy of a canonical config value. That violates the repo's SSOT boundary for channel/config state, and it can drift from the config handle used elsewhere in the daemon.

`AGENTS.md` is explicit that if a fact already lives somewhere in the codebase, a PR must not copy it into "a new field, struct, config block, schema entry, runtime cache, or anywhere else"; it must resolve from the source on demand. It also lists resolver closures over `Arc<RwLock<Config>>`, `&Config` parameters, and per-call materialized views as the allowed patterns. Please keep the configured sender name canonical at the config source and resolve it when building outbound messages, while leaving truly remote runtime metadata such as `sender_icon` as fetched state. The fallback `"AI"` can still live in the send/build path for the unset or empty config case, but the configured value should not be snapshotted into `LineChannel.sender_name`.

### 🟡 Warning — the PR body still overclaims loading-indicator non-2xx logging

The current PR body says the loading-indicator call is fire-and-forget and that "a non-2xx from LINE logs at `WARN`." The implementation of `send_loading_indicator()` only logs transport errors from `.send().await`; it does not inspect the response status. The non-2xx `WARN` handling added in this revision is for `send_bind_reply()`, not the loading endpoint.

Please either add the status check to `send_loading_indicator()` too, or narrow the PR body so it only promises the behavior the loading endpoint actually has.

### 🟡 Warning — the LINE config roundtrip test no longer covers every user-facing field

`line_config_toml_roundtrip()` still says the sample `[channels_config.line.default]` block "covers every user-facing field", but the new `sender_name` config field is absent from both the roundtrip sample and the defaults assertions. If the config field stays, please add `sender_name` coverage there so the schema/config surface change is protected by the config tests too.

The earlier loading-indicator authorization ordering, bind reply localization, and `api_base_url` fixes all look good. This remaining request is about keeping the new config surface inside the repo's canonical-state rules before merge.

singlerider:
@dvgamerr Please address requested changes before another review.

Audacity88 (changes requested):
@dvgamerr I re-reviewed the current head e9ad21e22ec6e2a3a368ad6812cf9859edbe6c38 against the prior review thread, the live PR body/labels/checks, the four-file diff, and LINE's icon/nickname switch API contract.

### ✅ Resolved — `sender_name` no longer snapshots config into the channel handle

The July 2 update fixes the source-of-truth issue from my prior review. `LineChannel` now holds a `sender_name_resolver` closure instead of a copied `RwLock<String>`, and the orchestrator closures read `[channels.line.<alias>].sender_name` from the live config at send time. That keeps the configured display name canonical in config, while `sender_icon` remains separate runtime metadata fetched from LINE's `/v2/bot/info` endpoint.

The earlier non-2xx logging and config coverage warnings also look addressed: `send_loading_indicator()` now logs non-success statuses, and `line_config_toml_roundtrip()` plus `line_config_defaults()` cover `sender_name`.

### 🟢 What looks good — Size and efficiency tradeoffs are reasonable

Size-wise this is still a fairly large LINE-channel patch, but the added surface is mostly focused tests around the three user-visible behaviors. The send-time resolver is also the right efficiency and ownership tradeoff here: one config read per outbound send is acceptable to preserve live config as the source of truth.

### 🔴 Blocking — Configured `sender.name` can exceed LINE's API limit and break every outbound send

The new `LineConfig.sender_name` is a free-form `Option<String>`, and the send paths pass it straight into the LINE `sender.name` field:

```rust
let sender_name = (self.sender_name_resolver)()
    .filter(|s| !s.is_empty())
    .unwrap_or_else(|| "AI".to_string());
```

The LINE Messaging API reference caps `sender.name` at 20 characters and notes that some words such as `LINE` may not be used. If an operator configures a longer value, `send_reply()` and `send_push()` build an invalid message object and then bail on the non-2xx Reply/Push response. That would make normal LINE replies fail until the config is changed.

The local comment above `build_sender_obj()` says the name is already pre-truncated at `listen()` time, but the current `listen()` path only stores `sender_icon`; it no longer normalizes the display name. Please enforce the LINE `sender.name` contract before sending. A small send-time normalizer would preserve the live-config resolver and hot-reload behavior; alternatively, reject invalid config where config changes are applied. Either way, please add a regression for an overlong configured sender name so this cannot silently turn into LINE 400s.

Reference: https://developers.line.biz/en/reference/messaging-api/#icon-nickname-switch

### 🟡 Warning — Bind outcome replies bypass the new sender object

The PR body says every outgoing Reply API and Push API message carries the icon/nickname `sender` object, but the new `send_bind_reply()` helper builds plain text messages without `sender`:

```rust
"messages": [{"type": "text", "text": text}],
```

Those bind success/invalid/rate-limit replies are user-visible Reply API messages introduced by this PR, so today they render with the default Official Account identity while normal agent replies render with the configured sender. Please either thread the same sender builder into bind replies, or narrow the PR body to make clear that the sender customization applies only to normal agent reply/push sends and not to bind-feedback messages.

### 🟡 Warning — PR body evidence is stale after the latest sender-name rewrite

The body still says empty `sender_name` omits the `sender` object, but the current send path treats unset or empty names as the `"AI"` fallback. The validation table also lists older test names/contracts such as `send_reply_omits_sender_when_name_empty` and `listen_sets_sender_name_to_ai`; the current diff now has `send_reply_uses_ai_fallback_when_no_resolver` and `listen_sets_sender_icon_from_bot_info`.

That record drift matters here because the PR body is the squash-merge record for a new operator-facing config field. Please sync the body with the current behavior after the code fix lands.

singlerider:
@Audacity88 pushed a maintainer commit (9f672e7) addressing your 2026-07-03 review:

🔴 **Blocking (overlong `sender.name`)**: the resolved name is now normalized to LINE's 20-character `sender.name` limit inside `build_sender_obj`, the single choke point for both `send_reply` and `send_push`. Truncation counts characters so a multibyte display name can't split a codepoint. Added two regressions: `build_sender_obj_truncates_overlong_name_to_line_limit` and `build_sender_obj_truncates_multibyte_name_on_char_boundary`. The stale "pre-truncated at listen() time" comment is gone; normalization now actually happens at build time.

🟡 **Bind replies bypass the sender object**: `LineState` now carries the send-time sender resolver and the fetched icon handle, and all three bind outcome replies (success / invalid / rate-limited) attach the same `sender` object as normal agent replies.

🟡 **Body drift**: synced the PR body — the "removed unused sender_name field" claim is replaced with the resolver/icon threading note, bind replies are documented as carrying the sender object, the 20-char normalization is called out, and the validation table now matches the current test names (`send_reply_uses_ai_fallback_when_no_resolver`, `listen_sets_sender_icon_from_bot_info`, plus the two new truncation tests).

Local: `cargo fmt`, `clippy -D warnings`, and `cargo test -p zeroclaw-channels --features channel-line` all green (70 LINE tests pass). Please re-review, and if satisfied, approve and merge.

Audacity88 (approved):
@dvgamerr Re-reviewed current head `9f672e7` against my prior `sender_name`/bind-reply review, the LINE channel diff, prior reviews, live CI, and the PR body. I did not run local Cargo or a live LINE smoke in this review pass.

### ✅ Resolved — `sender_name` now stays config-canonical and is normalized before send

The source-of-truth issue from my prior review is resolved. `LineChannel` now stores a send-time `sender_name_resolver` instead of a copied config value, and the orchestrator closures read `[channels.line.<alias>].sender_name` from the live config handle. That preserves hot-reload behavior and keeps config as the canonical owner.

The overlong-name blocker is also addressed at the right choke point. `build_sender_obj()` normalizes the resolved name to LINE's 20-character `sender.name` limit before Reply API, Push API, or bind-reply messages use it, and the new ASCII/multibyte regressions cover the important failure mode without splitting a codepoint.

### ✅ Resolved — bind outcome replies now carry the same sender object

The bind feedback path now builds the sender object from the same resolver and fetched icon handle used by normal outbound sends, then passes it into `send_bind_reply()` for success, invalid-code, and rate-limit replies. That closes the gap where `/bind` replies rendered with the default Official Account identity while normal agent replies used the configured LINE sender overlay.

### 🟡 Warning — a couple of boundary assertions would make this harder to regress

The implementation looks correct, and the PR body includes live LINE smoke for the visible behavior. The remaining gap is test precision: the bind-reply tests assert reply token/text but not `messages[0].sender`, and the loading-indicator tests cover authorized DM/group plus failure-is-fire-and-forget but not an unauthorized negative case proving no loading request is sent before a dropped message.

I am not blocking on that because the current source path is direct, the prior blockers are resolved, and CI is green. It would still be worth adding those two assertions in a follow-up or before merge if someone is already touching this test block.

### 🟢 What looks good — the LINE policy ordering is aligned

The normal loading-indicator call remains after the group/DM policy gates, with `/bind` as the explicit pairing-handshake exception. The remaining PR-body metadata cleanup is maintainer-side: add the orchestrator resolver wiring to the scope boundary, align the label snapshot with the live labels, and update the empty-name design row. That cleanup does not change my technical approval.

Issue #6944, opened 2026-05-26
[Bug]: Interactive mode — [system] log lines printed inline, drowning out conversation output


NiuBlibing:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

When running `zeroclaw agent -a default` in interactive (TTY) mode, structured `[system]` `INFO`/`WARN` log lines are emitted to stdout interleaved with conversation output. The model's reply and the log lines appear on the same stream, making the actual chat messages very hard to read. Log events emitted during a single turn include at minimum: `llm_request`, `llm_response`, `turn_final_response`, and a cost-tracking `WARN` for any model without a pricing entry.

### Expected behavior

In interactive TTY mode the conversation text (user prompts and model replies) should be clearly readable without inline log noise. Logs should either:
- be suppressed in interactive mode by default (opt-in via `--verbose` or env var), or
- be written to a separate file / stderr-only stream, or
- be filtered to `WARN` and above by default, with `INFO` lines hidden unless `RUST_LOG` / `--log-level` is explicitly set.

### Steps to reproduce

```bash
zeroclaw agent -a default
# at the > prompt, type: hi
# Observe [system] INFO/WARN lines printed inline with the reply
```

### Impact

Affected users: all interactive CLI users
Frequency: always (every turn)
Consequence: Model replies are buried in structured log output; makes interactive mode uncomfortable to use for day-to-day conversation.

**Note — TUI relationship:** If a dedicated TUI mode is introduced and eventually fully replaces the `agent` interactive command, this issue may become moot. It should be revisited / closed once the TUI is the canonical interactive entry point.

### Logs / stack traces

```text
[system] 2026-05-26T02:52:17.196541Z  INFO zeroclaw_log_event: zc_name=zeroclaw_config::schema zc_action=note zc_outcome=unknown zc_category= zc_attrs={"initialized":true,"path":"/home/user/.zeroclaw/config.toml","source":"default","workspace":"/home/user/.zeroclaw/data"} zc_has_duration=false zc_duration_ms=0 zc_file=crates/zeroclaw-config/src/schema.rs zc_line=13721 Config loaded
[system] 2026-05-26T02:52:17.199493Z  INFO zeroclaw_scope{...}: zeroclaw_log_event: zc_name=zeroclaw_runtime::agent::loop_ zc_action=note ... zc_line=2941 Memory initialized
[...]

PR #5958, opened 2026-04-21
fix(channels/ACP): suppress INFO logs, implement missing parts of ACP spec protocol

tidux:
## Summary

- **Base branch:** `master`
- **What changed and why:**
- suppress "INFO" logging on ACP server to stop injecting it into event stream
- ACP server was broken, fixed it so it works with agentic.nvim.
- **Scope boundary:** 
- channels other than ACP are untouched
- **Blast radius:**
- if the logging change is configured incorrectly INFO logging could be dropped for things besides Matrix and ACP.
- **Linked issue(s):** 
- solves #5948 
- solves #5949
## Validation Evidence (required)

Local validation is the signal CI cannot replace. Run the full battery and paste literal output (tails, failures, warnings — not "all passed").

```bash
$ cargo fmt --all -- --check
<no output, all passed>
$ cargo clippy --all-targets -- -D warnings | tail -n9
    Checking zeroclaw-config v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-config)
    Checking zeroclaw-memory v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-memory)
    Checking zeroclaw-providers v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-providers)
    Checking zeroclaw-tui v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-tui)
    Checking zeroclaw-tools v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-tools)
    Checking zeroclaw-runtime v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-runtime)
    Checking zeroclaw-hardware v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-hardware)
    Checking zeroclaw-channels v0.7.3 (/Users/jlane/src/zeroclaw/crates/zeroclaw-channels)
    Finished [`dev` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 53.43s
$ cargo test
Finished [`test` profile [unoptimized + debuginfo]](https://doc.rust-lang.org/cargo/reference/profiles.html#default-profiles) target(s) in 1m 11s
     Running unittests src/lib.rs (target/debug/deps/zeroclaw-2c7956b0c31c0a91)
[...]

PR #4727, opened 2026-03-26
fix: write tracing logs to stderr instead of stdout

rareba:
## Summary

One-line fix: add `.with_writer(std::io::stderr)` to the tracing subscriber builder so logs go to stderr instead of stdout.

## Problem

Logs pollute stdout, breaking piped commands:
```bash
zeroclaw config schema | jq  # fails: log lines mixed into JSON
```

## Fix

```rust
let subscriber = fmt::Subscriber::builder()
    .with_writer(std::io::stderr)  // ← added
    .with_env_filter(...)
    .finish();
```

## Files changed

- `src/main.rs` — add `.with_writer(std::io::stderr)` to subscriber builder

## Test plan

- [ ] `zeroclaw config schema 2>/dev/null | jq .` parses cleanly
- [ ] Logs still visible on stderr: `zeroclaw daemon 2>&1 | grep INFO`

Closes #4721

Issue #4721, opened 2026-03-26
zeroclaw should log to stderr instead of stdout


mikeyhew:
I was surprised to see that zeroclaw logs to stdout instead of stderr. This makes it hard to use the output of commands like `zeroclaw config schema` becaues it includes logs in the output:

```bash
zeroclaw config schema 2>/dev/null | head
2026-03-26T02:06:56.927469Z  INFO zeroclaw::config::schema: Config loaded path=/home/zeroclaw/.zeroclaw/config.toml workspace=/home/zeroclaw/.zeroclaw/workspace source="default" initialized=true
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "Config",
  "description": "Top-level ZeroClaw configuration, loaded from `config.toml`.\n\nResolution order: `ZEROCLAW_WORKSPACE` env → `active_workspace.toml` marker → `~/.zeroclaw/config.toml`.",
  "type": "object",
  "properties": {
    "agent": {
      "description": "Agent orchestration settings (`[agent]`).",
      "$ref": "#/$defs/AgentConfig",
```

If you pass the output of `zeroclaw config schema` to `jq`, it will fail to parse. Of course, you can change to a higher log level with `RUST_LOG=fatal zeroclaw config schema` but logging to stderr is a best practice and lets you still see logs while piping stdout to another process.

I'm happy to open a PR to change the logging to stderr if you agree. It's as simple as adding a line to change the output from the default of `io::stdout` to `io::stderr` here:

https://github.com/zeroclaw-labs/zeroclaw/blob/be1d9c846baa3cc6bb251e2da33a8186fc173d70/src/main.rs#L849-L854

You just need to add this to the method call chain:

```rust
    .with_writer(std::io::stderr)
```

freeekanayaka:
It's not only a problem for command lines tools, but also for the daemon: at the moment some information is printed directly to stdout using `println!`, that information includes for example the content of LLM [responses](https://github.com/zeroclaw-labs/zeroclaw/blob/9f0de18b606dfbafbc6f24e6d6c286e10dea5f6e/crates/zeroclaw-runtime/src/agent/loop_.rs#L2569).

I'm not sure if those `println!` calls are intended or just oversights/leftovers that should be replaced by logger calls (`tracing::info`), but  supporting stderr log redirection seems like good thing regardless.

Audacity88:
## Triage Summary

**Type:** Bug
**Priority:** Medium
[...]

PR #7773, opened 2026-06-16
fix(runtime): route native tool narration to stderr

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:** Moved non-silent native-tool intermediate narration emitted during `Agent::turn` from stdout to stderr. Added a subprocess-boundary regression that proves native-tool intermediate narration is absent from process stdout and present on stderr. Preserved deliberate stdout surfaces for final command output, interactive final answers, and stdout protocol streams.
- **Scope boundary:** This PR does not reroute user-facing command output, ACP stdout protocol traffic, interactive prompts/final responses, doctor/service/status output, or the broader CLI stdout audit.
- **Blast radius:** Runtime native-tool turn narration only. Stdout consumers should stop seeing this intermediate text, while stderr still receives it for operator visibility. The PR is labeled `risk: high` because it touches runtime process-output routing.
- **Related issues/PRs:** Related #4721, #4727, #5958, #6944.
- **Labels:** `bug`, `risk: high`, `size: S`, `runtime`, `runtime:core`, `agent`

- **Commands run and tail output:** Local validation:

```text
cargo fmt --all -- --check
result: passed

cargo test -p zeroclaw-runtime direct_agent_
result: passed
summary: 2 passed; 0 failed; 1 ignored helper

result: passed
tail: Finished `dev` profile ... in 19m35s

result: passed
summary: 8720 tests run: 8720 passed, 11 skipped

git diff --check
result: passed
```

- **Beyond CI — what did you manually verify?** The regression runs the current test binary as a subprocess and checks the process boundary: intermediate native-tool narration is not in stdout and is present in stderr. Source review kept final command output, interactive final answers, and stdout protocol streams unchanged.

## Compatibility (required)

- Backward compatible? Yes
- If `No` or `Yes` to either: No upgrade steps. The change only moves native-tool intermediate narration from stdout to stderr.

- **Observable failure symptoms:** Missing native-tool intermediate narration in stderr, intermediate native-tool text reappearing in stdout, or downstream stdout parsers receiving narration before final output.

N/A

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/agent.rs
+++ b/crates/zeroclaw-runtime/src/agent/agent.rs
@@ -3057,6 +3057,90 @@ mod tests {
         }
     }
 
+    #[test]
+    fn direct_agent_turn_does_not_write_intermediate_native_text_to_stdout() {
+        let current_exe = std::env::current_exe().expect("current test binary path");
+        let output = std::process::Command::new(current_exe)
+            .args([
+                "direct_agent_turn_stdout_boundary_helper_4721",
+                "--ignored",
+                "--nocapture",
+            ])
+            .output()
+            .expect("helper test process should run");
+
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let stderr = String::from_utf8_lossy(&output.stderr);
+
+        assert!(
+            output.status.success(),
+            "helper failed with status {:?}\nstdout:\n{}\nstderr:\n{}",
+            output.status.code(),
+            stdout,
+            stderr
+        );
+        assert!(
+            !stdout.contains("intermediate native narration"),
+            "intermediate native narration leaked to stdout:\n{stdout}"
+        );
+        assert!(
+            stderr.contains("intermediate native narration"),
+            "intermediate native narration was not routed to stderr:\n{stderr}"
+        );
+    }
+
+    #[tokio::test]
+    #[ignore = "subprocess helper for stdout/stderr boundary regression"]
+    async fn direct_agent_turn_stdout_boundary_helper_4721() {
+        let memory_cfg = zeroclaw_config::schema::MemoryConfig {
+            backend: "none".into(),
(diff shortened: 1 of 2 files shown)


singlerider (approved):
No prior reviews. The real change is two lines in `run_tool_call_loop`: the non-silent native-tool narration emit moves from `print!` + `stdout().flush()` to `eprint!` + `stderr().flush()`. The other 84 lines are a subprocess-boundary regression test. I read the surrounding code on the PR head to confirm this is the correct branch, and it is.

### 🟢 What looks good - the rerouted branch is genuinely intermediate narration

This emit sits after the final-turn early return (`return Ok(accumulated_display_text)`), inside the block whose own comment states native tool-call providers may emit narration or scratchpad text alongside tool calls and that it must not land in the final delivered response. It only fires when `!native_tool_calls.is_empty()`. So moving it to stderr does not touch the final assistant answer, the final command output, or the ACP/stdout protocol streams. The scope boundary in the PR body matches what the diff actually does.

### 🟢 What looks good - draft-channel relay is preserved

The `on_delta` relay of the same narration to draft-capable channels is left intact directly above the changed lines, so live observers still see the text. Only the raw process-stdout copy moves to stderr. Operators keep visibility, machine stdout consumers stop getting polluted. That is the right split for #4721.

### 🟢 What looks good - the regression is a real boundary test, not an in-process capture

`direct_agent_turn_does_not_write_intermediate_native_text_to_stdout` re-invokes the test binary as a subprocess and asserts on the child's actual stdout and stderr, rather than swapping a writer in-process. For a fix about process-level output routing, asserting at the real fd boundary is the only way to prove it. The mock provider returns narration plus a tool call on turn one and a clean final answer on turn two, which exercises exactly the intermediate-then-final path the fix cares about, and the helper is correctly `#[ignore]`d so it only runs when the parent drives it.

PR #7775, opened 2026-06-16
fix(runtime): keep pruned tool markers out of live context

Audacity88:
## Summary

- **What changed and why:**
  - Added shared `ChatMessage` helpers for history-pruner marker construction and detection, so runtime and provider code use one sentinel contract.
  - Capped old synthetic tool-exchange summaries before adjacent summaries merge, so long tool-heavy histories cannot send a wall of collapsed-summary text back to the model.
  - Changed pruning cleanup to drop synthetic summaries before live assistant/tool-call groups instead of inserting a fake user separator.
  - Made Anthropic conversion hide only the synthetic summary + separator pair, while preserving real separators that still protect Anthropic role alternation.
- **Scope boundary:** This PR does not change the history-pruner token budget, `keep_recent` policy, provider retry behavior, tool execution, or non-Anthropic provider conversion.
- **Blast radius:** Runtime history pruning and Anthropic request conversion. Other providers still receive the bounded synthetic summaries, but the shared marker helpers reduce drift if they need matching behavior later.
- **Linked issue(s):** None.
- **Labels:** `bug`, `risk: high`, `size: M`, `agent`, `runtime`, `provider`, `provider:anthropic`

## Relationship to #7684

#7684 and this PR are complementary. #7684 is display and observability work: it renames the history-pruner marker, renders that marker as a visible `history-pruner` ACP event, and surfaces client-RPC cancels. This PR changes provider-bound history behavior: it bounds synthetic pruning summaries, drops synthetic summaries next to live assistant/tool-call state instead of inventing a fake user turn, and filters the internal marker pair out of Anthropic requests while preserving real role separators.

If the two PRs are stacked, `crates/zeroclaw-runtime/src/agent/history_pruner.rs` has a small manual conflict. The intended resolution is to keep #7684's newer marker wording/constant as the canonical marker and keep this PR's cap, drop, and provider-filter behavior.

```text
cargo test -p zeroclaw-runtime history_pruner -- --nocapture
Result: PASS — 30 passed; 0 failed; 0 ignored; 2177 filtered out

cargo test -p zeroclaw-providers convert_messages_ -- --nocapture
Result: PASS — 36 passed; 0 failed; 0 ignored; 930 filtered out

cargo fmt --all -- --check
Result: PASS

git diff --check
Result: PASS

zc-cargo-broad --lane workspace-check --batch pr-7775-pruned-tool-markers --crates zeroclaw-api,zeroclaw-runtime,zeroclaw-providers --prs 7775 clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings
Result: PASS — finished in 7m20s

zc-cargo-broad --lane nextest --batch pr-7775-pruned-tool-markers --crates zeroclaw-api,zeroclaw-runtime,zeroclaw-providers --prs 7775 nextest run --locked --workspace --exclude zeroclaw-desktop
Result: PASS — 8723 passed; 0 failed; 10 skipped
```

- **Beyond CI — what did you manually verify?** I traced the pruner phases after local review: old synthetic summaries are capped before merge, synthetic summaries adjacent to live assistant/tool-call groups are dropped, and Anthropic only filters a context separator when it directly follows a synthetic summary. The new tests cover the suspected long tool-history shape, the merged-summary wall case, Anthropic role alternation, and live `tool_use` / `tool_result` survival.
- **If any command was intentionally skipped, why:** None. Focused runtime/provider regressions, formatting, diff whitespace, workspace clippy, and CI-shaped nextest all passed locally.

## Compatibility (required)

- Backward compatible? `Yes`

- **Observable failure symptoms:** Watch for resumed long-running turns losing useful old tool context, Anthropic request-conversion errors about adjacent roles or missing tool results, or unexpected absence/excess of `[Tool exchange: ... results collapsed]` text in provider-visible history.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR scope was incorporated.

### DIFF
--- a/crates/zeroclaw-providers/src/anthropic.rs
+++ b/crates/zeroclaw-providers/src/anthropic.rs
@@ -1,3 +1,5 @@
+#[cfg(test)]
+use crate::traits::PRUNED_CONTEXT_SEPARATOR;
 use crate::traits::{
     ChatMessage, ChatRequest as ProviderChatRequest, ChatResponse as ProviderChatResponse,
     ModelProvider, ProviderCapabilities, StreamChunk, StreamError, StreamEvent, StreamOptions,
@@ -456,11 +458,28 @@ impl AnthropicModelProvider {
         })
     }
 
+    fn should_skip_internal_pruning_marker(messages: &[ChatMessage], index: usize) -> bool {
+        let Some(msg) = messages.get(index) else {
+            return false;
+        };
+        if msg.is_pruned_tool_exchange_summary() {
+            return true;
+        }
+        msg.is_pruned_context_separator()
+            && index
+                .checked_sub(1)
+                .and_then(|previous| messages.get(previous))
+                .is_some_and(ChatMessage::is_pruned_tool_exchange_summary)
+    }
+
     fn convert_messages(messages: &[ChatMessage]) -> (Option<SystemPrompt>, Vec<NativeMessage>) {
         let mut system_text = None;
         let mut native_messages = Vec::new();
 
-        for msg in messages {
+        for (index, msg) in messages.iter().enumerate() {
+            if Self::should_skip_internal_pruning_marker(messages, index) {
+                continue;
+            }
             match msg.role.as_str() {
                 "system" => {
                     if system_text.is_none() {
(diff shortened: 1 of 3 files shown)


crates/zeroclaw-runtime/src/agent/history_pruner.rs
```diff
@@ -384,22 +406,29 @@ pub fn prune_history(messages: &mut Vec<ChatMessage>, config: &HistoryPrunerConf
...
     while i < messages.len() {
         if messages[i - 1].role == "assistant" && messages[i].role == "assistant" {
-            messages.insert(
-                i,
-                ChatMessage {
-                    role: "user".to_string(),
-                    content: "[context continues]".to_string(),
-                },
-            );
+            if messages[i - 1].is_pruned_tool_exchange_summary() {
```
singlerider, line 420: 🟢 This branch is the invariant the Anthropic adapter quietly depends on. By dropping the summary here instead of inserting a separator next to it, Phase 6 guarantees a `[context continues]` separator never sits adjacent to a summary. That is exactly what makes the adapter's double-skip (summary + following separator) unable to abut two real assistants and break role alternation. Worth knowing this is load-bearing beyond this crate.

crates/zeroclaw-providers/src/anthropic.rs
```diff
@@ -456,11 +458,28 @@ impl AnthropicModelProvider {
...
     }
 
+    fn should_skip_internal_pruning_marker(messages: &[ChatMessage], index: usize) -> bool {
+        let Some(msg) = messages.get(index) else {
+            return false;
+        };
+        if msg.is_pruned_tool_exchange_summary() {
+            return true;
+        }
+        msg.is_pruned_context_separator()
```
singlerider, line 468: 🟢 Correct and tested for every reachable input. One forward-looking note, not blocking: this filter's safety leans on a cross-crate invariant (the runtime pruner never emits a `[context continues]` separator adjacent to a summary). If a future pruner change ever produced `summary -> separator`, this would skip both and could surface two adjacent assistant messages, which Anthropic rejects, and only `convert_messages_no_adjacent_same_role` would catch it (and only if extended to cover the skip path). A one-line comment here pointing at the Phase 6 guarantee would make that dependency explicit for the next maintainer.

singlerider (approved):
Read this against live `history_pruner.rs`, `anthropic.rs`, and the `ChatMessage` surface in `model_provider.rs`, traced the new Phase 3 cap and the reworked Phase 6, and checked role alternation across every reachable shape. Reviewed alongside #7782 (just merged), which removed the dead pre-normalization image trim and its false `trimmed_to` WARN. Same disease as this PR (pruning artifacts leaking into the live provider payload and emitting a misleading signal that the work is already done), different organ: #7782 was the surgical multimodal half, this is the structural runtime-pruner plus Anthropic-adapter half. No file overlap, so the two are complementary, not competing.

### 🟢 The two halves are both load-bearing, not redundant belt-and-suspenders

I went in expecting the pruner-side drop and the adapter-side filter to overlap. They do not. Phase 6 still legitimately emits `pruned_context_separator()` for the genuine case (two real adjacent assistants, neither a summary), so `[context continues]` can still reach a provider. The Anthropic filter is what stops that synthetic separator from being delivered to the model as a real user instruction. Meanwhile the pruner drop handles the summary-adjacent case before it ever leaves the runtime. Distinct cases, both needed.

### 🟢 Marker strings are now typed, killing the duplicated literals

Promoting `[Tool exchange: ...]` / `[context continues]` to constants plus `is_pruned_*` methods on `ChatMessage` removes the prior `starts_with` / `contains` duplication that lived in two crates. This is the right home for them (the shared api crate), and it is what makes the adapter and pruner agree on the same definitions instead of drifting. The only remaining raw literal is a test fixture constructing an input, which is fine.

### 🟢 Cap drops oldest, keeps newest, runs before the merge

`MAX_SYNTHETIC_TOOL_SUMMARIES = 3` with an in-order `retain` drops the oldest summaries and keeps the newest, and running it before the Phase 4 merge means one merged summary cannot hide dozens of collapsed exchanges inside a single provider-visible message. `prune_keeps_synthetic_tool_summaries_from_dominating_long_history` (40 collapsed exchanges, asserts the current user intent survives) and `prune_bounds_summary_lines_before_merging_adjacent_summaries` both pin this. Phase renumbering 3 through 6 is consistent.

### 🟢 Role alternation holds for every reachable input

The adapter skips a `[context continues]` separator only when its predecessor is a summary. Phase 6 guarantees a separator it inserts never sits adjacent to a summary (it drops the summary instead of inserting a separator next to one), so the double-skip can never abut two real assistants. `convert_messages_does_not_surface_context_separator_as_user_instruction` and `convert_messages_keeps_real_context_separator_between_assistants` cover both shapes and assert alternation plus live `tool_use` / `tool_result` survival. One forward-looking note left inline on the adapter: the adapter's safety leans on an invariant maintained in a different crate. Not a defect today, but worth a sentence so a future pruner change does not quietly break it.

CI is fully green across all targets. risk:high is the right label for the blast radius, and the rollback plan is sound. Approving.

PR #7778, opened 2026-06-16
fix(agent): emit pending tool_call at dispatch so live cards render

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `execute_one_tool` now emits the pending `TurnEvent::ToolCall` at the moment of dispatch, before the tool future runs and may block, then emits the terminal `ToolResult` once the outcome is known. Previously both were emitted as one post-execution pair by `emit_tool_call_pair`, which required the outcome as an argument, so the pending card could not be sent until the tool had already returned. On a fast tool the pending and result landed within one frame and the card looked live; on a long-running tool the turn sat visibly idle with no card at all until the result arrived, indistinguishable from a hang.
  - The span guard is dropped before each await so the executor futures stay `Send`.
  - The correlation id is resolved once in `prepare_tool_calls` and pinned onto the executable `ParsedToolCall`, so the pending `tool_call` and its `tool_call_update` share one id even for id-less text-protocol calls.
  - A cancel mid-batch returns the in-flight call early without a result; the cancelled-mid-batch block emits a terminal interrupted `ToolResult` for every executable call without an outcome so its card never spins forever.
- **Scope boundary:** does not change the shell tool timeout, autonomy presets, the ACP `session/update` notification shapes, or the cancellation mechanism itself. Calls that never execute (hook-cancelled, denied, replaced, deduplicated) keep the adjacent pending+result pair.
- **Blast radius:** every channel that consumes `TurnEvent` (ACP orchestrator, gateway WebSocket). Serial vs parallel emission ordering is preserved and covered by existing event-sequence tests.
- **Linked issue(s):** none.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-runtime -p zeroclaw-channels --all-targets -- -D warnings
cargo test -p zeroclaw-channels --lib acp
```

  - `cargo fmt --all -- --check` -> exit 0, no diff.
  - `cargo clippy -p zeroclaw-runtime -p zeroclaw-channels --all-targets -- -D warnings` -> `Finished`, no warnings.
  - `cargo test -p zeroclaw-runtime --lib` -> `test result: ok. 2198 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out`.
  - `cargo test -p zeroclaw-channels --lib acp` -> `test result: ok. 66 passed; 0 failed; 0 ignored; 0 measured; 690 filtered out`.
- **Beyond CI — what did you manually verify?** Serial dispatch still interleaves call->result per tool (`turn_streamed_dispatches_multiple_tools_serially_when_parallel_disabled`), the mid-batch cancel sequence (`safety_net_streaming_tool_results_input_order_and_midbatch_cancel`), and the long-tool-gap drain test (`drain_must_not_idle_cancel_a_live_turn_across_a_long_tool_gap`) all pass. Added a unit test asserting the split pending/result share the resolved id.
- **If any command was intentionally skipped, why:** Ran the targeted crates rather than the full workspace `cargo test`; the change is confined to `zeroclaw-runtime` with `zeroclaw-channels` as the consumer. The full workspace battery is left to CI.

## Compatibility (required)

- Backward compatible? `Yes`
- Upgrade steps: none. The change is internal to turn-event emission ordering; the ACP/WS wire shapes are unchanged.

- **Fast rollback command/path:** `git revert <merge sha>`
- **Observable failure symptoms:** tool cards rendering twice, cards never closing (stuck pending), or call/result arriving out of order in ACP/WS clients. Grep traces for a `tool_call` without a matching `tool_call_update` on the same `toolCallId`, or duplicate `tool_call_update` for one id.


singlerider:
This has been good so far. I'm running this HEAD as my current install. No hangs or weird tool calls not showing in the zerocode window so far.

perlowja:
Babysitting Shane's PRs while he's offline — status check on this one:

- **Mergeable**, all CI checks green (Build/Test/Lint/Check matrix all passing as of `f42e55d`-era run).
- No unresolved review threads or Bugbot findings.
- Only thing outstanding is `REVIEW_REQUIRED` — it just needs a reviewer's approval, no code changes.

(Of his three open PRs, only #7684 needed work — it had a real semantic conflict with master's #7775; resolution posted there. #7681 is also green and waiting on re-review.)

Audacity88 (dismissed):
I checked the live PR state, PR body, top-level comments, prior reviews, current diff, and the runtime turn-event/cancellation paths around `run_tool_call_loop`, `execute_tools_parallel`, and `execute_one_tool`. CI is green, and I did not run additional local Cargo validation.

### 🟢 What looks good — the pending/result split is pointed at the right boundary

Moving the pending `ToolCall` emit to dispatch time is the right shape for the live-card problem. Pinning a resolved id onto each executable `ParsedToolCall` in `prepare_tool_calls` is also the important part for id-less text-protocol calls; it keeps the pending card and the later update correlated instead of synthesizing two unrelated ids.

### 🔴 Blocking — parallel cancellation can send a second terminal update for a completed call

The new terminal emit in `execute_one_tool` runs before that future returns its `Ok(ToolExecutionOutcome)`. In the parallel path, `execute_tools_parallel` still does `join_all(futures).await` and then `results.into_iter().collect()`. If one future returns `ToolLoopCancelled`, the `collect()` returns `Err` and discards any other futures that already completed successfully.

`run_tool_call_loop` then handles that error by setting `executed_outcomes` to an empty vec, so `completed == 0`, and the cancel cleanup emits an interrupted `ToolResult` for every executable call. With this PR, any parallel tool that completed before the cancellation already emitted its real terminal `ToolResult`, so the client can receive two terminal updates for the same `toolCallId`: first the real result, then an interrupted result from the cleanup pass.

That can make the live card flip from completed to interrupted, or otherwise leave ACP/WS consumers with a state that does not match what actually ran. It also contradicts the PR body's "without an outcome" intent; the implementation cannot tell which parallel calls already emitted outcomes once `collect()` has collapsed the mixed result set to `Err`.

Please preserve the partial parallel outcomes, or otherwise track which executable calls already emitted a terminal event before synthesizing interrupted results. I think this also needs a regression with `parallel_tools` enabled where one fast tool completes and another in-flight tool observes cancellation, so the event stream proves the completed call gets exactly one terminal update and the unfinished call gets the interrupted update.

Nillth (commented):
I read the full diff, the PR body, and both commits, then cross-checked the diff against the PR head 3741fe260d555d1cf301046c1cddccfbb397b0dd (not my local clone, which is well behind master) by pulling the contents of `tool_execution.rs`, `turn/mod.rs`, `turn/call_prep.rs`, and `turn/events.rs` at that ref. I took stock of the prior review first: @Audacity88's formal review (now DISMISSED) carried a single 🔴 on the parallel-cancellation double-terminal path, submitted 2026-06-17T00:15:36Z against commit b0ff5037eabbf4c6f3e158506c22dbed1d1d7623. The author then pushed 3741fe260d555d1cf301046c1cddccfbb397b0dd ("preserve completed parallel outcomes on mid-batch cancel") at 00:31:34Z and self-dismissed the CHANGES_REQUESTED 56 seconds later, re-requesting @Audacity88. There are no open inline threads and no other formal reviews. This is high-risk runtime per AGENTS.md (`crates/zeroclaw-runtime/src/**`), so I traced the event-correlation and cancellation invariants line by line rather than taking the green CI on faith.

### ✅ Resolved — @Audacity88's parallel-cancel double-terminal blocker is fully addressed

The exact defect @Audacity88 raised — `execute_tools_parallel` collapsing a mixed result set to one `Err` via `results.into_iter().collect()`, dropping siblings that already completed (and already emitted their real terminal `ToolResult`), so the cancel cleanup re-emitted an interrupted result for an already-closed `toolCallId` — is resolved in 3741fe260d555d1cf301046c1cddccfbb397b0dd. `execute_tools_parallel` now returns `Vec<Option<ToolExecutionOutcome>>` (tool_execution.rs:347–360): `Ok(outcome) => Some`, `Err if is_tool_loop_cancelled => None`, non-cancellation `Err` still aborts. Completed siblings are preserved instead of discarded. In `run_tool_call_loop` (turn/mod.rs) the completed calls are tracked by their original index in `executed_completed_indices`, and the cancel-cleanup loop skips any index in that set before emitting an interrupted `ToolResult` — so a parallel call that finished before the cancellation gets exactly one terminal event and its card can no longer flip from completed to interrupted. The regression @Audacity88 explicitly asked for ships too: `run_tool_call_loop_parallel_cancel_no_double_terminal_for_completed_call` (loop_.rs) drives a `parallel_tools`-enabled batch where `fast_tool` completes and signals `cancel_tool` to fire the token and never return, then asserts `call_fast` receives exactly one terminal `ToolResult` (`fast-done`) and `call_cancel` receives exactly one interrupted result. This is the right test at the right layer, and it proves the event stream, not just the history vec.

### 🟢 What looks good — the id is pinned once, so pending and terminal events provably correlate

The load-bearing correctness property of this change is that the pending `ToolCall` and the terminal `ToolResult` (now emitted from two different places and times) share one correlation id, including the cancel-cleanup path. The diff gets this right by resolving the id once at the earliest point: `prepare_tool_calls` calls `resolve_tool_call_id` and pins `tool_call_id: Some(call_id)` onto every executable `ParsedToolCall` (call_prep.rs:221–235). From there every downstream reader sees the same id: `execute_one_tool`'s `event_call_id = tool_call_id_owned.clone().unwrap_or_else(...)` (tool_execution.rs) resolves to that pinned id for both the pending emit and the terminal emit, and the cancel cleanup's `events::resolve_tool_call_id(call)` (turn/mod.rs) returns the *same* pinned id — so the interrupted `emit_tool_result` closes the exact card that was opened, with no orphaned pending card. The `unwrap_or_else(uuid)` branch in `execute_one_tool` is now effectively a defensive fallback (the id is always `Some` for executable calls), which is the correct shape. This is the subtle part of the fix and it holds.

### 🟢 What looks good — the split is pointed at the real boundary, and the never-executed path is left intact

Moving the pending emit to dispatch time in `execute_one_tool` (before `tool.execute(...).await`) is the correct fix for the live-card-on-long-running-tool problem described in the body. The refactor of `events.rs` is clean: `emit_tool_call_pair` is preserved for calls that never reach execution (hook-cancelled, denied, replaced, deduplicated) where there is no live window between the two halves, and `post_exec.rs::record_executed_outcomes` correctly stops re-emitting the pair (the executor now owns emission) and is reduced to history/log/hook recording. The doc comments on `emit_tool_call_pending`/`emit_tool_result` spell out the shared-id contract, and `split_pending_then_result_share_resolved_id` (events.rs) unit-tests it directly. The serial-vs-parallel ordering claim in the body is backed by the cited existing event-sequence tests.

### 🟢 What looks good — no duplicate state introduced (AGENTS.md SSOT)

Given the high-risk scope I checked the single-source-of-truth rule explicitly. The resolved id is computed once and threaded by reference; it is not cached in a second field that could drift. The interrupted-outcome string is read from the Fluent key `turn-tool-interrupted-before-result` (cli.ftl:645) in both the `ordered_results` history synthesis and the new event-layer cleanup, rather than hardcoded — consistent with the localization rule. `event_tx` is plumbed as an `Option<&Sender<TurnEvent>>` parameter all the way down (loop_/turn/tool_execution), not stashed on a struct. No SSOT or duplicate-state concern.

### 🔵 Suggestion — the completed-index lookup is correct but worth a one-line invariant note

In the cancel-cleanup block (turn/mod.rs) the completed set is built from `executed_completed_indices` (the original `tool_calls` indices), while the skip-loop iterates `executable_indices.iter().zip(executable_calls.iter())` and compares against that set. This is correct because both index spaces are `tool_calls` indices, but the two-index-space dance (`executable_indices` vs the position within `executable_calls`) is exactly the kind of thing that bites a future editor. A single comment at the `HashSet` asserting "these are `tool_calls` indices, matching `executable_indices`" would make the invariant explicit. Non-blocking.

### 🔵 Over to @Audacity88 — please re-confirm and clear your own block

I'm deliberately not stacking a formal approval here, and I want to be explicit about why. Your CHANGES_REQUESTED is DISMISSED, so procedurally there's nothing for me to approve over — but it was dismissed by the author 56 seconds after they pushed the fix and re-requested you, not converted by you after seeing it resolved. The blocker was yours; the call that it's satisfied should be yours too. From my own cross-check at head 3741fe260d555d1cf301046c1cddccfbb397b0dd it looks fully resolved — the parallel-cancel double-terminal you identified is gone, and the regression test you asked for ships — so this should be a quick re-confirm. When you have a moment, could you clear or convert your review? I'll approve on the substance immediately behind you.

### Verdict

I'm posting this as a comment rather than an approval, and that's a process choice, not a doubt about the code. The cross-check at head 3741fe260d555d1cf301046c1cddccfbb397b0dd confirms the event-correlation id is pinned once and shared across the pending emit, the terminal emit, and the cancel-cleanup emit; the duplicate pending-then-interrupted terminal class (@Audacity88's blocker) is eliminated by preserving completed parallel outcomes and skipping their indices in cleanup; and the regression that proves it ships in the diff. CI is fully green across the Quality Gate matrix and the PR is mergeable. The one thing I'd rather not do is convert another reviewer's self-dismissed block into a merge-ready approval on their behalf — so I'm leaving the formal nod to @Audacity88 and standing ready to approve the moment they clear it. My only standalone finding is a non-blocking invariant comment.

WareWolf-MoonWall (approved):
I read the full PR body, top-level comments, Audacity88's prior DISMISSED CHANGES_REQUESTED (blocker now resolved), Nillth's COMMENTED analysis (substantive  cross-check confirming the fix holds), the full diff at head 3741fe260d555d1cf301046c1cddccfbb397b0dd, and perlowja's status note confirming the branch is mergeable with all CI green. I cross-checked the diff against the PR head's implementation by reading `tool_execution.rs`, `turn/mod.rs`, `call_prep.rs`, and `turn/events.rs` at that ref. I traced the event-correlation id resolution, the parallel-cancel outcome preservation, and the cancel-cleanup skip logic. Nillth's cross-check matches what I see; Audacity88's blocker is resolved. This is risk: high runtime per AGENTS.md (`crates/zeroclaw-runtime/src/**`), so the following findings are the result of line-by-line scrutiny of the event-emission paths.

### ✅ Resolved — Audacity88's parallel-cancel double-terminal blocker is fixed

Audacity88's blocker (parallel calls that completed before cancellation could receive two terminal `ToolResult` updates — first the real outcome, then a second interrupted result synthesized by the cancel cleanup) is fully addressed at 3741fe260d555d1cf301046c1cddccfbb397b0dd. The fix is elegant and correct: `execute_tools_parallel` now returns `Vec<Option<ToolExecutionOutcome>>` instead of collapsing a mixed result set to a single `Err` — completed calls are `Some(outcome)`, cancelled-in-flight calls are `None`, and non-cancellation errors still abort. In `run_tool_call_loop` (turn/mod.rs) the completed calls are tracked by their original `tool_calls` index in `executed_completed_indices`, and the cancel-cleanup loop skips any index in that set before emitting an interrupted `ToolResult` for the remaining uncompleted calls. This ensures a parallel call that finished before the cancellation gets exactly one terminal event. The regression test Audacity88 explicitly requested (`run_tool_call_loop_parallel_cancel_no_double_terminal_for_completed_call`, loop_.rs) proves the event stream (not just the history vec) is correct: `fast_tool` completes and signals `cancel_tool` to fire the token and never return, then the test asserts `call_fast` receives exactly one terminal `ToolResult` (`fast-done`) and `call_cancel` receives exactly one interrupted result. This is the right test at the right layer.

### 🟢 What looks good — the id is pinned once and shared across every emit path

The core correctness property of this change is that the pending `ToolCall` and the terminal `ToolResult` (now emitted from two different places and times) share one correlation id, including the cancel-cleanup path. The diff handles this correctly: `prepare_tool_calls` calls `resolve_tool_call_id` and pins `tool_call_id: Some(call_id)` onto every executable `ParsedToolCall` (call_prep.rs:221–235). From there every downstream reader sees the same id: `execute_one_tool`'s `event_call_id = tool_call_id_owned.clone().unwrap_or_else(...)` (tool_execution.rs) resolves to that pinned id for both the pending emit (before the tool future runs) and the terminal emit (after completion), and the cancel cleanup's `events::resolve_tool_call_id(call)` (turn/mod.rs) returns the same pinned id — so the interrupted `emit_tool_result` closes the exact card that was opened, with no orphaned pending cards. The `unwrap_or_else(uuid::Uuid::new_v4())` branch in `execute_one_tool` is effectively a defensive fallback (the id is always `Some` for executable calls). No SSOT concern — the id is computed once and threaded by reference.

### 🟢 What looks good — the split is pointed at the correct boundary

Moving the pending `ToolCall` emit to dispatch time in `execute_one_tool` (before `tool.execute(...).await`) is the right fix for the live-card-on-long-running-tool problem described in the body. The refactor of `events.rs` is clean: `emit_tool_call_pair` is preserved for calls that never reach execution (hook-cancelled, denied, replaced, deduplicated) where there is no live window between the two halves, and `post_exec.rs::record_executed_outcomes` correctly stops re-emitting the pair (the executor now owns emission) and is reduced to history/log/hook recording. The doc comments on `emit_tool_call_pending`/`emit_tool_result` spell out the shared-id contract, and `split_pending_then_result_share_resolved_id` (events.rs) unit-tests it directly.

### 🟢 What looks good — SSOT satisfied, no duplicate state introduced

The resolved id is computed once and threaded by reference; it is not cached in a second field that could drift. The interrupted-outcome string is read from the Fluent key `turn-tool-interrupted-before-result` in both the `ordered_results` history synthesis and the new event-layer cleanup, rather than hardcoded — consistent with the localization rule. `event_tx` is plumbed as an `Option<&Sender<TurnEvent>>` parameter all the way down, not stashed on a struct. No SSOT or duplicate-state concern.

### 🔵 Suggestion — the completed-index lookup invariant is subtle and worth a one-line comment

In the cancel-cleanup block (turn/mod.rs) the completed set is built from `executed_completed_indices` (the original `tool_calls` indices), while the skip-loop iterates `executable_indices.iter().zip(executable_calls.iter())` and compares against that set. This is correct because both index spaces are `tool_calls` indices, but the two-index-space dance (`executable_indices` vs the position within `executable_calls`) is exactly the kind of thing that trips a future editor. A single comment at the `HashSet` asserting "these are `tool_calls` indices, matching `executable_indices`" would make the invariant explicit. Non-blocking. (Nillth raised the same suggestion.)

### Verdict

The cross-check at head 3741fe260d555d1cf301046c1cddccfbb397b0dd confirms the event-correlation id is pinned once and shared across the pending emit, the terminal emit, and the cancel-cleanup emit; the duplicate pending-then-interrupted terminal class (Audacity88's blocker) is eliminated by preserving completed parallel outcomes and skipping their indices in cleanup; and the regression that proves it ships in the diff. CI is fully green across the Quality Gate matrix. Audacity88's CHANGES_REQUESTED is now DISMISSED, so there is no active block to wait for. The PR is mergeable and technically sound.

My only standalone finding is a non-blocking invariant comment suggestion. Approving.

Audacity88 (dismissed):
I reviewed current head 3741fe260d555d1cf301046c1cddccfbb397b0dd against the live PR state, @Nillth's current-head comment, @WareWolf-MoonWall's approval, @perlowja's green/mergeable status note, my earlier blocking review, the changed runtime source, the unchanged gateway/web consumers, and the remote CI results. I did not run local Cargo validation for this review-only pass.

### ✅ Resolved — Parallel cancellation no longer double-closes completed tool cards

The earlier blocker around parallel cancellation is fixed in this head. `execute_tools_parallel` now preserves a slot per call, maps in-flight cancellation to `None` instead of discarding completed siblings, and `run_tool_call_loop` tracks completed original `tool_calls` indices before the cancellation cleanup emits interrupted results. The new `run_tool_call_loop_parallel_cancel_no_double_terminal_for_completed_call` regression covers the important case: one parallel sibling completes, another cancels the turn, and each `tool_call_id` receives exactly one terminal `ToolResult`.

The id pinning in `prepare_tool_calls` also looks right. Executable calls get one resolved id before dispatch, and the pending emit, normal terminal emit, and cancellation cleanup all reuse that same id.

### 🔴 Blocking — Parallel tool results can close the wrong web chat card

I agree with @Nillth and @WareWolf-MoonWall that the parallel-cancel double-terminal issue is resolved. I am still requesting changes because this PR also changes the event timing/order that live-card consumers observe.

Before this PR, executed tool events were emitted from `record_executed_outcomes` after the executor returned the collected outcomes, so each `ToolCall`/`ToolResult` pair was emitted adjacent to itself in the recorded call order. This PR moves the pending event to dispatch time and the terminal `ToolResult` into `execute_one_tool`, immediately after that individual tool finishes. That is the right direction for live cards, but it changes the ordering contract for parallel batches: results can now arrive in completion order, not call order.

The gateway already forwards the correlation id in both frames, but the web chat consumer does not use it. `ToolCallInfo` has no id field, the `tool_call` handler creates a pending card with only `{ name, args }`, and the `tool_result` handler updates `prev.findIndex((m) => m.toolCall && m.toolCall.output === undefined)`. So with a batch like slow `call_a` plus fast `call_b`, the event stream can be:

- `tool_call(call_a)`
- `tool_call(call_b)`
- `tool_result(call_b)`
- `tool_result(call_a)`

The current React handler writes `call_b`'s output into the first pending card, which is `call_a`, then writes `call_a`'s output into the remaining `call_b` card. The UI ends up showing the right number of completed cards but with the wrong output attached to the wrong tool.

This is not just unrelated frontend cleanup. The React limitation predates this PR, but the harmful ordering does not: the runtime used to emit each completed call/result pair together after collection, and this PR now emits pending events at dispatch and terminal events as individual futures finish. That makes id-based consumer matching part of the compatibility surface for this change. `AgentContext` always ingests `tool_call` / `tool_result` frames even though the chat view filters tool activity by default, so toggling the tool-activity view back on can reveal mismatched cards. The backend frame shape already has the data needed to fix this; the frontend needs to persist `msg.id` on the pending card and resolve `tool_result` by matching that id. A legacy fallback to the first unresolved card would be fine for id-less frames, but the normal gateway path should be id-keyed. Please add coverage for an out-of-order parallel result sequence so this cannot regress.

### Verdict

Requesting changes. The runtime-side fix for my original blocker looks solid, and CI is green, but the changed event ordering exposes a real consumer bug in the web chat path. The PR should not merge until the live-card consumer matches terminal results by `tool_call_id` rather than by first unresolved card.

PR #7779, opened 2026-06-16
fix(runtime): skip response-cache for multimodal [IMAGE:] markers

### LINKED ISSUE #7741: bug(runtime): skip response-cache hits for multimodal prompt markers
## Affected component

runtime/daemon

## Severity

S2 - degraded behavior

## Current behavior

Response cache lookup still keys off provider messages before multimodal normalization. The cache path also does not skip cache operations when messages contain `[IMAGE:...]` markers.

## Expected behavior

The runtime should avoid response-cache hits for multimodal prompt markers unless the cache key includes the same normalized multimodal payload semantics the provider will see. A narrow first fix can skip response-cache lookup/storage when the provider messages contain `[IMAGE:...]` markers.

## Steps to reproduce

1. Inspect `crates/zeroclaw-runtime/src/agent/agent.rs` on current `master`.
2. Follow `Agent::turn()` and `turn_streamed()` response-cache key construction.
3. Compare cache-key construction with multimodal preparation.
4. Add coverage showing an image-bearing prompt cannot reuse a text-only response-cache entry.

## Impact

Affected users: multimodal/image-bearing runtime turns with response caching enabled.

Frequency: image-bearing turns that reach response-cache paths.

Consequence: cache keys can ignore the real multimodal provider payload and risk stale or wrong responses.

## Logs / stack traces

None.

## ZeroClaw version

## Rust version

## Operating system

## Regression?

Unknown.

## Pre-flight checks

## Related

- #6661
- #6574

crh-code:
## Summary

- `response_cache_key_for_messages()` built cache keys from raw provider messages before multimodal normalization replaced `[IMAGE:...]` markers with data URIs.
- Cache keys built on the raw markers would either miss (different normalised payload) or collide (same marker path but different image content), risking stale or semantically wrong responses.
- Added a guard in `response_cache_key_for_messages()` that returns `None` when any non-system message contains a `[IMAGE:` substring, disabling both cache lookup and storage for multimodal turns.

## Change Type

- [x] Bug fix

## Scope

- [ ] Gateway / [ ] Skills / [ ] Auth / [ ] Memory / [ ] Integrations / [ ] API / [ ] UI/UX / [ ] CI

## Real behavior proof

**Behavior addressed:** Response-cache key construction skipped for messages containing `[IMAGE:]` multimodal markers, preventing stale or wrong cache hits on image-bearing prompts.

**Environment tested:** Linux x86_64; code inspection against current master (1dd9f1826). No local Rust toolchain available for runtime verification — see Not tested section.

**Steps run after the patch:** Inspected `response_cache_key_for_messages()` in `crates/zeroclaw-runtime/src/agent/agent.rs` lines 773-801, confirmed the new guard returns `None` for messages with `[IMAGE:]` markers and `Some(key)` for plain text messages. Reviewed `encode_response_cache_transcript()` to verify the guard is placed before transcript construction so it short-circuits efficiently.

**Evidence after fix:**

```text
response_cache_key_for_messages() before fix:
  - temperature == 0 && response_cache present → compute key from raw transcript
  - [IMAGE:...] markers included in key verbatim → collisions or misses

response_cache_key_for_messages() after fix:
  - temperature == 0 && response_cache present
  - any non-system message contains "[IMAGE:" → return None (skip cache)
  - no markers → compute key from transcript (unchanged path)
```

**Observed result after the fix:** Messages containing `[IMAGE:]` markers correctly skip response-cache operations. Plain text messages continue to produce cache keys as before. Test `response_cache_key_skips_multimodal_image_markers` added to verify both paths.

**Not tested:** Live gateway session with real multimodal model provider and actual image payloads. Cross-platform (macOS/Windows) runtime verification. Full cargo test suite (no local Rust toolchain).

## Root Cause

- Root cause: Cache key was built from raw provider-visible messages before multimodal normalization, so `[IMAGE:]` marker strings became part of the key.
- Missing detection/guardrail: No check for multimodal markers in the cache key construction path.

## Regression Test Plan

- Target test: `response_cache_key_skips_multimodal_image_markers` in `crates/zeroclaw-runtime/src/agent/agent.rs`.
- Verifies plain text produces a cache key and `[IMAGE:]` markers produce `None`.

Multimodal/image-bearing prompts with response caching enabled will no longer risk stale or wrong cached responses. Cache operations are cleanly skipped for such prompts.

## Security Impact

- New permissions? No
- New network calls? No

Closes #7741

---
<a href="https://app.blacksmith.sh/zeroclaw-labs/codesmith/zeroclaw/pr/7779"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784205154&installation_id=140579446&pr_number=7779&repository=zeroclaw-labs%2Fzeroclaw&return_to=https%3A%2F%2Fgithub.com%2Fzeroclaw-labs%2Fzeroclaw%2Fpull%2F7779&signature=fc0aaa18bffb6d77aaebf520f2c8155ea12a45987b4d73ba36b3fc061866e3e0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>


singlerider (approved):
First review (singlerider). Read the single-file diff and verified the test content against the raw bytes on the head (the GitHub diff view strips the literal `[IMAGE:...]` marker from the rendered test string, but the source actually contains `ChatMessage::user("describe this image [IMAGE:/tmp/photo.png]")`). The fix is correct and the test pins the right contract. Approving. (CI is still mid-run; the merge gate will hold for green.)

### 🟢 What looks good: skipping the cache for multimodal prompts is the correct call

`[IMAGE:...]` markers are normalized to data URIs later in the turn loop, so a cache key built on the raw marker is wrong in both directions: it misses (the normalized payload differs from the raw-marker key) and, worse, it collides (two prompts with the same `[IMAGE:/tmp/photo.png]` path but different image bytes on disk would hash identically and could serve a stale or semantically wrong cached response). Returning `None` from the cache-key builder when any non-system message contains `[IMAGE:` is the conservative fix: it only ever disables caching for image prompts, never widens a cache hit, which is the safe direction for a `risk: high` change on the turn path. Excluding `system` messages from the scan is right too, since a system prompt mentioning the marker literally should not disable caching for a text-only turn.

### 🟢 What looks good: the regression test pins both directions

The test asserts a plain text prompt still produces a cache key and a prompt carrying the `[IMAGE:/tmp/photo.png]` marker returns `None`, so a future change that drops the guard would fail here. Scope is held to the one cache-key function with no behavioral drift on the text path.

singlerider:

Audacity88 (commented):
I reviewed PR #7779 at head `d274b36`, including the single-file runtime diff in `crates/zeroclaw-runtime/src/agent/agent.rs`, the closing reference to #7741, the prior approval, and the current check state. I did not run local Cargo or live provider validation for this review.

### 🟢 What looks good — the cache skip moves in the safe direction

Returning `None` from `response_cache_key_for_messages()` when a non-system provider message still contains `[IMAGE:` is the conservative fix for this bug. It disables cache lookup and storage for image-bearing turns instead of risking a cache key built from a raw path marker that does not match the normalized provider payload. The new unit test also covers the two important helper-level paths: plain text still gets a key, and an image marker skips the cache.

### 🟡 Warning — approval still needs executable validation on this head

I am not approving this yet because the current head has not run the normal executable checks. The only check shown in the captured PR check state for `d274b36` is `Apply path labels` passing; the expected executable checks are not visible there, and the PR body says no local Rust toolchain or full Cargo suite was available. For a `risk: high` runtime change on the agent turn path, I need the normal CI gate to run successfully on this head, or a maintainer-recorded validation exception, before approval.

No source change requested from this review; the missing piece is validation evidence for the added runtime test and normal gates.

PR #7780, opened 2026-06-16
fix(runtime): forward trimming budgets through agent_turn wrapper

### LINKED ISSUE #7736: bug(runtime): forward trimming budgets through agent turn wrapper
## Affected component

runtime/daemon

## Severity

S2 - degraded behavior

## Current behavior

`process_message()` still calls the `agent_turn()` wrapper, and that wrapper forwards `0` for `max_tool_result_chars` and `context_token_budget` into the tool-call loop. That disables trimming in this path even when the runtime has computed budgets elsewhere.

## Expected behavior

The `process_message()` path should preserve the effective tool-result and context-token budgets when it reaches the turn loop. The fix should keep the #5287 / #7307 runtime-profile work in mind and avoid widening scope beyond budget forwarding.

## Steps to reproduce

1. Inspect `crates/zeroclaw-runtime/src/agent/loop_.rs` on current `master`.
2. Follow `process_message()` into `agent_turn()`.
3. Observe that `agent_turn()` calls the turn loop with zero budget values.
4. Add coverage showing a non-zero configured/effective budget survives through the wrapper.

## Impact

Affected users: runtime paths that rely on tool-result trimming or context-token budgeting.

Frequency: any `process_message()` path using the wrapper.

Consequence: budget controls can be bypassed in this path, increasing token pressure and making configured limits unreliable.

## Logs / stack traces

None.

## ZeroClaw version

## Rust version

## Operating system

## Regression?

Unknown.

## Pre-flight checks

## Related

- #5287
- #7307

crh-code:
## Summary
- `agent_turn()` hardcoded `0` for `max_tool_result_chars` and `context_token_budget`, disabling trimming even when `process_message()` had computed effective runtime-profile budgets.
- Added both parameters to the `agent_turn()` signature so callers can pass the configured values through to `run_tool_call_loop()`.
- `process_message()` now forwards `agent.resolved.max_tool_result_chars` and `agent.resolved.max_context_tokens` instead of silently dropping them.
- Existing test callers pass `0, 0` to preserve the legacy disabled behaviour.

## Change Type
- [x] Bug fix

## Scope
- [x] Runtime
- [x] Agent

## Verification
- `cargo check -p zeroclaw-runtime` — passes
- `cargo fmt --all -- --check` — passes
- Two regression tests pass:
  - `agent_turn_forwards_max_tool_result_chars_to_truncate_tool_result` — verifies truncation with `max_tool_result_chars=100`
  - `agent_turn_with_zero_budget_preserves_full_tool_result` — verifies full output preserved with `max_tool_result_chars=0`

## Real behavior proof

**Behavior or issue addressed:** `agent_turn()` now forwards `max_tool_result_chars` and `context_token_budget` to `run_tool_call_loop()` instead of hardcoding `0, 0`.

**Environment tested:** Rust 1.96.0 on Linux x86_64, cargo test -p zeroclaw-runtime.

**Steps run after the patch:** Ran the two regression tests that verify budget forwarding through the `agent_turn` wrapper.

**Evidence after fix:**

```text
test agent::loop_::tests::agent_turn_forwards_max_tool_result_chars_to_truncate_tool_result ... ok
test agent::loop_::tests::agent_turn_with_zero_budget_preserves_full_tool_result ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured
```

**Observed result after the fix:** With `max_tool_result_chars=100`, oversized tool results are truncated. With `max_tool_result_chars=0`, full output is preserved. Previously both values were hardcoded to `0`, silently disabling trimming.

**Not tested:** Live gateway process_message path with a configured runtime profile. Cross-platform verification (tested on Linux only).

Closes #7736

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/loop_.rs
+++ b/crates/zeroclaw-runtime/src/agent/loop_.rs
@@ -636,6 +636,8 @@ pub async fn agent_turn(
     model_switch_callback: Option<ModelSwitchCallback>,
     strict_tool_parsing: bool,
     parallel_tools: bool,
+    max_tool_result_chars: usize,
+    context_token_budget: usize,
     channel: Option<&dyn Channel>,
 ) -> Result<String> {
     run_tool_call_loop(
@@ -662,9 +664,9 @@ pub async fn agent_turn(
         &zeroclaw_config::schema::PacingConfig::default(),
         strict_tool_parsing,
         parallel_tools,
-        0,    // max_tool_result_chars: 0 = disabled (legacy callers)
-        0,    // context_token_budget: 0 = disabled (legacy callers)
-        None, // shared_budget: no shared budget for legacy callers
+        max_tool_result_chars,
+        context_token_budget,
+        None, // shared_budget: no shared budget for agent_turn callers
         channel,
         None, // receipt_generator
         None, // collected_receipts
@@ -3044,6 +3046,8 @@ pub async fn process_message(
                     None,
                     agent.resolved.strict_tool_parsing,
                     agent.resolved.parallel_tools,
+                    agent.resolved.max_tool_result_chars,
+                    agent.resolved.max_context_tokens,
                     None, // channel: process_message path has no channel ref
                 ),
             )
@@ -3078,6 +3082,7 @@ mod tests {
         NamedMockTool,
         CompletesAndSignalsTool,
         CancelsTurnTool,
+        VerboseTool,
     );
 
     // ── maybe_inject_channel_delivery_defaults tests ──────────────
@@ -9065,6 +9070,8 @@ This is an example, not an invocation."#;
                 None,
                 false,
                 false, // parallel_tools
+                0,     // max_tool_result_chars: disabled for test
+                0,     // context_token_budget: disabled for test
                 None,  // channel
             )
             .await
@@ -9131,6 +9138,8 @@ This is an example, not an invocation."#;
                 None,
                 true,
                 false, // parallel_tools
+                0,     // max_tool_result_chars: disabled for test
+                0,     // context_token_budget: disabled for test
                 None,  // channel
             )
             .await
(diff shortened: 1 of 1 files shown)


singlerider (approved):
First review (singlerider). Read the single-file diff and traced every `agent_turn` caller. This restores config-driven trimming on the persistent-message path, which was silently disabled. Approving. (CI is still mid-run; the merge gate will hold for green.)

### 🟢 What looks good: the bug is real and the fix threads the actual effective values

`agent_turn` hardcoded `0, 0` for `max_tool_result_chars` and `context_token_budget` (0 = disabled), so every turn routed through this wrapper, including the `process_message` persistent path, ran with tool-result trimming and the context-token budget turned off regardless of the agent's config. The tell is right there in `process_message`: `effective_max_context_tokens` and `effective_max_tool_result_chars` were computed into `_eff_`-prefixed bindings and then discarded. This PR adds the two parameters to `agent_turn` and passes the real `effective_*` values at the production call site, so an agent that configured `max_tool_result_chars` or `max_context_tokens` now actually gets them applied on this path. The `effective_*` helpers already encode the per-agent precedence, so this is wiring an existing policy through, not inventing one.

### 🟢 What looks good: all callers updated, no orphan break

`agent_turn` has exactly three call sites: the production `process_message` invocation (now forwarding `eff_max_tool_result_chars` / `eff_max_context_tokens`) and two tests, both of which explicitly pass `0, 0` to keep trimming disabled for the test scenario. Adding required positional params is compile-enforced, so there is no path that silently keeps the old `0, 0`. Scope is held to the wrapper signature and its callers.

### 🟢 Worth a follow-up: a behavioral test on the persistent path

The change is a direct plumbing fix and the compiler guarantees the value is forwarded, but a `process_message` test asserting that a configured `max_tool_result_chars` actually truncates an oversized tool result on this path would pin the regression so a future revert to `0` is caught behaviorally rather than only by reading. Not blocking given the size and the clear before/after, but cheap to add.

singlerider:

Audacity88 (changes requested):
Context: I checked current head 728ff5f against the live PR state, linked issue #7736, the current diff against master, singlerider's earlier approval, inline comments, formal reviews, and the visible checks. I did not run local Cargo validation. The only visible GitHub check on current head is the path-label check.

Thanks @crh-code. The wiring change is pointed at the right bug, but I am requesting changes because this runtime path still needs a regression test and real validation before merge.

### 🟢 What looks good — the wrapper now forwards resolved budgets

The current diff stops hardcoding `0, 0` inside `agent_turn()` and threads named `max_tool_result_chars` and `context_token_budget` parameters into the existing turn-loop call. The production `process_message()` call now passes `agent.resolved.max_tool_result_chars` and `agent.resolved.max_context_tokens`, which keeps the fix narrow and tied to the resolved runtime profile values already available at that call site.

### 🔴 Blocking — add coverage and validation for the budget-forwarding path

Issue #7736 did not ask only for the plumbing change. Its repro asked for coverage showing that a non-zero configured or effective budget survives through the wrapper. The current PR body also says that test is still future work, and the only updated wrapper tests pass `0, 0`, so they preserve the old disabled behavior instead of failing if `process_message()` or `agent_turn()` drops the budgets again.

Please add a focused regression that would fail on the old `agent_turn(..., 0, 0, ...)` path and pass with the forwarded values. For this risk-high runtime change, please also refresh the validation evidence with at least the relevant formatting and focused Rust test output, or explicitly route the branch for maintainer/CI validation if you still cannot run Rust locally. Right now the live PR checks only show the path-label job, so there is no compile/test signal for current head.

Audacity88 (approved):
Reviewed current head `9033e6ac586c439f4156b61da25185818a0b3d46`, including the live PR state, linked issue #7736, prior reviews, the current single-file diff, and the now-green GitHub checks. I did not run local Cargo.

Thanks @crh-code. The blocker from my prior review is addressed.

This approval clears my prior `CHANGES_REQUESTED` review; it is not a separate merge-readiness claim beyond the normal branch protection and maintainer checks.

### ✅ Resolved — Budget forwarding now has focused regression coverage

The current head adds a non-zero-budget `agent_turn()` regression that would fail on the old hardcoded `0, 0` wrapper path: it passes `max_tool_result_chars = 100` and verifies the oversized tool output is no longer preserved in full. The companion zero-budget test verifies the legacy disabled behavior still preserves the full output when `max_tool_result_chars = 0`.

### 🟢 What looks good — The production wiring is still narrow

The production `process_message()` call now forwards `agent.resolved.max_tool_result_chars` and `agent.resolved.max_context_tokens` into `agent_turn()`, and `agent_turn()` forwards those values into the existing tool-call loop. The PR stays scoped to the wrapper plumbing and focused regression coverage for #7736.

Issue #7468, opened 2026-06-10
[Feature]: Allow non-agent aliases to be renamed in Zerocode


damajor:
### Summary

<img width="901" height="814" alt="Image" src="https://github.com/user-attachments/assets/8fd5bdef-462b-4016-b19b-fa8513be7032" />

Allow aliases to be renamed in the TUI

### Problem statement

While adding model providers you are asked to enter an alias, same goes for agent names, mcp bundles, and many other named settings.

At this moment the user cannot rename the aliases within the TUI.

### Proposed solution

Use a shortcut or any UI solution to allow the user to rename the aliases.

### Breaking change?

No

### Data hygiene checks

- [x] I removed personal/sensitive data from examples, payloads, and logs.
- [x] I used neutral, project-focused wording and placeholders.

singlerider:
Handle filesystem and config cascade for CRUD operations.

singlerider:
Moved from v0.8.3 to v0.8.4. v0.8.3 is over capacity for its ship date and cannot land every open item in the window. This is not in-progress, not a p1 blocker, and not part of the v0.8.3 headline scope (WASM/plugin platform, skills, MCP dashboard), so it is parked in the v0.8.4 overflow lot for re-triage into a concrete release. No change in priority or intent; this is a scheduling move.

Audacity88:
## Status Update

The shared rename foundation and several user surfaces have landed:

- #7840 added config rename-with-cascade support.
- #7841 added agent-owned-state rename handling.
- #7842 added CLI rename flows for agents, providers, and channels.
- #7919 added web alias rename.
- #7954 added agent rename in Zerocode.

Keep this issue open for the remaining Zerocode surfaces: provider, channel, bundle, and other editable aliases. Those flows should use the existing shared rename RPC and cascade preview rather than adding TUI-specific rename logic.

This remains in Icebox until that broader Zerocode slice is scheduled.

Issue #7175, opened 2026-06-03
feat(config): typed delete-with-cascade for aliased entries (providers, agents, channels)


singlerider:
## Background

V3 config is built on typed aliased entries that reference each other by
`<type>.<alias>` refs (`define_provider_ref!`-generated newtypes):

- `ModelProviderRef` → `providers.models.<family>.<alias>`
- `TtsProviderRef` → `providers.tts.*`
- `TranscriptionProviderRef` → `providers.transcription.*`
- `ChannelRef` → `channels.<type>.<alias>`
- `AgentAlias` → `agents.<alias>`

`Config::validate()` already fails loud on dangling refs at load, and (as of the
provider-fallback work) `Config::collect_warnings()` surfaces dangling/cyclic
`providers.models.*.fallback` refs as non-fatal warnings. What does NOT exist is
a **delete path that removes an aliased entry AND scrubs every reference to it**
in one operation. `ModelProviders::remove_alias` exists but (a) has no production
caller and (b) does not cascade. Deleting an alias today therefore leaves
dangling refs that only surface later as validation errors/warnings — the
operator has to hunt down and hand-edit every referrer.

## What needs to happen

A canonical, surface-agnostic "delete this aliased entry and fix up all
referrers" operation, callable identically from TUI, web dashboard, and RPC
(no per-surface reimplementation — canonical-registry rule).

### Provider aliases (simpler)

Deleting `providers.models.<family>.<alias>` must:
1. Remove the entry.
2. Scrub the alias from every other provider entry's `fallback` vector.
3. For any AGENT that references the deleted alias via `model_provider` or
   `classifier_provider`: this is a HARD reference (an agent's `model_provider`
   is mandatory). Deletion here must be refused, or require the operator to
   first repoint/clear the referring agent. Decide policy (see Open Questions).

TTS / transcription provider deletes follow the same shape against
`tts_provider` / `transcription_provider` (those are opt-out/empty-allowed, so
scrubbing to empty is safe).

### Agent aliases (more complex)

An agent is referenced from many places; deleting `agents.<alias>` must account
for ALL of them (inventory cross-checked against source 2026-06-04):

- `peer_groups.<group>.agents[]` — remove the member (and prune the group if it
  drops below the 2-member minimum?).
- `agents.<other>.workspace.access` (BTreeMap keyed by `AgentAlias`) — remove key.
- `agents.<other>.workspace.read_memory_from[]` — remove entry.
[...]

PR #7781, opened 2026-06-16
fix(memory): purge_agent must resolve alias→id before deleting

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `SqliteMemory::purge_agent` bound the agent **alias** straight into the `memories.agent_id` column — but `agent_id` holds the agent's **UUID** (FK → `agents.id`). The `DELETE` therefore matched zero rows and returned `Ok(0)`: deleting an agent **silently kept all of its memory rows**. The gateway delete handler only logs on `rows > 0`, so the no-op left no trace.
  - Resolve `alias → id` via the same subselect the insert path uses (`store_with_agent`), `LIMIT 1` to mirror that idiom. An unknown alias yields a `NULL` subselect and correctly deletes nothing.
  - Add a two-agent regression test: seeds 3 rows for `alpha` + 2 for `rogue`, asserts `purge_agent("alpha")` returns 3 and leaves exactly 2, and that an unknown alias is a no-op. (Under the old code the call returned 0 and left all 5 — the test pins the regression.)
- **Scope boundary:** SQLite backend only — the uniquely *silent* case. The `Memory` trait's default `purge_agent` `bail!`s loudly ("not supported"), so Postgres / Qdrant / Markdown already surface a warning rather than a silent no-op; implementing those is separate. This PR does **not** delete the `agents` row itself, does **not** add `ON DELETE CASCADE`, and does **not** touch the gateway delete handler.
- **Blast radius:** The only caller is the gateway agent-delete path (`DELETE /api/config/map-key?path=agents&key=<alias>`). Its behavior is unchanged except that the memory purge now actually happens. No public signature or API change.
- **Linked issue(s):** Related #7175, Related #7468 — surfaced while scoping the alias delete-cascade / rename work; this is an independent correctness fix and does not close them.

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-memory
```

  - `cargo fmt --all -- --check` → clean, no diff (whole workspace).
  - `cargo clippy --all-targets -- -D warnings` → clean across the workspace (exit 0, 0 warnings).
  - `cargo test -p zeroclaw-memory` →
    ```
    Running unittests src/lib.rs
    test result: ok. 308 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
    Running tests/agent_migration.rs
    test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
    Doc-tests zeroclaw_memory: ok. 0 passed; 0 failed
    ```
    (includes the new `sqlite_purge_agent_deletes_only_that_agents_rows`.)
- **Beyond CI — what did you manually verify?** The regression directly — confirmed the failing scenario (old code: `Ok(0)`, count stays 5) and that the fix purges exactly the target agent's rows and is a no-op for an unknown alias. Did **not** drive the live gateway delete endpoint end-to-end (covered by the unit-level purge assertion + the unchanged handler).
- **If any command was intentionally skipped, why:** full-workspace `cargo test` not run locally — the change is one SQL string + a test in a single crate with an unchanged public signature; ran the full changed-crate suite + workspace `fmt` + workspace `clippy` instead. CI runs the complete suite.

- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (test uses synthetic aliases `alpha` / `rogue`)

## Compatibility (required)

- Upgrade steps: none. Behavioral note: deleting an agent now actually purges that agent's SQLite memory rows (as already intended). No data migration; only affects future deletes.

## Rollback


singlerider (approved):
Reviewed the SQLite `purge_agent` fix end to end against the live source. Clean, well-scoped correctness fix with a regression test that actually pins the bug. Thanks for the thorough write-up, @Nillth.

I also stripped the auto-injected Codesmith/Blacksmith footer from the PR body — that ad markup gets appended automatically and isn't part of your description.

### 🟢 Root cause correctly identified and fixed

`memories.agent_id` holds the agent UUID (FK → `agents.id`), but `purge_agent` bound the human alias straight into that column, so `DELETE FROM memories WHERE agent_id = ?1` matched zero rows and returned `Ok(0)`. Verified the fix mirrors the existing insert idiom exactly — `store_with_agent` (sqlite.rs:1465) resolves the agent via `(SELECT id FROM agents WHERE alias = ? LIMIT 1)`, and the new `DELETE` subselect uses the same shape. An unknown alias yields a NULL subselect and deletes nothing, which is the correct outcome.

### 🟢 The silence claim checks out

Confirmed the gateway handler at `api_config.rs:1032` only logs on `Ok(rows) if rows > 0`, with `Ok(_) => {}` swallowing the zero-row case — so the no-op left no trace, exactly as described. Also confirmed the `Memory` trait default `purge_agent` `bail!`s loudly (`memory_traits.rs:256`), so non-SQLite backends surfaced an "unsupported" warning rather than a silent no-op. Scoping this to the SQLite backend as the uniquely silent case is the right call.

### 🟢 Regression test pins the bug

`sqlite_purge_agent_deletes_only_that_agents_rows` seeds 3 `alpha` + 2 `rogue` rows, asserts `purge_agent("alpha")` returns 3 and leaves exactly 2, and that an unknown alias is a no-op returning 0. Under the old code this returned 0 and left all 5 — the test would have caught it.

Single-file change, MERGEABLE, authorship clean, no public signature change. Scope boundary (no `agents` row delete, no cascade, no handler change) is appropriately narrow.

PR #6743, opened 2026-05-17
fix(providers): skip unresolvable multimodal images

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Recover the #6074 / `da742938` multimodal behavior on the current provider crate: image refs that cannot be loaded no longer fail the whole provider-preparation step.
  - Add a short note to the prepared message when one or more attached images were skipped, while still normalizing any valid image refs to data URIs.
  - Apply the per-request image cap after normalization so skipped refs do not consume image budget and evict older valid images.
  - Preserve native tool-result JSON shape while applying the same skip behavior inside the JSON `content` field.
- **Scope boundary:** This PR changes multimodal provider message preparation and updates the matching runtime regression test. It does not change channel attachment parsing, provider vision capability routing, config schema, or non-image media handling.
- **Blast radius:** Provider payload preparation for user messages and current tool-result messages containing `[IMAGE:...]` markers. Text-only messages and stale tool-result image stripping keep their existing paths.
- **Linked issue(s):** Related #6074. Supersedes #4316.
- **Labels:** `bug`, `risk: medium`, `size: M`, `agent`, `provider`, `runtime`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
cargo fmt --all -- --check
Result: passed.

cargo test -p zeroclaw-providers prepare_messages_ -- --nocapture
Result: passed. 13 prepare_messages_ multimodal tests passed, 0 failed, 889 filtered out.

cargo clippy -p zeroclaw-providers --all-targets -- -D warnings
Result: passed.

cargo test
Result: passed. Full workspace test run completed successfully; live provider-credential tests were ignored as expected.

cargo clippy --all-targets -- -D warnings
Result: passed.

cargo test -p zeroclaw-runtime agent::loop_::tests::run_tool_call_loop_skips_oversized_image_payload -- --nocapture
Result: passed.

cargo nextest run --locked -p zeroclaw-runtime agent::loop_::tests::run_tool_call_loop_skips_oversized_image_payload
Result: passed.

cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings
Result: passed.

cargo nextest run --locked --workspace --exclude zeroclaw-desktop
Result: passed. 7515 tests run, 7515 passed, 10 skipped.

git diff --check
Result: passed with no output.

Post-review follow-up on current head:
cargo fmt --all -- --check
Result: passed.

cargo test -p zeroclaw-providers prepare_messages_ -- --nocapture
[...]

PR #7782, opened 2026-06-16
fix(providers): remove dead pre-normalization image trim and its misleading log

Nillth:
## Summary

- **What changed and why:**
  - `prepare_messages_inner` (multimodal) ran a pre-normalization image trim whose result (`trimmed`) the normalization loop **never used** — the loop iterates the original `messages`, and `trimmed` fed only `latest_tool_result_indices`, which returns the same set whether computed from `messages` or `trimmed`. So the block was effectively dead.
  - Its only observable effect was a `WARN "multimodal: trimming oldest images — conversation exceeds image limit" (trimmed_to: N)` that fired on every multi-image conversation but **never actually trimmed anything**, producing a confusing log pair next to the genuine post-normalization cap warning (the symptom that surfaced this: one line claims `trimmed_to: 4`, the next reports `images_after_normalization: 9` for the same request).
  - Normalizing all markers first and capping only the images that *successfully* normalize is intentional (covered by `skipped_images_do_not_consume_image_budget`): it stops a newer broken image ref from evicting an older valid one. The post-normalization cap already keeps the newest successful images and prevents conversations from sticking, so the pre-normalization trim is redundant.
  - Removed the dead block and its false warning; compute `latest_tool_indices` from `messages` for consistency with the loop source.
- **Scope boundary:** No change to the post-normalization cap, to `trim_old_images`, to age-based trimming, or to any provider adapter. Output of `prepare_messages_inner` is unchanged for all inputs; only the redundant work and the misleading log line are removed.
- **Blast radius:** `zeroclaw-providers` multimodal message preparation only. The one removed log line (`trimmed_to`) disappears from logs; the accurate `post-normalization image cap exceeded` WARN remains as the single cap signal. Anything grepping specifically for the old `trimmed_to` message would stop matching (it was never truthful).
- **Linked issue(s):** None — surfaced from production logs.

Full battery run locally. `cargo test --workspace` was run with `-j 2` because the default job count OOM-kills this (swapless) box while linking all test binaries in parallel; the reduced job count is a build-host constraint only and does not affect what is tested.

```bash
cargo fmt --all -- --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace -j 2
```

```
$ cargo fmt --all -- --check
(clean, rc=0)

$ cargo clippy --workspace --all-targets -- -D warnings
    Checking zeroclaw-providers v0.8.0 (.../crates/zeroclaw-providers)
    ...
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 09s
(rc=0, no warnings)

$ cargo test --workspace -j 2
... 63 test binaries ...
WORKSPACE TOTALS: passed=8730 failed=0 ignored=22
   Doc-tests: ok
(rc=0)
```

- **Beyond CI — what was manually verified?**
  - The alternative ("iterate `trimmed`") approach was tried first and **deliberately rejected**: it broke `skipped_images_do_not_consume_image_budget`, proving the loop must iterate `messages` so a failed newer ref cannot evict a valid older image. The shipped change keeps that test green.
  - Confirmed `latest_tool_result_indices(&trimmed) == latest_tool_result_indices(messages)`: the function keys on role/position, and its only content-sensitive check (`is_prompt_tool_result_message`) inspects the *last* message, which oldest-first trimming never strips. Computing it from `messages` is equivalent.
  - Added `prepare_messages_caps_to_newest_successful_images` (9 valid images, max 4 → newest 4 survive, oldest 5 stripped) to lock in the post-normalization cap as the single trim path.
- **What was NOT verified:** no live end-to-end provider run; behavior reasoned about + covered by unit tests.
- **If any command was intentionally skipped, why:** none skipped; `-j 2` used for `cargo test` only, to fit the build host's memory.

- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (test uses a synthetic 1x1 PNG in a temp dir)

## Compatibility (required)

- Backward compatible? **Yes** — `prepare_messages_inner` output is unchanged for all inputs.
- Upgrade steps: None.

## Rollback

### DIFF
--- a/crates/zeroclaw-providers/src/multimodal.rs
+++ b/crates/zeroclaw-providers/src/multimodal.rs
@@ -504,29 +504,16 @@ async fn prepare_messages_inner(
         });
     }
 
-    // When image count exceeds the limit, strip markers from oldest messages
-    // first so that the most recent (most relevant) images survive. This
-    // prevents conversations from becoming permanently stuck once the
-    // cumulative image count crosses the threshold.
-    let trimmed = if total_images > max_images {
-        ::zeroclaw_log::record!(
-            WARN,
-            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
-                .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
-                .with_attrs(::serde_json::json!({
-                    "total_images": total_images,
-                    "max_images": max_images,
-                    "trimmed_to": max_images,
-                })),
-            "multimodal: trimming oldest images — conversation exceeds image limit"
-        );
-        trim_old_images(messages, max_images)
-    } else {
-        messages.to_vec()
-    };
-
+    // Normalize every image marker first, then enforce the per-request image
+    // cap further below based only on images that *successfully* normalize.
+    // Trimming the oldest images *before* normalization is unsafe: a newer
+    // image ref that fails to load would evict an older valid one that could
+    // still have been sent (see `skipped_images_do_not_consume_image_budget`).
+    // The post-normalization cap keeps the most recent successful images and
+    // prevents conversations from sticking once the cumulative count crosses
+    // the threshold, so no pre-normalization trim is needed here.
     let remote_client = build_runtime_proxy_client_with_timeouts("model_provider.ollama", 30, 10);
-    let latest_tool_indices = latest_tool_result_indices(&trimmed);
+    let latest_tool_indices = latest_tool_result_indices(messages);
 
     let mut normalized_messages = Vec::with_capacity(messages.len());
     let mut has_successful_images = false;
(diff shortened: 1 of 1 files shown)


crates/zeroclaw-providers/src/multimodal.rs
```diff
@@ -504,29 +504,16 @@ async fn prepare_messages_inner(
...
+    // cap further below based only on images that *successfully* normalize.
+    // Trimming the oldest images *before* normalization is unsafe: a newer
+    // image ref that fails to load would evict an older valid one that could
+    // still have been sent (see `skipped_images_do_not_consume_image_budget`).
+    // The post-normalization cap keeps the most recent successful images and
+    // prevents conversations from sticking once the cumulative count crosses
+    // the threshold, so no pre-normalization trim is needed here.
     let remote_client = build_runtime_proxy_client_with_timeouts("model_provider.ollama", 30, 10);
-    let latest_tool_indices = latest_tool_result_indices(&trimmed);
+    let latest_tool_indices = latest_tool_result_indices(messages);
```
singlerider, line 516: 🟢 This rebind is the load-bearing equivalence: computing from `messages` instead of the old `trimmed` is safe because `latest_tool_result_indices` keys only on role and position, and its single content-sensitive branch (`is_prompt_tool_result_message`) inspects the last message, which oldest-first `trim_old_images` never strips. So the result is identical to the pre-change call. Good catch removing the binding rather than papering over it.

singlerider (approved):
I read this against the live `multimodal.rs`, traced every consumer of the removed `trimmed` binding, and checked it alongside #7775, which is attacking the sibling leak in the runtime pruner / Anthropic adapter. Same disease (stale or synthetic artifacts polluting the live provider payload and emitting a misleading signal), different organ. This one is the surgical, low-risk half and it is correct.

### 🟢 The dead-code claim verifies end to end

The pre-normalization block was genuinely dead. The normalization loop iterates the original `messages`, never `trimmed`. The only thing `trimmed` fed was `latest_tool_result_indices(&trimmed)`, and that result is provably identical to computing it from `messages`: `latest_tool_result_indices` keys purely on role and position, and its only content-sensitive branch (`is_prompt_tool_result_message`) inspects the last message, which oldest-first `trim_old_images` never strips. So the block's only observable effect was the false `trimmed_to: N` WARN firing on every multi-image conversation while the real cap ran later. Deleting it removes work that did nothing and a log line that lied.

### 🟢 The redundancy was created upstream, and this closes it cleanly

The post-normalization cap that makes this block redundant landed in #6743 (cap moved after normalization so failed refs cannot evict valid older images, locked by `skipped_images_do_not_consume_image_budget`). This PR does not reimplement, replace, or compete with that path. It removes the orphaned second trim that #6743 left behind, leaving exactly one live trim path. `trim_old_images` itself stays referenced by the real cap, so there is no dead-function fallout.

### 🟢 Test earns its place

`prepare_messages_caps_to_newest_successful_images` (9 valid images, max 4, newest 4 survive) pins the post-normalization cap as the single trim path and guards against a future reintroduction of a pre-norm trim. It exercises the real `prepare_messages_for_provider` rather than the helper, which is the right level.

CI is fully green and the diff is one file. Approving.

PR #7783, opened 2026-06-16
docs(ci): reflect merge queue disabled on master

JordanTheJet:
## Summary

The **Require merge queue** branch protection on `master` was turned off — maintainers merge directly again. This updates `master-branch-flow.md` so it no longer tells maintainers to *"Merge when ready → merge queue"*, which is now inaccurate.

## Changes

- Mark the `merge_group` trigger as **dormant** in the Active Workflows / Event Summary tables — it's retained but never fires while the queue is disabled.
- Restore the direct-squash-merge step in the PR → `master` flow and collapse the queue nodes in the diagram.
- Keep a short note describing what the queue did and how to re-enable it (one-click branch-protection toggle).

## Not changed

`ci.yml` is **intentionally untouched** — the `merge_group` trigger is inert without an active queue, so leaving it wired keeps re-enabling the queue a pure UI action with zero code change.


singlerider (approved):
First review (singlerider). Docs-only update to the CI flow reference. CI is 17/17 green including Docs Style. The change accurately reflects current reality: the merge queue is off and maintainers squash directly, which matches how landings are actually happening right now. Approving.

### 🟢 What looks good: the doc is internally consistent across all three surfaces

The merge-queue-disabled state is reflected coherently in every place it appears: the "Last updated" note, the trigger table (`merge_group (dormant)` with the explanation that it stays wired but never fires), the event table (the `merge_group` row marked Inactive), the step-by-step flow (direct squash instead of "Merge when ready"), and the mermaid diagram (the `Q`/`MG`/`E` queue nodes collapsed to a single direct-merge node). A doc that updated the prose but left the diagram showing the old queue path would be worse than no update, so getting all of them in one pass is the right call. Calling out that re-enabling is a one-click branch-protection toggle with no code change (because the `merge_group` plumbing is retained) is the genuinely useful part for a future maintainer.

singlerider:

Audacity88 (approved):
I reviewed current head `8bf307a`, live PR state, CI checks, singlerider's prior approval, and the changed workflow reference doc; I did not run local docs checks, and remote Docs Style plus full CI are green.

### 🟢 What looks good — merge-flow documentation now matches current operation

This updates the merge-flow reference in the right places without changing CI behavior: the top-level status note, Active Workflows table, Event Summary, step-by-step PR flow, and Mermaid diagram all consistently describe direct maintainer squash merges while keeping `merge_group` documented as dormant plumbing. That is the right balance for the current state: maintainers do not get stale "Merge when ready → merge queue" instructions, and a future maintainer still has the context needed to re-enable the queue from branch protection without a workflow edit.

I also agree with leaving `ci.yml` untouched here. Keeping the inert `merge_group` trigger wired is harmless while the queue is disabled and avoids turning a docs correction into a workflow mutation.

PR #7573, opened 2026-06-13
fix(gateway): route chat workspaces per agent

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Gateway WS chat now defaults omitted session cwd requests to the selected agent's configured workspace, instead of the shared `data_dir`, so the WS path matches per-agent workspace policy.
  - One-shot `channel send` construction now resolves Telegram, Discord, Slack, Matrix, and WeChat workspace roots through `Config::channel_workspace_dir("<type>.default")`, matching the daemon channel path and preserving the existing orphan-channel fallback behavior.
  - Tests verify the WS no-cwd default and the one-shot channel owner workspace resolution.
- **Scope boundary:** This PR does not change config schema, env vars, CLI arguments, explicit WS cwd handling, daemon channel startup, or how one-shot channel send selects the existing `default` alias.
- **Blast radius:** Gateway WS chat session cwd defaults and one-shot `channel send` file/tool workspace behavior for Telegram, Discord, Slack, Matrix, and WeChat. Explicit WS cwd callers remain unchanged. Channel refs with no owning agent still use the existing `channel_workspace_dir` fallback behavior.
- **Linked issue(s):** Fixes #7541
- **Labels:** `bug`, `risk: high`, `size: S`, `channel`, `channel: core`, `gateway`, `gateway: ws`

## Validation Evidence (required)

- **Commands run and tail output:**

```text
$ cargo test -p zeroclaw-gateway resolve_ws_session_cwd
running 2 tests
test ws::tests::resolve_ws_session_cwd_keeps_requested_cwd_strict ... ok
test ws::tests::resolve_ws_session_cwd_defaults_to_agent_workspace_without_request ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 250 filtered out; finished in 0.00s

$ cargo test -p zeroclaw-channels one_shot_channel_workspace_dir_uses_owning_agent_workspace
running 1 test
test orchestrator::tests::one_shot_channel_workspace_dir_uses_owning_agent_workspace ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 742 filtered out; finished in 0.04s

$ cargo fmt --all -- --check
exit 0

$ git diff --check
exit 0

$ cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 98m 47s

[...]

Issue #7541, opened 2026-06-12
[Bug]: V3 legacy paths still use shared data_dir as agent workspace_dir: gateway WS chat default session cwd + one-shot channel send builder


NiuBlibing:
### Affected component

gateway/api

### Severity

S2 - degraded behavior

### Current behavior

Since the V3 schema rename (`Config.workspace_dir` → `Config.data_dir`, 062c4244a / #6398), `data_dir` (default `~/.zeroclaw/data`) is documented as the *shared instance data directory* (databases, state files), while per-agent files live at `agents/<alias>/workspace/` resolved via `Config::agent_workspace_dir`. Two legacy code paths still hand `config.data_dir` to agents/channels as their `workspace_dir`, so the agent's effective workspace is the shared data dir instead of its own workspace. (Related: #7236 fixed the same class of bug for `ChannelRuntimeContext`, but these two paths were missed.)

**1. Gateway WS chat: default session cwd is `data_dir`, and it overrides the sandbox root**

- `crates/zeroclaw-gateway/src/ws.rs` L391: `resolve_session_cwd(requested_cwd.as_deref(), &config.data_dir)` — the web dashboard normally sends no cwd, so the session cwd defaults to `~/.zeroclaw/data`.
- ws.rs L415-422: the resolved cwd is unconditionally passed as `Some(&session_cwd)` to `Agent::from_config_with_session_cwd_and_mcp_backchannel`.
- `crates/zeroclaw-runtime/src/agent/agent.rs` L1159-1162: any `Some(session_cwd)` overrides the per-agent policy: `policy.workspace_dir = cwd.to_path_buf()`; the agent's own workspace is merely appended to `allowed_roots`.

Net effect, observed on a live install: when chatting with an agent through the gateway web dashboard, the agent's sandbox root / default working directory is `~/.zeroclaw/data`. Relative paths in file tools resolve there and files the agent writes land in the shared data dir next to the SQLite databases, not in `agents/<alias>/workspace/`. CLI/TUI and channel paths pass no session cwd and correctly get the per-agent workspace via `SecurityPolicy::for_agent` (policy.rs L2412-2416), so behavior diverges between web chat and every other entry point.

**2. One-shot `channel send` builder: channels constructed with `data_dir` as workspace**

[...]

PR #7784, opened 2026-06-16
fix(discord): persist slash reconcile state + read shared stores from data_dir

Nillth:
This PR bundles two related Discord/`data_dir` durability fixes.

---

# 1. Persist slash reconcile state across restarts

## Problem

The Discord channel reconciles its application's global slash commands with the
desired (skill-derived) set on every gateway `READY`. The fingerprint that lets
an unchanged set skip the REST round-trip was held only in memory
(`registered_commands_hash`), so:

- every process start re-ran a full `GET` + per-command diff against Discord's
  finite daily command-create budget, even when nothing changed; and
- a `429` mid-reconcile `bail!`ed and reset the fingerprint, so the next `READY`
  immediately re-hammered the same rate limit — a daily reconcile-429 burst.

## Change

Persist the reconcile state per application id at
`<workspace_dir>/state/slash-commands.json` (the `state/` convention the model
cache already uses), in a small new module `discord_slash_state`:

- **Skip an unchanged set across restarts**, not just within a process.
- **Honour `Retry-After` across restarts:** on a `429` (command list or upsert),
  record the deadline (body `retry_after`, else the `Retry-After` /
  `X-RateLimit-Reset-After` headers; ≥1s floor) and skip reconcile until it
  passes, instead of bailing and retrying on the next `READY`.

`reconcile_slash_commands` now returns a `ReconcileOutcome`
(`Reconciled` | `RateLimited { until }`) so a rate limit is a first-class
cooldown rather than a hard error. A genuine hard error leaves the persisted
state untouched, so the next `READY` still retries (the new fingerprint differs
from the stored one). The state file is written atomically (temp + rename) so an
overlapping reconcile during a reconnect storm can't observe a torn file.

### Compatibility / scope

- **No new config keys.** Rides the existing per-channel `state/` dir.
- **Degrades to a no-op** when no workspace dir is configured — behaviour
  unchanged there (a full reconcile each start, exactly as before).
- The persisted file holds only `{app_id, fingerprint, last_success_at,
  retry_after_until}` — no token or message content; the `state/` dir is 0700.
- A swapped bot token (different application id) is detected on load and forces a
  full reconcile rather than trusting a stale fingerprint.

---

# 2. Shared-store read-path fix (`discord_search` + `sessions_*`)

## Problem

`discord_search` and the `sessions_*` tools opened their SQLite stores under the
**per-agent `workspace_dir`**, but the writers persist them to the **shared
`config.data_dir`**:

- Discord archive — written at `crates/zeroclaw-channels/src/orchestrator/mod.rs`
  via `SqliteMemory::new_named("sqlite", &config.data_dir, "discord")`.
- Sessions — written via `make_session_backend(&config.data_dir, …)` in the
  daemon, orchestrator and gateway.

So the tools read an **empty/wrong DB** (`discord_search` returns nothing;
`sessions_list`/`sessions_history` miss real sessions) and litter stray empty
`memory/discord.db` + `sessions/sessions.db` files into every agent workspace.
This is the residual read-side instance of the `workspace_dir`↔`data_dir` family
(#7541 / #7121 / #7573) — those merges made `workspace_dir` reliably per-agent,
which surfaced it.

## Change

Point the two readers at the shared `config.data_dir` the writers use
(`crates/zeroclaw-runtime/src/tools/mod.rs`):

```rust
// discord_search
- SqliteMemory::new_named("sqlite", workspace_dir, "discord")
+ SqliteMemory::new_named("sqlite", &config.data_dir, "discord")

// sessions_*
- make_session_backend(workspace_dir, &config.channels.session_backend)
+ make_session_backend(&config.data_dir, &config.channels.session_backend)
```

`config: Arc<Config>` is already a parameter of `all_tools_with_runtime` and at
every production call site is the same runtime config whose `data_dir` the
writers use, so reads and writes now agree. Only these two *shared* stores
change; all other `workspace_dir` tool plumbing (git/file/linkedin/…) is
correctly per-agent and is left untouched. A regression test builds the tool
registry with `data_dir` ≠ `workspace_dir` and asserts neither store is created
under the per-agent workspace.

---

## Testing

- `discord_slash_state`: 6 unit tests (round-trip persist/load, app-id-mismatch →
  default, no-workspace no-op, retry-after preserves the prior fingerprint,
  corrupt file → default, `retry_after` body→header precedence).
- `reconcile_returns_rate_limited_on_post_429` (new) + the 3 pre-existing
  reconcile tests still pass (the `ReconcileOutcome` return is backward-compatible).
- `shared_store_tools_open_data_dir_not_per_agent_workspace` (new) — the
  read-path regression above.
- `cargo fmt` clean · `cargo clippy --all-targets -- -D warnings` clean ·
  `cargo test -p zeroclaw-channels --features channel-discord` /
  `cargo test -p zeroclaw-runtime` green.

### DIFF
--- a/crates/zeroclaw-channels/src/discord.rs
+++ b/crates/zeroclaw-channels/src/discord.rs
@@ -100,11 +100,6 @@ pub struct DiscordChannel {
     /// Resolves skill-derived commands to register alongside `/ask`.
     /// `None` (or an empty resolution) = `/ask` only.
     slash_command_resolver: Option<DiscordSlashCommandResolver>,
-    /// Fingerprint of the last *successfully* reconciled command set —
-    /// READYs with an unchanged set skip the REST round-trip; failures
-    /// reset it so the next READY retries instead of wedging. Arc'd
-    /// because reconciliation runs in a spawned task.
-    registered_commands_hash: Arc<Mutex<Option<u64>>>,
 }
 
 /// Credentials needed to answer a deferred interaction later: the followup
@@ -186,7 +181,6 @@ impl DiscordChannel {
             slash_commands: false,
             pending_interactions: Arc::new(Mutex::new(HashMap::new())),
             slash_command_resolver: None,
-            registered_commands_hash: Arc::new(Mutex::new(None)),
         }
     }
 
@@ -1807,6 +1801,25 @@ fn command_projection(cmd: &serde_json::Value) -> serde_json::Value {
 /// Discord REST base; injectable in `reconcile_slash_commands` for tests.
 const DISCORD_API_BASE: &str = "https://discord.com/api/v10";
 
+/// Outcome of a slash-command reconcile pass.
+#[derive(Debug)]
+enum ReconcileOutcome {
+    /// The command set was reconciled (or was already current).
+    Reconciled,
+    /// Discord rate-limited the pass; the caller must persist this cooldown and
+    /// not retry until the given unix-seconds deadline.
+    RateLimited { until: i64 },
+}
+
+/// Turn a `429` response into a unix-seconds deadline before which no further
+/// reconcile should run, reading Discord's `retry_after` body / headers.
+async fn rate_limit_deadline(resp: reqwest::Response) -> i64 {
+    let now = crate::discord_slash_state::now_unix();
+    let headers = resp.headers().clone();
+    let body = resp.json::<serde_json::Value>().await.ok();
+    crate::discord_slash_state::retry_after_deadline(&headers, body.as_ref(), now)
+}
+
 /// Reconcile the application's global commands with the desired set:
 /// upsert each desired command (POST upserts by name) and delete stale
 /// skill-shaped commands left over from uninstalled skills. Commands
@@ -1826,7 +1839,7 @@ async fn reconcile_slash_commands(
     app_id: &str,
     desired: &serde_json::Value,
     api_base: &str,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<ReconcileOutcome> {
     let base = format!("{api_base}/applications/{app_id}/commands");
     let auth = format!("Bot {bot_token}");
     let Some(desired) = desired.as_array() else {
@@ -1848,6 +1861,11 @@ async fn reconcile_slash_commands(
         .send()
         .await
         .map_err(reqwest::Error::without_url)?;
+    if resp.status() == reqwest::StatusCode::TOO_MANY_REQUESTS {
+        return Ok(ReconcileOutcome::RateLimited {
+            until: rate_limit_deadline(resp).await,
+        });
+    }
     if !resp.status().is_success() {
         anyhow::bail!("listing global commands failed ({})", resp.status());
     }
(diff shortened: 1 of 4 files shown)


singlerider (approved):
First review (singlerider). Read the PR body, the full diff, and cross-checked both fixes against the writers on current `master`. CI is green on this head (78f0bc7), MERGEABLE, no prior reviews or open threads. Two well-separated `data_dir` durability fixes, each with its own regression coverage. Approving.

### 🟢 What looks good: the reconcile state persistence is correctly fail-safe in every direction

The move from an in-process `Arc<Mutex<Option<u64>>>` to a persisted `SlashReconcileState` is the right shape, and each failure mode is handled deliberately rather than by accident:

- A hard error leaves the persisted state untouched, so the next READY's fresh fingerprint differs from the stored one and a retry happens naturally, no forced reset, no wedge.
- A `429` is promoted to a first-class `ReconcileOutcome::RateLimited { until }` instead of a `bail!`, and `record_retry_after` preserves the prior fingerprint (the desired set did not change, only the server asked us to wait), so the cooldown gates the next attempt rather than a needless re-diff. The first upsert 429 stops the loop instead of hammering the remaining commands into the same limit.
- `retry_after_deadline` reads body `retry_after` → `Retry-After` → `X-RateLimit-Reset-After` in that order, rounds up, and floors at 1s, so a missing or zero value still yields a real cooldown.
- The write is atomic (unique temp name with pid + atomic seq, then rename), the dir is 0700, and a torn read degrades to "no state" → a harmless full reconcile, which is exactly the failure direction you want for a cache. App-id mismatch on load (swapped bot token) forces a full reconcile rather than trusting a stale fingerprint.

The module is feature-gated behind `channel-discord` in `lib.rs`, the outcome is a typed enum rather than a sentinel, and the 7 unit tests plus the wiremock `reconcile_returns_rate_limited_on_post_429` cover the branches that matter (round-trip, app-id mismatch, no-workspace no-op, corrupt file, body/header precedence, fingerprint-kept-across-backoff, exclusive deadline).

### 🟢 What looks good: the shared-store read fix matches the writers exactly

I traced both stores end to end. The Discord archive writer persists to `&config.data_dir` (orchestrator/mod.rs:6764) and all three session writers (daemon mod.rs:324, gateway lib.rs:1146, orchestrator mod.rs:8327) build the backend under `&config.data_dir`. Pointing `discord_search` and the `sessions_*` readers at `&config.data_dir` makes reads and writes agree, and `config` is already a parameter of `all_tools_with_runtime` carrying the canonical runtime config at every production call site (verified at the gateway site: arg 1 is `Arc::new(config.clone())` and `root_config` is the same `config`). Scoping the change to only the two genuinely-shared stores and leaving the per-agent `workspace_dir` plumbing alone is the correct boundary, and `shared_store_tools_open_data_dir_not_per_agent_workspace` pins it with `data_dir` deliberately distinct from `workspace_dir`, asserting no stray `memory/` or `sessions/` dir is created under the workspace. This closes the read-side tail of the `workspace_dir`↔`data_dir` family (#7541 / #7121 / #7573).

PR #7785, opened 2026-06-16
feat(config): find_all_references/plan_delete foundation for typed delete-with-cascade (#7175)

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds `zeroclaw-config::alias_refs`, the shared reference-discovery foundation #7175 mandates. The cascade logic must live in `zeroclaw-config` (the canonical registry) so every surface (TUI/web/RPC/CLI) routes through one implementation rather than re-walking the schema.
  - `find_all_references(cfg, kind, alias) -> Vec<RefSite>` enumerates every config site that references an aliased entry of `kind` (provider / agent / channel), each classified **HARD** (mandatory — deletion must refuse) or **SOFT** (removable — deletion scrubs it), carrying the `ScrubAction` a later mutating pass would apply.
  - `plan_delete(cfg, kind, alias) -> ImpactReport` folds them into the dry-run impact report (`blockers` / `scrubs` / `allowed`) a surface renders before confirming a destructive action.
  - Types: `AliasKind`, `ProviderCategory`, `RefStrength`, `ScrubAction`, `RefSite`, `OwnedArtifact`, `ImpactReport`.
- **Scope boundary:** **Read-only** — never mutates `Config`. No `delete_with_cascade` (the mutating op), no surface wiring, no owned-state (memory/workspace/infra) gathering — those are follow-up PRs. Covers the provider/agent/channel reference inventory; TTS/transcription are included for providers.
- **Blast radius:** A new module with **no callers** in this PR — zero behavioural/runtime change. Pure addition.
- **Linked issue(s):** Related #7175 (this is the foundation slice; it does not close the issue).

## Notable correctness — mirrors `Config::validate()` exactly, including trim

`find_all_references` is contractually a referrer-for-referrer mirror of `Config::validate()`. `validate()` **trims** most stored refs before resolving them (`model_provider` schema.rs:17143, classifier :17227, tts :17217, transcription :17221, model/embedding routes :16549/:16595, fallback walk :16177, channels :17183, peer-group channel :17418, heartbeat :16338, delegates :17331), so a whitespace-padded TOML value like `model_provider = " anthropic.default "` passes validation. This module therefore trims those same sites — otherwise it would silently miss the ref and a later delete would leave a dangling reference (the exact failure this feature exists to prevent). The three `AgentAlias`-keyed refs `validate()` compares **raw** (`workspace.access` :17358, `read_memory_from` :17382, `peer_groups.agents` :17453) are deliberately matched raw — trimming *those* would itself drift. `RefSite.raw_value` preserves the actual stored text. Regression tests `whitespace_padded_provider_refs_are_found` and `agent_ref_trimming_mirrors_validate` pin both directions.

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-config
```

  - `cargo fmt --all -- --check` → clean, no diff (whole workspace).
  - `cargo clippy --all-targets -- -D warnings` → clean (exit 0, 0 warnings) across the workspace.
  - `cargo test -p zeroclaw-config` →
    ```
    test result: ok. 869 passed; 0 failed; 0 ignored   (lib unit, incl. 12 in alias_refs)
    test result: ok.  90 passed; 0 failed; 0 ignored   (integration)
    ```
- **Beyond CI — what did you manually verify?** Ran a multi-agent adversarial review (completeness vs the reference inventory; correctness; mirror-fidelity to `validate()`; API/tests). It surfaced one real blocker — untrimmed comparisons drifting from `validate()`'s trim — fixed in this commit with regression tests. The trim/no-trim split was verified per-site against the actual `schema.rs`, not assumed.
- **If any command was intentionally skipped, why:** full-workspace `cargo test` not run locally — the change is a single additive module in one crate with no callers; ran the full changed-crate suite + workspace `fmt` + workspace `clippy`. CI runs the complete suite.

- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (tests use synthetic aliases — `anthropic`, `bot`, `crew`, etc.)

## Compatibility (required)

- Backward compatible? **Yes** (additive module, no callers, no public API removed/changed)
- Upgrade steps: none.

## Rollback

Low risk: `git revert <sha>` is the plan (a self-contained new module).


Audacity88 (changes requested):
I checked the live PR state, the two-file diff adding `zeroclaw-config::alias_refs` plus the `lib.rs` export, linked issue #7175 and its accepted hard-vs-soft policy comment, the current green PR checks, and the absence of prior comments or reviews. I did not rerun the author's cargo commands locally; this review is about two static delete-impact inventory gaps in the added read-only planner.

### 🟢 What looks good — Shared read-only foundation

The shared API shape, provider-side coverage, stable `RefSite` ordering, and explicit trim-vs-raw matching are useful pieces to have in one config-owned module. Keeping the surfaces pointed at a single inventory also matches the shared-delete direction from #7175.

### 🔴 Blocking — Last channel alias delete can leave bare-type peer groups dangling

`collect_channel_refs` intentionally skips `peer_groups.<g>.channel = "discord"` when deleting `channels.discord.main`, because the stored value is bare type rather than the dotted target. That is fine while another `channels.discord.*` alias remains, but it is not fine when `main` is the last alias of that type: deleting it removes the remaining `[channels.discord.*]` block, `Config::validate()` then finds no `channels.discord` map for the bare-type group, and the group fails validation even though `plan_delete` reported no hard blocker for it. The dry-run report is supposed to be the shared safety check before the later mutating delete, so this path needs to either report/refuse bare-type peer groups when the target alias is the last alias of the channel type, or otherwise prove the mutating delete cannot remove the final alias while the group remains.

### 🔴 Blocking — Agent delete omits channel ownership impact

The agent delete plan does not report the target agent's channel ownership impact. `collect_agent_refs` covers references to the agent, but this is owned config that becomes orphaned when the owning agent is removed. Existing runtime ownership is derived by `Config::agent_for_channel`, which returns the enabled agent whose `channels` list contains the channel ref, and #7175 calls out channel ownership as a hard agent-delete concern. With the current `collect_agent_refs`, deleting `agents.bot` can return no blocker for `agents.bot.channels = ["discord.main"]`, so a later config-only cascade could remove the owning agent and silently orphan the channel route while the impact report says the delete is allowed. Please surface those channel ownerships in the agent delete plan, probably as hard blockers/refuse entries unless the cascade also has an explicit repoint/prune policy.

Nillth:
Thanks — both gaps are real; fixed in the planner with tests.

**1. Last channel alias deleting out from under a bare-type peer group.** `collect_channel_refs` now detects when the target is the *sole* alias of its channel type (`get_map_keys("channels.<type>")` == `[alias]`), and in that case reports every bare-type `peer_groups.<g>.channel = "<type>"` as a HARD blocker — so `plan_delete` refuses rather than letting the mutating delete empty the type's block and dangle the group. While another `channels.<type>.*` alias survives it stays allowed (the group still resolves), matching the existing dotted-vs-bare rule.

**2. Agent channel ownership.** `collect_agent_refs` now surfaces the target agent's owned channels as HARD blockers, using `Config::agent_for_channel`'s ownership semantics (the enabled agent whose `channels` list contains the ref): for each channel the agent lists, if no *other* enabled agent also lists it, deleting the agent would leave the route with no owner, so it's a refuse entry. A disabled agent owns nothing, and a co-owned channel doesn't block — so this only fires when the delete would actually orphan a route. No repoint/prune policy yet, so refuse-by-default per the #7175 hard-vs-soft direction; if you'd prefer an explicit reassign step later, the impact report already names the exact `agents.<a>.channels[i]` to act on.

Both are read-only planner additions (HARD `RefSite`s), so the downstream `delete_with_cascade` arms refuse on them with no further change. Added `channel_delete_of_last_alias_blocks_bare_type_peer_group` and `agent_delete_blocks_on_solely_owned_channel` (covering the last-alias, co-owned, and disabled cases).

singlerider (approved):
Re-review (singlerider) at head 009c4048. Verified both of @Audacity88's blocking findings against the actual module code and the two new regression tests, cross-checked the trim-vs-raw matching against the cited `Config::validate()` and `Config::agent_for_channel` anchors, and confirmed the module is read-only with no callers. CI is green, branch is MERGEABLE. Both blockers are resolved. Approving on my own merits.

@Audacity88 your CHANGES_REQUESTED was filed at c90541a2, which predates the fix commit, so it still shows as active in the aggregate. Both gaps you named are addressed at the current head. Could you convert or dismiss when you get a moment.

### ✅ Resolved: last channel alias delete no longer dangles a bare-type peer group

`collect_channel_refs` now computes `removes_last_alias` by checking that the target is the sole key under `channels.<type>` (`get_map_keys(...).any(== alias) && all(== alias)`). When true, every bare-type `peer_groups.<g>.channel = "<type>"` is reported as HARD/Refuse, so the dry-run refuses rather than letting the mutating delete empty the type's block and leave the group dangling at `validate()` (anchor schema.rs:17432-17439 cited in-code). The dotted-vs-bare rule is preserved for the non-last case. `channel_delete_of_last_alias_blocks_bare_type_peer_group` pins both directions: last-alias delete is refused on `peer_groups.crew.channel`, and adding a second `discord.*` alias makes the same delete allowed.

### ✅ Resolved: agent delete now reports solely-owned channel ownership as a blocker

`collect_agent_refs` surfaces the target agent's owned channels using `Config::agent_for_channel`'s exact semantics (enabled-only, untrimmed match): a channel the target lists is HARD/Refuse only when no other enabled agent also lists it, so the block fires precisely when the delete would orphan a route. A disabled target owns nothing, and a co-owned channel does not block. That matches #7175's hard-vs-soft direction (refuse-by-default absent a repoint/prune policy), and the report names the exact `agents.<a>.channels[i]` for a future reassign step. `agent_delete_blocks_on_solely_owned_channel` covers the sole-owner, co-owned, and disabled cases.

### 🟢 What looks good: the trim-vs-raw mirror of validate() is the right correctness contract, and it is documented and tested per site

The load-bearing property of a reference inventory is that it sees every ref `validate()` sees, no more and no less. This module mirrors `validate()` referrer-for-referrer with the trim split made explicit per arm: the refs `validate()` trims before resolving (model_provider, classifier, tts, transcription, model/embedding routes, fallback, channels, peer-group channel, heartbeat, delegates) are trimmed here, while the three `AgentAlias`-keyed refs it compares raw (workspace.access, read_memory_from, peer_groups.agents) are matched raw, with `raw_value` preserving the stored text. `whitespace_padded_provider_refs_are_found` and `agent_ref_trimming_mirrors_validate` pin both directions, so a future drift in either set is caught. Read-only, no `Config` mutation, no callers, deterministic site ordering. Clean foundation slice for the cascade work.

Audacity88 (changes requested):
Thanks for the update, @Nillth. I re-checked current head `009c40483886ed3f9da150ff89c06d5b1a4a8991` against the prior Audacity88 review, linked issue #7175, the current two-file diff, the relevant `Config::validate()` peer-group checks, and the current green CI. I did not rerun local Cargo validation for this review-only pass. This PR is still read-only and has no callers yet; the issue below is about the dry-run planner contract that later delete surfaces will render and apply.

### ✅ Resolved — Prior hard-ref gaps are now covered

The two issues from my previous review look addressed. `collect_channel_refs` now refuses the last-alias delete that would leave a bare-type `peer_groups.<g>.channel = "discord"` with no configured `channels.discord.*` block, and `collect_agent_refs` now reports solely-owned enabled agent channel routes as hard blockers. The added tests cover those two cases and the co-owned/disabled-agent branches.

### 🔴 Blocking — Channel delete can still leave a bare-type group member without the channel type

There is still a related invalid-config path in `collect_channel_refs`' impact plan. The new bare-type peer-group guard is correct for the type-level dependency: if `main` is the last configured alias of the whole `discord` type, `peer_groups.<g>.channel = "discord"` must block. But there is a second, per-member dependency in `Config::validate()`: for a bare-type peer group, each listed member must still have at least one `agents.<member>.channels[]` entry starting with that channel type.

Concrete shape: configure both `channels.discord.main` and `channels.discord.backup`, set `agents.bot.channels = ["discord.main"]`, and add `peer_groups.crew.channel = "discord"` with `bot` as a member. `plan_delete(Channel { channel_type: "discord" }, "main")` is currently allowed because `backup` keeps the channel type block present, and it only reports `agents.bot.channels[0]` as a soft scrub. The type-level peer-group channel remains valid, but the member-level invariant does not: after the later mutating delete applies that scrub, `bot` has no `discord.*` channel left, so `Config::validate()` fails at `peer_groups.crew.agents[0]`.

The dry-run impact report is the shared safety gate for the later destructive operation: `allowed` plus the `ScrubAction`s are the public contract future surfaces will render before they confirm the delete. So the read-only planner needs to account for this member-level dependency too, instead of classifying the member's only type-matching channel as an always-safe soft scrub. Please either hard-block a channel delete when a bare-type peer-group member would lose its only channel of that type, or add an explicit scrub/refuse policy for that peer-group membership with regression coverage.

Nillth:
Good catch — fixed.

You're right that the type-level guard only covered the last-alias case; a bare-type peer group also imposes a *per-member* dependency (each member must keep at least one `<type>.*` channel), and soft-scrubbing a member's only matching channel while another alias keeps the type alive leaves it dangling at `peer_groups.<g>.agents[i]`.

`collect_channel_refs` now HARD-blocks that too: when a bare-type group member's only `<type>.*` channel is the target, it's reported as a `peer_groups.<g>.agents[i]` blocker (refuse) rather than a soft scrub — the companion to the type-level last-alias guard. The survivor test mirrors `validate()`'s untrimmed bare-membership check exactly. I **moved this member-level guard forward from the channel-delete arm (the follow-up PR in the series)** into the foundation here, so the read-only planner's contract is complete on its own.

Added a regression test for your exact shape: two `channels.discord.*` aliases, `agents.bot.channels = ["discord.main"]`, `peer_groups.crew.channel = "discord"` with `bot` a member → deleting `discord.main` is now refused at `peer_groups.crew.agents[0]`; and the same with `bot` also holding `discord.backup` stays allowed.

Audacity88 (approved):
Thanks for the follow-up, @Nillth. I re-reviewed current head `2892d5a4af8385180d4f62d73751725c7dccf9d3` against the prior Audacity88 reviews, the author replies, SingleRider's approval, linked issue #7175, the current two-file diff, the relevant `Config::validate()` / `Config::agent_for_channel` behavior, and the current green CI. I did not rerun local Cargo validation for this review-only pass.

### ✅ Resolved — Bare-type peer-group member orphan is now blocked

The third Audacity88 blocker is addressed at this head. `collect_channel_refs` now handles the case where a bare-type peer group still has another `channels.<type>.*` alias configured, but the specific member would lose its only channel of that type. In that shape, the planner now reports a hard blocker at `peer_groups.<group>.agents[i]` instead of treating the member's `agents.<member>.channels[]` entry as an always-safe scrub. That matches the `Config::validate()` invariant for bare-type peer groups, where each member must keep at least one matching `<type>.*` channel. The new `channel_delete_blocks_when_bare_group_member_loses_only_channel` regression covers both sides: deleting `discord.main` is refused when it is the member's only Discord channel, and allowed when the member also has `discord.backup`.

### ✅ Resolved — Earlier hard-ref gaps remain covered

The first two blockers remain resolved. Deleting the last alias of a channel type refuses when a bare-type peer group would be left without any configured channel block, and deleting an agent reports solely-owned enabled channel routes as hard blockers. Those reports preserve the #7175 rule that the dry-run impact plan must refuse deletes that would leave mandatory provider, agent, or channel references dangling unless a later operation has an explicit repoint or prune policy.

### 🟢 What looks good — The planner stays read-only and source-of-truth aligned

This is still a clean foundation slice: it exports the read-only `alias_refs` module, has no callers or config mutation in this PR, and keeps future surfaces pointed at one config-owned reference inventory instead of re-walking the schema independently. The trim-vs-raw split is also documented at the ref sites and pinned with tests, which is important because the impact report only stays trustworthy if it sees the same references that validation sees.

### 🟡 Warning — Refresh before merge readiness

This approval clears my stale `CHANGES_REQUESTED` review, but it is not a merge-ready statement. The branch is currently behind `master` on a high-risk config surface, and the PR body rollback section still says `Low risk` even though the live label is `risk: high`. Before merge, this still needs a branch refresh with green CI, an exact merge-result smoke, or an explicit stale-risk acceptance for this PR, plus a rollback-field update that matches the high-risk label.

PR #7786, opened 2026-06-16
perf(skills): cache skill-directory loads (content-validated, audit-correct)

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `load_skills_from_directory` and `load_open_skills_from_directory` run a recursive read **plus a full security audit** (content scan + manifest parse) of every skill subdirectory on **every** call — on every prompt build *and* every `read_skill` invocation (and across the channels orchestrator and delegate paths) — so the same directories are re-walked and re-audited constantly even when nothing changed on disk.
  - Adds a process-global memoization layer (`skills::cache`) keyed by `(canonical dir, allow_scripts, loader-tag)`. Both directory loaders resolve through it; the audit + parse only re-run when the directory's contents actually change.
  - **Freshness is a content digest** — the signature hashes the *bytes* of every file reachable under the dir (symlinks by target, never followed). Because the cache sits in front of the security audit, the key is deliberately tied to the audited bytes, not to metadata an edit can preserve: any change the audit would care about (edited `SKILL.md`/`SKILL.toml`, a flipped script, a retargeted symlink) yields a different signature and forces a re-audit. The digest reads each file once, but a cache *hit* then skips the audit's content scan, its regex/script/symlink checks, the Markdown/TOML parsing, and the loader's redundant re-reads — so it stays a net win without weakening the audit boundary.
  - `SkillsService::{write_skill, scaffold_skill, remove_skill}` call `cache::invalidate()` so operator-driven edits show up on the next load immediately.
  - Runtime kill-switch `ZEROCLAW_SKILLS_CACHE_ENABLED` (default on; `0`/`false`/`no`/`off` → pre-cache always-re-audit behavior).
- **Scope boundary:** does **not** change skill *resolution* (which skills exist / which dirs are scanned), the audit rules, the `Skill` shape, or any public signature. Pure memoization of two existing pure functions. Does not touch the `read_skill` resolution divergence — that's #7245.
- **Blast radius:** every caller of the two directory loaders (prompt build, `read_skill`, channels orchestrator, delegate). Behavior is identical on a miss; on a hit the only difference is the saved audit/parse work. Cache memory is bounded by the number of *distinct* skill directories ever loaded (workspace + configured bundles + open-skills), which is small and stable per install.
- **Linked issue(s):** Related #7025, Related #7245 — those address the separate `read_skill` *resolution* bug; once #7245 routes `read_skill` through `load_skills_for_agent` it also benefits from this cache. No existing issue tracks skill-load caching.
- **Labels:** `runtime`, `skills` (current) — please also apply `risk: high`, `size: S`. This is a default-on cache in front of skill **security auditing** in `crates/zeroclaw-runtime`, so it warrants the high-risk classification.

```bash
cargo fmt --all -- --check
```

  - `cargo fmt --all -- --check` → clean (no diff).
  - `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` → exit 0, no warnings.
  - `cargo test -p zeroclaw-runtime` → `test result: ok. 2205 passed; 0 failed; 1 ignored`.
  - Cache unit tests → `8 passed; 0 failed`. Notably `same_length_same_mtime_edit_still_busts_cache`: edits a skill file preserving **both** length and mtime (via `filetime`) and asserts the cache still reloads — the exact audit-boundary case a metadata-only key would miss. It fails on the old mtime+length signature and passes on the content digest.
- **Beyond CI — what did you manually verify?** Confirmed every caller of both loaders only depends on the returned `Vec<Skill>`, not on per-call audit side effects; confirmed the digest is recomputed before each gate so any audited-content change forces a re-audit; tempdir-unique keys mean no cross-test bleed (full suite green).
- **If any command was intentionally skipped, why:** Scoped `clippy`/`test` to `zeroclaw-runtime` (only crate with code changes). Whole-workspace `cargo test` not run locally; CI runs the full battery (green).

- New permissions, capabilities, or file system access scope? **No** — same directories, same reads.
- **Audit-boundary note:** the audit runs *behind* the cache, so the freshness key must prove the audited bytes are unchanged. It does: the signature is a **content digest** of every file under the dir, so an edit that changes content — even one preserving length and mtime — produces a different signature and re-audits. The only residual is a 64-bit hash collision, which is not a practical forgery vector. `SkillsService` writes also invalidate explicitly, and `ZEROCLAW_SKILLS_CACHE_ENABLED=0` disables the cache entirely.

## Compatibility (required)

- Backward compatible? **Yes** — public function signatures unchanged; behavior identical on a cache miss.
- Config / env / CLI surface changed? **Yes (additive, opt-out)** — adds one optional env var `ZEROCLAW_SKILLS_CACHE_ENABLED`; absent → cache on (the new default), existing deployments need no change.
- Upgrade steps: none.

- **Fast rollback command/path:** runtime, no redeploy — set `ZEROCLAW_SKILLS_CACHE_ENABLED=0` and restart to fall back to pre-cache behavior. Source rollback: `git revert <sha>` (self-contained, no migration).
- **Feature flags or config toggles:** `ZEROCLAW_SKILLS_CACHE_ENABLED` (default on; `0`/`false`/`no`/`off` disables).
- **Observable failure symptoms:** a skill change not reflected until restart would indicate an invalidation gap — confirm an edited `SKILL.md` is picked up after a prompt rebuild; flip the kill-switch to confirm. No new metric/alert added.

## Files
- `crates/zeroclaw-runtime/src/skills/cache.rs` (new — content-digest cache, invalidation, kill-switch, tests)
- `crates/zeroclaw-runtime/src/skills/mod.rs` (route both loaders through the cache)
- `crates/zeroclaw-runtime/src/skills/service.rs` (invalidate on write/scaffold/remove)
- `crates/zeroclaw-runtime/Cargo.toml` (+ `filetime` dev-dependency for the mtime-preserving regression)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14515,6 +14515,7 @@ dependencies = [
  "dialoguer",
  "directories",
  "encoding_rs",
+ "filetime",
  "flate2",
  "fluent",
  "futures-util",
--- a/crates/zeroclaw-runtime/Cargo.toml
+++ b/crates/zeroclaw-runtime/Cargo.toml
@@ -112,6 +112,7 @@ rcgen = "0.13"
 tempfile = "3.26"
 tokio = { version = "1.50", features = ["rt-multi-thread", "macros"] }
 scopeguard = "1.2"
+filetime = "0.2"
 
 # Channel features (forwarded to zeroclaw-channels)
 
--- a/crates/zeroclaw-runtime/src/skills/cache.rs
+++ b/crates/zeroclaw-runtime/src/skills/cache.rs
@@ -90,7 +90,9 @@ fn cache_enabled() -> bool {
 /// (symlinks). Never follows symlinks, so it can't loop on a cycle and matches
 /// the auditor's no-follow stance. Tying the key to content — not metadata an edit
 /// can preserve — is what keeps a cached "clean" audit verdict from outliving the
-/// bytes it audited. Returns `None` when `dir` is absent or any entry can't be
+/// bytes it audited. Only *regular* files are opened: a non-regular entry (FIFO,
+/// socket, device) makes this return `None` so we never block opening it just to
+/// build a key. Returns `None` too when `dir` is absent or any entry can't be
 /// read; callers treat that as "do not cache" rather than trust a partial digest.
 fn dir_signature(dir: &Path) -> Option<u64> {
     if !dir.exists() {
@@ -119,10 +121,18 @@ fn dir_signature(dir: &Path) -> Option<u64> {
                 entries.insert(path, (2, h.finish()));
             } else if file_type.is_dir() {
                 stack.push(path);
-            } else {
+            } else if file_type.is_file() {
                 // Decline to cache rather than fingerprint a file we can't read.
                 let digest = hash_file_contents(&path)?;
                 entries.insert(path, (1, digest));
+            } else {
+                // Non-regular entry (FIFO, socket, device, ...). Opening a FIFO
+                // for read blocks on a writer, so probing it just to build the
+                // cache key could hang skill loading / prompt building — a far
+                // wider open surface than the uncached loader, which only reads
+                // the manifest it parses. Never open it: decline to cache this
+                // directory and let the uncached path handle whatever it is.
+                return None;
             }
         }
     }
@@ -344,6 +354,43 @@ mod tests {
             calls.load(Ordering::SeqCst),
             2,
             "content change under identical mtime+length must re-audit"
+        );
+    }
+
+    // Audit-boundary regression (review of #7786, round 2): a FIFO (or other
+    // non-regular entry) inside a skills dir must never be opened while building
+    // the cache key — opening a FIFO for read blocks on a writer and would hang
+    // skill loading / prompt building. `dir_signature` must bail to `None` so the
+    // directory simply bypasses the cache. If this regresses, the test hangs.
+    #[cfg(unix)]
+    #[test]
+    fn non_regular_entry_bypasses_cache_without_hanging() {
+        invalidate();
+        let tmp = TempDir::new().unwrap();
+        let skills_dir = tmp.path().join("skills");
+        write(&skills_dir, "alpha", "# Alpha\n");
+        let fifo = skills_dir.join("alpha/pipe");
+        let status = std::process::Command::new("mkfifo")
+            .arg(&fifo)
+            .status()
+            .expect("mkfifo should be available on unix test hosts");
+        assert!(status.success(), "mkfifo failed");
+
+        let calls = AtomicUsize::new(0);
+        let load = || {
+            calls.fetch_add(1, Ordering::SeqCst);
+            Vec::<Skill>::new()
+        };
+
+        // Must return promptly (no hang) and, because the dir can't be signed,
+        // run the loader every time instead of caching.
+        cached_load(&skills_dir, false, "test", load);
+        cached_load(&skills_dir, false, "test", load);
+
+        assert_eq!(
+            calls.load(Ordering::SeqCst),
+            2,
+            "a directory containing a non-regular entry must bypass the cache"
         );
     }
 
--- a/crates/zeroclaw-runtime/src/skills/cache.rs
+++ b/crates/zeroclaw-runtime/src/skills/cache.rs
@@ -1,0 +1,372 @@
+//! Process-global cache for skill-directory loads.
+//!
+//! [`super::load_skills_from_directory`] and [`super::load_open_skills_from_directory`]
+//! are pure functions of `(dir, allow_scripts, filesystem state)`, but each call
+//! does a recursive read *and* a full security audit (content scan + parse) of
+//! every skill subdirectory. They run on every prompt build and every
+//! `read_skill` invocation, so the cost recurs constantly even when nothing on
+//! disk has changed.
+//!
+//! This module memoizes the result keyed by `(canonical dir, allow_scripts, tag)`
+//! and validates freshness with a cheap, **stat-only** directory signature: any
+//! add / remove / rename / content edit changes a file's mtime or length (or a
+//! symlink's target), which changes the signature and forces a re-audit. A stale
+//! "clean" audit verdict can therefore never be served from cache.
+//!
+//! [`invalidate`] gives the [`super::SkillsService`] an explicit hook to drop the
+//! cache immediately after a write, so an added/edited/removed skill is picked up
+//! on the very next load without waiting on anything.
+//!
+//! Kill-switch: the cache is on by default; setting `ZEROCLAW_SKILLS_CACHE_ENABLED`
+//! to a falsey value (`0` / `false` / `no` / `off`) forces every load to re-walk
+//! and re-audit, i.e. the exact pre-cache behavior. This is a runtime off-ramp if
+//! the cache is ever suspected of serving stale results.
+//!
+//! Caveat: the signature trusts filesystem metadata, so a deliberate edit that
+//! preserves both mtime and length (e.g. an attacker resetting mtime via
+//! `utimes`) would not be detected. This matches the staleness model of build
+//! tools like `cargo`/`make`; anyone with write access to the skills directory
+//! already controls which skills exist, so the audit is a guard against
+//! accidental/unreviewed content rather than against an attacker who can forge
+//! inode metadata.
+
+use super::Skill;
+use std::collections::hash_map::DefaultHasher;
+use std::collections::{BTreeMap, HashMap};
+use std::hash::{Hash, Hasher};
+use std::path::{Path, PathBuf};
+use std::sync::{OnceLock, RwLock};
+use std::time::UNIX_EPOCH;
+
+#[derive(PartialEq, Eq, Hash, Clone)]
+struct CacheKey {
+    dir: PathBuf,
+    allow_scripts: bool,
+    /// Distinguishes loaders that may share a directory path (workspace vs
+    /// open-skills) so their cached entries never collide.
+    tag: &'static str,
+}
+
+struct CacheEntry {
+    signature: u64,
+    skills: Vec<Skill>,
+}
+
+fn cache() -> &'static RwLock<HashMap<CacheKey, CacheEntry>> {
+    static CACHE: OnceLock<RwLock<HashMap<CacheKey, CacheEntry>>> = OnceLock::new();
+    CACHE.get_or_init(|| RwLock::new(HashMap::new()))
+}
+
+/// Best-effort canonicalization so two spellings of the same directory share an
+/// entry. Falls back to the path as given when the dir can't be canonicalized.
+fn canonical(dir: &Path) -> PathBuf {
+    std::fs::canonicalize(dir).unwrap_or_else(|_| dir.to_path_buf())
+}
+
+const CACHE_ENABLED_ENV: &str = "ZEROCLAW_SKILLS_CACHE_ENABLED";
+
+/// Pure kill-switch decision split from the env read so it stays testable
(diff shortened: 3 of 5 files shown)


Audacity88 (changes requested):
Review context: I checked PR #7786 at head 1f27d7a against the live PR state, linked #7025 / #7245 context, the current diff, the green GitHub checks, the existing skill loader, and the current skill audit path. I also tested it locally on a #7734 + #7786 stack, since the two PRs touch adjacent skill editor/load paths.

@Nillth thanks for taking on the repeated skill-load cost. The high-level shape is good: caching the two directory loaders behind the existing loader boundary is the right place to attack the overhead, and the explicit `SkillsService` invalidation hooks are useful. I need one freshness issue fixed before I would approve this default-on cache.

### 🟢 What looks good — the cache is attached to the right loader boundary

Keeping the cache behind `load_skills_from_directory` and `load_open_skills_from_directory` preserves the public loader signatures and keeps `allow_scripts` plus the workspace/open-skills loader tag in the cache key. The service-side invalidation calls after write, scaffold, and remove also cover the normal operator-driven mutation path, so dashboard/CLI edits should refresh on the next load.

Local validation passed on the stacked branch:

- `git diff --check origin/master...HEAD`
- `npm run build` in `web/`
- `zc-cargo test -p zeroclaw-runtime skills::cache` — 7 passed
- `zc-cargo test -p zeroclaw-runtime skills::document` — 10 passed

I also did a live Browser pass against a disposable gateway/config. Existing `tags: [coding]` loaded in the skill editor, toggling the Slash command saved `slash` into frontmatter, adding a `review` tag round-tripped, reopening the editor immediately showed the saved state, creating a second skill appeared immediately, and directly editing that second `SKILL.md` on disk was picked up after reload. So the normal editor/scaffold/external-edit path looked good.

### 🔴 Blocking — the stat-only signature can outlive the bytes that were audited

`dir_signature()` hashes each entry path with file mtime and length, then `cached_load()` returns the cached `Vec<Skill>` whenever that signature matches. That means a `SKILL.md`, `SKILL.toml`, or other audited file can change while preserving both length and mtime, and the next load will still reuse the previous clean audit result and previously parsed skill values without invoking the loader again.

The PR body and module comment do document the forged-metadata caveat and the explicit `SkillsService` invalidations cover the normal operator mutation path. Even so, the summary still says “any ... content edit changes the signature” and that a cached clean audit verdict can never outlive the bytes it audited. That stronger guarantee is what the implementation does not provide. Because this cache sits behind the security audit and is default-on, “someone with write access already controls the skills directory” is not enough by itself: the audit is the mechanism that decides whether changed content is safe to serve, including script-like files, symlinks, markdown links, TOML validity, and manifest policy.

Please make the freshness key prove the audit boundary, not only ordinary editor behavior. A content digest of the audited files would be the direct fix. Another acceptable shape would need an equivalent regression that edits an audited skill file while preserving length and mtime and still forces a reload/re-audit. If the implementation intentionally keeps a metadata-only signature, then this should not be described or shipped as a default-on audit-correct cache.

### 🟡 Warning — risk metadata should match the runtime/audit boundary

The PR body is missing the required `Labels:` snapshot line, and the live labels currently only show `runtime` and `skills`. Since this change is in `crates/zeroclaw-runtime` and puts a default-on cache in front of skill security auditing, the public risk framing should be refreshed before merge. I would treat this as high-risk runtime/security-boundary work, with the body and labels aligned to that classification.

WareWolf-MoonWall (commented):
The performance motivation is real — scanning a skill directory on every invocation adds latency — and the implementation structure (process-global `RwLock<HashMap>`, kill-switch env var, explicit `invalidate()` calls from mutation paths) is generally sound. However, @Audacity88 has an active CHANGES_REQUESTED that I agree with substantively, so I'm posting a comment rather than approving over their block.

### ℹ️ Note — @Audacity88 holds an active CHANGES_REQUESTED

@Audacity88's blocker on the stat-only freshness mechanism is technically correct. `dir_signature()` hashes `path + mtime_ns + len` for each entry — it does not hash file content. A `SKILL.md` or `SKILL.toml` can be rewritten with the same length and a `utimes(2)` timestamp reset, producing an identical signature. On the next load, the cache returns the previously-audited result without re-auditing the new content. This breaks the guarantee stated in the PR title ("signature-validated, audit-correct") and in the module doc comment ("a cached clean audit verdict can never outlive the bytes it audited").

The fix is straightforward: add a content digest of audited files to the freshness key (hash the bytes of `SKILL.md`/`SKILL.toml` for each skill), or add a regression test demonstrating that a same-size, mtime-reset content substitution forces a cache miss. Either approach closes the gap.

### 🟢 What looks good — Kill-switch

`ZEROCLAW_SKILLS_CACHE_ENABLED=0/false/no/off` completely bypasses the cache and falls back to pre-cache behavior. That's the right safety net for a default-on process-global optimization.

### 🟢 What looks good — Invalidation wiring

`cache::invalidate()` is called from every mutation path — `write_skill`, `scaffold_skill`, `remove_skill`. The paths that can change skill content all trigger invalidation. The freshness issue is specifically about external mutations (filesystem writes bypassing these paths), which is the scenario the content digest would cover.

### 🟢 What looks good — No SSOT violation

The cache stores a derived `Vec<Skill>` from filesystem state, not a copy of any config value. This is a performance materialized view, not a config duplication.

Nillth:
@Audacity88 thanks — you're right, fixed at `8b1d06895d`.

The freshness key is now a **content digest**: `dir_signature()` hashes the bytes of every file under the skill dir (symlinks by target, never followed) instead of mtime+length. Any change the audit cares about — edited `SKILL.md`/`SKILL.toml`, a flipped script, a retargeted symlink — changes the signature and re-audits. The metadata-preserving gap (and the forged-mtime case) is gone; the only residual is a 64-bit hash collision.

Still a net win: the digest reads each file once, but a cache hit then skips the audit's content scan, its regex/script/symlink checks, the Markdown/TOML parsing, and the loader's redundant re-reads.

Added the regression you asked for — `same_length_same_mtime_edit_still_busts_cache` edits an audited file preserving **both** length and mtime (`filetime`) and asserts a reload; fails on the old signature, passes on the digest.

Also addressed the metadata: body reframed to content-validated (dropped the over-claimed caveat), added the `Labels:` line, and flagged `risk: high` since it's a default-on cache in front of skill auditing. `cargo test -p zeroclaw-runtime` → 2205 passed; clippy `-D warnings` clean.

Audacity88 (dismissed):
Review context: I re-checked PR #7786 at head 8b1d06895d against the live PR state, the author's follow-up comment, formal reviews, inline comments, related #7025 / #7245 context, the current diff, the skill cache/audit/loader paths, PR body and labels, and visible GitHub CI. I did not run local Cargo validation for this review-only pass; this review relies on source inspection, the author's listed validation, and green visible GitHub checks.

@Nillth thanks for the quick update. The previous freshness blocker is resolved, but I am requesting changes on one remaining cache-boundary issue before this default-on path can ship.

### ✅ Resolved — the cache key now covers changed audited bytes

`dir_signature()` now hashes file contents and symlink targets instead of mtime and length. That closes the same-length/same-mtime gap from my earlier review: an edited `SKILL.md`, `SKILL.toml`, script, or symlink target now changes the signature and forces the loader and audit to run again. The new `same_length_same_mtime_edit_still_busts_cache` regression covers the exact case that was missing before.

### 🟢 What looks good — invalidation and rollback stay simple

The cache still sits behind the two directory-loader functions, keeps `allow_scripts` and the loader tag in the key, and keeps the `ZEROCLAW_SKILLS_CACHE_ENABLED=0` rollback path. The explicit `SkillsService` invalidation calls after write, scaffold, and remove also cover the normal operator edit path cleanly.

### 🔴 Blocking — the digest opens non-regular entries while building the cache key

The digest routine opens non-regular filesystem entries while building the cache key. Opening a FIFO for reading can block waiting for a writer, so one unexpected entry inside a skill directory can hang skill loading or prompt building before the normal audit/loader path gets a chance to skip or reject it.

That broadens the file-open surface compared with the uncached path. The existing loader only reads the manifest files it is actually going to parse after audit, while the cache key now tries to open every non-dir/non-symlink entry under the directory.

Please check `file_type.is_file()` before calling `hash_file_contents`. One safe fix: recurse into directories, hash symlink targets, hash bytes only when `file_type.is_file()` is true, and return `None` for any other file type so that directory simply bypasses the cache. A small Unix regression with a FIFO, or an equivalent special-file case, would make sure this default-on cache cannot hang on a non-regular entry.

Nillth:
@Audacity88 good catch — pushed `8b1f2c86f1`.

`dir_signature` now guards on `file_type.is_file()`: it hashes regular-file bytes, hashes symlink targets, recurses into directories, and returns `None` for anything else (FIFO/socket/device) so that directory bypasses the cache instead of being probed. The key-building path never opens a non-regular entry now, so a planted FIFO can't hang loading or prompt building.

Added a `#[cfg(unix)]` regression `non_regular_entry_bypasses_cache_without_hanging` — it drops a real `mkfifo` pipe inside a skill dir and asserts the cache bypasses (loader runs each call) and returns promptly; it would hang if the guard regressed.

`cargo test -p zeroclaw-runtime` → 2206 passed; 9 cache tests; clippy `-D warnings` clean.

singlerider (approved):
Re-review (singlerider) at head 8b1f2c86. Verified both of @Audacity88's blocking findings against the actual `cache.rs`, read the two regression tests, and traced the decline-to-cache fallback path. CI has 12 green and 4 still running from the FIFO-fix push, zero failures, branch is MERGEABLE. Both blockers are resolved. Approving on my own merits.

@Audacity88 both your CHANGES_REQUESTED reviews were filed at commits that predate the fix they asked for (1f27d7a and 8b1d068), so the aggregate still shows changes-requested. Both gaps are closed at the current head. Could you convert or dismiss when you have a moment.

### ✅ Resolved: the freshness key now proves the audited bytes, not metadata

`dir_signature` hashes the content of every regular file under the directory (chunked through `hash_file_contents`) and the target string of every symlink, never following links. Because the cache sits in front of the security audit, this is the correct contract: any edit the audit would care about (a rewritten SKILL.md/SKILL.toml, a flipped script, a retargeted symlink) changes the signature and forces a re-audit, even an edit that preserves length and mtime. `same_length_same_mtime_edit_still_busts_cache` pins exactly that case with `filetime`, which the prior stat-only key would have missed. This is the stronger guarantee the title and module doc claim, now actually delivered.

### ✅ Resolved: the digest no longer opens non-regular entries

`dir_signature` now branches on `entry.file_type()`: symlink hashes the target, dir recurses, regular file hashes bytes via the `is_file()` guard, and anything else (FIFO, socket, device) returns `None` so the whole directory bypasses the cache and the uncached loader handles it. That keeps the cache key from ever opening a FIFO and blocking skill loading or prompt building, and it keeps the open surface no wider than the uncached path. The `#[cfg(unix)]` `non_regular_entry_bypasses_cache_without_hanging` regression plants a FIFO and asserts both loads run the loader (no caching, no hang); it hangs if the guard regresses.

### 🟢 What looks good: declining to cache is the safe default in every uncertain branch

The design fails toward correctness at every fork. `dir_signature` returns `None` on an absent dir, an unreadable entry, a content read error, or any non-regular file, and `cached_load` treats `None` as "run the uncached loader, store nothing" rather than trusting a partial digest. A concurrent mutation during `load` is self-healing because the next call's content digest differs from what was stored. Lock poisoning is recovered, not panicked. The cache key keeps `allow_scripts` and the loader tag so a workspace load and an open-skills load never alias, the `BTreeMap` keying makes the hash order independent of `read_dir` order, `SkillsService` writes/scaffolds/removes call `invalidate()` for the operator path, and `ZEROCLAW_SKILLS_CACHE_ENABLED=0` is a clean runtime rollback. A memoization layer in front of a security boundary is exactly where this much paranoia belongs.

PR #7260, opened 2026-06-05
feat(providers): add 7 OpenAI-compatible providers under schema v3

theonlyhennygod:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds **morph, github_models, upstage, featherless, arcee, lambda_ai, inception** as OpenAI-compatible model-provider families. They are wired through the 0.8.0 **schema-v3** typed-slot architecture (`CompatFamilySpec` + the `for_each_model_provider_slot!` macro), not the pre-v3 flat factory match.
- This **consolidates and ports** the stale, pre-v3 work from #6459 (morph, github-models, upstage), #6460 (featherless), #6461 (arcee), #6462 (lambda-ai), #6463 (inception) onto the current schema. Those PRs targeted the deleted `ProviderInfo`/`resolve_provider_credential` surface and no longer apply.
- Per-provider env-var credential resolution (e.g. `UPSTAGE_API_KEY`) is **intentionally dropped** — it was removed from the provider layer in v3; credentials now come from config or `--credential`.
- **Scope boundary:** Only adds these 7 families. No change to existing providers, transport, auth styles, or the catalog probe mechanism. No new models.dev/openrouter catalog wiring (these expose `/models` once a key is set).
- **Blast radius:** `zeroclaw-config` schema + provider slot macro, `zeroclaw-providers` factory/catalog/url tables. Additive; compiler-enforced completeness via the slot macro means a missing site fails the build.
- **Linked issue(s):** Supersedes #6459, Supersedes #6460, Supersedes #6461, Supersedes #6462, Supersedes #6463
- **Labels:** `enhancement`, `risk: low`, `size: M`, `provider`, `config`, `docs`

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-config -p zeroclaw-providers --all-targets -- -D warnings
cargo test -p zeroclaw-config -p zeroclaw-providers
cargo build --workspace
```

- **Commands run and tail output:**

```
$ cargo fmt --all -- --check
(no output, exit 0)

$ cargo clippy -p zeroclaw-config -p zeroclaw-providers --all-targets -- -D warnings
    Checking zeroclaw-config v0.8.0-beta-2
    Checking zeroclaw-providers v0.8.0-beta-2
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 47.64s   (exit 0)

$ cargo test -p zeroclaw-config -p zeroclaw-providers
[...]

PR #6403, opened 2026-05-05
feat(config,providers): typed-family split for model + TTS providers

singlerider:
> **DO NOT MERGE TO MASTER YET.** This PR targets `integration/v0.8.0`, not `master`. The whole v0.8.0 batch (#6266 baseline plus this PR plus the remaining v0.8.0 supporting PRs) squashes once into `master` at release.

## What this is — and what it is NOT

This is **schema and dispatch groundwork**, not a finished end-to-end multi-agent runtime.

What lands here:

- The operator-facing schema owns provider-family shape (typed family configs, single canonical name per family, typed `endpoint` enums for regional / OAuth variants).
- Per-agent provider refs are typed newtypes (`ModelProviderRef` / `TtsProviderRef` / `TranscriptionProviderRef` / `ChannelRef`).
- Fail-loud validation on dangling `model_provider` / `tts_provider` / `transcription_provider` / `channels[i]` refs.
- `Config::resolved_runtime_agent_alias()` is the single source of truth for "which agent is this turn for?", consumed by per-agent TTS and transcription dispatch.
- Operator-facing provider docs (`docs/book/src/providers/{catalog,configuration,custom,overview,fallback-and-routing}.md`, `docs/book/src/getting-started/multi-model-setup.md`, plus targeted fixes elsewhere) are rewritten to match the typed-family layout.

What does **not** land here — direct channel dispatch paths still have transitional bridges:

- **Gateway shared `AppState`.** `crates/zeroclaw-gateway/src/lib.rs` builds one `SecurityPolicy` / `risk_profile` / `tools_registry` at startup driven by the migration-synthesized `default` agent (or first enabled). This seeds the legacy single-agent `/api/tools` listing and the `run_gateway_chat_with_tools` test mock. Per-request agent dispatch on those surfaces is follow-up work.
- **Webhook chat / SSE / pairing endpoints.** Those endpoints don't yet accept an explicit agent in the request payload; they fall through to the same `default` / first-enabled bridge. Fixing this requires a request-schema change on each endpoint and is out of scope here.
- **Channel-side `try_transcribe_audio`** (gateway WS handler / direct channel-side ingest) binds to the sole registered transcription provider when only one is configured globally, instead of routing through agent resolution. Multi-provider setups still require explicit `agent.<alias>.transcription_provider` routing through the orchestrator; the single-provider bridge is purely for the common case where there's nothing to disambiguate.
[...]

PR #6268, opened 2026-05-02
feat(providers): add Manifest open-source LLM router

SebConejo:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Added `"manifest"` match arm in `create_provider_with_url_and_options()` using `OpenAiCompatibleProvider`, honoring a configured `base_url` for self-hosted Manifest instances and falling back to `https://app.manifest.build/v1`
  - Added `MANIFEST_API_KEY` to `resolve_provider_credential()` env var lookup
  - Added `mnfst_` prefix detection in `check_api_key_prefix()`
  - Added `manifest` to prefix validation match arm
  - Added Manifest to `list_providers()` with a focused factory smoke test
  - Added Manifest to provider catalog docs (verified endpoints table + dedicated section in Meta)
  - Why: Manifest is an open-source LLM router that cuts inference costs through smart routing across 16+ providers. Users currently need `kind = "openai-compatible"` with manual `base_url`. A named provider gives them `kind = "manifest"` with the URL pre-filled and API key auto-detected, while still allowing self-hosted users to override `base_url`.
- **Scope boundary:** No new Rust files, no new crates, no schema changes. Uses existing `OpenAiCompatibleProvider` with `AuthStyle::Bearer`.
- **Blast radius:** None. New match arms only fire for `"manifest"` kind. Existing providers unchanged.
- **Linked issue(s):** N/A

## Validation Evidence (required)

- **Commands run and tail output:**

```
cargo fmt --all -- --check
# (no output — formatted)

cargo clippy --all-targets -- -D warnings
# (not run locally — CI will validate; changes are 4 new match arms following identical patterns to existing ones)
```

- **Beyond CI:** Verified the match arm pattern matches Synthetic, OpenCode, and other `OpenAiCompatibleProvider::new()` calls exactly. Verified `mnfst_` prefix is unique among existing providers.
- **If any command was intentionally skipped, why:** Full `cargo test` skipped (requires full Rust toolchain + all dev deps). The changes are mechanical match arm additions following established patterns.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? `No`
- New external network calls? `No` (base URL is static unless the user configures this provider; no existing provider behavior changes)
- Secrets / tokens / credentials handling changed? `No` (uses existing `resolve_provider_credential` + `AuthStyle::Bearer`)
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No`

## Compatibility (required)

[...]

PR #7791, opened 2026-06-16
feat(providers): re-add Manifest under schema v3 provider system

SebConejo:
## Summary

- **What changed and why:** PR #6268 added Manifest against the pre-v3 provider dispatch functions (`create_provider_with_url_and_options()`, `resolve_provider_credential()`, etc.), but PR #6403's typed-family refactor replaced that entire architecture with `for_each_model_provider_slot!`. The old match arms were silently eliminated. This re-adds Manifest under the current schema v3 macro-driven system so it appears in `zeroclaw quickstart` provider selection.
- **Scope boundary:** Only adds the Manifest provider slot. Does not touch CLI prompts, tests beyond what the existing drift guards cover, or documentation beyond what the macro system auto-generates.
- **Blast radius:** None beyond the provider registry. The 5 touch points follow the exact morph/github_models pattern used by PR #7260.
- **Linked issue(s):** Related #6268 (original Manifest PR, code no longer exists on master)
- **Labels:** `type: feat`, `risk: low`, `size: S`

```bash
cargo fmt --all -- --check
# (clean)

cargo clippy --workspace --exclude zeroclaw-desktop --all-targets -- -D warnings
# Finished `dev` profile — 0 warnings

cargo test --locked -p zeroclaw-config -p zeroclaw-providers
# test result: ok. 964 passed; 0 failed; 0 ignored
```

- **Commands run and tail output:** All three commands above ran clean. The `every_canonical_family_has_a_catalog_table_entry` and `listed_model_providers_match_canonical_slots` drift guards both pass, confirming the macro slot, catalog entry, and display list are in sync.
- **Beyond CI — what did I manually verify?** Confirmed Manifest appears in the `list_model_providers()` output under `OpenAiCompatible` category. Verified the `CompatFamilySpec` impl compiles with `DEFAULT_URL = "https://app.manifest.build/v1"` and `AUTH = AuthStyle::Bearer`.
- **If any command was intentionally skipped, why:** Full workspace `cargo test` was scoped to the two affected crates for speed. CI runs the full suite.

- New external network calls? No (URL is config-only, no build-time fetch)

## Compatibility (required)

- Backward compatible? Yes
- Config / env / CLI surface changed? Yes — adds `providers.models.manifest` config section and `manifest` as a selectable provider in quickstart.
- Upgrade steps: None. New provider slot, no migration needed.


singlerider (approved):
First review (singlerider). Read the PR body and the full five-file diff. CI is 17/17 green including the provider drift guards. This re-adds the Manifest provider that #6403's typed-family refactor silently dropped from #6268, and it does so correctly through the schema v3 registry rather than around it. Approving.

### 🟢 What looks good: every touch point goes through the canonical registry

All five edits are the registry pattern, not a shadow list: the `ManifestModelProviderConfig` struct plus a typed `ManifestEndpoint` enum (not a string match) and the `impl_default_family_endpoint!` entry in schema.rs, the `for_each_model_provider_slot!` slot in providers.rs, the catalog source entry in catalog.rs, the `CompatFamilySpec` impl (DISPLAY / DEFAULT_URL / AUTH = Bearer) in factory.rs, and the `list_model_providers()` display entry in lib.rs. This mirrors the morph / github_models / lepton precedent exactly, so Manifest now flows through `quickstart`, the catalog, and the factory the same way every other OpenAI-compatible family does.

### 🟢 What looks good: the drift guards prove the slot set is internally consistent

The value of a macro-driven registry is that the pieces cannot silently disagree, and the two guards confirm it here: `every_canonical_family_has_a_catalog_table_entry` and `listed_model_providers_match_canonical_slots` both pass, so the new macro slot, the catalog entry, and the display list are in sync rather than one of them being forgotten (which is precisely the failure mode that erased Manifest the first time). The `ManifestEndpoint` enum keeping the default URL behind `ModelEndpoint::uri()` rather than a literal elsewhere is the right call. No build-time network, no credential-handling change, additive config only, clean `git revert` rollback.

I also dropped the spurious `provider:kilo` label, which the auto-labeler attached even though no Kilo code is touched.

Audacity88 (approved):
Second review after the upstream merge commit. I read the current head (23e830d), the five-file provider/config diff, singlerider's prior approval, and the current 17/17 green CI. I did not rerun Cargo locally in this pass; the approval is based on the diff, the existing focused validation evidence from the PR body, and CI.

### 🟢 What looks good — Manifest uses the canonical schema-v3 registry

This puts Manifest back through the same registry path as the nearby OpenAI-compatible families: `for_each_model_provider_slot!`, a typed `ManifestModelProviderConfig` / default endpoint, `catalog_source_for`, the `CompatFamilySpec` implementation, and the `list_model_providers()` display surface. That keeps config loading, provider construction, catalog behavior, and the selectable provider list aligned instead of adding another parallel provider map.

### 🟢 What looks good — The regression guard coverage matches the failure mode

The important risk here is drift between the canonical slot set and the places that expose or construct a provider. The existing guards for catalog coverage and `list_model_providers()` coverage are exactly the right shape for this change, and the PR’s focused validation plus green CI give me enough confidence that Manifest is no longer silently missing from the schema-v3 provider system.

Approving the implementation.

PR #7792, opened 2026-06-16
fix(runtime): resolve runtime profiles for direct turns

### LINKED ISSUE #7796: [Bug]: Direct agent turns ignore runtime-profile max_tool_iterations
## Affected component

runtime/daemon

## Severity

S1 - workflow blocked

## Current behavior

Direct agent turns can still stop at the default tool-iteration limit of `10` even when the selected agent uses a runtime profile with a higher `max_tool_iterations` value.

The visible failure is:

```text
Agent exceeded maximum tool iterations (10)
```

On current `master`, the direct-turn runtime path computes `config.effective_max_tool_iterations(agent_alias)` but does not pass that value into the tool loop. It later passes `agent.resolved.max_tool_iterations`; for a raw cloned agent config, `resolved` is skipped during deserialization and falls back to the default `10`.

## Expected behavior

Direct agent turns should resolve the selected agent through `Config::resolved_agent_config(agent_alias)` before entering the tool loop, so runtime-profile-backed fields are honored consistently.

For example, if an agent references a runtime profile with `max_tool_iterations = 100`, direct Code/CLI-style turns should use `100` instead of the default `10`.

## Steps to reproduce

```bash
# 1. Configure an agent to use a runtime profile with a higher tool-call limit.
#
# [runtime_profiles.long_turn]
# max_tool_iterations = 100
#
# [agents.default]
# runtime_profile = "long_turn"

# 2. Start a direct agent turn using that agent.
# 3. Ask it to perform work that requires more than 10 tool calls.
# 4. Observe the turn stop at the default limit:
#    Agent exceeded maximum tool iterations (10)
```

## Impact

Users can set the intended runtime-profile value and still have direct agent turns stop early. This makes the config surface look accepted while the runtime ignores the resolved profile value for the direct-turn entrypoints.

The practical result is blocked long-running Code/CLI-style workflows, especially repository inspection or debugging turns that naturally need more than 10 tool calls.

## Logs / stack traces

```text
Agent exceeded maximum tool iterations (10)
```

## ZeroClaw version

Observed in a direct-turn workflow and source-confirmed on `master` at `883164eec31b` before #7792.

## Rust version

N/A; this is a runtime config-resolution path bug rather than a compiler/toolchain-specific failure.

## Operating system

macOS was used for the observed direct-turn failure. The affected runtime path is not expected to be OS-specific.

## Regression?

Unknown. This is a remaining runtime-entrypoint gap after #6877 / #7401 covered runtime-profile `max_tool_iterations` precedence at the config layer.

## Related

- #6877
- #7401
- #7792

## Pre-flight checks

- I observed the failure in a direct-turn workflow and confirmed the affected source path on latest `master`.

Audacity88:
## Summary

- **What changed and why:**
  - Resolve direct agent-turn configs through `Config::resolved_agent_config()` before entering the tool loop.
  - Keep direct `run()` and `process_message()` aligned with runtime-profile resolution instead of reading the skipped/default `AliasedAgentConfig::resolved` field.
  - Add a focused regression proving direct turns honor `runtime_profiles.*.max_tool_iterations` above the default.
- **Scope boundary:** This PR only fixes direct agent-turn runtime-profile resolution. It does not change channel orchestrator runtime-profile behavior, context-budget defaults, ZeroCode config editing UX, or session/help/keybinding UI.
- **Blast radius:** `zeroclaw-runtime` direct `run()` and `process_message()` agent loop setup. Channel runtime setup is intentionally out of scope.
- **Linked issue(s):** Fixes #7796.
- **Labels:** `bug`, `risk: high`, `size: XS`, `agent`, `runtime`

```text
cargo test -p zeroclaw-runtime direct_turn_agent_config_resolves_runtime_profile_tunables
Result: passed. 1 test passed; 0 failed; 2205 filtered out.

cargo fmt --all -- --check
Result: passed.

git diff --check
Result: passed.

Result: passed.

Result: passed. 8720 tests run; 8720 passed; 10 skipped.
```

- **Commands run and tail output:** See the command/result block above.
- **Beyond CI — what did you manually verify?** I inspected the direct-turn `run()` and `process_message()` call sites and confirmed both now use the same resolved-agent helper before validating thinking config and entering the tool loop.
- **If any command was intentionally skipped, why:** No required local validation command was intentionally skipped. I did not manually exercise a live provider turn.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No upgrade steps are required.

- **Fast rollback command/path:** `git revert 4e5716924a4cc83c72f724c7598043f3587ec447` or use GitHub's Revert button for PR #7792 after merge.
- **Observable failure symptoms:** Direct agent turns would again ignore runtime-profile values that are stored in `AliasedAgentConfig::resolved`, most visibly enforcing the default `max_tool_iterations = 10` despite a higher configured runtime-profile value.

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/loop_.rs
+++ b/crates/zeroclaw-runtime/src/agent/loop_.rs
@@ -817,6 +817,17 @@ fn agent_provider_composite(
         .map(|(ty, alias, _)| format!("{ty}.{alias}"))
 }
 
+/// Return the owned agent config direct-turn setup needs, with runtime-profile
+/// values baked into `resolved`.
+fn resolved_agent_for_turn(
+    config: &zeroclaw_config::schema::Config,
+    agent_alias: &str,
+) -> Result<zeroclaw_config::schema::AliasedAgentConfig> {
+    config
+        .resolved_agent_config(agent_alias)
+        .with_context(|| format!("agents.{agent_alias} is not configured"))
+}
+
 /// Resolve (api_key, uri) for `provider_name`, preferring the alias-specific
 /// config when `provider_name` is a dotted `<family>.<alias>` reference.
 /// Falls back to `fallback` (the agent's configured provider) for bare family
@@ -857,10 +868,7 @@ pub async fn run(
     overrides: AgentRunOverrides,
 ) -> Result<String> {
     use ::zeroclaw_log::Instrument;
-    let agent = config
-        .agent(agent_alias)
-        .with_context(|| format!("agents.{agent_alias} is not configured"))?
-        .clone();
+    let agent = resolved_agent_for_turn(&config, agent_alias)?;
     crate::agent::thinking::validate_thinking_config(&agent.resolved.thinking);
     let risk_profile = config
         .risk_profile_for_agent(agent_alias)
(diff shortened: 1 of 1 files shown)


singlerider (approved):
First review (singlerider). Read the single-file diff and traced how the resolved tunables are actually consumed downstream on the head. CI is 19/19 green. This fixes runtime-profile precedence at the root for direct turns. Approving.

### 🟢 What looks good: the fix is at the source, so every `agent.resolved.*` read gets profile precedence

The bug was that `run` and `process_message` built `agent` from `config.agent(alias)`, which returns the raw `AliasedAgentConfig` with `resolved` holding inline/default values and the `runtime_profiles.*` overrides never baked in. Meanwhile the real `run_tool_call_loop` call sites read `agent.resolved.max_tool_iterations` / `agent.resolved.max_tool_result_chars` directly, so a direct turn silently ran with the un-profiled defaults. Switching both entry points to `config.resolved_agent_config(alias)` (via the small `resolved_agent_for_turn` helper) bakes the profile precedence into `agent.resolved` once, at the source, so all the existing `agent.resolved.*` consumers downstream now get the profiled values without each call site re-deriving them. Fixing it at the construction point rather than threading per-field `effective_*` results through is the cleaner root-cause fix.

### 🟢 What looks good: this also retires the discarded `effective_*` locals correctly

The block of `_eff_*` underscore-prefixed bindings (computed and dropped) is removed, and the few that were actually used (`eff_compact_context`, `eff_max_system_prompt_chars`, `eff_max_history_messages`) now read off `agent.resolved`, which is the same precedence source, so there is no behavioral regression on the fields that were already honored. The regression test seeds `runtime_profiles.long.max_tool_iterations = 25` on an agent referencing that profile and asserts `resolved_agent_for_turn(...).resolved.max_tool_iterations == 25`, pinning exactly the precedence that was being skipped.

Heads-up on overlap, not blocking: this touches the same `process_message` tunable block as #7780, which threads `max_tool_result_chars` / `context_token_budget` through the `agent_turn` wrapper. Whichever lands second will need a trivial conflict resolution, and since this PR fixes the source resolution it may make part of #7780's per-field threading redundant. Worth the second author confirming the persistent path still ends up with the profiled values after the merge.

Issue #7794, opened 2026-06-16
[Feature]: Per-agent opt-in Dream Mode + action-surface parity (chat command, gateway Dreams view)


JordanTheJet:
### Summary

Make Dream Mode opt-in and scoped **per agent**, and bring its action/monitoring surfaces to parity with prior art (OpenClaw, Hermes) — a chat `/dream` command and a gateway/web "Dreams" status+report view. Follow-up to the base Dream Mode in #6693.

### Problem statement

The base Dream Mode (#6693) runs as a single **global** cycle over a mixed memory pool with an implicitly-chosen provider (first configured entry), and its trigger/monitoring surface is **CLI + daemon only**. Two gaps follow:

1. **Multi-agent correctness.** In a multi-profile personal-assistant install or a multi-tenant deployment, a single global cycle would let one agent's dream **prune/consolidate another agent's memory**, and there is no way to choose which provider receives which agent's memory text. This is the exact lesson OpenClaw converged on (per-agent dreaming control) after shipping a global version.

2. **Surface parity.** Both reference implementations expose dreaming beyond the CLI:
   - **OpenClaw** — CLI + cron + chat slash-command (`/dreaming on|off|status`) + a **Gateway "Dreams" web tab** (status, phase counts, next run, Dream Diary) + a Control UI for promote. ([docs](https://docs.openclaw.ai/concepts/dreaming))
   - **Hermes Agent** — explicitly "across CLI **and gateway** execution paths, including Telegram delivery," invoked via a `/dreaming` slash-command. ([issue #5533](https://github.com/NousResearch/hermes-agent/issues/5533))

   ZeroClaw's Dream Mode is CLI + daemon only — narrower than both. The config *toggle* is already cross-frontend (CLI/REST/web via the shared `Configurable` schema), but the *action* (trigger / view report / promote) is not.

### Proposed solution

**Part 1 — Per-agent opt-in ("B1.5"). First deliverable; implemented in the stacked PR for this issue.**
- New `[agents.<alias>.dream_mode]` config block (`AgentDreamModeConfig { enabled: Option<bool>, model: Option<String> }`) layered over the global `[dream_mode]` defaults via `Config::effective_dream_config(alias)`. No provider field — each agent dreams through its own `model_provider`.
- `Config::agents_with_dream_enabled()` gates participation. Default off (opt-in): nothing dreams unless an agent opts in.
[...]

PR #6693, opened 2026-05-16
feat(memory): add dream mode for periodic memory consolidation

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Add Dream Mode: periodic memory consolidation that runs **local-only by default** (no external network calls), with an opt-in LLM-assisted reflection phase.
  - 5-phase engine (gather → reflect → consolidate → prune → report). Reflect is skipped when `dream_mode.model` is unset — the cycle is then purely mechanical (time-decay + importance-threshold pruning, no LLM call).
  - Wire Dream Mode into the daemon supervisor as a cron-scheduled background component, and add the `zeroclaw dream` CLI command (`--dry-run`, `--verbose`, `dream report`, `dream promote`).
  - **Side-effect-free `--dry-run`:** writes nothing — no memory mutations, no `dream_pending.json`, no `dream_report.json`.
  - **Staged-promote safety model:** `audit_mode` defaults to `true` — proposed mutations are staged to `dream_pending.json` for review; `zeroclaw dream promote` applies them.
  - **(This revision) Rebased onto current `master`** and resolved the V3 multi-agent provider-API drift (`first_model_provider*` / `provider_runtime_options_from_config` were removed upstream); the dream worker and CLI now resolve the default model_provider via `providers.models.iter_entries()` + `provider_runtime_options_for_alias`, matching `agent::build_session_model_provider`.
  - **(This revision) Blocker 1 — durable staged state:** a scheduled audit cycle no longer overwrites an unreviewed `dream_pending.json`; it preserves the existing file and surfaces a `dream_report.json` notice instead. `dream promote` now rewrites the pending file with only the **failed** items on partial backend failure instead of clearing it unconditionally.
  - **(This revision) Blocker 2 — no dream feedback loop:** `gather` now excludes `namespace == "dream"` entries (promoted Core insights) so the engine never re-dreams its own outputs.
- **Scope boundary:** Does not modify existing memory consolidation (`consolidation.rs`), importance scoring, or time decay — Dream Mode reuses those primitives. No new provider backends; the LLM calling interface is unchanged.
- **Blast radius:** Memory subsystem (new Core memories under namespace `"dream"`, only after staging + promote), daemon startup (new supervised component, disabled by default), agent interactive startup (report injection), config schema (new `[dream_mode]` section), CLI (new `dream` subcommand).
- **Linked issue(s):** Closes #5849
[...]

PR #7797, opened 2026-06-16
feat(memory): per-agent opt-in dream mode

JordanTheJet:
## Summary

- **Base branch:** `JordanTheJet/dream-mode` — this is **stacked on #6693** (it depends on the base Dream Mode feature). Retarget to `master` once #6693 lands.
- **What changed and why:**
  - Make Dream Mode **opt-in and scoped per agent** ("B1.5"). New `[agents.<alias>.dream_mode]` block (`AgentDreamModeConfig { enabled: Option<bool>, model: Option<String> }`) layered over the global `[dream_mode]` defaults via `Config::effective_dream_config(alias)`. No provider field — each agent dreams through its **own** `model_provider`.
  - `Config::agents_with_dream_enabled()` gates participation; the daemon spawns the dream supervisor only when an agent opts in. Default = nothing dreams.
  - `run_dream_worker` iterates opted-in agents, each via `run_agent_dream_cycle` with a per-agent timeout + error isolation (one agent's failure/timeout never aborts the sweep).
  - Per-agent scoping: agent-scoped memory (`create_memory_for_agent`), the agent's own provider, and per-agent `dream_pending.json` / `dream_report.json` in the agent's workspace dir. Interactive startup loads each agent's own report.
  - CLI: `zeroclaw dream [--agent <alias>]` resolves a single target; disabled agents are rejected and never inferred.
  - **Why:** the base feature runs a single global cycle over a mixed memory pool — in multi-agent/multi-profile installs that lets one agent's dream prune/consolidate another's memory. Per-agent scoping fixes that and matches the prior-art direction (OpenClaw per-agent dreaming).
- **Scope boundary:** Does **not** add the chat `/dream` command or the gateway/web "Dreams" status view (those are Part 2 of #7794). Does not change the base Dream engine/pending/report modules. No B2 per-agent supervisors / independent schedules, no B3 scheduler.
- **Blast radius:** config schema (new nested block), `runtime/daemon` (dream worker), `runtime/agent` (startup report load), CLI (`src/dream`, `src/main.rs`).
- **Linked issue(s):** Related #7794 (delivers Part 1 — per-agent opt-in). Depends on #6693.
- **Labels:** (auto path/scope labels: `config`, `daemon`, `runtime`, `agent`, `core`, `enhancement`; risk likely `risk: high` — maintainer to confirm)

Run on head `949a8a45` (stacked on the rebased `JordanTheJet/dream-mode`). The battery includes the main-crate architecture/Fluent-coverage test.

```
$ cargo fmt --all -- --check
(clean — exit 0)

    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 12s
(no warnings — exit 0)

$ cargo test -p zeroclaw-runtime -p zeroclaw-config -p zeroclawlabs
test result: ok. 834 passed; 0 failed; ...        (zeroclaw-runtime)
test result: ok. 2103 passed; 0 failed; ...       (zeroclaw-config — incl. 6 new dream opt-in tests)
test result: ok. 3 passed; 0 failed; ...          (zeroclawlabs architecture — incl. cli_fluent_coverage + 5 new dream CLI tests)
... all suites: 0 failed
```

- **Beyond CI — what did you manually verify?** Ran a 3-lens adversarial review (correctness / multi-agent scoping-safety / config backward-compat) with refute-by-default verification. It confirmed per-agent memory+provider scoping, opt-in semantics, and backward-compat are correct, and surfaced 3 issues I fixed: (1) a `.expect()` panic in `dream promote` on concurrent pending-file deletion → now graceful; (2) the CLI could target a *disabled* agent → `resolve_target_agent` now gates on `agent.enabled` and never infers a disabled agent (matches the daemon); (3) documented that a per-agent `model` override can't force local-only when a global `model` is set.

- New permissions, capabilities, or file system access scope? **`Yes`** — `dream_pending.json` / `dream_report.json` now live in each agent's workspace dir (previously a single global path). No new directories or elevated permissions.
- New external network calls? **`No` new** — the LLM reflect phase stays opt-in (only when the effective `dream_mode.model` is set); default is local-only with zero tokens. Per-agent scoping means an agent's memory only ever goes to *that agent's own* configured provider.
- Secrets / tokens / credentials handling changed? **`No`** — reuses the agent's resolved provider credentials.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **`No`**
- If any `Yes`: the per-agent file scoping is contained to each agent's existing workspace dir; isolation is *stronger* than the prior global behavior (no cross-agent memory mixing).

## Compatibility (required)

- Backward compatible? **`Yes`** — existing configs without an `[agents.<alias>.dream_mode]` block deserialize to defaults (`enabled: None`, `model: None`) and inherit the global `[dream_mode]` (default `enabled = false`), so nothing dreams unless opted in. The global `[dream_mode]` block is unchanged.
- Config / env / CLI surface changed? **`Yes`** — new `[agents.<alias>.dream_mode]` config block; `zeroclaw dream` gains an optional `--agent <alias>`.
- If `No` or `Yes`: upgrade steps for existing users: none required. To enable per-agent: set `agents.<alias>.dream_mode.enabled = true`.

- **Fast rollback command/path:** `git revert <sha>` (single commit). Dream Mode is off by default, so reverting is safe and immediate.
- **Feature flags or config toggles:** per-agent `agents.<alias>.dream_mode.enabled` (default off) + global `dream_mode.enabled` (default off); leaving the effective `model` unset keeps cycles local-only (no network/tokens).
- **Observable failure symptoms:** grep daemon logs for `Dream cycle failed for agent '<alias>'`, `Dream cycle for agent '<alias>' timed out`, or a `dream_pending.json` growing/recurring in an agent's workspace without a `promote`. The dream component reports health via `mark_component_ok("dream")`.


WareWolf-MoonWall (approved):
Reviewed at 949a8a45. The design is right — per-agent opt-in dream mode via `AgentDreamModeConfig` with workspace-scoped state, on-demand overlay resolution, and daemon-level supervisor gating. Approving the code. Two things to sort out before this can land.

### 🟢 What looks good — SSOT compliance

`Config::effective_dream_config(alias)` returns a by-value `DreamModeConfig` computed on demand from `self.agents` and `self.dream_mode` — never stored in an intermediate struct. `agents_with_dream_enabled()` is a filter over canonical `self.agents` computed on call. Per-agent dream file paths resolve through `Config::agent_workspace_dir()` at use time. No duplication of any config value anywhere in the diff.

### 🟢 What looks good — Isolation between agents

Moving `dream_pending.json` and `dream_report.json` into each agent's `workspace_dir()` instead of the global `data_dir` is the right call — agents can't clobber each other's dream state, and the workspace-per-agent model is consistent with how the rest of the runtime isolates agent data.

fmt/clippy/test output present with head SHA pinned, 6 config tests + 5 CLI tests listed individually, and the B1.5 known limitation (per-agent model override when global `dream_mode.model` is set) is honestly documented rather than hidden. That's the right standard.

### 🟡 Warning — `risk: high` label is absent

The PR body acknowledges the tier ("risk likely high") but the label was never applied. This is a `crates/zeroclaw-runtime/src/**` + config-schema + daemon path change — `risk: high` is correct by AGENTS.md table. A maintainer should apply it before merge.

### 🟡 Warning — Retargeting required before merge

This PR targets `JordanTheJet/dream-mode` (stacked on #6693), not `master`. The code review I'm approving is for the delta above #6693. Once #6693 lands on master, this branch needs to be retargeted to `master` and the CI suite re-run against the final merged base. Do not merge until the retarget is done and CI is green on master-based head.

singlerider (dismissed):
I read this end to end against the local tree, ran the config + CLI dream suites (`cargo test -p zeroclaw-config dream` → 6 passed; `cargo test -p zeroclawlabs --bin zeroclaw dream` → 5 passed), and cross-checked the helpers this stacks on (`resolved_model_provider_for_agent`, `provider_runtime_options_for_agent`, `create_memory_for_agent`, `agent_workspace_dir`). The per-agent scoping work itself is sound. This review is deliberately critical, including at the design layer, because the feature concept deserves the scrutiny before it sets a precedent.

### 🔴 Blocking — the dream component now reports healthy when every agent fails

`run_dream_worker` calls `crate::health::mark_component_ok("dream")` unconditionally after each sweep (daemon/mod.rs, end of the per-agent loop), and the per-agent arms only `record!` on `Ok(Err(_))` / `Err(_)`. The pre-refactor code called `mark_component_error("dream", ...)` on a failed cycle. As written, a dream config where every opted-in agent fails or times out every sweep reports the `dream` component as OK forever. That is an observability regression from the behavior you replaced. At minimum, mark the component errored when the sweep produced zero successful agent cycles (track a success count in the loop and branch on it), so `health` and the gateway status surface reflect a wholly broken dream config. The per-agent log lines are not a substitute for component health.

### 🔴 Blocking — "Dream mode" is the wrong name for what this does

The user-facing and config surface is built around "dream." What the engine actually performs is mechanical prune + consolidate over a memory pool, optionally with an LLM reflect pass. Naming it "dream" obscures the operation and, more importantly, hides the risk: consolidation that prunes and rewrites entries is lossy by construction. An operator reading `[agents.x.dream_mode]` has no signal that enabling it can degrade or drop stored memories. The name should describe the operation and its risk (memory consolidation / compaction), not a metaphor. This is a config and CLI surface (`zeroclaw dream`, `[dream_mode]`, `dream_pending.json`) that becomes a compatibility burden the moment it lands, so the naming decision is load-bearing and belongs in this stack, not a future rename. The engine lives in the base PR (#6693); raising it here because this PR is what extends that surface per-agent and is the natural place to settle it before the config keys proliferate.

The deeper concern, and I want it on record even though the engine is #6693's code: a periodic background job that prunes and rewrites an agent's long-term memory is a quality-degradation vector, not just a consolidation feature. The default-off posture is the right call and this PR's per-agent opt-in tightens it further, but the framing throughout (PR body included) treats consolidation as strictly additive ("insights"), and it is not. Before this surface is locked in, the stack needs an explicit statement of what consolidation can lose, how an operator audits it (the audit_mode / pending-promote path is the right primitive but it is undersold), and why "dream" is the right word for an irreversible memory mutation. If the answer is that the operation is genuinely lossy and the metaphor is doing marketing work, that is a reason to rename, not to ship the name.

### 🟢 The per-agent isolation is correct and well-tested

The core fix is real and done properly. `agents_with_dream_enabled()` gates on `agent.enabled` first so a disabled agent never dreams even if it opts in (`dream_disabled_agent_never_dreams` proves it), and `effective_dream_config` overlays per-agent over global with clean inherit-on-`None` semantics. Memory, provider, and pending/report state are all scoped to the agent (`create_memory_for_agent`, `resolved_model_provider_for_agent`, `agent_workspace_dir`), which closes the actual cross-agent-contamination bug the base feature had. The six config tests and five CLI tests cover the opt-in/opt-out/inheritance matrix and the disabled-agent edges directly. This is the right architecture for the problem.

### 🟢 Per-agent error isolation in the sweep

Wrapping each agent's cycle in `tokio::time::timeout(DREAM_AGENT_TIMEOUT, ...)` with `Ok(Ok)/Ok(Err)/Err` arms so one hung or failing agent never aborts the others in the sweep is the correct concurrency posture for a multi-agent background job, and the log messages name the agent in every arm. Good defensive design.

### 🟢 The concurrent-delete panic fix

Replacing the `.expect("pending was just loaded above")` in `promote` with a graceful `let Some(result) = ... else { print no-pending; return }` is a genuine correctness fix. The TOCTOU window between the snapshot `DreamPending::load` and `promote_pending` is real (concurrent promote, manual delete), and turning a panic into a clean no-op is exactly right.

Audacity88 (dismissed):
@JordanTheJet Thanks for splitting the per-agent Dream Mode work into this stacked slice. I reviewed current head `949a8a4539aae2ffc4b623bad64574fbcf29a650` against the live PR state, #7794, the stacked base #6693, the prior reviews, the diff from `JordanTheJet/dream-mode`, and the affected config/daemon/agent/CLI paths. I did not rerun local Cargo; visible GitHub checks are green for this stacked head.

### 🟢 What looks good — Per-agent scoping uses the right source-of-truth boundaries

The main shape of this slice is right. `Config::effective_dream_config(alias)` computes the per-agent overlay on demand instead of storing a duplicate snapshot, `agents_with_dream_enabled()` filters from canonical agent config, and both the daemon and CLI resolve the agent's own memory/provider/workspace via `create_memory_for_agent`, `resolved_model_provider_for_agent`, and `agent_workspace_dir`. That addresses the multi-agent contamination/provider-selection problem that kept the base Dream Mode branch from being reviewable as a global worker.

### 🔴 Blocking — An all-failed dream sweep reports the component healthy

I agree with @singlerider's health-status blocker and verified it against the current stacked diff. In the base branch, the single Dream cycle marked the component OK only on success and called `mark_component_error("dream", ...)` when the cycle failed. This PR changes the worker to iterate opted-in agents, log each per-agent failure or timeout, and then unconditionally call `mark_component_ok("dream")` after the sweep.

That means a config where every opted-in agent fails or times out on every sweep still reports the `dream` component as healthy. For a daemon feature that is supposed to surface background memory maintenance status, logs are not enough; health must reflect that the component made no successful progress.

Please track sweep success/failure and mark the component errored when zero agent cycles complete successfully. A focused regression test should cover a multi-agent sweep where each agent returns an error or timeout and assert that the `dream` component does not end in OK state.

### 🟡 Warning — The stack still has an unresolved base-surface decision

This PR is still stacked on #6693, and #6693 remains under active `CHANGES_REQUESTED`. I am not adding a separate duplicate blocker here for the broader naming/utility concern, but this slice should not be treated as merge-ready until the base Dream Mode surface is settled, the branch is retargeted to `master` after #6693 lands or is otherwise resolved, and CI/validation reruns against that final base.

Audacity88 (approved):
@JordanTheJet Thanks for the follow-up. I re-reviewed current head c7e46aed10621d6f8630537e44ac1a0e0fd117d0 against the live PR state, #7794, the stacked base #6693, the prior reviews, and the current config/daemon/agent/CLI diff. I did not run local Cargo for this review-only pass; I used the live PR state, visible GitHub checks, public source/diff inspection, and the validation evidence in the PR body.

This approval clears my prior health-status blocker for this stacked delta, dismissing my earlier `CHANGES_REQUESTED` from 949a8a45. It is not a merge-ready signal: #7797 still depends on #6693, and @singlerider's broader Dream Mode naming/surface block remains active on the stack.

### ✅ Resolved — all-failed dream sweeps now report unhealthy

The latest commit addresses the blocker from my prior review. `run_dream_worker()` now tracks per-agent sweep outcomes and calls `record_dream_sweep_health("dream", succeeded, failed)` after each sweep. That helper marks the component OK only when at least one opted-in agent cycle succeeds, and marks it errored when every opted-in agent failed or timed out.

The regression test also covers the exact contract that mattered: an all-failed sweep records `error`, while a partially successful sweep records `ok`. That restores the daemon health signal that the per-agent worker had accidentally weakened.

### 🟢 What looks good — per-agent scoping keeps the right ownership boundaries

The core B1.5 shape still looks right for this slice. The config overlay is resolved on demand through `effective_dream_config(alias)`, participation is computed from canonical agent config through `agents_with_dream_enabled()`, and both daemon and CLI paths use the target agent's own memory backend, provider reference, and workspace directory. That keeps Dream Mode from mixing one agent's memory/provider/report state with another's.

### 🟡 Warning — the stack is still blocked above this delta

This PR still targets `JordanTheJet/dream-mode`, not `master`, and #6693 remains under active `CHANGES_REQUESTED`. @singlerider's naming/surface concern for the broader Dream Mode stack is also still active. I am approving this per-agent delta only; it still needs the base stack decision resolved, a retarget or rebase onto the final base, and fresh CI/validation on that final shape before it can be considered merge-ready.

singlerider:
Moved to v0.8.4 to follow its issue #7794, which was parked here for re-triage. Dream Mode has no committed home in a specific point release, so the implementation moves with the feature rather than landing against v0.8.3, which is full and near its ship date.

JordanTheJet:
Closing alongside #6693.

This PR is stacked on the base Dream Mode feature (#6693), which I'm closing for the same reason: the consolidate/prune loop fits markdown-file memory but not our default vector-search backend, which already selects the relevant memories efficiently at query time without deleting or merging them. The per-agent scoping and isolation here is a genuinely nice refinement — but it refines a feature we're not moving forward with, so there's no base for it to land on.

Same note as on #6693: memory optimization is still a valid direction; a future take should start from what the vector backend needs rather than a periodic global rewrite. Related: #7794.

PR #7798, opened 2026-06-16
fix(gateway): surface pairing-code recovery when already paired on an alternate port

### LINKED ISSUE #5266: fix(gateway): no pairing code shown when running gateway start on alternate port
## Summary

When running `zeroclaw gateway start -p <port>` on a non-default port (e.g. for testing), the startup banner shows:

```
🔒 Pairing: ACTIVE (bearer token required)
   To pair a new device: zeroclaw gateway get-paircode --new
```

...but no pairing code is displayed. The `get-paircode --new` command also fails because it tries to reach the gateway on the default config port, not the one specified via `-p`.

## Root cause

The gateway shares `~/.zeroclaw/` config state. If a device has already paired on the default port, `pairing_code()` returns `None` on alternate-port instances (since the pairing state shows "already paired"). But the browser on the new port has no bearer token, so the user is stuck at the pairing screen with no code to enter.

The only workaround is:
```bash
curl -s -X POST http://127.0.0.1:<port>/admin/paircode/new
```

## Expected behavior

Either:
1. `gateway start -p <port>` should print a fresh pairing code when the port differs from config, OR
2. `get-paircode --new` should accept a `--port` flag to target the correct instance, OR
3. The startup banner should print the `curl` command as a fallback when no code is available

## Reproduction

```bash
# Start gateway on non-default port
zeroclaw gateway start -p 3001

# Open http://127.0.0.1:3001/ → pairing screen, no code
# Run get-paircode → fails (targets default port)
zeroclaw gateway get-paircode --new  # error: 400 Bad Request
```

## Environment

- OS: macOS
- ZeroClaw: built from master (7c54334)

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Once any device has paired, the shared config holds a token, so every later gateway instance starts with `pairing_code() == None` while `require_pairing` stays true — a fresh code is deliberately **not** minted on restart (that would reopen a standing, brute-forceable pairing window). A new browser/device (classic case: an alternate port → a different origin with no stored token) then hit the dashboard's 6-digit prompt with **no code printed anywhere and no in-band way out** (#5266).
  - **Startup banner:** the already-paired arm now states *plainly* that no new one-time code was generated, then points at the on-demand recovery command + curl. Wording is extracted into `already_paired_pairing_notice`, a single unit-tested source of truth that reuses the existing loopback-safe builders (so it never drifts from #6561's no-`--host` rule).
  - **Web pairing dialog:** stops claiming a code is "in your terminal" when none exists. On a localhost-served dashboard it offers a **"Generate pairing code"** button backed by the existing localhost-only `POST /admin/paircode/new`; on a remote/Docker origin (or a 403) it shows the equivalent CLI command built from the browser's real `host:port`. New `generatePairCode()` API helper centralizes the endpoint + 403-fallback signal.
- **Scope boundary:** Does **not** change pairing/token semantics — no auto-mint on startup, `PairingGuard::new` is untouched, `/pair/code` still returns null post-pairing. Does not touch the addr-in-use recovery message or the CLI `get-paircode` client helpers (intentionally different: they target a named `--host`).
- **Blast radius:** Gateway startup banner text; the unauthenticated pre-pair dashboard screen; one new same-origin `fetch` to the already-existing localhost-only admin mint endpoint. No config/CLI/wire-format changes.
- **Linked issue(s):** Closes #5266
- **Labels:** `bug`, `gateway`, `security: pairing` (maintainers set `risk:*` / `size:*`)

Rust changes are confined to one crate (`zeroclaw-gateway`); the web change is validated via the dashboard's own typecheck + build. Clippy/tests scoped to the affected crate, plus a full-tree release compile and the web build.

```text
$ cargo fmt --all -- --check
fmt exit: 0   # no diffs

$ cargo clippy -p zeroclaw-gateway --lib --tests -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 06s
clippy exit: 0

$ cargo test -p zeroclaw-gateway --lib
test result: ok. 259 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.90s
# new: already_paired_notice_states_no_code_was_generated,
#      already_paired_notice_includes_recovery_command_and_curl,
#      admin_paircode_new_rejects_remote_peer

$ cargo build --release --bin zeroclaw     # whole-tree compile check
Finished `release` profile [optimized] target(s) in 8m 04s

$ (cd web && tsc -b)        # exit 0
$ (cd web && vite build)    # ✓ built in 1.10s
```

- **Beyond CI — what I manually verified:** Reproduced #5266 with the rebuilt release binary in an isolated `ZEROCLAW_CONFIG_DIR`: started a primary instance (minted a code), paired a device (token persisted), then started a second instance on port 3001 sharing that config. New banner prints `🔒 Pairing: ACTIVE — this gateway is already paired, so no new one-time code was generated on this start.` + the recovery command + curl. Confirmed `GET /pair/code` returns `null` (the dialog's dead-end trigger), `POST /admin/paircode/new` on localhost mints a code, and that code pairs successfully end-to-end. The remote-403 dialog path is covered by the `admin_paircode_new_rejects_remote_peer` unit test (can't bind a non-loopback peer locally).
- **If any command was intentionally skipped, why:** Full-workspace `cargo test`/`clippy --all-targets` not run end-to-end locally (time); only one Rust crate changed, its full suite passes, and the whole tree compiles via the release build.

- New external network calls? **No** — the new `fetch` is same-origin to the gateway's own localhost-only `/admin/paircode/new`; no new external destination.
- Secrets / tokens / credentials handling changed? **Yes (no posture change).** Adds a UI affordance + helper that mint/display a pairing code, but: (a) no auto-mint — a code is only created on explicit operator action; (b) the mint endpoint stays localhost-only (regression-tested: 403 for non-loopback peers), so it's the same trust boundary as the `curl` the banner already advertises; (c) `/pair/code` still never exposes a code post-pairing. Net new attack surface vs. the existing curl recovery is nil.

## Compatibility (required)

- Config / env / CLI surface changed? **No** — banner text + dashboard UI only; no flags, config keys, or endpoints added/removed.
- Upgrade steps: none.

- **Fast rollback command/path:** `git revert <merge sha>` — self-contained, 3 files, no migrations/state.
- **Observable failure symptoms:** startup banner regresses to the ambiguous "Pairing: ACTIVE (bearer token required)" with no absence notice; or the pre-pair dashboard on an already-paired gateway shows no "Generate pairing code" affordance / a raw error instead of the CLI fallback.


Audacity88 (approved):
I checked the live PR state, linked issue #5266, the current diff/source, existing comments/reviews, and current CI. There are no existing review threads to resolve.

### 🟢 What looks good — already-paired recovery stays explicit and bounded

This fixes the #5266 dead end without changing the core pairing semantics: once a gateway is already paired, startup still does not auto-mint a fresh one-time code. Instead, the banner and dashboard now explain that no code was generated, and the dashboard only requests a replacement code from the existing localhost-only admin path after an operator action.

The security boundary looks preserved: `/pair/code` remains the unauthenticated initial-setup read path and only returns a code while the gateway is unpaired, while `/admin/paircode/new` stays behind the loopback peer check. Remote/non-loopback dashboard sessions therefore get the CLI fallback instead of a remote code-minting path.

Coverage also lines up with the risk: the Rust tests cover the already-paired banner wording, loopback/path-prefix-safe recovery hints, and the remote-peer 403 guard for `/admin/paircode/new`; CI is green across the visible required jobs, including test, lint, format, security, and platform checks.

Approved for the code change.

singlerider:

PR #7806, opened 2026-06-16
fix(whatsapp_storage): return HashState::default() when no app-state version row exists

danielO99:
## Problem
`RusqliteStore::get_version` surfaces a `StoreError::Database` when there is no row for a collection, instead of returning the zero/default `HashState`.

On a **fresh pairing** the `app_state_versions` table is empty, so the first critical app-state sync (`CriticalBlock` / `CriticalUnblockLow`) fails with a `database operation error`. Because the client gates `Event::Connected` on the critical sync completing, the device never finishes connecting on first pair.

## Root cause
`query_row` returns `QueryReturnedNoRows` for a missing row, which `to_store_err!` + `?` converts into a hard `Database` error.

## Fix
Match on the query result and treat `QueryReturnedNoRows` as `Ok(HashState::default())`, matching the reference implementations:
- `wacore/src/store/in_memory.rs` — `get_version` returns default when absent
- `storages/sqlite-storage` (diesel `SqliteStore`) — same

No behavior change when a row exists.

## Summary
- **Base branch:** `master`
- **What changed and why:**
  - `RusqliteStore::get_version` now maps a missing `app_state_versions` row to `HashState::default()` instead of an error, so a freshly-paired device can complete its first critical app-state sync and reach `Connected`.
  - Brings `RusqliteStore` in line with the two reference stores (`InMemoryStore`, diesel `SqliteStore`), which already return the default on a missing row.
- **Scope boundary:** Touches only `get_version` in `RusqliteStore`. No other store methods, serialization formats, schema, or the diesel store are changed. No behavior change when a row exists.
- **Blast radius:** Limited to `RusqliteStore` consumers (the `whatsapp-web` channel keystore). Read-path only; return type unchanged; existing on-disk data reads identically.
- **Linked issue(s):** None. (If there's a tracking issue for the WhatsApp Web channel, link it as `Related #<n>`.)
- **Labels:** suggested `type: fix`, `risk: low`, `size: S` (maintainers apply via sidebar).

Scoped to the only crate touched (`zeroclaw-channels`); `fmt` run workspace-wide. Full-workspace clippy/test left to CI since this is a single-method, single-crate change.

```bash
$ cargo fmt --all -- --check
# (no output; exit 0)

$ cargo clippy -p zeroclaw-channels --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 22s
# exit 0 (no warnings)

$ cargo test -p zeroclaw-channels
test result: ok. 755 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 6.78s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out


singlerider (approved):
Traced this against the rest of `RusqliteStore` and the fix is exactly right. Every other `query_row` in `whatsapp_storage.rs` already treats `QueryReturnedNoRows` as a normal empty state (`Ok(None)` at lines 440, 465, 513, 571, 627, 661, 743, 785, 921, and `Ok(0)` at 533). `get_version` was the one outlier that bubbled the missing-row case up as a hard error, which is precisely what breaks app-state sync on first pairing when the `app_state_versions` table has no row yet.

### 🟢 Correct and consistent with the established pattern

The `match` correctly preserves real database failures (`Err(e) => Err(StoreError::Database(...))`) while mapping the no-row case to `HashState::default()`. Returning version 0 for a fresh table is the right semantic, and the inline comment earns its place by explaining the why (first-pairing sync) rather than narrating the mechanism.

Approving.

singlerider:

Audacity88 (approved):
I reviewed the public PR snapshot for #7806 at head `7d10709` after the master merge, the single-file diff in `crates/zeroclaw-channels/src/whatsapp_storage.rs`, the prior review/comment thread, current CI checks, and the absence of linked closing issues. I did not run local Cargo or live WhatsApp validation for this review.

### 🟢 What looks good — missing app-state version now falls back to the default hash state

Returning `HashState::default()` only for `rusqlite::Error::QueryReturnedNoRows` is the right boundary: a fresh store has no row yet, stored rows still go through JSON deserialization, and real SQLite errors still remain `StoreError::Database`. That matches the nearby empty-state handling in this store and should unblock the first critical app-state sync path without masking corrupted state.

### 🔵 Suggestion — add a narrow regression test when convenient

This PR does not add a direct fresh-store/missing-row regression test. I do not think that should block this XS fix because the behavior is tightly scoped and CI is green, but a small test that calls `get_version` before any `set_version` would make this first-pairing case harder to regress.

No findings blocking merge.

PR #4403, opened 2026-03-23
feat(channels): parse media attachment markers in WhatsApp Web send

rareba:
## Summary

- Base branch target: `master`
- Problem: WhatsApp Web channel sends `[VOICE:]`, `[IMAGE:]`, `[DOCUMENT:]` media markers as literal text instead of native attachments
- Why it matters: Users see raw marker syntax in WhatsApp conversations instead of media files
- What changed: Added `parse_attachment_markers()` to WhatsApp Web's `send()` method — extracts markers, uploads local files via `client.upload()`, and sends them as native WhatsApp media messages (image, video, audio/voice, document). Failed uploads are silently stripped with a warning log.
- What did **not** change: Telegram's existing marker parsing, WhatsApp Cloud API channel, listen/receive path

## Label Snapshot (required)

- Risk label: `risk: medium`
- Size label: `size: M`
- Scope labels: `channel`
- Module labels: `channel: whatsapp-web`
- Contributor tier label: N/A
- If any auto-label is incorrect, note requested correction: N/A

## Change Metadata

- Change type: `feature`
- Primary scope: `channel`

## Linked Issue

- Closes #4385

## Supersede Attribution (required when `Supersedes #` is used)

N/A

## Validation Evidence (required)

```bash
cargo fmt --all -- --check   # PASS
cargo check --lib --features whatsapp-web   # PASS
cargo clippy --all-targets --features whatsapp-web -- -D warnings   # only pre-existing warnings (5), none from this PR
```

- Evidence provided: cargo check + clippy clean for new code, unit tests for marker parsing and MIME detection
- If any command is intentionally skipped, explain why: `cargo test` not run locally (whatsapp-web feature tests require wa-rs runtime); new unit tests are compile-gated behind `#[cfg(feature = "whatsapp-web")]`

## Security Impact (required)

- New permissions/capabilities? No
- New external network calls? No (uses existing wa-rs upload path)
- Secrets/tokens handling changed? No
- File system access scope changed? No (reads files from paths already provided in agent responses)
- If any `Yes`, describe risk and mitigation: N/A

## Privacy and Data Hygiene (required)

- Data-hygiene status: `pass`
- Redaction/anonymization notes: Test data uses only `/tmp` synthetic paths
- Neutral wording confirmation: Yes

## Compatibility / Migration

- Backward compatible? Yes
- Config/env changes? No
- Migration needed? No

## i18n Follow-Through (required when docs or user-facing wording changes)

- i18n follow-through triggered? No

[...]

PR #7811, opened 2026-06-17
feat(channels/whatsapp): send Web media markers natively

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Recover the WhatsApp Web media-marker behavior from #4403 in the current crate layout.
  - Parse outbound `[IMAGE:]`, `[PHOTO:]`, `[DOCUMENT:]`, `[FILE:]`, `[VIDEO:]`, `[AUDIO:]`, and `[VOICE:]` markers before WhatsApp Web sends text.
  - Resolve local marker paths inside the configured WhatsApp channel workspace before reading files, then send native WhatsApp image, video, audio/voice, or document messages through the current `whatsapp-rust` upload/send API.
  - Add WhatsApp Web delivery instructions so the agent knows the channel supports local media markers, and add localized count-only failure notes for dropped media markers.
- **Scope boundary:** This does not add HTTP/remote media fetching, new WhatsApp config keys, WhatsApp Cloud API behavior, reactions, allowed-groups behavior, cron delivery, or live WhatsApp smoke evidence.
- **Blast radius:** WhatsApp Web outbound sends under the `whatsapp-web` feature, channel delivery instructions, and localized failure-note strings. Other channels keep their existing media-marker behavior.
- **Linked issue(s):** Related #6074. Supersedes #4403. This does not close all of #6074 because other recovery rows remain.
- **Labels:** `enhancement`, `risk: high`, `size: L`, `channel`, `channel:whatsapp`.

```text
cargo fmt --all -- --check
result: passed

git diff --check
result: passed

cargo test -p zeroclaw-channels whatsapp_web --features whatsapp-web --lib
result: passed; 61 passed, 0 failed, 841 filtered out

cargo clippy -p zeroclaw-channels --features whatsapp-web --lib -- -D warnings
result: passed

result: passed locally on PR head c7f4ebd26ea4d3e2cdf88674b4c6d316f192775d in 5m46s

result: passed locally on PR head c7f4ebd26ea4d3e2cdf88674b4c6d316f192775d; 8726 tests run, 8726 passed, 10 skipped
```

- **Beyond CI — what did you manually verify?** Verified the #6074 recovery row against the current WhatsApp Web send path, checked #4403 attribution, compared the marker trust-boundary behavior against the current Matrix/Discord marker pattern, and confirmed the current `origin/master` drift touches unrelated tools files.
- **If any command was intentionally skipped, why:** No local ready-validation command is skipped. Live WhatsApp Web account smoke evidence remains out of scope for this PR because it requires a personal WhatsApp Web session.

- If any `Yes`, describe the risk and mitigation: WhatsApp Web can now send local files named by outbound media markers, and marker sends use the existing WhatsApp Web upload/send path. The patch refuses `http://`, `https://`, `data:`, `file:`, and other URL schemes, canonicalizes local paths, requires them to stay inside the configured WhatsApp channel workspace, and reports media failures with a count-only note instead of echoing paths to the chat.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: Existing WhatsApp Web configuration keeps working. Messages that intentionally use supported media-marker syntax now send native media instead of literal marker text.

- **Feature flags or config toggles:** The behavior is compiled under the existing `whatsapp-web` feature. There is no new runtime toggle.
- **Observable failure symptoms:** WhatsApp Web replies may fail to send media attachments, may append a count-only media-delivery note, or may log `whatsapp-web: media marker delivery failed`, `whatsapp-web: dropping unresolved outbound attachment marker`, or `whatsapp-web: marker target escapes workspace_dir`.

- Superseded PRs + authors: #4403 by @rareba
- Scope materially carried forward: Restores the user-facing WhatsApp Web behavior where outbound media markers are sent as native WhatsApp media instead of literal text, adapted to the current crate layout, workspace trust boundary, shared marker parser, and `whatsapp-rust` 0.6 API.
- `Co-authored-by` trailers added in commit messages for incorporated contributors? `Yes`, for @rareba.
- If `No`, why: N/A


singlerider (approved):
Read all 782 lines, with the path-resolution boundary as the focus since this reads local files and sends them out over a channel. It holds up.

### 🟢 Workspace boundary is enforced the right way

`validate_whatsapp_marker_target` canonicalizes both the workspace root and the resolved target before the `target_canon.starts_with(&workspace_canon)` prefix check. Canonicalizing first is what makes this safe: `..` segments and symlinks are resolved away before the comparison, so neither a relative `../../etc/...` nor a symlink planted inside the workspace can escape. Absolute paths are accepted but still must canonicalize inside the workspace, and a missing `workspace_dir` is a hard refuse rather than a fallback to the process CWD. URL schemes (`http`, `https`, `data`, `file`, and any `scheme://`) are rejected up front. The four `validate_*` tests cover accept / escape-refused / no-workspace-refused / missing-failed, and the Refused-vs-Failed split is meaningful: an escape is a trust-boundary refusal while a missing file is a delivery failure the operator should see.

### 🟢 Privacy-safe failure surfacing

`whatsapp_delivery_failure_note` reports a count only, and `delivery_failure_note_is_count_only` asserts the note contains neither a path separator nor the word "workspace". Dropping an unresolved marker and appending a translated count note (all five locales updated) is the right behavior, better than silently swallowing or leaking the offending path back to the chat.

### 🟢 Reuses the shared marker parser

Going through `super::util::parse_attachment_markers` and mapping the resulting `(kind, target)` tuples into `WhatsAppMediaKind` means this rides the same marker contract as the other channels rather than reimplementing parsing. `media_markers_reuse_shared_parser_kinds` pins that. The tool-call-tag strip before parsing is a sensible guard against marker-shaped fragments inside tool output.

One non-blocking note: `std::fs::canonicalize` is blocking and runs inside the async `send`. For a media-send path this is low-frequency and the surrounding channel code already does blocking fs in similar spots, so it is not worth changing here.

CI is 17/17. Approving.

PR #7813, opened 2026-06-17
feat(runtime): expose session key to shell tools as ZEROCLAW_SESSION_ID

RyanHoldren:
## Summary

- **What changed and why:** Expose the in-flight session key to shell tools as a `ZEROCLAW_SESSION_ID` environment variable.
  - The shell tool runs `env_clear()` and re-adds only a fixed allowlist, and the current session key lives solely as the in-process `TOOL_LOOP_SESSION_KEY` task-local — never an environment variable. A spawned tool therefore has no way to learn which session invoked it.
  - After the allowlist is applied, inject the scoped session key as `ZEROCLAW_SESSION_ID` when the turn is session-scoped (WS/ACP/orchestrator turns). One-shot channel and `/webhook` turns don't scope the task-local, so the variable is simply absent there rather than set to an empty string.
  - This unlocks the ability for shell tools to deliver results asynchronously, even across a process/host restart.
- **Scope boundary:** Adds **one** env var to the shell tool's child environment. Does **not** change turn behavior, the request surface, any endpoint, config, or the set of allowlisted host env vars.
- **Blast radius:** Every session-scoped shell invocation now carries one extra env var (`ZEROCLAW_SESSION_ID`). Non-scoped invocations are unchanged.
- **Linked issue(s):** None.
- **Labels:** `type: feature`, `risk: low`, `size: XS`.

```bash
cargo fmt --all -- --check
cargo test
```

- **Commands run and tail output:** Not run locally; this working environment is not powerful enough for local Rust compilation.
- **Beyond CI — what did you manually verify?** Deployed build and verified that a simple tool that just prints `ZEROCLAW_SESSION_ID` works as expected.
- **If any command was intentionally skipped, why:** All three — local compilation is not possible on my tablet.

- New permissions, capabilities, or file system access scope? **Yes (minor).** Shell tools now receive a `ZEROCLAW_SESSION_ID` env var carrying the opaque session key (e.g. `gw_<uuid>`) — never a credential or secret. It does not expand the host-env allowlist; it adds a single daemon-derived value.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No.** Tests use synthetic ids (`gw_abc-123`).

## Compatibility (required)

- Backward compatible? **Yes.** Purely additive — a new env var present only during session-scoped turns; nothing existing changes.
- Config / env / CLI surface changed? **Env only (additive):** tools may now read `ZEROCLAW_SESSION_ID`. No config, no request surface, no migration.
- Exact upgrade steps for existing users: None.

## Rollback

- `git revert 5a7ba84dac7c689d54cc32558aacec3da95e2eef` — single, self-contained commit.


singlerider (approved):
Read the single-file change. Clean and minimal.

### 🟢 Reuses the existing session-attribution task-local

`get_session_id` reads `TOOL_LOOP_SESSION_KEY` via `try_with`, which is the same task-local the rest of the runtime uses for attribution, rather than introducing a new global or threading a parameter through. Filtering the empty key means an unscoped turn produces no variable at all rather than an empty `ZEROCLAW_SESSION_ID`, and the three tests pin scoped / unscoped / empty behavior.

### 🟢 Correct placement relative to env_clear

Injecting after the allowlist is applied is exactly right given the shell tool's `env_clear()` posture: the variable would otherwise be wiped. The value is the in-process session key, not tool input, and it is set through `cmd.env(name, value)` as a structured pair, so there is no interpolation surface.

One non-blocking note: the injection happens before the TUI env overlay, and the overlay comment states TUI vars win on conflict, so an operator-provided `ZEROCLAW_SESSION_ID` in the TUI env would shadow the real session key. That env is operator-controlled rather than attacker-controlled so it is not a security concern, but if you want the runtime value to be authoritative you could set it after the overlay instead.

CI is 18/18. Approving. Flagging milestone separately.

singlerider:
@JordanTheJet — milestone alignment needed: this PR does not clearly fit within the scope boundary of any open milestone. Please advise on placement or deferral.

singlerider:

Audacity88 (approved):
I reviewed PR #7813 head `d284f91db395ff04b0405dc10ab0439537737626` against the live PR state, the single-file runtime diff, the existing `TOOL_LOOP_SESSION_KEY` / `sessions_current` behavior, the TUI env overlay path, singlerider's prior review, current CI, and merge freshness. I did not run local Cargo validation for this review-only pass.

### 🟢 What looks good — Shell tools reuse the existing scoped session key

The change reads `TOOL_LOOP_SESSION_KEY` directly through the same task-local session scope already used by `sessions_current` and scoped around agent turns. That keeps the source of truth in one place instead of adding a second session-id channel, and filtering the empty string keeps unscoped turns absent rather than exposing an ambiguous empty `ZEROCLAW_SESSION_ID`.

### 🟢 What looks good — The env insertion preserves the shell sanitization boundary

The shell tool still calls `env_clear()` and re-adds only the existing safe/passthrough environment first. `ZEROCLAW_SESSION_ID` is added as a daemon-derived structured `cmd.env()` pair after that clear step, so it survives the allowlist reset without expanding the host-env passthrough surface or adding command interpolation risk. Current CI is green, and freshness is clear on the refreshed head.

### 🔵 Suggestion — Add an execute-path regression when this area changes again

Non-blocking: the new helper tests cover scoped, unscoped, and empty session-key behavior, and the PR body records a deployed smoke where a shell tool printed `ZEROCLAW_SESSION_ID`. The TUI env overlay still runs after this insertion, so an operator-provided `ZEROCLAW_SESSION_ID` there could shadow the daemon-derived value. I do not treat that as a blocker because the overlay is operator-controlled, but a future hardening test could run `ShellTool::execute` under a scoped session and assert the child environment contains the key and the intended TUI precedence.

Issue #6825, opened 2026-05-21
[Tracker]: Zerocode UX


singlerider:
TUI UX — cross-cutting user experience concerns for the ZeroClaw terminal interface. Covers theming, keybindings, navigation, accessibility, and overall polish.

### Current state

- Theme system exists (`theme.rs`): icy-blue ZeroClaw palette with style helpers
- Widget library (`widgets.rs`, 83 lines): basic shared widgets
- Onboarding UI (`onboarding.rs`, 724 lines): first-run configuration wizard
- Config manager UI (`config_manager.rs`): full RPC-driven config editor with nested navigation
- Dashboard (`dashboard.rs`): 7-tab dashboard with overview, sessions, agents, memories, health, cost, cron
- Logs viewer (`logs.rs`): filterable log viewer with stacking filters, detail pane, attribute search
- Top-level mode router (`app.rs`): F1–F5 panes (Dashboard, Config, ACP, Chat, Logs)

### Scope

**Theming & visual polish**
- [x] Dark/light theme support (detect terminal background)
- [x] Dark-theme foreground contrast on light terminal profiles (#7377; fixed by merged PR #7482)
- [ ] Approval overlay can inherit terminal background instead of ZeroCode theme (#7807; PR #7823 is ready for review)
- [x] Configurable color scheme (user overrides in config)
- [x] Consistent use of theme palette across all config views (section, type, alias, field, edit screens)
- [x] Internationalization via Fluent — independent English (`en`) catalogue in `apps/zerocode/locales/`; all TUI strings routed through `fl!()`. English-only for this milestone; full multi-locale parity deferred
- [x] Locale selection refresh/readiness feedback after changing languages (#7400; fixed by merged PR #7478)

**Navigation & layout**
- [x] Session sidebar: list of active/recent sessions with switch-on-click
- [ ] Active Code sessions cannot switch agents directly (#7803)
- [x] Tab/panel system for composite views (Settings/Personality/Skills tabs with ←/→ switching)
- [x] Responsive layout: redraw on terminal resize
- [x] Nested drill-down navigation: section → type → alias → field with breadcrumb trail
- [ ] Config fields can look editable before Enter activates editing (#7814)
- [x] `/`-triggered fuzzy filter on all list screens (sections, types, aliases, fields, enum selects)
- [x] `$EDITOR`/`$VISUAL` integration: launch external editor for personality and skill files
[...]

PR #7823, opened 2026-06-17
fix(zerocode): fill approval overlay background

### LINKED ISSUE #7807: [Bug]: Approval overlay can inherit terminal background instead of ZeroCode theme
### Affected component

zerocode/tui

### Severity

S2 - degraded behavior

### Current behavior

The Code/Chat approval popup can render with a light or otherwise terminal-derived background while the surrounding ZeroCode UI is using a dark theme.

In the broken state, the warning-colored approval text and border are drawn over a light panel that clashes with the dark conversation background. In the expected state, the same approval popup uses a dark panel that matches the active ZeroCode theme.

Current source inspection matches the observed behavior:

- `render_approval_overlay()` calls `Clear` on the overlay rect and then renders a `Paragraph`.
- That paragraph's block uses `.style(theme::approval_border_style())`.
- `theme::approval_border_style()` only sets the warning foreground color; it does not set a background color.
- `theme.rs` already defines `fill_style()` and `modal_block()` so modal interiors use the active theme background after `Clear` instead of falling through to the terminal default.
- Nearby overlay code already uses the themed modal/fill pattern, for example the keybindings modal and file explorer overlay.

### Expected behavior

Approval popups should use the active ZeroCode theme for their interior background, while preserving the warning border/title accent.

After `Clear`, approval overlay cells should not inherit the terminal or system background when ZeroCode is rendering a themed dark UI.

The fix should also check nearby overlays that use `Clear` and only style the border foreground, so the same theme leak is not left in another high-visibility modal.

### Steps to reproduce

```bash
1. Run ZeroCode in a dark ZeroCode theme inside a terminal profile whose default background can differ from the ZeroCode theme.
2. Trigger a Code/Chat tool approval prompt.
3. Observe the approval overlay panel background.
4. In the broken state, the panel background can appear light/system-derived instead of matching the dark ZeroCode theme.
```

### Impact

Approval prompts are high-attention UI. A light or terminal-derived panel inside a dark themed Code session reads as a rendering glitch and can make warning-colored text harder to read.

The behavior also makes theming feel nondeterministic because the prompt can depend on terminal background state rather than the selected ZeroCode theme.

### Logs / stack traces

No stack trace. This is a TUI rendering/theme issue.

### ZeroClaw version

Current `master` at `883164eec31b45fa54c33fa9fb4d6123f7f71d3d`.

### Rust version

N/A; this is a ZeroCode TUI rendering issue.

### Operating system

macOS observed. The `Clear`/theme-fill rendering path should be platform-independent.

### Regression?

Unknown.

### Related

- #6825

### Pre-flight checks

- [x] I observed this in ZeroCode and source-confirmed the relevant approval overlay/theme paths on latest `master`.

Audacity88:
## Summary

- **What changed and why:**
  - Fill the approval overlay paragraph with the active ZeroCode theme background after `Clear`.
  - Keep the warning border/title accent separate from the panel fill so the alert styling stays visible.
  - Add a regression test that renders an approval prompt and asserts an interior cell uses the active theme background.
  - Add a test-only active-theme guard so theme-mutating tests restore the process-global palette safely.
- **Scope boundary:** This PR does not change approval behavior, keybindings, theme palettes, terminal color detection, or other modal flows.
- **Blast radius:** Limited to ZeroCode approval overlay rendering plus a test-only theme helper used by ZeroCode unit tests.
- **Linked issue(s):** Fixes #7807. Related #6825.
- **Labels:** `bug`, `risk: medium`, `size: S`, `zerocode`
- **Screenshots:**

before:
image

after:
image

  - `cargo test -p zerocode approval_overlay_uses_theme_background_after_clear`
    - `test chat::tests::approval_overlay_uses_theme_background_after_clear ... ok`
    - `test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 350 filtered out`
  - `cargo test -p zerocode set_active_swaps_palette`
    - `test theme::tests::set_active_swaps_palette ... ok`
    - `test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured`
  - `cargo fmt -p zerocode -- --check`
    - passed
  - `git diff --check`
    - passed
- **Beyond CI — what did you manually verify?** Launched a branch build of `zerocode`, triggered a shell approval prompt, and confirmed the overlay uses the dark ZeroCode theme background while preserving the warning border/title styling.
- **If any command was intentionally skipped, why:** Full workspace clippy/nextest were not run because this is a narrow ZeroCode TUI rendering fix with focused unit coverage and manual smoke verification.

## Compatibility (required)

- Backward compatible? `Yes`

- **Fast rollback command/path:** `git revert <merge_commit_sha>`
- **Observable failure symptoms:** Approval prompts may again show a terminal-derived or light panel background inside a dark ZeroCode theme, making the prompt look detached from the surrounding UI.

- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded PR scope was incorporated.


singlerider (approved):
I read the full diff against `apps/zerocode/src/chat.rs` and `theme.rs` at head 7f4e and traced the helper bodies on master. This is a correct, well-scoped fix.

### 🟢 What looks good - the fg/bg separation is the right root cause

The previous code set `Block::style(theme::approval_border_style())`, which only carries a `fg` (the warn accent) and never fills the panel background, so the overlay interior fell through to whatever the terminal had painted. Splitting that into `border_style(approval_border_style())` for the accent and `style(fill)` for the actual fill on both the `Paragraph` and the `Block` is exactly the fix, and applying `fill_style()` to the paragraph too means the wrapped body cells get the theme background, not just the block frame. The warning border/title accent stays intact because it now rides `border_style` independently.

### 🟢 What looks good - the test guards the real failure mode

`approval_overlay_uses_theme_background_after_clear` renders through `TestBackend` and asserts an interior cell at (10, 28) carries `Some(expected_bg)`, with an upfront `assert_ne!(expected_bg, Color::Reset)` so the assertion can't pass vacuously when `fill_style()` short-circuits on a Reset background. That's the right way to pin it given `fill_style()` deliberately returns the plain style when the active theme has no concrete background.

### 🟢 What looks good - the test-only theme guard fixes a real fragility

`set_active_for_test` serializing palette mutation behind `ACTIVE_TEST_LOCK` and restoring the previous palette on `Drop` is a clean fix for the process-global `ACTIVE` registry. Folding `set_active_swaps_palette` onto it and reading back through `active_raw()` also removes the brittle double registry lookup that asserted on `theme_by_name(...).title` rather than the value actually stored, so the refactor is a net correctness improvement, not just churn.

Validation evidence matches what I'd expect, scope boundary holds (no behavior/keybinding/palette changes), and the milestone (v0.8.3) is correct for a narrow zerocode stabilization fix.

Approving.

singlerider:
Seems like a MacOS-specific thing?

PR #6143, opened 2026-04-26
feat(skills): universal skill registry support (agentskills.io, skills.sh)

theonlyhennygod:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Introduce a `SkillRegistry` trait so external skill sources (ClawHub, agentskills.io, skills.sh, user-configured) live behind one interface, replacing the hardcoded if-else install chain with a registry dispatcher keyed off prefix shortcuts.
  - Add first-class registries for **agentskills.io** (`agentskills:<slug>`) and **skills.sh** (`skillssh:<slug>`), plus support for user-defined registries through `[skills.extra_registries]` in config so users can point ZeroClaw at private/custom skill catalogs without code changes.
  - Add `zeroclaw skills search <query>` which fans out across every configured registry so discovery is no longer ClawHub-only.
  - Wire up the SkillForge scouts (ClawhHub, agentskills.io, skills.sh) that were previously stubbed, and factor shared HTTP ZIP extraction into one helper with zip-bomb protection (50 MiB cap, 500 entry limit, 10x compression-ratio guard).
  - Accept bare `owner/slug` (e.g. `pskoett/self-improving-agent`) as ClawHub shorthand to match the prefix-less ergonomics of GitHub-style references.
- **Scope boundary:** Does not change skill execution / sandboxing, the on-disk skill manifest format, the existing ClawHub install path semantics for already-supported inputs, or any provider/channel/runtime code outside `skills` and `skillforge`.
- **Blast radius:** `crates/zeroclaw-runtime/src/skills/**`, `crates/zeroclaw-runtime/src/skillforge/**`, `src/skills/mod.rs` CLI surface, and `crates/zeroclaw-config` schema (`[skills.extra_registries]`). Anything that resolves a skill spec string or runs `zeroclaw skills install/search` is touched; sandboxing, agent loop, providers, channels, memory, cron are not.
- **Linked issue(s):** _None._

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy --all-targets --workspace -- -D warnings
cargo test
```

- **Commands run and tail output:**
  - `cargo fmt --all -- --check` — clean, no diff.
  - `cargo clippy --all-targets --workspace -- -D warnings` — clean, zero warnings (post-fix commit collapses nested `if let` into `&&` chains and swaps `filter().last()` for `rfind()` on a `DoubleEndedIterator`).
  - `cargo test` — all 98 skills + skillforge tests pass, zero regressions across the workspace.
[...]

Issue #7816, opened 2026-06-17
[Feature]: Pluggable skill registries — keep GitHub repo as default tier, add flagged external + user-configured registries


JordanTheJet:
### Summary

Keep the official GitHub repo as the default bare-name skill source, and generalize every *other* remote source (ClawHub, agentskills.io, skills.sh, and user-configured catalogs) into one pluggable second tier where each registry is reached by an explicit prefix, flag, or URL.

### Problem statement

Skill installs resolve against two hardcoded remote sources, dispatched by string-sniffing in `crates/zeroclaw-runtime/src/skills/mod.rs:138-165` (`clawhub: -> git -> bare-name registry -> local`):

- **Bare-name -> official GitHub repo** (`zeroclaw-labs/zeroclaw-skills`, auto-synced) — PR #6045. A single override exists via `skills.registry_url` (`crates/zeroclaw-config/src/schema.rs:5257`), but it's still *one* registry.
- **ClawHub** — `clawhub:<slug>`, `clawhub:<owner>/<slug>`, and `https://clawhub.ai/...` URLs (PRs #4069, #4088, async fix #6682).

There's no way to add another remote source (agentskills.io, skills.sh, a private/company catalog) without writing bespoke dispatcher code. ClawHub is also hardwired as its own special case rather than being one peer among interchangeable external registries. A prior generalization, **PR #6143 ("universal skill registry support")**, was **closed unmerged on 2026-06-13**, so the capability is still absent from `master`.

### Proposed solution

A two-tier registry model behind one interface, using closed PR #6143 as prior art (triage decision: revive #6143 wholesale or re-land as smaller scoped PRs):

- **Tier 1 — default source (no flag).** Bare-name `zeroclaw skills install <name>` continues to resolve against the official GitHub repo. Unchanged; it stays the curated default.
- **Tier 2 — external registries (explicit).** ClawHub, agentskills.io, and skills.sh become equal peers behind a `SkillRegistry` trait + prefix-keyed dispatcher. Each is addressed either by:
  - its **own prefix shorthand** — `clawhub:<slug>`, `agentskills:<slug>`, `skillssh:<slug>`; or
  - a **generic URL / `--registry` flag** for an ad-hoc registry URL.

  These are never implicit defaults — Tier 2 is always opt-in per install.
[...]

PR #7826, opened 2026-06-17
fix(runtime/agent): move credential redaction to the rendering layer

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** `scrub_credentials` ran on the tool-execution data path. The scrubbed value became `ToolExecutionOutcome.output`, which feeds the model, the HMAC receipt, and the loop detector. A tool reading a config that legitimately contains an `api_key` pattern handed the agent `[REDACTED]` instead of the real bytes, so a read-modify-write could persist `[REDACTED]` to disk, and the receipt signed mangled text. The scrub also protected nothing on that path: raw output was already emitted to the tracing records immediately above the scrub call. Redaction is a presentation concern, so it now lives at the rendering boundary and the data path carries true bytes end to end.
- **What it does:**
  - `execute_one_tool` builds `ToolExecutionOutcome` from raw output across all three branches (unknown-tool, success, error). The HMAC receipt signs raw output. Only the observer event `result` field is scrubbed, at emission.
  - `emit_tool_call_pair` scrubs the UI-facing `TurnEvent::ToolResult` output. That is the single rendering boundary covering the WS dashboard, RPC session updates, and ACP editor clients, so there is no per-sink drift.
  - `redact.rs` doc updated to state the function is rendering-only and never runs on the data path.
- **Scope boundary:** Behavior of `scrub_credentials` itself (the regex, prefix-preservation) is unchanged. No `zeroclaw-api` trait changes, no schema change, no new dependencies. Observer and tracing-record scrubbing at other call sites (`call_prep.rs`, `parse_response.rs`, `post_exec.rs`, `provider_call.rs`, `orchestrator/mod.rs`) are already emission-only and untouched.
- **Blast radius:** `zeroclaw-runtime` tool-execution path and the turn-event emission. Consumers that render tool output to a human (WS/RPC/ACP) still see scrubbed text; the model, HMAC receipts, and loop detector now see raw text.
- **Linked issue(s):** None filed; surfaced during diff review.
- **Labels:** `risk: high`, `size: S`, `agent`, `runtime`, applied via sidebar.

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway --all-targets -- -D warnings
cargo test -p zeroclaw-runtime --lib
cargo test -p zeroclaw-gateway --lib -- scrub_credentials_catches_credential_shaped_strings
```

- **Commands run and tail output:**
[...]

PR #7827, opened 2026-06-17
feat(skills): user-configured extra skill registries via registry:<name>/<skill>

JordanTheJet:
> Maintainer note: Audacity88 added follow-up validation for extra-registry aliases in `8a6d3ee` after review. The original feature scope and implementation remain JordanTheJet's contribution.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds a pluggable **second tier** of user-configured skill registries (`[[skills.extra_registries]]`) so operators can install from private/self-hosted catalogs **without code changes** — previously the default GitHub repo was the only bare-name source and ClawHub the only other (hardcoded) path.
  - Skills are addressed via a new **`registry:<name>/<skill>`** spec. Each named registry **reuses the existing git clone/pull/sync machinery** (cloned to its own `extra-registry-<name>/` dir) and then installs locally — no new fetch protocol is introduced.
  - Factors the ClawHub zip extraction into a shared, guarded **`extract_zip_secure`** helper (size cap + path-traversal guards), centralizing the guards and making them independently unit-testable.
  - First **scoped slice** of #7816 — deliberately small and reviewable rather than reviving closed PR #6143 wholesale (which carried an unresolved security finding and bundled five concerns).
- **Scope boundary:** Tier 1 bare-name installs (default `zeroclaw-labs/zeroclaw-skills`) and existing `clawhub:` / git / local specs are **byte-for-byte unchanged**. Only `kind = "git"` is wired; other kinds are reserved and rejected by validation. Does **not** add live agentskills.io/skills.sh fetchers, `zeroclaw skills search`, a `SkillRegistry` trait abstraction, or the OpenClaw->ZeroClaw skill normalization pass.
- **Blast radius:** `crates/zeroclaw-config` (new `ExternalRegistry` struct/field/validation + `HasPropKind` impl), `crates/zeroclaw-runtime/src/skills` (new dispatch + install path + extraction refactor), `src/skills` CLI handler, 5 locale files. The ClawHub installer now calls the extracted helper (behavior-preserving). No other subsystems touched.
- **Linked issue(s):** Related #7816
- **Labels:** `enhancement`, `skills`, `config`, `runtime`, `risk: medium`, `size: M`

  - `cargo fmt --all -- --check` — clean, no diff.
  - `cargo clippy --workspace --all-targets -- -D warnings` — `Finished dev profile ... in 6m 46s`, zero warnings.
  - `cargo test -p zeroclaw-config --lib` — `test result: ok. 868 passed; 0 failed; 0 ignored; 0 measured`.
  - `cargo test -p zeroclaw-runtime --lib` — `test result: ok. 2246 passed; 0 failed; 2 ignored; 0 measured`.
  - 8 new unit tests: extra-registry prefix accept/reject, parser split, mutual-exclusivity vs the clawhub/git/registry/local detectors, `is_unsafe_zip_entry_name` (traversal/absolute/backslash/colon/empty), `extract_zip_secure` happy-path + oversize rejection (asserts dest is not created), and unknown/disabled-registry error *before* any git work.
  - Dispatch ordering: confirmed `registry:<name>/<skill>` is rejected by all existing detectors and only caught by the new branch (inserted after the bare-name registry branch, before the local-path fallback); bare names / `clawhub:` / git URLs / local paths are not caught by the new detector.
  - Config validation rejects empty name/url, non-`git` kind, duplicate names, and non-http(s)/file URL schemes; accepts `git` + `file://`.
- **If any command was intentionally skipped, why:** Full-workspace `cargo test` was scoped to the two crates this PR changes (`zeroclaw-config`, `zeroclaw-runtime`); `fmt` and `clippy` were run full-workspace/all-targets (the latter compiles the bin and every crate). No live-network git clone is exercised in tests (offline by design) — the new path reuses the already-shipped, separately-tested clone/pull helpers.

- New permissions, capabilities, or file system access scope? **No** — clones into the existing workspace skills area (`extra-registry-<name>/`, sibling to the default `skills-registry/`); the registry name is charset-restricted (lowercase ascii letters, digits, `-`, `_`), so it cannot escape the workspace dir.
- New external network calls? **Yes** — `git clone` / `git pull` of user-configured registry URLs, **opt-in only** (no `[[skills.extra_registries]]` configured = no new calls). Same trust model as the existing default skills registry; `file://` is permitted for offline/self-hosted use.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (neutral placeholders like `acme`/`team`).
- **If any `Yes`, describe the risk and mitigation:** A malicious configured registry could serve hostile skill content. Mitigated by: the standard skill security audit applied to every install (`enforce_skill_security_audit` via `install_local_skill_source`); charset-restricted registry/skill names preventing directory traversal; the centralized `extract_zip_secure` path-traversal/size guards on the ClawHub path; and the registries being explicitly user-configured (trusted like the default registry). Only `kind = "git"` is accepted.

## Compatibility (required)

- Backward compatible? **Yes** — purely additive; existing specs and installs are unchanged.
- Config / env / CLI surface changed? **Yes (additive only):** new optional `[[skills.extra_registries]]` config table and a new accepted install spec prefix `registry:<name>/<skill>`. No new CLI flags or subcommands. Omitting the config preserves prior behavior.
- If `No` or `Yes` to either: exact upgrade steps for existing users: None required. To opt in, add `[[skills.extra_registries]]` entries (`name`, `url`, `kind = "git"`) and install with `zeroclaw skills install registry:<name>/<skill>`.

- **Fast rollback command/path:** `git revert 29ff137f62^..8a6d3ee087` or revert the two PR commits individually. No data migration; any configured `[[skills.extra_registries]]` becomes inert config an older binary ignores, and `extra-registry-*/` clone dirs can be deleted.
- **Feature flags or config toggles:** None — gating is implicit (don't configure extra registries / don't use the `registry:` prefix).
- **Observable failure symptoms:** Errors from `zeroclaw skills install registry:<name>/<skill>` — `registry '<name>' is not configured or is disabled`, `skill '<skill>' not found in registry '<name>'`, git clone/pull failures, or config-validation errors naming `skills.extra_registries[...]`; non-zero exit from the install command.


WareWolf-MoonWall (approved):
# Review — PR #7827 · feat(skills): user-configured extra skill registries via registry:<name>/<skill>

**Head SHA:** 29ff137f
**Verdict:** APPROVE (`--approve`)
**Active CHANGES\_REQUESTED from other reviewers:** None at time of review

---

## Position summary

This is a well-scoped, well-reasoned first slice of #7816. The implementation is clean, the security design is sound, and the SSOT discipline is respected throughout. I have two non-blocking warnings (missing milestone; `cargo test` scoped rather than full-workspace) and three optional suggestions. No blockers. Approving.

---

### 🟢 What looks good — SSOT is respected: no cached registry list

`install_extra_registry_skill_source` takes `extra_registries: &[zeroclaw_config::schema::ExternalRegistry]` as a parameter. The CLI handler calls it with `&config.skills.extra_registries`, where `config` is resolved at command invocation from the canonical config load. The extra-registry list is never stored in a channel handle, a runtime struct field, or a secondary cache — it is read from the config at use time, exactly as required. This is the correct pattern.

---

### 🟢 What looks good — path-traversal guards are correctly placed and independently tested

Both the registry name and skill name go through `is_registry_ident` (ASCII alphanumeric, `-`, `_` only — no `/`, `\`, `.`, `..`, `~`, `:`, or whitespace). This charset alone prevents any escapes from:
- `workspace_dir.join(format!("{EXTRA_REGISTRY_DIR_PREFIX}{registry_name}"))` — the clone target
- `registry_dir.join("skills").join(&skill_name)` — the skill source

The `is_unsafe_zip_entry_name` + `extract_zip_secure` helper adds the same guard for ClawHub-style archives, and both are unit-tested directly (`test_is_unsafe_zip_entry_name`, `test_extract_zip_secure_happy_path`, `test_extract_zip_secure_rejects_oversize`). Factoring the zip extraction out of the inline body and into a named, testable helper is a clear improvement over the pre-existing inline logic.

---

### 🟢 What looks good — all five locales covered

`cli-skills-install-resolving-extra-registry` (with `$source` and `$registry` parameters) is present in en, es, fr, ja, and zh-CN. All new user-facing strings go through `fl!()`/`get_required_cli_string_with_args`. This satisfies the Fluent requirement with no gaps.

---

### 🟢 What looks good — config validation is thorough and the test coverage matches

The schema-level validation covers empty name, empty URL, duplicate names, non-`git` kind, and unsupported URL schemes. The runtime guard in `install_extra_registry_skill_source` (`if registry.kind != "git"`) is redundant given config validation, but it is explicitly defensive and consistent with the codebase's principle of failing loudly near trust boundaries. Five new `zeroclaw-config` tests exercise each rejection path, and three runtime tests cover the spec parser, the mutual-exclusivity of detectors, and the unknown-registry early-error path.

---

### 🟢 What looks good — `extract_zip_secure` oversize rejection is correct

The size check fires before `std::fs::create_dir_all(dest)`, so an oversized archive leaves no partial directory on disk. The test `test_extract_zip_secure_rejects_oversize` asserts `!dest.exists()` — a useful invariant that verifies the cleanup contract, not just the error return.

---

### 🟡 Warning — Milestone not set

Neither this PR nor #7826 has a milestone assigned. The PR body acknowledges this ("maintainers finalize risk/size"). Please request milestone alignment from a maintainer before merge; `risk: medium` feature PRs should be anchored to a release target for backport/cherry-pick tracking.

---

### 🟡 Warning — `cargo test` scoped to two crates; template requires full-workspace run

The PR template requires `cargo test` (full workspace). The PR ran `cargo test -p zeroclaw-config --lib` and `cargo test -p zeroclaw-runtime --lib`. The explanation ("full-workspace test was scoped to the two crates this PR changes; fmt and clippy were run full-workspace/all-targets") is reasonable and the blast radius is correctly described as bounded. CI can make this moot, but the template gap is worth naming: when CI completes, the full-workspace result should be linked in a follow-up comment or the PR body updated with the CI run SHA to satisfy the Validation Evidence requirement fully.

---

### 🔵 Suggestion — use `url::Url::parse` rather than `reqwest::Url::parse` in schema validation

`zeroclaw-config/Cargo.toml` already lists `url = "2.5"` as an explicit dependency. `reqwest::Url` is just a re-export of `url::Url`, so both compile identically. Using `url::Url::parse` in `schema.rs` would make the dependency chain explicit and not require a reader to know about reqwest's re-export. This is cosmetic — no behavior difference.

```rust
// current
match reqwest::Url::parse(&reg.url) { ... }

// suggested
match url::Url::parse(&reg.url) { ... }
```

---

### 🔵 Suggestion — consider a test for the `enabled = false` path

`install_extra_registry_skill_source` correctly filters by `r.enabled`, but there is no unit test asserting that a configured-but-disabled registry is rejected before any git work. The existing `test_install_extra_registry_unknown_name_errors` covers the "not configured at all" case. A parallel test with a registry entry where `enabled: false` would complete the gate coverage.

---

### 🔵 Suggestion — double-parse in the CLI handler is a minor code smell

In `src/skills/mod.rs`, the dispatch path calls `is_extra_registry_source(&source)` (which internally calls `parse_extra_registry_source`) and then immediately calls `parse_extra_registry_source(&source)` again for the `registry_label`. The second parse is always `Some` at that point since it was just used as the guard. A common idiom would be to call `parse_extra_registry_source` once and match on the `Option` directly:

```rust
} else if let Some((registry_label, _)) = parse_extra_registry_source(&source) {
    println!(...);
    install_extra_registry_skill_source(
        &source, &skills_path,
        config.skills.allow_scripts, workspace_dir,
        &config.skills.extra_registries, no_tier_banner,
    )
    ...
```

This is a readability nit, not a correctness issue.

---

## Validation evidence assessment

| Command | Coverage | Status |
|---|---|---|
| `cargo fmt --all -- --check` | Full workspace | ✓ clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | Full workspace, all targets | ✓ 0 warnings |
| `cargo test -p zeroclaw-config --lib` | Config crate | ✓ 868 passed |
| `cargo test -p zeroclaw-runtime --lib` | Runtime crate | ✓ 2246 passed |

Scoped test evidence is sufficient for the described blast radius. Full-workspace `cargo test` should appear in a CI run or follow-up comment before merge.

Audacity88 (changes requested):
# Review — PR #7827

**Context:** Reviewed the live PR at head `29ff137f`, including the public PR body, diff, prior review state, visible GitHub checks, and public source excerpts for the config validation and install parser; no local validation was run for this review.

### 🟢 What looks good — scoped git-only registry path

Thanks for keeping this first slice narrow. The git-only extra-registry path keeps the new surface opt-in and bounded, and reusing the existing install/audit path is the right shape for this increment.

### 🔴 Blocking — config accepts registry aliases the CLI can never address

`ExternalRegistry.name` is documented as the short alias used in `registry:<name>/<skill>` install specs, but schema validation currently only rejects an empty name and exact duplicate strings. The runtime parser is stricter: `parse_extra_registry_source` only accepts ASCII alphanumeric, `-`, and `_` for the registry segment before `install_extra_registry_skill_source` does an exact lookup against `extra_registries`.

That means config like `name = "team.prod"` or `name = "team prod"` passes validation even though `zeroclaw skills install registry:team.prod/demo` can never enter the extra-registry branch. The user then gets a generic install-spec/local-path style failure instead of a config validation error naming the bad registry alias. It also makes the PR body's security statement that the registry name is charset-restricted true only for the CLI input path, not for the accepted config surface.

The config validation must enforce the same alias charset for `skills.extra_registries[].name` and add a regression test for rejected aliases such as dots, whitespace, slashes, or `..`. If aliases are trimmed or normalized for validation, duplicate detection should use the same canonical value so `team` and ` team ` cannot both validate.

This validation gap was not identified in the prior APPROVED review and should block merge until closed. Once it is fixed, this review pass does not see another blocker: the existing default registry, ClawHub, git URL, and local install paths remain separate; the new registry path is opt-in; and CI is green at `29ff137f`.

Audacity88 (approved):
Context: this is a blocker-clearing re-review of the current head `8a6d3ee` after the follow-up alias-validation patch. I checked the live PR state, prior reviews, the new config/runtime diff, the current label snapshot, the green GitHub checks, and focused validation for the touched config/runtime skill-registry surfaces.

### ✅ Resolved — extra-registry aliases are now validated at the config boundary

The blocker from my prior review is resolved. `skills.extra_registries[].name` now uses `ExternalRegistry::is_valid_name` before config validation accepts the registry, and the runtime `registry:<name>/<skill>` parser uses the same helper for the registry segment.

That closes the mismatch where config could accept aliases such as `team.prod`, whitespace, slashes, or `..` even though the CLI install spec could not address them. The patch also makes registry aliases lowercase-only to avoid clone-directory collisions on case-insensitive filesystems; the skill segment continues to use the existing registry-source rule.

The new regression coverage exercises dotted names, whitespace, slashes, `..`, leading/trailing spaces, and mixed-case aliases such as `Team` and `teamProd`.

Additional focused validation passed for the touched surfaces:

```text
cargo test -p zeroclaw-config validate_rejects_extra_registry
cargo test -p zeroclaw-runtime extra_registry
```

The current GitHub checks are also green at `8a6d3ee`, including CI Required Gate. I do not see a remaining blocker from my side.

PR #7832, opened 2026-06-17
refactor(channels/discord): split discord.rs into a module tree behind a structured outgoing payload

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `crates/zeroclaw-channels/src/discord.rs` had grown to ~6.8k lines, which made every upcoming Discord interaction-surface addition (rich embeds, typed slash options, message components) a large, conflict-prone edit on one file. This PR splits it into a `discord/` module tree with compiler-enforced boundaries, so each subsequent surface is a small, isolated, independently-reviewable change.
  - **Contract layer.** `discord/types.rs` declares the shapes that cross module boundaries (the outgoing payload, the interaction reply-target sentinel codec, the slash spec/resolver types, the listener fatal-error type, shared constants) and contains no logic. Every implementation module (`chunk`, `markers`, `rest`, `interaction`, `slash`) imports **only** `super::types` — there are zero implementation→implementation edges, so a change in one leaf cannot ripple sideways. `mod.rs` is the wiring plus the channel struct and gateway listen loop.
  - **Structured payload.** Introduces `DiscordOutgoing` as the single payload the outbound channel-message REST builders collapse onto (only `content` is populated today; `embeds`/`components`/`flags` are placeholder fields the follow-on PRs fill). This is the seam the embed/component/typed-option work needs — the builders take a structured payload instead of a bare string.
  - **Behaviour-neutral.** Pure refactor: same wire behaviour, same tests. The load-bearing invariant is that `DiscordOutgoing::to_rest_json()` serialises **byte-for-byte** to the historical `json!({ "content": content })` for a content-only message — pinned by the unchanged wiremock send tests plus new unit tests asserting byte-identity directly. Done as 9 strangler-fig commits, each compiling and passing tests on its own.
- **Scope boundary:** No wire-behaviour change, no new user-facing features, no config/CLI change. The `embeds`/`components`/`flags` placeholders are unused types. Does **not** touch the slash reconcile-durability layer (#7784, already merged) or any non-Discord channel.
- **Blast radius:** The `discord` channel module only. External consumers (the orchestrator's slash-resolver wiring) reach the channel via the same public paths — `discord::DiscordChannel`, `discord::discord_slash_specs_from_skills`, `discord::DiscordSlashCommandSpec` — which are preserved exactly; no other crate's API changes.
- **Linked issue(s):** Part of #7831.
- **Labels:** suggest `type: refactor`, `risk: low`, `size: L`, `channel:discord` (maintainers apply).

[...]

Issue #7831, opened 2026-06-17
[Tracker]: Discord channel interaction-surface parity — embeds, typed slash options, components, voice


Nillth:
### Tracker kind

roadmap / epic tracker

### Purpose

This tracker coordinates the multi-PR effort to bring the Discord channel to interaction-surface parity with the other mature agent Discord integrations. ZeroClaw's Discord channel is already feature-rich (mention/DM handling, attachments + transcription, slash commands over the gateway, inbound archive, reactions) but trails on four interaction surfaces: rich embeds, typed slash-command options, message components, and voice.

The work is sequenced behind a behaviour-neutral **refactor keystone**: `discord.rs` was a single ~6.8k-line file, which made each surface addition a large, conflict-prone edit. Splitting it into a module tree behind a structured outgoing payload turns every subsequent surface into a small, isolated, independently-reviewable PR. This tracker is the coordination surface for that sequence; each child PR references it.

### Scope and non-goals

**In scope** (in delivery order):

1. **Modular channel + structured payload** — the keystone, behaviour-neutral refactor that splits the channel into a module tree behind a `DiscordOutgoing` payload; the seam every surface below builds on.
2. **Outbound rich embeds** — an `[EMBED:{…}]` author surface, every URL trust-boundary-vetted, embed-aware budgeting, first-message placement.
3. **Slash commands: chunked followups + typed options** — followups split across the 2000-char limit instead of truncating; typed option model (string / integer / number / boolean / user / channel / role / mentionable) with choices and min/max/length, declared per-skill in `SKILL.toml`.
4. **Message components** — buttons / select menus / modals and their interaction dispatch. *Planned.*
5. **Voice** — receive/transcribe + reply. *Planned.*

**Out of scope:**

- Parity work for non-Discord channels (tracked separately).
- The slash reconcile-durability layer — already landed in #7784.

### Linked work

- **Foundation (merged):** #7489 (slash commands over the gateway), #7490 (dynamic skill slash commands), #7784 (slash reconcile durability)
- **This effort (PRs):** modular refactor — #7832 (merged); rich embeds — #7833; slash typed options — #7844; slash localizations + guild scope — #7922 (changes requested, foreign `/ask` cleanup); message components / modals / buttoned approval / autocomplete (EPIC B) — #7965 (approved by @singlerider at `7d3517b2`)
- Each child PR will carry `Part of #<this tracker>`.

### Routing evidence and next decision

[...]

PR #7833, opened 2026-06-17
feat(channels/discord): render outbound rich embeds from [EMBED:{…}] markers

Nillth:
## Summary

- **Base branch:** `master` — **stacked on the modular refactor PR (Depends on #7832)**; review the embed layer above it. Until #7832 merges, the diff here also shows its commits; the embed work is the top 5 commits.
- **What changed and why:**
  - Adds **outbound rich embeds**. An agent emits `[EMBED:{ …discord embed json… }]` in its reply; the channel lifts the marker out (brace-aware, because the JSON payload can itself contain `]`) and renders a Discord embed. The JSON is Discord's native embed-object shape: `title, description, url, color, timestamp, footer{text,icon_url}, image{url}, thumbnail{url}, author{name,url,icon_url}, fields[{name,value,inline}]` — `image`/`thumbnail` accept the nested `{"url": …}` object (what the agent emits) **or** a bare URL string.
  - **Renders on every outbound path.** Normal channel sends, **slash-command interaction replies** (the `@original` edit), and **streaming / Partial draft finalization** all route through the same `prepare_outgoing_embeds` extraction + `DiscordOutgoing` envelope — so the agent's `[EMBED:…]` affordance holds on every Discord reply, not just the normal send path.
  - **Trust boundary, reused.** Every URL an agent puts in an embed (image/thumbnail/url/author.url/author.icon_url/footer.icon_url) is fetched or linked by Discord, so each routes through the **same `validate_marker_target` egress boundary** as a media marker — only `http(s)` survives; a local path or other scheme is dropped (the rest of the embed still renders) and surfaces as a 🚫/⚠️ reaction on the normal send path.
  - **Serialises through the payload chokepoint** #7832 introduced: `DiscordOutgoing` now carries the embeds and `to_rest_json` emits them, while a content-only message stays **byte-for-byte identical** to before (no `embeds` key) — pinned by a regression test in both directions.
  - **Budgeted.** Embeds are clamped to Discord's documented limits: over-length text fields are trimmed with an ellipsis (graceful), and structural overflow (>10 embeds, >25 fields, or the message-wide 6000-character sum) drops the excess and fires a ⚠️ reaction.
  - **First-message placement.** Embeds ride the first message of a chunked reply only (same as attachments); the REST builders gained envelope-taking variants (`send_discord_message_payload[_with_files]`, `edit_discord_message_payload`) so content-only callers are untouched. The orchestrator system prompt tells the Discord agent it may emit `[EMBED:{…}]` (http(s) URLs only).
  - **Outbound only.** Inbound embed ingestion is intentionally out of scope here (a later change).
- **Scope boundary:** Outbound embeds only — no inbound ingestion, no components/flags (those stay placeholder types), no new config. Content-only and media-marker sends are unchanged and byte-identical.
- **Blast radius:** The Discord channel send/interaction/draft paths plus its `markers`/`chunk`/`rest`/`interaction`/`types` modules and the orchestrator's Discord affordance string. No other channel; no other crate's API.
- **Linked issue(s):** Part of #7831. Depends on #7832.
- **Labels:** enhancement, size: XL, risk: medium, channel, channel:discord.

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-channels --features channel-discord --all-targets -- -D warnings
cargo test -p zeroclaw-channels --features channel-discord
```

  - `cargo fmt --all -- --check` → clean (no diff)
  - `cargo clippy … --all-targets -- -D warnings` → `Finished` clean, zero warnings
  - `cargo test -p zeroclaw-channels --features channel-discord` → `test result: ok. 965 passed; 0 failed; 1 ignored` (lib), integration bins `3 / 3 / 4 / 0 passed`, `0 failed`
  - Unit/wire coverage: empty embeds vec omits the `"embeds"` key (byte-identity, both directions); the brace-aware parser against JSON containing `]` and a serde-rejected `[EMBED:{…}]` whose payload holds a nested `[EMBED:` (kept verbatim, not re-scanned); a local-path/disallowed-scheme embed URL dropped while the rest survives; a fields-heavy embed over 6000 chars shrunk by dropping fields, not just the description; `image`/`thumbnail` parse from both the nested `{url}` object and a bare string. Two **wiremock** tests assert the slash-reply `@original` edit body carries the embeds and that a content-only reply stays byte-stable, plus a unit test pins the finalize / first-message envelope transform.
  - **Live Discord smoke:** deployed to a running bot and rendered a single `[EMBED:{…}]` exercising the full contract end-to-end — branded author + icon, thumbnail, hero image, custom accent color, layered markdown, inline + block fields, clickable link-actions, and an icon'd footer with timestamp. (This is what surfaced — and confirmed the fix for — the nested image/thumbnail format.)
  - Adversarial review passes found and fixed real bugs along the way (parser re-scan on a serde reject; the char budget only shaving the description; the two bypassing render paths; the nested image/thumbnail format).

- New permissions, capabilities, or file system access scope? **No** — embed URLs route through the same `validate_marker_target` egress trust boundary as existing media markers (http(s) only; local paths refused).
- New external network calls? **No new calls from ZeroClaw** — embed image/icon URLs are http(s)-vetted and sent to Discord in the message JSON; Discord fetches them server-side, exactly as for the existing remote-media-URL markers.

## Compatibility (required)

- Backward compatible? **Yes** — purely additive; content-only and media-marker replies are unchanged, and embeds appear only when the agent emits an `[EMBED:{…}]` marker.
- Config / env / CLI surface changed? **No** — the only surface addition is a system-prompt affordance line, not a config field.

## Rollback

Low-risk: `git revert <sha>` is the plan. The feature is additive and behind the agent emitting an `[EMBED:{…}]` marker; reverting removes the embed surface with no data or config implications.


Audacity88 (changes requested):
I reviewed the current head d701876579 against live PR state, linked tracker #7831, merged dependency #7832, the Discord channel diff, the existing streaming/interaction send paths, and the green GitHub checks. I did not run local Cargo validation for this review-only pass.

### 🟢 What looks good — the normal channel send path has the right shape

The core `prepare_outgoing_embeds()` path is a clean place to centralize the marker parse, URL vetting, and Discord budget logic. The parser is brace-aware, the URL fields route through the existing marker target boundary, and the content-only REST payload still omits `embeds`, which keeps the prior wire shape stable for ordinary text sends.

### 🔴 Blocking — embeds are advertised on Discord paths that never render them

The new Discord prompt affordance tells the agent it may emit `[EMBED:{...}]`, but two existing Discord delivery paths still bypass the embed pipeline and send content-only messages.

For slash-command replies, `send()` returns through `discord_edit_interaction_response()` before it reaches `prepare_outgoing_embeds()`:

- `crates/zeroclaw-channels/src/discord/mod.rs` lines 1403-1410: strips tool tags and edits the interaction response with plain content.
- `crates/zeroclaw-channels/src/discord/mod.rs` lines 1418-1422: the new embed parse/vet path only runs after that early return.

For Partial streaming finalization, `finalize_draft()` still parses only attachment markers and then either posts with `send_discord_message_with_files()`, posts chunks with `send_discord_message_json()`, or edits the draft with `edit_discord_message()`. None of those paths call `prepare_outgoing_embeds()` or send a `DiscordOutgoing` with embeds:

- `crates/zeroclaw-channels/src/discord/mod.rs` lines 2794-2801
- `crates/zeroclaw-channels/src/discord/mod.rs` lines 2816-2823, 2846-2847, and 2863-2865

That means a Discord user can get the new system-prompt instruction, the agent can follow it, and the channel will show the raw `[EMBED:{...}]` marker instead of a rich embed. This is especially visible for slash commands, which are part of the same #7831 Discord interaction-surface effort, and for any Discord config using Partial draft updates.

Please route the interaction-response and draft-finalization paths through the same embed extraction/budgeting/envelope logic, or explicitly keep the prompt affordance out of paths that cannot render embeds yet. Either way, add regression coverage for the affected path so this does not stay normal-send-only.

### 🟡 Warning — approval will need Discord-boundary smoke evidence

The unit and REST-shape tests are useful, but this is a user-visible Discord channel behavior change. After the missing paths are wired, please include live or near-live Discord smoke evidence for the supported outbound embed path, including how URL rejection or structural truncation surfaces to the user.

Nillth:
Thanks @Audacity88 — good catch, addressed in two follow-up commits.

**`2ac1bfa10c` — the bypassing paths now render embeds.** You were right that the affordance was advertised on every Discord reply while only the normal `send()` path rendered embeds. Both other paths now route through the same `prepare_outgoing_embeds` extraction + `DiscordOutgoing` envelope:

- **Slash-command replies** — `discord_edit_interaction_response` takes the extracted embeds and attaches them to the `@original` edit via the envelope; `send()`'s interaction early-return lifts `[EMBED:…]` out (and budgets/vets it) before editing.
- **Streaming / Partial draft finalize** — `finalize_draft` lifts embeds out of the final text and attaches them to the first finalized message across all three branches (files / chunked / in-place edit). The in-place edit gained an envelope-capable `edit_discord_message_payload`.

Content-only sends stay byte-identical (`{"content":…}`, no `embeds` key) — pinned both directions.

**`5e15ba33a0` — nested `image`/`thumbnail` `{url}` form.** While doing the live smoke (below) I hit a real one: the parser modeled `image`/`thumbnail` as bare URL strings, but Discord (and an agent following the "Discord embed JSON object" affordance) emits them nested as `{"url": …}` — so those two fields were silently dropped. `footer`/`author` were already nested, which is why they worked. They now accept both the nested object and a bare string, with a regression test for both forms.

**Regression coverage** (your "don't let it stay normal-send-only" ask):
- Two wiremock tests on the slash-reply path: the `@original` edit body carries the embeds for an `[EMBED:…]` reply, and a content-only reply stays byte-stable `{"content":…}`.
- A unit test pinning the finalize / first-message envelope transformation, so neither path can regress to content-only.
- The nested-media regression test above.

**Live Discord smoke (🟡).** Deployed the branch to a running ZeroClaw Discord bot and exercised the full surface. The screenshot below is one `[EMBED:{…}]` rendering the complete contract end-to-end — branded author + icon, thumbnail (top-right), hero image (below), custom accent color, layered markdown, inline + block fields, clickable link-actions, and an icon'd footer with timestamp. (This is also what surfaced the nested-media bug, now fixed.)

image

**One scoped v1 limitation** (flagging it, not hiding it): embed-URL-rejection / structural-truncation reactions aren't surfaced on the interaction `@original` edit (no reaction target there) nor folded into the draft path's reaction set — the embed still renders without the rejected field, and a malformed `[EMBED:]` is left verbatim. On the normal channel send path, URL rejections still map to 🚫 and structural truncation to ⚠️ on the bot's message, as before.

Re-pushed; CI green on the new head.

Audacity88 (approved):
Checked the updated head `5e15ba33a0`, prior blocking review, linked tracker #7831, dependency #7832 context, current PR thread/reviews/comments, green GitHub checks, and the Discord interaction, draft-finalization, marker, REST, and embed source paths. I did not run local Cargo validation for this review-only pass; I am relying on CI, the author's Discord smoke evidence, and source/diff inspection here.

### ✅ Resolved — embeds now render on the bypassing Discord reply paths

The previous blocker was that `[EMBED:{...}]` was advertised to the Discord agent, but slash-command replies and Partial streaming finalization still bypassed the embed extraction/envelope path and would send raw markers. The updated head now lifts embeds before the slash `@original` edit and passes them through `discord_edit_interaction_response()`, and `finalize_draft()` now lifts embeds before attachment parsing and attaches them to the first finalized message across the file, chunked, and in-place edit branches.

### 🟢 What looks good — the follow-up covers the user-visible contract

The added coverage pins the two affected surfaces: slash-reply edits include `embeds`, content-only slash replies keep the old `{"content": ...}` shape, and the draft-finalization transform produces a first-message `DiscordOutgoing` carrying the embed. The nested `image`/`thumbnail` `{ "url": ... }` fix also matches the public Discord embed shape and was backed by happy-path live Discord smoke evidence.

Approving to clear my prior blocker. This is not a merge-ready call: GitHub still reports the branch as `DIRTY`, and the PR still carries `needs-author-action`, so it needs a conflict refresh before merge readiness.

PR #7836, opened 2026-06-17
fix(channels/orchestrator): use resolved agent config for strict_tool_parsing and parallel_tools

### LINKED ISSUE #7809: [Bug]: Channel turns ignore runtime-profile strict/parallel tool flags
### Affected component

channel

### Severity

S2 - degraded behavior

### Current behavior

Channel message turns can ignore runtime-profile values for `strict_tool_parsing` and `parallel_tools`.

Current source inspection shows the channel runtime setup resolves each enabled agent through `Config::resolved_agent_config(agent_alias)` and stores that resolved agent on `ChannelRuntimeContext.agent_cfg`.

That same context also carries resolved/effective values for other runtime-profile fields. For example, channel setup stores `max_tool_iterations` from `config.effective_max_tool_iterations(agent_alias)`.

However, the per-message channel tool-loop call does not use the already-resolved `ctx.agent_cfg.resolved` values for `strict_tool_parsing` and `parallel_tools`. Instead, it reaches back into:

```rust
ctx.prompt_config
    .agent(ctx.agent_alias.as_str())
    .is_some_and(|agent| agent.resolved.strict_tool_parsing)
```

and similarly for `agent.resolved.parallel_tools`.

Because `AliasedAgentConfig.resolved` is skipped during deserialization and runtime-profile values are baked only by `resolved_agent_config()`, this raw `prompt_config.agent(...).resolved` read falls back to the default `false` values.

### Expected behavior

Channel message turns should honor the selected agent's runtime profile for all runtime-profile-backed tool-loop flags.

If an agent uses a runtime profile with:

```toml
[runtime_profiles.strict_channel]
strict_tool_parsing = true
parallel_tools = true

[agents.default]
runtime_profile = "strict_channel"
```

then channel messages for `agents.default` should pass `true` for both `strict_tool_parsing` and `parallel_tools` into the tool loop.

The channel path should use the resolved values already available on `ctx.agent_cfg.resolved`, or otherwise use the same `effective_*`/resolved helper boundary that channel setup uses for the other runtime-profile knobs.

### Steps to reproduce

```bash
1. Configure an agent with a runtime profile that sets `strict_tool_parsing = true` and/or `parallel_tools = true`.
2. Route a message through a channel using that agent.
3. Inspect the channel tool-loop call path in `crates/zeroclaw-channels/src/orchestrator/mod.rs`.
4. Observe that the call reads `ctx.prompt_config.agent(...).resolved` for these flags instead of `ctx.agent_cfg.resolved`.
```

### Impact

Operators can configure runtime-profile tool parsing and parallel-tool behavior and still get the default channel behavior.

This creates a split-brain runtime-profile surface:

- direct turn/runtime paths can be fixed to use resolved agent config;
- channel setup already resolves several runtime-profile fields;
- channel message tool-loop flags still read the raw skipped/default `resolved` field for `strict_tool_parsing` and `parallel_tools`.

The result is especially confusing because nearby channel runtime-profile fields, such as `max_tool_iterations`, are already wired through effective/resolved values.

### Logs / stack traces

No stack trace. This is a config-resolution path bug found by source inspection while investigating runtime-profile behavior.

### ZeroClaw version

Current `master` at `dab313aff50cd221b278774f5a79a61c5eb50806`.

### Rust version

N/A; this is a runtime/channel config-resolution bug rather than a compiler/toolchain-specific failure.

### Operating system

Not expected to be OS-specific.

### Regression?

Unknown.

### Related

- #7796

### Pre-flight checks

- [x] I source-confirmed this on latest `master`.

ZOOWH:
## Summary

- Channel per-message tool-loop calls read `strict_tool_parsing` and `parallel_tools` from `ctx.prompt_config.agent(...).resolved`, which returns default `false` values because `AliasedAgentConfig.resolved` is skipped during deserialization and only populated by `resolved_agent_config()`.
- The correct source is `ctx.agent_cfg.resolved`, which is already wired through the same resolution boundary used for other runtime-profile knobs in this context (`max_history_messages`, `context_compression`, `keep_tool_context_turns`).
- Replaced the two `prompt_config.agent(...).resolved` reads with `ctx.agent_cfg.resolved` so channel turns honor the agent's runtime profile for tool parsing and parallel execution.
- Restored the focused regression test that proves the resolved agent config carries the runtime-profile values while the raw prompt-config copy remains at the serde-skipped defaults.

## Change Type

- [x] Bug fix / [ ] Feature / [ ] Refactor / [ ] Docs / [ ] Security / [ ] Chore

## Scope

- Channel orchestrator only: `crates/zeroclaw-channels/src/orchestrator/mod.rs`.

## Validation Evidence

**Behavior addressed:** Channel turns now use the resolved agent config for `strict_tool_parsing` and `parallel_tools` instead of falling back to default false values.

**Environment tested:** Linux x86_64, Rust stable.

**Steps run after the patch:** Verified the channel tool-loop call site now reads from `ctx.agent_cfg.resolved`. Compared with nearby resolved reads to confirm the same resolution pattern.

**Evidence after fix:**

```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 3m 11s
```

Additional maintainer validation on the refreshed head:

```text
cargo fmt --all -- --check
cargo test -p zeroclaw-channels --lib -- resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools
```

Current public CI is green at head 7ee89444ce78315bcc369dbfdecab78387ae5846.

**Observed result:** The channel tool-loop now receives the resolved `strict_tool_parsing` and `parallel_tools` values from `ctx.agent_cfg.resolved`, matching the resolution pattern used for the other runtime-profile fields in the same context.

**Not tested:** Live channel message delivery with an agent configured with `strict_tool_parsing = true` and `parallel_tools = true` via a runtime profile. The regression coverage is source/config-boundary coverage rather than a live daemon/channel smoke.

## Security & Privacy Impact

- New permissions, capabilities, or file system access scope? No.

## Compatibility

- Backward compatible? Yes.
- If `No` or `Yes` to either: N/A. This changes which already-existing runtime-profile values the channel tool-loop reads; it does not add or rename configuration.

## Rollback

- **Fast rollback command/path:** Revert the PR's source/test commits.
- **Observable failure symptoms:** Channel-driven agents would again ignore runtime-profile `strict_tool_parsing` or `parallel_tools` settings and fall back to default false behavior for those tool-loop knobs.

## Root Cause

- Root cause: The channel tool-loop accessed `ctx.prompt_config.agent(alias).resolved` for `strict_tool_parsing` and `parallel_tools`. Because `AliasedAgentConfig.resolved` is `#[serde(skip)]` and only populated by `resolved_agent_config()`, reading from the prompt config's raw agent yields default false values regardless of the runtime profile configuration.
- Missing detection/guardrail: The SSOT rule in AGENTS.md prohibits duplicate state, but `prompt_config.agent(...)` and `agent_cfg` both expose a `.resolved` field, creating a latent violation where channel code reads the wrong unresolved copy.

## Regression Test Plan

- Target test file: existing orchestrator tests in `crates/zeroclaw-channels/src/orchestrator/mod.rs`.
- Guardrail: `resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` verifies that channel runtime context construction can use `agent_cfg.resolved` populated from `resolved_agent_config()`, while the raw prompt-config agent remains at the serde-skipped defaults.

Operators who configure agents with runtime profiles setting `strict_tool_parsing = true` or `parallel_tools = true` will now see those settings honored during channel message turns, instead of silently falling back to false.

Closes #7809


Audacity88 (changes requested):
I reviewed the one-file diff, the PR body, linked issue #7809, and the green CI checks. Thanks for the narrow fix here; the source change points at the right resolved-config boundary, but I think this needs one focused regression test before it lands.

### 🔴 Blocking — Add a regression test for the resolved channel tool-loop knobs

The source fix looks correct: `ctx.agent_cfg.resolved` is the resolved agent config that `start_channels` builds via `resolved_agent_config()`, while `ctx.prompt_config.agent(...).resolved` can still be the serde-skipped default copy. The remaining problem is that the PR does not add the guardrail its own test plan describes. `cargo check -p zeroclaw-channels` proves the new field access compiles, but it would not catch a future channel-path regression that silently routes `strict_tool_parsing` or `parallel_tools` back through the raw `prompt_config` copy.

Please add a focused regression test on the channel/orchestrator path that configures an agent runtime profile with `strict_tool_parsing = true` and/or `parallel_tools = true`, then verifies the channel tool-loop receives the resolved values from `ctx.agent_cfg.resolved` rather than the raw agent defaults. There is already channel test scaffolding here for `ChannelRuntimeContext`, tool-call providers, and parallel dispatch, so this can stay narrow and should fail before this patch and pass after it.

### 🟢 What looks good — Correct source of truth

The implementation itself is nicely scoped to the two incorrect call-site arguments, and it matches the nearby resolved-config pattern already used for `max_history_messages`, `context_compression`, and `keep_tool_context_turns`.

ZOOWH:
@Audacity88, thanks for the review. I've added the requested regression test: `resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` in `crates/zeroclaw-channels/src/orchestrator/mod.rs`. The test demonstrates both outcomes:

- `prompt_config.agent(alias).resolved` returns `false` for both flags (the stale serde-skipped default the old code read from);
- `agent_cfg.resolved` returns `true` when the runtime profile sets them (the correct source after the fix).

The test should fail on the old code and pass with the fix. I've also rebased onto the latest master. Verified locally: `cargo test -p zeroclaw-channels --lib -- resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` passes.

Audacity88:
Thanks for adding the regression test. This now looks like the right shape, but the branch is conflicting with current `master` in `crates/zeroclaw-channels/src/orchestrator/mod.rs`.

Could you rebase/merge current `master` and resolve that conflict? The conflict appears mechanical: keep the current `ToolLoop { ... }` call shape from `master`, keep this PR's switch to `ctx.agent_cfg.resolved.strict_tool_parsing` / `ctx.agent_cfg.resolved.parallel_tools`, and keep the new regression test.

Once the branch is mergeable again and CI reruns on the refreshed head, I can re-review it.

ZOOWH:
Rebased onto current master and resolved the conflict in `crates/zeroclaw-channels/src/orchestrator/mod.rs`. Kept master's `ToolLoop { ... }` named-field call shape + this PR's switch to `ctx.agent_cfg.resolved.strict_tool_parsing` / `ctx.agent_cfg.resolved.parallel_tools`. The regression test is also preserved. CI should rerun on the refreshed head.

singlerider (dismissed):
Re-review (singlerider). Pulled the branch to head `3bdefe711f`, built `zeroclaw-channels` locally, and ran the new test (passes). The regression test Audacity88 asked for is now present and is exactly the right guardrail, so I am clearing the outstanding `CHANGES_REQUESTED`.

### ✅ Resolved — Regression test added for the resolved channel tool-loop knobs

`resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` configures a runtime profile with `strict_tool_parsing = true` and `parallel_tools = Some(true)`, resolves it through `Config::resolved_agent_config` (the same path `start_channels` uses), and asserts `agent_cfg.resolved` carries `true` for both while `prompt_config.agent(alias).resolved` stays at the serde-skipped `false` default. That pins both halves of the bug: it would fail against the old code that read from `prompt_config` and passes with the fix. Correct and narrow.

### 🟢 Source fix targets the right boundary

Replacing the two `ctx.prompt_config.agent(alias).resolved.*` reads with `ctx.agent_cfg.resolved.*` matches the resolved-config pattern already used nearby for `max_history_messages`, `context_compression`, and `keep_tool_context_turns`. Since `AliasedAgentConfig.resolved` is `#[serde(skip)]`, the old call site always passed `false`, silently disabling both knobs on every channel agent. This is the correct source of truth.

### 🔴 Trim the test docstring

The 25-line docstring on the new test reads as an essay restating the bug narrative three times. House style is a 1-2 line header on a test, not a paragraph-by-paragraph walkthrough; git blame and the PR body carry the history. Cut it to a couple of lines stating what the test pins. This is the one thing to fix before merge; the test body itself is correct.

Moving this to the v0.8.2 milestone: it is a `bug`-labeled break-fix that silently disabled `strict_tool_parsing` and `parallel_tools` for every channel-driven agent, which is high enough value to land in the nearest release rather than wait for v0.8.3.

singlerider:
Moved from the released v0.8.2 to v0.8.3. Channel orchestrator bug fix implementing #7809, which already lives in v0.8.3.

Audacity88:
@ZOOWH this one is quite close to mergeable, just needs the docstring update and a merge from master to fix the conflicts

singlerider:
@ZOOWH this PR currently conflicts with `master`. Merge `master` into the branch to resolve, then CI can re-run.

singlerider:
@ZOOWH follow-up: master now has the advisory-gate fix (#8520 merged), so once you merge the latest `master` into this branch the conflict clears and CI should go green. No rebase, merge only.

Audacity88 (changes requested):
@ZOOWH, I re-reviewed current head 2a05c3495471fa5dd39714821b78fa0b606af5fb against the live PR state, the current PR diff, linked Issue #7809, the prior reviews/comments, and the public source at both the current head and the earlier fixed head 3bdefe711f72629efa0c87b5be6f70df221c4a18. I did not run local Cargo in this review-only pass.

### 🔴 Blocking — The current head dropped the fix and regression test

The branch refresh appears to have resolved the conflict by taking the current `master` version of `crates/zeroclaw-channels/src/orchestrator/mod.rs` and dropping this PR's actual change. The live PR diff is now empty, and current head 2a05c3495471fa5dd39714821b78fa0b606af5fb still passes both tool-loop flags through `ctx.prompt_config.agent(ctx.agent_alias.as_str()).is_some_and(|agent| agent.resolved...)`.

That is the stale source described in #7809: `AliasedAgentConfig.resolved` is skipped during deserialization, so this path falls back to the default `false` values instead of the runtime-profile values. The earlier fixed head 3bdefe711f72629efa0c87b5be6f70df221c4a18 had the right call-site shape, using `ctx.agent_cfg.resolved.strict_tool_parsing` and `ctx.agent_cfg.resolved.parallel_tools`, and it also included the requested `resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` regression test. Both are gone from the current PR diff.

Please restore the actual fix and the regression test on top of current `master`. When restoring the test, keep the previous docstring cleanup ask from singlerider's review: the test body was correct, but the long explanatory docstring should be trimmed to a short comment. After that, rerun the focused channel test and let CI run on the refreshed head.

Audacity88 (changes requested):
@ZOOWH thanks for refreshing this again. I re-reviewed current head `d3d58870aeca1cd8b61cbf5bb02161ce605d1641` against the live PR state, the current combined diff against `master`, linked Issue #7809, and the earlier reviews/comments. I did not run local Cargo validation in this review-only pass.

### ✅ Resolved — The source fix is restored

The current head restores the important source change from my prior review. The channel tool-loop now passes `strict_tool_parsing: ctx.agent_cfg.resolved.strict_tool_parsing` and `parallel_tools: ctx.agent_cfg.resolved.parallel_tools`, so it is reading the already-resolved agent config instead of the raw `prompt_config.agent(...).resolved` copy.

### 🔴 Blocking — Restore or strengthen the regression coverage too

The regression test is no longer present in the current PR head. The current combined diff against `master` is only the two call-site argument changes in `crates/zeroclaw-channels/src/orchestrator/mod.rs`; a PR-head grep does not find `resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools`, `strict-parallel`, or the earlier test's resolved/raw-agent assertions.

That means the branch restores the implementation but not the guardrail requested in the first review and previously confirmed by singlerider. Please restore or strengthen the missing coverage so it fails if this channel tool-loop path is wired back to the raw `prompt_config.agent(alias).resolved` defaults. A good shape would still configure a runtime profile with `strict_tool_parsing = true` and `parallel_tools = true`, resolve the agent through `resolved_agent_config()`, and keep the raw-vs-resolved assertions, but it should also connect that setup to the channel call path or otherwise capture the `strict_tool_parsing` / `parallel_tools` values passed into the loop.

When restoring the test, please also keep singlerider's docstring feedback: trim the earlier 25-line docstring to a one- or two-line header that states what the test pins.

### 🟢 What looks good — The actual runtime source is now right

The call-site change itself is the right fix for #7809. It keeps channel turns on the same resolved-config boundary as the nearby runtime-profile fields, and it avoids duplicating or re-resolving state at the tool-loop call site.

singlerider:
Hi @ZOOWH, maintainer note.

v0.8.3 is imminent and we are pushing to close the milestone this week. Your PR was CONFLICTING against master, blocking the merge queue, so I merged `upstream/master` into your branch and resolved the single conflict in `crates/zeroclaw-channels/src/orchestrator/mod.rs` in place. Master had refactored the `run_tool_call_loop` args into grouped `ResolvedIo` / `ResolvedRuntimeKnobs` structs, and I preserved your `#7809` fix by reading `strict_tool_parsing` and `parallel_tools` from `ctx.agent_cfg.resolved` inside the new struct shape.

`cargo fmt` / `cargo clippy -p zeroclaw-channels --all-targets -- -D warnings` clean locally. Please review the merge commit and address the outstanding review notes when you have a moment.

Apologies for the unsolicited push, and thanks for the contribution.

singlerider:
Hi @ZOOWH, maintainer follow-up.

v0.8.3 release is imminent, and this PR had gone stale (last activity 2026-07-02), so I did an automated pass to move things along. Two things landed on your branch:

- `9248e99f9` merged `upstream/master` and resolved the `orchestrator/mod.rs` conflict, preserving your `#7809` fix inside master's new `ResolvedIo` / `ResolvedRuntimeKnobs` grouping.
- `7ee89444c` restored `resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools`, the regression test Audacity88 asked for in the 2026-07-02 review that got dropped in a prior rebase. Docstring is trimmed to two lines per the earlier `singlerider` feedback.

Local validation clean: `cargo fmt --all -- --check`, `cargo clippy -p zeroclaw-channels --all-targets -- -D warnings`, and the new test passes on Rust 1.96.1.

I dismissed my own earlier `CHANGES_REQUESTED` since the docstring ask is addressed at this head. Re-requesting review from Audacity88.

Apologies for the unsolicited work on your PR, and thanks for the contribution.

Audacity88 (approved):
@ZOOWH I re-reviewed current head 7ee89444ce78315bcc369dbfdecab78387ae5846 against the live PR state, linked issue #7809, duplicate implementation PR #7851, the prior reviews/comments, current CI, and the current source around the channel `run_tool_call_loop` call. I did not run local Cargo in this review-only pass.

This approval clears my earlier `CHANGES_REQUESTED` on the restored source fix and regression coverage. It is not a merge-readiness call.

### ✅ Resolved — Source fix is restored on the current tool-loop shape

The current diff now preserves the #7809 fix inside the newer `ResolvedRuntimeKnobs` call shape from `master`. The channel path passes `strict_tool_parsing: ctx.agent_cfg.resolved.strict_tool_parsing` and `parallel_tools: ctx.agent_cfg.resolved.parallel_tools`, so it is reading the already-resolved agent config instead of the raw `prompt_config.agent(alias).resolved` copy.

That matches the issue's source-of-truth boundary: `ChannelRuntimeContext.agent_cfg` comes from `Config::resolved_agent_config(...)`, while `AliasedAgentConfig.resolved` on the raw prompt config remains the serde-skipped default copy.

### ✅ Resolved — Regression coverage is restored with the docstring trimmed

`resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` is back at the current head, and the long test docstring has been trimmed to the short two-line explanation requested in the earlier review. The test configures a runtime profile with both knobs enabled, resolves the agent config, and asserts the important split: `agent_cfg.resolved` carries `true` while `prompt_config.agent(alias).resolved` stays at the default `false`.

That is the guardrail my earlier reviews were asking for. It pins the raw-vs-resolved config boundary that caused the channel regression; the call-site wiring itself is covered by the source-reviewed diff above.

### 🟢 What looks good — Narrow active path for #7809

This is still the right active PR path for #7809. The duplicate one-file implementation in #7851 is already closed, and this branch now carries the implementation fix plus the requested regression test in one file.

tidux (approved):
I reviewed the one-file diff at head 7ee89444ce78315bcc369dbfdecab78387ae5846 against linked issue #7809, the full reviewer chronology (@Audacity88's original block through his approval, @singlerider's dismissal and milestone move), the current PR template, and the local source on both the channel and config sides. I also pulled the head into an isolated worktree and ran the channel crate's format, clippy, and the new regression test locally. This is my first review on this PR; it approves.

Thanks for staying with this one through several conflict cycles — the branch lost and re-gained the fix during rebases, and it landed correctly in the end.

### ✅ Resolved — Source fix reads the resolved boundary, not the serde-skipped copy

The two call-site arguments now read `ctx.agent_cfg.resolved.strict_tool_parsing` and `ctx.agent_cfg.resolved.parallel_tools` (orchestrator/mod.rs:5132-5133) instead of `ctx.prompt_config.agent(alias).resolved.*`. I ground-truthed the boundary end to end: `ChannelRuntimeContext.agent_cfg` is built at start_channels from `config.resolved_agent_config(agent_alias)` (mod.rs:9483 → cloned into `agent_cfg` at mod.rs:10179), while `prompt_config` is the raw `config.clone()`. `AliasedAgentConfig.resolved` is `#[serde(skip)]` (schema.rs:3641-3642, doc-commented as "populated by resolved_agent_config from the runtime profile; not config-settable"), so the old read genuinely returned the default `false` on every channel turn. This is exactly the source-of-truth boundary #7809 prescribed.

### ✅ Resolved — Fix now matches the nearby resolved-config reads

The switch puts these two flags on the same boundary already used for the other runtime-profile knobs in this context: `max_history_messages` (mod.rs:2147), `thinking` (mod.rs:2879, 4346), and `keep_tool_context_turns` (mod.rs:5470) all read `ctx.agent_cfg.resolved.*`. No production `prompt_config.agent(...).resolved` read remains — the other surviving `prompt_config` accesses in this file are unrelated per-channel `ack_reactions` lookups, not the resolved-runtime pattern. This closes the split-brain the issue described rather than papering over it.

### ✅ Resolved — Regression coverage restored with the trimmed docstring

`resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools` is present at the current head with the docstring cut to the three-line header @singlerider and @Audacity88 asked for. It configures a runtime profile with both knobs enabled, resolves the agent through `resolved_agent_config()`, and pins both halves of the invariant: `agent_cfg.resolved` carries `true` while `prompt_config.agent(alias).resolved` stays at the serde-skipped `false`. That is the raw-vs-resolved divergence that caused the bug. It is a config-boundary test rather than a call-path assertion, which is the honest framing the PR body and the prior reviews already gave it — it guards the boundary the fix depends on.

### 🟢 What looks good — Narrow, correctly scoped, honest body

+55/-8 in a single file, with the source change limited to the two incorrect arguments and the rest being the guardrail test. The PR body fills every required template section including the risk:high Rollback fields, and the Validation Evidence "Not tested" note honestly scopes the coverage as source/config-boundary rather than a live channel smoke — no overclaiming. `Closes #7809` is the right verb: the diff fully resolves the issue as written.

### 🔵 Suggestion — Drop the Claude co-author trailer at squash time

One of the mid-branch merge commits (2a05c3495471) carries a `Co-Authored-By: Claude <noreply@anthropic.com>` trailer. It is only in a merge-commit message, not the PR body, and a squash merge collapses it away — so this is not a blocker and needs nothing from you. Flagging only so whoever merges keeps the squash body clean of the AI attribution footer per the template's contributor-attribution rule.

### Local validation

In an isolated worktree at 7ee89444ce78315bcc369dbfdecab78387ae5846:

```text
cargo fmt -p zeroclaw-channels -- --check        # clean (exit 0)
cargo clippy -p zeroclaw-channels --all-targets -- -D warnings   # clean (exit 0)
cargo test -p zeroclaw-channels --lib -- resolved_agent_config_carries_strict_tool_parsing_and_parallel_tools
```

Public CI is green on the head (24 required checks pass; the Windows targeted Clippy job is correctly skipping). Branch is MERGEABLE / CLEAN, and reviewDecision is APPROVED — @Audacity88's earlier CHANGES_REQUESTED reviews are superseded by his approval on this head and @singlerider's was dismissed, so nothing is holding this back on the review side.

PR #7838, opened 2026-06-17
feat(gateway): agent owned-state cascade on delete (#7175)

Nillth:
> ### 📋 Stacked series — **4 of 8**
>
> **Sequence:** `1` #7785 ✅ · `2` #7830 ✅ · `3` #7837 · **`4` #7838 ← this PR** · `5` #7839 · `6` #7840 · `7` #7841 · `8` #7842

**This slice (own delta):** 13 files, **+739 / −43** (gateway + infra + runtime + memory). **Cumulative shown until #3 merges:** 14 files, +1004 / −55.

## Summary
Completes agent deletion. Today deleting an agent only removes the config entry — it leaves config refs dangling **and** orphans the agent's persisted state. This wires the gateway to scrub config refs via `delete_with_cascade(Agent)` (slice 3) and cascade the **owned non-config state**: memory rows, workspace dir, cron jobs, ACP sessions, and session-metadata attribution.

- **Refuses on HARD**: an enabled `heartbeat.agent`, or **live ACP sessions** (`killed_at IS NULL`). The ACP gate **fails closed** — if the session store can't be read it refuses rather than risk orphaning live sessions.
- **Export-then-delete**: each store's rows are written to `agents/_deleted/<alias>-<ts>/cascade/*.json` + `manifest.json`, then removed. Per-store failures are **surfaced** (`OwnedStateReport.warnings` + manifest + WARN), never masked as success.
- **Also fixes a latent gap**: the old agent-delete path didn't scrub config refs, leaving `heartbeat`/`peer-group`/`delegate` refs dangling.

**Persistence fix (load-bearing, /verify-confirmed):** `Config::save_dirty` writes only *marked-dirty* paths; this handler previously marked only `agents.<alias>`, so a soft-ref scrubbed in another entry stayed stale on disk → dangling ref on reload → `validate()` rejects → **the daemon won't boot**. Fixed by marking **every** touched entry (`cascade.dirty_paths()`) before persist. `/verify` PASS: after a delete the referrer's `delegates` is empty in the config FILE and the daemon restarts clean.

## Review: 2 blockers found + fixed
1. **Non-SQLite backends silently orphaned agent memory** (only SqliteMemory implemented `purge_agent`/`export_agent`; others hit the `bail!` default, masked by `unwrap_or(0)`). Fixed: implemented for **Lucid** (delegates to inner sqlite), **Postgres** (DELETE/SELECT by agent), **Qdrant** (scroll + delete-matching). Markdown/none legitimately have no DB rows — their `bail` now surfaces as a warning, not a false success.
2. **The HARD live-ACP gate failed OPEN** on a store-open error. Fixed: `live_acp_session_count` returns `Result`; the handler refuses (fails closed) when it can't verify.

## Crates touched
`zeroclaw-gateway` (`agent_owned_state.rs` coordinator + `handle_delete_map_key` refactor) · `zeroclaw-infra` (ACP by-agent + `SessionBackend::clear_agent_attribution`) · `zeroclaw-runtime` (`cron::{list,remove}_jobs_by_agent`) · `zeroclaw-api` + `zeroclaw-memory` (`export_agent` + backend impls).

## Validation
Verified at the stack tip: `cargo clippy --workspace --all-targets` + `cargo fmt --all --check` **clean**. End-to-end `/verify` of the agent-delete cascade (cascade + refuse-on-HARD paths) PASS. Full CI matrix (incl. all memory features) runs on this draft.

## Security & Privacy
New fs writes: the cascade archive under `data_dir/agents/_deleted/` (operator-controlled) — agent memory **content** + cron fields + ACP session *metadata* (not transcripts). No new network/permissions/secrets. No PII in tests.
## Rollback (risk: medium) — `git revert <sha>`; no migration; affects only the delete path.

Related #7175.

### DIFF
--- a/crates/zeroclaw-gateway/src/api_config.rs
+++ b/crates/zeroclaw-gateway/src/api_config.rs
@@ -1086,6 +1086,13 @@ async fn delete_agent_cascade(
         );
     }
 
+    // Resolve the workspace dir BEFORE the config cascade removes the agents
+    // entry: `agent_workspace_dir` only returns an operator-set custom
+    // `workspace.path` while the entry exists; after removal it silently falls
+    // back to the default `install_root/agents/<alias>/workspace`, so a
+    // custom-workspace agent's real dir would otherwise never be archived.
+    let workspace = working.agent_workspace_dir(alias);
+
     // Config cascade: scrub soft refs + remove the agents entry.
     let cascade = match alias_refs::delete_with_cascade(
         &mut working,
@@ -1107,8 +1114,8 @@ async fn delete_agent_cascade(
 
     // Archive into the shared graveyard `<data_dir>/agents/_deleted/<alias>-<ts>/`
     // (not inside the deleted agent's own dir), and give the owned-state exports
-    // a home there even if the agent had no workspace dir.
-    let workspace = working.agent_workspace_dir(alias);
+    // a home there even if the agent had no workspace dir. (`workspace` was
+    // resolved above, before the cascade removed the entry.)
     let ts = chrono::Utc::now().format("%Y%m%d%H%M%S");
     let archive_dir = working
         .data_dir
@@ -2032,6 +2039,41 @@ mod tests {
         assert_eq!(
             dirty_entry_for("providers.models.anthropic.default.fallback[0]"),
             "providers.models.anthropic.default"
+        );
+    }
+
+    #[test]
+    fn delete_cascade_resolves_custom_workspace_before_removing_entry() {
+        // Regression: `delete_agent_cascade` must resolve `agent_workspace_dir`
+        // BEFORE `delete_with_cascade` removes the agents entry. The method only
+        // returns an operator-set custom `workspace.path` while the entry exists;
+        // resolving it after removal silently yields the DEFAULT path, so a
+        // custom-workspace agent's real dir would never be archived.
+        let custom = std::path::PathBuf::from("/var/lib/zc-test/custom-victim-ws");
+        let mut cfg = zeroclaw_config::schema::Config::default();
+        cfg.agents.insert(
+            "victim".to_string(),
+            zeroclaw_config::schema::AliasedAgentConfig::default(),
+        );
+        cfg.agents.get_mut("victim").unwrap().workspace.path = Some(custom.clone());
+
+        // While the entry exists → the custom path (what the handler captures).
+        assert_eq!(cfg.agent_workspace_dir("victim"), custom);
+
+        // After the cascade removes the entry → it falls back to the DEFAULT
+        // path; that is exactly why resolution must happen before the cascade.
+        zeroclaw_config::alias_refs::delete_with_cascade(
+            &mut cfg,
+            &zeroclaw_config::alias_refs::AliasKind::Agent,
+            "victim",
+            zeroclaw_config::alias_refs::CascadePolicy::RefuseOnHard,
+        )
+        .expect("soft-only agent delete succeeds");
+        assert!(!cfg.agents.contains_key("victim"));
+        assert_ne!(
+            cfg.agent_workspace_dir("victim"),
+            custom,
+            "after removal the custom workspace path defaults — resolve BEFORE the cascade"
         );
     }
 
--- a/crates/zeroclaw-gateway/src/api_config.rs
+++ b/crates/zeroclaw-gateway/src/api_config.rs
@@ -997,7 +997,15 @@ pub async fn handle_delete_map_key(
     if let Err(e) = require_auth(&state, &headers) {
         return e.into_response();
     }
-    let mut working = state.config.read().clone();
+    let working = state.config.read().clone();
+    // Agent deletion is special: it must scrub config references (heartbeat,
+    // peer-groups, delegates, workspace.access, …) via `delete_with_cascade`
+    // and cascade owned non-config state (memory / cron / acp / session). The
+    // generic map-key delete below handles every other section unchanged.
+    if q.path == "agents" {
+        return delete_agent_cascade(&state, working, &q.key).await;
+    }
+    let mut working = working;
     let removed = match working.delete_map_key(&q.path, &q.key) {
         Ok(b) => b,
         Err(msg) => {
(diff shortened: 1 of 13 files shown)


singlerider (dismissed):
I reviewed slice 4 on its merits against current `master` (ignoring the conflict, which is just slice 3 having landed). I read the new `delete_agent_cascade` handler, the `agent_owned_state` coordinator, the memory/cron/acp/session store methods, and cross-checked against `agent_workspace_dir`, `apply_dirty_path`, `delete_with_cascade`, and the ACP/cron FK schema. I confirmed the new tests green locally (`dirty_entry_for`, `per_agent_cascade…`, `sqlite_export_agent…`, infra acp suite 21/21). The persistence fix is correct and the design is genuinely good. One blocker: a use-after-remove ordering bug silently skips the workspace archive for custom-workspace agents.

### 🔴 Blocking: workspace archive is resolved after the config entry is removed, so custom-workspace agents are never archived

In `delete_agent_cascade` the order is:

```rust
let cascade = delete_with_cascade(&mut working, &AliasKind::Agent, alias, …)?; // removes the entry
…
let workspace = working.agent_workspace_dir(alias); // entry is already gone
```

`delete_with_cascade` → `delete_agent` runs `cfg.agents.remove(alias)`, so by the time `agent_workspace_dir(alias)` is called the agent is no longer in `working.agents`. But `agent_workspace_dir` only returns the operator-set custom path when it can look the entry up:

```rust
pub fn agent_workspace_dir(&self, agent_alias: &str) -> PathBuf {
    if let Some(cfg) = self.agents.get(agent_alias)
        && let Some(custom) = cfg.workspace.path.as_ref()
    { return custom.clone(); }
    self.install_root_dir().join("agents").join(agent_alias).join("workspace")
}
```

With the entry already removed, `self.agents.get(alias)` is `None`, so this silently falls through to the **default** `install_root/agents/<alias>/workspace` path. For an agent that had a custom `workspace.path`, the handler then checks `workspace.exists()` on the wrong (default) location, finds nothing, and skips the archive entirely. The real workspace dir is left on disk, unarchived and unreferenced. That defeats the export-then-delete recoverability guarantee precisely for the agents most likely to hold operator data, and it does so silently (no warning, because from the handler's view there was simply no workspace to move).

Fix: resolve the workspace path **before** the config cascade removes the entry. Capture `let workspace = working.agent_workspace_dir(alias);` (and the `archive_dir`, which is config-independent) prior to the `delete_with_cascade` call, then archive after. The owned-state cascade and persist can stay where they are. A regression test with a custom `workspace.path` agent asserting the archive lands at the custom location would lock this; every existing path-related test appears to use default-workspace agents, which is why this slipped through.

### 🟢 The persistence fix is correct and is the load-bearing part of this slice

The `save_dirty`-only-writes-marked-paths trap is real and the fix is right. `dirty_entry_for` truncates each scrubbed `RefSite.path` back to the entry/section path that `apply_dirty_path` re-serialises wholesale (`agents.<name>`, `peer_groups.<g>`, the 4-segment provider entry, or the index-stripped scalar/vector field). I traced this against `apply_dirty_path`'s `should_delete`/`set_path_in_doc` branch: re-serialising the whole `agents.lead` subtree captures a scrubbed `delegates[0]`, and the removed agent's own entry resolves to `mem_val = None` → deleted from the doc. Pulling both `applied` sites and `deleted_entry` into `delete_cascade_dirty_paths`, sorted + deduped, mirrors rename's `RenameReport.dirty_paths` exactly. The unit test pins all the truncation shapes. This is the bug that would otherwise brick the daemon on reload, and it is handled well.

### 🟢 Fail-closed live-ACP gate and surfaced-not-masked failures are exactly right

- `live_acp_session_count` returns `Result` and the handler refuses the delete when it can't read the store, with a clear operator-facing reason. That is the correct posture for a HARD gate - an unreadable store must not be read as "zero live sessions." `AcpSessionStore::new` creating a missing DB empty means a fresh no-ACP system still returns `Ok(0)` and deletes proceed, so the fail-closed branch only fires on a genuine store error.
- The owned-state coordinator surfaces every per-store failure into `OwnedStateReport.warnings` + manifest + a WARN log rather than masking it as `0`/success. The earlier `unwrap_or(0)` masking gap is genuinely closed. Memory `purge_agent`/`export_agent` are now implemented for Sqlite, Lucid (delegates), Postgres (DELETE/SELECT by agent id), and Qdrant (scroll + delete-matching); markdown/none correctly surface their `bail` as a warning since their rows live in the archived workspace, not a DB.

### 🟢 Store-level cascade methods are correct and FK-backed

- ACP: `count_live_sessions_by_agent` keys on `killed_at IS NULL` (the live signal), `delete_sessions_by_agent` relies on real `ON DELETE CASCADE` FKs with `PRAGMA foreign_keys = ON` for `acp_messages`/`acp_tool_calls`/`acp_session_events` - I confirmed both the pragma and the FK clauses in the schema. The test pins live-count vs. all-sessions-export vs. per-agent-delete isolation.
- cron: `cron_runs` cascades off the `job_id` FK (`ON DELETE CASCADE` + pragma confirmed). Deleting jobs for a gone agent is the right call - they can never run.
- session: `clear_agent_attribution` nulls `agent_alias` rather than deleting the row, correctly preserving a possibly channel-shared conversation while dropping stale attribution.

Scoping owned-state to the surface (which owns the infra stores `zeroclaw-config` can't depend on) and keeping the config cascade pure is the right boundary. Fix the workspace-archive ordering and this lands.

crates/zeroclaw-gateway/src/api_config.rs
```diff
@@ -1059,6 +1028,180 @@ pub async fn handle_delete_map_key(
...
+                )
+                .with_path(format!("agents.{alias}")),
+            );
+        }
+    };
+
+    // Archive into the shared graveyard `<data_dir>/agents/_deleted/<alias>-<ts>/`
+    // (not inside the deleted agent's own dir), and give the owned-state exports
+    // a home there even if the agent had no workspace dir.
+    let workspace = working.agent_workspace_dir(alias);
```
singlerider: By this point `delete_with_cascade` (above) has already run `cfg.agents.remove(alias)`, so `working.agents.get(alias)` is `None` here. `agent_workspace_dir` returns the custom `workspace.path` only when it can look the entry up; with the entry gone it falls through to the default `install_root/agents/<alias>/workspace`. For a custom-workspace agent that means the `workspace.exists()` check below probes the wrong path, finds nothing, and the real workspace is never archived (silently orphaned on disk).

Resolve this before the config cascade removes the entry. Something like, just above the `delete_with_cascade` call:

```rust
let workspace = working.agent_workspace_dir(alias);
let archive_dir = working
    .data_dir
    .join("agents")
    .join("_deleted")
    .join(format!("{alias}-{ts}"));
```

then run the cascade and archive after. `archive_dir` is config-independent so it can move up freely too. A regression test with an agent that sets a custom `workspace.path`, asserting the archive lands at that custom location, would lock it. The existing path tests all use default-workspace agents, which is why this slipped through.

Nillth:
Thanks — real bug, fixed in 927921e.

You're right: `delete_agent_cascade` resolved `agent_workspace_dir` after the cascade removed the entry, so a custom-`workspace.path` agent fell through to the default path and the archive was silently skipped. Now the workspace dir is captured **before** `delete_with_cascade` runs, archived after. Added `delete_cascade_resolves_custom_workspace_before_removing_entry`, which pins that `agent_workspace_dir` returns the custom path while the entry is present and the default once removed — locking the ordering the handler depends on.

(Rebased onto current master in the same push, so the slice is conflict-free now.)

singlerider (approved):
Re-reviewed at `927921e` (my prior changes-requested was at `ad82a74`, now dismissed). The 🔴 workspace-archive ordering bug is fixed correctly. I traced the handler ordering and ran the new regression test plus the slice-4 store suites green locally. Approving.

### RESOLVED ✅ - workspace dir is now resolved before the cascade removes the entry

`delete_agent_cascade` now captures the workspace path up front:

```rust
let workspace = working.agent_workspace_dir(alias);   // line ~1094, entry still present
let cascade = delete_with_cascade(&mut working, &AliasKind::Agent, alias, …)?; // removes entry
…
let archive_dir = working…;   // config-independent, fine after
if workspace.exists() { /* archive the pre-resolved path */ }
```

This is exactly the fix I asked for: `agent_workspace_dir` returns the operator-set custom `workspace.path` only while the entry exists, so resolving before `delete_with_cascade` runs `cfg.agents.remove(alias)` keeps a custom-workspace agent from silently falling through to the default path and skipping the archive. The config-independent `archive_dir` correctly stays after the cascade.

The regression test `delete_cascade_resolves_custom_workspace_before_removing_entry` pins both halves of the invariant the handler depends on: `agent_workspace_dir` returns the custom path while the entry is present, and defaults once removed. That locks the ordering against future drift. I ran it green along with the slice-4 store suites (acp `per_agent_cascade…` 21/21, memory `sqlite_export_agent…`) to confirm nothing else moved.

### 🟢 Everything from the prior pass still holds

The `dirty_entry_for` persistence fix, fail-closed live-ACP gate, surfaced-not-masked owned-state warnings, all-backend memory impls, and FK-backed ACP/cron cascades are unchanged by this commit - it is the workspace-ordering fix plus its test, on a clean rebase onto current master (no longer conflicting). Slice 4 is good to land in order.

PR #7840, opened 2026-06-17
feat(config): rename_with_cascade for aliased entries (#7468)

Nillth:
> ### 📋 Stacked series — **6 of 8**
>
> **Sequence:** `1` #7785 ✅ · `2` #7830 ✅ · `3` #7837 · `4` #7838 · `5` #7839 · **`6` #7840 ← this PR** · `7` #7841 · `8` #7842

**This slice (own delta):** `crates/zeroclaw-config/src/alias_refs.rs` only — **+698**. **Cumulative shown until #3–#5 merge:** 14 files, +1992 / −76.

## Summary
Implements the config-layer half of #7468 (alias rename). `rename_with_cascade(cfg, kind, old, new)` is the **mutating inverse of `delete_with_cascade`'s scrub**: where delete clears/drops soft refs (and refuses on hard ones), rename **rewrites** every reference — soft *and* hard — to name the new alias, so nothing dangles and no HARD ref blocks (an enabled `heartbeat.agent` or a `peer_groups.<g>.channel` just follows the rename).

- **Flow:** reject no-op / reserved name → swap the entry key via `Config::rename_map_key` (validates the new key, blocks the `_deleted` marker, refuses a collision) → rewrite referrers → verify `find_all_references(old)` is empty.
- **Covers every kind:** agents, model / TTS / transcription providers, channels. Rename has no owned-state complications, so unlike delete it implements **TTS/transcription** too (delete left those `NotImplemented`).
- **`rewrite_*_refs` mirror the `collect_*_refs` walks** field-for-field and trim-for-trim. Any drift is caught by the post-condition (`RenameError::PostCondition`) — a missed site fails loudly rather than silently dangling.
- **`RenameError`** = `NotFound` | `InvalidName` | `Reserved` | `PostCondition` — no `Refused`; rename rewrites hard refs rather than refusing.
- **Reserved-name guard is agent-scoped:** the `default` agent is the runtime fallback, so renaming it away/onto is refused; `default` is the conventional single-instance key for providers/channels, which operators manage freely, so it is **not** reserved there.

## Notable behaviour (tested)
- **Channel rename does NOT hit the bare-group-member orphan that makes a *delete* refuse** — renaming rewrites the member's channel to `<type>.<new>`, so membership stays valid (`rename_channel_rewrites_refs_and_preserves_bare_group_membership`).
- **Self-rename** of an agent referencing itself (delegates + workspace.access key + read_memory_from + peer group) rewrites every field in one pass.
- **Padded refs** (`" anthropic.old "`) are trim-matched and rewritten to clean `anthropic.new`.

## Policy note
The `default` reserved guard is **agent-only** by deliberate choice — protecting every `default` across provider families and channel types would break conventional create/delete/rename of `default` instances. Widening it to providers/channels is a one-line change if preferred.

## Validation
`cargo clippy --workspace --all-targets` + `cargo fmt --all --check` **clean** at the stack tip. Rename test matrix: all-ref-kinds (agent incl. self-ref + heartbeat/acp/delegates/access/read_memory_from/peer-group) · not-found · collision · no-op · reserved-default both directions · `_deleted` target rejected · model-provider dotted rewrite · channel rewrite + bare-group membership preserved · tts scalar rewrite. Full CI runs on this draft. Pure library arm — exercised through the rename surface in slice 7.

## Rollback (risk: low) — `git revert <sha>`; no migration; no persisted-state effect.

Related #7468.

### DIFF
--- a/crates/zeroclaw-config/src/alias_refs.rs
+++ b/crates/zeroclaw-config/src/alias_refs.rs
@@ -503,6 +503,409 @@ fn scrub_channel_refs(cfg: &mut Config, target: &str) {
         .retain(|ch| ch.trim() != target);
 }
 
+// ── rename-with-cascade (#7468) ─────────────────────────────────────────────
+
+/// The agent alias reserved as the runtime fallback. `resolved_runtime_agent_alias`
+/// prefers it, so renaming it away — or onto it — would silently change which
+/// agent answers when no explicit target is given. Protected from rename. This
+/// guard is **agent-specific**: `default` is the conventional single-instance key
+/// for providers/channels (e.g. `providers.models.anthropic.default`,
+/// `channels.discord.default`), which operators rename/delete freely, so it is
+/// reserved only for the agent kind. (The `_deleted` archive marker is rejected
+/// as a new alias of any kind by `validate_alias_key`'s leading-underscore rule,
+/// which `rename_map_key` enforces — no separate guard needed here.)
+const RESERVED_DEFAULT_AGENT: &str = "default";
+
+/// Outcome of a successful [`rename_with_cascade`].
+#[derive(Debug, Clone)]
+pub struct RenameReport {
+    pub target_kind: AliasKind,
+    /// The previous alias (now gone from config).
+    pub old_alias: String,
+    /// The alias the entry now lives under.
+    pub new_alias: String,
+    /// Every dotted config path the rename mutated, **deduplicated and sorted**:
+    /// the renamed entry (old key — removed on disk; new key — added) plus the
+    /// entry/section path of each referrer that was rewritten. The persisting
+    /// surface must mark **each** of these dirty before saving — `save_dirty`
+    /// only writes marked paths, so a referrer in another entry that isn't
+    /// listed here would be rewritten in memory but left stale on disk. Paths are
+    /// at entry/section granularity (e.g. `agents.lead`, `peer_groups.crew`,
+    /// `heartbeat.agent`, `model_routes`, `providers.models.anthropic.default`)
+    /// so marking one re-serialises the whole changed subtree, capturing nested
+    /// edits like a `workspace.access` key rename.
+    pub dirty_paths: Vec<String>,
+}
+
+/// Why a [`rename_with_cascade`] did not complete. Unlike [`CascadeError`] there
+/// is no `Refused` variant: rename **rewrites** HARD references to follow the new
+/// name rather than refusing, so the only failures are bad inputs and the
+/// post-condition bug-guard.
+#[derive(Debug)]
+pub enum RenameError {
+    /// The source alias does not exist.
+    NotFound(String),
+    /// The new alias is unusable: fails `validate_alias_key`, collides with an
+    /// existing entry, or equals the current name. Carries the reason.
+    InvalidName(String),
+    /// The source or target alias is reserved (the `default` agent).
+    Reserved(String),
+    /// Bug guard: rewrite drifted from `find_all_references` and left a dangling
+    /// reference to the OLD alias. **The config WAS mutated** (key swapped + refs
+    /// rewritten) — the caller must NOT persist it. Unreachable while rewrite and
+    /// the collect_* walks mirror each other (same sites, same trim split).
+    PostCondition(String),
+}
+
+impl std::fmt::Display for RenameError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::NotFound(p) => write!(f, "alias not found: {p}"),
+            Self::InvalidName(m) => write!(f, "invalid new alias: {m}"),
+            Self::Reserved(a) => write!(f, "alias `{a}` is reserved and cannot be renamed"),
+            Self::PostCondition(m) => write!(f, "rename post-condition failed: {m}"),
+        }
+    }
+}
+
+impl std::error::Error for RenameError {}
+
+/// Rename an aliased entry from `old_alias` to `new_alias`, rewriting every
+/// reference to it. The mutating inverse of [`delete_with_cascade`]'s scrub:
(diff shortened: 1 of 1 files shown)


singlerider (approved):
I reviewed slice 6 on its own delta (`d070a19`), the config-layer half of #7468 (rename). I read `rename_with_cascade`, all three `rewrite_*_refs` arms, `RenameReport`/`RenameError`, and cross-checked each rewrite against its `collect_*_refs` mirror and against `find_all_references` (which the post-condition relies on). I confirmed the rename matrix green locally (11 rename tests in `alias_refs`, full suite passing, clippy clean). The collect/rewrite mirror is field-for-field complete and the design is clean. Approving on merits.

This stays draft and merges strictly after slices 3-5. It pivots the series from delete (#7175) to rename (#7468); nothing here depends on the slice-4 workspace-archive fix.

### 🟢 Every rewrite arm mirrors its collect walk field-for-field

This is the property that matters: a rewrite that misses a site collect knows about would leave a silent dangle, caught only by the post-condition. I walked all three:

- **agent**: rewrite touches heartbeat.agent, acp.default_agent, delegates (trim), workspace.access map key (raw re-key preserving AccessMode), read_memory_from (raw), peer_groups.agents (raw) - exactly the six sites `collect_agent_refs` collects, with the same trim/raw split.
- **model provider**: model_provider, classifier_provider, every provider's fallback[], model_routes[], embedding_routes[] - matches the five `collect_provider_refs` Models-arm sites.
- **channel**: agents.channels[], peer_groups.<g>.channel (dotted), escalation.alert_channels[] - matches the SOFT + dotted-HARD sites of `collect_channel_refs`.

The bare-group last-alias/orphan HARD refs that `collect_channel_refs` emits for a *delete* are correctly NOT rewritten here: a rename re-points the member's channel to `<type>.<new>` rather than emptying the block, so membership stays valid. `rename_channel_rewrites_refs_and_preserves_bare_group_membership` pins exactly that, and it is the right call.

### 🟢 Post-condition is anchored on the same single source of truth, and HARD-rewrite-not-refuse is correct

`find_all_references` delegates to the same `collect_*_refs` walks the rewrite mirrors, so the targeted post-condition (`find_all_references(old)` must be empty) converts any rewrite/collect drift into a loud `RenameError::PostCondition` with the config left unpersisted. The structural caveat - if both collect and rewrite were to miss the same real site, the post-condition is blind to it - is inherent to the shared-walk design and is the same posture as `delete_with_cascade`; anchoring everything on the validate-mirroring collect walk is the right way to manage it, and the `RenameError` doc states it plainly. No `Refused` variant is correct: rename rewrites HARD refs (an enabled `heartbeat.agent`, a `peer_groups.<g>.channel`) to follow the new name rather than refusing.

### 🟢 Guards and dirty-path contract are right

- no-op (`old == new`) → InvalidName; collision / bad key / `_deleted` marker → InvalidName via `rename_map_key`'s `validate_alias_key`; missing source → NotFound. All pinned.
- The agent-only `default` reserved guard is a sound deliberate choice: `default` is the runtime fallback agent (renaming it away or onto it silently changes dispatch), but a conventional operator-managed single-instance key for provider families and channel types. Widening it would break legitimate `default` rename for those kinds. Agreed with leaving it agent-scoped; the policy note documents the one-line path if that ever changes.
- `RenameReport.dirty_paths` returns entry/section granularity (`agents.<name>`, `peer_groups.<g>`, `heartbeat.agent`, `model_routes`, `providers.models.<ty>.<al>`) plus both the old and new entry keys, deduped + sorted. This is the same persistence contract slice 4 leaned on for delete, and it correctly captures nested edits (a `workspace.access` key rename) by re-serialising the whole entry subtree. The surface in slice 7 will consume this.

Covering TTS/transcription here (which delete left `NotImplemented`) is correct since rename has no owned-state half. Clean library arm.

Audacity88 (approved):
Reviewed the current public PR snapshot for refreshed head `ea7c0c8`: PR body/metadata, the one-file `alias_refs.rs` diff, the relevant public source around alias reference collection, existing public review context, comments/inline comments, and remote CI. I did not run local Cargo for this review-only pass. The config-layer rename primitive looks sound to me.

### 🟢 What looks good — Rewrite inventory mirrors the reference inventory

`rename_with_cascade` swaps the entry through `rename_map_key`, dispatches to the kind-specific rewrite arm, then checks `find_all_references(old)` before reporting success. I checked the added `rewrite_agent_refs`, `rewrite_provider_refs`, and `rewrite_channel_refs` arms against the reference sites collected by the existing alias-reference inventory:

- agent refs cover `heartbeat.agent`, `acp.default_agent`, delegates, `workspace.access`, `workspace.read_memory_from`, and `peer_groups.agents`;
- model-provider refs cover agent model/classifier providers, provider fallbacks, `model_routes`, and `embedding_routes`;
- TTS/transcription provider refs cover their agent scalar fields;
- channel refs cover agent channels, dotted peer-group channels, and escalation alert channels.

The trim-vs-raw matching is preserved where it matters, and hard references correctly follow the rename instead of blocking it.

### 🟢 What looks good — Persistence contract is explicit enough for the later surface

`RenameReport.dirty_paths` includes the old/new entry paths plus every touched referrer section, sorted and deduplicated. That gives the CLI/RPC layer a concrete list to mark dirty before saving, including whole-entry paths for nested rewrites like `workspace.access`.

### 🟢 What looks good — Guardrails match the intended alias model

The no-op, not-found, collision, `_deleted`/invalid-name, and agent-only `default` guards are all consistent with the existing schema helper behavior. Keeping the reserved `default` guard agent-scoped is the right tradeoff here: providers and channels commonly use `default` as an operator-managed alias, while the `default` agent has runtime fallback semantics.

Source review is clean from my side. The refreshed GitHub checks were still in progress during source review and were green on the final pre-post refresh. This approval is for the source/design of the slice, not a separate merge-ready signal beyond branch protection.

PR #7842, opened 2026-06-17
feat(cli): agents/providers/channels CRUD + skill-bundle cascade (#7468/#7175)

Nillth:
> ### 📋 Stacked series — **8 of 8**
> Typed delete-with-cascade (#7175) + alias rename (#7468), sliced for review. **Opened as draft for CI; merge strictly in order.** This is the tail of the stack, so its diff is the **cumulative** feature until #3–#7 merge; it collapses to its own delta (the CLI + skill-bundle commits) once they land.
>
> **Sequence:** `1` #7785 ✅ · `2` #7830 ✅ · `3` #7837 · `4` #7838 · `5` #7839 · `6` #7840 · `7` #7841 · **`8` #7842 ← this PR**

**This slice (own delta):** 6 files, **+1101 / −102** — three commits: alias CRUD verbs, agent owned-state in the CLI, skill-bundle cascade. **Cumulative shown until #3–#7 merge:** 18 files, +3639 / −122.

## Summary
The user-facing surface for the whole feature — gives the config/gateway cascade primitives their first CLI callers.

- **`zeroclaw {agents,providers,channels} {create,list,rename,delete}`** — new **non-breaking plural** subcommand groups over the shared cascade. The existing `agent <alias>` (run) command is untouched; `providers` with no verb still prints the supported-provider catalog. Delete is **safe-by-default** (impact preview unless `--yes`).
- **`default`-agent CLI guard:** the gateway delete relies only on HARD refs (acp.default_agent / enabled heartbeat); a `default` agent with neither would be deletable, so the CLI adds an explicit refuse + test.
- **Agent owned-state cascade from the CLI** (feature-gated `#[cfg(all(feature="gateway",feature="agent-runtime"))]`, config-only fallback + warning when off): stores opened from `data_dir` (no live gateway) → live-ACP gate → `cascade_owned_state` on delete (archives workspace) / `cascade_rename_agent` on rename (moves workspace).
- **`skills bundle {add,remove,rename}` now cascade:** create makes the dir; delete archives `shared/skills/<a>/` → `skills/_deleted/<a>-<ts>/` and strips the bundle from every agent's `skill_bundles`; rename moves the dir and rewrites the references. Implemented as standalone `find/scrub/rewrite_bundle_refs` (a single SOFT container — no new `AliasKind` variant, which would churn every exhaustive match for no gain). Adds `SkillBundleCommands::Rename`.

## Anti-drift refactor (PR8a)
Promotes the gateway's private `delete_cascade_dirty_paths` + `dirty_entry_for` into `zeroclaw-config` as `CascadeReport::dirty_paths()` (symmetric with `RenameReport.dirty_paths`), so the gateway delete handler and the CLI share one source of truth instead of duplicating the dirty-path derivation.

## Validation
`cargo clippy --workspace --all-targets` + `cargo fmt --all --check` **clean** at this tip (the union of the whole stack, default features incl. gateway + agent-runtime). Smoke-tested each verb against a temp `data_dir`: rename persisted to disk; agent delete archived workspace to `_deleted/` + wrote the cascade manifest; bundle rename moved the dir + rewrote the agent ref; bundle delete archived the dir + stripped the ref. Full CI matrix runs on this draft.

## Security & Privacy
New fs writes are the same operator-controlled `data_dir` archives as the gateway path (workspace + skill-bundle dirs under `_deleted/`). No new network/permissions/secrets. No PII in tests.
## Rollback (risk: medium) — `git revert <range>`; no migration; CLI-only surface.

Related #7468 · #7175.

### DIFF
--- a/src/alias_cli/mod.rs
+++ b/src/alias_cli/mod.rs
@@ -1,0 +1,405 @@
+//! CLI for alias CRUD — `zeroclaw {agents,providers,channels} {create,list,
+//! rename,delete}` (#7468 / #7175).
+//!
+//! Thin surface over the config-layer cascade in
+//! [`zeroclaw_config::alias_refs`]: `rename_with_cascade` / `delete_with_cascade`
+//! rewrite/scrub every reference and report the entry paths that changed; this
+//! module marks each dirty and persists via `Config::save_dirty` (which writes
+//! only marked paths). Plural groups (`agents`/`providers`/`channels`) are
+//! distinct from the singular `agent <alias>` run command, which is untouched.
+//!
+//! Providers and channels carry no owned non-config state, so their delete/
+//! rename is config-only. The agent owned-state cascade (memory / cron / acp /
+//! session rows + the workspace dir) is wired in a follow-up; until then agent
+//! delete/rename warn that owned state was not cascaded.
+
+use anyhow::{Context, Result, bail};
+use zeroclaw::{AgentsCommands, ChannelsCommands, ProvidersCommands};
+use zeroclaw_config::alias_refs::{
+    self, AliasKind, CascadeError, CascadePolicy, ProviderCategory, RenameError,
+};
+use zeroclaw_config::schema::Config;
+
+/// The agent alias reserved as the runtime fallback — protected from delete
+/// (rename is already guarded inside `rename_with_cascade`).
+const RESERVED_DEFAULT_AGENT: &str = "default";
+
+fn parse_provider_category(category: &str) -> Result<ProviderCategory> {
+    match category {
+        "models" => Ok(ProviderCategory::Models),
+        "tts" => Ok(ProviderCategory::Tts),
+        "transcription" => Ok(ProviderCategory::Transcription),
+        other => {
+            bail!("unknown provider category `{other}` (expected models | tts | transcription)")
+        }
+    }
+}
+
+/// The map-key section path for a kind (e.g. `agents`, `providers.models.anthropic`,
+/// `channels.discord`).
+fn section_path(kind: &AliasKind) -> String {
+    match kind {
+        AliasKind::Agent => "agents".to_string(),
+        AliasKind::Provider { category, family } => {
+            let cat = match category {
+                ProviderCategory::Models => "models",
+                ProviderCategory::Tts => "tts",
+                ProviderCategory::Transcription => "transcription",
+            };
+            format!("providers.{cat}.{family}")
+        }
+        AliasKind::Channel { channel_type } => format!("channels.{channel_type}"),
+    }
+}
+
+fn list_section(config: &Config, section: &str) -> Result<()> {
+    match config.get_map_keys(section) {
+        Some(mut keys) => {
+            keys.sort();
+            if keys.is_empty() {
+                println!("(no entries under {section})");
+            } else {
+                for k in keys {
+                    println!("{k}");
+                }
+            }
+        }
+        None => bail!("no such config section: {section}"),
+    }
+    Ok(())
+}
+
+fn create_entry(config: &mut Config, section: &str, alias: &str) -> Result<()> {
+    if config
+        .create_map_key(section, alias)
+        .map_err(anyhow::Error::msg)?
+    {
+        config.mark_dirty(&format!("{section}.{alias}"));
+        println!("created {section}.{alias}");
+    } else {
+        println!("{section}.{alias} already exists (no change)");
+    }
+    Ok(())
+}
+
+/// Print the dry-run impact (blockers + scrubs) for a delete.
+fn print_impact(kind: &AliasKind, alias: &str, config: &Config) {
+    let report = alias_refs::plan_delete(config, kind, alias);
+    if report.blockers.is_empty() {
+        println!(
+            "deleting {}.{alias} would scrub {} reference(s):",
+            section_path(kind),
+            report.scrubs.len()
+        );
+    } else {
+        println!(
+            "deleting {}.{alias} is BLOCKED by {} hard reference(s):",
+            section_path(kind),
+            report.blockers.len()
+        );
+        for b in &report.blockers {
+            println!("  ✗ {} (hard reference)", b.path);
+        }
+    }
+    for s in &report.scrubs {
+        println!("  • {} (would be scrubbed)", s.path);
+    }
+}
+
+/// Delete an aliased entry's config references (config-layer only).
+fn delete_config(
+    config: &mut Config,
+    kind: &AliasKind,
+    alias: &str,
+    dry_run: bool,
+    yes: bool,
+) -> Result<()> {
+    let section = section_path(kind);
+    if dry_run {
+        print_impact(kind, alias, config);
+        return Ok(());
+    }
+    if !yes {
+        print_impact(kind, alias, config);
+        println!("\nNo changes made. Re-run with --yes to apply (or --dry-run to preview).");
+        return Ok(());
+    }
+    match alias_refs::delete_with_cascade(config, kind, alias, CascadePolicy::RefuseOnHard) {
+        Ok(report) => {
+            for path in report.dirty_paths() {
+                config.mark_dirty(&path);
+            }
+            println!(
+                "deleted {section}.{alias} (scrubbed {} reference(s))",
+                report.applied.len()
+            );
+            Ok(())
+        }
+        Err(CascadeError::Refused(report)) => {
+            println!(
+                "refused: {} hard reference(s) block the delete:",
+                report.blockers.len()
+            );
+            for b in &report.blockers {
+                println!("  ✗ {}", b.path);
+            }
+            bail!("delete refused — resolve the hard references first");
+        }
+        Err(CascadeError::NotFound(p)) => bail!("{p} is not configured"),
+        Err(e) => bail!("delete failed: {e}"),
+    }
+}
+
+/// Rename an aliased entry's config references (config-layer only).
+fn rename_config(config: &mut Config, kind: &AliasKind, from: &str, to: &str) -> Result<()> {
+    match alias_refs::rename_with_cascade(config, kind, from, to) {
+        Ok(report) => {
+            for path in &report.dirty_paths {
+                config.mark_dirty(path);
+            }
+            println!(
+                "renamed {sec}.{from} → {sec}.{to} (rewrote {} reference path(s))",
+                report.dirty_paths.len(),
+                sec = section_path(kind)
+            );
+            Ok(())
+        }
+        Err(RenameError::NotFound(p)) => bail!("{p} is not configured"),
+        Err(RenameError::InvalidName(m)) => bail!("invalid new alias: {m}"),
+        Err(RenameError::Reserved(a)) => bail!("alias `{a}` is reserved and cannot be renamed"),
+        Err(RenameError::PostCondition(m)) => bail!("rename cascade post-condition failed: {m}"),
+    }
+}
+
+async fn save(config: &mut Config) -> Result<()> {
+    Box::pin(config.save_dirty())
+        .await
+        .context("failed to persist config")
+}
+
+// ── agents ──────────────────────────────────────────────────────────────────
+
+pub async fn handle_agents(cmd: AgentsCommands, config: &mut Config) -> Result<()> {
+    match cmd {
+        AgentsCommands::List => list_section(config, "agents"),
+        AgentsCommands::Create { alias } => {
+            create_entry(config, "agents", &alias)?;
+            save(config).await
+        }
+        AgentsCommands::Rename { from, to } => {
+            rename_config(config, &AliasKind::Agent, &from, &to)?;
+            warn_agent_owned_state();
+            save(config).await
+        }
+        AgentsCommands::Delete {
+            alias,
+            dry_run,
+            yes,
+        } => {
+            if alias == RESERVED_DEFAULT_AGENT {
+                bail!("the `default` agent is reserved and cannot be deleted");
+            }
+            delete_config(config, &AliasKind::Agent, &alias, dry_run, yes)?;
+            if yes && !dry_run {
+                warn_agent_owned_state();
+                save(config).await?;
+            }
+            Ok(())
+        }
+    }
+}
+
+fn warn_agent_owned_state() {
+    eprintln!(
+        "note: config references were updated, but the agent's owned state \
+         (memory rows, workspace dir, cron/acp/session rows) was NOT cascaded \
+         by this CLI yet — use the gateway API for the full owned-state cascade."
+    );
+}
+
+// ── providers ─────────────────────────────────────────────────────────────────
+
+pub async fn handle_providers(cmd: ProvidersCommands, config: &mut Config) -> Result<()> {
+    match cmd {
+        ProvidersCommands::List { category } => {
+            let cats = match category {
+                Some(c) => vec![parse_provider_category(&c)?],
+                None => vec![
+                    ProviderCategory::Models,
+                    ProviderCategory::Tts,
+                    ProviderCategory::Transcription,
+                ],
+            };
+            for cat in cats {
+                let cat_name = match cat {
+                    ProviderCategory::Models => "models",
+                    ProviderCategory::Tts => "tts",
+                    ProviderCategory::Transcription => "transcription",
+                };
+                // Enumerate families under this category, then their aliases.
+                if let Some(families) = config.get_map_keys(&format!("providers.{cat_name}")) {
+                    let mut families = families;
+                    families.sort();
+                    for family in families {
+                        if let Some(mut aliases) =
+                            config.get_map_keys(&format!("providers.{cat_name}.{family}"))
+                        {
+                            aliases.sort();
+                            for a in aliases {
+                                println!("{cat_name}.{family}.{a}");
+                            }
+                        }
+                    }
+                }
+            }
+            Ok(())
+        }
+        ProvidersCommands::Create {
+            category,
+            family,
+            alias,
+        } => {
+            let cat = parse_provider_category(&category)?;
+            let section = section_path(&AliasKind::Provider {
+                category: cat,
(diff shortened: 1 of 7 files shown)


singlerider (dismissed):
I reviewed slice 8 on its own delta (the four CLI + skill-bundle + i18n commits), the user-facing surface for the whole feature. I read the alias CRUD module, the CLI owned-state cascade, the skill-bundle find/scrub/rewrite, and the PR8a `dirty_paths()` promotion, cross-checking against the slice-4/7 coordinators it reuses. Config tests green (925/925 incl. the moved `dirty_entry_for` + new bundle test), CLI binary builds clean with default features. The reuse discipline is excellent: no duplicated cascade logic, the anti-drift refactor is exactly right. One blocker: the CLI delete path repeats the same workspace-archive ordering bug I flagged on slice 4, even though the CLI rename path right next to it avoids it correctly.

### 🔴 Blocking: CLI agent delete archives the workspace after the config entry is removed, so custom-workspace agents are never archived

In the `AgentsCommands::Delete` arm the order is:

```rust
agent_delete_precheck(config, &alias)?;
apply_delete(config, &AliasKind::Agent, &alias)?;   // delete_with_cascade → cfg.agents.remove(alias)
agent_delete_owned_state(config, &alias).await?;    // resolves workspace internally, entry already gone
```

`apply_delete` runs `delete_with_cascade`, which calls `cfg.agents.remove(alias)`. Then `agent_delete_owned_state` resolves the workspace internally:

```rust
let workspace = config.agent_workspace_dir(alias);
if workspace.exists() { /* archive */ }
```

By that point `config.agents.get(alias)` is `None`, so `agent_workspace_dir` falls through to the default `install_root/agents/<alias>/workspace`. For an agent with a custom `workspace.path`, the `exists()` check probes the wrong path, finds nothing, and the custom workspace is left on disk, unarchived. This is the identical bug to #7838 (slice 4), and it is silent for the same reason.

The fix is the pattern the rename arm in this very file already uses correctly. Compare `AgentsCommands::Rename` directly above:

```rust
// Capture the workspace path while the `from` entry still exists
let old_ws = config.agent_workspace_dir(&from);
rename_config(config, &AliasKind::Agent, &from, &to)?;
agent_rename_owned_state(config, &from, &to, &old_ws).await?;
```

Mirror that in the Delete arm: capture `let old_ws = config.agent_workspace_dir(&alias);` before `apply_delete`, and pass it into `agent_delete_owned_state` (which then archives `old_ws` instead of re-resolving the removed entry). A CLI test with a custom-`workspace.path` agent asserting the archive lands at the custom location would lock it - same coverage gap as slice 4. Since rename already does this right, this is an internal inconsistency, not a missing pattern.

### 🟢 The PR8a anti-drift refactor is exactly the right call

Promoting the gateway's private `delete_cascade_dirty_paths` + `dirty_entry_for` into `zeroclaw-config` as `CascadeReport::dirty_paths()` + public `dirty_entry_for` (symmetric with `RenameReport.dirty_paths`) is the correct consolidation: the gateway delete handler now calls `cascade.dirty_paths()`, the CLI calls the same, and the logic + its test live in one place. The diff cleanly deletes the gateway copies and the call site updates to the method. This is precisely the duplication this stack should be removing as it grows a second caller, not carrying twice.

### 🟢 The CLI reuses the cascade primitives instead of reimplementing them

Delete routes through `delete_with_cascade` + `report.dirty_paths()`, rename through `rename_with_cascade` + `report.dirty_paths`, and both owned-state paths reuse the slice-4/7 `cascade_owned_state` / `cascade_rename_agent` / `live_acp_session_count` coordinators, opening the stores from `data_dir` with no live gateway. The feature-gating (`#[cfg(all(feature="gateway",feature="agent-runtime"))]` with a config-only fallback + warning when off) is the right way to keep the owned-state half optional without a second cascade implementation. Owned-state warnings are surfaced to stderr in both paths. Safe-by-default delete (impact preview unless `--yes`, `--dry-run` to preview) is good CLI hygiene.

### 🟢 CLI `default`-agent delete guard closes a real gap, and the skill-bundle cascade is well-scoped

- The gateway delete refuses only on HARD refs (acp.default_agent / enabled heartbeat), so a `default` agent with neither would be deletable; the explicit CLI refuse (`the default agent is reserved and cannot be deleted`) + test closes that. Rename was already guarded inside `rename_with_cascade`.
- The skill-bundle cascade (`find/scrub/rewrite_bundle_refs`) is a single SOFT container (`agents.<X>.skill_bundles[]`), three mirror fns flattened from the channel arm, deliberately **not** a new `AliasKind` variant - the right call, since a variant would churn every exhaustive match for no gain on a one-container soft ref. Delete archives `shared/skills/<a>/` to `skills/_deleted/<a>-<ts>/` and strips the ref; rename moves the dir + rewrites. `bundle_refs_find_scrub_rewrite` pins all three.
- Routing CLI output through Fluent i18n keeps the new surface consistent with the rest of the CLI.

Fix the delete-arm workspace ordering to match the rename arm and this lands. Nice work tying the whole stack off with a coherent CLI surface.

src/alias_cli/mod.rs
```diff
@@ -0,0 +1,722 @@
...
+                print_impact(&AliasKind::Agent, &alias, config);
+                println!(
+                    "\nNo changes made. Re-run with --yes to apply (or --dry-run to preview)."
+                );
+                return Ok(());
+            }
+            // Owned-state HARD gate (live ACP sessions) runs BEFORE the config
+            // cascade so a refusal mutates nothing.
+            agent_delete_precheck(config, &alias)?;
+            apply_delete(config, &AliasKind::Agent, &alias)?;
```
singlerider, line 417: Same workspace-archive ordering bug as #7838 (slice 4): `apply_delete` runs `delete_with_cascade` → `cfg.agents.remove(alias)`, then `agent_delete_owned_state` resolves the workspace internally with `config.agent_workspace_dir(alias)` - but the entry is already gone, so a custom `workspace.path` agent falls through to the default path, the `exists()` check fails, and the custom workspace is never archived (orphaned silently).

The rename arm directly above already does this right - mirror it here:

```rust
let old_ws = config.agent_workspace_dir(&alias);
agent_delete_precheck(config, &alias)?;
apply_delete(config, &AliasKind::Agent, &alias)?;
agent_delete_owned_state(config, &alias, &old_ws).await?;  // archive old_ws, don't re-resolve
```

A CLI test with a custom-`workspace.path` agent asserting the archive lands at the custom location would lock it.

Audacity88 (changes requested):
@Nillth I reviewed slice 8 at head 870a4c9, focusing on the own delta from slice 7 rather than the cumulative stack. I checked the live PR state, current green GitHub checks, singlerider's current review, the new CLI alias/bundle surfaces, the promoted dirty-path helper, the skill-bundle reference helpers, and the final CLI source after the Fluent follow-up. I did not run local Cargo validation for this review-only pass.

I agree with singlerider's active blocker on the CLI custom-workspace delete ordering and will not duplicate that finding here. Two additional blockers remain in the same head.

### 🟢 What looks good — the CLI surface is additive and the dirty-path helper belongs in config

The plural `agents` / `providers` / `channels` groups are a clean additive surface next to the existing singular `agent` run command, and delete being preview-first unless `--yes` is the right UX. Promoting `CascadeReport::dirty_paths()` into `zeroclaw-config` is also the right source-of-truth move; both gateway and CLI need the same entry-granularity persistence contract.

### 🔴 Blocking — CLI side effects can commit before the config change persists

The new CLI paths perform irreversible filesystem or DB side effects before `save_dirty()` proves the config change reached disk:

- `agents rename` rewrites the in-memory config, then `agent_rename_owned_state` moves the workspace and re-points memory/cron/ACP/session rows, then `save(config)` runs.
- `agents delete` removes the in-memory config entry, then archives/purges owned state, then saves.
- `skills bundle remove` archives the bundle directory before deleting and saving the config entry.
- `skills bundle rename` moves the bundle directory before saving the renamed config and rewritten agent refs.

If `save_dirty()` fails after any of those side effects, the command returns an error with config and non-config state already split. Examples: config can still reference a bundle whose directory has been moved to `_deleted`, or still name an agent whose workspace and owned rows were moved/purged. This is especially risky because these commands are presented as the safe user-facing surface for the cascade.

Please either persist the validated config change before the non-config side effects, or add explicit preflight plus rollback/recovery for every side effect that happens before `save_dirty()`. The current ordering is not safe enough for delete/rename operations that move or purge operator-owned data.

### 🔴 Blocking — some alias CRUD output still bypasses Fluent

The PR adds the Fluent keys and converts many strings, but the final `src/alias_cli/mod.rs` source still has user-facing bare literals in the alias CRUD paths. For example, `apply_delete` still prints `deleted {section}.{alias}...`, `refused: ...`, and hard-ref rows directly, then returns bare `bail!` messages like `delete refused — resolve the hard references first`; `rename_config` still prints `renamed ...` and returns bare `bail!` messages for not found, invalid name, reserved alias, and post-condition failure.

That conflicts with the repository localization rule for CLI messages and with this PR's own claim that the alias CLI output has been routed through Fluent. Please route the remaining user-facing success, refusal, warning, and error text through Fluent as well. Plain alias names and ref paths that are data can stay as data, but the surrounding messages should come from the catalog.

Nillth:
Thanks — both fixed in a3d8ec0.

**1. Side-effects-before-persist.** Reordered all four CLI flows (agents delete/rename, skills bundle remove/rename) to mutate + persist the config *first*, then run the irreversible fs/DB side effects. `agent_delete_owned_state` now takes the workspace path resolved before the cascade removes the entry — which also fixes the CLI's own copy of the slice-4 custom-workspace bug — and the bundle directory is captured before delete/rename and acted on after the save lands. A side-effect failure after persist now surfaces as a warning against a consistent on-disk config, rather than splitting state.

**2. Remaining localization.** `apply_delete` / `rename_config` success + refusal text and the bare `bail!` errors (not-configured, delete-failed, invalid-name, reserved, post-condition, unknown-provider-category, no-such-section, live-ACP, reserved-default, bundle not-configured / rename-failed) now route through Fluent `cli-*` keys. Plain alias names and ref paths stay as data. `cli_fluent_coverage` is green.

Audacity88 (dismissed):
@Nillth thanks for the update.

I re-reviewed current head a3d8ec0 against my prior #7842 findings. I checked the live PR state, your fix comment, the earlier reviews, the final `src/alias_cli/mod.rs` and `src/skills/mod.rs` control flow, the Fluent catalog entries in `crates/zeroclaw-runtime/locales/en/cli.ftl`, and the current green GitHub checks. I did not run local Cargo validation for this review-only pass.

### ✅ Resolved — CLI side effects now run after the config change is durable

This resolves my side-effects-before-persist blocker. Agent delete now captures the workspace path while the config entry still exists, applies the config cascade, calls `save_dirty()`, and only then archives workspace / owned state. Agent rename follows the same shape: capture old workspace, rename config references, persist, then move workspace and re-point owned state.

The skill-bundle remove/rename paths also now persist the config mutation before archiving or moving the bundle directory. A later filesystem failure can still warn, but it no longer leaves the on-disk config pointing at state that was already moved away.

### ✅ Resolved — Main alias CRUD success/refusal/error output now goes through Fluent

The bare user-facing strings I called out in `apply_delete` / `rename_config` have been moved behind `mt` / `mta` keys, with matching `cli-alias-*` entries in the catalog. The new skill-bundle messages also use catalog keys. The remaining printed path lines are the data being reported, not the surrounding user-facing sentence, so they look fine to me.

### 🔴 Blocking — The owned-state-unavailable fallback warning still bypasses Fluent

One user-facing alias CLI string still bypasses the catalog. In `warn_agent_owned_state()`, the no-owned-state fallback prints a bare `eprintln!` note:

```rust
note: config references were updated, but the agent's owned state ...
```

That path is compiled for non-`gateway` / non-`agent-runtime` feature combinations, and the module already has `mt()` specifically so these CLI messages can route through Fluent when `agent-runtime` is present and through an English fallback when it is not. Please route this warning through a `cli-alias-*` key as well, for example `cli-alias-owned-state-unavailable`, so the alias CLI localization contract is complete across feature combinations.

This is the only blocker I still see from my side on this slice. The PR is still draft/stacked, and singlerider's active `CHANGES_REQUESTED` remains in force until they clear it.

Nillth:
Thanks — fixed in e421a6cf5f.

Good catch, that fallback warning was the last bare string. `warn_agent_owned_state` now routes through `cli-alias-owned-state-unavailable` via `mt()`, so the alias-CLI localization is complete across feature combinations — Fluent under `agent-runtime`, English fallback without. `cli_fluent_coverage` is green and the no-default-features path compiles.

singlerider (approved):
RE-REVIEW at head `e421a6cf`. I re-checked my dismissed `CHANGES_REQUESTED` and both of Audacity88's blockers against the current source: `src/alias_cli/mod.rs`, `src/skills/mod.rs`, and the Fluent catalog. All three blockers across the two prior rounds are resolved. CI is 18/18 green, MERGEABLE. This is the draft tail of the 8-PR stack, so it lands only after #3–#7; the verdict below is on the own delta (the CLI + skill-bundle + i18n commits).

### ✅ RESOLVED — CLI agent delete now archives the workspace before the config entry is removed

My blocker is fixed. The `Delete` arm now resolves the workspace while the entry still exists, then persists, then archives:

```rust
let workspace = config.agent_workspace_dir(&alias);
apply_delete(config, &AliasKind::Agent, &alias)?;
save(config).await?;
agent_delete_owned_state(config, &alias, &workspace).await
```

`agent_delete_owned_state` now takes `workspace: &Path` from the caller instead of re-resolving the removed entry, so a custom `workspace.path` is preserved through the archive instead of falling through to the default path and being silently skipped. This mirrors the rename arm exactly, closing the internal inconsistency. The same fix lands the CLI's copy of the slice-4 (#7838) bug.

### ✅ RESOLVED — CLI side effects now run after the config change is durable

Audacity88's side-effects-before-persist blocker is fixed across all four flows. Each now mutates + `save_dirty()` first, then runs the irreversible fs/DB side effects: agent delete (config cascade → save → archive/purge owned state), agent rename (capture `old_ws` → rename config → save → move workspace + re-point rows), `skills bundle remove` (scrub refs → save → archive dir), `skills bundle rename` (rewrite refs → save → move dir). A post-persist side-effect failure now warns against a consistent on-disk config instead of splitting state.

### ✅ RESOLVED — owned-state-unavailable fallback now routes through Fluent

Audacity88's final blocker is fixed. `warn_agent_owned_state()` now resolves `cli-alias-owned-state-unavailable` via `mt()` (Fluent under `agent-runtime`, English fallback without), and the key is present in `locales/en/cli.ftl:786`. `cli_fluent_coverage` is green. The remaining bare `println!` lines in the module print pure data (alias names, ref paths, bundle paths), which the localization rule permits.

### 🟢 The reuse discipline and anti-drift refactor hold up under re-read

The CLI still routes delete through `delete_with_cascade` + `report.dirty_paths()` and rename through `rename_with_cascade`, reusing the slice-4/7 `cascade_owned_state` / `cascade_rename_agent` coordinators opened from `data_dir` with no live gateway. The PR8a promotion of `CascadeReport::dirty_paths()` into `zeroclaw-config` remains the right source-of-truth consolidation. Skill-bundle `find/scrub/rewrite_bundle_refs` stays a single flattened SOFT container rather than a new `AliasKind` variant — correct for a one-container soft ref. Feature-gating (`all(gateway, agent-runtime)` with a config-only fallback + warning) is intact.

All blockers cleared. Approving on the own delta; the stack-ordering merge discipline in the PR body governs when it actually lands.

WareWolf-MoonWall (approved):
I read the diff at head e421a6cf on the own delta (the alias CRUD CLI verbs, agent-owned-state cascade, and skill-bundle cascade commits), cross-checked the SSOT and cascade-correctness properties, and reviewed the prior review history. singlerider APPROVED at this head after their earlier CHANGES_REQUESTED was addressed. Audacity88's CHANGES_REQUESTED (bare Fluent literals in `apply_delete`/`rename_config`) was resolved and they dismissed their review before singlerider's approval.

### ✅ Resolved — Fluent localization in alias CRUD paths

Audacity88's 🔴 blocker (user-facing bare literals in `apply_delete` and `rename_config` — success/refusal/error text bypassing Fluent) is addressed at a3d8ec0e. The final `src/alias_cli/mod.rs` routes all user-facing output through `cli-*` Fluent keys and `cli_fluent_coverage` passes. Audacity88 confirmed and dismissed their review.

### 🟢 What looks good — cascade write path is single-source-of-truth

`delete_with_cascade` and the alias-rename path both write through the one canonical config mutator — no second cache or parallel state store. The `CascadeReport` surfaces impact before mutation and the `--yes` gate ensures safe-by-default deletion. The `default`-agent guard (refuses delete when the alias is a hard ref in `acp.default_agent` or enabled heartbeat) is well-scoped and test-covered.

### 🟢 What looks good — skill-bundle cascade is correct

The skill-bundle find/scrub/rewrite correctly identifies and removes bundle references from all agent configs when the owning bundle or agent is deleted, without touching unrelated agents. No orphaned bundle refs can survive a delete. The `dirty_paths()` promotion from slice 7 is clean.

### 🟡 Warning — branch is CONFLICTING (expected for stacked PR — merge strictly in order)

This is slice 8 of 8 and the branch shows `mergeable: CONFLICTING`. This is expected: slices #3–#7 (#7837–#7841) need to land first, at which point the cumulative diff collapses to this PR's own delta. **Do not merge this PR before its predecessors are squash-merged in sequence.** The content of this slice is clean; the conflict is structural, not a code problem.

### 🔵 Suggestion — i18n placeholder coverage for non-English locales

The new `crates/zeroclaw-runtime/locales/en/cli.ftl` entries (+46 lines) are English-only. The prior Fluent pattern for this codebase requires placeholder entries in es/fr/ja/zh-CN even when translation is pending. If those aren't present (they weren't visible in this diff), they should be added before merge.

### Verdict

Content is technically sound. Both prior reviewer CRs resolved. Approving the content with the note that this PR must not land until slices #3–#7 precede it, and the non-English locale placeholder gap should be confirmed before squash-merge.

PR #7843, opened 2026-06-17
fix(telegram): bypass mention_only gate for replies to bot messages

### LINKED ISSUE #5866: [Bug]: in the Telegram group bot ignores replies on its messages when mention_only=true

mazhuima:
## Summary

When `mention_only=true` in a Telegram group, users replying directly to the bot's messages were silently ignored because the reply text typically does not contain an `@mention`. This PR detects `reply_to_message.from.id` and bypasses the mention check when it matches the bot's own user ID.

Fixes #5866

## Changes

- Added `bot_id: Mutex<Option<i64>>` field for caching the bot's numeric user ID
- Enhanced `fetch_bot_username` to also extract and cache the bot ID from `getMe` API response
- Added `is_reply_to_bot()` helper that checks `reply_to_message.from.id == bot_id`
- Updated `parse_update_message()` to bypass mention check for replies to bot
- Updated `check_media_mention_gate()` to bypass mention check for replies to bot

## Test plan

- [ ] CI passes (cargo check, cargo test)
- [ ] Manual verification: Telegram group with `mention_only=true`, reply to bot message without @mention → bot should respond

## Affected components

- `zeroclaw-channels` — Telegram channel

### DIFF
--- a/crates/zeroclaw-channels/src/telegram.rs
+++ b/crates/zeroclaw-channels/src/telegram.rs
@@ -536,6 +536,7 @@ pub struct TelegramChannel {
     last_draft_edit: Mutex<std::collections::HashMap<String, std::time::Instant>>,
     mention_only: bool,
     bot_username: Mutex<Option<String>>,
+    bot_id: Mutex<Option<i64>>,
     /// Base URL for the Telegram Bot API. Defaults to `https://api.telegram.org`.
     /// Override for local Bot API servers or testing.
     api_base: String,
@@ -615,6 +616,7 @@ impl TelegramChannel {
             typing_handle: Mutex::new(None),
             mention_only,
             bot_username: Mutex::new(None),
+            bot_id: Mutex::new(None),
             api_base: TELEGRAM_OFFICIAL_API_BASE_URL.to_string(),
             transcription: None,
             transcription_manager: None,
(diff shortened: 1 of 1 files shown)


Audacity88:
## Triage Summary

#5866 already has an active implementation path in #7723. That PR has the narrowed behavior reviewers asked for; the remaining blocker is regression coverage for reply-to-this-bot, reply-to-another-bot, and normal unmentioned group messages.

This PR targets the same bug but does not say whether it supersedes #7723, and it does not add that coverage. Next step: either fold this into #7723, or update this PR to state why it should replace #7723 and include the same three tests.

Until then, #7723 stays the current canonical review path for #5866.

mazhuima:
Thanks for the triage @Audacity88. I agree this is a duplicate of #7723 which already has the narrowed reviewer-requested behavior. Closing in favor of #7723.

One note: #7723 uses `bot_username` to identify replies to self. An alternative worth considering is using the bot numeric `id` from `getMe` — numeric IDs never change whereas usernames can. But this is a minor point and #7723's approach is fine as-is.

PR #7844, opened 2026-06-17
feat(channels/discord): typed slash-command options + chunked interaction followups

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **Typed slash-command options.** A `slash`-tagged skill can now declare typed command options in its `SKILL.toml` under `[[skill.slash_options]]` — `string`/`integer`/`number`/`boolean`/`user`/`channel`/`role`/`mentionable`, with `choices`, `min`/`max`, and `min_length`/`max_length`. These register as real Discord application-command options, and the values the user submits are folded into the synthesized agent prompt (`name: value` lines). Previously every skill command had exactly one free-text `input` option; this lifts it to the full typed model OpenClaw/Hermes expose.
  - **Backward-compatible.** A skill that declares no options keeps the legacy single required `input` string (which also remains the ownership marker used for stale-command reaping). The manifest field defaults empty, so existing skills are unchanged.
  - **Input is validated, never trusted to Discord.** An authoring mistake in a manifest (an option name with spaces/punctuation, a non-numeric choice value on an integer option, inverted `min`/`max`, or duplicate names) would otherwise produce a registration body Discord rejects with `400`, and the reconcile would retry it on every READY — the same re-registration burst #7784 fixed. `map_skill_slash_options` slugs/de-dups names and drops bad choices/bounds with a WARN, so nothing invalid reaches Discord.
  - **Chunked interaction followups.** The deferred-reply path used to truncate any answer over Discord's 2000-char limit with a WARN (a flagged TODO). It now splits the answer with the shared chunker: the first chunk edits the `@original` deferred message and the rest are delivered as followup webhook POSTs, so long `/ask` answers arrive in full.
- **Scope boundary:** Flat typed options only — sub-commands/groups are a noted follow-on. **Not** in this PR: name/description localizations, guild-scoped deployment (both planned). User/channel/role/mentionable option values currently fold into the prompt as the raw snowflake id (resolving to a display name via `data.resolved` is a follow-on). No config/env/CLI flags added.
- **Blast radius:** The Discord channel's slash registration + reconcile projection + interaction dispatch + the followup path, plus a new optional `slash_options` field on the runtime `Skill`/`SkillMeta` manifest (additive, default empty; consumed only by the Discord channel).
- **Linked issue(s):** Part of #7831. (Builds on the modular refactor #7832, now merged.)
- **Design note for maintainers:** the `[[skill.slash_options]]` table is a **new user-facing `SKILL.toml` surface**. The Discord-side model + chunked followups are ordinary channel work in the lineage of #7489/#7490, but I'd value a sounding on the manifest schema shape (field names, where typed options live) before it sets precedent — happy to adjust.
- **Labels:** suggest `type: feature`, `risk: medium`, `size: M`, `channel:discord`, `skills` (maintainers apply).

```bash
cargo fmt --all -- --check
cargo clippy -p zeroclaw-channels --features channel-discord --all-targets -- -D warnings
cargo test -p zeroclaw-channels --features channel-discord
cargo test -p zeroclaw-runtime slash_options
```

  - `cargo fmt --all -- --check` → clean (no diff)
  - `cargo clippy … --all-targets -- -D warnings` → `Finished dev profile …`; zero warnings
  - `cargo test -p zeroclaw-channels --features channel-discord` → `test result: ok. 950 passed; 0 failed; 1 ignored` (lib), integration bins `3 / 3 / 4 / 0 passed`, `0 failed`
  - `cargo test -p zeroclaw-runtime slash_options` → manifest-parse tests `ok. 2 passed; 0 failed`
- **Beyond CI — what did you manually verify?** Reconcile change-detection stability: the projection of a typed command normalises numeric bounds (int-vs-float) and choice values so a desired-vs-Discord echo difference doesn't force a spurious re-registration (a regression test pins int `1` vs float `1.0`). The registration-input validation (slug/dedup names, drop bad choices, drop inverted bounds) is test-pinned. An all-optional typed command invoked with no arguments still invokes the skill. An adversarial review pass confirmed the projection logic sound and surfaced the two issues above, both fixed.

- New external network calls? **No** — the same Discord REST surface; followups use the existing interaction webhook endpoint.
- Secrets / tokens / credentials handling changed? **No** — the interaction followup token handling is unchanged.
- If any `Yes`, describe the risk and mitigation: Slash-command option values are user-supplied and folded into the synthesized agent prompt — the same trust level as the existing `/ask` and skill `input` free text (already user-supplied since #7489/#7490); no new untrusted-input path.

## Compatibility (required)

- Backward compatible? **Yes** — skills with no declared options keep the legacy single `input`; the manifest field defaults empty; chunked followups only change behaviour for replies that previously got truncated past 2000 chars.
- Config / env / CLI surface changed? **Manifest only** — `SKILL.toml` gains an optional `[[skill.slash_options]]` table. Additive and default-none; existing manifests are unaffected. No config/env/CLI flags.
- If `No` or `Yes` to either: exact upgrade steps for existing users: None required; authors opt in by adding `[[skill.slash_options]]` to a `slash`-tagged skill.

## Rollback

Low-risk: `git revert <sha>` is the plan. Typed options are opt-in per skill and the chunked-followup change is confined to the deferred-reply path; reverting restores the single-`input` command shape and the prior truncation with no data or config implications.

### DIFF
--- a/crates/zeroclaw-channels/src/discord/interaction.rs
+++ b/crates/zeroclaw-channels/src/discord/interaction.rs
@@ -6,7 +6,7 @@
 
 use serde_json::json;
 
-use super::types::{DISCORD_MAX_MESSAGE_LENGTH, DiscordOutgoing};
+use super::types::DiscordOutgoing;
 
 /// Credentials needed to answer a deferred interaction later: the followup
 /// webhook is addressed by application id + interaction token.
@@ -100,34 +100,24 @@ pub(crate) async fn discord_reject_interaction(
     Ok(())
 }
 
-/// Deliver the agent's answer by editing the deferred interaction response
-/// (`PATCH /webhooks/{app_id}/{token}/messages/@original`). The token is valid
-/// for 15 minutes; no bot auth header is required for the followup webhook.
+/// Edit the deferred interaction's @original message
+/// (`PATCH {api_base}/webhooks/{app_id}/{token}/messages/@original`). The token
+/// is valid for 15 minutes; no bot auth header is required for the followup
+/// webhook. `content` must be within Discord's 2000-char limit — callers whose
+/// reply may exceed it chunk first and post the remainder via
+/// [`discord_post_interaction_followup`].
 pub(crate) async fn discord_edit_interaction_response(
     client: &reqwest::Client,
     app_id: &str,
     interaction_token: &str,
+    api_base: &str,
     content: &str,
 ) -> anyhow::Result<()> {
-    let url = format!(
-        "https://discord.com/api/v10/webhooks/{app_id}/{interaction_token}/messages/@original"
-    );
-    let trimmed: String = content.chars().take(DISCORD_MAX_MESSAGE_LENGTH).collect();
-    if trimmed.chars().count() < content.chars().count() {
-        ::zeroclaw_log::record!(
-            WARN,
-            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
-                .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
-                .with_attrs(::serde_json::json!({
-                    "content_chars": content.chars().count(),
-                })),
-            "interaction reply truncated to Discord's 2000-char limit (chunked followups are a planned follow-up)"
-        );
-    }
+    let url = format!("{api_base}/webhooks/{app_id}/{interaction_token}/messages/@original");
     // without_url: transport errors embed the token-bearing URL.
     let resp = client
         .patch(&url)
-        .json(&DiscordOutgoing::text(trimmed).to_rest_json())
+        .json(&DiscordOutgoing::text(content).to_rest_json())
         .send()
         .await
         .map_err(reqwest::Error::without_url)?;
@@ -135,6 +125,31 @@ pub(crate) async fn discord_edit_interaction_response(
         let status = resp.status();
         let err = resp.text().await.unwrap_or_default();
         anyhow::bail!("interaction followup edit failed ({status}): {err}");
+    }
+    Ok(())
+}
+
+/// Post an additional interaction followup message
+/// (`POST {api_base}/webhooks/{app_id}/{token}`), used to deliver the answer
+/// chunks beyond the first when a reply exceeds Discord's 2000-char limit.
+pub(crate) async fn discord_post_interaction_followup(
+    client: &reqwest::Client,
+    app_id: &str,
+    interaction_token: &str,
+    api_base: &str,
+    content: &str,
+) -> anyhow::Result<()> {
+    let url = format!("{api_base}/webhooks/{app_id}/{interaction_token}");
+    let resp = client
+        .post(&url)
+        .json(&DiscordOutgoing::text(content).to_rest_json())
+        .send()
+        .await
+        .map_err(reqwest::Error::without_url)?;
+    if !resp.status().is_success() {
+        let status = resp.status();
+        let err = resp.text().await.unwrap_or_default();
+        anyhow::bail!("interaction followup post failed ({status}): {err}");
     }
     Ok(())
 }
--- a/crates/zeroclaw-channels/src/discord/mod.rs
+++ b/crates/zeroclaw-channels/src/discord/mod.rs
@@ -2095,34 +2095,8 @@ impl Channel for DiscordChannel {
                                                 None => Vec::new(),
                                             };
                                             match specs.into_iter().find(|spec| spec.slug == command) {
-                                                Some(spec) if spec.options.is_empty() => {
-                                                    // Legacy single free-text `input` option.
-                                                    if input.is_empty() {
-                                                        None // known command, empty input
-                                                    } else {
-                                                        Some(format!(
-                                                            "Use the '{}' skill for this request: {input}",
-                                                            spec.skill_name
-                                                        ))
-                                                    }
-                                                }
                                                 Some(spec) => {
-                                                    // Typed options: fold the submitted name/value pairs
-                                                    // (extracted above) into the prompt (Discord enforces
-                                                    // the required ones).
-                                                    if submitted.is_empty() {
-                                                        None
-                                                    } else {
-                                                        let rendered = submitted
-                                                            .iter()
-                                                            .map(|(n, v)| format!("{n}: {v}"))
-                                                            .collect::<Vec<_>>()
-                                                            .join("\n");
-                                                        Some(format!(
-                                                            "Use the '{}' skill for this request:\n{rendered}",
-                                                            spec.skill_name
-                                                        ))
-                                                    }
+                                                    skill_command_prompt(&spec, &input, &submitted)
                                                 }
                                                 None => None, // stale or foreign command
                                             }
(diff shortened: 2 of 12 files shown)


singlerider (approved):
First review (no prior reviews, comments, or inline threads). Author @Nillth. Part of #7831, builds on the merged modular refactor #7832. I read all 12 files of the own delta — the new `slash_options` contract module, the registration/projection/reconcile changes in `slash.rs`, the chunked-followup path, the dispatch extraction in `mod.rs`, and the runtime `SkillSlashOption` manifest type — and cross-checked the trust surface, the module-boundary discipline, and the reconcile change-detection. Built `zeroclaw-channels --features channel-discord` + `zeroclaw-runtime`, ran the slash-option / typed-option / chunk tests locally (all green: 7 + 15 + manifest-parse 2), clippy clean with `-D warnings`. CI 18 checks green, MERGEABLE. This is clean, well-tested work. All green.

### 🟢 The typed-option model is sound and the input-validation discipline is exactly right

`map_skill_slash_options` sanitizes every authoring mistake before it reaches Discord rather than passing it through to a `400`: an unknown `type` drops the option; the name is slugged to Discord's charset and de-duped within the command; numeric choices whose value doesn't parse to the option type are dropped; inverted `min`/`max` bounds are dropped with a WARN; required options are sorted first (Discord rejects a required option after an optional one). This directly avoids re-creating the re-registration burst #7784 fixed — an invalid registration body would 400 and `reconcile_slash_commands` would retry it on every READY. Each rule is test-pinned (`map_drops_unknown_kinds_and_sorts_required_first`, `map_slugs_option_names_and_dedups`, `map_drops_a_nonparsing_numeric_choice`, `map_drops_inverted_bounds`). The `OptKind::is_scalar` gate keeps choices/bounds off user/channel/role/mentionable kinds where Discord rejects them.

### 🟢 The reconcile projection is stabilized against the int-vs-float echo difference

`command_projection` now normalises choice values (numbers → `f64` via `normalize_scalar`) and min/max bounds (`as_f64`) so a desired command carrying integer bounds compares equal to Discord's echo of the same command with float bounds. `projection_is_stable_across_int_vs_float_bounds` pins it (sent `min_value: 1` vs echoed `1.0`, plus server-side `id`/`version` decorations, project equal). Without this a typed command would re-register on every READY — the same spurious-churn class the projection exists to prevent. Good instinct to test it explicitly.

### 🟢 The chunked-followup path correctly replaces the silent truncation

`deliver_interaction_answer` splits with the shared `split_message_for_discord`, edits the deferred `@original` with the first chunk, and posts the remainder via `discord_post_interaction_followup` — so a long `/ask` answer arrives in full instead of being cut at 2000 chars with a WARN (a flagged TODO, now retired). The chunking lives in the wiring layer (`mod.rs`) so `interaction.rs` need not depend on `chunk`, preserving the no-impl-to-impl boundary the new `slash_options` module also respects (contract-tier, imports no sibling impl). `interaction_answer_over_2000_chars_chunks_into_edit_plus_followup` (PATCH×1 + POST×1) and `short_interaction_answer_only_edits_original` (POST×0) lock both arms via wiremock.

### 🟢 Backward compatibility and the trust surface are both correct

A skill with no declared options keeps the legacy single required `input` string (still the ownership marker for reaping); the manifest field is `#[serde(default)]` empty, so existing manifests are untouched, and `no_options_falls_back_to_the_legacy_input` proves it. On trust: option `value`s fold into the synthesized prompt as `name: value` lines, the same trust level as the existing free-text `input` (user-supplied since #7489/#7490) — no new untrusted-input path. The option `name` is slugged (no newline/injection vector) and `skill_name` is sanitized of `\n`/`\r`/`'` at spec-build, so neither the name nor the skill label can break the prompt framing. The PR body discloses this accurately.

### 🟢 Documented limitations are the right calls

- **Typed-command reaping:** `is_skill_command_shape` recognises only the legacy single-`input` shape, so a typed command is not auto-reaped when its skill is uninstalled (it is still upserted while installed). This is conservative — it never risks deleting a foreign command — so the failure mode is a stale-but-harmless registered command, not data loss or a re-registration loop. Documented inline with a follow-on plan. Correct to defer.
- **Snowflake resolution:** user/channel/role/mentionable values fold as the raw snowflake id rather than a `data.resolved` display name. Documented as a follow-on; acceptable for a first cut since the agent receives an unambiguous id.

### 🟢 One design sounding on the manifest schema, since you asked

The `[[skill.slash_options]]` shape reads well as a new `SKILL.toml` surface, and modelling it channel-agnostically on the runtime `Skill` (with the Discord channel owning the wire mapping) is the right layering — a future slash-capable channel maps `kind` to its own wire type without touching the manifest. Two small things worth a thought before it sets precedent, neither blocking:

1. `kind` is a free-text `String` validated late (the channel drops unknown values with a WARN). That's a deliberate channel-agnostic choice and it fails safe, but it means an authoring typo (`"interger"`) silently yields no option with only a runtime WARN. If you want earlier feedback you could surface unknown kinds at manifest-load as a soft diagnostic; I would not gate on this.
2. `min`/`max` are `f64` even for integer options (coerced at registration). Fine on the wire, but a manifest author writing `min = 1` for an integer option gets a float in the typed model. Harmless given the coercion; flagging only so the `f64`-for-int choice is intentional and not an oversight.

Nothing here blocks. Approving. Nice, self-contained slice with validation discipline that respects the prior re-registration lessons.

WareWolf-MoonWall (approved):
I read the full diff at head 92143f9e, the PR body, singlerider's approval, the validation evidence, and the slash_options.rs module. This is my first review; singlerider reviewed thoroughly and found no blockers. I cross-checked the key correctness and SSOT properties independently.

### 🟢 What looks good — typed option validation is safe-by-default

`map_skill_slash_options` in `slash_options.rs` slugs and de-dups names, drops invalid choice values on typed options, and drops inverted min/max bounds with a WARN before anything reaches Discord's API. This is the right defense: a manifest authoring mistake produces a logged warning and a safe degraded registration rather than a 400 loop that would re-register on every READY (the same re-registration storm #7784 addressed). The test suite pins both the slug/de-dup behaviour and the projection's numeric normalisation (`int 1` vs `float 1.0` equality) so a future refactor can't quietly regress the guard.

### 🟢 What looks good — SSOT: options are resolved from SKILL.toml at manifest load, not cached

Typed options come from the skill manifest (`[[skill.slash_options]]`), parsed in `slash_options.rs` and consumed by `discord_slash_specs_from_skills`. No second store or runtime struct field caches them — the canonical source is the manifest file and the skills service resolves it on demand. AGENTS.md SSOT rule is satisfied.

### 🟢 What looks good — backward compatibility is exactly right

A skill that declares no options gets the legacy single required `input` string unchanged. The `ownership marker` role of the `input` option is preserved so stale-command reaping continues to work. Existing Discord registrations are unaffected on first reload since the projection equality check prevents spurious re-registrations for unchanged commands.

### 🟢 What looks good — chunked followup handles the 2000-char Discord limit correctly

The deferred-reply chunked path loops through `send_message_chunk` with the existing interaction token, falling back to channel followup only on token expiry. Edge cases (empty final chunk, single-chunk response that fits) are covered by tests. This is a well-scoped fix for a real truncation bug.

### 🔵 Suggestion — `SKILL_COMMAND_OPTION_DESCRIPTION` is an English-only constant shown to Discord users

The constant (`"What to send to the skill command"`) is passed directly to the Discord API as the option description string visible in the slash-command UI. This follows the existing project pattern for Discord-API-facing strings (which aren't routed through Fluent). Worth a brief comment noting it's an API contract string and English-only by deliberate choice, so a future contributor doesn't flag it as a localization gap.

### Verdict

Template is complete, validation evidence is solid (fmt/clippy/test with literal output), security analysis is thorough, and the content is correct. No blockers. Approving.

PR #7667, opened 2026-06-15
refactor(api): make before_llm_call take mutable borrows

fanchanghu:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Changed `HookHandler::before_llm_call` to accept `&mut Vec<ChatMessage>` and `&mut String` instead of taking them by value. This avoids cloning the entire provider-visible transcript on every LLM request, which matters because `Vec<ChatMessage>` can be large (long tool histories, image/data URLs).
  - Updated `HookRunner::run_before_llm_call` in zeroclaw-runtime to match the new trait signature.
- **Scope boundary:** This PR only changes the hook API and dispatcher; it does not add any call sites. Hook invocation at LLM call sites will be added in a follow-up PR. No other modifying hooks are changed.
- **Blast radius:** Very low. `before_llm_call` currently has no callers in the repository. The only consumers affected are future plugin implementations that override `before_llm_call`; they will need to use the new mutable-borrow signature.
- **Linked issue(s):** None
- **Labels:** `type: refactor`, `risk: low`, `size: S`

## Validation Evidence (required)

```bash
cargo fmt --all -- --check
cargo clippy --all-targets -- -D warnings
cargo test -p zeroclaw-api -p zeroclaw-runtime --lib
cargo test -p zeroclaw-runtime --tests
```

- **Commands run and tail output:**
  - `cargo fmt --all -- --check`: passed (no output)
  - `cargo clippy --all-targets -- -D warnings`: passed
  - `cargo test -p zeroclaw-api -p zeroclaw-runtime --lib`: `test result: ok. 2163 passed; 0 failed; 1 ignored`
  - `cargo test -p zeroclaw-runtime --tests`: `test result: ok. 2163 passed; 0 failed; 1 ignored` plus `scheduled_run_does_not_leak_conversation_memory_into_provider_request ... ok`
- **Beyond CI — what did you manually verify?** Confirmed the new signature eliminates the `Vec<ChatMessage>` clone in `HookRunner::run_before_llm_call`. Verified the previously failing CI test `turn_cache_hit_emits_agent_end_with_none_tokens` passes after rebasing onto latest `zero/master`.
- **If any command was intentionally skipped, why:** Full `cargo test` was not run locally because the repository’s full test suite is large and CI already covers it; targeted tests for the affected crates were run instead.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? (`No`)
- New external network calls? (`No`)
[...]

Issue #7822, opened 2026-06-17
RFC: WASM plugin lifecycle observer subscriptions (PluginCapability::Observer)


dakaii:
> **Revision history**
>
> **2026-08-24 - Revision 2 (maintainer takeover; observer capability clarification).** Folds in the 2026-08-03 discussion feedback. Replaces the proposed new `PluginCapability::Hook` with the existing reserved `PluginCapability::Observer`, pins the host-owned completion source, removes the Phase 1 textual-summary option, and defines per-plugin timeout, backpressure, failure, and permission-composition rules. This is a material vote-contract revision; discussion runs through 2026-08-26 05:00 UTC before a default two-thirds Core vote may open if the body remains stable.

## Problem

ZeroClaw supports built-in Rust hooks and several WASM plugin capabilities, but a third-party WASM component cannot subscribe to agent lifecycle events through the plugin ABI.

Operators therefore need core patches or indirect telemetry when they want a plugin to perform post-turn auditing, metrics, or integration work. A lifecycle-listener capability belongs in the plugin model only if it preserves the WIT compatibility boundary, does not expose sensitive turn content by default, and cannot change or fail the completed turn.

### Add a WIT-defined observer lifecycle-listener capability

ZeroClaw may implement the existing reserved `PluginCapability::Observer` as the WASM lifecycle-listener surface.

This RFC does not add a parallel `PluginCapability::Hook`. Built-in Rust hooks remain an internal runtime surface, and the public plugin compatibility contract uses the observer capability name that already appears in plugin manifests and documentation.

The guest contract must be a versioned WIT world under `wit/v0`, with matching host bridge code in `crates/zeroclaw-plugins`. It must not expose the Rust `HookHandler` trait directly or use an ad hoc `on_hook(event, json)` ABI.

The plugin manifest must declare the observer capability and an explicit allowlist of subscribed event names. Unknown event names and capability/world mismatches are configuration errors. Declaring the observer capability does not grant filesystem, network, memory, config, or other host permissions.

### Phase 1 is observe-only turn completion

The first event is `on_turn_complete`.

The host supplies a typed, versioned summary after the turn outcome has been established. The payload may include bounded attribution and operational metadata such as a correlation identifier, agent alias, channel, typed outcome, duration, token usage, and capped tool names or counts.

Raw prompts, model responses, tool arguments, tool results, credentials, memory contents, unrestricted session state, and free-form textual summaries are not part of the Phase 1 payload.

[...]

PR #7846, opened 2026-06-17
feat(runtime): wire before_llm_call hook into LLM call paths

fanchanghu:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Wired the `before_llm_call` modifying hook into `run_tool_call_loop` in `crates/zeroclaw-runtime/src/agent/turn/mod.rs`.
  - The hook runs **before** the framework's vision-provider routing and per-iteration message preparation. This gives user-defined hooks priority over runtime fallbacks (e.g. image-marker stripping when no vision provider is configured), allowing plugins to rewrite messages, swap models, or cancel the call based on the original request state.
  - Cancellations abort the current tool-loop iteration via `ToolLoopCancelled`.
- **Scope boundary:** This PR only adds the call site for `before_llm_call`; it does not change the hook trait signature (already merged in #7667). No other hooks or agent entry points are modified.
- **Blast radius:** High by label classification because this touches runtime/agent LLM-call flow. The hook only fires when a `HookRunner` is attached (currently only embedded `Agent` paths pass one). When no hooks are registered, behavior is unchanged.
- **Linked issue(s):** Depends on #7667 (the mutable-borrow signature change for `before_llm_call`)
- **Labels:** `enhancement`, `size: XS`, `risk: high`, `agent`, `runtime`

```bash
cargo fmt --all -- --check
cargo check -p zeroclaw-runtime --all-targets
cargo test -p zeroclaw-runtime --lib hooks::runner::tests
cargo test -p zeroclaw-runtime --lib "agent::turn::vision_route::tests::prepare_messages_for_iteration_populates_and_reuses_image_cache" -- --exact
```

  - `cargo fmt --all -- --check`: passed (no output after formatting)
  - `cargo clippy --all-targets -- -D warnings`: passed
  - `cargo check -p zeroclaw-runtime --all-targets`: passed
  - `cargo test -p zeroclaw-runtime --lib hooks::runner::tests`: `test result: ok. 4 passed; 0 failed`
  - `cargo test -p zeroclaw-runtime --lib "agent::turn::vision_route::tests::prepare_messages_for_iteration_populates_and_reuses_image_cache" -- --exact`: `test result: ok. 1 passed; 0 failed`
- **Beyond CI — what did you manually verify?** Confirmed the hook placement is before `resolve_vision_provider` and `prepare_messages_for_iteration`, so hook modifications take priority over framework vision fallback behavior.
- **If any command was intentionally skipped, why:** Full `cargo test` was not run locally because the repository’s full test suite is large and CI already covers it; targeted tests for the affected crate were run instead.

## Compatibility (required)

  - The hook is opt-in; callers without a `HookRunner` see no behavior change.

- Revert this PR (`git revert <merge-sha>`) to remove the `before_llm_call` invocation from the LLM turn loop.
- If a production regression needs immediate mitigation before revert lands, disable the affected `HookRunner` attachment path so `before_llm_call` is not invoked for embedded Agent calls.
- Re-run the targeted runtime hook/vision-provider tests and relevant CI after rollback.

### DIFF
--- a/crates/zeroclaw-runtime/src/agent/turn/mod.rs
+++ b/crates/zeroclaw-runtime/src/agent/turn/mod.rs
@@ -296,6 +296,26 @@ pub async fn run_tool_call_loop(
             .into());
         }
 
+        // Modifying hook before LLM call (may rewrite messages/model or cancel).
+        // Runs before the framework's vision-provider routing and message
+        // preparation so user-defined strategies take priority over runtime
+        // fallbacks such as image-marker stripping.
+        let mut active_model = model.to_string();
+        if let Some(hooks) = hooks {
+            match hooks.run_before_llm_call(history, &mut active_model).await {
+                crate::hooks::HookResult::Continue(()) => {}
+                crate::hooks::HookResult::Cancel(reason) => {
+                    ::zeroclaw_log::record!(
+                        INFO,
+                        ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                            .with_attrs(::serde_json::json!({"reason": reason.to_string()})),
+                        "llm call cancelled by hook"
+                    );
+                    return Err(crate::agent::loop_::ToolLoopCancelled.into());
+                }
+            }
+        }
+
         let iteration_tool_specs = build_iteration_tool_specs(
             model_provider,
             tools_registry,
@@ -320,10 +340,13 @@ pub async fn run_tool_call_loop(
                 .vision_model_provider
                 .as_deref()
                 .unwrap_or(provider_name);
-            let vm = multimodal_config.vision_model.as_deref().unwrap_or(model);
+            let vm = multimodal_config
+                .vision_model
+                .as_deref()
+                .unwrap_or(&active_model);
             (vp_box.as_ref(), vp_name, vm)
         } else {
-            (model_provider, provider_name, model)
+            (model_provider, provider_name, &active_model)
         };
 
         let prepared_messages = prepare_messages_for_iteration(


Audacity88 (changes requested):
Reviewed `9f7e5bff8236b46c6f4837a9efbb697b3e2d1828` against the live PR state, green check snapshot, empty review/comment threads, dependency PR #7667, accepted hook RFC #7822, the one-file diff, and the adjacent embedded `Agent::turn` / `turn_streamed` replay paths. I did not rerun Cargo locally.

### 🟢 What looks good — The hook is placed before provider routing fallbacks

The intended call-site placement is directionally right. Running `before_llm_call` before vision-provider routing and image fallback preparation gives hooks a chance to inspect or rewrite the original request state before runtime fallbacks strip or transform multimodal markers. Keeping cancellation fail-closed through `ToolLoopCancelled` also fits the existing turn-loop cancellation shape.

### 🔴 Blocking — Hook-mutated messages bypass the Agent replay log

The new call passes the loop's canonical `history` directly into `hooks.run_before_llm_call(history, &mut active_model)`. That lets a hook append, delete, or rewrite messages that the provider will see. But the embedded `Agent` wrappers do not persist `loop_history` directly. They pass `Some(&mut loop_new_messages)` / `Some(&mut round_added)` into `run_tool_call_loop`, then replay only that append log into `self.history` and returned `new_messages` after the loop.

That means a hook can change the provider-visible transcript without the stored conversation recording the same change. For example, a hook that rewrites the latest user message or injects a policy/user instruction would affect the model response, but `Agent::turn` would still store the original conversation plus the assistant response. The streamed path has the same shape for returned `new_messages`. This breaks the turn-loop append-log contract and makes later replay/debugging see a different transcript than the provider actually answered.

Please make the hook semantics explicit before merging: if `before_llm_call` changes are meant to be ephemeral provider-request transforms, run the hook on a per-request message copy after the durable transcript boundary instead of mutating `history`; if they are meant to be durable transcript edits, route them through the same append-log/replay path that `Agent::turn` and `turn_streamed` persist. Either way, add a loop/Agent-level regression test that proves a `before_llm_call` message rewrite reaches the provider and that the stored/returned conversation state matches the intended semantics.

### 🟡 Warning — PR body risk metadata is stale

The PR body label snapshot says `type: feat`, `risk: low`, `size: S`, `runtime`, but the current live labels are `enhancement`, `size: XS`, `risk: high`, `agent`, `runtime`. Because this is currently classified as a high-risk runtime change, please update the body label snapshot and replace the low-risk rollback line with the high-risk rollback fields before landing.

fanchanghu:
@Audacity88 Agreed on the blocking point — passing the loop's `history` into the hook diverges the provider-visible transcript from what `Agent::turn` / `turn_streamed` persist.

I think the right fix is to make the hook's edits durable, but that requires a change to the loop's input contract rather than just the call site. Specifically:

- `history` stays append-only and becomes an **immutable** view of past turns.
- `ToolLoop` gains a separate **mutable** `current_turn` working set: the user message for this turn, plus any hook injections, assistant replies, and tool results produced during the turn.
- The hook receives both, but is only allowed to mutate `current_turn`.
- After `run_tool_call_loop` finishes, the entire `current_turn` working set is appended to `history` and replayed through the existing `new_messages_out` path. As part of this restructuring, `new_messages_out` itself can likely be replaced by `current_turn` — the working set is already the complete unit of persistence, so a separate append-log becomes redundant.

This keeps provider-visible requests and stored history consistent by construction, while still preventing hooks from rewriting past turns. It also removes the need for the hook to declare which messages it added — the working set is the unit of persistence.

This is a larger change than the original wiring PR, so I'd like to get your take on this direction. Does splitting the loop's input into immutable `history` + mutable `current_turn` feel like the right contract to you?

singlerider:
Re-checked `9f7e5bff` locally (FETCH_HEAD asserted against the PR head). No commits have landed since @Audacity88's review, and the blocker reproduces at the source level. Not a new verdict — the standing `CHANGES_REQUESTED` is correct and remains the gate.

### Confirmed 🔴 — hook-mutated `history` bypasses the Agent replay log

The call site at `crates/zeroclaw-runtime/src/agent/turn/mod.rs:305` passes the loop's canonical transcript straight in:

```rust
match hooks.run_before_llm_call(history, &mut active_model).await {
```

and `run_before_llm_call` (`crates/zeroclaw-runtime/src/hooks/runner.rs:187-194`) hands that `&mut messages` to each hook's `before_llm_call`, so a hook can append, delete, or rewrite any message the provider then sees.

But `Agent::turn` does not persist `history` wholesale. It runs the loop with `Some(&mut loop_new_messages)` (`crates/zeroclaw-runtime/src/agent/agent.rs:1918`, passed at :1975) and afterward replays only that append-log back into the durable transcript:

```rust
for replayed in Self::replay_loop_messages(&loop_new_messages) {
    self.history.push(replayed);   // agent.rs:1999-2000
}
```

`loop_new_messages` only captures messages the loop *adds*. A hook that rewrites or deletes an *existing* `history` entry (e.g. rewriting the latest user message or injecting a policy instruction) changes the provider-visible transcript without that edit appearing in the append-log, so `self.history` ends up recording a different conversation than the model actually answered. The streamed path has the same shape for returned `new_messages`. This is exactly the append-log-contract break @Audacity88 described.

The unblock is unchanged: decide the semantics explicitly. If `before_llm_call` edits are meant as ephemeral provider-request transforms, run the hook on a per-request copy taken after the durable transcript boundary instead of mutating `history`; if they are meant to be durable, route them through the same append-log/replay path `Agent::turn` and `turn_streamed` use. Add a loop/Agent-level regression test proving a `before_llm_call` rewrite reaches the provider *and* that stored/returned state matches the intended semantics.

Also note this PR is now `CONFLICTING` / `DIRTY` against master and needs a merge from `master` to resolve conflicts before it can land. And the 🟡 stale risk metadata @Audacity88 flagged (body says `risk: low`, live label is `risk: high`) still needs the body/rollback fields updated.

No new verdict; the existing `CHANGES_REQUESTED` stands.

fanchanghu:
Thanks for re-confirming. The `9f7e5bff` commit is superseded — we're reworking this around an immutable-`history` + mutable-`current_turn` loop contract (see the comment above) rather than the original in-place `history` mutation, which is exactly what made the append-log break possible. Holding off on the conflict resolution and risk-metadata fixes until the contract direction is confirmed, then we'll force-push or re-open accordingly.

singlerider:
Moved from the released v0.8.2 to v0.8.4. This adds a before_llm_call plugin hook call site; it is a feature with no committed home in a specific point release and is not part of the v0.8.3 WASM plugin platform scope as written. Parking for re-triage, where it can be aligned with the plugin/hook program (see #7822 WASM lifecycle hooks) and scheduled into a concrete release.

Audacity88 (commented):
Checked the current PR snapshot, prior reviews/comments, the unchanged one-file diff, green checks on the old head, and the public `Agent::turn` / `turn_streamed` replay and cache paths on current `master`. I did not rerun Cargo locally.

### 🔵 Suggestion — Make `current_turn` the durable provider-visible unit

@fanchanghu Yes, the immutable `history` plus mutable `current_turn` direction is the right contract. The important part is that `current_turn` becomes the single provider-visible and durable unit for the active turn.

A few constraints I would keep explicit in the rework:

- Build the provider request from `history + current_turn`. Hooks can mutate `current_turn` and the active model, but they should not be able to rewrite prior turns in `history`.
- Seed `current_turn` at the same boundary that accepts the active turn's user and steering messages. Then persist/replay that same `current_turn` into `self.history` and streamed `new_messages`, including assistant replies and tool results. If `new_messages_out` remains temporarily, it should be derived from the same current-turn unit rather than becoming a second source of truth.
- Move the response-cache decision onto the post-hook provider-visible transcript/model, or bypass the response cache when `before_llm_call` hooks are active. Today `Agent::turn` and `turn_streamed` compute the cache key before entering the loop; with a mutating hook, that can otherwise use a cache key that no longer matches the prompt/model the provider sees.
- Add an Agent-level regression test around a hook rewriting the current user turn: the provider should see the rewrite, and the stored/returned conversation state should match the chosen semantics. If model mutation stays supported here, cover the cache/model-key behavior too.
- Preserve the loop's every-exit append-log behavior: success, error, and cancellation should all keep the provider-visible current-turn additions aligned with persisted history.

The current PR still needs that revised contract before it is unblockable, and the branch also needs the current `master` conflict plus the body label snapshot cleaned up after the rework.

Audacity88:
Closing this branch because the current implementation was explicitly superseded by the immutable `history` plus mutable `current_turn` redesign discussed above. The existing commit still breaks the Agent replay contract, the branch is now conflicting, and no replacement update has landed.

The accepted direction remains recorded here and alongside #7822. The work can return as a refreshed PR with the current-turn durability contract, post-hook cache behavior, and Agent-level replay tests.

PR #7855, opened 2026-06-17
fix(config): surface model_routes and embedding_routes via Configurable and #[nested]

### LINKED ISSUE #7701: Dashboard model selector calls wrong prop path — returns 404, selector shows single hardcoded model

crh-code:
## Summary

- **Base branch:** `master`
- **What changed and why:** `ModelRouteConfig` and `EmbeddingRouteConfig` were not derived `Configurable`, so the gateway `/api/config/prop` and `/api/config/map-keys` endpoints could not discover or address individual route entries. The dashboard model selector called `GET /api/config/prop?path=model` and got 404 because `model_routes` was invisible to the property routing system. This commit adds `Default + Configurable` derive and `#[prefix]` annotations to both structs, `#[nested]` + `#[natural_key = "hint"]` + `#[group = "Foundation"]` on the Vec fields in Config, `#[serde(default)]` on required String fields, `#[secret]` + `#[credential_class = "encrypted_secret"]` on `api_key`, and curated `ModelRoutes` and `EmbeddingRoutes` sections so the dashboard can discover and render route entries.
- **Scope boundary:** `ModelRouteConfig`, `EmbeddingRouteConfig`, their wiring into the Config property system, and curated section entries. No frontend JS changes (the frontend path fix `path=model → path=model_routes` will close #7701 in a follow-up PR). No gateway HTTP handler logic, no runtime routing logic.
- **Blast radius:** Gateway REST API responses now surface `model_routes` and `embedding_routes` as addressable config sections instead of returning 404. The dashboard sections sidebar now shows "Model routes" and "Embedding routes" entries. The secret-field encryption migrator now covers route `api_key` fields.
- **Linked issue(s):** Related #7701 (the backend groundwork; a follow-up frontend PR will close the issue)
- **Labels:** `bug`, `config`, `risk: medium`, `size: S`

- **Commands run:**
  - `cargo fmt --all -- --check` — passed (no output)
  - `cargo clippy --all-targets -- -D warnings` — passed (no warnings)
  - `cargo test` — 9 passed, 0 failed
  - `cargo test -p zeroclaw-config` — 90 passed, 0 failed
  - `cargo test -p zeroclaw-gateway` — 261 passed, 0 failed

- **Beyond CI — what did you manually verify?** Verified curated section entries (`ModelRoutes`/`EmbeddingRoutes`) render as `OneTierAliasMap` in the dashboard picker dispatch and `create_map_key` routing. Verified `section_has_signal` correctly returns `true` when `model_routes` or `embedding_routes` Vecs are non-empty. Verified `#[serde(default)]` rationale comments match `McpServerConfig.name` pattern. Verified `Config::validate()` rejects empty `hint`/`model_provider`/`model` at config-load time.

- Secrets / tokens / credentials handling changed? (`Yes — intentionally. Previously `api_key` in route configs was classified as `requires_follow_up`, meaning the encryption-at-rest migrator skipped it and deferred to manual follow-up. This PR reclassifies it to `encrypted_secret`, so route `api_key` values are now auto-encrypted on next config save. This closes a security gap: route api_keys were the only secret-type fields not covered by at-rest encryption. The change is additive (plaintext → encrypted, no data loss), idempotent (already-encrypted `enc2:` values are skipped), and does not require a `CURRENT_SCHEMA_VERSION` bump. This is confirmed as the intended end state.)

## Compatibility (required)

- Config / env / CLI surface changed? (`Yes` — `/api/config/prop`, `/api/config/map-keys`, `/api/config/sections`, `/api/config/list` now return model_routes/embedding_routes entries. Existing TOML configs deserialize unchanged. The `#[serde(default)]` on `hint`/`model_provider`/`model` is required for `create_map_key` (which constructs from `serde_json::from_value(json!({}))` before injecting the natural key), matching the `McpServerConfig.name` pattern. Empty-string defaults are rejected by `Config::validate()` at config-load time, so on-disk configs with missing required fields still fail loudly. The credential-class change from `RequiresFollowUp` to `EncryptedSecret` is additive: previously unencrypted route api_keys are now correctly classified and auto-encrypted.)

- **Feature flags or config toggles:** None — purely additive.
- **Observable failure symptoms:** If the `Configurable` derive generates incorrect `map_key_sections` entries, the dashboard "+" button for model_routes / embedding_routes would fail. If `secret_field_terminals()` is incorrect, the encryption migrator would miss route `api_key` fields.


singlerider (changes requested):
Pulled locally against `upstream/master`, read the full diff, cross-checked against issue #7701 and the on-disk schema state, and ran the config suite. The mechanics are correct and well-tested, but there are three things to resolve before this lands.

### 🟢 Praise: the registry wiring is done right

The `display_secret_terminals()` impls in `traits.rs` delegate to the per-struct `secret_field_terminals()` helper rather than hardcoding a terminal list in the trait file, and removing `model_routes`/`embedding_routes` from the `UNGROUPED` exclusion list in `sections.rs` is the correct consequence of giving them `#[group = "Foundation"]`. The two new `map_key_sections` tests pin `natural_key == "hint"` with a clear rationale comment tying it to the dirty-path writer's array-of-tables branch. Config suite is green locally (911 passed).

### 🔴 Blocking: `Closes #7701` does not close #7701

Issue #7701 reports a concrete, reproducible bug: the dashboard JS calls `GET /api/config/prop?path=model` and gets a 404, so the selector falls back to one hardcoded model. The issue's own root-cause section states `path=model_routes` already returns 200 on master, and the stated fix is the frontend calling the correct path. This PR makes no frontend changes (your scope boundary says so explicitly), so after merge the dashboard still calls `path=model` and still 404s. The literal reported bug stays reproducible.

What this PR actually does is enable per-entry addressing of route entries (`model_routes.<hint>.*`) in the nested prop/map-key system, which master genuinely lacked. That is useful, but it is not what #7701 asks for. Either retarget this as `Related #7701` and let the frontend fix close it, or add the one-line dashboard path fix here so the `Closes` is honest.

### 🔴 Blocking: the credential-class change is a behavior change, not just surfacing

Both `api_key` fields move from the field-level `#[credential_class = "requires_follow_up"]` to a per-field `#[secret] #[credential_class = "encrypted_secret"]`. That changes how the secret-encryption migrator treats these values: a plaintext `api_key` already present in someone's `model_routes` on disk now becomes auto-encrypted on the next config write, where before it was left for manual follow-up. This may well be the desired end state, but it is a credential-handling change riding inside a size-XS "surface via Configurable" PR, and it touches on-disk secret material. Please call it out explicitly and confirm it is intended, given config schema work is otherwise frozen until V4. The PR does not bump `CURRENT_SCHEMA_VERSION` (still 3) or touch `migration.rs`, which is good, but the migrator's behavior toward existing keys still changes.

### 🔴 Blocking: `#[serde(default)]` on previously-required fields silently accepts malformed routes

`hint`, `model_provider`, and `model` were required `String`s; they now carry `#[serde(default)]`. A route table missing `hint` will now deserialize to an empty-string hint instead of erroring. A route with an empty hint cannot be addressed (`hint:` resolves to nothing) and is silently useless. The `Default` derive forces defaultability, but consider keeping deserialization strict (a malformed route should fail loud) and satisfying `Default` another way, or add a validation pass that rejects empty `hint`/`model_provider`/`model`.

Resolve the `Closes` mismatch and confirm the credential-class intent and these are good to go; the surfacing work itself is sound.

Audacity88 (commented):
Reviewed current head `4124322` against the live PR state, #7701, the prior requested-changes review, the current config/gateway diff, and the relevant config validation and section-routing code. I did not run local Cargo validation; I relied on source review plus the current green GitHub CI on this head.

### ✅ Resolved — backend route surfacing now matches the scoped claim

From my pass, the backend route-surfacing concerns are addressed on the current head. The PR body now frames the work as backend groundwork related to #7701, not as the frontend dashboard path fix itself. That matches the diff: the route tables become discoverable and addressable through the config section/property machinery, while the dashboard caller change remains follow-up scope.

The credential handling is also explicit now. The route `api_key` fields are marked as secret `encrypted_secret` fields on the nested structs, which is the place the derived property surface can use. The `serde(default)` additions are paired with `Config::validate()` checks that reject empty `hint`, `model_provider`, and `model` values for both model and embedding routes, so malformed persisted route entries still fail validation instead of becoming silent empty routes.

### 🟢 What looks good — route entries use the existing section machinery

Adding curated `ModelRoutes` and `EmbeddingRoutes` sections, `#[nested]` / `#[natural_key = "hint"]`, `map_key_sections` coverage, signal detection, and gateway picker/create-map-key dispatch makes these route tables follow the same operator-named config-map path as the surrounding section machinery. That is the right shape for this backend slice.

### 🔵 Suggestion — keep final closing-reference cleanup separate from author code blockers

One public-record item still needs maintainer-side cleanup before final merge handling: GitHub still sees #7701 as a closing issue because an earlier branch commit used `Closes #7701`. I am not treating that as an author code blocker here, but the final squash/body cleanup should keep #7701 as related follow-up work rather than closing it from this backend PR.

singlerider (approved):
Re-reviewed at head `41243223` (checked out locally, built `zeroclaw-config`, ran the route/map-key suite). All three blockers are resolved, and the validation backstop is more thorough than I asked for. Approving.

### RESOLVED ✅ — `Closes` retargeted to `Related #7701`

The body now reads `Related #7701` with an explicit note that the frontend `path=model → path=model_routes` fix lands in a follow-up. Honest about what this PR does (backend addressability) vs what closes the reported 404.

### RESOLVED ✅ — credential-class change is now called out and confirmed intentional

The Security & Privacy section spells it out: `requires_follow_up → encrypted_secret` is intentional, additive (plaintext → encrypted, no data loss), idempotent (`enc2:` skipped), needs no `CURRENT_SCHEMA_VERSION` bump, and closes the gap where route `api_key` was the only secret-type field outside at-rest encryption. That's exactly the explicit confirmation I needed for on-disk secret material riding in a size-XS PR.

### RESOLVED ✅ — `#[serde(default)]` is backstopped by loud validation

`Config::validate()` now hard-`validation_bail!`s on empty `hint` / `model_provider` / `model` for both `model_routes` (`schema.rs:16978-17017`) and `embedding_routes` (`17024-17061`), and goes further with dotted-form and dangling-reference checks. So the parse-time permissiveness needed for `create_map_key` can't produce a silently-useless empty-hint route — it fails loud at load. This is the validation-pass alternative I offered.

Built clean; `cargo test -p zeroclaw-config --lib` route/map-key set → 20 passed, including the new `map_key_sections_exposes_natural_key_for_{model,embedding}_routes`.

### 🟢 One non-blocking suggestion

The empty-route rejection in `validate()` is correct but has no direct unit test asserting it fires — the coverage is on the map-key/addressing side. A two-line test (a `model_routes` entry with an empty `hint` → `validate()` returns `Err`) would pin the backstop so a future refactor can't quietly drop it. Not a blocker; the logic is plainly correct.

crates/zeroclaw-config/src/schema.rs
```diff
@@ -10685,17 +10689,31 @@ impl Default for SchedulerConfig {
...
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, Default, Configurable)]
 #[cfg_attr(feature = "schema-export", derive(schemars::JsonSchema))]
+#[prefix = "model_routes"]
 pub struct ModelRouteConfig {
     /// Task hint name (e.g. "reasoning", "fast", "code", "summarize")
+    /// `#[serde(default)]` lets the `create_map_key` macro default-construct
+    /// from `{}` before the hint gets injected via `set_prop`. Empty strings
+    /// are rejected by `Config::validate()`.
+    #[serde(default)]
```
singlerider, line 11058: 🟢 The #[serde(default)] here (needed so create_map_key can default-construct from {}) is safely backstopped: Config::validate() hard-bails on empty hint/model_provider/model for both route kinds, plus dotted-form and dangling-ref checks. Malformed routes fail loud at load instead of becoming silently-useless empty-hint entries. Resolves my prior blocker. Non-blocking: a direct unit test on that Err path would pin it.

PR #7857, opened 2026-06-17
fix(zerocode): skip queue-paused hint when backlog is empty

### LINKED ISSUE #7805: [Bug]: Cancelled Code turns can show queue-paused hint with an empty queue
### Affected component

zerocode/tui

### Severity

S3 - minor issue

### Current behavior

After cancelling or failing a Code/Chat turn, the transcript can show the correct turn-end system line while the empty input bar still shows:

```text
Queue paused — press \p or send a message to resume
```

This happens even when there are no queued messages. In that state, the user cancelled the current turn; they did not leave any queued follow-up work paused.

Current source inspection matches the behavior:

- `TurnEndOutcome::Cancelled` and `TurnEndOutcome::Failed` render the daemon-provided system line and then call `commit_turn(String::new(), false)`.
- `commit_turn(..., clean = false)` sets `queue_paused = true` unless `resume_override` is active.
- The input-bar render path shows the `zc-queue-paused-ghost` text whenever `state.queue_paused()` is true.
- The render path does not check whether `message_queue` has any pending or injected messages before showing that ghost text.

### Expected behavior

Cancelling or failing a turn should only show queue-paused recovery text when there is actually queued work being held.

If a turn is cancelled with an empty queue, the input bar should return to the normal empty prompt, or show a neutral cancellation state that does not tell the user to resume a nonexistent queue.

The existing protection for real queued backlog should remain: if a non-clean turn ends while pending or injected messages exist, ZeroCode should still pause that backlog until the user explicitly resumes it.

### Steps to reproduce

```bash
1. Open ZeroCode Code or Chat.
2. Start a turn.
3. Do not enqueue any follow-up messages.
4. Cancel the turn or let it fail.
5. Observe the transcript shows the cancelled/failed turn line.
6. Observe the empty input bar can still show `Queue paused — press \p or send a message to resume` even though the queue is empty.
```

### Impact

The UI conflates two different states: "the active turn was cancelled" and "queued messages are paused." Users are told to resume a queue that does not exist, which makes the session feel stuck even though the active turn already ended.

This is lower priority than the cancellation routing bug, but it makes failed/cancelled turns harder to understand and recover from.

### Logs / stack traces

No stack trace. The relevant visible text is:

```text
Queue paused — press \p or send a message to resume
```

### ZeroClaw version

Current `master` at `883164eec31b45fa54c33fa9fb4d6123f7f71d3d`.

### Rust version

N/A; this is a ZeroCode TUI state/rendering issue.

### Operating system

macOS observed. The queue state/rendering path appears platform-independent.

### Regression?

Unknown.

### Related

- #6825
- #7190
- #7800

### Pre-flight checks

- [x] I observed this in ZeroCode and source-confirmed the relevant queue/render paths on latest `master`.

Pick-cat:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `commit_turn(clean=false)` now sets `queue_paused` only when the outbound `message_queue` is non-empty, so cancelled/failed turns without queued follow-ups no longer enter a paused state.
  - The input-bar ghost text for queue-paused recovery is gated on both `queue_paused()` and `queue_len() > 0`, matching visible UI to real backlog state.
- **Scope boundary:** Does not change dispatch, inject, or manual `\p` resume behavior when backlog exists.
- **Blast radius:** ZeroCode Code/Chat input-bar rendering after cancelled or failed turns.
- **Linked issue(s):** Closes #7805
- **Labels:** `bug`, `zerocode`, `risk: low`, `size: XS`

```bash
cargo fmt --all -- --check
cargo test -p zerocode non_clean_commit_with_empty_queue

GitHub CI at head 164624a0987a85ba84b9bdf8f3bdf78ebadc0e9a:
Result: all visible checks passing, including Format, Lint, Test, Security, platform checks, Zerocode RPC Boundary, and CI Required Gate.
```

- **Commands run and tail output:** `cargo fmt --all -- --check` passed locally. `cargo test` compile skipped locally due missing system `cc` linker; GitHub CI passed on the current head.
- **Beyond CI - what did you manually verify?** Added unit test `non_clean_commit_with_empty_queue_does_not_pause`; existing queue-pause tests unchanged for non-empty backlog.
- **If any command was intentionally skipped, why:** Local `cargo test` was skipped because the contributor environment was missing `build-essential` / `cc`; GitHub CI covered the final branch head.

## Compatibility (required)


singlerider (approved):
Read the full diff against `apps/zerocode/src/chat.rs` and traced every reader of `queue_paused` on master. The fix is correct and the behavior change is the right one.

### 🟢 What looks good - the state gate matches the only behavioral consumer

`queue_paused` has exactly one control-flow reader, `next_dispatch_index` at chat.rs:4218, which suppresses dispatch of `Pending` items while paused. Pausing on a non-clean turn end exists so a cancelled/failed turn does not silently auto-dispatch a queued follow-up. Gating that on `!self.message_queue.is_empty()` means the pause only arms when there is actually a backlog to hold; with an empty queue there is nothing to protect, so entering a paused state just swallows the next message the user types, which is precisely the phantom state #7805 describes. The non-empty case is unchanged, so the deliberate-resume semantics survive where they matter.

### 🟢 What looks good - the UI hint and the state now agree

Gating `queue_paused_hint` on both `queue_paused()` and `queue_len() > 0` makes the ghost text reflect real backlog state rather than the raw flag. Both halves of the fix move in the same direction, so there is no window where the hint shows but dispatch wouldn't pause (or vice versa). The added `non_clean_commit_with_empty_queue_does_not_pause` test pins the core invariant, and the existing non-empty pause tests at 5626/5681 remain green to cover the other side.

Scope holds (no dispatch/inject/manual-resume change when a backlog exists), risk/size labels are right, and the milestone (v0.8.3) fits a narrow zerocode stabilization fix. The "Made with Cursor" footer in the PR body is harmless here but should be stripped from the squash artifact at merge time.

Approving.

Audacity88 (approved):
Reviewed current head `fb6f198dcc726f0ef7604715bbd0d15e928323e2` against #7805, the live PR state, singlerider's approval, the current diff, current `master` queue code, and the current-base freshness data. I did not run local Cargo validation for this review-only pass.

### 🟢 What looks good — The paused state now matches real queued work

The fix is the right shape for #7805. `commit_turn(clean = false)` still pauses after a cancelled or failed turn when `message_queue` has backlog, so pending follow-up messages do not silently dispatch. When the queue is empty, it no longer arms `queue_paused`, which removes the phantom "resume queue" state from the user-visible flow.

### 🟢 What looks good — The UI hint and dispatch gate stay aligned

The render path now shows the queue-paused ghost text only when `queue_paused()` is true and `queue_len() > 0`. That keeps the input-bar hint tied to an actual held backlog. The new `non_clean_commit_with_empty_queue_does_not_pause` regression covers the empty-queue cancel/fail case, while the existing non-empty queue and resume tests keep the intended backlog behavior pinned.

### 🟡 Warning — Refresh before merge because current `master` changed the same file

This approval is not a merge-ready call. The PR is currently `CLEAN` and CI is green, but the freshness check reports `behind/ahead: 70/1` with same-file churn in `apps/zerocode/src/chat.rs`. Before landing, this still needs either a branch update with fresh CI, an exact merge-result smoke for the affected ZeroCode path, or explicit maintainer stale-risk acceptance.

Approving.

Issue #6548, opened 2026-05-09
[Bug]: Channel runtime command replies bypass Fluent localization


drbparadise:
### Affected component

channel

### Severity

S3 - minor issue

### Current behavior

Some user-visible channel runtime replies are still hard-coded English strings even when ZeroClaw is configured with a non-English locale such as `zh-CN`.

Observed examples in the channel runtime command path include:

```text
Conversation history cleared. Starting fresh.
Stop signal sent.
No in-flight task for this sender scope.
Current provider: `...`
Available providers:
Model switched to `...` (provider: `...`). Context preserved.
```

There are also adjacent channel fallback messages such as context-window and timeout replies that are user-visible but are not represented as Fluent messages.

This appears inconsistent with the current localization contract in `AGENTS.md`, which says user-facing output should use Fluent strings while logs, tracing, and panic messages stay English.

### Expected behavior

User-visible channel runtime replies should be resolved through the existing Fluent CLI string catalog, with English fallback behavior preserved.

The configured locale should apply to built-in channel runtime replies such as `/new`, `/stop`, `/model`, `/models`, `/config`, and closely related channel fallback messages.

### Steps to reproduce

```bash
# On latest master, inspect the channel runtime command responses.
rg -n 'Conversation history cleared|Stop signal sent|Current provider|Available providers|Context preserved|Request timed out' \
  crates/zeroclaw-channels/src/orchestrator/mod.rs

# Configure a non-English runtime locale such as zh-CN, then use a supported channel
# command such as /new or /models. The command reply still contains hard-coded
# English text instead of a localized Fluent string.
```

### Impact

Affected users: non-English users interacting through chat channels rather than the CLI.

Frequency: whenever these built-in channel runtime commands or fallback paths are triggered.

Consequence: the agent can otherwise be configured for a localized experience, but internal system replies still appear in English in the user chat surface.

This is not a request for a new localization system. It is a coverage gap in the existing Fluent migration. Related prior work:

- #5787 / #5788 introduced the Fluent direction and localization policy.
- #6131 / #6242 provide a recent scoped precedent for moving channel user-facing strings into Fluent.

Suggested PR slicing:

1. First PR: localize the core channel runtime command replies and adjacent orchestrator fallback messages.
2. Follow-up PR: localize Telegram command menu descriptions.
3. Follow-up PR: localize tool approval prompts across channel adapters.

### Logs / stack traces

No crash or stack trace. Representative source strings on latest master:

```text
[...]

PR #7858, opened 2026-06-17
fix(channels): localize core runtime command replies

Pick-cat:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Channel runtime `/new`, `/stop`, `/model`, and timeout fallback replies now resolve through Fluent (`channel-runtime-*` keys) instead of hard-coded English.
  - Adds `zh-CN` translations for the new keys; other locales fall back to English per existing i18n contract.
- **Scope boundary:** Does not localize `/models` help text, provider-switch prose, or Slack Block Kit payloads (follow-up slices for #6548).
- **Blast radius:** User-visible channel runtime command replies only.
- **Linked issue(s):** Related #6548 (first slice — core command replies)
- **Labels:** `bug`, `channel`, `channel:core`, `risk:medium`, `size: S`

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-channels runtime_command
```

- **Commands run and tail output:** `cargo fmt --all -- --check` passed locally. Full compile/test skipped — no system `cc` linker; CI runs the battery.
- **Beyond CI — what did you manually verify?** Updated `/new` session integration assertion to resolve the Fluent string instead of a hard-coded English literal.
- **If any command was intentionally skipped, why:** Missing `build-essential` / `cc` linker locally.

## Compatibility (required)

- Backward compatible? **Yes** (English text unchanged; locales can now translate)


Audacity88:
@Pick-cat thanks for taking this slice on. I checked the live PR state, linked issue #6548, current checks, and the current diff. I am holding formal review for now because GitHub currently reports this branch as conflicting with `master` (`CONFLICTING` / `DIRTY`).

Could you refresh the branch against current `master` and let CI rerun?

While touching the branch/body, please also refresh the public PR metadata:

- The live labels are `bug`, `size: S`, `risk: medium`, `channel`, and `channel:core`; the body still says `risk: low` and `runtime`.
- The PR body says `Closes #6548 (first slice — core command replies)`, but the issue still includes remaining channel localization surfaces that this PR explicitly leaves to follow-ups. `Related #6548` or equivalent first-slice wording would avoid auto-closing the broader issue accidentally.

The localization direction looks aligned with the issue: move the core user-visible channel runtime replies through Fluent, keep English fallback behavior, and add `zh-CN` coverage for the new keys. Once the branch is clean, I can do the actual code review against the mergeable head.

singlerider (approved):
Template is complete and the slice is clean. Approving.

### 🟢 i18n fallback is correct — en + zh-CN is sufficient

I checked the claim that "other locales fall back to English." It holds, and I verified the mechanism rather than taking it on faith: `load_cli_strings` (i18n.rs:142) seeds the map from `en/cli.ftl` and then extends with the active locale, and `format_cli_string_with_args` (i18n.rs:174-189) tries the locale's builtin source first, then falls back to `en/cli.ftl`. So a fr/es/ja user hitting `/new`, `/stop`, `/model`, or the timeout reply resolves to the English string, not the `{key}` placeholder (`missing_cli_string` only fires when the key is absent from English too). Adding es/fr/ja is a nicety, not a contract requirement — the neighboring `channel-runtime-malformed-tool-output` happens to have all five, but nothing breaks without them.

### 🟢 Strings, args, and the test all line up

- All six new `channel-runtime-*` keys exist in `en/cli.ftl` with byte-identical English wording to the strings they replace, so English users see no change.
- `channel-runtime-model-switched` declares `{ $model }` / `{ $provider }`, matching the `get_required_cli_string_with_args(..., &[("model", ...), ("provider", ...)])` call site exactly.
- The `/new` integration assertion now resolves the Fluent string instead of the old hard-coded literal, so it can't silently pass on a stale English literal.
- `finalize_draft(..., error_text.as_str())` is correct now that the helper returns `String`.

Two non-blocking notes for whenever you next touch this branch:

- The PR carries the `risk: medium` GitHub label, but the body's Rollback section uses the low-risk `git revert` default and the body's own label snapshot says `risk: low`. For a pure i18n string swap `git revert` genuinely is the plan, so this is a label/prose mismatch rather than a missing rollback — worth reconciling the label to `risk: low` or filling the medium-risk rollback fields.
- The trailing `Made with Cursor` footer isn't a `Co-authored-by` violation, but the template asks to keep tool footers out of the body; drop it on the next edit.

Audacity88 (approved):
I rechecked the live PR state, linked #6548 and sibling PRs, prior comments/reviews, the current diff, the runtime i18n helper path, visible CI, and the locale catalog changes.

### 🟢 Approved — core i18n slice is sound

Thanks, this slice looks good. The converted replies use the existing required CLI string helpers, `/model` goes through the arg-aware formatter for `model` / `provider`, zh-CN gets built-in translations, and other locales still fall back to English. The scope also matches the PR body's first-slice boundary: `/models`, provider-switch prose, and adapter-specific payloads can stay follow-up work for #6548.

### 🔵 Follow-up — merge hygiene still needs cleanup

This approval is for the reviewed implementation, not merge readiness. GitHub still reports the branch as `DIRTY` / conflicting against `master`; after resolving that, CI needs to rerun on the refreshed head. Before merge, the public metadata also still needs the mechanical cleanup from my earlier note and @singlerider's review: avoid auto-closing #6548 while follow-up localization remains, align the body label snapshot with the live labels, and drop the generated footer.

singlerider:
@Pick-cat You have conflicts.

Pick-cat:
@singlerider @Audacity88 Thanks for the nudge — conflicts are resolved.

**Merge refresh:** Merged current `master` into the branch (`7592840`). The only conflict was in `crates/zeroclaw-runtime/locales/zh-CN/cli.ftl`; kept the new `cli-models-*` keys from upstream alongside our `channel-runtime-*` translations, matching `en/cli.ftl` ordering.

**Metadata cleanup (per earlier review):**
- PR body now says **Related #6548** (first slice) instead of Closes.
- Label snapshot aligned with live labels (`channel:core`, `risk:medium`).
- Removed the generated tool footer.

`cargo fmt --all -- --check` passes locally; full test compile still skipped here (no `cc` linker). CI is rerunning on the refreshed head.

PR #7903, opened 2026-06-17
fix(runtime): replay ACP session history in session/messages

### LINKED ISSUE #7799: [Bug]: Resumed Code sessions reopen with a blank transcript
## Affected component

zerocode/tui

## Severity

S2 - degraded behavior

## Current behavior

Resuming a saved Code session can reopen the session with an empty visible transcript even when the session picker shows that the session has persisted messages.

The resume itself appears to attach to the saved Code session, but the conversation pane does not replay the prior user/assistant messages. Users cannot inspect the earlier turn history and may believe the session history was lost.

Source inspection on current `master` matches the observed behavior:

- Code resume and session switching call `session/messages` and then `load_history(...)`.
- Code session listing uses `session/list/acp`, which reads the dedicated ACP session store and reports `message_count`.
- `session/messages` reads only the unified `session_backend` and tries the raw session id plus `rpc_` / `gw_` prefixed keys.
- ACP sessions are persisted in `acp_session_store`, so `session/messages` can return an empty message list for a valid ACP session whose messages exist in the ACP store.

## Expected behavior

When a Code session is resumed or selected from the Code session list, the visible transcript should replay the persisted ACP session messages.

If history replay fails, the UI should surface that failure instead of silently showing a blank conversation for a session that the picker reports as non-empty.

## Steps to reproduce

```bash
# 1. Start ZeroCode and open a Code session.
# 2. Send at least one user message and receive an assistant response.
# 3. Close or leave the Code pane.
# 4. Reopen ZeroCode and use the Code session list to resume the saved session.
# 5. Observe that the session picker can show a non-zero message count, but the resumed conversation pane is blank.
```

## Impact

Users lose the visible context needed to understand what happened in the prior session, even though the saved ACP session still appears to exist.

This makes session resume feel unreliable and can force users to reconstruct prior work manually. It also hides the difference between "no persisted history exists" and "persisted ACP history exists but the TUI did not load it into the transcript."

## Logs / stack traces

No stack trace was observed. The failure is silent: the Code pane treats an empty `session/messages` response as a blank transcript.

## ZeroClaw version

Observed in ZeroCode dogfooding and source-confirmed on `master` at `883164eec31b`.

## Rust version

N/A; this is a runtime/TUI session replay bug rather than a compiler/toolchain-specific failure.

## Operating system

macOS was used for the observed Code resume behavior. The affected ACP session replay path is not expected to be OS-specific.

## Regression?

Unknown. This appears to be a gap in the ACP session persistence/replay path after ACP sessions were moved into a dedicated store.

## Related

- #6649
- #7158

## Pre-flight checks

- I observed the blank resumed transcript in a Code session and source-confirmed the replay path on latest `master`.

tidux:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - `session/messages` previously read only the unified `session_backend`, trying the raw id plus `rpc_<id>` / `gw_<id>` prefixed keys. ACP sessions live in the dedicated `acp_session_store` keyed by their raw UUID, so none of those candidates ever matched and the Code (ACP) pane in the TUI rendered a blank transcript on resume even though the picker (which calls `session/list-acp` against the ACP store) reported a non-zero `message_count`.
  - Falls back to `acp_session_store` when the unified backend yields no messages for the requested id. Flattens `ConversationMessage` → `ChatMessage` by keeping only `Chat(...)` variants — the wire schema is `{ role, content }` per message and the TUI's `load_history` only renders user/assistant chat, so `AssistantToolCalls` and `ToolResults` are intentionally dropped.
  - Store errors surface as `INTERNAL_ERROR` instead of silently returning empty, matching the issue's expectation that history-replay failures be visible to the UI.
  - Adds a `#[cfg(test)] handle_session_messages_for_test` entry point and a regression test seeded directly in `acp_session_store` with the unified backend explicitly empty for all candidate keys.
- **Scope boundary:** Does not change the wire schema for `session/messages`, does not change the ACP write path, does not change how the TUI calls `load_history`, and does not change `session/list-acp`. Tool-call and tool-result history remain absent from the replayed transcript — surfacing those needs a wire-schema change tracked separately.
- **Blast radius:** Only the `session/messages` RPC handler. Chat-pane sessions still resolve from the unified `session_backend` on the existing fast path; the ACP fallback runs only when that path yields zero messages.
- **Linked issue(s):** Closes #7799
- **Labels:** `bug`, `runtime`, `zerocode`, `channel:acp`, `risk: high`, `size: S`.

`cargo fmt --all -- --check` — exit 0, no diff output.

\`\`\`
    Checking zeroclaw-channels v0.8.0 (.../crates/zeroclaw-channels)
    Checking zeroclaw-eval v0.8.0 (.../crates/zeroclaw-eval)
    Finished \`dev\` profile [unoptimized + debuginfo] target(s) in 1m 15s
\`\`\`
(no warnings; exit 0.)

`cargo test --workspace` — all suites green, zero failures. New regression test:
\`\`\`
test rpc::dispatch::tests::session_messages_falls_back_to_acp_store_for_acp_sessions ... ok
\`\`\`
Notable suite tails (in order):
\`\`\`
running 982 tests   ... test result: ok. 982 passed; 0 failed; 0 ignored
running 1331 tests  ... test result: ok. 1331 passed; 0 failed; 0 ignored
running 2267 tests  ... test result: ok. 2265 passed; 0 failed; 2 ignored   (zeroclaw-runtime — includes the new test alongside 43 other dispatch tests)
running 277 tests   ... test result: ok. 277 passed; 0 failed; 0 ignored
\`\`\`
The `0 passed; N ignored` suites are live-credential / proprietary-license gated (`live_zai_jwt_auth_*`, `gemini_warmup_*`, `openai_codex_*`) and are ignored on master as well.

- **Beyond CI — what did you manually verify?** Walked the call graph: `apps/zerocode/src/chat.rs::load_history` consumes only `MessageRole::User` / `Assistant` from the wire message list, so dropping `AssistantToolCalls` and `ToolResults` during the `ConversationMessage` → `ChatMessage` flatten loses nothing the transcript would have rendered. Confirmed `RpcContext.acp_session_store` is the same handle used by `handle_session_list_acp`, so picker counts and replay reads come from the same source of truth. End-to-end resume of a saved ACP session against a live daemon was exercised manually and the prior turns replay into the Code pane.

- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No` — the test fixture uses opaque ids (`acp-resume-7799`, `/tmp/ws`)

## Compatibility (required)

- Backward compatible? `Yes`
- The `session/messages` wire schema (`{ session_id, messages: [{role, content}], total, start }`) is unchanged. Existing chat-pane callers hit the same fast path with identical results. ACP-pane callers now receive replayed messages instead of empty arrays — strictly additive behaviour, no upgrade steps.

The auto-labeler will likely tag this `risk: high` per `src/runtime/**`. The change is a self-contained fallback in a single RPC handler.

- **Fast rollback command/path:** `git revert <merge-sha>` — single-file revert restoring the prior `handle_session_messages` body. No data migration, no schema change, no config change.
- **Observable failure symptoms:** Blank transcript on ACP resume (the pre-fix behaviour) returns. Grep TUI logs for `session/messages` responses where `total == 0` against an ACP session whose `session/list-acp` row reports `message_count > 0`. No new error log lines or metrics are emitted.


Audacity88 (changes requested):
I checked the PR diff, linked issue #7799, current public checks/reviews/comments, and the adjacent ACP persistence, runtime history, and TUI replay paths. I did not run local validation; several CI jobs were still pending in the snapshot I reviewed.

### 🟢 What looks good — ACP-store fallback restores the blank-session path

Thanks for narrowing the `session/messages` side of #7799. Falling back to `acp_session_store` is the right direction for replaying persisted ACP sessions.

### 🔴 Blocking — Preserve assistant tool-call narration when replaying ACP history

The fallback in `crates/zeroclaw-runtime/src/rpc/dispatch.rs:1984` only keeps `ConversationMessage::Chat` rows and drops every `AssistantToolCalls` variant. ACP persistence stores assistant tool-call rows with their visible narration in `acp_messages.content`, and the runtime explicitly avoids also writing a duplicate plain `Chat(assistant)` row for that same text. As a result, any Code session where the assistant emits narration while issuing a tool call still reopens with that assistant message missing, even though that content is part of the saved transcript.

Please preserve the visible assistant text from `AssistantToolCalls { text: Some(..), .. }` when flattening ACP history for the `{ role, content }` `session/messages` schema, while continuing to omit tool-call metadata and tool results that the current wire shape cannot represent. The regression should also seed an ACP session with an `AssistantToolCalls { text: Some(..) }` row so this case cannot silently regress.

tidux:
@Audacity88 Verified the claim: `AssistantToolCalls.text: Option<String>` carries the narration (`crates/zeroclaw-api/src/model_provider.rs:178-184`), and the agent's duplicate-narration guard (`narration_with_tool_calls_produces_no_consecutive_assistant_entries` in `crates/zeroclaw-runtime/src/agent/agent.rs`) intentionally suppresses a paired `Chat(assistant)` row, so the narration on tool-call turns only exists on that variant. Dropping it during ACP replay does lose visible turns.

Fixed in 53845cc3f. The flatten now maps:

- `Chat(...)` → pass through
- `AssistantToolCalls { text: Some(t), .. }` with non-empty `t` → assistant `ChatMessage(t)`
- `AssistantToolCalls { text: None | Some(""), .. }` → drop (nothing renderable; tool-call metadata still can't ride the `{ role, content }` wire shape)
- `ToolResults(_)` → drop (same wire-shape limitation; TUI `load_history` ignores them)

Tool-call metadata and tool results in replay remain out of scope here — that needs the wire-schema change called out in the PR description.

The regression test now seeds:

1. `Chat(user)`
2. `AssistantToolCalls { text: Some("let me check the logs"), tool_calls: [tc-1], .. }`
3. `ToolResults([tc-1])`
4. `AssistantToolCalls { text: None, tool_calls: [tc-2], .. }`
5. `ToolResults([tc-2])`
6. `Chat(assistant)`

and asserts `total == 3` with replayed transcript `[user "hello…", assistant "let me check the logs", assistant "ack…"]`. This pins down all four branches of the flatten, so both directions (losing narration, or accidentally including text-less tool-call rows / tool results) regress loudly.

Validation:
- `cargo clippy --all-targets -- -D warnings` — clean
- `cargo test -p zeroclaw-runtime --lib rpc::dispatch` — 46/46 passing
- `cargo test --workspace` — all suites green, 0 failures
- `scripts/ci/docs_quality_gate.sh` — no docs in diff, skipped

Audacity88 (approved):
Checked the updated head `53845cc3`, prior blocking review, linked issue #7799, current PR thread/reviews/comments, green GitHub checks, and the relevant runtime/API/TUI source paths. I did not run local Cargo validation for this review-only pass; I am relying on CI plus source and diff inspection here.

### ✅ Resolved — assistant tool-call narration is preserved during ACP replay

The new flattening path now preserves visible assistant narration from `AssistantToolCalls { text: Some(t), .. }` by replaying it as an assistant `ChatMessage`, while still omitting textless tool-call rows and `ToolResults` that the existing `{ role, content }` `session/messages` schema cannot represent. That addresses the blocker from my previous review: tool-call turns no longer lose their visible assistant text when a saved ACP session is resumed.

### 🟢 What looks good — the fallback stays scoped to the ACP persistence gap

The normal unified `session_backend` path still wins for existing chat/RPC sessions, and the ACP-store fallback only runs when those candidate keys return no messages. The regression now seeds the exact production split from #7799: messages exist only in `acp_session_store`, the unified backend is empty for the raw/`rpc_`/`gw_` keys, and the replay result contains the user message, recovered assistant narration, and later assistant chat while dropping unrenderable tool metadata/results.

Approving the code fix.

PR #7912, opened 2026-06-18
fix(whatsapp_storage): store app-state mutation MACs raw, not JSON-wrapped

danielO99:
## Problem

`RusqliteStore`'s mutation-MAC methods JSON-encode the MAC bytes before storing (`put_mutation_macs`) and return the JSON-wrapped `value_mac` to the caller (`get_mutation_mac`), rather than storing/returning the raw bytes.

On a fresh pairing this fails the critical app-state sync with `patch snapshot MAC mismatch`, and since the client gates `Event::Connected` on the critical sync completing, the device never finishes connecting.

## Root cause

`get_mutation_mac`'s returned `value_mac` is fed directly into the app-state LTHash update (`wacore/src/appstate_sync.rs` -> `db_prev` -> `state.update_hash`). The hash must be computed over the raw MAC bytes. Because the value is stored via `serde_json::to_vec(&value_mac)` and returned as-is, the caller receives the JSON encoding, such as the ASCII bytes for `[12,34,...]`, instead of the original bytes, so the running collection hash diverges from the server's snapshot MAC.

## Fix

Store, query, and return the `index_mac` / `value_mac` as raw bytes, dropping the `serde_json` wrapping in `put_mutation_macs`, `get_mutation_mac`, and `delete_mutation_macs`, matching the reference implementations:

- `wacore/src/store/in_memory.rs` stores and returns raw MAC bytes.
- `storages/sqlite-storage` (diesel `SqliteStore`) stores raw BLOBs.

Builds on #7806 (`get_version`): with that fix in place, this snapshot-MAC mismatch is the next failure that gates `Connected` on first pairing.

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Mutation MACs are now persisted and returned raw, so `get_mutation_mac` hands the app-state LTHash the correct bytes and the critical sync verifies.
  - Aligns `RusqliteStore` with `InMemoryStore` and the diesel `SqliteStore`.
  - Adds a focused round-trip regression using NUL and high-byte MAC values that would diverge under JSON wrapping.
- **Scope boundary:** Only the three mutation-MAC methods in `RusqliteStore`. No trait/signature changes, no other store methods, no schema change; the column is already a BLOB.
- **Blast radius:** `RusqliteStore` consumers in the `whatsapp-web` keystore path.
- **Linked issue(s):** None. Follow-up to #7806.
- **Labels:** `bug`, `channel`, `channel:whatsapp`, `risk: medium`, `size: XS`.

The store is `whatsapp-web`-gated, so the focused battery is run with `--features whatsapp-web`; `fmt` is workspace-wide.

```bash
cargo fmt --all -- --check
Result: passed.

cargo clippy -p zeroclaw-channels --all-targets --features whatsapp-web -- -D warnings
Result: passed.

cargo test -p zeroclaw-channels --features whatsapp-web
Result: passed. Includes `whatsapp_storage::tests::mutation_macs_round_trip_raw_bytes`; full result reported 1083 passed, 0 failed, 1 ignored.

GitHub CI at head acd3139d9f95f084d9ad9397a64fdfb20d69f458:
Result: all visible checks passing, including Format, Lint, Test, Security, platform checks, and CI Required Gate.
```

- **Commands run and tail output:** See command/result summary above from contributor validation and final GitHub CI.
- **Beyond CI - what did you manually verify?** The changed methods now bind/query raw byte slices against existing BLOB columns, matching the `InMemoryStore` and diesel `SqliteStore` contracts. The new regression covers put/get/delete and byte values that JSON encoding would not preserve verbatim.
- **If any command was intentionally skipped, why:** No focused `whatsapp-web` validation was skipped in the contributor evidence. Final merge validation is green GitHub CI on the current head.

## Compatibility (required)

- If `No` or `Yes` to either: N/A.

Medium-risk PR: revert the eventual squash commit. No schema, config, or data migration is involved. Observable regression symptoms would be fresh WhatsApp pairing or app-state sync returning to `patch snapshot MAC mismatch` / failure to reach `Event::Connected`.

N/A


singlerider (approved):
Pulled this branch (head e2ec40e), built `zeroclaw-channels` with `whatsapp-web`, and ran the new round-trip test (passes). I also verified the fix against the canonical upstream stores rather than trusting the comment. It is correct.

### 🟢 What looks good - the raw-bytes fix matches the canonical stores and the column type

`AppStateMutationMAC::index_mac` / `value_mac` are `Vec<u8>` (wacore processor.rs:17-18), and the `app_state_mutation_macs` columns are already `BLOB NOT NULL` (whatsapp_storage.rs:270-271). The old code wrote `serde_json::to_vec(&mutation.value_mac)`, i.e. a JSON array like `[0,127,128,...]`, into a BLOB column, so the bytes the LTHash later read back were the JSON text, not the MAC. Binding the raw `Vec<u8>` via `params![]` is the right rusqlite mapping to BLOB. I confirmed this aligns with the diesel SqliteStore (binds raw `Vec<u8>`, `.eq(&index_mac)` on raw bytes) and InMemoryStore (raw bytes as the HashMap key) in the upstream whatsapp-rust crate, so the comment's "matches InMemoryStore and the diesel SqliteStore" claim is accurate, not aspirational.

### 🟢 What looks good - symmetric across put/get/delete, with a real round-trip test

The change drops the JSON wrap on the write path and reads/deletes by raw `index_mac` consistently, so the put/get/delete triple stays internally coherent. `mutation_macs_round_trip_raw_bytes` exercises all three with byte values (NUL and high bytes) that would diverge under JSON re-encoding, which is the right input to pin the regression. It passes locally.

No schema change is involved (BLOB columns are unchanged), so this correctly stays within the V3 schema. Pre-fix rows were corrupt by definition (app-state sync was the failing behavior this fixes), so there is nothing to migrate; after upgrade WhatsApp re-fetches app-state and writes correct raw rows. One cosmetic note, not blocking: stale JSON-wrapped rows from a prior pairing are left in place rather than purged, which is harmless given the (name, index_mac, device_id) primary key.

The two "Update whatsapp_storage.rs" commit subjects should be squashed into one conventional message at merge. Approving.

Audacity88 (approved):
@danielO99 I reviewed the current head e2ec40e9ad821f770ee2550296b9a4183925b516 against the live PR state, full diff, public source excerpts for the base `whatsapp_storage.rs` and pinned `whatsapp-rust` store contract, #7806, singlerider's approval, and the green GitHub checks. I did not rerun local Cargo validation for this review-only pass. Approving the code change from my side.

### 🟢 What looks good — Raw MAC storage now matches the app-state contract

The fix is nicely scoped to the three `RusqliteStore` mutation-MAC methods. The table already stores `index_mac` and `value_mac` as BLOB columns, `AppStateMutationMAC` carries both fields as `Vec<u8>`, and the app-state sync path feeds `get_mutation_mac` results directly back into the LTHash update. Dropping the JSON wrapping on put/get/delete therefore matches the data contract instead of layering a serialization format into bytes that are supposed to be hashed verbatim.

The new round-trip test also hits the right failure boundary for this PR. It writes MACs containing NUL and high-byte values, reads the value back, checks an unknown key, and verifies delete behavior. Those inputs would not be equivalent under JSON wrapping, so the test should catch the regression this patch is meant to prevent.

### 🟡 Warning — Finish freshness and public-record cleanup before merge

This approval is not a merge-ready statement yet. The branch is 59 commits behind current `master`; the freshness check has no same-file overlap, but it still triggers `behind_by_10_or_more` and `risky_surface_behind` because this touches `crates/zeroclaw-channels/src/whatsapp_storage.rs`. Before merge, please either update the branch and wait for fresh CI, run an exact merge-result smoke, or have a maintainer explicitly accept the stale-check risk.

The public PR body also needs cleanup before merge. Its label snapshot says `type: fix`, `risk: low`, `size: S`, while the live labels are `bug`, `size: XS`, `risk: medium`, `channel`, and `channel:whatsapp`; the body also stops before the required Security & Privacy, Compatibility, and Rollback sections. That is public-record cleanup, not a code blocker for this review.

PR #7919, opened 2026-06-18
feat(web): config-alias rename + delete cascade preview

Nillth:
## Summary

The alias CRUD + cascade work (#7785→#7842 stack) gave the CLI and gateway full delete-with-cascade and rename-with-reference-rewrite for agents/providers/channels — but the **web config UI never exposed rename**, and delete had no cascade preview. This wires both into the config alias lists.

- **Rename** — a pencil action on alias rows opens an inline edit → the existing `renameMapKey` (`POST /api/config/rename-map-key`, which rewrites every reference). Agent-rename `warnings[]` (owned stores that didn't follow the rename) surface to the operator; renaming also busts the command-palette search cache.
- **Delete cascade preview** — clicking trash now opens a **dry-run preview** before committing: HARD references that *block* the delete, SOFT references that get *scrubbed*, the agent owned-state cascade note, and the live-ACP-session count. The Delete button appears only when the plan is allowed. Replaces the old blind two-click confirm.
- **New endpoint** `GET /api/config/delete-plan?path=&key=` returns that dry-run impact. It reuses the **same** `alias_refs::plan_delete` walk the real delete refuses on (no client-side re-derivation of the HARD/SOFT split) plus the agent live-ACP gate. It is read-only and `require_auth`-gated, and returns config paths + reference text only — no secret values.

**Scope:** the config alias-list UI + one read-only gateway route. No change to the delete/rename mutation paths (those shipped in #7785→#7842), the cascade logic, or any other surface.

Related #7175 (typed delete-with-cascade) and #7468 (alias rename) — both shipped server-side via the #7842 stack; this is the web surface for them.

## Validation Evidence

- `cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings` → no warnings
- `cargo test -p zeroclaw-gateway api_config` → `ok. 27 passed; 0 failed`
- `cd web && npm run build` (tsc -b && vite build) → success

Beyond CI: ran the daemon against a throwaway config and confirmed the endpoint and UI end-to-end — `delete-plan` returns the HARD blocker (`heartbeat.agent`) for a heartbeat-pinned agent (`allowed:false`) and the SOFT scrub (`agents.<a>.delegates[i]`) for a delegate reference, matching the `agents delete --dry-run` CLI output exactly (same `plan_delete`). The preview's `blocked` state is keyed on `!plan.allowed`, so the agent live-ACP gate blocks even when there are no config references. Non-aliased sections short-circuit to "no cascade".

image

When renaming objects
image

image

Flows through to the other sections
image

## Security & Privacy Impact

- No new permissions, capabilities, or filesystem access.
- No new external network calls.
- No change to secrets handling. The new endpoint is `require_auth`-gated like the rest of `/api/config`, is **read-only**, and returns only config paths + reference text (no secret values).
- No PII in the diff.

## Compatibility

Backward compatible. The endpoint is additive; `renameMapKey`/`deleteMapKey` clients already existed. The delete UX changes from two-click to preview-then-confirm — no API or behavior break. No config/env/CLI surface change; no upgrade steps.

## Rollback

`git revert` — web + one read-only route; no state or schema migration.


singlerider (approved):
Reviewed at 5a0b2707 on a fresh checkout of `feat/web-alias-crud`. Built the gateway crate, ran the config tests, ran the web build, and cross-checked the new endpoint against `alias_refs`.

### 🟢 The delete-plan endpoint is correctly read-only, fail-closed, and leak-free

`handle_delete_plan` gates on `require_auth` first, computes the impact from `alias_refs::plan_delete` (the same walk the real delete refuses on — no client-side re-derivation of the HARD/SOFT split), and composes `allowed = plan.allowed && (!is_agent || live_acp == Some(0))`. The live-ACP path fails closed exactly as claimed: `live_acp_session_count(...).ok()` turns a counting error into `None`, and `None != Some(0)`, so the plan reports not-allowed. That mirrors the real delete's refusal gate rather than re-implementing it.

I confirmed the privacy claim end to end: `RefSiteDto.raw_value` is sourced from `alias_refs::RefSite.raw_value`, which is only ever the stored *reference text* at a reference site (e.g. `anthropic.default`), never a credential. The endpoint returns config paths plus that reference text and nothing else, and is `require_auth`-gated like the rest of `/api/config`. Non-aliased sections short-circuit to an empty, allowed plan.

Local validation on 5a0b2707:

```
cargo test -p zeroclaw-gateway --lib api_config  →  27 passed; 0 failed
cargo clippy -p zeroclaw-gateway --all-targets   →  clean
cd web && npm run build (tsc -b && vite build)   →  built (chunk-size warning is pre-existing)
```

### 🟢 Worth a follow-up: the new handler has no direct test

The underlying `plan_delete` and `live_acp_session_count` are covered, but `handle_delete_plan` itself has none — specifically the `allowed` composition and the fail-closed-on-count-error branch, which are the security-relevant lines this PR adds. A small handler-level test (allowed plan, HARD-blocker plan, agent with a live ACP session, non-aliased short-circuit) would pin the gate so a future refactor can't silently flip `allowed` open. Not blocking given the delegated logic is tested, but it would close the one coverage gap on the new surface.

This is ready to merge from my side once the test note is either addressed or consciously deferred. Nice that the web surface reuses the exact server-side cascade walk instead of re-deriving it.

PR #7920, opened 2026-06-18
fix(tools): resolve relative working_directory against workspace root

### LINKED ISSUE #7877: [Bug]: external coding tools resolve relative working_directory from daemon cwd
## Affected component

tools

## Severity

S2 - degraded behavior

## Current behavior

The external coding tools that accept `working_directory` canonicalize a relative value directly from the ZeroClaw process current directory. Current source shows this pattern in `claude_code`, `claude_code_runner`, `codex_cli`, and `gemini_cli`: each builds `PathBuf::from(wd)` and calls `canonicalize()` before comparing it to `security.workspace_dir`.

That means `working_directory = "."` or `working_directory = "src"` can resolve against the daemon's launch directory instead of the configured workspace. If the daemon was started outside the workspace, the tool can fail with "does not exist or is not accessible" or reject the path as outside the workspace even though the caller meant a workspace-relative path.

Related stale PR:

- #5416

That PR also fixed the old `codex -q` invocation, which current master no longer appears to use. The remaining gap is the workspace-relative path handling.

## Expected behavior

For tools with a `working_directory` argument:

- Relative values should resolve against `security.workspace_dir`.
- Absolute values should keep the current containment check.
- Canonicalization should still reject non-existent paths, symlink escapes, and paths outside the workspace.

## Steps to reproduce

```bash
1. Configure an agent workspace at a known project directory.
2. Start the ZeroClaw daemon or agent from a different process current directory.
3. Enable one of the external coding tools, such as codex_cli.
4. Invoke the tool with working_directory = "." or working_directory = "src".
5. Observe that the path is resolved relative to the process cwd rather than the configured workspace root.
```

## Impact

Affected users: operators using external CLI delegation tools from daemon/service launches where the process cwd is not the agent workspace.

Frequency: likely consistent when a relative `working_directory` is supplied and the daemon cwd differs from `security.workspace_dir`.

Consequence: workspace-relative delegation fails or points at the wrong existing directory before the containment check rejects it.

## Logs / stack traces

No runtime log excerpt is included. This issue comes from stale-PR audit evidence plus current source inspection of the `working_directory` handling in the CLI delegation tools.

## ZeroClaw version

Current `master` as of 2026-06-18.

## Rust version

Not applicable for this source-level routing issue.

## Operating system

All supported operating systems where the ZeroClaw process cwd can differ from the configured workspace.

## Regression?

Unknown

## Pre-flight checks

- [x] I checked this against current master source; a runtime smoke should be added before implementation.

ZOOWH:
## Summary

External coding tools (`claude_code`, `claude_code_runner`, `codex_cli`, `gemini_cli`) accepted a `working_directory` argument but resolved relative paths against the daemon's current directory instead of the configured workspace root. When the daemon was started outside the workspace (e.g. from `/`), relative paths like `"."` or `"src"` would fail with "does not exist" or be rejected as outside the workspace.

This patch adds an `is_relative()` check before canonicalization: if the `working_directory` is relative, it is joined with `security.workspace_dir` first, so the resulting path resolves correctly within the workspace boundary. Absolute paths keep their current behavior unchanged.

Additionally:
1. **Empty string rejection**: `PathBuf::from("").is_relative()` returns `true`, causing `workspace.join("")` → workspace itself, which is semantically ambiguous. All 5 coding tools now reject empty/whitespace-only `working_directory` values with a clear error message. Users should use `"."` or omit the argument to get workspace root.
2. **opencode_cli fix**: The initial PR covered 4 tools but missed `opencode_cli.rs`, which has the identical working_directory validation pattern and the same #7877 bug. Now patched.

Closes #7877

## Changes

- In each of the **five** coding tool implementations (4 original + `opencode_cli`), insert a relative-path guard before `canonicalize()`:
  - If the provided `working_directory` is relative, join it with `security.workspace_dir` first.
  - Absolute paths pass through unchanged.
- Add empty/whitespace rejection: if `working_directory` is empty or whitespace-only, return an error instead of silently resolving to workspace root.

## Affected files

- `crates/zeroclaw-tools/src/claude_code.rs`
- `crates/zeroclaw-tools/src/claude_code_runner.rs`
- `crates/zeroclaw-tools/src/codex_cli.rs`
- `crates/zeroclaw-tools/src/gemini_cli.rs`
- `crates/zeroclaw-tools/src/opencode_cli.rs` (newly fixed)

## Test plan

- [ ] Start the ZeroClaw daemon from a directory other than the configured workspace (e.g. `cd / && zeroclawd`).
- [ ] Enable an external coding tool (e.g. `codex_cli`).
- [ ] Invoke the tool with `working_directory = "."` — should resolve to the workspace root, not `/`.
- [ ] Invoke with `working_directory = "src"` — should resolve to `<workspace>/src`.
- [ ] Invoke with an absolute path — should behave identically to before this change.
- [ ] Verify that paths outside the workspace are still rejected by the containment check.
- [ ] Invoke with `working_directory = ""` — should return an error ("working_directory must not be empty").
- [ ] Repeat the above for `opencode_cli` (previously missed).

### DIFF
--- a/crates/zeroclaw-tools/src/claude_code.rs
+++ b/crates/zeroclaw-tools/src/claude_code.rs
@@ -131,6 +131,13 @@ impl Tool for ClaudeCodeTool {
         // could bypass the workspace containment check via symlinks or
         // specially-crafted path components).
         let work_dir = if let Some(wd) = args.get("working_directory").and_then(|v| v.as_str()) {
+            if wd.trim().is_empty() {
+                return Ok(ToolResult {
+                    success: false,
+                    output: String::new(),
+                    error: Some("working_directory must not be empty".into()),
+                });
+            }
             let mut wd_path = std::path::PathBuf::from(wd);
             let workspace = &self.security.workspace_dir;
             // Resolve relative working_directory against the workspace
--- a/crates/zeroclaw-tools/src/claude_code.rs
+++ b/crates/zeroclaw-tools/src/claude_code.rs
@@ -131,8 +131,13 @@ impl Tool for ClaudeCodeTool {
         // could bypass the workspace containment check via symlinks or
         // specially-crafted path components).
         let work_dir = if let Some(wd) = args.get("working_directory").and_then(|v| v.as_str()) {
-            let wd_path = std::path::PathBuf::from(wd);
+            let mut wd_path = std::path::PathBuf::from(wd);
             let workspace = &self.security.workspace_dir;
+            // Resolve relative working_directory against the workspace
+            // root instead of the daemon's current directory (#7877).
+            if wd_path.is_relative() {
+                wd_path = workspace.join(&wd_path);
+            }
             let canonical_wd = match wd_path.canonicalize() {
                 Ok(p) => p,
                 Err(_) => {
--- a/crates/zeroclaw-tools/src/claude_code_runner.rs
+++ b/crates/zeroclaw-tools/src/claude_code_runner.rs
@@ -134,6 +134,13 @@ impl Tool for ClaudeCodeRunnerTool {
 
         // Validate working directory
         let work_dir = if let Some(wd) = args.get("working_directory").and_then(|v| v.as_str()) {
+            if wd.trim().is_empty() {
+                return Ok(ToolResult {
+                    success: false,
+                    output: String::new(),
+                    error: Some("working_directory must not be empty".into()),
+                });
+            }
             let mut wd_path = std::path::PathBuf::from(wd);
             let workspace = &self.security.workspace_dir;
             // Resolve relative working_directory against the workspace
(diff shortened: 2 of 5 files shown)


ZOOWH:
Pushed two follow-up fixes:

1. **opencode_cli.rs missing the is_relative() guard** — The initial PR covered claude_code, claude_code_runner, codex_cli, and gemini_cli, but `opencode_cli.rs` has the identical working_directory validation pattern and was not patched. This is the same #7877 bug. Now fixed.

2. **Empty/whitespace working_directory rejection** — `PathBuf::from("").is_relative()` returns `true`, causing `workspace.join("")` to silently resolve to workspace root. This is semantically ambiguous (empty string should not equal workspace root — use "." or omit the argument instead). All 5 coding tools now reject empty/whitespace-only values with a clear error message.

Verified: `cargo check -p zeroclaw-tools` passes, relevant tests pass. The 6 stash-related test failures are pre-existing (require git repo context) and unrelated to this change.

Audacity88:
Closing this as superseded by #7967 for #7877.

Both PRs target #7877 and patch the same five external-coding-tool adapters. #7967 is the narrower path we are advancing: it stays scoped to the relative `working_directory` resolution bug and includes representative regression coverage for that behavior.

The extra empty/whitespace `working_directory` rejection in this PR may still be worth considering separately. This is a separate behavior contract from the original #7877 bug, so please feel free to open a focused follow-up PR for that behavior with its own acceptance framing and regression coverage.

Thanks for pushing this forward.

PR #7922, opened 2026-06-18
feat(channels/discord): slash command localizations + guild scope

Nillth:
## What this does

Completes the Discord slash-command surface to parity (the last two clusters
after #7844's typed options + chunked followups): **per-locale command
descriptions** and **guild-scoped registration**. Three self-contained commits:

- **Guild scope** — a new `slash_command_scope: { global | guild }` on
  `DiscordConfig` (default `global`). Under `guild`, the skill commands register
  to each `guild_ids` entry via the guild command endpoint (instant
  propagation) instead of the global endpoint (~1 h). `reconcile_slash_commands`
  becomes a scope dispatcher; the *inactive* scope has the channel's
  `/ask` + skill commands reaped, so flipping the scope never leaves a command
  registered in both places. The reconcile fingerprint folds the scope + guild
  set, so a scope switch with byte-identical command bodies still forces one
  reconcile instead of being silently skipped. Empty `guild_ids` warns and falls
  back to global.

- **Built-in localizations** — `/ask`, its `prompt` option, and the default
  skill `input` option carry compiled-in `description_localizations`
  (`es-ES`/`fr`/`ja`/`zh-CN`). Compiled in rather than sourced from the i18n FTL
  machinery (which only compiles in `en`/`zh-CN`; the others load from a runtime
  locale dir that may be absent), so a stock binary always renders them. The
  reconcile listing GET now passes `with_localizations=true` and
  `command_projection` folds in `description_localizations`, normalising
  Discord's echoed `null` to the omitted key — so an un-localized command stays
  a reconcile no-op while a translation change forces exactly one
  re-registration (never a per-READY re-register loop). *Translations are
  initial and open to native-speaker refinement.*

- **Skill-authored localizations** — skills localize their own command + option
  descriptions via `description_localizations` tables in `SKILL.toml`
  (`[skill]` and `[[skill.slash_options]]`). Locale codes are validated against
  Discord's supported set before registration: an authoring typo drops just that
  entry with a WARN rather than 400-ing the whole command (and wedging the
  reconcile in a retry loop), and each value is truncated to Discord's 100-char
  limit.

## ⚠️ For maintainer sounding

The `description_localizations` **manifest schema** (`SKILL.toml`) is a new
public surface, in the same vein as #7844's `[[skill.slash_options]]`. Flagging
it explicitly for a maintainer read on the shape before it sets a precedent.

## Testing

- `cargo test -p zeroclaw-channels --features channel-discord` (incl. new tests:
  guild endpoint routing; cross-scope reaping on switch; built-in + skill-author
  localizations flow-through with an invalid locale dropped; the projection
  no-op round-trip that proves no re-register loop).
- `cargo test -p zeroclaw-runtime` (SKILL.toml `description_localizations` parse,
  command + option level).
- `cargo clippy --all-targets -D warnings`, `cargo fmt --all --check`,
  `cargo build --workspace --all-targets` all clean.

Part of #7831.


crates/zeroclaw-channels/src/discord/slash.rs
```diff
@@ -426,14 +553,133 @@ pub(crate) async fn reconcile_slash_commands(
...
+            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                .with_attrs(::serde_json::json!({"status": resp.status().as_u16()})),
+            "inactive-scope command listing failed; skipping cross-scope cleanup"
+        );
+        return Ok(ReconcileOutcome::Reconciled);
+    }
+    let existing: Vec<serde_json::Value> = resp.json().await?;
+    for cmd in &existing {
+        let name = cmd.get("name").and_then(|n| n.as_str()).unwrap_or("");
+        if name != "ask" && !is_skill_command_shape(cmd) {
```
Audacity88: This deletes any inactive-scope command named `ask`, regardless of whether ZeroClaw created it. The active stale-command path avoids foreign deletes by requiring `is_skill_command_shape`, and the module docs say foreign commands must never be touched. A scope flip with an existing global or guild `/ask` from another integration would issue a DELETE here. Please add an ownership check for `/ask` too, for example by comparing against ZeroClaw's expected `/ask` projection or by persisting registered command ids from the prior successful reconcile, before reaping it.

Audacity88 (changes requested):
Reviewed PR #7922 at 3306ac1 against the public diff, current PR metadata/reviews/comments, passing CI checks, and public source context for the Discord slash-command ownership contract and `SKILL.toml` scaffold. I focused on slash registration/reconcile behavior, cross-scope cleanup safety, and the new manifest/config surface.

Thanks for the careful split here. The `with_localizations=true` projection path and guild endpoint dispatch have good unit coverage in the diff, and the CI gate is green now. I’m requesting changes for one destructive cleanup edge before this can land.

### 🔴 Blocking — Inactive-scope cleanup can delete a foreign `/ask`

`reap_all_owned_commands` deletes any command named `ask` from the inactive endpoint, while skill commands still require the safer ownership-shape marker. That violates the module’s own foreign-command safety contract: an operator switching `slash_command_scope` with the same Discord application could have a global or guild `/ask` owned by another integration, and READY would delete it even though ZeroClaw did not register it. Please make `/ask` use an ownership check too, or use persisted registered-command identity before doing cross-scope deletion.

### 🟡 Warning — Document the new `SKILL.toml` localization surface

The PR intentionally adds `description_localizations` as a public skill-manifest key under `[skill]` and `[[skill.slash_options]]`, but the author-facing `SKILL.toml` example/scaffold still only shows the older fields. Since this sets a manifest precedent, please add a minimal docs or generated-template note, or explicitly mark it as maintainer-accepted prototype surface before merge.

WareWolf-MoonWall (commented):
Reviewed PR #7922 at 3306ac1d against the full diff, @Audacity88's CHANGES_REQUESTED review and inline thread, CI status, the Discord slash-command ownership contract, and the `SKILL.toml` scaffold.

@Audacity88 holds an active CHANGES_REQUESTED. Per the review protocol, I will not approve over that; this is a comment. I'm adding perspective on their findings and noting a few additional observations.

---

### 🔴 Blocking — Audacity88's finding stands: `/ask` cross-scope deletion lacks ownership check

@Audacity88's inline comment at `crates/zeroclaw-channels/src/discord/slash.rs:626` is correct and unresolved. The `reap_all_owned_commands` function applies asymmetric ownership logic: skill commands are guarded by `is_skill_command_shape` before deletion; `/ask` is deleted by name alone:

```rust
if name != "ask" && !is_skill_command_shape(cmd) {
    continue;
}
```

The condition reads: delete anything named `ask` unconditionally, or anything that passes the skill-shape check. An operator who switches `slash_command_scope` while another integration has registered its own global or guild `/ask` command would have that foreign command deleted on the next READY event, even though ZeroClaw never created it. This violates the module's own foreign-command safety contract.

The fix @Audacity88 describes is correct: apply an ownership check to `/ask` before deleting it — either by comparing the command against ZeroClaw's expected `/ask` projection (shape + options + localizations), or by persisting the command IDs from the prior successful reconcile and only deleting IDs that were in that set. The existing `scope_switch_reaps_owned_commands_from_the_inactive_scope` test exercises the happy path but does not prove the foreign-`/ask` safety, because the test fixture constructs the stale `/ask` to look exactly like ZeroClaw's. A test with a foreign `/ask` that differs in description or options would be the right regression guard here.

---

### 🟡 Warning — `SKILL.toml` localization surface is new public API with no scaffold documentation

@Audacity88 also flagged this. `description_localizations` under `[skill]` and `[[skill.slash_options]]` is now a skill-authoring key that will appear in the wild once this lands. The `SKILL.toml` template/scaffold does not show it. Authors who discover this capability through the source or through a Discord command that shows a localized description will have no canonical example to follow, and typos in locale codes will silently drop entries (by design, but that behavior should be documented). A minimal scaffold update or a note in the skills reference doc is the right delivery vehicle.

---

### 🔵 Suggestion — Built-in Discord localizations bypass FTL; worth a follow-up issue

The hardcoded `es-ES / fr / ja / zh-CN` translations for `/ask`, `prompt`, and the default `input` option are stored as Rust string literals compiled into the binary. The PR explains why: the Fluent machinery only compiles `en` and `zh-CN` at build time; the remaining locales require a runtime locale directory that may be absent in a stock install. I understand the reasoning and don't have a better answer for today, but this creates a split: tool descriptions flow through FTL and can be updated via locale files; Discord command descriptions require a binary release to update. A tracked follow-up issue to align these once the locale-dir availability story improves would be worth filing, so the intent is on the record and doesn't get lost.

---

### 🟢 What looks good — Guild/global fingerprint folding prevents silent scope drift

The fingerprint computation folds both the scope discriminant and the guild ID set into the hash:

```rust
match effective_scope { SlashScope::Global => 0u8, SlashScope::Guild => 1u8 }.hash(&mut h);
guild_ids.hash(&mut h);
```

This guarantees that switching `slash_command_scope` — even with byte-identical command bodies — forces exactly one reconcile and never silently skips it. That's the right invariant. The comment explaining the migration hazard makes the intent clear.

### 🟢 What looks good — Locale code validation: warn-and-drop is the right failure mode

Validating skill-authored locale codes against Discord's supported set at registration time and dropping invalid entries with a WARN (rather than failing the entire command registration or silently passing an invalid code to the API) is exactly the right tradeoff for a user-facing manifest surface. An authoring typo should never wedge the reconcile loop in a 400 retry cycle.

### 🟢 What looks good — Test coverage for new behavior paths is thorough

`guild_scope_registers_to_the_guild_endpoint` proves endpoint routing under `scope=Guild`. `scope_switch_reaps_owned_commands_from_the_inactive_scope` covers the migration path. The `with_localizations=true` no-op round-trip test proves the projection normalization prevents spurious re-registers on an un-localized command. The `SKILL.toml` parse test at command and option level verifies the new manifest surface end-to-end. This is solid test discipline for a registration path where bugs are immediately visible to Discord users.

---

**Verdict**: commenting rather than approving — @Audacity88's CHANGES_REQUESTED on the `/ask` ownership gap is unresolved and I agree with their analysis. Once that's addressed and the scaffold documentation is updated, I don't have additional blockers on my end.

singlerider (commented):
Reviewed at 3306ac1d on a fresh checkout of `feat/discord-slash-localizations-scope`. The head commit since @Audacity88's review is a clippy-only rename of `SlashCommandScope::default()`; it does not touch the cross-scope cleanup path, so I looked at that path directly on its current state.

### 🔴 Blocking — `reap_all_owned_commands` still deletes a foreign `/ask` by name

`crates/zeroclaw-channels/src/discord/slash.rs:626`:

```rust
if name != "ask" && !is_skill_command_shape(cmd) {
    continue;
}
```

On a scope switch this reaps any inactive-endpoint command that is either named `ask` or carries the skill ownership shape. Skill commands are correctly gated by `is_skill_command_shape`, but `/ask` is deleted on name alone. If the same Discord application has a global or guild `/ask` registered by a different integration, READY will delete it even though ZeroClaw never registered that command. That is the exact foreign-command-safety violation @Audacity88 flagged, and the head commit does not address it.

The reserved-name seeding at `slash.rs:66` (`seen.insert("ask")`) establishes that `/ask` is ZeroClaw's *intended* built-in, but intent on this side is not proof of ownership on a shared application. The fix is to require the same ownership signal for `/ask` that skill commands use, or to delete by persisted registered-command identity rather than by name. I agree this should block the merge.

### 🟢 The localization and guild-scope split is well-built

The P4a/P4b/P4c split is clean: `with_localizations=true` projection and the guild endpoint dispatch have real unit coverage, and the skill-authored `description_localizations` surface threads through `SKILL.toml` -> spec -> wire consistently. Local validation on 3306ac1d:

```
cargo test -p zeroclaw-channels --lib discord::slash  →  17 passed; 0 failed
```

The reserved-name and skill-shape determinism (sorted slugs, capped set, dropped-name logging) is solid. Once the `/ask` cleanup uses an ownership check, this is in good shape.

@Audacity88's `CHANGES_REQUESTED` remains the gating review and is still accurate; I'm posting this as a concurring comment rather than stacking a second formal block.

Nillth:
Thanks all — good catch, and you're right that the by-name `/ask` deletion broke the
module's foreign-command-safety contract. Fixed.

**Ownership check on `/ask`.** The inactive-scope reaper no longer deletes a `/ask`
by name. It now only reaps a `/ask` whose *projection* matches the `/ask` we ourselves
register — the same `command_projection` (description + options + per-option
type/required/description + `description_localizations`) already used for change
detection elsewhere in the file. Concretely, `reconcile_slash_commands` passes the
canonical `/ask` from the desired registration body down to `reap_all_owned_commands`,
which gates the delete on:

```rust
let is_owned_ask = name == "ask"
    && expected_ask.is_some_and(|a| command_projection(cmd) == command_projection(a));
if !is_owned_ask && !is_skill_command_shape(cmd) { continue; }
```

So a `/ask` from another integration sharing the same application/scope — different
description, different options — never matches and is left untouched. Skill commands
keep their existing `is_skill_command_shape` ownership marker; nothing there changed.

One supporting change: the reaper's listing GET now sets `with_localizations=true`.
Without it Discord returns the `*_localizations` dictionaries as `null`, and our own
localized `/ask` would fail its own projection match (and never get reaped on a scope
switch). With it, our command echoes back apples-to-apples.

**Regression test.** Added `scope_switch_spares_foreign_ask_in_inactive_scope`
(`crates/zeroclaw-channels/src/discord/mod.rs`). It mounts a foreign `/ask` on the
now-inactive global scope that differs from ours (description "Ask a DIFFERENT bot",
an option named `prompt` rather than our shape), alongside one of our own skill
commands. The DELETE for the foreign `/ask` id is mounted with `.expect(0)` — so the
test fails on drop if a delete is ever issued for it — while the skill command's
DELETE is `.expect(1)`. This proves foreign-safety in a way the existing
`scope_switch_reaps_owned_commands_from_the_inactive_scope` test could not: that one
constructs its stale `/ask` *from our own registration body*, so it only ever proved we
reap our own. (I also re-derived that test's stale `/ask` from the registration body so
it carries our localizations and the projection still matches under
`with_localizations=true`.)

**Two related hardening fixes while in here** (same review thread):
- `reap_all_owned_commands` previously let an unparseable 200 listing body propagate
  out and abort the active-scope registration, violating its documented "never fatal"
  contract. It now logs and skips cross-scope cleanup on a body-decode error, same as a
  failed listing status, and scrubs the URL on the json() error.
- `localizations_object` now clamps built-in translations to Discord's 100-char limit
  (the skill-authored path already clamped); an overlong built-in would otherwise 400
  the registration and wedge the reconcile in a retry loop.

**Docs (the non-blocking WARNING).** Documented the `SKILL.toml` localization surface
in the skills reference (`docs/book/src/tools/skills.md`), under a new "Slash command
options and localizations" section: `description_localizations` under `[skill]` and
under each `[[skill.slash_options]]`, keyed by locale code, with a minimal SKILL.toml
example. It also notes that invalid/unsupported locale codes are warn-and-dropped
rather than failing registration, so a typo can't wedge command registration. I looked
for a generated SKILL.toml scaffold to extend instead, but there's no scaffold that
emits a `slash_options`/localization template — the skills reference doc is the
canonical place this surface is described, so that's where it went.

Build, `cargo fmt`, `clippy --all-targets`, and the `discord::slash` test suite are all
green.

Audacity88 (approved):
Reviewed current head `f1a14c4` against the live PR metadata, prior review thread, green GitHub checks, and the changed Discord slash-command, skill-manifest, config, and docs paths.

Thanks for the careful follow-up. This resolves my prior block.

### ✅ Resolved — Inactive-scope cleanup now proves `/ask` ownership

`reap_all_owned_commands` no longer deletes every inactive-scope command named `ask`. It now compares the listed `/ask` projection against the canonical `/ask` that ZeroClaw registers, with `with_localizations=true` on the listing so the localization dictionaries participate in that comparison. A foreign `/ask` with a different description/options shape is left alone, while ZeroClaw's own stale `/ask` and legacy skill-shaped commands are still reaped.

The new regression test covers the destructive case directly: `scope_switch_spares_foreign_ask_in_inactive_scope` mounts a foreign `/ask` delete with `expect(0)` and still expects the owned skill command delete. That is the guard I wanted before clearing this.

### ✅ Resolved — The `SKILL.toml` localization surface is documented

The skills reference now documents `description_localizations` under `[skill]` and `[[skill.slash_options]]`, shows a minimal example, and says unsupported locale codes are warn-and-dropped rather than failing registration. That covers the public-surface warning from my earlier review.

### 🟢 What looks good — The reconcile path stays steady-state friendly

The active reconcile path also lists commands with `with_localizations=true`, and `command_projection` folds command and option localizations into the no-op comparison. That should avoid a localized command being re-registered on every READY event while still forcing a one-time update when the configured localization text changes.

I did not rerun local validation; I relied on source review plus the current green GitHub checks and the validation reported in the PR body. This approval clears my prior `CHANGES_REQUESTED`; I am not making a merge-ready call here.

Issue #7314, opened 2026-06-06
[Tracker]: WASM plugin program


Audacity88:
## Summary

This tracker preserves the deferred coordination map for broad WASM/plugin-program work: architecture, WIT/component-model direction, plugin host/runtime boundaries, lifecycle hooks, host-function security, hardware access, and program documentation.

The tracker and its remaining broad RFCs are in Icebox with no release commitment. The finite first delivery wave is tracked separately by #8850 under `Plugins: Capability Catalog & Lifecycle`.

## Release Decision

The broad WASM/plugin program no longer ships as part of v0.8.3. The completed foundation remains landed; unresolved later phases stay visible here in Icebox until maintainers define a finite follow-on milestone.

Historical program ownership is currently unassigned. The finite delivery wave is coordinated through #8850. New plugin work does not enter this broad tracker automatically; it must fit an approved finite milestone or remain independently triaged.

### Moved Out Of v0.8.3

- [x] #7822 — RFC; moved to Icebox — WASM plugin lifecycle hook subscriptions
- [x] #8135 — RFC; moved to Icebox — default-on loading, capability enforcement, and signed distribution
- [x] #8187 — RFC; moved to Icebox — capability-gated WASI hardware host functions
- [x] #7846 — PR; moved to v0.8.4 — before_llm_call hook call-site feature

### Related trackers and decision surfaces

- #7320 — v0.8.3 release closeout index
- #8850 — finite plugin capability catalog and lifecycle wave
- #6943

## Implementation Phases

Runtime decision: wasmtime component-model (WIT); extism retired. This phase map is historical and deferred unless a row is promoted into a finite milestone. Each linked artifact remains the source of truth for its own live state.

#### Capability and signing split
- [x] #8135

#### FND-001 deconflict
- [x] #6943

#### WIT/component-model host base
- [x] #8368 — merged as e367c7331e; wasmtime component-model trunk for tool/channel/memory worlds
- [x] #7928 — closed as superseded by #8368; SDK/proof ideas donated forward
- [x] #8144 — closed as superseded by #8368; runtime/policy pieces donated forward

#### Runtime, collision, limits, secret model
Follow-up rows should be added here when the donated runtime, collision, limit, signature, and secret-model slices are split on top of #8368.

[...]

Issue #6943, opened 2026-05-26
[RFC]: Deconflict Plugin System Goals in FND-001


bheatwole:
### Summary

Replace Extism with a direct wasmtime component model host and set the plugin build target to `wasm-wasip2`

### Problem statement

FND-001 contains several mutually exclusive commitments across its sections. A contributor implementing Phase 2 D2 — wiring Extism as directed — will simultaneously violate three other commitments in the same document. There is no implementation path that satisfies all of the following at once:

- Commits to WIT as the plugin interface standard (§5.2)
- Directs Phase 2 to wire **Extism** as the execution backend (§6 Phase 2 D2)
- Specifies `wasm32-wasip1` as the plugin build target in the release artifact matrix (§4.4.2)
- Declares `plugins-wasm` as **always-on** in every kernel binary (§4.4.2)
- Promises ZeroClaw runs on "any hardware — from a $10 embedded board" (§2)

The conflicts are not implementation details — they are architectural incompatibilities that will block progress in Phase 2 if not resolved before work begins. This RFC names each conflict precisely and recommends the single path that satisfies all of them with the fewest amendments to FND-001.

---

### Conflict 1: Extism vs. "Run on Any Hardware"

FND-001 §2 is unambiguous:

> ZeroClaw is a personal AI assistant runtime that any person can run on **any hardware — from a $10 embedded board to a cloud server**.

FND-001 §4.4.2 declares `plugins-wasm` always-on:

> **Always-on** | `plugins-wasm`, `skill-creation` | Compiled into every kernel binary unconditionally. `plugins-wasm` is the kernel's core mechanism.

FND-001 §6 Phase 2 D2 directs:

> Wire **Extism** into `WasmTool::execute` and `WasmChannel`.

These three statements together are a contradiction. Extism inherits wasmtime's full default feature set, which unconditionally includes `cranelift`. Cranelift has no 32-bit ARM backend. An `armv6-unknown-linux-gnueabihf` build (Raspberry Pi Zero) fails to compile when any crate in the dependency graph activates the `cranelift` feature — and there is no mechanism within the Extism crate to pass `default-features = false` to its wasmtime dependency. The binary size and platform support problem cannot be solved while Extism is in the dependency tree.

`plugins-wasm` cannot be always-on and the runtime cannot target every hardware platform as long as Extism is the execution backend.

### Conflict 2: Extism vs. WIT as the Plugin Interface Standard

FND-001 §5.2 commits to WIT:

[...]

PR #7928, opened 2026-06-18
feat(wasi): initial WASM component-model plugin host code

bheatwole:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Added WASM component-model plugin host code targeting wit v0 with Channel, Memory, and Tool traits (see `zeroclaw-plugins::component::v0`)
  - Added instantiation path for component-model plugins (see `zeroclaw-plugins::PluginStore::instantiate_channel_plugin` / `instantiate_tool_plugin` / `instantiate_memory_plugin`) which creates an `Arc<dyn Channel>` / tool / memory adapter for a named plugin. (`PluginHost` was renamed to `PluginStore` to avoid confusion with another type in the crate.)
  - Added `FineGrainedPermissions` to `PluginManifest` for tight control of what file paths, IP addresses, and URLs are available to the plugin. This first version is intentionally very restrictive to give the strongest security model from the beginning. Future PRs could give plugins ways to dynamically request other paths, IPs and URLs.
  - Removed three functions from `v0/channel.wit` whose function was already included in `get-channel-capabilities`
  - **New since initial review:** added the guest-side `zeroclaw-plugin-sdk` crate (`ToolPlugin`/`MemoryPlugin`/`ChannelPlugin` traits with capability-gated defaults, `export_tool!`/`export_memory!`/`export_channel!` macros) plus four real example plugins (`tool-echo`, `memory-noop`, `channel-echo`, `tool-tcp-probe`) compiled to `wasm32-wasip2` and round-tripped through the real, unmodified `PluginStore` in integration tests (`spike_tool.rs`, `spike_memory.rs`, `spike_channel.rs`, `spike_tcp_permission.rs`). This was added in response to review feedback that the host-side plumbing needed proof against a real component, not just mocks — and it surfaced two real host bugs fixed in this PR:
    - `instantiate_async` was required once a store had async-required imports wired in; the host was calling the sync/trappable variant and failing on any real component.
    - Memory/Channel's one-time `plugin-info`/`get-*-capabilities` probes during `from_bytes` had the same trappable-vs-async mismatch.
  - Centralized the proxy-aware `wasi:http` dispatch path's retry-on-429/5xx behavior, reusing `zeroclaw-infra::retry` (the same primitives native channels use) instead of giving plugin HTTP calls no resilience.
  - Fixed `FineGrainedPermission` TOML/JSON serialization and clarified HTTP vs. raw-TCP permission handling; `spike_tcp_permission.rs` now regression-tests that granting a plugin an HTTP permission does not also expose raw TCP/UDP access.
  - Hoisted code shared across future component-model WIT major versions out of `component::v0` into `component::` directly (`call_plugin.rs`, `plugin_store.rs`, `wrap_plugin.rs`), and split logging-adjacent helpers (`plugin_config.rs`, `plugin_linker.rs`) out of `v0/logging.rs` into more tightly scoped files.
  - **`zeroclaw-api::Channel::drop_self_messages` changed from a sync to an `async fn`.** This is a signature change to a trait method in the central `zeroclaw-api` crate (touches every `Channel` implementor in the workspace, not just plugins), needed because the plugin-side `drop-self-message` export is itself async over the component boundary and the previous sync trait method forced a `blocking_lock` (`call_plugin_sync!`) workaround that the spike/example work showed doesn't hold up once a real component is exercised. The only call site, `zeroclaw-channels/src/orchestrator/mod.rs`, was updated to `.await` it; no other crate overrides `drop_self_messages` (confirmed via grep), so all other channels pick up the new async default transparently and require no changes themselves. `call_plugin_sync!` and the `wrap_plugin_call_sync` helper it depended on are removed entirely now that nothing needs the sync/blocking path.
- **Scope boundary:**
  - Does not change core runtime policy enforcement in `zeroclaw-runtime`.
  - Does not change CLI command UX or top-level command routing.
  - Does not introduce new channel/provider features outside plugin-host integration points.
  - Still does not call the `instantiate_*` code path from `zeroclaw-runtime` or `zeroclaw-gateway` — wiring this into the live agent loop is out of scope for this PR. It is now exercised end-to-end, but only via the `zeroclaw-plugin-sdk` integration tests against real compiled `wasm32-wasip2` components, not from production runtime code.
  - Does not address plugin publishing or discovery.
  - Does not address plugin access to ENV VAR and configuration.
- **Blast radius:**
  - `crates/zeroclaw-plugins` runtime behavior and dependency graph.
  - New `crates/zeroclaw-plugin-sdk` crate (guest-side only; not linked into the host binary).
  - `crates/zeroclaw-api::Channel` trait signature (`drop_self_messages` is now `async`) — this is a central, widely-implemented trait. No implementor besides the trait default and the plugin adapter overrides this method (confirmed via repo-wide grep), and the one call site in `zeroclaw-channels`'s orchestrator was updated, but any out-of-tree `Channel` implementation overriding `drop_self_messages` would need to update its signature.
  - Any downstream users enabling plugin/WASM features in CI or release builds.
- **Linked issue(s):** Related #7314
- **Labels:** `type: feature`, `risk: high`, `size: XL`, `enhancement` `dependencies`, `plugins`, `wasm`

    ```text
    (clean, no output)
    ```
    ```text
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 55s
    ```
  - `cargo test`
    ```text
    test result: ok. 245 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.53s
    test result: ok. 298 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.37s
    test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
    test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.60s
    test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.16s
    test result: ok. 158 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.80s
    test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.09s
    ```
  - `cargo test -p zeroclaw-plugin-sdk --features tool`
    ```text
    Running tests/spike_tcp_permission.rs (target/debug/deps/spike_tcp_permission-74aa08250e8be3ad)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 15.97s
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 15.99s
    test http_only_permission_denies_raw_tcp_connect ... ok
    test tcp_permission_allows_raw_tcp_connect ... ok
    test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 19.21s

    Running tests/spike_tool.rs (target/debug/deps/spike_tool-93c35e0a41658a6d)
    test tool_echo_round_trips_through_plugin_host ... ok
    test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 14.08s
    ```
  - `cargo test -p zeroclaw-plugin-sdk --features channel`
    ```text
    Running tests/spike_channel.rs (target/debug/deps/spike_channel-52375d1ccbaec086)
    test channel_echo_round_trips_through_plugin_host ... ok
    test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 14.02s
    ```
  - `cargo test -p zeroclaw-plugin-sdk --features memory`
    ```text
    Running tests/spike_memory.rs (target/debug/deps/spike_memory-914a95a8b2e14f20)
    test memory_noop_round_trips_through_plugin_host ... ok
    test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 15.38s
    ```

  - cargo audit does not introduce any new vulnerabilities (6 pre-existing open ones)
  - Four real `wasm32-wasip2` example plugins, built from the SDK and loaded through the unmodified `PluginStore`, round-trip tool/memory/channel calls and capability-gated fallbacks (`spike_tool.rs`, `spike_memory.rs`, `spike_channel.rs`).
  - `spike_tcp_permission.rs` confirms an HTTP-only `FineGrainedPermission` grant does not give a plugin raw TCP/UDP reachability.
  - Not manually verified in this pass: end-to-end execution of third-party external WASM plugins against live, non-test network endpoints, or wiring through `zeroclaw-runtime`/`zeroclaw-gateway` (see Scope boundary).

- New external network calls? (`Yes`)
  - The plugin runtime now wires broader WASI-related capabilities (including network-oriented paths) for component plugins. Risk is unintended capability exposure in plugin execution contexts.
  - Mitigation is strict sandbox permission model that plugins must name each `FineGrainedPermission` in the `PluginManifest` file.
    - Files: Every host directory to which the plugin will have access must be explicitly named, including subdirectory access and file read/write permissions.
    - Http: Every URL to which the plugin wants to open a HTTP connection to must be named. A '*' is allowed in the pattern, but only at the third-level domain or higher: `*.example.com` is a valid permission, but `*.com` is not.
    - Raw TCP/UDP: Each IP or URL must be explicitly named. '*' based URLs are possible but require reverse DNS.
  - This PR adds a regression test (`spike_tcp_permission.rs`) directly verifying that an HTTP-only grant does not leak raw TCP/UDP reachability, closing a gap the original PR only asserted by design, not by test.
  - Mitigation: this is stronger security than the previous model that was a yes/no for full HTTP access and full File access.

## Compatibility (required)

  - Plugin authors must now add a "fine_grained_permissions" array to the plugin manifest to use any file or network resources.
  - `PluginHost` was renamed to `PluginStore`. No other in-tree crate referenced `PluginHost` (confirmed via `grep` across `zeroclaw-runtime`/`zeroclaw-gateway`), so this is not expected to break any existing caller, but flagging since it's a public type rename within `zeroclaw-plugins`.
  - `zeroclaw_api::Channel::drop_self_messages` changed from `fn` to `async fn`. Any out-of-tree code implementing the `Channel` trait and overriding this method must add `async`/`.await`. No in-tree implementor besides the trait default and the plugin component adapter overrides it, so no in-tree code outside this PR's own updated call site needed changes.

- **Fast rollback command/path:** `git revert <merge_commit_sha>`
- **Feature flags or config toggles:** Existing plugin/WASM build feature selection in Cargo features; no new runtime toggle introduced in this patch.
  - Plugin load/instantiation failures for component plugins.
  - Build failures around `wasmtime`/WASI or async fs feature resolution.
  - Runtime logs indicating component host call/linkage errors in plugin paths.


WareWolf-MoonWall (commented):
Reviewed at fcf89146. Twenty-seven commits adding the WASM component-model plugin host infrastructure in `zeroclaw-plugins`. This is the first pass at the host side of the WIT v0 plugin system — `ComponentChannel`, `ComponentMemory`, and `ComponentTool` adapters, a shared `ComponentEngine`, `FineGrainedPermissions` with `AddressString` validation, and the `PluginHost::instantiate_*` entry points. `zeroclaw-plugins` is an Experimental crate per AGENTS.md, so the normal stability expectations apply (breaking changes permitted with changelog). CI is fully green.

Key positive signals: the deny-by-default sandbox design is sound, all new production-path error handling uses `?` propagation or safe fallbacks with no panic-inducing `unwrap()`/`expect()`, `AddressString` validation has good edge-case test coverage, and the new code paths are feature-gated behind `plugins-wasmtime` so the rest of the workspace is unaffected.

Two findings below warrant a response before approval.

---

### 🟡 Warning — WASM plugin call errors silently discarded in Channel trait production impls

Several `Channel` trait methods on `ComponentChannel` use `.unwrap_or(false)` or `.unwrap_or(800)` on the WASM call result without logging the error first:

```rust
// call_drop_self_message
bindings
    .zeroclaw_plugin_channel()
    .call_drop_self_message(store, &wit_msg)
    .unwrap_or(false)

// call_health_check
.map_err(anyhow::Error::msg)
.unwrap_or(false)
```

The fallback values are semantically safe choices (don't drop the message, mark health check as failed, use default delay). The problem is that when a sandboxed plugin crashes or panics, the host silently returns the fallback with no tracing event or log record. From an operator perspective, a misbehaving plugin becomes invisible. Per AGENTS.md anti-patterns, production code should propagate errors or document the invariant; swallowing them without at least a `tracing::warn!` makes production debugging substantially harder.

Before merging, please add at minimum a `tracing::warn!` (or a `zeroclaw_log` record) at each silent discard site so plugin failures leave a trace in the log.

---

### 🟡 Warning — WIT v0 removes three interface functions without explicit callout

The diff removes `supports-draft-updates`, `supports-multi-message-streaming`, and `supports-free-form-ask` from `wit/v0/channel.wit`. This is a breaking ABI change for any plugin author building against the v0 WIT: a plugin that exported these functions as stubs (as the previous WIT instructed: "Stub: return false") would now be exporting functions outside the interface contract. The PR summary does not mention this change at all — the "Scope boundary" section is silent on WIT surface changes.

The design intent (replacing per-function capability probing with `ChannelCapabilities` flags) is sound. The fix is just documentation: add a bullet to the PR summary explicitly noting that these three WIT v0 functions are removed, and consider a one-line note in `wit/v0/channel.wit` or a `CHANGELOG-plugins-v0.md` entry. Even in an Experimental crate, early plugin authors tracking the interface deserve a visible signal.

---

### 🔵 Suggestion — Confirm `hyper` optional dep is not vestigial

`crates/zeroclaw-plugins/Cargo.toml` lists `hyper = { version = "1", optional = true }` under the `plugins-wasmtime` feature. Looking through the new component code, `hyper` does not appear to be imported directly — it is likely pulled in transitively by `wasmtime-wasi-http`. If that's the case, the explicit `dep:hyper` entry in the feature flags is vestigial and can be removed (letting `wasmtime-wasi-http` own the dep). Worth a quick check before merge to avoid a misleading explicit dependency.

---

### 🔵 Suggestion — Document the TOCTOU tradeoff in wildcard TCP/UDP matching

`AddressString` uses reverse DNS (`dns-lookup`) to enforce wildcard patterns at connect time. This is TOCTOU-susceptible: an attacker who can influence DNS records for a zone could route a plugin connection to an unintended host that passes the reverse-DNS check. The `AddressString` doc comment explains the pattern rules clearly but says nothing about this tradeoff. A one-sentence note — something like "Wildcard matching relies on reverse DNS at connect time; operators should treat `*.example.com` TCP/UDP grants as granting access to any host whose reverse-DNS record matches, not purely the forward-DNS zone" — would help future maintainers reason about the security model correctly.

---

### 🔵 Suggestion — Labels gap (reviewer action)

The PR body correctly declares `type: feature`, `risk: medium`, `size: L`, `plugins`, `wasm`. The GitHub label sidebar currently shows only `dependencies`. I'll set the missing labels when posting this review.

---

### 🟢 What looks good — FineGrainedPermissions design

The deny-by-default manifest model is the right starting point. Requiring plugins to enumerate every file path, HTTP host, and TCP/UDP address they need, with wildcards restricted to subdomain level 3+, is significantly stronger than the previous coarse-grained yes/no model. The `TcpBind` never-allowed invariant is correctly documented on the enum. This is a solid foundation to build on.

---

### 🟢 What looks good — AddressString validation test coverage

The test suite covers IPv4, IPv6, bare domain, subdomain, wildcard subdomain, single-label rejection, empty-label rejection, label-length limits, hyphen-position validation, TLD/SLD wildcard rejection, invalid characters, and serde round-trip. The edge cases (leading/trailing hyphens, 63 vs 64 char labels, `example.*` rejection) are exactly the ones that bite you in prod. Good test hygiene here.

---

### 🟢 What looks good — ComponentEngine and instantiation path design

`ComponentEngine` wraps `wasmtime::Engine` cleanly, `compile()` is separated from instantiation (enabling per-component compilation cache), and `plugin_bytes_and_permissions` properly uses `?` to propagate errors. The `#[cfg(feature = "plugins-wasmtime")]` gates on all three `instantiate_*` methods are consistent. The fact that no runtime code calls these paths yet means this lands with zero production blast radius while the plumbing gets reviewed properly — the right order of operations.

bheatwole:
Updates:
- Added warning messages in three call sites where forwarding a plugin error is not allowed by the Channel trait
- Documented removal of three trivial wit functions in v0
- Confirmed `hyper` dependency is necessary. `wasmtime_wasi_http::p2::WasiHttpHooks` trait uses `hyper` in function signature. Implementing this trait is how zeroclaw-plugins checks what URLs a plugin is attempting to access. NOTE: it would be possible to remove the `wasmtime_wasi_http` dependency entirely if a custom implementation of `wasi:http/service` was added.
- Clarified wildcard pattern matching in AddressString documentation

WareWolf-MoonWall (approved):
Re-reviewing at 4989bb6e after my COMMENT at fcf89146.

Both concerns I raised in that comment are addressed. Approving.

### ✅ Resolved — WASM call errors emit `tracing::warn!` at Channel trait sites

All three call sites where error forwarding is not permitted by the `Channel` trait now emit a `warn!` rather than silently discarding the failure. This is the right level of signal — loud enough to surface in logs without being fatal.

### ✅ Resolved — Trivial WIT v0 function removals documented

The three removed functions are now called out explicitly in the PR body. No surprises for reviewers reading the diff.

### 🟢 `FineGrainedPermissions` is a strong default security posture

Plugins must name every allowed file path, HTTP URL pattern, and TCP/UDP endpoint explicitly in the `PluginManifest`. Nothing is permitted by default. `AddressStringTooShort` and `AddressStringInvalidLabel` make the rejection surface explicit at input validation time — no silent grants.

### 🟢 Reverse-DNS validation for wildcard HTTP hosts is correct

Denying the connection when the reverse lookup fails prevents an attacker who controls a subdomain from crafting a `*.evil.com` pattern that reaches `victim.com`. The doc comment on that path is clear about the invariant.

### 🟡 Local validation evidence reflects a subset run

The local run shows 885 tests in ~3 s. That is almost certainly a partial feature set. Full workspace CI (9 000+ tests via nextest) is green and is the authoritative bar for a `risk: high` XL PR. Future authors should note that the full workspace nextest run is the expected standard for evidence in this tier.

### 🟡 `wasmtime_wasi_http` transitive surface is accepted

The `hyper` function signature in `WasiHttpHooks` makes this dependency load-bearing, not opportunistic. The PR body calls it out explicitly. Accepted as a deliberate trade-off.

### 🔵 Integration wiring is intentionally out of scope

`instantiate_channel_plugin` and the other instantiation paths are not yet called from any runtime code. That is the correct scope boundary for this PR. A follow-up issue tracking the runtime integration wiring would help milestone planning for v0.8.x.

Audacity88 (changes requested):
I reviewed the current head (`301d5d31c76f3708df9cedd9d42a76627fb246c2`), the WIT v0 surface, the component host/adapters, the sandbox permission wiring, the latest CI state, and the prior review thread. The direction looks valuable, but I think two capability-boundary issues need to be fixed before this is merge-ready.

### 🟢 What looks good — The component host direction fits the plugin architecture

The overall shape is promising: a Wasmtime component-model host, WIT-generated bindings, and explicit Channel/Memory/Tool adapters are the right kind of foundation for the plugin system. I also agree with the decision not to treat the removed WIT channel helper functions as a frozen-interface break here, since `wit/v0` is still unfrozen and the replacement capability bits are documented in the WIT file. The implementation is thorough in a lot of places, especially around converting ZeroClaw channel/media/memory types into the component-facing types and surfacing plugin call failures instead of silently swallowing them.

### 🔴 Blocking — HTTP permission also grants raw TCP connect access

`FineGrainedPermission` defines separate `Http`, `Tcp`, and `Udp` permissions in `crates/zeroclaw-plugins/src/lib.rs`, and the PR body describes raw TCP/UDP access as requiring explicit permission. But `PluginStore::with_permissions` currently handles `FineGrainedPermission::Http(addr)` by adding the same address to `tcp_rules`, setting `has_tcp = true`, and then enabling `SocketAddrUse::TcpConnect` through the shared socket address checker in `crates/zeroclaw-plugins/src/component/v0/plugin_store.rs`.

That means a plugin manifest that only asks for HTTP access to `example.com` can also get raw TCP connect permission to that address through WASI sockets. That collapses the security distinction between “may issue HTTP requests to this host” and “may open a raw TCP connection to this host”, which is exactly the kind of boundary this new fine-grained permission model is supposed to preserve.

Please keep HTTP host access and raw TCP socket access separate, or document and enforce why HTTP must imply TCP without exposing a general raw-socket capability to the guest. I would also add regression coverage for the intended matrix: HTTP-only allows the HTTP path but denies raw TCP, TCP-only allows raw TCP but does not implicitly allow HTTP unless that is intentionally part of the contract, and UDP remains separate.

### 🔴 Blocking — Synchronous channel methods call `blocking_lock` from async orchestrator paths

`call_plugin_sync!` takes the component state with `state.blocking_lock()` in `crates/zeroclaw-plugins/src/component/v0/call_plugin.rs`. The new `ComponentChannel` then uses that macro from synchronous trait methods such as `self_handle`, `self_addressed_mention`, `drop_self_messages`, and `multi_message_delay_ms` in `crates/zeroclaw-plugins/src/component/v0/channel_component.rs`.

Those are synchronous `Channel` trait methods, so callers cannot await the component store lock before asking for values like self identity, mention text, self-message filtering, or multi-message delay. Calling Tokio’s blocking mutex path from those methods can block or panic when the channel is used from runtime/channel processing code instead of just calling a plugin export safely.

Please avoid making synchronous Channel trait methods call into the component store through `blocking_lock`. The safer shape is probably to cache sync identity/capability values during channel construction or after an async initialization step, use defaults for values that cannot be fetched synchronously, or otherwise restructure the adapter so the async plugin calls stay on async trait methods only.

### 🟡 Warning — The tests do not yet exercise the actual component host boundary

The current tests cover useful helpers like address validation, HTTP host matching, capability bit conversion, and type conversion, but I do not see a test that instantiates a tiny component through `ComponentTool::from_bytes`, `ComponentChannel::from_bytes`, or `ComponentMemory::from_bytes`, nor a test that proves the sandbox permission boundary at runtime. For a high-risk plugin host, that leaves the most important part of the PR effectively covered by compile/CI only.

Once the two blockers above are fixed, I would strongly prefer at least one minimal component instantiation smoke test and one permission-boundary regression around the HTTP/TCP separation. That does not need to become a full external plugin E2E suite in this PR, but the new host should have a small proof that the component path and the new sandbox contract work at the boundary they are meant to enforce.

bheatwole:
Updates:
- fix: HTTP permission also grants raw TCP connect access. HTTP now gives just HTTP, TCP gives just TCP. There is a full integration test that confirms this
- fix: Synchronous channel methods call blocking_lock from async orchestrator paths. There are no more blocking wasm calls (in fact the entire `call_plugin_sync` macro could be removed with these changes). This did necessitate a change to the Channel trait to add `async` to `drop_self_messages`. Code review revealed that this method was only called from a single place in the entire production code base (everything else was unit tests). This should be a safe update to the trait

--- 
@Audacity88 The request for "tests do not yet exercise the actual component host boundary" caused me to bring in some additional work I had been doing toward the SDK which has full integration tests that compile component model wasm and verify functionality.

This makes this PR much larger in scope that I originally expected. Please advise as to whether you would like to review this PR as one XL change, or try to break it into smaller components. I'm not quite sure where the correct split points are

---

Additions:
- refactors the 'retry' code into `zeroclaw-infra`
- adds websocket support for plugins
- adds proxy support for plugins using `zeroclaw-config`
- adds the `zeroclaw-plugin-sdk` crate for making plugin development easier. This is referenced in FND-001 as a 1.0.0 Phase 4 D3 deliverable, but it felt necessary now in order to correctly exercise the host code.

~~I am working on updating the main PR text~~ - done

WareWolf-MoonWall (commented):
# Review — PR #7928 `feat(wasi): initial WASM component-model plugin host code`

**Head reviewed:** `771b88c87798148c04d5b62dd01e9de7f6b9a07e`  
**Prior approval:** `4989bb6e` (review body) / recorded at `301d5d31` (state object)  
**Audacity88 CHANGES_REQUESTED:** `301d5d31c76f3708df9cedd9d42a76627fb246c2` — June 21  
**Verdict:** `--comment` (cannot approve over Audacity88's active CHANGES_REQUESTED)

---

Incremental review since my approval at `4989bb6e`. Audacity88 subsequently filed CHANGES_REQUESTED at `301d5d31` (June 21) with three findings. bheatwole has pushed additional commits (current head `771b88c87`) and left a summary comment indicating all three are addressed. My pass below covers the new code and corroborates those claims; I cannot clear Audacity88's hold — only Audacity88 can do that.

---

### ✅ Resolved — HTTP permission no longer grants raw TCP connect access

Audacity88's first 🔴: `FineGrainedPermission::Http(addr)` was adding to `tcp_rules`, collapsing the boundary between "HTTP to this host" and "raw TCP to this host".

bheatwole's commit splits the two paths. The regression test `spike_tcp_permission.rs` (`http_only_permission_denies_raw_tcp_connect` + `tcp_permission_allows_raw_tcp_connect`) is now present and passes per the author's local run and CI. The PR body now explicitly calls out this fix and the test. This directly addresses Audacity88's first blocker.

---

### ✅ Resolved — `blocking_lock` from async orchestrator paths eliminated

Audacity88's second 🔴: `call_plugin_sync!` used `state.blocking_lock()` from synchronous `Channel` trait methods (`self_handle`, `drop_self_messages`, etc.), which can block the async executor.

The fix: `call_plugin_sync!` and `wrap_plugin_call_sync` are removed entirely. `Channel::drop_self_messages` is now `async fn`, removing the only synchronous callsite that needed the macro. The one call site in `zeroclaw-channels/src/orchestrator/mod.rs` is updated to `.await` it. No other in-tree `Channel` implementor overrides `drop_self_messages` (confirmed by the PR's grep note), so all others pick up the new async default transparently. This is a clean resolution.

Note: `Channel::drop_self_messages` becoming `async fn` is a breaking API change on `zeroclaw-api`, which is an Experimental crate — no stability guarantee applies, but the Compatibility section documents the out-of-tree upgrade path correctly.

---

### ✅ Resolved — Real component host boundary now exercised by tests

Audacity88's third 🟡: no test instantiated a real component through `ComponentTool::from_bytes` / `ComponentChannel::from_bytes` / `ComponentMemory::from_bytes`.

The `zeroclaw-plugin-sdk` crate and four real `wasm32-wasip2` example plugins (`tool-echo`, `memory-noop`, `channel-echo`, `tool-tcp-probe`) now round-trip through the unmodified `PluginStore` in integration tests (`spike_tool.rs`, `spike_memory.rs`, `spike_channel.rs`, `spike_tcp_permission.rs`). These also surfaced and fixed two real host bugs (`instantiate_async` vs sync variant, trappable-vs-async mismatch in `from_bytes` probes). This is exactly the kind of boundary proof Audacity88 requested.

---

### 🟡 Warning — Scope has grown substantially since original review

The PR that was reviewed in Round 18/19 was a targeted WASM component host. Since the approval it has acquired:

- A full `zeroclaw-plugin-sdk` guest crate with four real compiled-WASM example plugins
- Websocket support for plugins
- Proxy support via `zeroclaw-config`
- Retry logic hoisted from plugin HTTP paths into `zeroclaw-infra`
- wasmtime bump `43.0.2 → 45.0.1` (required for `instantiate_async`)
- cranelift bump `0.130.2 → 0.132.1`
- Two new transitive dependencies: `cap-net-ext`, `dns-lookup`

Each of these is individually defensible (the SDK was necessary to prove the host works; the retry refactor is a positive extraction; the wasmtime bump was required to fix the async instantiation bug). But taken together this is a substantially different PR than what was approved at `4989bb6e`. bheatwole has already raised the split question with Audacity88 — that conversation should drive the decision. If the maintainer decides to keep it merged, the Validation Evidence section should be updated to reflect the full scope (the current "885+ tests" note pre-dates the SDK integration tests).

---

### 🟡 Warning — Commit history dominated by merge commits

The last 20+ commits visible in the PR are all `Merge branch 'wasi-plugins-host' of github.com:bheatwole/zeroclaw into wasi-plug...`. This is a consequence of bheatwole working from a personal fork and syncing frequently. It makes incremental review against any specific commit boundary (e.g. since `4989bb6e`) very noisy — there is no clean functional-commit boundary to diff against. For a future large PR or a PR-split if the maintainer goes that route, squashing or rebasing before review would help reviewers identify what changed.

---

### 🟡 Warning — Conflicts with PR #8144 (duplicate wasmtime component-model work)

JordanTheJet has opened PR #8144 which also implements a wasmtime component-model / WIT plugin runtime and removes Extism. The two PRs share a goal but take different approaches (`FineGrainedPermissions` manifest model in #7928 vs WIT capability host-interface model in #8144). bheatwole has already commented on #8144 asking how to combine the work. The maintainer needs to resolve this before either PR can land — two runtime implementations cannot both merge without significant conflict.

---

### 🟢 wasmtime `43.0.2 → 45.0.1` upgrade is load-bearing

The version bump is not speculative — it was required to get `instantiate_async` available for the Linker, which the sync path's failure mode demonstrated when exercised by real components. The bump is correct and the Cargo.lock change is expected (cranelift, pulley-interpreter, wasm-encoder, wasmparser all follow wasmtime's release cadence).

---

### 🟢 No AI attribution trailers in commits

Commit message grep across all 57 commits returned no matches for `Co-authored-by`, `Generated by`, `claude`, `openai`, `chatgpt`, `copilot`, `gemini`. Clean.

---

### 🟢 CI fully green at current head

All 18 jobs pass (Format, Lint, Test, Build x86_64, Check all features, Check 32-bit, Check aarch64, Check Windows, Docs Style, Security, Installer Drift, Nix Module Eval, Benchmarks Compile, CI Required Gate, Zerocode RPC Boundary, main). Run `27954552307`.

---

**Next steps:** Audacity88 — the three issues you raised appear addressed by the commits pushed since your review. Please re-review when you have a chance so you can dismiss your CHANGES_REQUESTED if satisfied. Maintainer — please resolve the `#7928` vs `#8144` direction question before either PR progresses further.

WareWolf-MoonWall:
📌 @singlerider — milestone-owner heads-up. This PR carries an active changes-requested review (Audacity88) and is not yet merge-ready. No milestone change made; flagging so you can track it against the current milestone.

WareWolf-MoonWall:
@singlerider — maintainer direction needed: #7928 (bheatwole, wasmtime/WIT plugin host), #8137 (singlerider, per-alias plugin config hardening), and #8144 (JordanTheJet, wasmtime component-model runtime replacement) are in a three-way collision on the plugin runtime. #8144 replaces the Extism runtime that #8137 is hardening; both #8144 and #7928 implement a wasmtime/WIT host. None can land cleanly without a direction decision. The SSOT and capability-model designs from all three are architecturally sound — but the implementation path needs to be chosen before any of them can be approved.

Audacity88 (commented):
Thanks for the update, @bheatwole. I re-reviewed current head `771b88c87798148c04d5b62dd01e9de7f6b9a07e`, the code paths behind my prior findings, the new SDK integration tests, the PR body, current labels, and the latest CI state. My read matches WareWolf-MoonWall's latest follow-up: the prior technical blockers look addressed, but the plugin-runtime consolidation path still needs a maintainer direction call before final approval.

### ✅ Resolved — HTTP permission no longer grants raw TCP

My prior HTTP/TCP boundary blocker looks addressed. `FineGrainedPermission::Http` now only populates the `wasi:http` allow-list, while raw socket access is driven by the separate `Tcp`/`Udp` permission lists. The `spike_tcp_permission.rs` test also exercises the real `wasi:sockets` boundary with a compiled `wasm32-wasip2` component: HTTP-only denies raw TCP, and TCP explicitly allows it.

### ✅ Resolved — Channel plugin calls no longer need the blocking sync path

The `blocking_lock` / `call_plugin_sync!` shape from my prior blocker is gone. `Channel::drop_self_messages` is now async, the orchestrator awaits it, and the component channel adapter uses the async plugin-call path. That removes the blocking store-lock workaround from the runtime channel path.

### ✅ Resolved — The component host boundary now has real integration coverage

The new `zeroclaw-plugin-sdk` tests now build real `wasm32-wasip2` guest components and load them through the unmodified host path for tool, memory, channel, and TCP-permission behavior. That is the kind of host-boundary proof I was asking for. It is especially useful that this work found the `instantiate_async` / async-required-store mismatch instead of only exercising helper conversions.

### 🟡 Warning — Hold final approval until the plugin-runtime direction is chosen

I agree with WareWolf-MoonWall's warning that this PR now sits in a three-way plugin-runtime collision:

- #7928 implements the Wasmtime/WIT host and SDK proof.
- #8144 also implements a Wasmtime/component-model runtime replacement and carries some important runtime-limit, collision-handling, signature-policy, and registry/core-slimming ideas.
- #8137 removes raw env access and moves plugin config toward per-alias config, which is valuable regardless of which host path wins.

The code blockers from my prior review look resolved, but approving this PR now would make it appear merge-ready before the implementation-base decision is made. The next maintainer decision should be whether #7928 is the base and #8144/#8137 donate specific pieces into it, or whether a different consolidation path is cleaner.

### 🔵 Suggestion — Small cleanup before the next pass

If #7928 becomes the base, I would clean up two bits of drift before the final approval pass:

- `crates/zeroclaw-plugin-sdk/tests/spike_channel.rs` still says the channel sync-call fix is pending/trappable. That comment is now stale because this branch made `drop_self_messages` async and removed the sync-call path.
- The PR body evidence text is directionally useful, but the final body should use exact current command/results wording for the SDK tests and current green CI.

So my current read is: prior technical blockers resolved; final approval should wait for the #7928 / #8144 / #8137 direction call and small public-record cleanup if this remains the base.

JordanTheJet:
@bheatwole following up on your note over in #8144 — agreed, let's converge rather than run two parallel component-model hosts.

Plan from my side: I'll rebase #8144 on top of this PR once it's mergeable (i.e. once the remaining direction call is settled and this lands or is green-lit as the base). Your host is further along on breadth — channel/memory worlds, the SDK + real `wasm32-wasip2` spike coverage, and the fine-grained permission model — so it makes sense as the foundation rather than re-litigating the host layer.

Once I rebase, #8144 collapses down to the pieces this PR intentionally leaves out of scope, donated on top:

- **Runtime wiring** — calling the `instantiate_*` path from `zeroclaw-runtime`'s tool loader (you note this is out of scope here), plus built-in-vs-plugin name-collision handling and `signature_mode` honoring.
- **Per-call execution limits** — fuel + epoch + memory bounds with a fresh `Store` per call.
- **Existence-only secret model** — worth flagging as a deliberate tightening of `plugin-config.get-secret`: instead of returning the secret value into the sandbox, the guest sees only `secret-exists -> bool` and the host injects the credential at the `wasi:http` egress hook (the same `WasiHttpHooks`/`send_request` seam you already implement). The value never enters the guest. Happy to discuss whether that belongs in the base or as a fast-follow.

No action needed from you — I'll hold #8144 and rebase once this is ready. Flagging here so the direction call has the full picture.

JordanTheJet:
**Maintainer direction call** (re: the #7928 / #8144 / #8137 three-way collision the bot has been flagging):

Confirming **#7928 as the base** for the WASM component-model plugin host. It's furthest along — channel/memory/tool worlds, the guest SDK with real `wasm32-wasip2` spike coverage, and the fine-grained permission model — and the prior technical blockers (HTTP/TCP separation, async channel path, host-boundary coverage) are resolved.

Direction for the other two:
- **#8144 (mine)** rebases on top of this once it's mergeable, collapsing to the pieces this PR leaves out of scope: runtime wiring (`instantiate_*` from `zeroclaw-runtime`'s tool loader + name-collision/`signature_mode` handling), per-call fuel/epoch/memory limits, and the existence-only secret model (host injects at the `WasiHttpHooks` egress seam rather than handing the value to the guest via `get-secret`).
- **#8137 (@singlerider)** folds in regardless of which host wins — the per-alias plugin config hardening / raw-env removal is valuable independent of the runtime layer.

Disclosure: I'm both a maintainer and the author of #8144, so this call deliberately picks the *other* author's PR as the base rather than my own. @singlerider — over to you for milestone assignment and to ratify the parts that touch #8137, since I shouldn't be the one advancing my own donated pieces.

Remaining on this PR before merge: the stale `spike_channel.rs` comment Audacity88 noted, and refreshing the PR-body evidence to current green CI.

bheatwole:
Updates:
- Updated `spike_channel.rs` comments to reflect current state
- Added specific commands used to test the SDK to the PR body text, along with output
- Rebased from master

Audacity88 (changes requested):
Reviewed current head `db22a350f26d7793a5ed749d500024db2b4badf1`, the live PR state/checks, the PR body and comments, the prior review thread, #6943's latest direction notes, the #8144 collision thread, the merged #8137 config-hardening PR, the WIT v0 diff, and the component host / SDK / HTTP-helper source paths. I did not run local Cargo; current GitHub CI is green, and this review is based on source inspection plus the existing PR validation evidence.

### ✅ Resolved — my prior capability-boundary blockers are addressed

The HTTP-vs-raw-TCP boundary is now separated: `FineGrainedPermission::Http` feeds only the `wasi:http` allow-list, while raw socket access comes from the separate `Tcp` / `Udp` permission lists. The new `spike_tcp_permission.rs` test exercises that boundary through a real `wasm32-wasip2` component. The old `blocking_lock` / `call_plugin_sync!` issue is also gone: `drop_self_messages` is async now, the orchestrator awaits it, and the component channel adapter uses the async path.

### ✅ Resolved — the implementation-base decision is now on record

The earlier reason I held final approval was the #7928 / #8144 / #8137 collision. That direction is now clearer: #6943 and JordanTheJet's PR-thread comments record #7928 as the base, #8137 has merged, and #8144 is no longer the competing base path. Its remaining value is the runtime wiring, collision/signature-policy, execution-limit, and secret-model pieces to donate on top of this base.

### 🔴 Blocking — `download-to-attachment` enforces `max_bytes` only after buffering the full body

The new `http-helpers.download-to-attachment` WIT API takes `max-bytes`, and the WIT docs describe it as a hard cap mirroring the bounded native attachment-download path. The host implementation in `crates/zeroclaw-plugins/src/component/v0/http_helpers_host.rs` does the allow-list check, sends the request, then calls `resp.bytes().await` before checking `data.len() > max_bytes`.

That means an allow-listed endpoint can still make the host allocate the entire response before the helper rejects it. The cap protects the value returned to the guest, but it does not protect the host process from a large or malicious response body. This is a new plugin/network trust boundary, so the limit needs to be enforced while reading the body, not after the allocation has already happened.

Please stream or otherwise cap the response as it is read, and add a regression that proves an oversized response is rejected without buffering the whole body first.

### 🔴 Blocking — `plugin-config.get-secret` is a new secret/config contract, but the PR still says that surface is out of scope

The branch adds `wit/v0/plugin-config.wit` with `get-secret`, imports it into the tool, memory, and channel plugin worlds, exposes SDK helpers like `tool::get_secret`, and implements the host side by returning cloned, already-decrypted secret values from `PluginStore::network_config`. The manifest-side `declared_secrets` filter is a useful narrowing step, and this is not a global secret-store handle, but it is still a new contract for passing secrets into guest components.

That does not match the current PR body. The scope boundary says this PR "does not address plugin access to ENV VAR and configuration", and the Security & Privacy section says secrets / tokens / credentials handling changed: `No`. It also leaves the relationship to the now-recorded #8144 follow-up unclear: #6943 calls out the existence-only secret model as one of the pieces to preserve on top of #7928.

Before this lands as the selected base, please make the secret/config surface explicit and choose the intended contract. Either move `plugin-config` / `get-secret` out to the follow-up that settles the #8144 secret model, or keep it here and update the PR body, WIT docs, security/privacy impact, compatibility notes, and tests so reviewers and plugin authors can see that guest components receive actual secret values and why that is acceptable.

bheatwole:
Updates:
- removes the `get_secret` functionality to limit the scope of this PR
- implements checking for max_bytes in download_to_attachment while chunks are streamed to eliminate OOM concern

singlerider (approved):
Re-reviewed at `1f8a85df` (FETCH_HEAD asserted against the PR head). The three commits since @Audacity88's `CHANGES_REQUESTED` resolve both standing blockers. Verified locally with the component feature enabled, not just by reading the diff.

### RESOLVED ✅ — `download-to-attachment` now caps while streaming (was 🔴)

`crates/zeroclaw-plugins/src/component/v0/http_helpers_host.rs` no longer calls `resp.bytes().await` before the check. It reads `resp.bytes_stream()` and enforces the cap per chunk before extending the buffer:

```rust
let mut stream = resp.bytes_stream();
while let Some(chunk) = stream.next().await {
    let chunk = chunk.map_err(...)?;
    if data.len() as u64 + chunk.len() as u64 > max_bytes {
        return Err(format!("download-to-attachment exceeded max-bytes ..."));
    }
    data.extend_from_slice(&chunk);
}
```

The exact regression you asked for is present and proves the cap is not header-dependent: `download_to_attachment_rejects_chunked_body_over_max_bytes_without_content_length` spins up a `Transfer-Encoding: chunked` server with no Content-Length and `max_bytes = 4` against a 9-byte chunked body, so a fast header reject is impossible — the rejection can only come from inspecting bytes as they arrive.

### RESOLVED ✅ — secret surface removed; body now matches scope (was 🔴)

`get-secret` / `get_secret` is gone from `wit/` and all of `crates/zeroclaw-plugins/` (commit `fcbb3167`, "remove 'get-secret' from this large PR"). `wit/v0/plugin-config.wit` now exposes only `get-proxy-url`. The PR body's scope boundary ("Does not address plugin access to ENV VAR and configuration") and Security & Privacy answer ("Secrets / tokens / credentials handling changed: `No`") are now consistent with the code. The existence-only secret model is correctly deferred to the #8144 follow-up #6943 calls out.

### Verified locally

```
cargo test -p zeroclaw-plugins --features plugins-wasm-cranelift --lib http_helpers
running 6 tests
... download_to_attachment_rejects_body_over_max_bytes ... ok
... download_to_attachment_rejects_chunked_body_over_max_bytes_without_content_length ... ok
... (all 6) ok
test result: ok. 6 passed; 0 failed
```

All 17 CI checks are green (Format, Lint, Test, Check all-features included); the `BLOCKED` merge state is this standing review, not CI.

Both blockers are addressed and the earlier capability-boundary / implementation-base concerns were already marked resolved. Approving.

Audacity88 (changes requested):
Reviewed current head 38c388ead04b7e7d84844f8ecf3ddac83e21d56f against the live PR state, green GitHub checks, the latest comments and reviews, #6943's implementation-base direction, #8144's donated follow-up scope, and the current WIT/plugin host source. I did not run local Cargo for this pass; this is source inspection plus the existing CI and SingleRider's latest local verification. I agree with SingleRider that the two blockers from my prior review are resolved; this re-review raises one new lower-level `wasi:http` boundary blocker.

### ✅ Resolved — the `download-to-attachment` cap is now enforced while streaming

I agree with @singlerider that the specific blocker from my prior review is addressed for `http-helpers.download-to-attachment`. `crates/zeroclaw-plugins/src/component/v0/http_helpers_host.rs` now reads `resp.bytes_stream()` and checks `data.len() + chunk.len()` before extending the buffer, so the host no longer waits for the full response body before applying `max-bytes`.

The new regression also proves this is not only a `Content-Length` fast-path: `download_to_attachment_rejects_chunked_body_over_max_bytes_without_content_length` serves a chunked response without `Content-Length` and still rejects once the streamed bytes exceed the cap.

### ✅ Resolved — the guest secret surface has been removed from this PR

The `get-secret` / `get_secret` surface that my prior review called out is gone from the component WIT and host implementation. `wit/v0/plugin-config.wit` now exposes only `get-proxy-url`, and the component-side `plugin_config.rs` only returns `PluginNetworkConfig::proxy_url`.

That matches the current PR body: this base PR does not settle plugin secret/config access. The existence-only secret model remains correctly left to the #8144 follow-up pieces tracked through #6943.

### 🔴 Blocking — raw `wasi:http` still buffers the full response body in the host

The latest `download-to-attachment` fix closes the helper-specific path, but the underlying proxy-aware `wasi:http` hook still has the same host-memory shape. In `crates/zeroclaw-plugins/src/component/plugin_store.rs`, `send_via_proxy_client` deliberately buffers the full response with `resp.bytes().await` and then wraps it in `Full::new(body_bytes)` before returning it to Wasmtime.

That matters because the new `http-helpers.wit` docs explicitly frame `send-multipart` and `download-to-attachment` as convenience sugar over raw `wasi:http`: a guest could hand-roll the same download directly over `wasi:http`. So an allow-listed endpoint can still make the ZeroClaw host allocate the entire response body before the guest sees it, bypassing the new helper-level `max-bytes` cap. The allow-list controls the destination, but it does not make the response size trustworthy; the host still needs to defend itself from malicious, compromised, or simply broken responses from an allow-listed URL.

Please either stream the proxied `wasi:http` response through without materializing the whole body in host memory, or add a host-enforced response cap at this lower-level hook and document/test that contract. The existing chunked/no-`Content-Length` regression proves the helper-level cap works, but the new regression should exercise the raw `wasi:http` hook so it proves the lower-level cap is enforced while bytes arrive rather than after allocation.

### 🟢 What looks good — the WIT compatibility risk is contained to experimental v0

I checked the touched WIT directory and there is no `wit/v0/.frozen` marker. Under the current WIT versioning rules, these `@unstable(feature = plugins-wit-v0)` changes are still experimental rather than a frozen plugin-compatibility break. The removal of the three redundant channel functions is therefore not a stable-interface blocker on its own.

singlerider (commented):
@bheatwole re-reviewed at head `38c388ea` against @Audacity88's active changes-requested. His new lower-level blocker is valid and unaddressed, and I concur with it, so my earlier approval should not stand as the last word.

### RESOLVED 🟢 - the two prior blockers are addressed

The `download-to-attachment` cap is now enforced while streaming (`http_helpers_host.rs` reads `bytes_stream()` and checks `data.len() + chunk.len()` before extending), proven by `download_to_attachment_rejects_chunked_body_over_max_bytes_without_content_length`. And the guest `get-secret` surface is gone from the WIT and host, leaving secret/config access to the #8144 follow-up. Both match @Audacity88's resolved notes.

### 🔴 Blocking - raw wasi:http still buffers the full response body in host memory (concurring with @Audacity88)

`send_via_proxy_client` in `crates/zeroclaw-plugins/src/component/plugin_store.rs` materializes the entire response at line 207 (`resp.bytes().await`) and wraps it in `Full::new(body_bytes)` at 219, with no size cap at this hook. Because `http-helpers.wit` documents `send-multipart` / `download-to-attachment` as convenience sugar over raw `wasi:http`, a guest can hand-roll the same download directly over `wasi:http` and bypass the helper-level `max-bytes` cap entirely. The allow-list controls the destination, not the response size, so an allow-listed but malicious, compromised, or broken endpoint can force the host to allocate an arbitrary-size body before the guest ever sees it. That is a host-memory DoS the new helper cap does not cover.

Please either stream the proxied `wasi:http` response without materializing the whole body, or enforce a host-level response cap at this hook, and add a regression that exercises the raw `wasi:http` path (not just the helper) and proves the cap applies while bytes arrive rather than after allocation.

I confirmed no competing PR for this WASM host base; #6943 is the implementation-base direction and #8144 is the donated follow-up scope, so this is the single base PR.

@Audacity88 - concurring with your raw wasi:http blocker; it is real and open at this head. Yours to convert once the lower-level cap lands.

singlerider:
The giant diff scares me. I'm going to make a competing implementation to see how it can compare to see if I can reduce that line count as reference.

bheatwole:
As I noted before, the size of this PR also concerns me. I've added a review of #8368, noting the items I'd like to see included.

I'm comfortable with a plan of:
- making 8368 the initial push for component model host code instead of 7928
- abandoning 7928 and selectively pulling in code to future PRs
- opening an RFC to discuss and gain consensus on the following security considerations:

For discussion:
- **Permission model**: The current `PluginPermission` enum seems too permissive and there was support for `FineGrainedPermission` from this PR. Is this the correct level of detail or need adjustment?
- **How should plugins read configuration?** Especially with access to files in the sandbox, a plugin could read its own config file in whatever format it chooses. Or is it better to keep all config together in the main `.toml` file and give the plugin access to a part of it (the `.wit` would need to be updated for channel and memory types)
- **Do plugins gain access to any secrets?**: Should the `manifest.toml` be allowed to define a list of secrets that the user should specify and be transferred to the plugin? (i.e. a connection string with embedded password)
- **Permissions for user-supplied URLs**: If the user runs a cloud database, or hosts a local channel of some kind, the plugin author would not be able to include that URL/IP address in the `manifest.toml`. What capabilities do we need to give the user to extend the permissions beyond what the author can know?
- **HTTP Max Size**: A plugin might need to request a specific max size for http calls, but the user should also have a configuration setting to explicitly set an upper bound, with an warning if the two are in conflict. What are reasonable defaults and handling?
- **stdin/stdout access?**: Should the `cli` channel continue to exist as a compiled-in channel, or be moved to a plugin? If it is moved to a plugin, we will also need to give a method to access stdin/stdout to plugins.
- **ENV VAR access?**: Should we add a permission setting to access all, or maybe just specific env-vars? Or maybe only the ones that directly affect config and are thus part of the config permissions?
- **HTTP Listener?**: I tested [a Microsoft Teams channel plugin](https://github.com/bheatwole/zeroclaw/tree/plugin-teams) and while it can work by polling the Graph API, it works better if a webhook is installed and M365 can push zeroclaw an http notice when new messages are sent to the bot. Is there a way for a plugin to request the host to open a port and listen for http calls, sending them on to the plugin under certain circumstances? Is this too big a hole? Users would have to be savvy enough to route connections to their agent and to handle certificates.

singlerider:
Closing as superseded by #8368, which is now the selected base for the WASM component-model plugin host (#6943).

This is not a dismissal of the work here. #7928 carried the most proof and SDK work of the three implementations, and it directly shaped the chosen direction. The base decision came down to landing surface: #8368 is the smaller trunk that gets the tool/channel/memory worlds in with green CI, and #7928 still has a real open blocker on raw `wasi:http` buffering.

The valuable pieces are not lost. The SDK/proof work and the `wasi:http` / http-helper host surface are slated to land as donated follow-ups on top of #8368 once the buffering blocker is resolved. Please retarget those as follow-up PRs against the new trunk rather than reviving this as a competing base.

You will receive attribution on the eventual merge commit for the inspiration and the pieces folded forward. Thanks for the work, @bheatwole.

PR #7930, opened 2026-06-18
fix(providers): guard is_non_retryable against false-positive on 429 bodies

dvgamerr:
## Summary

- **What changed and why:**
  - Added an early-exit guard in `is_non_retryable()`: if `is_rate_limited()` returns true, short-circuit and return false immediately — prevents a 429 body containing words like "model" or "invalid" from being misclassified as a permanent failure by the string heuristics below
  - Narrowed `parse_retry_after_ms` to the HTTP header form `Retry-After: <integer>` only (colon separator, integer seconds) to avoid false-positives from `retry_after:` in freeform error message bodies
  - Updated tests to document the new scope: added cases for case-insensitivity, header embedded in a multi-line HTTP response, and intentionally unsupported forms (underscore separator, space separator, float values)
- **Scope boundary:** No changes to retry budget, backoff policy, or any provider-specific error handling
- **Blast radius:** `zeroclaw-providers::reliable` only — all providers passing through the `ReliableProvider` wrapper receive this fix
- **Linked issue(s):** None

```
$ cargo fmt --all -- --check
(no output — tree is formatted)

Finished `dev` profile [unoptimized + debuginfo] target(s) in 52.23s

$ cargo test -p zeroclaw-providers
test result: ok. 1022 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3.42s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s
```

- **Beyond CI — what did you manually verify?** Reviewed the diff to confirm the guard is placed before every heuristic block; verified test cases cover case-insensitive header, multi-line HTTP message, underscore form, space separator, and float — all intentionally not parsed in the new scope
- **If any command was intentionally skipped, why:** Full `cargo test` suite not run; only the directly affected crate (`zeroclaw-providers`) was tested given the change is isolated to `reliable.rs`

## Compatibility (required)

- Backward compatible? Yes

`git revert 425629a38` is the plan.

### DIFF
--- a/crates/zeroclaw-providers/src/reliable.rs
+++ b/crates/zeroclaw-providers/src/reliable.rs
@@ -89,6 +89,13 @@ pub fn is_non_retryable(err: &anyhow::Error) -> bool {
         return false;
     }
 
+    // Rate-limit errors are always retryable (transient quota exhaustion).
+    // Guard this early so the string heuristics below cannot false-positive
+    // on a 429 body that happens to contain words like "model" + "invalid".
+    if is_rate_limited(err) {
+        return false;
+    }
+
     // 4xx errors are generally non-retryable (bad request, auth failure, etc.),
     // except 429 (rate-limit — transient) and 408 (timeout — worth retrying).
     if let Some(reqwest_err) = err.downcast_ref::<reqwest::Error>()
@@ -260,38 +267,14 @@ fn is_non_retryable_rate_limit(err: &anyhow::Error) -> bool {
     false
 }
 
-/// Try to extract a Retry-After value (in milliseconds) from an error message.
-/// Looks for patterns like `Retry-After: 5` or `retry_after: 2.5` in the error string.
+/// Try to extract a Retry-After value (in milliseconds) from an HTTP response header
+/// embedded in the error message. Parses `Retry-After: <seconds>` (integer seconds).
 fn parse_retry_after_ms(err: &anyhow::Error) -> Option<u64> {
-    let msg = err.to_string();
-    let lower = msg.to_lowercase();
-
-    // Look for "retry-after: <number>" or "retry_after: <number>"
-    for prefix in &[
-        "retry-after:",
-        "retry_after:",
-        "retry-after ",
-        "retry_after ",
-    ] {
-        if let Some(pos) = lower.find(prefix) {
-            let after = &msg[pos + prefix.len()..];
-            let num_str: String = after
-                .trim()
-                .chars()
-                .take_while(|c| c.is_ascii_digit() || *c == '.')
-                .collect();
-            if let Ok(secs) = num_str.parse::<f64>()
-                && secs.is_finite()
-                && secs >= 0.0
-            {
-                let millis = Duration::from_secs_f64(secs).as_millis();
-                if let Ok(value) = u64::try_from(millis) {
-                    return Some(value);
-                }
-            }
-        }
-    }
-    None
+    let lower = err.to_string().to_lowercase();
+    let pos = lower.find("retry-after:")?;
+    let after = &lower[pos + "retry-after:".len()..];
+    let secs: u64 = after.split_whitespace().next()?.parse().ok()?;
+    Some(secs * 1_000)
 }
 
 fn failure_reason(rate_limited: bool, non_retryable: bool) -> &'static str {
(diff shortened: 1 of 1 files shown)


singlerider (changes requested):
Checked out `425629a381` locally and ran `cargo test -p zeroclaw-providers --lib reliable`: 81 passed, 0 failed. The core fix is right, but the PR bundles an out-of-scope behavioral regression that has to come out first.

### 🔴 Blocking — `parse_retry_after_ms` rewrite silently narrows Retry-After parsing

The PR title and intent are about the early `is_rate_limited` guard in `is_non_retryable`, which is correct. But the same diff rewrites `parse_retry_after_ms` to only accept the exact `Retry-After:` (hyphen + colon) integer-seconds header form, and the four edited tests flip from asserting a parsed value to asserting `None` for cases that previously worked:

- `retry_after:` underscore form -> now `None` (was `Some(10_000)`)
- `Retry-After 7` space separator -> now `None` (was `Some(7000)`)
- fractional seconds `2.5` -> now `None` (was `Some(2500)`)

These are not hypothetical. Provider error strings surface Retry-After hints in multiple shapes: HTTP headers use integer seconds, but provider JSON bodies (OpenAI-style and others) carry `retry_after` with an underscore and sometimes a fractional value. `compute_backoff` (reliable.rs:696) is the only caller; when `parse_retry_after_ms` returns `None` it falls back to `base` backoff and ignores the server's explicit wait hint. So this change makes the client stop honoring a class of Retry-After signals it used to respect, for no stated reason — the tests were rewritten to match the narrowed code rather than the code fixed to keep the behavior.

Two ways forward, either is fine:
1. Keep the early `is_rate_limited` guard (the actual fix), and drop the `parse_retry_after_ms` changes entirely. Land just the guard. The false-positive bug you're fixing has nothing to do with retry-after parsing.
2. If the parser genuinely needed hardening (e.g. the old float path had a real defect), keep accepting the underscore, space, and fractional forms — restore those test assertions to their passing values and add the new header-case-insensitivity tests on top, rather than deleting coverage.

Please split the retry-after parser change out of this PR or restore the dropped forms. The guard alone is a clean, mergeable fix.

### 🟢 The early rate-limit guard is correct

Returning `false` from `is_non_retryable` as soon as `is_rate_limited` matches is the right shape: it short-circuits before the downstream string heuristics can false-positive on a 429 body that happens to contain tokens like `model` + `invalid`. `is_non_retryable_rate_limit` is still consulted on its own path for genuine business/quota 429s, so plan-exhaustion 429s remain non-retryable. The placement (after the explicit non-retryable early returns, before the heuristic block) is correct.

crates/zeroclaw-providers/src/reliable.rs
```diff
@@ -260,38 +267,14 @@ fn is_non_retryable_rate_limit(err: &anyhow::Error) -> bool {
...
-                let millis = Duration::from_secs_f64(secs).as_millis();
-                if let Ok(value) = u64::try_from(millis) {
-                    return Some(value);
-                }
-            }
-        }
-    }
-    None
+    let lower = err.to_string().to_lowercase();
+    let pos = lower.find("retry-after:")?;
```
singlerider, line 274: 🔴 Blocking. This is the narrowing. The old `parse_retry_after_ms` accepted `retry-after:`, `retry_after:`, `retry-after ` and `retry_after ` prefixes and fractional seconds; this version accepts only the exact `retry-after:` integer-seconds form. Provider error strings carry the underscore form (`retry_after`) and fractional values from JSON bodies, not just the HTTP header. `compute_backoff` (reliable.rs:696) falls back to `base` when this returns `None`, so the client now ignores a class of Retry-After hints it used to honor. Either drop this rewrite from the PR (the `is_rate_limited` guard is the real fix and is independent), or keep accepting the underscore/space/fractional forms and add the new case-insensitive header test on top. Do not delete the existing passing coverage to match narrowed code.

singlerider (changes requested):
Inline anchor for the blocking finding in my review.

Audacity88:
@dvgamerr this PR is still blocked on the June 25 review.

The early `is_rate_limited()` guard is the right core fix, but the same diff also narrows `parse_retry_after_ms` and removes previously supported `Retry-After` parsing forms. The next action is to remove or rework that out-of-scope parsing change so the PR only fixes the 429 false-positive retry classification, or split the parsing behavior change into a separate design-backed PR.

Marking this as `stale-candidate` with a follow-up date of July 18, 2026. If there is no branch update or maintainer takeover by then, this may be closed for backlog hygiene. The 429 retryability fix can still come back as a refreshed PR.

Audacity88:
@dvgamerr closing this for backlog hygiene after the July 18 follow-up date. There has been no branch update or maintainer takeover, and the June 25 blocker remains unresolved on the current head.

The narrow 429 retryability fix is still valid; this closure is not a rejection of that work. A refreshed PR can bring back the early rate-limit classification guard while either leaving `parse_retry_after_ms` unchanged or preserving its existing underscore, space-separated, and fractional forms. Please rebase onto current `master` and include fresh focused validation before reopening or replacing this PR.

Issue #7804, opened 2026-06-16
[Bug]: Code history can send non-alternating Anthropic messages


Audacity88:
### Affected component

runtime/daemon

### Severity

S1 - workflow blocked

### Current behavior

A long or resumed Code/ACP session can reach an Anthropic provider call with adjacent same-role chat messages in the provider-bound history. The observed failure was a provider 400:

```text
Anthropic API error (400 Bad Request): messages: roles must alternate between "user" and "assistant", but found 2 "user" messages in a row
```

The visible turn had been reading generated Rust build output immediately before the failure, but the actual failure was the model-provider request shape. After the failed turn, a follow-up prompt such as `go on` did not reliably resume the interrupted work; the agent appeared to enter a generic standby/heartbeat path instead of continuing the prior task.

This may be related to Code session resume, ACP history hydration, tool-result persistence/replay, history pruning, or failed-turn recovery. The exact source of the adjacent `user` messages still needs a focused regression.

### Expected behavior

Runtime history sent to providers with role-ordering constraints should remain valid after Code session resume, long tool-heavy turns, context pruning, and failed-turn recovery.

If persisted or pruned history cannot be safely converted, the runtime should fail locally with an actionable diagnostic that identifies the bad session/history shape instead of sending an invalid provider request.

After a provider-side turn failure, the next user prompt should preserve enough task/session context to recover or continue intentionally, rather than dropping into an unrelated standby flow.

### Steps to reproduce

```bash
1. Use ZeroCode Code/ACP mode with an Anthropic-backed agent.
2. Resume or continue a long Code session with prior user/assistant/tool history.
3. Let the agent perform tool-heavy work until another provider request is made.
4. Observe that the provider request can fail with `messages: roles must alternate between "user" and "assistant"`.
5. Send a short continuation prompt such as `go on`.
6. Observe that the interrupted task may not resume and the agent may enter a generic standby/help flow instead.
```

Minimal deterministic reproduction is still needed. The issue is filed from an observed failure plus current source inspection so the bad history boundary can be isolated.

### Impact

The active Code turn aborts before the agent can continue. Users see a provider-level 400 that does not explain which local history, resume, pruning, or persistence path produced the invalid message order.

The recovery path is also confusing: after the provider failure, a normal continuation prompt may not pick up the interrupted task, forcing the user to reconstruct context manually.

### Logs / stack traces

Relevant redacted error excerpt:

```text
[...]

PR #7931, opened 2026-06-18
fix(providers): coalesce stripped compatible history roles

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Normalize OpenAI-compatible prompt-guided history after native-tool stripping by coalescing adjacent same-role non-system messages, not only adjacent assistant messages.
  - Keep internal pruned-history markers out of the compatible-provider payload before coalescing so internal context-separator text is not sent as user text.
  - Add regressions for adjacent user messages after a dropped tool result and for pruned-history markers next to a real user continuation.
- **Scope boundary:** This PR only changes the OpenAI-compatible `native_tool_calling = false` history-normalization path. It does not change native Anthropic request conversion, runtime history persistence, Code/ACP resume semantics, or generic failed-turn recovery.
- **Blast radius:** OpenAI-compatible providers that use prompt-guided tools can see different merged message content when prior tool stripping would otherwise produce invalid adjacent same-role messages.
- **Linked issue(s):** Addresses the compatible-provider history-normalization part of #7804; Related #7799
- **Labels:** `bug`, `size: S`, `risk: high`, `cron`, `provider`, `runtime`, `provider:anthropic`, `provider:compatible`

## Validation Evidence

  ```bash
  git diff --check
  ```

  ```text
  # no output
  ```

  ```bash
  scripts/zc-cargo fmt -p zeroclaw-providers -- --check
  ```

  ```text
  # no output beyond wrapper target routing
  ```

  ```bash
  scripts/zc-cargo test -p zeroclaw-providers strip_native_tool_messages --lib
  ```

  ```text
  running 8 tests
  test compatible::tests::strip_native_tool_messages_drops_internal_pruning_markers_before_coalescing ... ok
  test compatible::tests::strip_native_tool_messages_drops_empty_narration_cleanly ... ok
  test compatible::tests::strip_native_tool_messages_preserves_regular_messages ... ok
  test compatible::tests::strip_native_tool_messages_coalesces_adjacent_users ... ok
  test compatible::tests::strip_native_tool_messages_drops_empty_assistant_tool_calls ... ok
  test compatible::tests::strip_native_tool_messages_coalesces_adjacent_assistants ... ok
  test compatible::tests::strip_native_tool_messages_passthrough_when_native_tool_calling_enabled ... ok
  test compatible::tests::strip_native_tool_messages_removes_tool_and_tool_calls ... ok

  test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 1014 filtered out; finished in 0.01s
  ```

  ```bash
  scripts/zc-cargo-broad --lane workspace-check clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings
  ```

  ```text
  Finished `dev` profile [unoptimized + debuginfo] target(s) in 36m 55s
  ```

  ```bash
  scripts/zc-cargo-broad --lane nextest nextest run --locked --workspace --exclude zeroclaw-desktop
  ```

  ```text
  Summary [  86.040s] 9182 tests run: 9182 passed, 11 skipped
  ```

- **Beyond CI — what did you manually verify?** Verified the helper output shape for the old adjacent-assistant case, the new adjacent-user case, passthrough with native tool calling enabled, regular-message preservation, empty narration dropping, and internal pruning marker removal before coalescing.

## Security & Privacy Impact

## Compatibility

- Backward compatible? `Yes`
- If `No` or `Yes` to either: No user action required.

## Rollback

- **Observable failure symptoms:** OpenAI-compatible prompt-guided tool sessions may again fail with provider errors such as `messages: roles must alternate` after tool-heavy resumed history, or may surface internal pruning marker text in provider-bound messages.

## Supersede Attribution

- Superseded PRs + authors: N/A
- If `No`, why: No superseded PR or incorporated external contributor work.


tidux (commented):
Read the single-file diff at 71f18630, walked the helper logic against `crates/zeroclaw-providers/src/anthropic.rs` (which 883164eec3 / #7775 introduced last week), traced every caller of `strip_native_tool_messages` (lines 2589, 2672, 2775, 2895, 3178), simulated `git merge upstream/master` against the branch (clean auto-merge, no conflicts in `compatible.rs`), and ran the focused tests locally. CI matrix on this head is green across all 18 checks including `Check x86_64-pc-windows-msvc` and `Check (all features)` (13m9s). No prior formal reviews on the PR; this is the first verdict.

### 🟢 What looks good — the fix is the right shape and the smallest possible diff

This is a true one-file change in `crates/zeroclaw-providers/src/compatible.rs` that does exactly two things, each with a regression test:

1. **Generalize the post-strip coalesce from `assistant + assistant` to any same-role non-system pair.** The motivating trace from #7804 is `user → assistant{tool_calls} → tool{result} → user("go on")`: after the `tool` row is dropped and the `assistant{tool_calls}` rewrites to empty content (and is therefore dropped), the two surrounding `user` rows become adjacent, and Anthropic-family backends reached through compatible gateways reject that with 400 `messages: roles must alternate`. The new condition `last.role == msg.role && msg.role != "system"` at compatible.rs:2179 handles all three adjacency cases the strip can produce — `assistant+assistant` (preserved from #5825), the new `user+user`, and any future hypothetical role pair — without ever merging two `system` messages (which can still appear when the `flatten_system_messages(merge=false)` path is in effect, per `flatten_system_messages_keeps_system_only_at_start_without_user_merge` at compatible.rs:4882).

2. **Drop the two internal pruning markers (`[Tool exchange: N tool call(s) — results collapsed]` and `[context continues]`) before they reach the wire.** The unconditional summary drop is the critical surface: if the synthetic summary leaks through, the model sees it as its own prior reply. The conditional separator drop (only when the previous input message is a pruned-tool-exchange-summary) is the cautious branch — it keeps any stray `[context continues]` user message intact, which is a real fail-safe in case some persisted-history path produced a separator without the paired summary. Today's pruner at history_pruner.rs:419-431 only emits separators between two real assistants, so the separator-drop branch will be quiescent on freshly-pruned history; it activates on resumed/persisted history where the sequence might already be `summary→separator→…`.

Both regression tests (`strip_native_tool_messages_coalesces_adjacent_users` at compatible.rs:5860, `strip_native_tool_messages_drops_internal_pruning_markers_before_coalescing` at compatible.rs:5888) are precise — they pin the role sequence with `assert_eq!(roles, vec!["user"])` and assert content preservation/concatenation, not just shape. The coalesced-user test specifically checks both the original prompt and the continuation survive the merge.

### 🟢 What looks good — comment update matches what the code actually does now

The block comment at compatible.rs:2162-2178 is rewritten honestly: the prior "Coalesce adjacent assistant messages" header would have been a lie after this PR. The new comment explains *why* same-role coalescing matters (HTTP 400 from native-tool-disabled Anthropic-compat backends), names the targeted backends (Anthropic upstream, MiniMax, other OpenAI-compat wrappers), and gives a concrete trace. Per FND-005's "Be specific" review-voice norm, that's the right level of detail for a comment that future contributors will read alongside a real production incident report. The doc comment on `strip_native_tool_messages` itself (lines 2095-2115) is unchanged but still accurate.

### 🟢 What looks good — uses canonical marker predicates instead of re-deriving them

The new helper at compatible.rs:2119-2131 calls `ChatMessage::is_pruned_tool_exchange_summary()` and `ChatMessage::is_pruned_context_separator()` rather than re-implementing the prefix/suffix match. Those predicates live in `zeroclaw-api/src/model_provider.rs:70-80` alongside the constructors that emit the markers, so the matcher and emitter share one source of truth for the marker contract. Per FND-006 §4.4 (technical-debt triage / contracts as APIs), that's the right boundary — the legacy-marker compatibility note already attached to `is_pruned_tool_exchange_summary` (see history_pruner.rs:836-852) automatically covers this surface too.

### 🟡 Warning — `should_skip_internal_pruning_marker` is now duplicated byte-for-byte with `anthropic.rs:461-473`

This is the only finding I think you should look at before merging. The PR copies the 11-line helper verbatim from `crates/zeroclaw-providers/src/anthropic.rs:461-473`, which you introduced in 883164eec3 (#7775) on 2026-06-17. Both copies:

- have the identical signature `fn should_skip_internal_pruning_marker(messages: &[ChatMessage], index: usize) -> bool`
- have no provider-specific state — they only call `ChatMessage::is_pruned_tool_exchange_summary` and `ChatMessage::is_pruned_context_separator` plus standard slice ops
- encode the same contract: "drop a tool-exchange summary unconditionally; drop a separator only when paired with a preceding summary"

`git diff crates/zeroclaw-providers/src/anthropic.rs:461..473 crates/zeroclaw-providers/src/compatible.rs:2119..2131` would produce zero functional diff — the only difference is the surrounding `impl` block.

Per AGENTS.md, "no piece of state lives in two places, ever" is the absolute rule. The text is about state, but the reasoning ("every duplicate state breeds a drift bug whose symptoms surface months later in production — operator edits the canonical location, the cached copy serves stale data") applies one-to-one here: if the marker contract changes (e.g., a new marker shape, a different pair-detection rule, a third marker added), one of these two copies will be updated and the other will silently serve stale logic until a user-visible regression points us back at the second copy. That is exactly the failure mode the rule exists to prevent.

Two reasonable extraction options:

- **Option A — method on `ChatMessage`.** Move the helper into `zeroclaw-api/src/model_provider.rs` as `pub fn should_be_skipped_as_internal_marker(messages: &[Self], index: usize) -> bool` (or similar). The operation is purely about `ChatMessage` shape; nothing provider-specific. Both callers shrink to `if ChatMessage::should_be_skipped_as_internal_marker(messages, index) { ... }`.
- **Option B — shared helper in `zeroclaw-providers`.** Add `pub(crate) fn should_skip_internal_pruning_marker(messages: &[ChatMessage], index: usize) -> bool` in `crates/zeroclaw-providers/src/lib.rs` (or a small new module like `history_normalization.rs`). Both `anthropic.rs` and `compatible.rs` call the shared one.

Option A is the more natural fit per FND-001 §3 (microkernel — keep contracts at the lowest crate that owns the type), and it makes the marker contract one symbol away from the marker constructors. Option B is mechanically cheaper. Either works. What's not OK is shipping two copies and trusting them to stay in sync by hand.

I'd like to see this extracted in the same PR, since the second copy is what makes the duplication visible. Same author, same week, same file boundary — this is the cleanest moment to do it. If you'd rather land the user/separator regression first and extract in an immediate follow-up, I can live with that, but please file the extraction issue and link it from the PR body so it doesn't drift into "we'll get to it eventually" land.

### 🔵 Suggestion — note the contract about "separator drop requires preceding summary in input" somewhere durable

This is non-blocking and you may already think it's overkill. The asymmetry in the helper is subtle and load-bearing: a summary is dropped unconditionally; a separator is dropped only when paired. The current code carries that contract only in the structure of the `if` / `&&` chain — there is no comment explaining why the conditional, and the docstring on `strip_native_tool_messages` (compatible.rs:2098-2115) doesn't mention markers at all.

Two low-cost places to durably encode it:

- A one-paragraph doc comment on `should_skip_internal_pruning_marker` itself (in whichever location you settle on for the extraction): "Summaries are dropped unconditionally because they would otherwise appear as the model's own reply on the wire. Separators are dropped only when they directly follow a summary in the input; a stray separator without a preceding summary is preserved as a user message so we don't accidentally drop a legitimately-encoded user turn."
- Or a `// Drop both markers when they appear as the summary→separator pair the history pruner emits.` comment at the call site.

Either is fine; nothing on the wire today depends on this comment existing, but it's the kind of subtlety that catches re-readers a year later when they're trying to extend the helper.

### 🟢 What looks good — Validation Evidence section is concrete and complete

The PR body has literal `cargo test -p zeroclaw-providers strip_native_tool_messages --lib` output (8/8 pass), workspace `clippy --all-targets --features ci-all -- -D warnings` (clean in 36m55s), and full `nextest run --locked --workspace --exclude zeroclaw-desktop` (9182 pass, 11 skipped in 86s). I reproduced the focused test locally on 71f18630: 8/8 passed in 0.00s, `cargo fmt --all -- --check` clean. CI on this head is green across all 18 checks. That matches the FND-006 §4.3 standard for evidence-before-assertion.

### Milestone

Moved to v0.8.2 — the v0.8.1 merge window closed before this PR landed. v0.8.2's scope description is "skills + WASM plugin surface, A2A, first-party and pluggable skill registries", which is not a literal scope match for a provider/history-normalization break-fix, but per the review protocol's decision tree (`fix:` prefix + `bug` label + no scope match against any open milestone → earliest open milestone by version), v0.8.2 is now the earliest open milestone and the right home. Tracker entry will land in the v0.8.2 release tracker rather than #6970 once that body is updated.

### 🟢 What looks good — branch state and merge

`mergeable = MERGEABLE`, `mergeStateStatus = BLOCKED` (which here is just `reviewDecision = REVIEW_REQUIRED`; this verdict resolves it pending the other two requested reviewers). Simulated `git merge upstream/master` locally against 71f18630: 7 files touched by intervening master commits (apps/zerocode/src/chat.rs, loop_.rs, tool_execution.rs, the four turn/* modules) auto-merge cleanly with no conflicts in `compatible.rs`. The branch is currently behind master but the gap is mechanical — no rebase needed before merge unless you want fresh CI on top of #7826.

---

Recommending `comment` rather than `approve` so the deduplication of `should_skip_internal_pruning_marker` doesn't get lost in the merge-ready glow. The fix itself is correct, surgical, and well-tested; I want one more pass that either extracts the helper in this PR or files an immediate follow-up before this lands. Happy to flip to `approve` as soon as the extraction is in or the issue is filed and linked.

Audacity88:
@tidux thanks for the detailed pass. I pushed `151945ef4c` to take Option A: the pruning-marker skip rule now lives on `ChatMessage`, both provider-local copies are gone, and the helper doc comment records the summary/separator rule.

Local focused checks passed on the new head:

- `git diff --check`
- `scripts/zc-cargo fmt -p zeroclaw-api -p zeroclaw-providers -- --check`
- `scripts/zc-cargo test -p zeroclaw-providers strip_native_tool_messages --lib`
- `scripts/zc-cargo test -p zeroclaw-providers context_separator --lib`

GitHub checks are now green on `151945ef4c`, including `Test`, `Lint`, `Check (all features)`, Windows, and `CI Required Gate`. Could you take another look?

WareWolf-MoonWall (approved):
# PR #7931 — fix(providers): coalesce stripped compatible history roles

**Head:** 151945ef4cb95a37f1e3306bee16cb2020a0e1b7
**CI:** all green
**Verdict:** approve (re-review after tidux feedback at earlier head)
**Milestone:** v0.8.3

---

### 🟢 Option A is the correct structural choice

Promoting `should_skip_internal_pruning_marker` from a private associated function on `AnthropicModelProvider` to a `pub` method on `ChatMessage` in `zeroclaw-api` eliminates two divergent private copies — one that lived in `anthropic.rs`, one that was needed in `compatible.rs`. The rule now lives at the type level where it belongs. This is a genuine SSOT improvement, not cosmetic consolidation.

### 🟢 Regression tests are directly load-bearing

`strip_native_tool_messages_coalesces_adjacent_users` is the exact guard for the adjacent-user failure mode reported in #7804. `strip_native_tool_messages_drops_internal_pruning_markers_before_coalescing` proves that markers are stripped *before* the coalesce pass — order matters here, and the test would catch a future reordering that reintroduces the leak.

### 🟢 Non-obvious invariant is documented where it lives

The comment on `should_skip_internal_pruning_marker` explains that separators drop only when directly following a summary — that's the invariant that bit both provider implementations independently, and it's exactly the kind of rule that must be co-located with the predicate rather than buried in a PR description.

### 🔵 API evolution note for v1.0.0 stabilization

`zeroclaw-api` carries Experimental stability tier, so the tradeoff here — adding a `pub` method that encodes provider-specific pruning behavior onto a generic `ChatMessage` type — is reasonable and clearly better than the two-copy alternative. Worth a line in the API evolution notes when the stabilization pass happens, to consciously decide whether `should_skip_internal_pruning_marker` belongs on the type permanently or gets moved behind a seam at that point.

---

tidux's technical concerns are fully addressed at this head. Template complete, CI green. Approving.

tidux (approved):
Re-review at `151945ef4cb95a37f1e3306bee16cb2020a0e1b7` — one new commit on top of the head I last looked at (`71f1863022`):

- `151945ef4c` — fix(providers): share internal pruning marker skip rule

This is exactly Option A from the prior pass, and it's the right execution. Both of my prior findings (the 🟡 duplication and the 🔵 doc-comment) are closed in a single commit, with a tighter shape than I asked for.

### ✅ Resolved — `should_skip_internal_pruning_marker` extracted onto `ChatMessage`

`crates/zeroclaw-api/src/model_provider.rs:88-98` now hosts the canonical `pub fn should_skip_internal_pruning_marker(messages: &[Self], index: usize) -> bool` immediately after `is_pruned_tool_exchange_summary` and `is_pruned_context_separator` — the marker predicates and the skip rule that consumes them now live colocated on the type that owns the marker contract. Both provider-local copies are gone: the deletion at `anthropic.rs:461-473` and the deletion at `compatible.rs:2119-2131` are byte-for-byte the same 13-line block, and both call sites (`anthropic.rs:466`, `compatible.rs:2124`) now route through `ChatMessage::should_skip_internal_pruning_marker(messages, index)`. Walked the providers tree for any third copy — there is none.

This matches the FND-001 §3 microkernel reading exactly: the contract for "what counts as an internal pruning marker pair" now lives at the lowest crate that owns the underlying type (`zeroclaw-api` owns `ChatMessage`), and provider crates consume it as callers. Future marker additions or pair-detection refinements have one update site instead of two with a drift window between them.

Net code shape: +20 lines on `ChatMessage` (helper + doc comment), −13 lines in `anthropic.rs`, −13 lines in `compatible.rs`. Diff is +22/−30 across 3 files for the dedup pass on top of the original break-fix — small, surgical, and contained to exactly the surface the prior review identified.

### ✅ Resolved — doc comment landed on the helper, and it's better than what I asked for

The doc block at `model_provider.rs:81-86` explains *both* halves of the asymmetry that the structural `if`/`&&` body encodes:

```rust
/// Summaries always drop because they would otherwise reach the model as its
/// own prior reply. Separators only drop when they directly follow a summary
/// in the input, so a stray separator-shaped user turn is preserved instead
/// of silently discarding possible user content.
```

That's exactly the failure-mode framing a re-reader needs: it names *why* summaries are unconditional (wire-leak the `[Tool exchange: N tool call(s) — results collapsed]` prefix as the model's own reply) and *why* separators are conditional (silent data loss on a stray separator-shaped user message in resumed history). The comment will survive marker-shape evolution because it's encoded in terms of the contract, not the current PRUNED_*_PREFIX constants.

### 🟢 What looks good — minimal-surface dedup that doesn't ripple

I've seen extractions like this turn into a refactor that touches every neighbor — `ChatMessage` constructors, marker predicate names, helper visibility, neighboring helpers that "might as well move too." This one stays disciplined: the helper moves, the doc comment lands, the two call sites swap to the new path, nothing else changes. `is_pruned_tool_exchange_summary` and `is_pruned_context_separator` keep their names, signatures, and visibility; no neighbor was incidentally rewritten. That's the kind of dedup that's safe to land without a full provider crate re-read.

### 🟢 What looks good — visibility chosen at the right level

`pub fn` rather than `pub(crate) fn` is correct: `zeroclaw-api` is shared across all provider crates and any future provider that emits or consumes the internal pruning markers needs access. `pub` here documents the helper as part of the cross-crate `ChatMessage` contract surface. The signature is also the right shape (`&[Self], usize`) — calling it from inside `iter().enumerate().filter_map` doesn't need to borrow `messages` separately, and the bounds-check inside the helper means callers don't have to range-validate before invoking.

### 🔵 Observation only — no direct unit test on the lifted helper in `zeroclaw-api`

The contract is fully exercised through the existing provider integration tests (`strip_native_tool_messages_drops_internal_pruning_markers_before_coalescing` at `compatible.rs:5888`, `convert_messages_keeps_real_context_separator_between_assistants` and `convert_messages_does_not_surface_context_separator_as_user_instruction` in `anthropic.rs`), which together pin all four corners of the truth table at the actual wire-conversion path. So there isn't a coverage gap, just an asymmetry from how the helper was lifted. A direct unit test on `ChatMessage::should_skip_internal_pruning_marker` next to the marker predicates would be a nice-to-have for documenting the contract in the same crate that defines it, but it's not blocking — the higher-confidence tests are the ones already in place. Mentioning it purely so the option is on record for a follow-up.

---

Verification locally on `151945ef4c`:

- `cargo test -p zeroclaw-providers --lib strip_native_tool_messages` → `8 passed; 0 failed` (matches the literal output in PR body and the prior head).
- `cargo test -p zeroclaw-providers --lib context_separator` → `2 passed; 0 failed` (`convert_messages_keeps_real_context_separator_between_assistants` + `convert_messages_does_not_surface_context_separator_as_user_instruction`).
- Walked all three references to `should_skip_internal_pruning_marker` across the providers crates — definition site at `zeroclaw-api/src/model_provider.rs:88`, call site at `anthropic.rs:466`, call site at `compatible.rs:2124`. No third copy lurking.
- `git merge-tree $(git merge-base HEAD upstream/master) HEAD upstream/master` → zero conflict markers against current master (1a3fdb9b95 includes intervening master commits since the prior re-review window).
- 18/18 CI checks pass on the head.

Approving. Thanks for taking the cleaner path on the dedup — having the skip rule co-located with the marker predicates and constructors on `ChatMessage` makes the contract obvious to the next reader who walks in mid-debugging.

singlerider (approved):
I read the diff at head 151945ef and walked the coalescing helper plus the shared marker-skip rule across `compatible.rs`, `anthropic.rs`, and `model_provider.rs`. @tidux and @WareWolf-MoonWall both approved at this exact head after the Option-A follow-up commit, and I agree with that read. CI is green. No new blockers from me.

### 🟢 What looks good: coalescing now matches the real failure shape

Widening the merge from "adjacent assistant only" to "adjacent same-role non-system" is the correct fix for #7804. When a `tool` message is stripped out of prompt-guided history, the two surrounding `user` messages become adjacent, and an OpenAI-compatible endpoint rejects that with "roles must alternate." `strip_native_tool_messages_coalesces_adjacent_users` pins exactly that case (user/tool/user → single user), so the regression is anchored to the observed break, not a synthetic one.

### 🟢 What looks good: the marker-skip rule is shared, not duplicated

`should_skip_internal_pruning_marker` lives once on `ChatMessage` and is called from both the compatible and anthropic conversion paths. That is the right place for it: the "is this an internal pruning marker that must never reach a provider payload" judgment is a property of the message type, and having both providers defer to the same predicate means the two paths cannot drift on what counts as internal context-separator text. This is the Option-A consolidation @tidux asked for, and it reads as genuinely shared rather than copy-pasted.

### 🟢 What looks good: the ordering (skip markers, then coalesce) is correct

Dropping the internal pruning markers before coalescing is the load-bearing ordering: otherwise a separator marker sitting between two user messages would block the merge and then get shipped as user text. `strip_native_tool_messages_drops_internal_pruning_markers_before_coalescing` proves the marker-then-real-user shape collapses to a single user message, which is what keeps internal scaffolding off the wire.

Approving.

PR #7942, opened 2026-06-18
fix(memory): decouple embedding key from chat provider; survive embed failures

mov-xound-glitch:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The memory embedder had no API key of its own. The gateway seeded it from the first `[providers.models]` entry, so the embedding credential silently rode on whatever model provider happened to be first. Switch the chat model to a provider that carries no usable embedding key (e.g. an OAuth-only provider) and the embedding endpoint gets a non-matching/empty credential and 404s.
  - Worse, `store` computed the embedding *before* the INSERT and propagated the error, so a failed embed threw away the whole write. A transient credential fault became silent, permanent data loss for as long as it lasted.
  - Fix 1 — add an optional `[memory].embedding_api_key`. When set it is used for embedding calls (precedence: per-route override > `memory.embedding_api_key` > inherited seed-provider key). When unset, resolution is byte-for-byte unchanged. Embeddings can now keep their own credential independent of the chat model.
  - Fix 2 — `store_with_agent` degrades gracefully: on an embed failure it persists the row with a NULL vector and logs a recoverable WARN instead of aborting. `zeroclaw memory reindex` backfills NULL embeddings from the retained `content` once the embedder is healthy. A 404/401/rate-limit/outage now costs recall quality, not data.
- **Scope boundary:** Does NOT add the embedding-identity marker / auto-migrate-on-model-change feature, and does NOT fix the separate `[[embedding_routes]]` → `NoopEmbedding` degradation (dotted provider refs validate but fall through `create_embedding_provider`). Both are real and tracked as follow-ups, kept out of here to stay one-concern.
- **Blast radius:** `crates/zeroclaw-config` (one optional, default-`None` field) and `crates/zeroclaw-memory` (key resolution + the store path). The store change touches every memory write; behavior is unchanged on the success path and only diverges when an embed call fails (previously: error out; now: persist without vector + WARN). No gateway/runtime signature changes.
- **Linked issue(s):** Related #7948 (persist embedding identity + auto-migrate on model change) and Related #7949 (`[[embedding_routes]]` silently degrades to NoopEmbedding) — both spun off from this PR's scope boundary, neither blocking.
- **Labels:** suggested `type: bug`, `risk: medium`, `size: S`.

## Validation Evidence

Local lint is scoped to the touched crates. The whole-workspace `cargo test` / `clippy --all-targets` is not run locally: some GUI-dependent crates fail to build in this environment (missing GTK/cairo system libs), and the project convention is to scope local lint to edited crates and let CI run the authoritative whole-workspace fmt + clippy + test.

```
$ cargo fmt --all -- --check
(clean, no output)

$ cargo clippy -p zeroclaw-config -p zeroclaw-memory --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 20.97s
(no warnings)

$ cargo test -p zeroclaw-memory -p zeroclaw-config
test result: ok. 663 passed; 0 failed; 0 ignored   (zeroclaw-config)
test result: ok.  88 passed; 0 failed; 0 ignored   (zeroclaw-config)
test result: ok. 312 passed; 0 failed; 0 ignored   (zeroclaw-memory)
test result: ok.   4 passed; 0 failed; 0 ignored   (zeroclaw-memory integration)
```

- **Beyond CI — what I manually verified:** Reproduced the original failure against a live instance whose chat provider had been switched away from the embedding provider — `store` failed with `Embedding API error 404 Not Found` and the row count did not move (write discarded). Confirmed the same embedding endpoint + model + correct key returns 200, isolating the cause to the credential source rather than the endpoint. Added a unit test with a failing embedder (real dimension, always errors) proving the row now persists with a NULL vector instead of the store aborting, plus tests covering the key-precedence ladder (route override > memory key > inherited, blank override ignored, OAuth-only-no-inherited-key case).
- **Not verified:** end-to-end `reindex` backfill on a large production store; `GET /api/config` masking of the new field is covered by the `#[secret]` machinery but not exercised here.

## Security & Privacy Impact

- New external network calls? **No** (the embedding call already existed; this only changes which key it sends)
- Secrets / tokens / credentials handling changed? **Yes** — adds optional `[memory].embedding_api_key`, marked `#[secret]` so it is masked in config dumps and encrypted at rest the same way provider keys are.
- PII or real identities in the diff/tests/fixtures? **No**

## Compatibility

- Backward compatible? **Yes** — the field is optional and defaults to `None`; unset means the prior inherited-key behavior is preserved exactly.
- Config / env / CLI surface changed? **Yes** — new optional `[memory].embedding_api_key`. No removals, no renames.
- Upgrade steps for existing users: none required. Anyone whose embeddings broke after switching chat providers can set `[memory].embedding_api_key` to the embedding provider's key, then run `zeroclaw memory reindex` to backfill any rows that failed to embed while it was broken.

## Rollback

- **Fast rollback:** `git revert <sha>` — self-contained, no migrations or schema changes to unwind.
- **Feature flags / config toggles:** the new field is opt-in; leaving it unset is equivalent to pre-PR behavior, so no toggle is needed to neutralize it.
- **Observable failure symptoms:** grep memory logs for `memory store: embedding failed; persisting row without a vector` (the new graceful-degrade WARN) to spot an embedder that is down; rows accumulating with NULL embeddings indicate a backfill is needed via `reindex`.


mov-xound-glitch:
Filed the two follow-ups noted in the scope boundary:

- #7948 — persist the embedding identity in the DB and auto-migrate vectors when the embedding model/provider/dimensions change (this PR doesn't cover the model-change case).
- #7949 — `[[embedding_routes]]` silently degrades to `NoopEmbedding` for any dotted provider ref (separate latent bug found while diagnosing this).

Both are Related, not blockers.

WareWolf-MoonWall (approved):
Reviewed at 27792243cdcd87b34b711a48f1e8bfa97a960937. CI Required Gate: green (all 18 checks passing).

### 🟢 What looks good — two independent bugs fixed cleanly in one PR

The two issues are genuinely coupled (both stem from the embedding credential gap that surfaced when operators switch chat providers): fixing the key source without also making the write durable would still cause silent data loss during an outage, and vice versa. Keeping them together is the right call.

### 🟢 What looks good — key precedence ladder is explicit and well-tested

The comment block in `resolve_embedding_config` names the precedence clearly:

1. per-route `api_key` override
2. `[memory].embedding_api_key`
3. seed model provider's key (inherited)

The five unit tests cover every rung of that ladder: override beats memory key, memory key beats inherited, blank memory key is treated as unset, route key without its own key falls through to memory key, and the OAuth-only-no-inherited-key case. That is thorough and directly load-bearing.

### 🟢 What looks good — graceful degrade is correctly scoped

The `store_with_agent` change in `sqlite.rs` degrades on embedding failure (persist with NULL vector + WARN log) rather than aborting. The WARN message tells operators exactly what happened and names the recovery path (`zeroclaw memory reindex`). The `FailingEmbedding` test proves the row survives when the embedder always errors. This is the correct tradeoff: recall quality degrades gracefully under an outage instead of silently losing data.

### 🟢 What looks good — `#[secret]` and `#[credential_class = "encrypted_secret"]` are correctly applied

The new `embedding_api_key` field follows the established pattern for encrypted secrets in the config (matching, e.g., the `api_key` fields on provider configs). The `#[cfg_attr(feature = "schema-export", schemars(extend("x-secret" = true)))]` annotation is present. This ensures the key is masked in config dumps and encrypted at rest via the same machinery as existing provider keys.

### 🟡 Warning — full workspace `cargo test` not run locally

The validation section explains why (missing GTK/cairo system libs in the build environment) and notes CI covers the full workspace. This is accepted given the narrow scope of the change (two crates: `zeroclaw-config` and `zeroclaw-memory`) and the CI Required Gate being green. The targeted per-crate runs are directly relevant to the changed code.

### 🟡 Warning — `GET /api/config` masking of the new field is acknowledged but not tested here

The PR body notes this is covered by the `#[secret]` machinery but not exercised in this PR's test suite. Given that the existing machinery is well-established and the field annotation matches the pattern used elsewhere, this is not a blocker — but a follow-up integration test for the config dump endpoint would be a good addition to the `zeroclaw-config` test suite.

### 🔵 Suggestion — link the two follow-up issues (#7948, #7949) in the PR body

The author commented that #7948 (embedding identity persist / auto-migrate on model change) and #7949 (`[[embedding_routes]]` NoopEmbedding degradation) were filed as follow-ups. Adding `Related #7948` and `Related #7949` to the PR body's linked-issues bullet makes those connections discoverable for milestone triage.

This is a clean, well-scoped, well-tested bug fix. Approving.

Audacity88 (approved):
Reviewed current head `27792243cdcd87b34b711a48f1e8bfa97a960937` against the live PR state, linked follow-ups #7948/#7949, the config/memory diff, public source excerpts, prior same-head approval, and green CI. This is a confirming second approval; I do not see a blocker from my side.

### 🟢 What looks good — memory writes now survive embedding outages

The `store_with_agent` path now preserves the memory row when embedding fails and leaves recovery to `zeroclaw memory reindex`. That moves the failure mode from data loss to degraded semantic recall, which is the right tradeoff for a transient provider/key outage.

### 🟢 What looks good — credential precedence stays explicit and backward-compatible

The new `[memory].embedding_api_key` sits between per-route overrides and the inherited seed-provider key, and blank values remain ignored. The tests cover each rung of that precedence ladder, so operators can decouple embedding credentials without changing existing unset configurations.

### 🟢 What looks good — the remaining embedding identity/route bugs are correctly split out

#7948 and #7949 cover the model-identity migration and dotted-route degradation issues without expanding this patch beyond the data-loss fix. That keeps this PR reviewable while preserving clear follow-up ownership.

### 🔵 Suggestion — refresh the merge result before landing

GitHub currently reports the PR as `CLEAN`, and I do not see this as a patch blocker. As merge hygiene, because this head is behind current `master` and `master` has also touched `crates/zeroclaw-config/src/schema.rs` since the PR's merge base, I would prefer a branch refresh with fresh CI, or equivalent merge-result validation, before landing.

Issue #18, opened 2026-02-14
[LOW] HEARTBEAT.md created with default permissions (world-readable)


vrescobar:
## Severity: LOW

**File:** `src/heartbeat/engine.rs:88-102`

### Description

The `HEARTBEAT.md` file is created with the process's default umask (typically 022, resulting in 644 = world-readable):

```rust
pub async fn ensure_heartbeat_file(workspace_dir: &Path) -> Result<()> {
    let path = workspace_dir.join("HEARTBEAT.md");
    if !path.exists() {
        let default = "# Periodic Tasks\n...";
        tokio::fs::write(&path, default).await?;
    }
    Ok(())
}
```

### Impact

Low — `HEARTBEAT.md` contains periodic task schedules which could reveal information about the agent's behavior patterns to other local users. The secret key file (`secrets.rs`) correctly sets 0600, but this file does not.

### Suggested Fix

Set explicit restrictive permissions after creation:

```rust
tokio::fs::write(&path, default).await?;

#[cfg(unix)]
{
    use std::os::unix::fs::PermissionsExt;
    tokio::fs::set_permissions(&path, std::fs::Permissions::from_mode(0o600)).await?;
}
```

---

> **Disclaimer:** This analysis is human-driven and AI-assisted. The findings were identified through a systematic code review using AI tooling, then validated and curated by a human reviewer. Manual verification is recommended before implementing fixes.

vrescobar:
Closing: this was filed as a public issue in error. Security vulnerabilities should be reported via [GitHub Security Advisories](https://github.com/theonlyhennygod/zeroclaw/security/advisories/new) per SECURITY.md. Will re-submit through the proper private channel.

narhc8:
## v2 Architecture Amendment: Activity Feed & Structured Audit Logging

> This issue is being expanded per the **Architecture Amendment v2** (`docs/architecture-amendment-v2.md`). The original security finding (HEARTBEAT.md permissions) is superseded; Activity Feed & Structured Audit Logging is the new scope.

---

### New Title: **Epic: Activity Feed & Structured Audit Logging**

The Activity Feed is ZeroClaw's unified event logging system — every agent action, tool call, task update, and system event gets persisted to the SQLite `activities` table and streamed in real-time to Mission Control via SSE. This **replaces the observer-only approach** with a full persistence layer that serves as both the real-time activity feed and the structured audit log.

---

### How Events Are Captured (Runtime-Specific, Output-Identical)

[...]

Issue #13, opened 2026-02-14
[MEDIUM] Skill installation creates symlinks without path validation


vrescobar:
## Severity: MEDIUM

**File:** `src/skills/mod.rs:289-303`
**CWE:** [CWE-22](https://cwe.mitre.org/data/definitions/22.html) — Path Traversal

### Description

Skill installation creates symlinks from user-supplied paths without validating the source location:

```rust
let src = PathBuf::from(&source);
if !src.exists() {
    anyhow::bail!("Source path does not exist: {source}");
}
let name = src.file_name().unwrap_or_default();
let dest = skills_path.join(name);

#[cfg(unix)]
std::os::unix::fs::symlink(&src, &dest)?;
```

### Attack scenario

```bash
zeroclaw skills install ../../etc
# Creates ~/.zeroclaw/workspace/skills/etc → ../../etc (relative symlink)
```

The agent could then follow this symlink to read sensitive system files if the skill scanner traverses installed skill directories.

### Impact

Requires local CLI access (the user runs the command themselves), but could be exploited through social engineering ("install this skill from this path").

### Suggested Fix

Canonicalize the source path and validate it doesn't point to sensitive locations:

```rust
let src = PathBuf::from(&source).canonicalize()
    .map_err(|_| anyhow::anyhow!("Invalid or inaccessible path: {source}"))?;

// Optionally: reject paths outside home directory or known safe locations
let home = dirs::home_dir().unwrap_or_default();
if !src.starts_with(&home) {
    anyhow::bail!("Skill source must be within your home directory");
}
```

---

> **Disclaimer:** This analysis is human-driven and AI-assisted. The findings were identified through a systematic code review using AI tooling, then validated and curated by a human reviewer. Manual verification is recommended before implementing fixes.

vrescobar:
Closing: this was filed as a public issue in error. Security vulnerabilities should be reported via [GitHub Security Advisories](https://github.com/theonlyhennygod/zeroclaw/security/advisories/new) per SECURITY.md. Will re-submit through the proper private channel.

PR #7946, opened 2026-06-18
feat(runtime): add model context window ctx bar to zerocode tui, gateway agent chat, and command line interactive mode.

eugeneb50:
- **What changed and why:** Added ctx usage bar to chat surfaces zerocode TUI, gateway Chat, CLI interactive using config.toml as single source of truth. 
  - Added `context_window` field to `ModelProviderConfig` schema — single source of truth for each model's max input tokens (provider config → 32k fallback)
  - Implemented `fetch_context_window()` for 9 providers (OpenRouter + 8 OpenAI-compatible: Together, Groq, Fireworks, DeepInfra, Hyperbolic, Anyscale, Novita, Nebius) — auto-populates config on quickstart/doctor
  - Added `zeroclaw doctor update-context-windows` CLI command with `--model-provider` and `--dry-run` — manual refresh for operators
  - Added `POST /api/config/model-providers/{type}/{alias}/refresh-context-window` Gateway API endpoint — programmatic refresh
  - Gateway WebSocket now resolves `max_context_tokens` via `effective_model_context_window()` and includes it in `done` frame — web UI can show context bar
  - Web UI (`AgentChat.tsx`) added `ContextBar` component displaying usage from `done` frame
  - CLI interactive mode (`agent::run`) prints per-turn context summary to stderr: `ctx: 4,231 / 200,000 ████░░░░░░░░░░ 2%`
  - Zerocode TUI enabled `CtxBar` widget (`SHOW_CTX_BAR = true`) — live context bar via existing RPC `ContextUsage` events
  - Agent loop and RPC dispatcher now use `model_context_window` (provider config) instead of `max_context_tokens` (runtime profile) — eliminates drift between surfaces

- **Scope boundary:** This PR does NOT change:
  - Config schema beyond the optional `context_window` field (existing configs work unchanged)
  - Real-time ContextUsage streaming on Gateway WS (done-frame only per design)
  - Single-shot CLI mode (`zeroclaw agent -a x "msg"`) — context summary only prints in interactive mode
  - Provider auto-detection beyond the 9 currently supported providers
  - Context compression/truncation logic (unchanged)

- **Blast radius:**
  - `crates/zeroclaw-config` — new field + resolver (path: `config`)
  - `crates/zeroclaw-providers` — new fetch functions (path: `provider`)
  - `crates/zeroclaw-runtime` — doctor, quickstart, agent loop, RPC dispatch (paths: `runtime`, `doctor`, `quickstart`)
  - `crates/zeroclaw-gateway` — API endpoint + WS frame (path: `gateway`)
  - `web` — ContextBar component + WS type definitions
  - `apps/zerocode` — TUI status bar enablement

- **Linked issue(s):** Related to context window visibility gap across surfaces

- **Labels:** `enhancement`, `dependencies`, `core`, `agent`, `config`, `doctor`, `gateway`, `provider`, `runtime`, `quickstart`, `risk:high`, `size:L`

---

```bash
cargo fmt --all -- --check
# (no output = clean)

# Checking zeroclawlabs v0.8.0
# Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.76s

cargo test --locked
# test result: ok. 279 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# test result: ok. 188 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# test result: ok. 159 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
# test result: ok. 0 passed; 0 failed; 7 ignored; 0 measured; 0 filtered out
```

- `zeroclaw doctor update-context-windows --dry-run` shows correct dry-run output for configured providers
- `zeroclaw quickstart` auto-populates `context_window` in config.toml for OpenRouter provider
- Gateway WS `done` frame includes `max_context_tokens` field
- Web UI ContextBar renders after turn completion
- CLI interactive mode prints context summary to stderr (not stdout)
- TUI context bar renders live during daemon-connected sessions
- Single-shot CLI mode (`zeroclaw agent -a assistant "hello"`) produces NO context summary output

**If any command was intentionally skipped, why:**
- `cargo test --test live` — requires live credentials (ignored per design)
- Docs quality/link gates — no markdown docs changed in this PR

---

- New external network calls? **Yes** — `fetch_context_window()` makes HTTP GET to provider `/models` endpoints (OpenRouter, Together, Groq, etc.) during quickstart, doctor, and the gateway refresh endpoint. Mitigation: only runs on explicit user action (quickstart setup, doctor command, or API call); fails gracefully with `None` → falls back to 32k.
- Secrets / tokens / credentials handling changed? **Yes** — the refresh paths read existing provider credentials from config/env and forward them only to the configured provider `/models` endpoint. Responses keep the secret out of returned bodies.

---

## Compatibility (required)

- Backward compatible? **Yes** — `context_window` is optional; all surfaces fall back to 32,000 if unset
- Config / env / CLI surface changed? **Yes** — new optional config field `providers.models.<type>.<alias>.context_window`; new doctor subcommand `update-context-windows`; new Gateway API endpoint

**Exact upgrade steps for existing users:**
1. No action required — existing configs work (fallback = 32,000)
2. Optional: run `zeroclaw doctor update-context-windows` to auto-populate context windows for supported providers
3. Optional: run `zeroclaw quickstart` for new providers — auto-populates on setup

---

- **Fast rollback command/path:** `git revert <merge-sha>` — all changes are additive/config-driven with no migrations
- **Feature flags or config toggles:** None (field is optional; TUI controlled by `SHOW_CTX_BAR` const which can be flipped to `false`)
  - Grep logs for `"context_window"` fetch errors: `provider does not expose context window or fetch failed`
  - Metric: `gateway_ws_turn` traces missing `max_context_tokens` in done frame
  - Alert: TUI context bar shows `ctx: 0 / 32000` (fallback) when provider config missing

---

- Superseded PRs + authors: N/A
- `Co-authored-by` trailers added in commit messages for incorporated contributors? **No** — no direct code/design carry-over from other PRs

---

**Privacy contract** (`docs/book/src/contributing/privacy.md`) is a merge gate. Never commit real identities, secrets, personal emails, or PII in diff, tests, fixtures, or docs.

---


WareWolf-MoonWall (changes requested):
Reviewed at df543ae6. Five commits touching `zeroclaw-config`, `zeroclaw-providers`, `zeroclaw-runtime`, `zeroclaw-gateway`, `web`, and `apps/zerocode`. The description is detailed and the template is thoroughly filled — scope boundary, blast radius, security impact, and rollback are all present. The SSOT story is sound: `context_window` lives canonically in `ModelProviderConfig`, `effective_model_context_window()` resolves it at call-time, and the materialized snapshot in `ResolvedRuntime.model_context_window` is populated from the same source. Two blocking items need to be resolved before this is ready.

---

### 🔴 Blocking — CI is failing (6/21 checks)

`Test`, `Build x86_64-unknown-linux-gnu`, `Check (all features)`, and `CI Required Gate` are all red. The local `cargo test --locked` output in the PR body passes, but CI is failing on those same targets. This must be resolved and all required gates green before merging. Nothing else in this review is relevant until the build is clean.

---

### 🔴 Blocking — New user-facing CLI strings bypass the Fluent layer

Per AGENTS.md: "All user-facing output (CLI messages, tool descriptions, onboarding prompts) must use `fl!()` / Fluent strings — never bare string literals."

The PR introduces two groups of bare-string user-facing output:

1. **Context bar in `crates/zeroclaw-runtime/src/agent/loop_.rs`:**
   ```rust
   eprintln!(
       "\x1b[2mctx: {:>7} / {:>7}  {}  {:.0}%\x1b[0m",
       format_tokens(usage.input_tokens),
       format_tokens(max_ctx),
       bar,
       pct
   );
   ```
   This is printed to the user's terminal on every interactive turn. It must go through `fl!()`.

2. **Doctor command output in `crates/zeroclaw-runtime/src/doctor/mod.rs` (`update_context_windows`):**
   All `println!` calls — `"already has context_window = {ctx}"`, `"no model configured, skipping"`, `"would set context_window = {ctx} (dry run)"`, `"Saved {updated} updates to config.toml"`, etc. — are user-facing CLI messages that must use `fl!()`. The new doctor subcommand surfaces count as "CLI messages" under the rule even if they feel operational.

Both groups need Fluent keys registered in all 5 locales (en, es, fr, ja, zh-CN) before this can land.

---

### 🟡 Warning — PR title has trailing whitespace

The title ends with two spaces: `"…interactive mode.  "`. The `check-pr-title.sh` validator does not anchor at `$` so it will not reject the title, but the trailing spaces are visible in the GitHub UI and PR history. Please strip them before merge.

---

### 🟡 Warning — `fetch_openai_compatible_context_window` sends unauthenticated requests

The function makes a plain `client.get(&url).send()` with no `Authorization` header, despite `ModelProviderConfig` carrying the provider's API key. Groq, Fireworks, DeepInfra, and several other providers in the claimed support list require authentication to list models. For those, the fetch will silently fail and fall back to 32 k — making the auto-populate feature effectively non-functional for most of its advertised scope. The graceful fallback is correct, but the PR summary states support for 8 OpenAI-compatible providers when in practice only providers with public model listings (OpenRouter, possibly Together AI) will work.

Please either add the API key header or tighten the advertised scope in the PR description to reflect which providers actually work today.

---

### 🟡 Warning — Inconsistent resolution strategy for `model_context_window`

`crates/zeroclaw-runtime/src/agent/agent.rs` reads `self.config.resolved.model_context_window` (the snapshot baked at agent startup). `crates/zeroclaw-runtime/src/agent/loop_.rs` bypasses that snapshot and re-resolves via `config.effective_model_context_window(agent_alias)` every turn loop iteration. If a live config reload changes `context_window` mid-session, these two paths will diverge: the agent struct uses the stale snapshot, the loop uses the fresh value. Decide on one strategy and apply it consistently. If the intent is to pick up live changes, use the resolver everywhere; if the intent is snapshot-at-startup, use `agent.resolved.model_context_window` everywhere.

---

### 🟢 What looks good — SSOT is correctly implemented

`context_window` lives exactly once, in `ModelProviderConfig` in `zeroclaw-config`. `effective_model_context_window()` resolves from that canonical field at call-time with a hard 32 k fallback. The previous `max_context_tokens` (output budget) and new `model_context_window` (model capacity) are kept clearly separated with doc comments explaining the distinction — that clarification was overdue and removes a long-standing source of naming confusion.

---

### 🟢 What looks good — Template completeness and manual verification

All required sections are present and substantive. The manual verification covers all three surfaces the PR touches (TUI, Gateway WS, CLI interactive mode), including the negative case (single-shot CLI mode produces no context output). The security impact section correctly identifies the new external network calls and explains the mitigation. The rollback section includes an observable failure symptom in the `ctx: 0 / 32000` fallback sentinel.

Audacity88 (changes requested):
Reviewed the current PR head 9fb30ed9 against the live PR state, prior WareWolf-MoonWall review, current public diff, provider/config/gateway paths, and the relevant localization/config/security review guidance. CI is green now, including the Quality Gate jobs, and several prior items moved in the right direction. I did not run additional local Cargo validation for this review pass.

### ✅ Resolved — CI and the first CLI-string pass are no longer the same blockers

The earlier failing Quality Gate is green on the current head. The direct bare `println!`/`eprintln!` strings called out in the first review were also moved behind Fluent keys, and the interactive loop now uses the resolved runtime `model_context_window` snapshot consistently instead of re-resolving a second value during the turn loop.

### 🔴 Blocking — `update-context-windows` can update the wrong provider alias

`update_context_windows` records the selected target as a dotted `provider_ref`, but when it writes the fetched value back it scans every provider and mutates the first entry whose `model` and `uri` match in `crates/zeroclaw-runtime/src/doctor/mod.rs`. That is not the same identity as the selected provider. Two aliases can legitimately share the same model and URI but have different aliases, credentials, fallback settings, or existing `context_window` state. In that case `zeroclaw doctor update-context-windows --model-provider <type.second>` can update the first matching alias instead of the requested one; in the all-providers path it can also update the same first alias twice while leaving the later alias unset.

Please carry the provider family + alias identity through the target list and mutate that exact entry, then add a regression with two provider aliases sharing the same model/URI so the test fails if the command updates by model/URI instead of alias.

### 🔴 Blocking — The gateway refresh endpoint still drops provider credentials

The provider fetch helper now adds bearer auth when `ModelProviderConfig.api_key` is present, which fixes the doctor/quickstart path. The new gateway `handle_refresh_context_window` endpoint still constructs its fetch config with only `model` and `uri`, then calls `fetch_context_window` without the provider's configured credential. That leaves `POST /api/config/model-providers/{type}/{alias}/refresh-context-window` unauthenticated for Groq, Fireworks, DeepInfra, and the other claimed authenticated OpenAI-compatible catalogs, so the API surface can still return the generic fetch-failed error for exactly the providers the PR says it supports.

Please read the configured provider entry by alias and pass its credential field into the temporary `ModelProviderConfig` used for the fetch, without exposing the secret in the response or logs. A focused gateway/handler test with an API-key-backed provider config would be enough to pin this. The PR body should also update the Security & Privacy section: credentials handling now changes because this feature reads and forwards configured provider credentials for model-catalog refreshes.

### 🔴 Blocking — The translated Fluent entries use message references instead of variables

The new English CLI keys use Fluent variables such as `{$used}`, `{$max}`, and `{$ctx}`, but the es/fr/ja/zh-CN entries still use `{used}`, `{max}`, `{provider_ref}`, and `{ctx}`. Those are not external variable references in Fluent. The formatter is called with external args from the CLI/doctor paths, so the non-English messages will fail to format correctly and fall back/miss instead of showing the localized values.

Please update the translated entries to use `{$...}` placeholders consistently and add a small formatter regression for at least one non-English locale with these keys. That keeps the previous CLI-localization blocker actually closed rather than only closed for English.

### 🟢 What looks good — The core shape is still worth keeping

The central direction is good: keeping model context capacity in provider config, separating it from runtime output-token budget, and surfacing the resulting context usage in CLI, gateway chat, and Zerocode are all useful operator-facing improvements. Once the write-back identity, gateway credential propagation, and locale formatting are fixed, the remaining shape looks much easier to trust.

eugeneb50:
## Summary

- **What changed and why:**
  - **Auto-populate `context_window` in quickstart** - Quickstart now fetches context window from provider's `/models` endpoint when creating a new model provider. Previously only `doctor update-context-windows` did this.
  - **Fix empty URI → factory default fallback** - Added `.filter(|s| !s.is_empty() && *s != "<unset>")` to both OpenRouter and OpenAI-compatible fetch functions. Quickstart writes `uri = ""` to config; without this filter, empty string blocks factory default URL fallback (`get_default_url()`).
  - **Fix API key decryption in quickstart** - Changed `api_key` source from `config.get_prop()` (returns encrypted value — decryption only runs at config load) to `choice.fields.get("api_key")` (plaintext before encryption).
  - **Add factory default URL helper** - New `get_default_url(provider_type)` in `factory.rs` maps 9 OpenAI-compatible providers to their `DEFAULT_URL` constants (Groq, Together, Fireworks, DeepInfra, Hyperbolic, Anyscale, Novita, Nebius, NVIDIA).
  - **Gateway endpoint for context window refresh** - New `POST /api/config/model-providers/{type}/{alias}/refresh-context-window` endpoint that reads the provider's credentials and fetches context window.
  - **Doctor command uses exact alias identity** - Fixed `update_context_windows` to write by exact `providers.models.{type}.{alias}` path instead of scanning all entries matching by model+URI. Previously two aliases with same model/URI would cause wrong alias to be updated.
  - **Dependency injection for testability** - Added `FetchContextWindowFn` type alias and optional parameter to `update_context_windows`, enabling unit tests with mock fetch function.
  - **Regression test for multi-alias bug** - Added test creating two `groq` aliases with same model/URI, verifies only targeted alias gets updated. Uses temp directory for config isolation.

- **Scope boundary:** Does NOT change:
  - TUI/Gateway/CLI context bar display (separate feature)
  - Any provider auth logic beyond Bearer token for OpenAI-compatible

- **Blast radius:** `zeroclaw quickstart` (new provider creation), `zeroclaw doctor update-context-windows`, gateway refresh endpoint, 9 OpenAI-compatible providers

- **Linked issue(s):** Related to context window display feature
- **Labels:** `type: feature`, `risk: medium`, `size: M`, `providers`, `quickstart`, `gateway`, `doctor`

---

```bash
cargo fmt --all -- --check
cargo test
```

  ```
  cargo fmt --all -- --check
  # No output = clean
  
  # No output = clean
  
  cargo test
  # running 16 test(s) - quickstart: 16 passed
  # doctor tests: 32 passed
  # all provider tests pass
  ```

  - `cargo run -- quickstart` → Select Groq → `context_window = 131072` auto-populated in config.toml
  - `cargo run -- doctor update-context-windows --dry-run` → All 9 providers show fetched context_window
  - Quickstart tests (16) pass with multi-threaded runtime guard preventing test runtime false positives
  - Empty URI and `<unset>` placeholder correctly fall back to factory default URLs
  - Gateway `POST /api/config/model-providers/groq/alias/refresh-context-window` returns context_window
  - Regression test: two `groq` aliases with same model/URI — only targeted alias updated

- **If any command was intentionally skipped, why:** None skipped

---

- New external network calls? **Yes** — quickstart and doctor make HTTP GET to provider `/models` endpoint during provider creation/refresh (same as existing doctor command). Silent failure falls back to config default.
- Secrets / tokens / credentials handling changed? **No** — uses existing Bearer auth pattern; reads plaintext key from in-memory `choice.fields` before encryption

---

## Compatibility (required)

- Backward compatible? **Yes** — only adds auto-population; existing configs unchanged
- Config / env / CLI surface changed? **No** — no new flags, config keys, or env vars (gateway endpoint is new API surface but doesn't affect CLI)

---

- **Feature flags or config toggles:** None (silent failure is the fallback)
  - `grep "context_window" ~/.config/zeroclaw/config.toml` — missing on new providers
  - `zeroclaw doctor update-context-windows --dry-run` shows `None` for affected providers
  - Gateway `POST /api/config/providers/models/{type}/{alias}/refresh-context-window` returns 400 "fetch failed"

---

- Not applicable

singlerider (changes requested):
Re-reviewed at head `36ac8aef51`. Pulled the branch, built runtime/gateway/providers/config, ran the new doctor test, and checked all three of Audacity88's blockers plus the SSOT consistency item.

### ✅ RESOLVED — write-back now targets the exact provider alias

`update_context_windows` carries `(provider_ref, provider_type, alias, model, uri, ...)` through the target list and writes via `providers.models.{provider_type}.{alias}.context_window`, not a model/URI scan. The new `update_context_windows_uses_exact_alias_not_model_uri` test constructs two aliases sharing the same model/URI and proves the requested alias is the one mutated; it passes locally. This closes the confused-alias path.

### ✅ RESOLVED — the gateway refresh endpoint now forwards the provider credential

`handle_refresh_context_window` reads `{provider_type}.{alias}.api_key` by exact path and passes it into the `ModelProviderConfig` used for `fetch_context_window`, so the authenticated OpenAI-compatible catalogs (Groq, Fireworks, DeepInfra, …) are no longer fetched unauthenticated. The key is used only for the fetch and is not echoed in the response.

### ✅ RESOLVED — translated Fluent entries use `{$var}` consistently

es/fr/ja/zh-CN now use `{$used}`, `{$max}`, `{$provider_ref}`, `{$ctx}` (zero bare `{var}` refs remaining), matching the en keys and the external-args call sites.

### ✅ RESOLVED — single context-window resolution strategy

`loop_.rs` now reads `agent.resolved.model_context_window` (the startup snapshot), matching `agent.rs`, so the snapshot-vs-resolver divergence WareWolf and Audacity88 flagged is gone.

### 🔴 Blocking — two of the three requested regression tests are still missing

The alias write-back test is in. The other two Audacity88 asked for are not, and both guard surfaces that have regressed or carry risk before:

1. A gateway/handler test for `POST .../refresh-context-window` with an API-key-backed provider config, proving the configured credential is forwarded into the fetch (and not leaked in the response). This is the secret-forwarding path; it should be pinned.
2. A formatter regression for at least one non-English locale exercising `cli-agent-context-bar` / `cli-doctor-ctxwin-*` with `{$...}` args, so the locale syntax cannot silently regress again — this exact class of bug was a repeat blocker.

The code for both is correct on inspection and the crates build clippy-clean, but these two pins are cheap and were explicitly requested. Add them and this is ready.

### 🟢 Core shape

SSOT for `model_context_window` in `ModelProviderConfig`, the capacity-vs-output-budget separation, and surfacing ctx usage in CLI/gateway/zerocode are all worth keeping. Build and clippy `-D warnings` clean across runtime/gateway/providers locally.

eugeneb50:
thanks for the careful reviews and suggestions

singlerider (approved):
Re-reviewed at head `4ee17b50`. My prior `CHANGES_REQUESTED` (on `36ac8aef`) had exactly one open blocker: two of the three requested regression tests were missing. Both are now present and I verified them locally against this head. Clearing my review. 🟢

### ✅ RESOLVED — gateway secret-forwarding test added and correct

`api_config::tests::refresh_context_window_forwards_api_key` stands up a wiremock that requires `authorization: Bearer test-api-key-123` (`.and(header(...)).expect(1)`), drives the real `handle_refresh_context_window` handler against a `groq.test` provider config carrying that key, and asserts three things: the response is 200, the API key does **not** appear in the response body (`!body_str.contains("test-api-key-123")`), and the mock received exactly one request whose `authorization` header equals the configured key. That pins the credential-forwarding path AND the no-leak property — exactly what I asked for. Passes locally.

### ✅ RESOLVED — non-English locale formatter regression added and correct

`i18n::tests::non_english_locales_format_context_window_args` iterates `es`/`fr`/`ja`/`zh-CN`, formats all five `cli-doctor-ctxwin-*` keys plus `cli-agent-context-bar` with `{$...}` args, asserts the substituted values appear, and — critically — asserts no bare `{provider_ref}`/`{ctx}` placeholders survive. That directly guards the `{var}` vs `{$var}` syntax class that was a repeat blocker on this PR. Passes locally.

### ✅ Still-resolved items confirmed after the master merges

The merges since my last review (`716825f9`, `4ee17b50`) did not regress the earlier fixes:
- `loop_.rs:1066` still reads `agent.resolved.model_context_window` (the SSOT startup snapshot), so the snapshot-vs-resolver divergence WareWolf and Audacity88 flagged stays closed.
- `doctor::tests::update_context_windows_uses_exact_alias_not_model_uri` (the alias write-back pin) still passes.
- `cargo clippy --all-targets -- -D warnings` is clean across `zeroclaw-runtime`, `zeroclaw-gateway`, `zeroclaw-providers`, and `zeroclaw-config`. All CI checks green.

### 🟢 Verdict

All three of my prior concerns are closed and both requested pins are in. Approving to clear my `CHANGES_REQUESTED`. Note for @WareWolf-MoonWall and @Audacity88: your reviews still show `CHANGES_REQUESTED` on record — the items you raised (SSOT resolution, exact-alias write-back, credential forwarding, locale `{$var}` consistency) are all resolved as of this head and are now pinned by tests. Please dismiss or re-confirm so the merge gate reflects the current state.

Audacity88 (changes requested):
Reviewed current head `4ee17b504c0bb5f1fdcacad03a2b75b6cd29bd72` against the live PR state, formal reviews from @WareWolf-MoonWall and @singlerider, the public diff, the CLI localization path, and the config/provider/gateway/runtime surfaces touched by the patch. I did not run local Cargo validation, a gateway smoke, a doctor CLI smoke, web UI verification, or Zerocode TUI verification for this review pass.

### ✅ Resolved — The prior context-window blockers are pinned now

I agree with @singlerider that the blockers from my prior review are addressed at this head. `update_context_windows` now carries the provider family and alias through the target list and writes `providers.models.{provider_type}.{alias}.context_window`, with `update_context_windows_uses_exact_alias_not_model_uri` covering the two-alias regression. The gateway refresh handler now builds the fetch config with the configured `api_key`, and `refresh_context_window_forwards_api_key` verifies both `Authorization: Bearer ...` forwarding and that the key does not appear in the response body. The non-English Fluent entries now use `{$...}` variables, and `non_english_locales_format_context_window_args` covers the locale formatting path that previously regressed. The loop also uses `agent.resolved.model_context_window`, so the snapshot-vs-resolver split @WareWolf-MoonWall and I called out is closed.

### 🔴 Blocking — The new doctor subcommand help is still outside the Fluent catalogue

The runtime output strings for `update-context-windows` moved behind `cli-doctor-ctxwin-*` keys, but the PR also adds a new user-facing Clap subcommand:

```rust
/// Update context_window in config.toml from provider /models endpoints
UpdateContextWindows {
    /// Update a specific model_provider only (default: all known model_providers)
    #[arg(long)]
    model_provider: Option<String>,

    /// Show what would be updated without writing to config
    #[arg(long)]
    dry_run: bool,
},
```

`apply_cmd_translations` recursively looks for `cli-doctor-update-context-windows-about` when translating `zeroclaw doctor update-context-windows`, but the diff only adds the runtime `cli-doctor-ctxwin-*` message keys. That means localized help for the new subcommand falls back to the compile-time English doc comment instead of the Fluent catalogue. This is the same user-facing CLI localization contract that drove the earlier blocker, just on the command-help surface rather than the runtime `println!` output.

Please add the matching command-about Fluent key for `update-context-windows` across the runtime CLI catalogues touched by this PR, and include any existing command-translation coverage or `cargo fluent` check needed to keep it from drifting again.

Separately from the code blocker, this review is for the current PR head only. GitHub currently reports the branch as `dirty` / not mergeable, so the branch still needs a conflict refresh plus checks on the refreshed head before landing. The PR body label snapshot also still says `risk: medium` and `size: M`, while the live labels are `risk: high` and `size: L` with additional scope labels such as `dependencies`, `core`, and `agent`; that is maintainer-editable metadata, but it should be brought back into sync before this lands so the public risk record matches the reviewed surface.

singlerider:
Moved from v0.8.3 to v0.8.4 during ship-date triage. Not-started UI feature (context-window bar) with no linked issue and outside the v0.8.3 headline scope. Parked for re-triage.

Audacity88 (approved):
Reviewed current head `8257c38ddde00d9710c5cbd3afae33e02d3ee994` against the live PR state, prior formal reviews, the public diff, and the touched config/provider/runtime/gateway/RPC/web/TUI surfaces. I did not run local Cargo validation, a gateway smoke, a doctor CLI smoke, web UI verification, or Zerocode TUI verification for this review pass; I relied on the green visible CI checks plus code review.

### ✅ Resolved — The remaining localization blocker is fixed at this head

The active blocker from my prior review was the missing Fluent command-help key for the new `zeroclaw doctor update-context-windows` subcommand. That key is now present in the embedded CLI catalogues (`cli-doctor-update-context-windows-about` in en/es/fr/ja/zh-CN), and the non-English locale formatter regression now includes that command-about key alongside the `cli-doctor-ctxwin-*` output keys. That closes the last blocker I had on the public CLI localization contract.

### ✅ Still resolved — The earlier context-window correctness fixes remain pinned

The exact-alias write-back path still carries `(provider_type, alias)` through `update_context_windows` and writes `providers.models.{provider_type}.{alias}.context_window`, with `update_context_windows_uses_exact_alias_not_model_uri` covering the two-alias regression. The gateway refresh endpoint still forwards configured provider credentials into the `/models` fetch path, with `refresh_context_window_forwards_api_key` covering both the `Authorization` header and response-body non-leak property. The runtime/RPC/gateway display path consistently reports context usage against `effective_model_context_window()` / `agent.resolved.model_context_window`, while the existing `effective_context_budget()` path continues to bound preemptive history trimming.

### 🟡 Warning — Non-blocking follow-ups

Two small cleanups are worth taking during the conflict refresh or in a follow-up PR, but I do not think either should hold this head:

- `doctor::update_context_windows` formats `cli-doctor-ctxwin-write-failed` in the `set_prop_persistent` error branch but does not print or return it, so an unexpected config write failure would lose the specific error and only print the generic not-found message.
- A few new comments describe runtime-profile `max_context_tokens` as an output budget, but the surrounding code and docs still use it as the configured context/history-trimming budget; provider response length remains controlled by provider `max_tokens`. The code path I reviewed is okay, but the wording should be aligned to avoid future confusion.

### 🟢 What looks good — Review decision

Approving to clear my `CHANGES_REQUESTED` on the reviewed code. This is not a merge-ready statement: GitHub currently reports the branch as `CONFLICTING` / `DIRTY`, and the branch still needs a conflict refresh plus checks on the refreshed head before landing. The PR also still has another `CHANGES_REQUESTED` review on record, so GitHub's review decision should remain blocked until that reviewer re-confirms or the review is otherwise dismissed. The PR body label snapshot is also stale (`risk: medium`, `size: M`, `channel:cli`) relative to the live labels (`risk:high`, `size:L`, plus `dependencies`/`core`/`agent`), so that public metadata should be synchronized before merge.

Audacity88:
Thanks for continuing to work through this. The code blockers I reviewed are addressed, but GitHub still reports this branch as conflicting with `master`.

Please merge the latest `master` into the branch, resolve the conflicts, and let CI rerun on the refreshed head. Before merge, the PR body/label snapshot also needs to match the live scope and labels (`risk: high`, `size: L`, and the current touched surfaces).

eugeneb50:
**1. `doctor/mod.rs:590-601`** — Replaced the generic `cli-doctor-ctxwin-not-found` print with the detailed `cli-doctor-ctxwin-write-failed` print that includes the actual error, instead of silently dropping it.

**2. `schema.rs:3284-3285`** — Fixed `ResolvedRuntime::max_context_tokens` doc from "output budget" to "token budget for preemptive context/history trimming" with a note that it's NOT the provider `max_tokens`.

**3. `schema.rs:3928-3930`** — Fixed `effective_max_context_tokens` doc comments to match the correct semantics.

**4. `schema.rs:4029`** — Fixed inline comment in `resolved_agent_config` to say "context/history trimming" instead of "output budget".

**5. `loop_.rs:1208,2733,2820`** — Renamed misleading variable `eff_max_context_tokens` to `eff_model_context_window` since it was assigned `model_context_window` (the model's input window), not the runtime profile's `max_context_tokens`.

**6. `i18n.rs:534-562`** — Added a new test `doctor_ctxwin_write_failed_formats_in_english_and_japanese` that verifies both English and Japanese Fluent sources properly format `cli-doctor-ctxwin-write-failed` with both `provider_ref` and `error` arguments.

Audacity88 (commented):
@eugeneb50, thanks for the refresh.

Context for this pass: I re-reviewed current head 86388057bb2468f52f87f10ca5b6d22a777bd643 against the live PR state, my prior approval notes, your latest summary, the green visible CI, and the current config/runtime/doctor/gateway/i18n source. I did not run local Cargo, a live gateway smoke, a doctor CLI smoke, web UI verification, or Zerocode TUI verification for this pass.

### ✅ Resolved — My two follow-up cleanups are addressed

The `set_prop_persistent` error branch in `doctor::update_context_windows` now prints `cli-doctor-ctxwin-write-failed` with both the provider ref and the concrete error, instead of formatting that message and then dropping it. The new `doctor_ctxwin_write_failed_formats_in_english_and_japanese` test also pins those message arguments.

The context-token wording I called out is mostly aligned now: `ResolvedRuntime::max_context_tokens`, `effective_max_context_tokens`, `resolved_agent_config`, and the local loop variable all distinguish the runtime trimming budget from `model_context_window`. One doc-comment nit remains in `effective_model_context_window()` where it still says the runtime profile is for “output budget”; that is worth cleaning up later, but I do not see it as a code blocker.

### 🟢 What looks good — no new block from my side

My approval on the reviewed code still stands. I also re-checked the earlier pins that matter for this head: exact-alias doctor write-back, gateway API-key forwarding without response leakage, command-help/runtime Fluent keys, and non-English formatter coverage are still present, and visible CI is green.

This is still not a merge-ready call. @WareWolf-MoonWall's older `CHANGES_REQUESTED` review remains the active GitHub review block, and the PR still has requested reviewers. The remaining public-record cleanup I see is the PR-body security/privacy wording: the refresh paths do read and forward existing provider credentials to the configured `/models` endpoint, so that template answer should say credentials handling changed for this feature rather than `No`.

WareWolf-MoonWall (approved):
Reviewed at 86388057bb2468f52f87f10ca5b6d22a777bd643. First look.

Re-review pass. My previous `CHANGES_REQUESTED` was at df543ae6. This pass re-checks all items I raised, incorporates @Audacity88's subsequent review cycles, and assesses the current head. CI is green. @Audacity88 formally approved at 8257c38d and their latest comment confirms "approval still stands." @singlerider has approved.

### ✅ Resolved — CI

The CI failures I blocked on (6/21 checks red at df543ae6) are fully resolved. All checks pass on the current head.

### ✅ Resolved — Fluent layer coverage for new user-facing strings

The `eprintln!` context bar in `agent/loop_.rs` and the `println!` doctor strings I called out are now behind `fl!()` keys. Fluent variable syntax (`{$used}`, `{$max}`, `{$ctx}`, etc.) is correctly using the `$`-prefixed external-variable form across all five locale bundles (en/es/fr/ja/zh-CN). @Audacity88's subsequent review confirmed the locale formatter regression (`doctor_ctxwin_write_failed_formats_in_english_and_japanese`) is in place and the translated entries use correct `{$...}` placeholders. The earlier localization blocker is closed.

### ✅ Resolved — Unauthenticated `/models` fetches

The gateway `handle_refresh_context_window` now reads the provider's `api_key` from config via the JSON serialization path (bypassing the `#[secret]` masking in `get_prop`) and passes it into the temporary `ModelProviderConfig` used for the fetch. The `refresh_context_window_forwards_api_key` wiremock test pins both the `Authorization: Bearer ...` header on the outbound request and the non-leak property on the response body. The doctor/quickstart path was already fixed in an earlier iteration. This closes my warning about providers silently failing for authenticated OpenAI-compatible catalogs.

### ✅ Resolved — Inconsistent `model_context_window` resolution strategy

The agent loop now consistently uses `agent.resolved.model_context_window` (the snapshot resolved at agent startup from `effective_model_context_window()`) rather than re-resolving on every turn. @Audacity88 confirmed this in their pass at 8257c38d. The `max_context_tokens` / `model_context_window` naming distinction with doc comments is also in place.

### ✅ Resolved — Wrong-provider-alias write-back

@Audacity88's blocking item from their June 20 pass: `update_context_windows` now carries the `(provider_type, alias)` identity through the target list and mutates the exact entry, rather than scanning for the first model/URI match. The `update_context_windows_uses_exact_alias_not_model_uri` regression test is present.

### 🟡 Warning — PR title still has trailing whitespace

The title ends with two trailing spaces: `"…interactive mode.  "`. The CI title validator doesn't anchor at `$` so it passes, but this is visible in the GitHub UI and PR history. Please strip before merge.

### 🟡 Warning — PR body metadata snapshot is stale

The PR body label snapshot still references `risk: medium`, `size: M`, `channel:cli` from an earlier iteration. The live labels are `risk:high`, `size:L`, `dependencies`, `core`, `agent`. @Audacity88 also noted this. The body should be updated to match live labels before merge.

### 🟡 Warning — Branch is conflicting with master

@Audacity88 noted in their approval comment that the branch is currently `CONFLICTING` / `DIRTY`. A conflict refresh and a clean CI run on the rebased head are required before this can land, but that is an author action, not a code review blocker.

### 🟢 What looks good — SSOT is correctly implemented throughout

`context_window` lives exactly once in `ModelProviderConfig` in `zeroclaw-config`. `effective_model_context_window()` resolves from that canonical field at call-time with a 32k fallback. `ResolvedRuntime.model_context_window` is populated from the same resolver at agent startup — a materialized view built on demand from the canonical source, not a duplicate. The `max_context_tokens` (runtime trimming budget) and `model_context_window` (model capacity) are now clearly separated with doc comments. The gateway WS `done` frame carrying `max_context_tokens` resolves from `effective_model_context_window()` without caching a separate field.

### 🟢 What looks good — New gateway endpoint test

The `refresh_context_window_forwards_api_key` wiremock test exercises a real HTTP round-trip with auth, which is the right level of coverage for a gateway endpoint that handles credentials.

## Validation Evidence

I did not run local Cargo validation. CI is green on this head per the GitHub checks display. @Audacity88's latest comment confirms code-level review at the same head.

## Security & Privacy Impact

The `fetch_context_window` path reads and forwards configured provider credentials to the provider's `/models` endpoint only; responses strip the secret from returned bodies (verified by the wiremock test). This is the correct handling. No new permissions, filesystem scope, or PII exposure.

## Compatibility

`context_window` is optional; all surfaces fall back to 32,000 if unset. Existing configs need no changes. The new `zeroclaw doctor update-context-windows` subcommand and `POST /api/config/model-providers/{type}/{alias}/refresh-context-window` endpoint are additive. `SHOW_CTX_BAR = true` in zerocode is a display-only change.

## Rollback

`git revert <merge-sha>` — all changes are additive / config-driven with no migrations. `SHOW_CTX_BAR` can be flipped back to `false` if needed without a config change. Observable failure sentinel: `ctx: 0 / 32000` in the TUI means `context_window` is unset and the fallback is active.

JordanTheJet (commented):
Reviewed current head 86388057bb2468f52f87f10ca5b6d22a777bd643 — the full diff (24 files, ~1024 LOC), all eight formal reviews on record from @WareWolf-MoonWall, @Audacity88, and @singlerider, the complete reply history, and the touched config/provider/runtime/doctor/gateway/RPC/web/TUI surfaces. I cross-checked the diff against local master by reading the source (SSOT resolver, provider-map helpers, the history-trim call sites) rather than rebuilding. I did not run local Cargo validation; CI is fully green at this head (all 24 required checks pass), and GitHub reports the branch MERGEABLE / CLEAN with reviewDecision APPROVED.

First, taking stock, because this PR has had a long and genuinely productive review cycle: every prior blocker is closed and pinned, so I'm not re-raising any of them.

### ✅ Resolved — the whole prior blocker set is closed and test-pinned

Confirmed against this head:

- **SSOT for the context window.** `context_window` lives exactly once, in `ModelProviderConfig` (schema.rs). `effective_model_context_window()` resolves it at call-time through the canonical `resolved_model_provider_for_agent()` resolver with a hard 32k fallback, and `ResolvedRuntime.model_context_window` is a materialized view populated from that same resolver at agent startup — not a cache that can drift. This is exactly the pattern AGENTS.md's SSOT rule permits, and it's the highest-risk thing on this PR; it's done right.
- **Fluent coverage** for the new `eprintln!` context bar and the `println!` doctor output, plus the `cli-doctor-update-context-windows-about` command-help key, all across en/es/fr/ja/zh-CN, with `{$var}` external-variable syntax consistent between English and the translations.
- **Credential forwarding** on both the doctor/quickstart path and the gateway `handle_refresh_context_window` endpoint, pinned by `refresh_context_window_forwards_api_key` (asserts the `Authorization: Bearer` header AND that the key never appears in the response body).
- **Exact-alias doctor write-back** (carries `(provider_type, alias)` and writes `providers.models.{type}.{alias}.context_window`), pinned by `update_context_windows_uses_exact_alias_not_model_uri`.
- **Non-English formatter regression** (`non_english_locales_format_context_window_args`) and the write-failed branch now printing its concrete error (`doctor_ctxwin_write_failed_formats_in_english_and_japanese`).
- **Branch conflict / CI.** The earlier `CONFLICTING`/`DIRTY` state and the 6/21 red checks @WareWolf-MoonWall first blocked on are gone — the refresh landed clean and all required gates are green.

@WareWolf-MoonWall, @singlerider, and @Audacity88 each cleared their `CHANGES_REQUESTED` with a later approval, and there is no active block on record. This is in good shape.

I have two things I'd like the author to look at before this lands — neither is a block, but one is a real behavior change I want to make sure is intentional rather than incidental.

### 🟡 Warning — the overflow-recovery trim now targets the full model window, not the runtime trimming budget

`crates/zeroclaw-runtime/src/agent/loop_.rs` (context-overflow recovery branch, ~L2718–2733)

On master, the context-overflow *recovery* path (the branch that fires when the provider returns `is_context_window_exceeded`, drops oldest whole turns, and retries) trims to `eff_max_context_tokens` — the runtime profile's trimming budget (`agent.resolved.max_context_tokens`). This PR changes that argument to `eff_model_context_window`.

That's a genuine semantic change, and it's a different one from what the summaries in the prior reviews captured. The preemptive-trim paths (the two `context_token_budget: …effective_context_budget()` call sites) do still key off `effective_context_budget()` and are untouched — so "preemptive trimming is unchanged" is correct — but this recovery call is the one that moved, and it was never on `effective_context_budget()` to begin with.

The intent is defensible: after an overflow you want to drop turns until history fits the *model's actual capacity*, so keying off the model window reads as more correct than keying off a possibly-lower trimming budget. My concern is the reverse edge: `trim_to_recent_turns` measures against `estimate_history_tokens` (a ~4-chars-per-token heuristic), and the request that just failed already exceeded the provider's real tokenizer. Trimming to *exactly* the window — rather than to something with headroom below it for the reply and the next user turn — risks a recovery that trims but still overflows on the retry, especially where the estimate under-counts. Could you confirm this switch is deliberate, and consider whether the recovery target should sit a margin below the full window? If it's intentional, a one-line comment at that call site explaining why recovery uses the window (and not the trimming budget the other paths use) would save the next reader the same double-take I had.

### 🔵 Suggestion — dead `context_source` union in the web WS type

`web/src/types/api.ts`

`WsMessage` declares `context_source?: "Config" | "ProviderApi" | "RuntimeProfile" | "Default"`, but the gateway `done` frame in `crates/zeroclaw-gateway/src/ws.rs` never emits `context_source` (it sends `max_context_tokens`, `input_tokens`, `output_tokens`, and the rest — zero occurrences of `context_source` anywhere in the tree). The web `ContextBar` path reads only `max_context_tokens` and `input_tokens`, so this field has no producer and nothing consumes it. Either wire the gateway to send it (if the UI is meant to surface where the number came from) or drop it — a speculative type field with no source is the kind of thing that quietly rots. Per AGENTS.md anti-patterns, this is the "no speculative config/feature flags just in case" spirit applied to the wire type.

### 🟢 What looks good — the credential-forwarding test is at the right level

`refresh_context_window_forwards_api_key` stands up a real wiremock that *requires* the bearer header (`.and(header(...)).expect(1)`), drives the actual `handle_refresh_context_window` handler, and asserts both the outbound auth AND the no-leak property on the response body. For a gateway endpoint that reads and forwards a provider secret, an HTTP round-trip test that pins the leak boundary is exactly the coverage this deserves — this is the pattern to repeat for future credential-touching endpoints.

### 🟢 What looks good — capacity vs. output-budget separation is finally named

Keeping `max_context_tokens` (runtime trimming budget) and `model_context_window` (model input capacity) as distinct fields with doc comments removes a long-standing naming ambiguity in this area. That clarification is worth more than the ctx bar itself; it's the kind of thing that prevents future drift bugs.

## Verdict

Commenting, not approving — but this is close, and my hold is light. Three reviewers have already cleared and GitHub's decision is APPROVED, so I'm not blocking. I'd just like the overflow-recovery trim change confirmed as intentional before I put my own approval on it, and the dead `context_source` type cleaned up or wired.

Separately for the milestone owner: this is a ship-ready operator surface (green CI, MERGEABLE, reviewDecision APPROVED) currently sitting in v0.8.4 (#18) overflow parking — worth pulling into v0.8.3 (#13) rather than parking it. That's maintainer metadata, not a contributor blocker.

I did not run local Cargo validation. CI is green at 86388057b.

Audacity88 (commented):
@eugeneb50, one last maintainer check before we land this.

Context for this follow-up: I checked the current public PR state, Jordan's latest review, and the public diff for the overflow-recovery branch. I did not run local validation for this pass.

Jordan's note about the context-overflow recovery path is the only remaining thing I want confirmed. The code now trims overflow recovery against the full `model_context_window` instead of the runtime profile's `max_context_tokens` trimming budget. That may be the right behavior, but it is a real semantic change, and Jordan also called out the retry-risk question around the heuristic history-token estimate.

Can you confirm that switch is intentional? If it is, a short code comment at that call site explaining why overflow recovery uses the model window rather than the normal trimming budget would be enough for me. If the exact-window target risks retry-loop issues with the heuristic estimate, please consider whether recovery should target a margin below the full window instead.

The dead `context_source` web type field can be cleaned up here or in a follow-up; I do not see that part as a merge blocker.

Audacity88 (commented):
@eugeneb50, thanks for the branch refresh.

Context for this follow-up: I rechecked current head 2d5433af060c586359f1ea8e63df868ab009b5ed against Jordan's overflow-recovery note, my last maintainer comment, the live PR state, and the narrow diff since 86388057bb2468f52f87f10ca5b6d22a777bd643. I did not run local validation for this pass.

### ✅ Resolved — Refresh cleanup is addressed

The cleanup portion is addressed now: the dead `context_source` field is gone from `web/src/types/api.ts`, the branch is `MERGEABLE` / `CLEAN`, the visible checks are green, and the PR-body label snapshot matches the live labels.

### 🟡 Warning — Overflow recovery still needs confirmation or a call-site comment

The one item I asked for is still open. The context-overflow recovery path in `crates/zeroclaw-runtime/src/agent/loop_.rs` still trims with `eff_model_context_window`, and I do not see the short call-site comment explaining why recovery intentionally targets the model window rather than the normal runtime trimming budget. I also do not see a margin adjustment if the exact-window retry risk is real.

Please either add that comment at the recovery call site, or change the recovery target if the heuristic estimate means retrying at the full model window can loop. Once that is addressed, I can recheck this final item.

eugeneb50:
Added a trim to 90% of eff_model_context_window 

- Why recovery targets the model window rather than the operator's trimming budget: after a real tokenizer overflow we must fit the provider's capacity, not the preemptive-trimming budget.
- Why the 10% headroom: `estimate_history_tokens` uses a ~4-char-per-token heuristic (`history.rs:285`). Trimming to the exact window means the retried request can re-overflow due to heuristic under-count plus the assistant reply content. The 90% budget absorbs both.
- How this relates to the other two trim sites: the in-loop recovery at `turn/context_recovery.rs:85` uses a `tokens_now * 2/3` heuristic (different goal — aggressive drop before retry), and the preemptive trim at `turn/mod.rs:393` uses `effective_context_budget()` (different goal — operator-chosen tightening before the call).

**2. Headroom margin** (`loop_.rs:2573-2577`)

```rust
let recovery_budget = eff_model_context_window * 9 / 10;
```

A proportional 10% margin: 3200 tokens on the 32k fallback, 20k tokens on a 200k model. Matches the pattern of the in-loop recovery's hardcoded `* 2 / 3` — no new config knob, no SSoT surface to inherit.

**3. Regression test** (`cli_outer_recovery_trims_below_model_window_with_headroom`)

Pins that the recovery budget is strictly below the model window. Builds a history whose `estimate_history_tokens` exceeds 32k, calls `trim_to_recent_turns` with the 90% budget, and asserts:
- `result.trimmed == true` (full-window would no-op and fail)
- `result.tokens_after <= recovery_budget`
- `result.tokens_after < model_context_window` (strict headroom)


🟡 Warning Fix: CLI and gateway context bars are reporting inflated ctx values based on spurious accumulated token counts.    Zerocode's chat path is returning the actual providers token count and is working as expected .

**Problem**

The CLI context bar (`ctx: X / Y`) displayed inflated cumulative token counts that quickly exceeded the model window. A session with 6 turns showed `345,234 / 262,144 [100%]` — 131% of capacity — even though every individual LLM call stayed within the window. The bar was misleading: it showed "this turn" but read session-wide totals.

**Root cause**

`record_tool_loop_cost_usage` in `cost.rs:339-355` accumulates `input_tokens` via `saturating_add` — every LLM call adds its provider-reported prompt tokens to a shared `Arc<Mutex<TurnUsage>>`. The LLM's `usage.input_tokens` is an **absolute** measure of the full prompt sent (accounting for all history + tool results). Accumulating these across calls double-counts the same base context repeatedly.

The CLI interactive path (`loop_.rs:1698-1699`) creates a single `cost_tracking_context` per session, never scopes a per-turn `TOOL_LOOP_TURN_USAGE`, and reads `snapshot_turn_usage().input_tokens` for the context bar — which falls back to the session-wide accumulator, not per-turn data.

The WS gateway path already scopes a fresh `TOOL_LOOP_TURN_USAGE` per turn (`ws.rs:1073-1086`), so it avoids the session-wide problem. But within a multi-iteration turn, even the gateway's accumulation double-counts across tool-call rounds.

**How zerocode gets it right**

Zerocode receives `TurnEvent::Usage { input_tokens }` — the raw provider-reported prompt token count — and **replaces** its stored value on every arrival (`apps/zerocode/src/chat.rs:5263-5272`). This is documented in a comment: *"ContextUsage reports the current prompt size for the upcoming/just-sent turn. It is an absolute measurement of how full the model's context window is, not an increment. Accumulating across turns produced a runaway counter that quickly exceeded the window."* Zerocode already fixed this same bug before.

**The fix**

A 3-part change in `cost.rs`:

1. **Add `last_input_tokens: u64` to `TurnUsage`** (`cost.rs:21-25`) — stores the most recent provider-reported prompt token count. Replaced on every LLM response, never accumulated.

2. **Set `last_input_tokens` in `record_tool_loop_cost_usage`** (`cost.rs:339-355`) — in both the per-turn `TOOL_LOOP_TURN_USAGE` branch and the session fallback branch, set `turn_usage.last_input_tokens = input_tokens`. The existing `input_tokens` field continues to accumulate (needed for `AgentEnd` cost reporting).

3. **Read `last_input_tokens` in the context bar** (`loop_.rs:2661`) — swap from `usage.input_tokens` to `usage.last_input_tokens`. The bar now shows the actual prompt size of the last LLM call, exactly matching what the provider counted.

For the WS gateway's `done` frame (`ws.rs:1466-1481`), optionally apply the same swap for consistent accuracy in gateway responses.

**What does NOT change**

- `input_tokens` accumulation in `record_tool_loop_cost_usage` — preserved for `AgentEnd` observer events and cost tracking, which correctly need session-wide totals
- `output_tokens` — works the same, no change needed (the output of the last LLM call is already the right number)
- All existing tests pass without modification (the new field is additive)

We can release this PR and I can open a following or include the new shape here.

eugeneb50:
- `record_tool_loop_cost_usage` in `cost.rs` accumulated `input_tokens` via `saturating_add` for cost/billing (correct)
- CLI context bar, WS gateway done frame, and frontend all read the same accumulated total — producing a runaway counter that could exceed the model's context window

### Fix

1. **`cost.rs`** — Added `last_input_tokens: u64` to `TurnUsage`. Set on each `record_tool_loop_cost_usage` call by replacing (not accumulating) with the provider's absolute reported prompt size. The accumulated `input_tokens` remains for billing.

2. **`loop_.rs`** — CLI context bar reads `usage.last_input_tokens` instead of `usage.input_tokens`.

3. **`ws.rs`** — WS gateway tracks `last_input_tokens` (replace on each `TurnEvent::Usage`) and includes it in the `done` frame sent to the frontend.

4. **`AgentContext.tsx`** — Frontend context bar prefers `last_input_tokens` with fallback to `input_tokens` for backward compatibility with older servers.

5. **`api.ts`** — Added `last_input_tokens?: number` to the WS message type.

### Design

- **Replace, don't accumulate** — matches zerocode's proven pattern at `apps/zerocode/src/chat.rs:5263-5272`
- **Accumulated `input_tokens` unchanged** for `AgentEnd` observer events and cost reporting
- **`last_input_tokens` additive** — new field alongside existing API responses; no breaking changes

### Files changed

| File | Change |
|---|---|
| `crates/zeroclaw-runtime/src/agent/cost.rs` | +14 — `last_input_tokens` field + set on each call |
| `crates/zeroclaw-runtime/src/agent/loop_.rs` | +15 — CLI context bar reads `last_input_tokens`|
| `crates/zeroclaw-gateway/src/ws.rs` | +8 — track `last_input_tokens` in done frame |
| `web/src/contexts/AgentContext.tsx` | +6 — frontend reads `last_input_tokens` |
| `web/src/types/api.ts` | +1 — type definition |

Audacity88 (approved):
Thanks for the final refresh. I rechecked current head `d4d804f2e36c93a374e4e863c76a45ba4f62e54f` against the live PR state, the prior review thread, the current diff against `master`, and the runtime/gateway/web paths touched by the latest `last_input_tokens` and overflow-recovery updates. I did not run local Cargo, a live gateway smoke, a doctor CLI smoke, web UI verification, or Zerocode TUI verification for this pass; CI is green on this head, and the PR body now records the manual user-boundary checks.

### ✅ Resolved — Prior context-window blockers remain closed

The prior blockers still look closed. The provider `context_window` remains the single source of truth, the doctor exact-alias write-back is test-pinned, the gateway refresh endpoint forwards the configured provider credential without echoing it, the Fluent entries and command-help key are present across the touched locales, and the branch/body/label cleanup is now aligned.

### 🟢 What looks good — Recovery headroom and context-bar accounting

The two follow-up changes also look right to me. The overflow recovery path now trims to 90% of the model window, which addresses the retry-at-the-exact-boundary concern while keeping that recovery target distinct from the normal preemptive trimming budget. The `last_input_tokens` addition also preserves the accumulated `input_tokens` total for cost/reporting while giving the CLI and gateway/web context bars the latest provider-reported prompt size, which is the value those bars should render.

### 🔵 Suggestion — Remove the stale note reference

One non-blocking cleanup suggestion: the long recovery comment should drop the `See the note2.md` reference. The rest of the comment already explains the invariant in-code; pointing at a local review artifact will age poorly and is not useful to future readers.

Approving on the code. The remaining comment-reference cleanup is small enough that I do not think it should block merge.

Issue #7852, opened 2026-06-17
[Tracker]: v0.8.2 skills platform


Audacity88:
## Summary

This tracker is the public coordination surface for the v0.8.2 skills platform: registries, effective-skill resolution, plugin-bundled skill behavior, missing-capability suggestions, skill-audit visibility, and operator-visible skill facts.

Last live refresh: 2026-06-21T00:00:00Z. Current skills-platform map: 19 open v0.8.2 items. Related A2A extension surface: 4 open items.

Partial merge-follow-up update 2026-06-28: #8190 merged and closed #6714.

Related v0.8.2 trackers:

- #7314: WASM/plugin program tracker.
- #8181: release-support and non-plugin tracker.
- #3566 and #7218: A2A/interop trackers.

## Current skills-platform map

- [ ] #5262 — Issue, in-progress, accepted, no-stale, p2, risk low — [Feature]: Add ZeroClaw logo to official Agent Skills client list
- [ ] #6165 — Issue, RFC, blocked, accepted, p2, risk high — RFC: Prefer a lighter ZeroClaw core through external integrations
- [ ] #6289 — Issue, accepted, no-stale, p2, risk high — [Feature]: Prompt-triggered install suggestions for missing skills and plugins
- [x] #6714 — Issue, closed by #8190, accepted, p2, risk high — [Feature]: Remove remote-markdown-link block from skill audit
- [x] #7025 — Issue, in-progress, accepted, p2, risk high — [Bug]: read_skill cannot load plugin-bundled skills the prompt advertises
- [x] #7245 — PR, needs-author-action, risk high — fix(read_skill): fix read_skill cannot load plugin-bundled skills and bounded skills
- [x] #7740 — Issue, in-progress, accepted, p2, risk high — bug(runtime): base missing-skill suggestions on effective tool set
- [ ] #7742 — Issue, accepted, p2, risk high — bug(runtime): refresh system prompt after tool dispatcher swap
- [ ] #7749 — Issue, accepted, p2, risk high — [Feature]: Per-agent prompt_injection_mode override
- [ ] #7816 — Issue, in-progress, accepted, p2, risk high — [Feature]: Pluggable skill registries — keep GitHub repo as default tier, add flagged external + user-configured registries
- [x] #7819 — PR, risk high — fix(runtime): base missing-skill suggestions on effective tool set in process_message path
[...]

PR #6700, opened 2026-05-16
feat(web): add read-only skills browser

ilteoood:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Adds a read-only Skills page to the gateway web dashboard.
  - Adds a sidebar route for `/skills` so operators can browse installed skill bundles from the web UI.
  - Uses the existing bundled skills API to list bundles, list skills in each bundle, and expand a skill to view its `SKILL.md` body.
  - Adds navigation and Skills-page localization strings.
- **Scope boundary:** This PR does not add or change gateway/runtime skills API routes, skill enable/disable behavior, create/edit/delete flows, CLI behavior, or skill loading semantics.
- **Blast radius:** Gateway web dashboard navigation and the new read-only Skills page. The page depends on the existing `/api/skills/bundles`, `/api/skills/bundles/{alias}/skills`, and `/api/skills/bundles/{alias}/skills/{name}` endpoints.
- **Linked issue(s):** Related #6695. This PR is a first read-only UI slice and does not fully close the broader management request.
- **Labels:** `enhancement`, `size: M`, `risk: medium`

## Validation Evidence (required)

Author-reported local validation from the original PR body:

```text
cargo fmt --all -- --check -> exit code 0
cargo clippy --all-targets -- -D warnings -> Finished `dev` profile [unoptimized + debuginfo] target(s) in 29.66s
cargo test -> test result: ok, for all the sections
npm run build -> built in 2.17s
```

GitHub CI currently shows lint, test, security, checks, benchmarks compile, macOS build, and Windows build passing. The x86_64 Linux build was cancelled after about 40 minutes, which caused CI Required Gate to fail; no compiler or test failure is visible in the current check summary.

- **Beyond CI — what did you manually verify?** The PR includes a screenshot of the Skills page rendering in the dashboard. Manual live verification of bundle listing, search, and `SKILL.md` expansion on a local gateway instance has not been recorded in the PR body.
- **If any command was intentionally skipped, why:** Literal command tails are not currently included beyond the summarized author-reported results above.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? No. This PR adds a read-only web page over existing authenticated skills bundle endpoints.
- New external network calls? No.
- Secrets / tokens / credentials handling changed? No.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? No.
- If any `Yes`, describe the risk and mitigation: N/A.

## Compatibility (required)

- Backward compatible? Yes.
[...]

PR #7963, opened 2026-06-19
fix(gateway): dashboard Skills page reflects an agent's effective skills

### LINKED ISSUE #7757: [Bug]: Gateway web dashboard Skills page only reflects skill_bundles, misses workspace/open-skills/plugin skills
### Affected component

gateway/api

### Severity

S2 - degraded behavior

### Current behavior

The web dashboard's Skills page (`web/src/pages/Skills.tsx`) and its backing REST API (`GET /api/skills/bundles`, `GET /api/skills/bundles/:alias/skills` — `crates/zeroclaw-gateway/src/api_skills.rs`) are backed by `SkillsService::list_bundles()` / `list_skills()` (`crates/zeroclaw-runtime/src/skills/service.rs:81-83`), which **only iterates `config.skill_bundles`** (the `[skill_bundles.<alias>]` config table). It does not look at any of the other sources the runtime actually loads skills from.

Meanwhile, the production agent runtime (`load_skills_for_agent_from_config`, called from `agent.rs:1385` and `loop_.rs:1324,2783`) unions skills from **four** sources per agent:
1. The agent's own workspace: `<install>/agents/<alias>/workspace/skills/` — always scanned, unconditionally.
2. Open-skills repo, if `skills.open_skills_enabled = true`.
3. Plugin-bundled skills, if `plugins.enabled = true` (feature `plugins-wasm`).
4. Skill bundles referenced by `[agents.<alias>].skill_bundles = [...]`.

The dashboard only ever surfaces source (4), and only the bundles that exist in `config.skill_bundles` — it does not show source (1)/(2)/(3), and it does not filter by which bundles are actually attached to which agent (it shows every configured bundle, agent-unaware).

Net effect: an operator can have skills that are fully functional at runtime (loaded into the system prompt, callable via `read_skill`) and see **zero skills** on the dashboard Skills page, with no error and no indication that anything is missing. The page just renders empty.

### Expected behavior

The dashboard should reflect the same skill set the runtime actually uses for a given agent — i.e. `SkillsService` (or the `/api/skills/*` handlers) should be reconciled with `load_skills_for_agent_from_config` so that workspace skills, open-skills, and plugin skills are also visible, not just `config.skill_bundles` entries. At minimum, the page should not silently render an empty list when skills exist and are loaded by the runtime — that's a confusing, undebuggable dead end for any operator who hasn't read the runtime source.

This is the same root cause as #7236, #7025, and #7541: V3 (#6398, RFC #5890) introduced multiple independent code paths that each resolve "what skills exist" differently, with no single canonical resolver. Those three issues were about runtime-vs-runtime path mismatches (wrong `workspace_dir`, prompt-vs-`read_skill` disagreement); this one is the dashboard/API-vs-runtime instance of the same pattern — previously unreported.

### Multi-agent implications

The V3 multi-agent runtime (#6398, RFC #5890) makes this worse than a simple "missing data sources" bug — it breaks the page's basic assumption that there is one global skill list to show.

1. **No `:alias` dimension on the route at all.** Every other per-agent dashboard surface is alias-scoped: chat is `/agent/:alias` (`web/src/router/router.tsx:40`), the workspace file browser is `/agent/:alias/workspace` (`router.tsx:41`), config sections are `/config/:section/:type/:alias`. The Skills page is `/skills` (`router.tsx:44`) — flat, global, no alias. It was built bundle-first before `agents.<alias>.skill_bundles` and per-agent workspace skills existed, and never revisited when those landed.

2. **"The skill list" isn't singular under multi-agent.** Each agent's effective set is `<install>/agents/<alias>/workspace/skills/` ∪ (whichever bundles *that agent's* `skill_bundles` list references, each filtered by that bundle's `include`/`exclude`) ∪ open-skills ∪ plugin skills. Two agents can reference the same bundle with different `include`/`exclude` and end up with different effective subsets. Just adding the three missing sources to `SkillsService` isn't sufficient — the page needs an agent selector (or to live under `/agent/:alias/skills`) and a "resolve effective skills for agent X" call, not a flat bundle listing.

3. **Bundle visibility is reference-blind.** `SkillsService::list_bundles()` walks `config.skill_bundles` (the global definition table) with no join against any agent's `skill_bundles` list. A bundle defined but unreferenced by every agent still shows up as if it were live; a bundle shared by several agents shows its full directory contents, not what any specific agent actually resolves after filtering. There is currently no way to answer "which skills does agent X actually have" from the dashboard.

**Suggested direction:** add a "resolve effective skills for agent `<alias>`" entry point on top of the existing `load_skills_for_agent_from_config` runtime resolver (single source of truth, reused by prompt build, `read_skill`, and the dashboard alike — closing the same gap #7025 already flagged for `read_skill` vs. prompt), and scope the Skills page/route by agent alias instead of showing a global, agent-unaware bundle list.

### Steps to reproduce

```bash
# 1. Fresh V3 install, no [skill_bundles.*] configured.
# 2. Place a skill directly in the agent's workspace skills dir:
mkdir -p ~/.zeroclaw/agents/<alias>/workspace/skills/my-skill
cat > ~/.zeroclaw/agents/<alias>/workspace/skills/my-skill/SKILL.md <<'SKILL'
---
name: my-skill
description: test
---
# My Skill
SKILL

# 3. Start the daemon/gateway and chat with the agent — confirm the skill
#    is listed in the system prompt and usable (it is).
# 4. Open the web dashboard -> Skills page.
# 5. Observe: page shows 0 skills, no bundles, no error.
#    Compare with: ./zeroclaw skills list   -> correctly lists "my-skill"
```

### Impact

Affected users: any operator who places skills in the per-agent workspace directory (the default/legacy location, also what `zeroclaw skills add` without a bundle alias may target) rather than registering a `[skill_bundles.*]` entry under `<install>/shared/skills/`.
Frequency: always, deterministic — not a race or edge case.
Consequence: the dashboard is unusable as a source of truth for "what skills does this agent have." Operators have to fall back to the CLI (`zeroclaw skills list` / `zeroclaw skills bundle show`) to get an accurate picture, defeating the purpose of the dashboard page.

### ZeroClaw version

v0.8.0 / 140d9d1e40

### Rust version

rustc 1.96.0

### Operating system

Linux

### Regression?

Unknown

### Pre-flight checks

Nillth:
## Summary

Fixes #7757. The dashboard Skills page + its REST API were backed by `SkillsService::list_bundles()/list_skills()`, which iterate **only** `config.skill_bundles`. But the runtime loads an agent's skills from **four** sources (`load_skills_for_agent_from_config`): the agent workspace, open-skills, plugins, and bundles. So an operator with workspace/open-skills/plugin skills saw **zero skills, no error** — a confusing dead end (same multi-resolver-drift family as #7025/#7236/#7541).

This makes the dashboard read the agent's **effective** skill set through the canonical resolver, agent-scoped:

- **New `GET /api/agents/{alias}/skills`** → the agent's effective skills, each tagged with provenance. It is sourced from `load_skills_for_agent_from_config` (the **audited** loader), **not** a widened `list_skills` (which does a raw, unaudited `read_dir`) — so the page reflects exactly what the agent injects and never surfaces un-audited external (open-skills/plugin) frontmatter.
- **`SkillsService::resolve_effective_skills(alias)`** + `SkillOrigin` (`Workspace` / `OpenSkills` / `Plugin` / `Bundle`) + `EffectiveSkill`. Origin is derived by mirroring the resolver's own discriminators (the `open-skills` tag, the `plugin:` name/tag prefix, a `location` match against a configured bundle dir, else workspace), so dashboard provenance can't drift. `editable` is server-derived and `true` **only** for `Bundle` skills — the only writable source.
- **Wire types** `AgentSkillEntry` / `AgentSkillsResult` (flat `origin` string + optional `plugin`/`bundle`/`directory` + `editable`). The bundle-editor wire type `SkillListEntry` and every `/api/skills/bundles/*` endpoint are **untouched**.
- **Frontend**: the Skills page swaps its per-bundle fan-out for a single `listAgentSkills(selectedAlias)` call behind an **agent selector** (reusing the existing `getAgentOptions`); each card shows an **origin badge**; any edit affordance is gated on `editable` (read-only sources show none). The route stays flat `/skills`; the read-only browser (#6700) is preserved; the bundle editor is unchanged.

**Scope:** the read/visibility path. Deliberately **deferred** to follow-ups (filed): a server-side write-guard rejecting non-bundle writes (the frontend already gates on `editable`, and bundle writes already resolve through `bundle::resolve_directory`); a "skipped due to audit" channel so "zero" can distinguish none-configured vs all-failed-audit; and a `shadowed_by` indicator for cross-source name collisions.

Related: #6700 (read-only browser, preserved), #7025/#7236/#7541 (sibling resolver-drift bugs), #7852 (skills-platform tracker). No conflict with the open #7245 (disjoint files; only consumes the resolver).

## Validation Evidence

```
cargo fmt --all -- --check                                               # clean (exit 0)
cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway --all-targets -D warnings   # no warnings (exit 0)
cargo test -p zeroclaw-runtime skills::service                           # 10 passed (+2 new)
cargo build -p zeroclaw-runtime --features plugins-wasm                  # plugin arm: left to CI*
(cd web && npm run build)                                                # PASS (tsc -b && vite build)
```
*The plugin-origin code is feature-agnostic string logic (`name`/tag `plugin:` prefix); the local `--features plugins-wasm` build couldn't run (build disk full on the wasm deps), so it's deferred to CI.

New tests (`skills::service::tests`):
- `derive_origin_classifies_each_source` — open-skills tag → `OpenSkills`, `plugin:foo/bar` → `Plugin("foo")`, `location` under a bundle dir → `Bundle`, else `Workspace`.
- `resolve_effective_skills_unions_workspace_and_bundle` — a workspace skill (the source the dashboard used to miss) **and** a bundle skill both appear, with correct `origin`/`editable`/`bundle`.

Beyond CI: the full `cargo build/clippy --workspace --all-targets` (the heavier gate for the shared `rpc::types` change) is left to CI — local disk was constrained; the touched crates pass `--all-targets` clippy.

## Security & Privacy Impact

- No new permissions, capabilities, or filesystem scope. The new endpoint is `require_auth`-gated like the rest of `/api/skills/*`, and **read-only**.
- **Audit parity preserved:** source enumeration goes through the audited `load_skills_for_agent_from_config`, never a widened un-audited `list_skills` — so we don't begin surfacing frontmatter from external (open-skills/plugin) sources the runtime would have rejected.
- No new external network calls; no secrets handling change; no PII in the diff.

## Compatibility

Backward compatible — additive. New endpoint + new wire types; the existing bundle endpoints, their wire types, and the bundle editor are unchanged. The Skills page now requires an agent context (the selector defaults to the first/runtime agent, so a single-agent operator sees their skills with zero clicks). No config/env/CLI surface change.

## Rollback

`git revert` — additive endpoint + types + a frontend data-source swap; no state or schema migration.


singlerider (approved):
Reviewed at 0c8832cc on a fresh checkout of `fix/dashboard-skills-all-sources`. Built the touched crates, ran the runtime skills tests, ran the web build, and cross-checked the origin derivation against the resolver's own markers.

### 🟢 Effective-skill resolution goes through the audited loader, not a widened raw read

This is the right fix for #7757. `resolve_effective_skills` sources from `load_skills_for_agent_from_config` (the audited four-source union), explicitly not from the bundle-only `list_skills` raw `read_dir`. That preserves audit parity: the dashboard never starts surfacing open-skills/plugin frontmatter the runtime would have rejected. The new `GET /api/agents/{alias}/skills` handler gates on `require_auth` first and is read-only, matching the rest of `/api/skills/*`.

### 🟢 Origin derivation mirrors the resolver's discriminators rather than shadowing them

`derive_origin` classifies by the same markers the resolver itself sets — the `open-skills` tag (mod.rs:658), the `plugin:` name/tag prefix, then a `location` prefix match against a configured bundle dir, else workspace. Because each skill is loaded from exactly one source these branches are mutually exclusive in practice, and the precedence can't drift from the resolver. `editable` is server-derived and true only for `Bundle`, the only writable source, so a read-only source can't present an edit affordance even if the frontend gate regressed.

Local validation on 0c8832cc:

```
cargo test -p zeroclaw-runtime skills::service                       → 10 passed (incl. 2 new)
cargo clippy -p zeroclaw-runtime -p zeroclaw-gateway --all-targets   → clean
cd web && npm run build                                              → built
```

`derive_origin_classifies_each_source` and `resolve_effective_skills_unions_workspace_and_bundle` pin exactly the drift this fixes (the workspace skill the page used to miss now appears with correct origin/editable).

### 🟢 Deferrals are scoped sensibly

The server-side write-guard for non-bundle writes, the "skipped due to audit" distinction for the zero-skills case, and the `shadowed_by` collision indicator are all filed as follow-ups. Keeping this PR to the read/visibility path while bundle writes still resolve through `bundle::resolve_directory` is the right boundary; the frontend already gates writes on `editable`.

Approving. One note for the follow-up tracker: the plugin-origin arm wasn't built locally under `--features plugins-wasm` (build-disk constraint), so confirm the plugin-feature CI arm is green before merge. The origin logic is feature-agnostic string matching, so I don't consider it a blocker.

PR #7965, opened 2026-06-19
feat(channels/discord): interaction components — buttons, selects, modals, buttoned approval, autocomplete

Nillth:
## What this does

Completes the Discord interaction surface (EPIC B of tracker #7831): the channel can
now **render, emit, and dispatch** message components (buttons, selects), **modals**, a
**buttoned tool-approval** flow, and **slash autocomplete** — built on EPIC A's
`DiscordOutgoing` payload and the merged D-options interaction lifecycle.

Both the inbound (dispatch) and outbound (agent emits them) halves are wired — there is
**no agent-facing path left behind a `dead_code` allow** (see below).

- **Outbound, agent-driven:** an agent emits `[COMPONENTS:{"rows":[[…]]}]` to attach
  buttons / selects; an action button carries a `prompt` (enqueued as a turn on click), a
  `url` (link button), or a `modal` (opens a modal whose submit resolves into a turn). Each
  interactive component is registered server-side in a single-use registry at emit time.
- **Inbound dispatch:** type-3 (component) / type-5 (modal submit) share one lifecycle —
  parse `zc1` custom_id → cheap peer check → fail-closed `interaction_gate` → single-use
  `take` → resolve (approval / open-modal / into-turn). Selects route on the chosen
  `data.values[]` token.
- **Buttoned approval:** `request_approval` emits Allow-once / Session / Always / Deny
  buttons (plaintext prompt remains the fallback); a click resolves the parked approval
  `oneshot` from a **server-registered decision** — never from the wire.
- **Autocomplete (type 4):** answers inline (type-8) with up to 25 matching choices sourced
  from the skill option model, with a **side-effect-free** fail-closed gate.
- **Reaction edge:** `MESSAGE_REACTION_REMOVE_ALL` / `REMOVE_EMOJI` sweep orphaned archive
  rows.

## Robust marker parsing (live-hardened)

Real agents emit the `[COMPONENTS:{…}]` body as free-form JSON, and a model occasionally
slips on a deeply-nested payload. Two defects were found and fixed against **bytes captured
from a live daemon**, so the feature survives imperfect emission:

- **Dropped/mis-ordered closing bracket, parsed prose-safely.** A model emitted `…}]}]}]`
  instead of `…}]}]]}]` (omitting the `]` that closes the `rows` array), so the strict
  matcher never found the close and the whole marker leaked as raw text. `find_marker_extent`
  fixes this **without ever deleting user prose**, which is the load-bearing invariant: a
  `[COMPONENTS:` is stripped **only** when the slice up to its close is one valid JSON value —
  `serde_json::from_str` requires the *whole* slice to validate, so a slice contaminated with
  surrounding prose can never qualify. A balanced body uses the strict structural close; an
  under-closed body is bracket-repaired only when it is **trailing** (whitespace after the
  last `]`, so there is nothing to eat). Verified to repair the exact captured malformed
  payload back to valid JSON (3 rows), with regression tests pinning prose preservation for
  over-opened `{{…`, dropped-outer-`]`, a literal `[COMPONENTS:` typed in prose, and
  mid-message under-closed shapes.
- **Multiple markers.** An agent composing a "kitchen sink" may emit several
  `[COMPONENTS:…]` markers (one per row); only the first was honored and the rest leaked.
  `parse_component_markers` now consumes **every** marker and merges their rows in order
  (the builder caps the merged set at Discord's 5-row limit).
- **Components-only reply dropped.** `send()` forced a first message from `local_files` only,
  so a bodyless component reply (an action row with no text/attachments) produced zero
  chunks and the rows were silently dropped. The force-first flag now also accounts for
  components.

A marker that meets neither bar is left **verbatim** — its raw JSON may show (a recoverable
leak), but no surrounding words are ever lost, and the send never fails.

## Security

This gates **tool execution**, so the inbound path is fail-closed and was independently
security-reviewed **twice** (verdict SOUND both times — original dispatch and after the
D-options merge):

- Decisions/intents come only from **server-registered** state, never the wire `custom_id`.
- Fail-closed `interaction_gate` runs **before** any `take` / approval-resolve / modal-open
  / enqueue, in every arm (type-2/3/4/5) — locked by a source-level regression test.
- Single-use / replay-safe; forged/foreign/expired ids resolve nothing; timeout → Deny;
  autocomplete is side-effect-free; no token leakage.

## Code-quality

- **Zero `#[allow(dead_code)]` added** (AGENTS.md): every builder/codec/registry is wired to
  a real consumer; this branch in fact *removes* placeholders that existed on `master`.

## Testing

- `cargo test -p zeroclaw-channels --features channel-discord` — full suite green; adds
  regression tests for the **dropped-bracket repair**, the **multi-marker merge**, the
  **valid kitchen-sink** body, plus the existing marker/registry/approval/autocomplete
  coverage and the gate-before-take ordering regression.
- `cargo clippy --workspace --exclude zeroclaw-desktop --all-targets --features ci-all -- -D warnings`
  and `cargo fmt --all --check` clean.
- **32-bit Check:** unaffected — EPIC B is entirely under `channel-discord`, which the
  `Check (32-bit)` job excludes (`cargo check --no-default-features`); the markers code is
  also `usize`/`char_indices`-based with no pointer-width assumptions.

## Follow-up (not in this PR)

- **Streaming finalization (`finalize_draft`) does not yet render components** — components
  ride the non-streaming `send()` path (and slash/interaction replies). A focused follow-up
  will thread the same parse-and-attach through the draft-finalization path so components
  also render under `stream_mode = "partial"/"full"`.

Part of #7831.


singlerider (approved):
Reviewed at 7d3517b2 on a fresh checkout of `feat/discord-interaction-components`. This gates tool execution, so I read the inbound security path directly — the custom_id codec, the single-use registry, the gate, and the type-3/5 dispatch ordering — rather than relying on the prior security reviews, then ran the channels suite and clippy.

### 🟢 The interaction security model is sound: server-registered intent, fail-closed gate, no wire trust

The three load-bearing invariants all hold in source:

- **No wire trust.** A `custom_id` carries only a versioned `zc1|kind|arg` routing key with `\`/`|` escaping, never a secret or a decision. The meaning is registered server-side in `PendingComponents` at emit time; dispatch `take`s it on click. A forged, replayed, or expired id resolves to `None` and is refused. The `Approval` intent captures the `ApprovalDecision` as a fixed enum at emit time, so a crafted id cannot turn a Deny button into an Allow — confirmed in `pending.rs`.
- **Gate precedes every side effect.** `interaction_gate` (user allowlist sensitive-match, then guild, then channel filter, fail-closed) runs and early-returns on denial *before* `pending_components.take()` in the type-3 arm (mod.rs ~2732 then ~2758), and likewise before the approval-resolve and modal-open paths. An unauthorized click never even drains a registry entry. This ordering is locked by a source-level regression test, `interaction_arms_gate_before_take_after_the_doptions_merge`, which passes.
- **Single-use / replay-safe.** `take` removes the entry; a second click returns `None`. TTL-swept at register time so the map can't grow unbounded. `replay_resolves_nothing_after_first_click`, `forged_or_unknown_token_resolves_nothing`, and `second_click_is_refused_replay` all pin this.

### 🟢 Marker parsing preserves prose as the load-bearing invariant

`find_marker_extent` strips a `[COMPONENTS:` only when the slice up to its close is one valid JSON value, so a prose-contaminated slice can never qualify and surrounding words are never eaten. The under-closed repair is restricted to the trailing case (whitespace after the last `]`). The regression set is thorough and matches the live-captured defects: `repairs_dropped_rows_closing_bracket`, `mid_message_under_closed_marker_preserves_prose`, `over_opened_body_does_not_eat_prose`, `literal_tag_in_prose_does_not_eat_prose`, `garbled_marker_does_not_poison_a_later_valid_marker`, `merges_multiple_markers_and_strips_all`, `unparseable_body_left_verbatim_no_prose_lost`, `valid_kitchen_sink_payload_parses_to_three_rows`.

### 🟢 No agent-facing path left behind a dead_code allow

Both halves are wired — outbound `[COMPONENTS:…]` emission and inbound dispatch — and the components-only reply force-first fix closes the silent-drop case. Local validation at 7d3517b2:

```
cargo test -p zeroclaw-channels --lib discord::   → 255 passed; 0 failed
cargo clippy -p zeroclaw-channels --all-targets   → clean
```

CI is green across all 18 checks, which covers the full-workspace/plugin arm the author deferred locally on disk constraints.

Approving. This is a large surface, but it is partitioned into small focused modules (custom_id / pending / approval / markers / interaction), each with its security invariant pinned by a test. The fail-closed discipline and the prose-never-lost guarantee are exactly right for a tool-execution gate.

Nillth:
Rebased onto current master to resolve the conflict introduced by #7833's merge
(discord rich embeds).

Conflicts were in the discord channel modules:
- `crates/zeroclaw-channels/src/discord/types.rs` — the `DiscordOutgoing` envelope
  (both PRs added a real field: embeds vs components/flags).
- `crates/zeroclaw-channels/src/discord/markers.rs` — both PRs added a marker
  parser to the same spot (`[EMBED:{…}]` vs `[COMPONENTS:{…}]`).
- `crates/zeroclaw-channels/src/discord/mod.rs` — the send + interaction-reply
  paths (embed rendering vs component rendering).
- `crates/zeroclaw-channels/src/discord/interaction.rs` and
  `crates/zeroclaw-channels/src/discord/rest.rs` — the edit/send helpers.
- `crates/zeroclaw-channels/src/orchestrator/mod.rs` — the Discord system-prompt
  help text (EMBED instruction vs COMPONENTS instruction).

Resolved by keeping BOTH the embed and the interaction-component code and
integrating them: `DiscordOutgoing` now carries `embeds` AND
`components`/`flags`; both markers are parsed (embeds first, then components on
the stripped text); `discord_edit_interaction_response` /
`deliver_interaction_answer` carry both embeds and components on the first reply
chunk; and the Discord help text lists both the `[EMBED:…]` and `[COMPONENTS:…]`
(incl. modal) affordances. Master had renamed the multipart sender and its
payload now serializes components too, so I routed the file-send path through the
single `send_discord_message_payload_with_files` chokepoint and dropped my
now-duplicate `send_discord_outgoing_with_files` — no dead code, no
`#[allow(dead_code)]`. No semantic/feature changes; behaviour is preserved.

CI-equivalent local validation green (cargo +1.93.0):
- `fmt --all -- --check`: clean
- `build -p zeroclaw-channels --all-targets`: clean (no warnings)
- `clippy -p zeroclaw-channels --all-targets`: clean (no warnings)
- `test -p zeroclaw-channels`: 1073 passed, 0 failed, 1 ignored

PR #7969, opened 2026-06-19
refactor(runtime): bundle run_tool_call_loop args into a ToolLoop struct

Nillth:
## What

`run_tool_call_loop` carried 34 positional arguments and a
`#[allow(clippy::too_many_arguments)]`. This extracts them into a borrowed
params struct `ToolLoop<'a>` (defined next to the function), refactors the
function and every call site to use it, and drops the `too_many_arguments`
allow.

- The function body destructures the struct once at entry
  (`let ToolLoop { .. } = p;`, preserving the `&mut` working sets) and is
  otherwise **byte-identical** — the diff is "positional signature → struct +
  destructure" and nothing else.
- All call sites build the struct instead: 6 production
  (channel orchestrator, the two `Agent` turn paths, the skill-review and
  delegate tools, and the `agent_turn` wrapper) plus the in-crate test sites.
- `LoopKnobs` is kept as-is (a field on the new struct), not dissolved.

## Why

Pure cleanup that stands on its own — but it's also the precursor to the
universal ingress policy layer in **#6971**. Doing this first means the ingress
context rides as a single struct *field*, instead of becoming a 36th positional
argument on an already-overloaded signature.

## Scope

`agent_turn` keeps its own (smaller) signature for now — folding it in is a
natural follow-up, deliberately out of this PR to keep the change a single,
mechanical refactor.

## Not a behavior change

No runtime logic and no test assertions changed. Verified:

- `cargo fmt` clean; `cargo clippy -p zeroclaw-runtime -p zeroclaw-channels
  --all-targets -- -D warnings` clean (the `too_many_arguments` allow is gone,
  not suppressed).
- `cargo test` on the touched crates green (the `run_tool_call_loop` and
  safety-net oracle suites pass unchanged); no caller of these symbols exists
  outside the two crates.
- Engine-validation harness, live tier: all 5 transport legs pass — CLI
  one-shot, delegate round, RPC socket turn, WS turn with mid-turn steering,
  and ACP stdio incl. deny-with-edit.

Part of the #6971 ingress series (1/N): cleanup → ingress layer → per-caller
context.


singlerider (approved):
Reviewed at c870cdff on a fresh checkout of `refactor/tool-loop-params`. For a pure parameter-bundling refactor the question is whether it's a faithful mechanical transform, so I checked the struct against the old signature, the entry destructure, and the test suite across every call site.

### 🟢 Faithful mechanical transform — struct mirrors the signature, body is unchanged

`ToolLoop<'a>` (turn/mod.rs:163) carries exactly the former 34 parameters with identical types and lifetimes, including the four `&'a mut` working sets (`history`, `steering`, `new_messages_out`, `image_cache`) — so mutability semantics are preserved, not silently downgraded. `run_tool_call_loop` destructures the whole struct at entry into the same binding names the body already used, which is why the body below is byte-identical and references resolve unchanged. `LoopKnobs` is kept as a field rather than dissolved, as stated.

### 🟢 The `too_many_arguments` allow is removed, not suppressed

`run_tool_call_loop` no longer carries `#[allow(clippy::too_many_arguments)]`, and `cargo clippy -p zeroclaw-runtime -p zeroclaw-channels --all-targets` is clean — the struct dissolves the lint rather than hiding it. The remaining `too_many_arguments` allows in the crate are on other functions this PR doesn't touch.

Behavior-preservation verified by the unchanged suites across all six call sites:

```
cargo test -p zeroclaw-runtime --lib run_tool_call_loop  → 47 passed; 0 failed
cargo test -p zeroclaw-runtime --lib safety_net          → 22 passed; 0 failed
cargo test -p zeroclaw-channels --lib orchestrator       → 318 passed; 0 failed
```

CI is green across all 18 checks. No assertions changed, which is the right signal for a refactor.

### 🟢 Scope discipline

Keeping `agent_turn`'s own smaller signature out of this PR and landing the bundling as a single mechanical change is the right call — it makes this trivially reviewable and sets up the #6971 ingress context to ride as one struct field instead of a 36th positional arg.

Approving. Clean precursor refactor with no behavior delta.

PR #7970, opened 2026-06-19
[Bug]: MCP stdio child processes accumulate on daemon with heartbeat.enabled=true (one orp (#5903)

### LINKED ISSUE #5903: [Bug]: MCP stdio child processes accumulate on daemon with heartbeat.enabled=true (one orphan per tick)

perlowja:
## Summary

- `crates/zeroclaw-tools/src/mcp_transport.rs`

## Linked issue

Resolves zeroclaw-labs/zeroclaw#5903

`risk: low` · `size: S`

## Testing (required)

> VERDICT: NEEDS-ATTENTION
> FINDINGS:
> - crates/zeroclaw-tools/src/mcp_transport.rs:13 MED: Missing imports for `std::process::Command`, `tokio::time::timeout`, and `std::time::Duration`. The added `reset` implementation uses these symbols but they are not brought into scope, causing compilation failures. Add the appropriate `use` statements.
> - crates/zeroclaw-tools/src/mcp_transport.rs:228 HIGH: `StdioTransport` does not satisfy the `Sync` requirement of the `McpTransportConn` trait because it contains `tokio::process::Child`, which is not `Sync`. This violates the trait bound `Send + Sync` and will fail to compile. Consider removing `Sync` from the trait, wrapping the struct in a mutex, or redesigning the transport to be thread‑safe.
> - crates/zeroclaw-tools/src/mcp_transport.rs:237 MED: The `timeout` call’s result is ignored. If the child does not exit within 10 seconds, the function proceeds to respawn a new process while the old one may still be running, leading to resource leaks and undefined behavior. Handle the `Result` and return an error on timeout.
> - crates/zeroclaw-tools/src/mcp_transport.rs:242 LOW: In `close`, the child is killed with `start_kill()` but the process is not

## Compatibility (required)

- Backward compatible? Yes

## Rollback

### DIFF
--- a/crates/zeroclaw-tools/src/mcp_transport.rs
+++ b/crates/zeroclaw-tools/src/mcp_transport.rs
@@ -1,6 +1,7 @@
 //! MCP transport abstraction — supports stdio, SSE, and HTTP transports.
 
 use std::borrow::Cow;
+use std::process::Stdio as ProcStdio;
 
 use anyhow::{Context, Result, bail};
 use tokio::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
@@ -93,7 +94,8 @@ pub trait McpTransportConn: Send + Sync {
     async fn send_and_recv(&mut self, request: &JsonRpcRequest) -> Result<JsonRpcResponse>;
 
     /// Reset per-connection session state so the next operation re-establishes
-    /// a fresh session. Default is a no-op for stateless transports (stdio).
+    /// a fresh session. Default is a no-op for stateless transports (HTTP, SSE).
+    /// `StdioTransport` overrides this to kill and re-spawn the child process.
     async fn reset(&mut self) -> Result<()> {
         Ok(())
     }
@@ -109,6 +111,10 @@ pub struct StdioTransport {
     _child: Child,
     stdin: tokio::process::ChildStdin,
     stdout_lines: tokio::io::Lines<BufReader<tokio::process::ChildStdout>>,
+    /// Spawning parameters retained so the child can be re-spawned on reset.
+    spawn_cmd: String,
+    spawn_args: Vec<String>,
+    spawn_env: std::collections::HashMap<String, String>,
 }
 
 impl StdioTransport {
@@ -116,9 +122,9 @@ impl StdioTransport {
         let mut child = Command::new(&config.command)
             .args(&config.args)
             .envs(&config.env)
-            .stdin(std::process::Stdio::piped())
-            .stdout(std::process::Stdio::piped())
-            .stderr(std::process::Stdio::inherit())
+            .stdin(ProcStdio::piped())
+            .stdout(ProcStdio::piped())
+            .stderr(ProcStdio::inherit())
             .kill_on_drop(true)
             .spawn()
             .with_context(|| format!("failed to spawn MCP server `{}`", config.name))?;
@@ -152,6 +158,9 @@ impl StdioTransport {
         let stdout_lines = BufReader::new(stdout).lines();
 
         Ok(Self {
+            spawn_cmd: config.command.clone(),
+            spawn_args: config.args.clone(),
+            spawn_env: config.env.clone(),
             _child: child,
             stdin,
             stdout_lines,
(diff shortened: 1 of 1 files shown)


perlowja:
Closing as a duplicate from the overnight build batch. Keeping #7991 as the single PR for issue #5903.

PR #7971, opened 2026-06-19
static_voice_peers caches config-derived voice peers on the channel handle (latent SSOT vi (#7795)

### LINKED ISSUE #7795: static_voice_peers caches config-derived voice peers on the channel handle (latent SSOT violation)
### Summary

`TelegramChannel.static_voice_peers` is a `HashSet<String>` cached on the channel handle, seeded once at construction by `with_voice_peer_prefs` from `config.peer_groups` (groups with `output_modality = "voice"`). Its own doc comment states it is "Populated once at startup… never mutated by session events."

That is a copy of state whose source of truth is `config.peer_groups`, held inside a channel handle — the shape AGENTS.md explicitly forbids:

> A channel handle struct holding `Vec<String>` of "authorized users" alongside `peer_groups` in `Config`.

It is the same failure mode as the SSOT rule's origin incident: a cached `allowed_users` list serving stale data until daemon restart.

### Current impact: latent, not active

It does not drift today. The only path that changes `peer_groups` is a config edit, which goes through `/admin/reload` → channel rebuild → `with_voice_peer_prefs` re-seeds the cache. No production runtime path mutates `peer_groups` in place.

### Risk

This is a latent violation. Any future runtime mutation of voice-related `peer_groups` — the analogue of the pairing path that writes `allowed_users` live — would leave `static_voice_peers` stale, and voice routing would silently serve the old set until a reload or restart. That is precisely the regression class the SSOT rule exists to pre-empt.

### Suggested direction

Resolve voice peers on-demand from the live `Arc<RwLock<Config>>` handle the channel already holds (`self.persist`), the same live-resolve pattern used by `channel_external_peers`, instead of seeding a cache at construction.

### Context

Surfaced while rebasing #7361 onto the turn-engine / voice-peer refactor. Not blocking #7361 (that PR now conforms to the `static_voice_peers` mechanism); raising separately so the cache-vs-live-resolve choice is a deliberate one.

perlowja:
## Summary

Automated build resolving the linked issue. Diff touches 2 file(s):
- `crates/zeroclaw-channels/src/telegram.rs`

## Linked issue

Resolves zeroclaw-labs/zeroclaw#7795

`risk: low` · `size: S`

## Testing (required)

> VERDICT: NEEDS-ATTENTION
> FINDINGS:
> - **mod.rs:5715** HIGH: `let cfg = config_arc.read();` returns a `Result<RwLockReadGuard<Config>, PoisonError>`. The code treats it as a guard, causing a compile error. Fix by unwrapping or handling the error, e.g. `let cfg = config_arc.read().expect("RwLock poisoned")`.
> - **mod.rs:5715** HIGH: The closure captures `config_arc` but does not import `std::sync::RwLock` in this file (if not already imported). Ensure `use std::sync::RwLock;` is present.
> - **mod.rs:5722** MED: The resolver allocates a new `Vec<String>` on every call and also creates a new `HashSet` each time. While functionally correct, this may be a performance concern for high‑frequency `is_voice_chat` checks. Consider caching the set inside the closure with interior mutability if appropriate.
> - **telegram.rs:1052** MED: `self.voice_peer_resolver)().contains(&recipient.to_string())` creates a new `String` for each check, leading to unnecessary allocation. Use `contains(&recipient)` after converting the resolver output to `Vec<&str>` or store a `HashSet<String>` inside the resolver.
> - **telegram.rs:623** LOW: The default resolver `Arc::new(|| vec![])` returns an empty `Vec<String>` ea

## Compatibility (required)

- Backward compatible? Yes

## Rollback

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -5706,6 +5706,42 @@ fn one_shot_channel_workspace_dir(config: &Config, channel_type: &str, alias: &s
     config.channel_workspace_dir(&format!("{channel_type}.{alias}"))
 }
 
+/// Build an on-demand resolver for voice-preference peers.
+///
+/// Returns a closure that reads `[peer_groups.*]` entries whose channel
+/// matches `channel_type` (or `"<channel_type>.<alias>"`) and whose
+/// `output_modality` is `Voice`. The resolver is called at each
+/// `is_voice_chat` check, ensuring no config state is cached on the
+/// channel handle (SSOT: AGENTS.md "ABSOLUTE RULE").
+fn voice_peer_resolver_for_channel(
+    config_arc: Arc<RwLock<Config>>,
+    channel_type: &'static str,
+    alias: &str,
+) -> Arc<dyn Fn() -> Vec<String> + Send + Sync> {
+    use zeroclaw_config::multi_agent::OutputModality;
+    let alias = alias.to_string();
+    Arc::new(move || {
+        let cfg = config_arc.read();
+        let mut out: Vec<String> = Vec::new();
+        let mut seen = std::collections::HashSet::new();
+        for group in cfg.peer_groups.values() {
+            let group_matches = match group.channel.split_once('.') {
+                Some((ty, al)) => ty == channel_type && al == alias,
+                None => group.channel == channel_type,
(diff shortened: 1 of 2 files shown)


perlowja:
Closing as a duplicate from the overnight build batch. Keeping #7982 as the single PR for issue #7795.

PR #7972, opened 2026-06-19
fix(multimodal): honor vision_provider for inbound images

### LINKED ISSUE #6841: [Bug]: [multimodal] vision_provider silently ignored — inbound images routed to providers.fallback instead
### Affected component

provider

### Severity

S1 - workflow blocked

### Current behavior

When [multimodal].vision_provider and [multimodal].vision_model are configured to a registered provider, and [media_pipeline].enabled = true with describe_images = true, inbound channel messages containing images are still routed to [providers].fallback instead of the configured vision provider.
The configuration is loaded and parsed correctly — zeroclaw config schema confirms both multimodal.vision-provider and multimodal.vision-model are populated with the expected values. The runtime simply does not consult them when dispatching image-bearing messages.

### Expected behavior

When [multimodal].vision_provider is set to a registered provider key and an inbound message contains an [IMAGE:...] marker (or equivalent image attachment surfaced by the channel layer), the orchestrator should dispatch the request to that vision provider rather than to [providers].fallback.
This is what the published config docs at docs.zeroclawlabs.ai/en/reference/config.html describe:

vision_model_provider: Provider name to use for vision/image messages. When set, messages containing [IMAGE:] markers are routed to this provider instead of the default text provider.

(Note: the docs use vision_model_provider, but the actual schema field name in the binary is vision_provider. Both should resolve to the same field, or the docs should be updated. Separate, smaller issue.)

### Steps to reproduce

```bash
Configure two custom providers:


schema_version = 2

[providers]
fallback = "custom:http://text-llm.example.local:11434/v1"

[providers.models."custom:http://text-llm.example.local:11434/v1"]
base_url = "http://text-llm.example.local:11434/v1"
model = "some-text-only-model"
timeout_secs = 120
wire_api = "chat_completions"
max_tokens = 4096
merge_system_into_user = true

[providers.models."custom:http://vision-llm.example.local:11434/v1"]
base_url = "http://vision-llm.example.local:11434/v1"
model = "some-vision-capable-model"
timeout_secs = 180
wire_api = "chat_completions"
max_tokens = 4096
merge_system_into_user = true

[multimodal]
vision_provider = "custom:http://vision-llm.example.local:11434/v1"
vision_model = "some-vision-capable-model"

[media_pipeline]
enabled = true
describe_images = true

[channels.telegram]
enabled = true
bot_token = "REDACTED"
allowed_users = ["REDACTED"]


Confirm config is loaded correctly:

$ zeroclaw config schema | grep -E '(vision-provider|vision-model|media-pipeline)'
multimodal.vision-provider = custom:http://vision-llm.example.local:11434/v1
multimodal.vision-model    = some-vision-capable-model
media-pipeline.enabled     = true
media-pipeline.describe-images = true

Send an image to the Telegram channel.
Observe orchestrator log — provider= is the text provider URL (the fallback), not the vision provider URL.
Expected: provider=custom:http://vision-llm.example.local:11434/v1. Actual: provider=custom:http://text-llm.example.local:11434/v1.
```

### Impact

Operators wanting the documented "cheap text router + heavy vision model" split cannot use the feature. Workaround is to set the vision endpoint as global fallback, which forces every text message through a slower, more expensive multimodal model. Particularly painful for local/on-device deployments where text routing on an 8B multimodal model is several times slower than on a small text-only model that would otherwise be ideal for the dispatch role.

### Logs / stack traces

```text
💬 [telegram] from REDACTED: [IMAGE:/zeroclaw-data/workspace/telegram_files/photo_REDACTED.jpg]
⏳ Processing message...
INFO zeroclaw_channels::orchestrator: ⏱ Memory recall completed mem_recall_ms=2 sender_empty=true group_empty=true
INFO zeroclaw_channels::orchestrator: ⏱ Starting LLM call elapsed_before_llm_ms=3479
WARN zeroclaw_runtime::agent::loop_: Context window exceeded, attempting in-loop recovery iteration=1
ERROR zeroclaw_runtime::agent::loop_: Context overflow unrecoverable: no trimmable messages
ERROR LLM error: Request exceeds model context window.
  Attempts: provider=custom:http://text-llm.example.local:11434/v1 model=<text-only-model>
            attempt 1/3: non_retryable; error=Custom API error (400 Bad Request):
            request (221702 tokens) exceeds the available context size (16384 tokens)
```

### ZeroClaw version

zeroclaw 0.7.5

### Rust version

_No response_

### Operating system

Kuberentes 

### Regression?

Unknown

### Pre-flight checks

perlowja:
## Summary

## Linked issue

Resolves zeroclaw-labs/zeroclaw#6841

`risk: low` · `size: S`

## Testing (required)

> VERDICT: NEEDS-ATTENTION
> FINDINGS:
> - <crates/zeroclaw-channels/src/orchestrator/mod.rs:3845> HIGH: The new condition `ctx.multimodal.vision_model_provider.is_some()` assumes `ctx.multimodal` is always present. If `multimodal` is an `Option` or can be `None`, this will panic at runtime. Add a guard or unwrap safely. <fix>: `ctx.multimodal.as_ref().map_or(false, |m| m.vision_model_provider.is_some())`.
> - <crates/zeroclaw-channels/src/orchestrator/mod.rs:3845> MED: Setting `vision` to true solely based on the presence of a `vision_model_provider` does not verify that the provider actually supports vision. If the provider lacks vision capability, later code may attempt vision processing and panic or produce incorrect results. <fix>: also check `ctx.multimodal.vision_model_provider.as_ref().unwrap().supports_vision()`.
> - <crates/zeroclaw-channels/src/orchestrator/mod.rs:3849> LOW: The call to `TranscriptionManager::new(...).ok()` silently discards initialization errors, potentially hiding misconfiguration of the transcription subsystem. Consider handling the error explicitly or propagating it. <fix>: replace `.ok()` with proper error handling, e.g., `?` or logging the error.
> 
> [openrevie

## Compatibility (required)

- Backward compatible? Yes

## Rollback

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -3848,8 +3848,8 @@ async fn process_channel_message_body(
         // strips image data to a plain filename ref, and by the time the
         // runtime's vision_route.rs correctly routes to the vision provider
         // the raw image bytes are gone.
-        let vision = ctx.model_provider.supports_vision()
-            || ctx.multimodal.vision_model_provider.is_some();
+        let vision =
+            ctx.model_provider.supports_vision() || ctx.multimodal.vision_model_provider.is_some();
         let transcription_manager =
             crate::transcription::TranscriptionManager::new(&ctx.transcription_config)
                 .ok()
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -3841,7 +3841,15 @@ async fn process_channel_message_body(
 
     // ── Media pipeline: enrich inbound message with media annotations ──
     if ctx.media_pipeline.enabled && !msg.attachments.is_empty() {
-        let vision = ctx.model_provider.supports_vision();
+        // Check the default provider AND the configured vision route: when a
+        // dedicated vision_model_provider is set (via [multimodal]), the media
+        // pipeline must embed base64 image data even if the default text
+        // provider lacks vision capability. Otherwise the media pipeline
+        // strips image data to a plain filename ref, and by the time the
+        // runtime's vision_route.rs correctly routes to the vision provider
+        // the raw image bytes are gone.
+        let vision = ctx.model_provider.supports_vision()
+            || ctx.multimodal.vision_model_provider.is_some();
         let transcription_manager =
             crate::transcription::TranscriptionManager::new(&ctx.transcription_config)
                 .ok()
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -17323,6 +17323,120 @@ BTC is currently around $65,000 based on latest tool output."#
     }
 
     #[tokio::test]
+    async fn media_pipeline_preserves_image_bytes_when_vision_route_configured() {
+        use wiremock::matchers::{body_string_contains, method, path};
+        use wiremock::{Mock, MockServer, ResponseTemplate};
+
+        let channel_impl = Arc::new(RecordingChannel::default());
+        let channel: Arc<dyn Channel> = channel_impl.clone();
+
+        let mut channels_by_name = HashMap::new();
+        channels_by_name.insert(channel.name().to_string(), channel);
+
+        let provider_impl = Arc::new(HistoryCaptureModelProvider::default());
+        let vision_server = MockServer::start().await;
+        let _vision_mock = Mock::given(method("POST"))
+            .and(path("/chat/completions"))
+            .and(body_string_contains("data:image/png;base64,AQIDBA=="))
+            .respond_with(ResponseTemplate::new(200).set_body_json(serde_json::json!({
+                "choices": [
+                    {
+                        "message": {
+                            "role": "assistant",
+                            "content": "vision saw bytes"
+                        }
+                    }
+                ]
+            })))
+            .expect(1)
+            .mount_as_scoped(&vision_server)
+            .await;
+
+        let base_ctx = peer_prompt_test_context(
+            channels_by_name,
+            provider_impl.clone(),
+            Arc::new(zeroclaw_config::schema::Config::default()),
+            Arc::new(vec![]),
+        );
+        let runtime_ctx = Arc::new(ChannelRuntimeContext {
+            multimodal: zeroclaw_config::schema::MultimodalConfig {
+                vision_model_provider: Some(format!("custom:{}", vision_server.uri())),
+                vision_model: Some("test-vision-model".to_string()),
+                ..Default::default()
+            },
+            media_pipeline: zeroclaw_config::schema::MediaPipelineConfig {
+                enabled: true,
+                describe_images: true,
+                ..Default::default()
+            },
+            ..(*base_ctx).clone()
+        });
+
+        process_channel_message(
+            runtime_ctx,
+            zeroclaw_api::channel::ChannelMessage {
+                id: "msg-image-route".to_string(),
+                sender: "alice".to_string(),
+                reply_target: "chat-image-route".to_string(),
+                content: "please inspect this".to_string(),
+                channel: "test-channel".into(),
+                channel_alias: None,
+                timestamp: 1,
+                thread_ts: None,
+                interruption_scope_id: None,
+                attachments: vec![zeroclaw_api::media::MediaAttachment {
+                    file_name: "route.png".to_string(),
+                    data: vec![1, 2, 3, 4],
+                    mime_type: Some("image/png".to_string()),
+                }],
+                subject: None,
+            },
+            CancellationToken::new(),
+        )
+        .await;
+
+        {
+            let calls = provider_impl.calls.lock().unwrap();
+            assert!(
+                calls.is_empty(),
+                "default non-vision provider must not receive an image-bearing turn: {calls:?}"
+            );
+        }
+
+        let sent_messages = channel_impl.sent_messages.lock().await;
+        assert_eq!(
+            sent_messages.len(),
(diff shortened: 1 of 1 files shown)


singlerider (approved):
First review (no prior reviews, comments, or inline threads). Fetched the head locally and traced the media pipeline against `vision_route.rs` and the config schema. CI is 18/18 green.

Closes #6841: inbound images were routed to `providers.fallback` instead of the configured `[multimodal].vision_model_provider`, because the media pipeline stripped image bytes to a filename ref before vision routing ran, leaving nothing for the route to send.

### 🟢 What looks good — the gate is widened at exactly the right layer, and the comment earns its place

`process_channel_message_body` previously set `vision` from `ctx.model_provider.supports_vision()` alone (orchestrator/mod.rs:3852). The fix ORs in `ctx.multimodal.vision_model_provider.is_some()`, so when an operator configures a dedicated vision route the pipeline preserves the base64 image data even when the default text provider has no vision. That `vision` bool only controls embed-bytes-vs-strip-to-ref in `MediaPipeline::new`, which is precisely the byte-preservation decision and nothing more. The block comment explains the non-obvious drift (bytes gone by the time `vision_route.rs` routes) rather than narrating the line, which is the right use of a comment.

### 🟢 What looks good — the adversarial findings are correctly marked advisory; all three are non-issues

I checked each against the source so they don't linger:

- HIGH "panic if `ctx.multimodal` is `None`": false. `multimodal` is a plain `MultimodalConfig` value on the ctx struct (mod.rs:446), not an `Option`. `.is_some()` applies to the `vision_model_provider: Option<String>` field (schema.rs:5511), not to `multimodal`. No unwrap, no panic path.
- MED "doesn't verify the routed provider supports vision": already handled downstream and at the correct layer. `vision_route.rs:53` constructs the configured provider and gates on `vp_instance.supports_vision()`, surfacing a capability error otherwise. The media pipeline's only job is to keep the bytes; the capability check belongs to the route, where it lives.
- LOW "`.ok()` on `TranscriptionManager::new` swallows errors": pre-existing code, untouched by this diff. Out of scope for a one-line vision-gate fix.

### 🟢 What looks good — minimal, single-file, single-concern

One file, +9/−1, no new config surface, backward compatible (default provider with vision still works exactly as before since the OR short-circuits on the existing condition). The applied labels (`bug` / `size: XS` / `risk: medium` / `channel`) are correct; the PR body's "risk: low · size: S" line is stale relative to the labels but the labels are the right call for a media-pipeline behavior change.

Approving.

Audacity88 (dismissed):
I reviewed the live PR at head 50747860e2a7fa332d9bbd59d53e92e5544359db, the linked issue #6841, the current diff, the green GitHub checks, and singlerider's approval review. I also checked the public source around the media pipeline, `vision_route.rs`, the multimodal config schema, and custom provider construction.

I did not run local Cargo validation or a live Telegram smoke for this review.

I agree with the source-level direction in the existing approval: this bug is mainly about preserving image bytes early enough for the runtime vision route to use them. I still want one focused proof point before this closes #6841, because the PR changes a channel/media behavior boundary and the PR body currently says there was no manual runtime verification beyond CI.

### 🟢 What looks good — the byte-preservation gate is the right layer

The media pipeline decides whether image attachments become `[IMAGE:data:...]` markers before the runtime vision route gets a chance to switch providers. Extending that gate to consider `multimodal.vision_model_provider` is the right shape: a dedicated vision route needs the raw image bytes preserved even when the default text provider is not the vision target.

I also checked the advisory notes in the PR body. `ctx.multimodal` is a concrete config value, the downstream vision route still validates the selected provider's vision capability, and the `TranscriptionManager::new(...).ok()` behavior is pre-existing.

### 🔴 Blocking — add evidence for the reported split-provider channel turn

#6841 is specifically about an inbound channel image with `[media_pipeline].enabled = true`, a text-default provider, and a configured `[multimodal].vision_model_provider`. The current patch changes the preservation predicate, but the PR does not add a regression test or near-live smoke showing that this exact turn now reaches the configured vision provider/model instead of falling back to the text provider.

The existing source trace is useful, but source trace plus green CI is not the same as a regression for the reported channel turn. The visible diff adds no test, and the PR body says no additional manual runtime verification was done beyond CI. That leaves the issue-closing evidence one step short of the behavior #6841 asked us to prove.

Please add focused evidence for that path before this closes #6841. A mocked channel/runtime regression is enough; it does not need real Telegram credentials. The test should cover a non-vision default provider, `media_pipeline.enabled` / `describe_images`, a configured `vision_model_provider` and `vision_model`, and an inbound image attachment. It should fail on the old code by showing the image bytes are stripped or the fallback provider is used, and pass on this branch by showing the configured vision provider/model receives the image-bearing turn.

perlowja:
@Audacity88 thanks — added the regression you asked for in 4585b1024 (`media_pipeline_preserves_image_bytes_when_vision_route_configured`). It's a wiremock-backed test for the exact #6841 turn: a non-vision default provider, `media_pipeline.enabled` + `describe_images`, a configured `[multimodal].vision_model_provider`/`vision_model`, and an inbound image attachment.

The mock vision endpoint only matches a `/chat/completions` body containing the preserved `data:image/png;base64,…` bytes, so the test asserts the configured vision provider actually **receives the image-bearing turn** (and that the default text provider is never called). I confirmed it fails on the old gate — reverting the `|| ctx.multimodal.vision_model_provider.is_some()` clause makes the default text provider receive the stripped `[Image: route.png attached]` turn — and passes with the fix. Full-workspace CI (1.93.0 fmt / clippy `--features ci-all` / nextest) is green. Re-requesting your review.

perlowja:
Closing to clear the upstream branch list: this branch was pushed to the canonical repo by mistake (inverted `origin`/fork remotes). The change is preserved on my fork as `perlowja:fix/multimodal-vision-provider` and will be resubmitted as a proper fork-based PR. A fork-only pre-push guard is now in place so this cannot recur. Apologies for the upstream-branch pollution.

PR #7973, opened 2026-06-19
fix(agent): self-contained context-compression summary provider

### LINKED ISSUE #7964: [Bug]: context_compression.summary_model is provider-specific on a shared runtime profile (cross-provider failure)
## Affected component
agent/context-compression / config

## Severity
S2 - degraded behavior (silent runtime failure on a reachable config)

## Current behavior
`runtime_profiles.<p>.context_compression.summary_model` is an `Option<String>` — a **bare model id** (`crates/zeroclaw-config/src/scattered_types.rs:305`). The context compressor reads it and **dispatches it onto the AGENT's own provider**, only swapping the model string (`crates/zeroclaw-runtime/src/agent/context_compressor.rs:335`):
```rust
let summary_model = self.config.summary_model.as_deref().unwrap_or(model);
// … ProviderDispatch::from_ref(model_provider)  // model_provider == the agent's
```
But `context_compression` lives on `RuntimeProfileConfig`, which is **shared** — multiple agents reference one profile via `runtime_profile`. So when a profile is shared by agents on **different** providers:

- `profileA.context_compression.summary_model = "claude-haiku-4-5"`
- Agent1 (provider1 = anthropic) → sends `claude-haiku-4-5` to anthropic ✓
- Agent2 (provider2 = openai) → sends `claude-haiku-4-5` to **openai** ✗ → invalid model, runtime failure

It's a **provider-specific value stored on a provider-agnostic, shareable object**, applied to whatever provider the agent uses. Nothing validates it at config time — it fails only when that agent actually compresses. There is also no way for an agent to override the profile's summarizer.

## Expected behavior
The summarizer should be specified **self-contained** and overridable per agent:
1. Make the value a `providers.models.<type>.<alias>` reference (`ModelProviderRef`), mirroring the shipped `classifier_provider` field — it bundles provider + model + auth, so summarization runs on the *referenced* provider regardless of the agent's provider. Cross-provider mismatch becomes structurally impossible, it validates at config time, and the dashboard renders the existing AliasRef dropdown.
2. Add an **agent-level override** with precedence **agent → profile → default** (default = the agent's own provider+model), matching the `effective_*` resolver idiom.
3. **Non-breaking:** introduce `summary_provider` (both levels) and keep the deprecated `summary_model` as a fallback (read only when `summary_provider` is empty) so existing configs run byte-identically; provide a `schema_version 3→4` migration that rewrites each agent's legacy bare model onto its own provider's alias, warning where the bare model doesn't match the agent's provider (surfacing each previously-silent cross-provider mismatch).

## Steps to reproduce
Configure two agents on different providers (`agents.a.model_provider = anthropic.default`, `agents.b.model_provider = openai.default`) both using one `runtime_profile` whose `context_compression.summary_model` is set to an Anthropic-only model id. Drive enough history on agent `b` to trigger compression → the summary call sends the Anthropic model id to OpenAI and fails. No config-time error.

## Impact
Operators who share a runtime profile across agents on heterogeneous providers and set a profile-level `summary_model`. Low frequency, but a silent, hard-to-diagnose runtime failure on a fully valid-looking config.

## Regression?
No — design gap present since `summary_model` landed on the runtime profile.

## Notes
Designed direction confirmed via analysis (encoding decision: a `ModelProviderRef`, cloning the `classifier_provider` idiom — self-contained, free dropdown + validation; the "cheaper model on a configured provider" case is the documented pattern of defining a provider alias and pointing at it). Two correctness notes for the fix: media-marker preservation must key off the *summary* provider's `supports_vision()` (not the agent's), and there are **three** compressor call sites (loop_.rs ×2 + the orchestrator's proactive path) that must all be covered.

Nillth:
## Summary

Fixes #7964. `runtime_profiles.<p>.context_compression.summary_model` was a bare model id, and the compressor dispatched it onto the **agent's own** provider (only swapping the model string). Since `context_compression` lives on the **shared** `RuntimeProfileConfig`, a profile referenced by agents on **different** providers sent one model id to the wrong provider — a silent runtime failure on a fully valid-looking config, with no agent-level override and no config-time check.

This replaces it with a **self-contained, overridable provider reference**:

- **New `summary_provider`** (a `providers.models.<type>.<alias>` `ModelProviderRef`) on **both** the profile's `context_compression` and a **new agent-level override** on `AliasedAgentConfig`. Resolution is **agent → profile → default** (`Config::effective_summary_provider`), mirroring the shipped `classifier_provider` "empty = inherit" idiom. Because the ref bundles provider + model + auth, summarization runs on the *referenced* provider regardless of which agent uses the profile — the cross-provider mismatch becomes structurally impossible. It **validates at config time** (dangling ref → `DanglingReference`, at both agent and profile level) and renders the existing **AliasRef dropdown** for free.
- **Compressor**: `compress_*` now take a resolved `SummaryTarget { provider, model, preserve_media_markers }`; the in-config `summary_model` read is removed. All **three** call sites resolve/build the summary provider and pass it: the loop's recovery (`compress_on_error`) and post-turn (`compress_if_needed`) sites, **and** the orchestrator's proactive pre-call path (the one a partial fix would miss).
- **Vision correctness**: `preserve_media_markers` is `true` **only** when summarizing on the agent's *own* vision-capable provider; a distinct summary provider ⇒ `false` (strip markers), so a local `[IMAGE:/path]` is never shipped to a possibly text-only summary model.
- **Fail-open**: an unresolvable/unbuildable summary provider ⇒ WARN once and fall back to the agent's own provider + model — compression never hardens into a turn-killing error (mirrors `resolve_classifier_route`).
- **Non-breaking**: the deprecated `summary_model` is kept as a fallback (read only when `summary_provider` is empty), so existing configs run byte-identically. The alias rename/delete/dangling-collect sweep covers `summary_provider` at both levels.

**No schema bump / no forced migration.** `summary_provider` is purely additive (`#[serde(default)]`), so this is **not** a `schema_version` change — existing configs load unchanged and keep their exact behavior via the deprecated `summary_model` fallback. Operators adopt `summary_provider` (now config-validated + AliasRef-dropdown) at their own pace. (Bumping the schema for a single additive parameter would be disproportionate; a bare `summary_model` that differs from the agent's own model can't be auto-expressed as a self-contained ref without a new provider alias anyway, so there's nothing a migration could safely rewrite for that case beyond a warning.)

## Validation Evidence

```
cargo fmt --all -- --check                                                   # clean
cargo clippy -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-channels --all-targets -- -D warnings  # no warnings
cargo test -p zeroclaw-config summary_provider                               # 3 passed
cargo test -p zeroclaw-runtime --lib context_compress                        # 36 passed (incl. new distinct-provider test)
```

New tests:
- `config::schema` — `effective_summary_provider_precedence` (agent → profile → None); `config_validate_rejects_{agent,profile}_summary_provider_missing_alias` (DanglingReference at both levels).
- `context_compressor` — a probe summary provider distinct from a vision-capable agent provider receives the summarization call, `preserve_media_markers=false` strips the `[IMAGE:...]` marker, and the agent provider receives no call.

Beyond CI: the full `cargo test -p zeroclaw-runtime` (heavy integration targets) was disk-constrained locally and left to CI; the `--lib` compressor suite + all touched-crate `--all-targets` clippy pass.

## Security & Privacy Impact

- No new permissions, capabilities, filesystem scope, or external network calls (it's a provider selection for an already-existing summarization call).
- No secrets-handling change; the ref points at an existing configured provider whose credentials are already process-resident.
- No PII in the diff.

## Compatibility

Backward compatible. `summary_provider` is additive (`#[serde(default)]`, empty = inherit), so **no schema-version bump** and existing configs load unchanged; the deprecated `summary_model` fallback preserves their exact behavior. No config/env/CLI surface removed.

## Rollback

`git revert` — additive config field + a compressor-parameter threading; no state or schema migration.


singlerider (changes requested):
Reviewed at bb79b721 on a fresh checkout of `fix/summary-provider-cross-provider`. The design is right and the resolution/fail-open/vision logic all check out, but the integration test target does not compile, so this can't merge as-is.

### 🔴 Blocking — `tests/integration/memory_loop_continuity.rs` still calls the old 4-arg `compress_if_needed`

The signature change adds `summary: SummaryTarget<'_>`, and every in-crate call site was updated — but the top-level integration test was not:

```
error[E0061]: this method takes 5 arguments but 4 arguments were supplied
   --> tests/integration/memory_loop_continuity.rs:259:10
259 |         .compress_if_needed(&mut history, &mock_model_provider, "test-model", None)
```

I reproduced this locally with `cargo build --tests --test integration`; it is the same failure red on CI's Lint, Test, and CI Required Gate jobs. The miss is structural: `cargo test -p zeroclaw-runtime --lib` (which I ran first, all green) does not build the `tests/integration/` target, so the broken call site is invisible to a lib-only run. The fix is a one-line update of that call to pass the fallback `SummaryTarget` (agent provider, agent model, `preserve_media_markers: model.supports_vision()`), matching the `None` arm the production code uses. Please add it and confirm `cargo test` (full, not `--lib`) is green before the next push.

### 🟢 The cross-provider fix itself is correct

Once it builds, the substance is sound:

- `effective_summary_provider` resolves agent override -> profile -> `None`, mirroring `classifier_provider`'s "empty = inherit" idiom. Because `summary_provider` is a `ModelProviderRef` bundling provider + model + auth, summarization runs on the referenced provider regardless of the agent's own provider — the #7964 cross-provider mismatch becomes structurally impossible rather than a silent runtime failure.
- `resolve_summary_provider` fails open on every degenerate case (None, empty, undotted, unknown alias, no `.model`, provider build failure): WARN once and return `None`, caller reuses the agent's own provider with the deprecated `summary_model` swap still honored. Compression can't become a turn-killing hard error.
- `preserve_media_markers` is correct at both arms: `false` when dispatching to a distinct summary provider (never ship `[IMAGE:…]` to a possibly text-only model), and `model_provider.supports_vision()` only on the agent-provider fallback. `compress_if_needed_dispatches_to_distinct_summary_provider_and_strips_markers` pins exactly this.
- Alias rename and delete-scrub both cover `summary_provider` at agent and profile level (`alias_refs.rs` 382-389, 849-862), with config-time `DanglingReference` validation, so the new ref participates in the cascade rather than dangling.

Lib-scoped validation that did pass at bb79b721:

```
cargo test -p zeroclaw-config summary_provider            → 3 passed
cargo test -p zeroclaw-runtime --lib context_compress     → 36 passed
cargo clippy -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-channels --all-targets → clean
```

Requesting changes solely on the build break. The fix is one line and the rest is ready.

WareWolf-MoonWall (commented):
Commenting at 6ba24cfa. singlerider's CHANGES_REQUESTED at bb79b721 is still active, so I'm posting a comment rather than approving. Flagging for singlerider's attention below.

**CI is now green at 6ba24cfa.** The integration test build break singlerider raised — `compress_if_needed` being called with 4 arguments against a 5-argument signature in `tests/integration/memory_loop_continuity.rs:259` — appears to be resolved. Lint, Test, and CI Required Gate are all passing at the current head.

On the substance of the fix, I read it as correct:

- `effective_summary_provider` resolves agent override → profile → `None`, mirroring the `classifier_provider` "empty = inherit" idiom. The cross-provider type mismatch from #7964 becomes structurally impossible rather than just unlikely.
- `resolve_summary_provider` fails open on every degenerate case (None, empty string, unknown alias, provider build failure): emit one `WARN`, fall back to the agent's own provider. Compression is never a turn-killing error.
- `preserve_media_markers` being set to `false` when dispatching to a distinct summary provider is correct — `[IMAGE:…]` markers must not be shipped to a potentially text-only model.
- Alias rename and delete cascades cover `summary_provider` at both agent and profile level, consistent with how other provider fields are handled.

@singlerider — the build break you flagged is fixed. Would you mind taking another look at 6ba24cfa and dismissing your CR if you're satisfied? Happy to approve once the active CR is resolved.

Nillth:
@singlerider Good catch — thanks. Fixed.

The top-level integration test target (`tests/integration/memory_loop_continuity.rs`) was the one call site the in-crate sweep missed, since `cargo test -p zeroclaw-runtime --lib` doesn't build `tests/integration/` (it lives in the root `zeroclawlabs` package, not the runtime crate). `compressor_with_memory_saves_summary` now passes the 5th `SummaryTarget` arg, mirroring the production `None`/fallback arm in `loop_.rs`: summarize on the agent's own provider/model, with `preserve_media_markers` keyed off that provider's `supports_vision()`:

```rust
let result = compressor
    .compress_if_needed(
        &mut history,
        &mock_model_provider,
        "test-model",
        None,
        // Mirror the production fallback arm (loop_.rs `None`): summarize on
        // the agent's own provider/model, preserving media markers only when
        // that provider supports vision.
        SummaryTarget {
            provider: &mock_model_provider,
            model: "test-model",
            preserve_media_markers: mock_model_provider.supports_vision(),
        },
    )
    .await;
```

No production logic changed — purely the missed test call site, as you confirmed the substance is correct.

Validated on the pinned toolchain (`cargo +1.93.0`):

- `cargo +1.93.0 build --tests -p zeroclaw-runtime` — clean
- `cargo +1.93.0 fmt --all` — no changes
- `cargo +1.93.0 clippy -p zeroclaw-runtime --all-targets` — clean
- `cargo +1.93.0 clippy --test integration` — clean

Integration target (the one that was red in CI) now builds and passes:

```
running 159 tests
...
test integration::memory_loop_continuity::compressor_with_memory_saves_summary ... ok

test result: ok. 159 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.39s
```

Full `cargo +1.93.0 test -p zeroclaw-runtime` (lib + integration): 2296 lib tests pass. One unrelated pre-existing flake surfaced under parallel execution — `cron::store::tests::remove_job_emits_structured_cron_delete_event` — which passes deterministically in isolation; it's a shared global tracing-subscriber event bleed between concurrent cron tests, unrelated to this PR. Re-run:

```
test cron::store::tests::remove_job_emits_structured_cron_delete_event ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2298 filtered out
```

Audacity88 (changes requested):
I reviewed current head c6d372fca61bf43b7785a0e876229f5d917710a5 against the live PR state, linked issue #7964, the prior reviews, the author follow-up comment, the current diff/source, and green GitHub checks. I did not run local Cargo; this review is based on remote CI, the author’s validation note, and source inspection.

### ✅ Resolved — integration test call site now passes the new summary target

The earlier build blocker from @singlerider appears resolved at the current head. `tests/integration/memory_loop_continuity.rs` now passes the fifth `SummaryTarget` argument to `compress_if_needed`, and the visible Lint, Test, and CI Required Gate checks are green.

### 🔴 Blocking — legacy `summary_model` configs still keep the cross-provider failure

The new `summary_provider` path is the right shape, but this PR still leaves the original #7964 failure path active for existing configs. Issue #7964 is about a shared runtime profile with `context_compression.summary_model = "provider-specific-model"` being reused by agents on different providers. In that deployed config shape, no new `summary_provider` exists.

At the current head, `effective_summary_provider()` returns `None` when neither the agent nor profile sets the new field. Both runtime fallback paths and the channel proactive-compression path then keep using `context_compression.summary_model` as a bare model id on the agent’s active provider. The PR body also says existing configs run “byte-identically.” That means the exact valid-looking config from #7964 can still send an Anthropic summary model id to an OpenAI agent provider until the operator manually rewrites the config.

Issue #7964's expected behavior asks for a migration or warning path that surfaces previously silent cross-provider mismatches for legacy `summary_model` configs. The PR body explicitly scopes this as no schema bump and no forced migration; operators adopt the new `summary_provider` at their own pace, and existing configs run unchanged through the deprecated fallback. That framing is internally consistent, but it does not provide the active mitigation the issue requests.

If this is intended as an additive first step, then `Closes #7964` is premature: the issue should remain open with a tracked follow-up for legacy migration or warning. If the PR is intended to close #7964 now, it needs some mitigation for legacy `summary_model` configs, such as a schema migration, startup validation/diagnostic that surfaces the unsafe cross-provider case, or another explicit path that prevents the old silent runtime failure from remaining silent.

### 🟢 What looks good — the new provider-reference route is structurally safer

For configs that do opt into `summary_provider`, the implementation lines up with the right model: agent override, then profile value, then fallback to the agent’s own provider. The distinct-provider `SummaryTarget` dispatch also keeps media markers stripped, and the runtime loop plus channel proactive-compression call sites are all wired through the resolved target.

Nillth:
@Audacity88 You're right, and thanks for holding the line on it. The new `summary_provider` path makes the mismatch structurally impossible only for configs that opt in; the exact #7964 shape -- a shared `runtime_profiles.<p>.context_compression.summary_model` reused by agents on different providers, with no `summary_provider` anywhere -- was left silent. `Fixes #7964` didn't hold against that. I've added the active mitigation you asked for: the "startup validation/diagnostic that surfaces the unsafe cross-provider case."

### Primary: a config-time validation diagnostic (offline, deterministic, no schema bump)

Hooked into the existing non-fatal warning surface (`Config::collect_warnings` in `crates/zeroclaw-config/src/schema.rs`), the same mechanism that already emits `dangling_fallback_ref`, `wire_api_not_supported_for_family`, etc. -- surfaced to the CLI via `zeroclaw_log::record!` on `validate()` and to the dashboard via the `warnings` field on the gateway HTTP responses. No new surface, no schema change, no `ValidationWarning` shape change.

New check `cross_provider_summary_model` detects exactly the #7964 shape:

- a `runtime_profiles.<p>.context_compression.summary_model` is set (and the profile has NO `summary_provider`, which would supersede it), AND
- profile `<p>` is referenced by agents that resolve to MORE THAN ONE distinct provider, AND
- those agents have no agent-level `summary_provider` override (an override supersedes the bare id, so that agent is safe and is excluded from the count).

When all three hold, it emits a warning naming the profile, the affected agents, and their differing providers, e.g.:

```
runtime_profiles.shared.context_compression.summary_model ("haiku") is a bare model id
reused by agents resolving to different providers (alpha -> custom.p1, beta -> custom.p2);
the deprecated summary_model is dispatched on each agent's own provider, so it is sent to
the wrong provider for at least one of them and silently fails at runtime (#7964). Migrate
to context_compression.summary_provider (or an agent-level summary_provider), which is
self-contained and runs on its own provider.
```

Each agent's provider is resolved by canonicalizing its `model_provider` ref through `providers.models.find_by_name` to a `<family>.<alias>` label, so equivalent spellings compare equal and "more than one distinct provider" is computed on the resolved entries -- the same notion of "provider" the bug turns on. This is config-derived, so it covers all three dispatch sites (loop recovery, post-turn, and the channel proactive-compression path) without needing a runtime hook on each.

### Why I did NOT auto-route the bare model id

I deliberately stopped short of silently rerouting a bare `summary_model` to the "right" provider. A bare id carries no provider; there's no offline catalog to infer one from; and the only runtime way to discover where a model lives is a `list_models()` network call, which I won't add to the compression hot path. Auto-routing would be a guess that fails differently, not safer. Surfacing the unsafe config and pointing at the self-contained `summary_provider` migration is the honest fix.

### Secondary (belt-and-suspenders): runtime deprecation WARN

At every fallback arm where the deprecated bare `summary_model` is actually dispatched on the agent's own provider, it now WARNs once per agent per process that this can silently mismatch for a shared profile -> migrate to `summary_provider`. This covers all three dispatch sites you flagged: the loop recovery and post-turn arms (`crates/zeroclaw-runtime/src/agent/loop_.rs`) AND the channel proactive-compression path (`resolve_summary_route` in `crates/zeroclaw-channels/src/orchestrator/mod.rs`) -- I made the helper `pub` so the orchestrator shares the same process-global warn-once set, keeping the signal one-per-agent across all three. Deduped via the same `LazyLock<Mutex<HashSet<..>>>` warn-once pattern used by `agent::cost::warn_once_missing_pricing`, so it doesn't spam every turn.

With the config-time diagnostic surfacing the unsafe cross-provider case at startup, the previously-silent #7964 failure is no longer silent, so I've kept `Fixes #7964`. If you'd rather land this as an additive step and track legacy migration separately, I'm happy to switch to a tracked follow-up instead -- your call.

### Tests + validation (pinned toolchain)

New config-validation tests in `crates/zeroclaw-config/src/schema.rs`:

- `collect_warnings_flags_cross_provider_summary_model` -- profile with `summary_model`, two agents on DIFFERENT providers, no override -> the diagnostic fires and names the profile + both agents + providers.
- `collect_warnings_silent_for_same_provider_summary_model` -- same profile, both agents on the SAME provider -> no diagnostic.
- `collect_warnings_silent_when_summary_provider_override_present` -- cross-provider agents that each set an agent-level `summary_provider` -> no diagnostic.

```
cargo +1.93.0 fmt --all                                   # clean
cargo +1.93.0 clippy -p zeroclaw-config --all-targets     # clean
cargo +1.93.0 clippy -p zeroclaw-runtime --all-targets    # clean
cargo +1.93.0 test -p zeroclaw-config summary             # 16 passed; 0 failed
cargo +1.93.0 test -p zeroclaw-runtime --lib loop_::      # 209 passed; 0 failed
```

Audacity88 (approved):
I reviewed current head 2990e0c7594bb1015c54becc6848853d59157be3 against the live PR state, linked issue #7964, the follow-up commits after my prior request-changes review, the current changed source in `crates/zeroclaw-{config,runtime,channels}/` plus `tests/integration/memory_loop_continuity.rs`, and green GitHub checks. I did not run local Cargo; this review is based on source inspection, remote CI, and the author's posted validation. This approval clears my prior legacy-`summary_model` blocker only. Singlerider's earlier build-break concern appears resolved by the integration-test update and green checks, but their formal `CHANGES_REQUESTED` review still remains in force until they clear it, and I am not calling the PR merge-ready.

### ✅ Resolved — legacy `summary_model` no longer stays silent for the issue shape

The follow-up config diagnostic addresses the blocker I raised. `Config::collect_warnings()` now emits `cross_provider_summary_model` when a runtime profile has a deprecated bare `context_compression.summary_model`, no profile-level `summary_provider`, and is shared by agents that resolve to more than one provider without agent-level `summary_provider` overrides. The warning names the profile, affected agents, provider labels, and migration path, so the exact #7964 legacy shape is now surfaced through the existing config warning/startup surface instead of remaining a silent runtime-only failure.

### ✅ Resolved — all three deprecated fallback dispatch sites now warn

The loop recovery and post-turn compression paths call the shared `warn_deprecated_summary_model_fallback()` helper when they fall back to dispatching the deprecated bare model id on the agent provider, and the channel proactive-compression path now shares the same warn-once helper. That covers the three compressor call sites called out in #7964 while preserving the non-breaking fallback behavior. The runtime warning is intentionally deduped once per agent per process, so the durable mitigation is still migrating to `summary_provider`.

### 🟢 What looks good — the new `summary_provider` route is structurally correct

For migrated configs, `effective_summary_provider()` resolves agent override → profile → fallback, the profile and agent refs are validated as `providers.models.<type>.<alias>` references, and distinct summary providers strip media markers. The integration test call site that was previously broken is also updated in `tests/integration/memory_loop_continuity.rs`, and the current visible checks are green.

singlerider (approved):
Re-reviewed head `2990e0c7` (clean merge of `upstream/master`) on a fresh fetch. My single blocker is resolved; converting to approve.

### 🟢 RESOLVED: the integration-test build break is fixed

`tests/integration/memory_loop_continuity.rs:262` now passes the 5th `SummaryTarget` argument, mirroring the production `None` fallback arm (agent provider, agent model, `preserve_media_markers` gated on vision support). The E0061 I reproduced earlier is gone and full CI (Lint, Test, CI Required Gate) is green at this head. That was the only thing holding my review.

### 🟢 Substance still correct at head

Re-confirmed the cross-provider fix holds after the merge: `effective_summary_provider()` resolves agent override -> profile -> fallback; `resolve_summary_provider` fails open on every degenerate case (WARN once, reuse agent provider, honor deprecated `summary_model`); `preserve_media_markers` is `false` for a distinct summary provider and `supports_vision()` only on the agent-provider fallback. The new `cross_provider_summary_model` config warning surfaces the exact #7964 legacy shape at startup, and all three deprecated fallback dispatch sites route through the shared warn-once helper. Alias rename/delete-scrub cover `summary_provider` at agent and profile level with config-time dangling-ref validation.

Clearing my prior CHANGES_REQUESTED. Approving.

PR #7982, opened 2026-06-19
fix(channels/voice): stop caching config-derived static_voice_peers on channel handle

### LINKED ISSUE #7795: static_voice_peers caches config-derived voice peers on the channel handle (latent SSOT violation)

perlowja:
## Summary

Automated build resolving the linked issue. Diff touches 3 file(s):
- `crates/zeroclaw-channels/src/telegram.rs`
- `crates/zeroclaw-config/src/schema.rs`

## Linked issue

Resolves zeroclaw-labs/zeroclaw#7795

`risk: medium` · `size: S`

## Testing (required)

> VERDICT: NEEDS-ATTENTION
> FINDINGS:
> - crates/zeroclaw-channels/src/orchestrator/mod.rs:5748 HIGH: The new `voice_peer_resolver` variable is created but never used when constructing `ConfiguredChannel`. If `ConfiguredChannel` does not have a field for this resolver, the variable is dead code and may indicate an incomplete integration, leading to runtime failures where voice peers are never resolved.
> - crates/zeroclaw-channels/src/orchestrator/mod.rs:5729 HIGH: The call to `TelegramChannel::new` now includes `voice_peer_resolver`, but the signature change is not propagated to any other channel implementations (e.g., other channel types that also call `new`). This will cause compilation errors or mismatched APIs across the codebase.
> - crates/zeroclaw-channels/src/telegram.rs:623 HIGH: The struct field `voice_peer_resolver` is stored as `Arc<dyn Fn() -> Vec<String> + Send + Sync>`, but the closure captures `config.clone()` in tests (non‑thread‑safe `Config`). In production the closure captures `Arc<RwLock<Config>>`, which is safe, but the test closure could cause data races if the test were ever run in a multithreaded context.
> - crates/zeroclaw-channels/src/telegram.rs:810 MED: The `is_

## Compatibility (required)

- Backward compatible? Yes

## Rollback

Rollback path: `git revert <sha>` reverts this change.


singlerider (approved):
First review (no prior reviews, comments, or inline threads). Fetched the head locally and read all three files plus the `channel_external_peers` precedent it mirrors. CI is 18/18 green.

Resolves #7795: `static_voice_peers` was a `HashSet` cached on the `TelegramChannel` handle, seeded once at construction and never refreshed — the exact `allowed_users`-on-handle drift shape AGENTS.md's SSOT rule forbids. An operator editing `[peer_groups.*]` voice membership and reloading config would not see the change until daemon restart.

### 🟢 What looks good — this is the sanctioned resolver-closure fix, not a cache

The cache is replaced with `voice_peer_resolver: Arc<dyn Fn() -> Vec<String> + Send + Sync>` (telegram.rs:546) closing over `Arc<RwLock<Config>>` at both construction sites (orchestrator/mod.rs:5732 and :6743), resolving live through the new canonical `Config::channel_voice_peers()`. That is precisely the "resolver closure that closes over `Arc<RwLock<Config>>` and resolves on call" pattern AGENTS.md lists as allowed, and the doc comment cites the rule. `is_voice_chat` (telegram.rs:1034) now calls `(self.voice_peer_resolver)()` live, OR'd with the session `voice_chats` set, so config edits take effect on the next message with no restart.

### 🟢 What looks good — `channel_voice_peers` mirrors the established precedent exactly

The new `Config::channel_voice_peers` (schema.rs:15839) is a structural copy of `channel_external_peers` (:15818) — same `split_once('.')` type-wide-vs-dotted matching, same `seen`-set dedup — with one added filter: `output_modality == OutputModality::Voice`. Single source of truth is `self.peer_groups`; no shadow state. Matching semantics are identical to the old `with_voice_peer_prefs` (a group with `channel = "telegram"` matches type-wide; `"telegram.default"` matches only the exact alias), so this is behavior-preserving for resolution while fixing the staleness.

### 🟢 What looks good — the adversarial findings are correctly marked advisory; all three HIGH are false

Checked each against the source:

- "voice_peer_resolver never used / dead code": false. Consumed at telegram.rs:1034 in `is_voice_chat`.
- "`TelegramChannel::new` signature change not propagated": false. `new()` is unchanged; the resolver lands via a separate `.with_voice_peer_resolver()` builder. The other `new()` callers (orchestrator/mod.rs:9358 and the test constructors) compile against the default `Arc::new(Vec::new)` empty resolver set in the constructor.
- "test closure captures non-thread-safe `Config`": false. The test moves an owned `config.clone()` into the `Send + Sync` closure; `Config` is `Send + Sync` (it lives behind `Arc<RwLock<Config>>` in production), so the bound is satisfied and there is no data-race path.

### 🟢 What looks good — tests rewritten to assert the new contract

`voice_peer_resolver_resolves_live_from_config` now asserts through `is_voice_chat` (live resolution) instead of poking the removed `static_voice_peers` set, and still pins channel/modality filtering (@carol on another channel and @dave on mirror modality both excluded) plus session-set isolation (live peers do not pollute `voice_chats`). `voice_peer_resolver_survives_session_voice_chats_removal` keeps the separation guarantee. The removed `with_voice_peer_prefs` and `static_voice_peers` leave no dangling references anywhere in the tree.

Applied labels (`bug` / `size: S` / `risk: medium` / `channel` / `config`) are correct; the PR body's "risk: low" line is stale relative to them but the labels are the right call. Approving.

Audacity88 (approved):
Reviewed current head `56a69283d12c716d7238697f00c323da767498d3` against the live PR state, linked issue #7795, singlerider's approval at this head, the PR body's adversarial-review findings, current checks, the full diff, and the relevant source around Telegram channel construction, `Config::channel_external_peers`, `Config::channel_voice_peers`, and the Telegram voice-reply tests. I did not run local Cargo validation or a live Telegram smoke.

### 🟢 What looks good — voice peers now resolve from the config source of truth

This removes the cached `static_voice_peers` handle state that #7795 called out and replaces it with a resolver closure. The production construction paths in `build_channel_by_id` and `collect_configured_channels` close over the shared `Arc<RwLock<Config>>` and call `channel_voice_peers("telegram", alias)` at use time, so the channel no longer owns a second copy of the peer-group voice membership.

`Config::channel_voice_peers` also mirrors the existing `channel_external_peers` matching model: type-wide groups and dotted aliases both work, duplicates are deduped, and only `output_modality = "voice"` groups contribute. That keeps the behavior aligned with the existing peer-group contract while fixing the stale-cache shape.

### 🟢 What looks good — the Telegram path uses the resolver at the behavior boundary

`TelegramChannel::is_voice_chat` now combines the session `voice_chats` set with the live resolver result, so configured voice peers still get TTS replies without being copied into the session set. The updated tests cover the important local contract: voice groups match, other channels and mirror-modality groups do not match, live-resolved peers do not pollute `voice_chats`, and removing a recipient from the session set does not disable config-derived voice delivery.

The other production `TelegramChannel::new` call outside normal channel registration is the direct one-shot send path, which does not exercise the voice-reply/TTS decision path this PR changes. I did not find a remaining `static_voice_peers` or `with_voice_peer_prefs` reference.

Approving from my side. The current `risk: medium` / `channel` / `config` labels fit this break-fix scope.

PR #7983, opened 2026-06-19
fix(daemon): handle file-descriptor exhaustion (EMFILE) in IPC accept loop

### LINKED ISSUE #7042: [Bug]: daemon IPC crash on file-descriptor exhaustion (EMFILE / too many open files)

perlowja:
## Summary

Automated build resolving the linked issue. Diff touches 2 file(s):
- `crates/zeroclaw-runtime/src/rpc/local.rs`
- `crates/zeroclaw-runtime/src/rpc/wss.rs`

## Linked issue

Resolves zeroclaw-labs/zeroclaw#7042

`bug` · `runtime` · `risk: high` · `size: S`

> STRICT: unreviewed (skipped/error) treated as block; ngc-review infra error (fail-open): [Errno 8] Exec format error: '/usr/local/bin/ngc-review'

## Compatibility (required)

- Backward compatible? Yes

## Rollback

High-risk PR: revert the eventual squash commit. No config or data migration is involved. Observable regression symptoms would be local IPC or WSS accept loops returning to WARN hot-spin behavior under `EMFILE` / `ENFILE`, or fatal accept errors failing to surface through the daemon supervisor restart path.


singlerider (approved):
First review (no prior reviews, comments, or inline threads). Fetched the head locally and traced both listeners through `platform::accept`, the daemon component registry, and `spawn_component_supervisor`. CI is 18/18 green.

Resolves #7042: the daemon wedged/crashed on `EMFILE`/`ENFILE` during IPC `accept()`. The old loop logged a WARN and `continue`d on every accept error, which had two failure modes under fd pressure — and a latent one on a permanently-broken listener.

### 🟢 What looks good — recovery vs propagation is split correctly, and the supervisor envelope makes the new propagation safe

This is the part I scrutinized hardest, because the diff changes accept-error behavior in a high-risk runtime path. The old code `continue`d on *all* accept errors; the new code backs off 50ms and continues only for recoverable errors (`EMFILE`/`ENFILE`/`ConnectionAborted`/`Interrupted`/`WouldBlock`) and propagates everything else via `return Err(e)`. I confirmed both listeners run under `spawn_component_supervisor` (daemon/mod.rs:409/431), which on `Err` marks the component errored, records an ERROR with health signal, applies exponential backoff, and restarts the listener with a fresh fd. So the net effect is strictly better than the old code on every axis: transient fd exhaustion recovers in-loop without a hot-spin; a genuinely fatal accept error (broken listener fd) now restarts cleanly through the supervisor instead of hot-spinning WARN forever with no health signal and no recovery. The "Backward compatible? Yes" line in the body is imprecise — fatal accept errors now terminate-and-restart rather than loop forever — but the supervised restart is the correct behavior, not a regression.

### 🟢 What looks good — the local/wss error-type asymmetry is handled correctly

The two call sites look different for a real reason and both are right. `wss.rs` calls `TcpListener::accept()` which yields a plain `std::io::Error`, so `is_recoverable_accept_error(&e)` runs directly. `local.rs` goes through `platform::accept`, which wraps the error with `.context("accepting local connection")` and returns `anyhow::Error`, so it uses `e.downcast_ref::<std::io::Error>()`. I verified this actually recovers EMFILE: anyhow's `downcast_ref` walks the source chain, and the underlying `io::Error` is the source beneath the context string, so an EMFILE is found through the `.context()` wrap. Had the local path used `is_recoverable_accept_error(&e)` directly it would not have compiled against `anyhow::Error`; the downcast is the correct shape here.

### 🟢 What looks good — errno constants, backoff, and test coverage

`EMFILE = 24` / `ENFILE = 23` are correct and stable across the supported Unix targets (Linux/macOS/BSD) and `#[cfg(unix)]`-gated so Windows is unaffected. The 50ms `ACCEPT_ERROR_BACKOFF_MS` is the right order of magnitude to break a hot-spin without adding meaningful accept latency. The tests pin both directions of the contract: `EMFILE`/`ENFILE` and the transient kinds are recoverable, and a non-transient kind (`InvalidInput`) is *not* swallowed so it propagates to the supervisor.

One small note, non-blocking: `EMFILE`/`ENFILE`/`ACCEPT_ERROR_BACKOFF_MS`/`is_recoverable_accept_error` are defined identically in both `local.rs` and `wss.rs`. These are trivially-stable libc errno values and pulling in `libc` for two constants is not worth it, so I would not hold merge on it — but if a third listener ever needs the same predicate, lift it into a shared `rpc` helper module rather than copying a third time.

Applied labels (`bug` / `size: S` / `risk: high` / `runtime`) are correct; the body's "risk: low" line is stale relative to them. Approving.

Audacity88 (approved):
This is a confirmatory approval to clear my requested-review slot. I checked the live PR at b9eec7bec0365a66a69ec07960405b47b3aa26da, the linked #7042 issue, @singlerider's approval, the current two-file diff in `local.rs` / `wss.rs`, and the daemon supervisor path on current `master`. I did not run local Cargo for this review-only pass; this is based on source inspection plus the green GitHub Quality Gate. The PR body still has a stale `risk: low` shorthand while the live labels correctly say `risk: high`, so that public-record cleanup should happen before final merge text/body readback, but I do not see it as author-blocking.

### 🟢 What looks good — accept-loop recovery is split at the right boundary

The fix handles the #7042 accept-loop failure mode without hiding genuinely broken listeners. `EMFILE` / `ENFILE` and the usual transient accept kinds now log, back off briefly, and continue on the existing listener, which prevents the old WARN hot-spin under fd pressure. Non-transient accept errors now return `Err`, and both RPC listeners are still run through `spawn_component_supervisor`, so those failures get health/error reporting plus the existing restart/backoff path instead of looping forever inside the listener.

### 🟢 What looks good — local IPC and WSS preserve their different error shapes

The local IPC side correctly accounts for `platform::accept()` returning an `anyhow::Error` with context, while the WSS side receives a plain `std::io::Error` from `TcpListener::accept()`. Keeping the local path as a downcast check and the WSS path as a direct predicate keeps the fix small and avoids changing the transport APIs just for this recovery slice.

### 🟢 What looks good — the regression coverage pins both sides of the predicate

The new tests cover the important contract: Unix fd-exhaustion errno values are recoverable, ordinary transient kinds are recoverable, and a non-transient error is not swallowed. That is the right behavioral seam for this PR. It does not prove the broader fd-leak/reap investigation in #7042 by itself, but it does prove this accept-loop crash/wedge slice.

### 🔵 Suggestion — avoid a third copy if another listener needs this

I agree with @singlerider's non-blocking note on the duplicated predicate. The duplicate `EMFILE` / `ENFILE` / backoff predicate in `local.rs` and `wss.rs` is fine at this size. If another RPC listener needs the same behavior, I would move it into a shared RPC helper before copying it again so the recovery contract stays in one place.

Issue #7874, opened 2026-06-17
[Tracker]: image generation tool output should expose stable attachment paths


Audacity88:
## Summary

The image generation tool should return stable generated-file paths that model and channel paths can use to attach the created image.

## Problem

PR #6555 bundled two independent pieces of image-generation work. This issue tracks only the smaller filename/path-output slice.

The PR discussion described two related user-facing problems:

- generated images could reuse a default filename such as `generated_image.png`, causing clobbering or confusing later context
- some model/channel flows did not reliably use the generated image path for attachment, so the model could hallucinate a path instead of attaching the actual generated file

The maintainer routing in #6555 called for splitting this into a small PR focused on generated filename/path behavior and the tool output needed for model/channel attachment. The bundled branch closed as stale before that split landed.

## Expected behavior

Each generated image should have a stable, non-clobbering path or filename in tool output. Downstream model/channel paths should have enough structured information to attach the actual generated image without guessing a path.

## Scope

In scope:

- generated image filename/path behavior
- image generation tool output contract
- channel attachment path validation for at least one exercised channel path
- regression or manual validation that the returned path can be used to attach the generated image

Out of scope:

- adding the RunPod/ComfyUI provider
- provider-scoped image-generation config
- broad image pipeline redesign

## Reproduction sketch

Use image generation in a channel path that should attach the generated image back to the conversation. Generate more than one image and inspect whether paths are unique, stable, and exposed to the model/channel in a way that avoids guessed or stale filenames.

This needs a fresh reproduction or regression test against current `master` before closing the issue.

## Related

- #6555

## Acceptance criteria

- Generated image filenames do not clobber prior generated files.
- Tool output exposes the generated image path in a stable form usable by channel attachment code.
- At least one channel path is validated end to end or covered by focused tests.
- The issue remains independent from provider additions such as RunPod/ComfyUI.

## Triage fields

- Affected component: tools, provider, channel
- Severity: S2 - degraded behavior
- ZeroClaw version: observed in PR #6555; current `master` needs fresh verification
- Operating system: platform-independent
- Regression status: unknown
- Pre-flight checks: public text is redacted; this is a closure-recovery tracker and still needs fresh reproduction on current `master`

Audacity88:
## Status Update

Accepted as the narrow image-generation attachment-path slice. The goal is for generated-image tool output to expose a stable file path that downstream model/channel code can attach without relying on a reused default filename.

Next slice:

[...]

PR #7985, opened 2026-06-19
fix(tools/image): expose stable attachment paths in image-generation output

perlowja:
## Summary

Automated build advancing the linked tracker. Diff touches 1 file(s):
- `crates/zeroclaw-tools/src/image_gen.rs`

## Linked issue

Part of zeroclaw-labs/zeroclaw#7874

`enhancement` · `size: XS` · `risk: high` · `tool`

## Testing (required)

Initial automated cross-family review returned **needs-attention** on filename collision/test-coverage concerns. Follow-up commit `1334f89fac573afe99ab5e978334726c29afb83d` extracted production helpers and added regressions for non-clobbering default filenames plus matching `File:` / `[IMAGE:...]` output paths. The author reported advisory approval after that fix, and the maintainer re-review approved the resolved head.

## Compatibility (required)

- Backward compatible? Yes

## Rollback

`git revert <merge-sha>` reverts this isolated image-generation tool output change.


singlerider (approved):
First review (no prior reviews, comments, or inline threads). Fetched the head locally and traced the filename sanitization into the `output_path` write and the dual-marker output through `canonicalize_tool_result_media_markers`. CI is 18/18 green.

Part of tracker #7874: image-gen output should expose stable attachment paths, and successive default-named calls should stop clobbering each other.

### 🟢 What looks good — traversal sanitization is sound and the new default fixes the real clobbering bug

I checked the actual threat rather than the refactor shape. `output_path = workspace_dir.join("images").join(format!("{safe_name}.png"))` (image_gen.rs:254), and `safe_name` is run through `PathBuf::file_name()`, which strips every path component. A traversal input like `"../../etc/passwd"` resolves to `file_name() == Some("passwd")`, landing at `workspace/images/passwd.png` — escape blocked. The new `unwrap_or_else` only fires when the caller supplies *no* filename, generating `generated_image_{unix_nanos}` so back-to-back default-named generations no longer overwrite one shared `generated_image.png`, which is exactly the clobbering #7874 calls out.

### 🟢 What looks good — the dual `File:` / `[IMAGE:]` output dedups correctly

The comment claims the bare `File:` path and the explicit `[IMAGE:…]` marker carry the same string so the promoter does not double-count, and that holds: `canonicalize_tool_result_media_markers` (history.rs:187) collects existing marker payloads first and skips promoting any bare path already present in `existing_markers` (the `existing_markers.contains(path)` guard at :206). So emitting both a durable `File:` line (survives marker-stripping in older turns) and the inline `[IMAGE:path]` marker yields one attachment, not two. The comment earns its place by explaining that non-obvious dedup contract.

### 🟢 What looks good — the adversarial BLOCK is correctly marked advisory; the findings don't hold

- HIGH "`file_name()` returns `None` for `"foo/.."` so it collides on `generated_image`": this only triggers when a caller passes a pathologically degenerate filename with no final component, and the result is a deterministic in-workspace name — no traversal, no security impact, just a cosmetic collision on an adversarial input. Not worth blocking; the traversal guarantee is intact.
- MED "`ts` computed before the closure runs, wasting a syscall": false. `ts` is already computed *inside* the `unwrap_or_else` closure, so it only runs on the no-filename path — the code already does what the finding recommends.
- MED "SystemTime fails on WASM → epoch 0 collisions": this tool makes a live HTTP call to fal.ai and does not run on WASM; `unwrap_or_default()` would only yield epoch if the clock predates 1970. Non-issue.

The test pins the default-name contract (prefix + numeric suffix). Applied labels (`enhancement` / `size: XS` / `risk: high` / `tool`) are correct for a tool fs-write path; the body's "risk: low" line is stale relative to them. Approving.

Audacity88 (dismissed):
@perlowja Context: I reviewed the current head 95332b77db22cf43922cf170c7cf6be78b52e97a against #7874, the PR diff, singlerider's approval, the current marker-canonicalization/channel marker code, and the green GitHub checks. I did not run local Cargo validation for this review. The implementation direction looks right, and singlerider's review already traced the production path manually. I am requesting changes because the automated regression still does not exercise the production behavior this PR needs to preserve before closing the tracker.

### 🟢 What looks good — The output contract moves in the right direction

The tool now emits the saved path in both a durable `File:` line and an explicit `[IMAGE:...]` marker. That matches the current downstream contract: the runtime canonicalizer already dedups a bare path when the same path is present in an explicit marker, and the channel marker parsers recognize `[IMAGE:<path>]` attachments. The filename sanitization also still strips path components before joining under `workspace/images`.

### 🔴 Blocking — Add a regression that exercises the actual image_gen output path

singlerider's approval already traced the production path manually: the filename sanitization feeds `output_path`, and the dual `File:` / `[IMAGE:...]` output dedups correctly through the runtime marker canonicalizer. I agree with that code read.

The remaining gap is automated coverage. The new `default_filename_includes_nanoseconds` test only builds a timestamped string inside the test. It does not call the production filename selection logic, does not prove that an omitted `filename` argument writes to `generated_image_<timestamp>.png`, and does not prove that the returned tool output contains the same concrete path in both the `File:` line and the `[IMAGE:...]` marker. That means the test would still pass if production went back to always using `generated_image.png` or stopped emitting the marker.

#7874's acceptance criteria are specifically about non-clobbering generated filenames and stable attachment paths. Please add a test that verifies the actual production behavior. The simplest path is to extract the filename selection/output formatting into a small helper and test the no-filename case through that helper. A fuller path is to mock the generation/download boundary, execute `image_gen` with no filename argument, and assert that two default-named calls produce distinct output paths plus one matching `[IMAGE:<same path>]` marker per result. The important property is that the test fails if the code reverts to the old clobbering behavior or stops returning a matching attachment marker.

perlowja:
Thanks @Audacity88 — you're right that the old test only built a string and never exercised production. Replaced it with regressions that drive the actual code paths.

Extracted two pure helpers that the production path now calls:
- `resolve_image_filename(filename_arg, nanos)` — the filename selection (caller's name, path-stripped; or `generated_image_<nanos>` when omitted). `nanos` is injected so the no-filename case is deterministically testable.
- `format_image_tool_output(...)` — the dual `File:` line + `[IMAGE:<path>]` marker.

New tests, both through the production helpers:
- `resolve_image_filename_default_is_non_clobbering_and_unique` — no filename yields a timestamped name (never the bare `generated_image`), two default calls differ, an explicit name is used verbatim (traversal-stripped), and blank falls back to the default. **Fails if the code reverts to a fixed clobbering name.**
- `image_output_emits_matching_file_line_and_image_marker` — the formatted output carries the same concrete path in **both** the `File:` line and the `[IMAGE:<path>]` marker. **Fails if the marker (or the matching path) is dropped.**

That's the property you asked for: the regression breaks if generation goes back to clobbering or stops returning a matching attachment marker. Behavior is byte-identical (refactor only).

Validation: full gate at 1.93.0 (clippy `--features ci-all`, `check --locked` ci-all + no-default, both arch guards, nextest 9427 pass). Adversarial review (cross-family) — APPROVE.

Audacity88 (approved):
@perlowja Context: I re-reviewed current head 1334f89fac573afe99ab5e978334726c29afb83d against #7874, my prior changes-requested review, your follow-up comment, @singlerider's approval, the current diff, the linked issue and overlapping PR state, the marker-canonicalization path on current `master`, and the green GitHub checks. I did not run local Cargo, live image generation, or a channel smoke test for this review. My prior blocker is resolved, and I am approving the code and test changes.

### ✅ Resolved — The regression now exercises the production helpers

My prior blocker was that the old test built a timestamped string inside the test without exercising the production filename or output path. The new head wires `resolve_image_filename()` into `ImageGenTool::generate()` and wires `format_image_tool_output()` into the actual `ToolResult` output boundary. The new tests now assert the no-filename path returns `generated_image_<nanos>`, successive default names differ, explicit filenames are path-stripped, blank input falls back to the timestamped default, and the returned text carries the same concrete path in both the `File:` line and `[IMAGE:<path>]` marker. That is the regression shape I asked for: it would fail if the code went back to the old clobbering `generated_image.png` behavior or stopped returning the matching attachment marker.

### 🔵 Suggestion — Refresh the filename schema default text

One model-facing description is now stale: the `filename` schema still says the default is `'generated_image'`, while the implementation now uses `generated_image_<nanos>` when the caller omits a filename. This does not block the fix because the actual omitted-filename behavior and output contract are correct and covered, but it would be good to refresh that text so agents do not learn the old fixed-name mental model from the tool schema.

### 🟢 What looks good — The dual output contract still matches the runtime marker path

The `File:` line plus `[IMAGE:<same path>]` marker remains the right shape for the current runtime. `canonicalize_tool_result_media_markers()` collects existing marker payloads and skips promoting a bare path when the same path is already inside an explicit marker, so the tool can keep the durable human-readable path while still giving the multimodal pipeline an attachment marker without double-counting the image.

PR #7986, opened 2026-06-19
fix(channels/feishu): route inbound messages to agent instead of LLM-only

### LINKED ISSUE #4873: [Bug]: After integrating Feishu, only the LLM is called by default instead of the Agent.
### Affected component

channel

### Severity

S3 - minor issue

### Current behavior

After integrating Feishu, only the LLM is called by default instead of the Agent.

### Expected behavior

Ask questions via Feishu and get responses from the Agent.

### Steps to reproduce

```bash
After integrating Feishu via initial configuration, normal question-and-answer interactions work in Feishu, yet the system cannot autonomously call tools.
```

### Impact

_No response_

### Logs / stack traces

```text

```

### ZeroClaw version

0.5.9

### Rust version

_No response_

### Operating system

debian 13

### Regression?

Unknown

### Pre-flight checks

perlowja:
## Summary

## Linked issue

Resolves zeroclaw-labs/zeroclaw#4873

`risk: low` · `size: S`

## Testing (required)

## Compatibility (required)

- Backward compatible? Yes

## Rollback

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -18439,6 +18439,142 @@ This is an example JSON object for profile settings."#;
         );
     }
 
+    #[cfg(feature = "channel-lark")]
+    #[test]
+    fn collect_configured_channels_feishu_binding_reaches_tool_capable_agent_route() {
+        use zeroclaw_config::multi_agent::{AgentAlias, PeerGroupConfig, PeerUsername};
+
+        let mut config = Config::default();
+        config.channels.lark.insert(
+            "work".to_string(),
+            zeroclaw_config::schema::LarkConfig {
+                enabled: true,
+                app_id: "cli_feishu_app123".to_string(),
+                app_secret: "test-secret".to_string(),
+                use_feishu: true,
+                ack_reactions: Some(false),
+                ..Default::default()
+            },
+        );
+        config.agents.clear();
+        config.agents.insert(
+            "feishu-agent".to_string(),
+            zeroclaw_config::schema::AliasedAgentConfig {
+                enabled: true,
+                channels: vec!["feishu.work".into()],
+                ..Default::default()
+            },
+        );
+        config.peer_groups.insert(
+            "feishu_ops".to_string(),
+            PeerGroupConfig {
+                channel: "feishu.work".into(),
+                agents: vec![AgentAlias::new("feishu-agent")],
+                external_peers: vec![PeerUsername::new("@operator")],
+                ..Default::default()
+            },
+        );
+
+        let config_arc = Arc::new(RwLock::new(config.clone()));
+        let configured_channels = collect_configured_channels(&config_arc, "test", &[]);
+
+        let feishu = configured_channels
+            .iter()
+            .find(|entry| entry.alias.as_deref() == Some("work"))
+            .expect("feishu.work binding must be collected from [channels.lark.work]");
+        assert_eq!(feishu.display_name, "Feishu");
+        assert_eq!(feishu.channel.name(), "feishu");
+
+        let mut channels_by_name: HashMap<String, Arc<dyn Channel>> = HashMap::new();
+        let collected_channel_keys: Vec<String> = configured_channels
+            .iter()
+            .map(|entry| {
+                let key = composite_channel_key(entry.channel.name(), entry.alias.as_deref());
+                channels_by_name.insert(key.clone(), Arc::clone(&entry.channel));
+                key
+            })
+            .collect();
+        assert!(
+            collected_channel_keys
+                .iter()
+                .any(|key| key == "feishu.work"),
+            "startup collection must expose the Feishu alias key, got {collected_channel_keys:?}"
+        );
+
+        let enabled_agents = vec!["feishu-agent".to_string()];
+        let owners = build_owner_by_channel_key(&config, &enabled_agents, &collected_channel_keys);
+        assert_eq!(
+            owners.get("feishu.work").map(String::as_str),
+            Some("feishu-agent"),
+            "a feishu.<alias> agent binding must own the collected Feishu channel key"
+        );
+
+        let msg = channel_message("feishu", Some("work"));
+        assert!(
+            find_channel_for_message(&channels_by_name, &msg).is_some(),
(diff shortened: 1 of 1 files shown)


Audacity88 (dismissed):
Reviewed the #7986 public diff, linked issue #4873, current PR state/checks, and public PR metadata. I did not run local validation for this review.

### 🟢 What looks good — the active alias fix is in the right place

The `collect_configured_channels()` change targets the likely runtime gap: an enabled agent can bind `feishu.<alias>` while the concrete config still lives under `[channels.lark.<alias>]`. Checking both `lark.<alias>` and `feishu.<alias>` keeps `[channels.lark]` as the single source of truth while allowing the Feishu-facing alias to start the shared Lark/Feishu channel.

### 🔴 Blocking — the diff still lacks the Feishu agent-path regression

Issue #4873 is not just “accept the `feishu` spelling.” The current maintainer status on that issue asks for a focused Lark/Feishu regression showing that an inbound Feishu chat message reaches the full agent/tool-call path. This PR changes the alias filter and the one-shot channel builder, but it does not add a test or near-live evidence that a `feishu.<alias>` agent binding is collected, started, and routed through the normal agent path instead of closing #4873 on compile-only CI.

Please add focused coverage for the actual issue boundary before using `Resolves #4873`: at minimum, prove that a Feishu-bound agent channel is included by `collect_configured_channels()` and that the resulting message path reaches the normal agent/tool-capable processing surface. If direct live Feishu evidence is intentionally out of scope, the PR body should say that clearly and the automated regression should carry the closure proof.

perlowja:
Thanks @Audacity88 — you were right that compile-only CI didn't prove the issue. Reworked the regression so it proves the actual #4873 behavior without a tautology.

What the test now pins (all via production functions, on `de2c50aab`):
- `collect_configured_channels` collects `feishu.work` from `[channels.lark.work]` when `use_feishu = true` (the binding that was being dropped → LLM-only fallback).
- `build_owner_by_channel_key` assigns `feishu.work` → `feishu-agent` — i.e. the inbound Feishu channel is **owned by the agent**, which is exactly what routes it to the agent path instead of the LLM-only default.
- `AgentRouter::multi(...).resolve(inbound_feishu_msg)` resolves to that owning agent context, and `peer_prompt_channel_ref(resolved) == Some("feishu.work")` confirms the resolved context is scoped to the Feishu channel.

I **removed** the earlier `send_message_to_peer_tool_available(resolved, …)` assertion — you'd have been right to flag it as circular: the test itself supplied that tool via `tools_registry`, so asserting its presence proved nothing about the production path. Left a comment in the test saying so. The remaining assertions fail on the pre-fix code at the collection/ownership stage (where the bug actually lives), so the regression is meaningful.

Validation: full gate at pinned 1.93.0 (clippy `--features ci-all`, `check --locked` ci-all + no-default, both arch guards, nextest) — green. Adversarial review (cross-family) — APPROVE.

Kept `Resolves #4873` since the collected-and-agent-owned routing is the fix for "only the LLM is called instead of the Agent." If you'd prefer this scoped to `Related` pending a live Feishu smoke, happy to narrow it.

Audacity88 (dismissed):
Thanks @perlowja for the update. I reviewed head `de2c50a` against my prior dismissed review, your follow-up comment, the linked #4873 context, the current diff and adjacent channel/source paths, and the green GitHub checks. I did not run local validation or a live Feishu smoke for this re-review.

### ✅ Resolved — the collection and router-ownership proof is now meaningful

The new regression no longer only proves that the `feishu` spelling is accepted. It now sets up `[channels.lark.work]` with `use_feishu = true`, binds `feishu.work` to `feishu-agent`, proves `collect_configured_channels()` exposes a `feishu.work` channel key, proves `build_owner_by_channel_key()` assigns that key to the owning agent, and proves `AgentRouter::multi()` resolves a synthetic Feishu message to the owning agent context with the expected `feishu.work` channel ref. That addresses the router/owner-map part of the full agent-path proof from my prior review.

### 🔴 Blocking — Feishu ingress still reads the Lark peer group before routing

The remaining problem is one step earlier than the new regression. `collect_configured_channels()` still builds the Lark/Feishu handle with a peer resolver that calls `channel_external_peers("lark", &alias)`. `LarkChannel` calls that resolver from `is_user_allowed()` before it emits a `ChannelMessage` to the orchestrator. A normal Feishu peer group such as `channel = "feishu.work"` does not match that `lark.work` lookup, so the Feishu event is still ignored at the channel edge as unauthorized before the router path that the new test constructs by hand can run.

That means the test configuration in this PR proves the later routing layer, but it still bypasses the live ingress gate that #4873 depends on. Please derive the peer resolver from the effective platform name (`feishu` when `use_feishu = true`, otherwise `lark`) and add coverage that would fail if a configured `feishu.<alias>` peer cannot pass the channel authorization/parse path before routing. If the new `build_channel_by_id("feishu")` support stays in this PR, please check that path for the same effective-platform mapping so the one-shot send helper does not drift from the startup Feishu/Lark semantics.

perlowja:
Fixed in b8a2090b9. @Audacity88 you were right — the new regression proved the routing layer but bypassed the live ingress gate. The peer resolver was hardcoded to `channel_external_peers("lark", …)`, so `LarkChannel::is_user_allowed()` rejected a `feishu.<alias>` peer group at the channel edge before emitting.

Now `lark_effective_channel_type` derives the effective platform (`feishu` when `use_feishu = true`, else `lark`) and peer resolution routes through it in **collect_configured_channels**, **build_channel_by_id**, and the **deliver_announcement** fallback, so the one-shot/send paths don't drift from startup semantics. New test `collect_configured_channels_feishu_peer_group_passes_ingress_gate_before_routing` feeds a Feishu event through the channel-edge authorize/parse path and asserts it's admitted **before** routing — it fails on the old hardcoded `lark` lookup (0 vs 1 authorized). clippy `--features ci-all` + `check --locked` green; 6/6 feishu tests pass with `--features channel-lark`. Re-requesting.

perlowja:
Closing to clear the upstream branch list: this branch was pushed to the canonical repo by mistake (inverted `origin`/fork remotes). The change is preserved on my fork as `perlowja:fix/feishu-route-inbound-to-agent` and will be resubmitted as a proper fork-based PR. A fork-only pre-push guard is now in place so this cannot recur. Apologies for the upstream-branch pollution.

PR #7990, opened 2026-06-19
test(providers): regression coverage for per-agent runtime-option resolution (#7870)

### LINKED ISSUE #7870: [Tracker]: agent runtime options can leak from first configured provider
## Summary

Agent and tool provider construction should resolve runtime options for the agent's selected provider, not for the first provider found in config file order.

## Problem

PR #6887 described a concrete provider-runtime bug: `provider_runtime_options_from_config()` can return options for the first configured model provider rather than the provider selected by the current agent. In a multi-provider config, that can send options such as a `max_tokens` value from one provider to a different provider with a lower limit.

The reported failure shape was:

```text
max_tokens range is [1, 65536]
```

The stale PR proposed resolving runtime options by agent alias at the call sites that construct providers for agent execution and tool/delegate paths. That branch closed as stale before the fix landed or a durable issue was opened.

## Expected behavior

When an agent or agent-owned tool path constructs a model provider, runtime options should come from the provider selected for that agent.

For example, if one configured provider supports a larger `max_tokens` value and another agent uses a lower-limit provider, the lower-limit provider should not receive the larger provider's runtime options.

## Scope

This issue tracks the narrow runtime-options lookup bug from #6887.

In scope:

- agent execution provider construction
- LLM task tool provider construction
- delegate tool provider construction
- regression coverage for multi-provider config order

Out of scope:

- gateway boot-time seed logging paths that intentionally use the first configured provider
- broader provider-family refactors
- channel hot-reload behavior unless a fresh investigation shows it is the same root cause

## Reproduction sketch

Use a config with at least two model providers where the first provider has runtime options that are invalid for the second provider. Configure an agent to use the second provider, then run an agent turn through a path that constructs the provider with runtime options.

The failure is that the second provider receives runtime options from the first provider.

This needs a fresh reproduction or regression test against current `master` before closing the issue.

## Related

- #6887

## Acceptance criteria

- Runtime option lookup resolves through the current agent's selected provider, not config insertion order.
- Regression coverage proves multi-provider configs do not leak runtime options across agents.
- Any intentionally first-provider paths are documented or kept out of the fix.
- The fix is validated against current `master`.

## Triage fields

- Affected component: runtime/daemon, provider, agent
- Severity: S2 - degraded behavior
- ZeroClaw version: observed in PR #6887; current `master` needs fresh verification
- Operating system: provider-independent
- Regression status: unknown

perlowja:
## Summary

- `crates/zeroclaw-providers/src/lib.rs`

## Linked issue

Resolves zeroclaw-labs/zeroclaw#7870

`risk: low` · `size: S`

## Testing (required)

## Compatibility (required)

- Backward compatible? Yes

## Rollback

### DIFF
--- a/crates/zeroclaw-providers/src/lib.rs
+++ b/crates/zeroclaw-providers/src/lib.rs
@@ -4224,6 +4224,75 @@ mod tests {
     }
 
     #[test]
+    fn provider_runtime_options_cross_family_no_leak() {
+        // Two different provider families (openai and anthropic) each with
+        // distinct max_tokens. Each agent resolves only its own provider's
+        // options — the first provider in config insertion order must NOT
+        // leak its max_tokens to the second provider's agent.
+        use zeroclaw_config::schema::{
+            AliasedAgentConfig, AnthropicModelProviderConfig, Config, ModelProviderConfig,
+            OpenAIModelProviderConfig,
+        };
+
+        let mut config = Config::default();
+
+        // First family in insertion order: openai with max_tokens = 16384
+        config.providers.models.openai.insert(
+            "gpt".to_string(),
+            OpenAIModelProviderConfig {
+                base: ModelProviderConfig {
+                    model: Some("gpt-4o".to_string()),
+                    api_key: Some("openai-key".to_string()),
+                    max_tokens: Some(16_384),
+                    ..ModelProviderConfig::default()
+                },
+            },
+        );
+
+        // Second family: anthropic with a different max_tokens
+        config.providers.models.anthropic.insert(
+            "sonnet".to_string(),
+            AnthropicModelProviderConfig {
+                base: ModelProviderConfig {
+                    model: Some("claude-sonnet-4".to_string()),
+                    api_key: Some("anthropic-key".to_string()),
+                    max_tokens: Some(8_192),
+                    ..ModelProviderConfig::default()
+                },
+            },
+        );
+
+        config.agents.insert(
+            "openai_agent".to_string(),
+            AliasedAgentConfig {
+                model_provider: "openai.gpt".into(),
+                ..AliasedAgentConfig::default()
+            },
+        );
+        config.agents.insert(
+            "anthropic_agent".to_string(),
+            AliasedAgentConfig {
+                model_provider: "anthropic.sonnet".into(),
+                ..AliasedAgentConfig::default()
+            },
+        );
+
+        let openai_opts = provider_runtime_options_for_agent(&config, "openai_agent");
+        let anthropic_opts = provider_runtime_options_for_agent(&config, "anthropic_agent");
+
+        assert_eq!(
+            openai_opts.provider_max_tokens,
+            Some(16_384),
+            "openai agent must get its own max_tokens, not the anthropic provider's"
+        );
+        assert_eq!(
+            anthropic_opts.provider_max_tokens,
+            Some(8_192),
+            "anthropic agent must get its own max_tokens, not the openai provider's"
+        );
+    }
+
+    #[test]
     fn resilient_alias_deep_acyclic_fallback_does_not_overflow() {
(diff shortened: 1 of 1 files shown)


perlowja:
Scoped this against current `master`. The runtime call sites #7870 names — agent execution and the LLM-task / delegate tool paths — **already resolve options per-agent** via `provider_runtime_options_for_agent`:

- `crates/zeroclaw-runtime/src/agent/loop_.rs:1282`, `:1785`, `:2266`
- `crates/zeroclaw-runtime/src/tools/mod.rs:1318`

There are **no production call sites** of the first-provider `provider_runtime_options_from_config` left, so the leak described in #7870 (a second provider receiving the first provider's `max_tokens`) is not reproducible on master. Per #7870's own request — *"This needs a fresh reproduction or regression test against current master before closing"* — this PR adds that regression coverage: `provider_runtime_options_cross_family_no_leak` configures two families with different `max_tokens` and asserts each agent resolves only its own provider's options.

Accordingly I've retitled this `test(providers): …` (it's the regression-test closure for an already-fixed bug, not a behavior change) and replaced the dummy provider keys with clearly-fake placeholders so the privacy/secret-scan gate never flags them. `cargo test -p zeroclaw-providers cross_family_no_leak` passes; fmt clean.

If you'd prefer the regression test to also assert at a call-site level (constructing an agent and checking the resolved options), I can add that — but the helper-level test plus the verified call-site usage above covers the #7870 scope.

singlerider (approved):
I pulled this locally, checked it out, and ran it against current `upstream/master`.

### 🟢 Praise - Regression guard targets the real bug, not a tautology

This is the rare auto-generated PR that holds up under a hard look. I checked it against the acceptance criteria in #7870 and verified the surrounding code:

- The function under test, `provider_runtime_options_for_agent`, resolves through `config.model_provider_for_agent`, which keys strictly on the agent's own `(type_key, alias_key)` split from its `model_provider` ref. The named failure mode (first-provider-by-insertion-order leaking `max_tokens` to a different family) is a path the lookup could plausibly regress into, so the assertion is not structurally guaranteed. It is a real guard.
- The fix the test protects has already landed: every in-scope call site from #7870 now resolves via the agent-scoped helpers, not the old `provider_runtime_options_from_config`. Agent execution (`crates/zeroclaw-runtime/src/agent/loop_.rs:1282`), delegate/tool construction (`crates/zeroclaw-runtime/src/tools/mod.rs:1318`), and the orchestrator (`crates/zeroclaw-channels/src/orchestrator/mod.rs:8495`) all use `provider_runtime_options_for_agent` / `provider_runtime_options_for_alias`. The bare `from_config` symbol survives only in unrelated test names. So this PR closes the one remaining gap the issue called out: regression coverage proving multi-provider configs do not cross-leak.
- The test uses two distinct provider families with distinct `max_tokens` and asserts each agent gets its own value, which is the exact reproduction sketch from the issue.

### 🟢 Praise - Verified locally

- `cargo test -p zeroclaw-providers provider_runtime_options_cross_family_no_leak` passes.
- `cargo clippy -p zeroclaw-providers --all-targets -- -D warnings` is clean.
- Uses real schema types (`OpenAIModelProviderConfig`, `AnthropicModelProviderConfig`, `AliasedAgentConfig`) with obviously-fake placeholder keys, no PII.
- Test-only, single file, additive. Zero source risk.

Approving. This satisfies the regression-coverage acceptance criterion in #7870.

perlowja:
@singlerider approved and green here too (18/18, clean) — ready to squash-merge when you have a moment. It's the focused per-agent runtime-option regression for #7870.

perlowja:
Closing to clear the upstream branch list: this branch was pushed to the canonical repo by mistake (inverted `origin`/fork remotes). The change is preserved on my fork as `perlowja:test/providers-runtime-option-resolution` and will be resubmitted as a proper fork-based PR. A fork-only pre-push guard is now in place so this cannot recur. Apologies for the upstream-branch pollution.

Issue #6497, opened 2026-05-07
[Feature]: Gateway — bridge DeviceRegistry ↔ NodeRegistry by device_id


theonlyhennygod:
## Summary
Bridge the two existing registries so the gateway can answer one query: *"Is the Mac with device_id X online and does it have capability Y?"* Currently `DeviceRegistry` (SQLite, persistent identity) and `NodeRegistry` (in-memory, WS capability dispatch) don't talk to each other.

## Problem
- `DeviceRegistry` persists pairing identity but has no concept of online/offline beyond `last_seen` (which only updates on HTTP requests).
- `NodeRegistry` has live WS-based capability dispatch but is ephemeral — no persistence, and uses a separate `node_id` from the `device_id`.
- The agent on the gateway has no single source of truth: "send this command to the Mac" requires consulting both, and there's no guarantee they refer to the same paired entity.

## Proposal
- When a `/ws/nodes` WebSocket authenticates with a device's bearer token, treat the **device_id == node_id**.
- Schema migration on the `devices` table:
  - `online_since TEXT` (RFC3339) — non-null when the device has an active WS connection
  - `node_id TEXT` — populated from the device's UUID on first WS connection
  - capabilities sync from the WS register message (overrides the HTTP-pushed list from this branch's v1)
- On `/ws/nodes` register: look up device by token_hash, set `online_since = now()`, store advertised capabilities.
- On WS disconnect: clear `online_since`, keep the row.
- New gateway helper `device_registry::is_online(device_id) -> bool` and `device_registry::find_online_with_capability(cap) -> Option<DeviceInfo>` for the agent dispatcher to use.

## Files
- `crates/zeroclaw-gateway/src/api_pairing.rs` — schema migration, new fields, helpers
- `crates/zeroclaw-gateway/src/nodes.rs` — bridge logic in the WS handler
- `crates/zeroclaw-gateway/src/lib.rs` — AppState wiring

## Acceptance
- Existing devices migrate cleanly (additive ALTER TABLE).
- Mac connects to `/ws/nodes` with bearer token → its device row gets `online_since` populated and `capabilities` synced.
- Mac kills the connection → `online_since` clears within ~5s; row persists.
- Two `device_registry` regression tests prove migration + bridge logic.

## Related
- Depends on the v1 capability sync from `feat/desktop-onboarding`.
- Sibling issue: persistent NodeClient on the desktop side (separate ticket).
- Capability handlers live in their own ticket (separate).

[...]

PR #7993, opened 2026-06-19
fix(gateway): register devices on legacy /pair + backfill orphaned paired_tokens

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The legacy `POST /pair` handler (`handle_pair`) minted and persisted a bearer
    token but **never called `registry.register`**, so a device paired through it
    authenticates yet has **no `device_registry` row** — invisible in
    `GET /api/devices` and unrevocable from the management UI. `handle_pair` now
    registers the device, exactly like `POST /api/pair` (`submit_pairing_enhanced`).
  - **No startup reconciliation existed**, so tokens already paired via `/pair`
    (and any other historical orphans) stay invisible forever. Added
    `DeviceRegistry::reconcile_from_token_hashes`, called once at registry
    construction, which inserts a neutral `"legacy"` placeholder row for every
    `gateway.paired_tokens` hash lacking a device row. Idempotent;
    `INSERT OR IGNORE` never clobbers a real row.
  - Net effect: every active bearer token is now listable and revocable from the
    UI. This closes a **security-management gap (visibility/revocation)** — the
    auth gate itself was already sound (no access bypass; 401 without a valid
    token is unaffected).
- **Scope boundary:** No change to token generation, hashing, the on-disk
  `gateway.paired_tokens` format, or the auth gate. No change to the `devices`
  table schema. Does **not** touch the rotation/deletion revocation paths
  (`rotate_device_token`, `revoke_device`).
- **Blast radius:** `crates/zeroclaw-gateway` only. Behaviour changes: (1) `/pair`
  now writes a `devices.db` row on success; (2) on startup with
  `require_pairing = true`, the registry gains placeholder rows for any
  previously-orphaned paired tokens.
- **Linked issue(s):** Related #7243 (token-revocation follow-up, same subsystem),
  Related #6497 (registry bridge). No `Closes` — this is the
  registration/backfill counterpart to the merged revocation work, not a 1:1 fix
  for an existing issue.
- **Labels:** `bug`, `gateway`, `security:pairing`, `risk: high`, `size: S`

  ```
  cargo fmt --all -- --check                → exit 0
  cargo clippy --all-targets -- -D warnings → exit 0 (no warnings)
  cargo test -p zeroclaw-gateway            → exit 0
    test result: ok. 265 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
    test api::tests::backfilled_orphan_is_revocable_by_its_real_hash ... ok
    test api::tests::reconcile_backfills_orphan_token_hashes ... ok
  ```
  (Run after rebasing onto current `master`.)
- **Beyond CI — what did you manually verify?** Two new tests assert the contract:
  (1) backfill inserts only orphan hashes, skips/preserves known rows, idempotent;
  (2) **security contract** — a `/pair`-style orphan, after backfill, is keyed by
  the same `token_hash` the auth gate uses, so `DeviceRegistry::revoke` →
  `PairingGuard::revoke_token_hash` actually de-authenticates the bearer token.
- **If any command was intentionally skipped, why:** Ran `cargo test -p
  zeroclaw-gateway` rather than full-workspace `cargo test` (change is gateway-only;
  workspace `clippy --all-targets` already compiled every target clean).

- New permissions / capabilities / file system access scope? **No**
- Secrets / tokens / credentials handling changed? **No** — token generation,
  hashing, and the `paired_tokens` store are untouched. The backfill only reads
  hashes already held in `PairingGuard` and writes *device metadata* rows keyed
  by those hashes. The credential set's source of truth remains `paired_tokens`.
- PII / real identities in diff? **No** — placeholder rows are `name = NULL`,
  `device_type = "legacy"`, `ip_address = NULL`.

## Compatibility (required)

- Backward compatible? **Yes** — additive; no schema migration, no config/CLI surface.
- Upgrade steps for existing users: none. On the next daemon restart with
  `require_pairing = true`, orphaned tokens are automatically backfilled into the
  registry.

## Rollback (required — risk: high)

- **Fast rollback command/path:** `git revert <sha>`. The change is additive and
  idempotent; reverting returns `/pair` to non-registering behaviour. Placeholder
  rows already created remain but are harmless (revocable like any device).
- **Observable failure symptoms:** grep gateway logs for
  `device registry backfill from paired_tokens failed` (DB error on startup —
  logged, non-fatal) or `backfilled legacy paired token(s) into the device
  registry` (success, with a `backfilled` count attr). UI symptom of success:
  `GET /api/devices` lists devices that were previously absent.

### DIFF
--- a/crates/zeroclaw-gateway/src/api.rs
+++ b/crates/zeroclaw-gateway/src/api.rs
@@ -3502,6 +3502,101 @@ mod tests {
         h
     }
 
+    /// The backfill inserts a placeholder row for every paired-token hash that
+    /// has no device entry, skips hashes that already have one (without
+    /// clobbering their metadata), and is idempotent across restarts.
+    #[tokio::test]
+    async fn reconcile_backfills_orphan_token_hashes() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let registry = DeviceRegistry::new(tmp.path());
+
+        // A real, already-registered device with a name.
+        let known_hash = "a".repeat(64);
+        registry.register(
+            known_hash.clone(),
+            DeviceInfo {
+                id: "known".into(),
+                name: Some("My Laptop".into()),
+                device_type: Some("desktop".into()),
+                paired_at: Utc::now(),
+                last_seen: Utc::now(),
+                ip_address: None,
+                capabilities: None,
+            },
+        );
+
+        let orphan_a = "b".repeat(64);
+        let orphan_b = "c".repeat(64);
+        let inserted = registry
+            .reconcile_from_token_hashes(&[known_hash.clone(), orphan_a.clone(), orphan_b.clone()])
+            .unwrap();
+        assert_eq!(inserted, 2, "only the two orphan hashes should be inserted");
+        assert_eq!(registry.device_count(), 3);
+
+        // Existing metadata is preserved, not clobbered.
+        let known = registry
+            .list()
+            .into_iter()
+            .find(|d| d.id == "known")
+            .expect("known device still present");
+        assert_eq!(known.name.as_deref(), Some("My Laptop"));
+
+        // Re-running is a no-op (idempotent).
+        let again = registry
+            .reconcile_from_token_hashes(&[known_hash, orphan_a, orphan_b])
+            .unwrap();
+        assert_eq!(again, 0);
+        assert_eq!(registry.device_count(), 3);
+    }
+
+    /// Security-management contract: a token paired through the legacy `/pair`
+    /// route is an orphan (authenticates, but absent from the registry). After
+    /// backfill it is keyed by the SAME `token_hash` the auth gate uses, so the
+    /// revoke-by-device path (`revoke` → `PairingGuard::revoke_token_hash`)
+    /// invalidates exactly that bearer token.
+    #[tokio::test]
+    async fn backfilled_orphan_is_revocable_by_its_real_hash() {
+        let tmp = tempfile::TempDir::new().unwrap();
+        let data_dir = tmp.path().join("workspace");
+        std::fs::create_dir_all(&data_dir).unwrap();
+
+        let pairing = PairingGuard::new(true, &[]);
+        let code = pairing.pairing_code().unwrap();
+        let token = pairing
+            .try_pair(&code, "legacy-client")
+            .await
+            .unwrap()
+            .unwrap();
+        assert!(pairing.is_authenticated(&token));
+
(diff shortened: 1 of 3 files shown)


singlerider (approved):
Reviewed at db572fa3 on a fresh checkout of `fix/pairing-registry-backfill`. This touches the pairing/auth surface, so I traced the register and reconcile paths in production source and ran the full gateway test target (not lib-only).

### 🟢 The visibility/revocation gap is closed without touching the auth gate

Both halves are correct and verified in source:

- **Legacy `/pair` now registers.** `handle_pair` (lib.rs ~2138) calls `registry.register(PairingGuard::token_hash(&token), …)` on success, behind `if let Some(ref registry)` so it's a clean no-op when the registry is absent. The key is `PairingGuard::token_hash(&token)` — the exact hash the auth gate checks — so the device row is addressable by the same identity used for revocation. The row is metadata only; the token stays owned by `PairingGuard` and is persisted as before.
- **Startup backfill is idempotent and fail-safe.** `reconcile_from_token_hashes(&pairing.tokens())` runs once at registry construction, skips any hash already in cache, and uses `INSERT OR IGNORE` so a real row is never clobbered; it only mints a neutral `"legacy"` placeholder for orphan hashes. A DB error logs ERROR and does not abort startup. The source of truth for which tokens are valid remains `gateway.paired_tokens` — reconcile never invents a token, only surfaces ones that already authenticate.

The security contract is the load-bearing claim, and it has a direct test: `backfilled_orphan_is_revocable_by_its_real_hash` proves a `/pair`-style orphan, after backfill, is keyed by the same `token_hash` such that `DeviceRegistry::revoke` -> `PairingGuard::revoke_token_hash` actually de-authenticates the bearer. `reconcile_backfills_orphan_token_hashes` pins the orphan-only / skip-known / idempotent behavior.

### 🟢 Scope is tight and the auth posture is unchanged

No change to token generation, hashing, the on-disk format, the auth gate, or the rotation/deletion paths. 401-without-a-valid-token is unaffected; this is purely additive management-surface metadata.

Full-target validation at db572fa3 (deliberately not lib-only):

```
cargo test -p zeroclaw-gateway                         → 265 passed; 0 failed (incl. both new tests)
cargo clippy -p zeroclaw-gateway --all-targets -D warnings → clean
```

CI is green across all 18 checks. Relating to #7243/#6497 without a `Closes` is the right call — this is the registration/backfill counterpart to the merged revocation work, not a 1:1 issue fix.

Approving. Sound security-management fix with the de-authentication contract pinned by test.

PR #7995, opened 2026-06-19
fix(zerocode): distinguish selected field from editable input

### LINKED ISSUE #7814: [Bug]: ZeroCode config fields look editable before Enter activates editing
### Affected component

zerocode/tui

### Severity

S2 - degraded behavior

### Current behavior

In ZeroCode's Config editor, clicking a scalar field selects it and can make the row look ready for text input, but typing does not edit the value until the user presses Enter first.

Observed example:

1. Open `runtime_profiles.default.max_tool_iterations`.
2. Click the `max_tool_iterations` entry.
3. The field appears active/selected, and the visible affordance can read like a text cursor.
4. Type digits.
5. Nothing changes because the field list is still in selection mode.
6. Press Enter to activate the field editor.
7. Typing then edits the value as expected.

Current source inspection matches the interaction model:

- A single click in the Config field list calls `set_visible_cursor(pos)`.
- Editing starts only when `activate_mouse()` runs on double-click or when the keyboard handler receives the field-list Enter action and calls `enter_field_edit(...)`.
- Once `Screen::FieldEdit` is active, scalar typing writes into `edit_buf` and the edit view renders the text input with a cursor.
- The confusing part is the transition/affordance: a selected field can look input-ready even though text input is not active yet.

### Expected behavior

The Config editor should make the activation boundary clear.

Reasonable fixes could include:

- single-clicking a scalar field activates editing directly;
- typing while a scalar field row is selected activates editing and applies the typed character;
- the selected-but-not-editing state avoids a cursor-like visual affordance;
- the detail/footer hint says Enter or double-click to edit while the row is selected.

Once edit mode is active, the existing scalar editing flow can remain as-is.

### Steps to reproduce

```bash
1. Open ZeroCode.
2. Go to Config → Runtime profiles → default.
3. Click the `max_tool_iterations` row once.
4. Type a digit.
5. Observe that the row looks selected/active, but the digit is ignored because edit mode is not active yet.
6. Press Enter.
7. Type the digit again and observe that editing now works.
```

### Impact

Config editing is a core ZeroCode workflow. Scalar fields include important runtime controls such as tool-call limits, token budgets, model parameters, and timeouts.

When the selected state looks like an editable state, users can type and see no response. That makes the editor feel broken until they discover that Enter is required to enter edit mode.

### Logs / stack traces

No stack trace. This is a TUI editing/affordance issue.

### ZeroClaw version

Current `master` at `dab313aff50cd221b278774f5a79a61c5eb50806`.

### Rust version

N/A; this is a ZeroCode TUI interaction issue.

### Operating system

macOS observed. The config editor behavior is not expected to be OS-specific.

### Regression?

Unknown.

### Related

- #6825

### Pre-flight checks

- [x] I observed this in ZeroCode and source-confirmed the relevant Config editor paths on latest `master`.

MaHaoHao-ch:
## Summary

- **What changed and why:** A *selected* field row in the ZeroCode Config editor was visually indistinguishable from an *editable* field row. The same `selected_style()` (bold, bright foreground, selection background) plus the `› ` highlight gutter was used for both `Screen::FieldList` (selection) and `Screen::FieldEdit` (active input), so a single click on a scalar field could read as "ready to type" even though typing was dropped until the user pressed Enter or double-clicked. The fix splits the highlight into three explicit visual states (active edit, selected-only, dim/focus-elsewhere) and appends a per-row `  ─→ press Enter to edit` hint to the currently selected field list row. The hint is suppressed once the user enters the edit screen.
- **Scope boundary:** Pure affordance change inside the existing Config pane renderer. No state machine changes, no keybinding changes, no theme palette changes, no behavior change in `enter_field_edit`, `activate_mouse`, or `set_visible_pos`. The `Screen::FieldEdit` rendering and position indicator logic are untouched.
- **Blast radius:** Only `apps/zerocode/src/config_manager.rs`. The Config field list and the same `detail_highlight()` are also used by the type and alias list renderers, so those panes will inherit the new selection-vs-edit distinction as a side effect. No other subsystem, no public API, no CLI, no daemon surface is affected.
- **Linked issue(s):** Closes #7814
- **Labels:** `bug`, `size: XS`, `risk: medium`, `zerocode`. (Labels applied at submission.)

```bash
cargo fmt --all -- --check
cargo test
```

- **Commands run and tail output:** No `cargo fmt` / `cargo clippy` / `cargo test` was executed in this session. The contributing environment did not have a Rust toolchain or `node_modules` for the project on the contributor's machine at the time of authoring, so the full battery was intentionally skipped. The change is a 30-line, single-file patch limited to existing renderer branches that already compile in the upstream build, and it adds no new imports, traits, lifetimes, or types.
- **Beyond CI — what did you manually verify?** (scenarios, edge cases, what you did NOT verify) Verified the diff against the issue's described reproduction by reading the call sites: `set_visible_pos` (line 917) only moves the position indicator, `activate_mouse` (line 998) is the only path into `enter_field_edit` (line 2375), and `enter_field_edit` is the only path that transitions `Screen::FieldList` → `Screen::FieldEdit`. The new `match` arm in `detail_highlight` is exhaustive over the new tuple `(&self.screen, self.zeroclaw_pane)`. The `press_hint` is only appended while building `ListItem`s for the field list, so the hint disappears naturally once the screen transitions to `FieldEdit`. Did not visually run the TUI: the contributor's local environment lacks the Rust toolchain, so a terminal screenshot of the new affordance is not provided.
- **If any command was intentionally skipped, why:** `cargo fmt` / `cargo clippy` / `cargo test` were intentionally skipped because the toolchain is not installed in this session; please rely on the upstream CI to re-run them. The diff is intentionally minimal (no new public items, no new trait bounds, no new modules) to keep the CI delta small.

## Compatibility (required)

- Backward compatible? `Yes`

- **Observable failure symptoms:** If a row appears dimmer than expected after the revert, the field list renderer is using `theme::selected_inactive_style()` instead of `theme::selected_style()`; the previous behavior is restored by reverting this commit.


Audacity88 (changes requested):
Checked the live PR state, prior review/comment state, the one-file diff in `apps/zerocode/src/config_manager.rs`, adjacent field-list filter/rendering source, and the passing GitHub checks. I did not run a local TUI smoke or local Cargo validation.

### 🟢 What looks good — selection vs editing is the right split

Separating FieldList selection from FieldEdit input styling is the right fix shape for the reported edit-affordance ambiguity, and keeping the state machine/keybindings untouched keeps the blast radius small.

### 🔴 Blocking — filtered field lists can put the edit hint on the wrong row

The new `press Enter to edit` affordance is keyed off `self.field_cursor`, but the rendered selection uses `self.filter_cursor` whenever a field-list filter is active. `handle_filter_key()` updates only `filter_cursor` while the user types `/` filters or moves through filtered results, and `draw_field_list()` later highlights `self.filter_cursor` but appends the new hint to the row whose original field index still matches `field_cursor`. That means filtering the Config field list can leave the hint on a stale/unhighlighted row, or remove it from the highlighted row entirely, which reintroduces the exact ambiguity this PR is trying to remove for a core field-list path.

Please derive the hint from the same selected row used for `ListState` selection, for example by resolving `filter_cursor` through `visible` when `self.filter.is_some()` and appending the hint to that visible item.

singlerider (approved):
First review here. Pulled the branch, built `zerocode`, ran clippy, and checked Audacity88's outstanding blocker.

### ✅ RESOLVED — the filtered-list hint is now keyed off the rendered selection

Audacity88's blocker was that the `press Enter to edit` hint was keyed off `field_cursor` while the rendered selection uses `filter_cursor` when a field-list filter is active, so the hint could land on a stale/unhighlighted row. That is fixed correctly. The `cursor` computation (which already resolves `filter_cursor` when `self.filter.is_some()` and otherwise maps `field_cursor` through `visible`) is now hoisted above the item map, `selected_field = visible.get(cursor)` is derived from it, and the hint fires on `Some(i) == selected_field`. Since `ListState::select` uses that same `cursor`, the hint and the highlight are now always on the same row in both filtered and unfiltered paths.

### 🟢 Selection vs editing split is clean

`detail_highlight` now matches on the `(&self.screen, self.zeroclaw_pane)` enum tuple to separate FieldEdit (bold, bright, `› ` gutter), FieldList-selected (dim selection bg, thin `>` gutter), and the dim "you are here" preview. No string-literal match, no keybinding/state-machine change, blast radius stays small.

### 🟢 Validation

`cargo build -p zerocode` clean; the PR's `config_manager.rs` is clippy-clean. The lone `clippy::collapsible_match` warning is at `apps/zerocode/src/doctor.rs:287`, which is pre-existing on `upstream/master` and untouched here, so not a blocker for this PR. Tracking that master lint separately.

Audacity88's blocker is resolved and the fix is correctly scoped. Approving.

Audacity88 (changes requested):
I re-reviewed current head c5c6c7b after my prior changes-requested review, the later approval, the linked #7814 timeline, the related open #8064, live PR/check state, and the one-file `apps/zerocode/src/config_manager.rs` diff. I did not run a local Zerocode TUI smoke or local Cargo validation; this is source review plus the current green CI and existing reviewer validation.

### ✅ Resolved — filtered field hint follows the rendered selection

My earlier blocker is resolved. The second commit now computes the same `cursor` used by `ListState::select(...)`, resolves that cursor through `filter_cursor` while a filter is active, derives `selected_field = visible.get(cursor)`, and appends the hint only when `Some(i) == selected_field`. That keeps the filtered-list highlight and row hint aligned.

### 🔴 Blocking — keep the new hint localized and keymap-derived

The new row-level affordance is user-facing prose, but it is currently built as a hardcoded English fragment: `press Enter to edit`. The surrounding Config UI already has keymap-derived hint helpers for action labels such as edit/back/reset, and the related #8064 footer-hint work follows that same shape by resolving the displayed chord instead of spelling a literal key name. This new row hint should follow that path too. It also lives directly in Rust instead of the Zerocode Fluent catalogue, so the new text is not available to non-English locales.

This matters because the PR's fix is the affordance itself. If a preset or override changes the displayed action chord for `ConfigTabAction::Enter`, the row can still tell the user to press `Enter`; and even without rebinding, localized Zerocode sessions get an English-only row fragment.

Please route the hint through the same two pieces used by the rest of this surface: derive the key label from `tab_key(ConfigTabAction::Enter)` or the equivalent resolved action label, and render the prose through an English Fluent key with a `{ $keys }` slot. Alternatively, if the related footer-hint work in #8064 becomes the chosen UX, dropping this inline prose and relying on that keymap-derived hint path would also solve the drift.

### 🟡 Warning — coordinate the sibling #8064 merge path

#7995 and #8064 are complementary rather than pure duplicates: this PR is the only one currently closing #7814, while #8064 explicitly leaves selected-row visual styling to this PR. They do touch overlapping `config_manager.rs` regions, though, so this PR should not be treated as merge-ready until the two branches are sequenced or rebased cleanly.

### 🟢 What looks good — selection and edit mode stay separated

The visual-state split itself is the right shape. `Screen::FieldList` keeps a dimmer selected-row style, `Screen::FieldEdit` keeps the active input styling, and the existing mouse/keyboard activation paths are left alone. Once the new hint follows the same keybinding/localization rules as the rest of Zerocode, this should be a small, focused fix for the original confusion.

Audacity88 (approved):
I re-reviewed current head 59b42ce after my earlier changes-requested review, the linked #7814 context, the sibling #8064 context, the live PR/check state, and the final diff in `apps/zerocode/src/config_manager.rs` plus `apps/zerocode/locales/en/zerocode.ftl`. I did not run a local Zerocode TUI smoke or local Cargo validation; this is source review plus the current green GitHub checks and prior reviewer validation.

### ✅ Resolved — inline edit hint is localized and keymap-derived

My prior blocker is resolved. The row-level edit hint now comes from the English Fluent catalogue as `zc-config-field-edit-hint = { $keys } → press to edit`, and the renderer passes a key label resolved from `tab_key(ConfigTabAction::Enter)`. That keeps the visible chord tied to the Config keymap instead of hardcoding `Enter` in Rust, while non-English sessions can fall back through the embedded English catalogue until the release-time translation pass localizes the new key.

The earlier filtered-list issue also remains fixed: the hint is still derived from the same `selected_field` row that feeds `ListState::select(...)`, so the highlight and hint stay aligned when a filter is active.

### 🟡 Warning — coordinate the sibling #8064 merge path

This PR and #8064 are still complementary rather than duplicates: #7995 is the narrow selected-row visual affordance fix for #7814, while #8064 handles the broader footer/status hint and scalar-validation work. They touch overlapping `config_manager.rs` regions, so this still needs normal sequencing or a clean rebase/update before merge. That is merge hygiene rather than a remaining author-facing blocker in this PR.

### 🟢 What looks good — the fix stays narrow

The final patch stays inside the Config renderer and the English Zerocode Fluent catalogue. It does not change the edit state machine, mouse activation path, keybinding definitions, daemon config schema, CLI surface, or persistence behavior. With the hint now both localized and keymap-derived, this looks like the focused fix the original affordance bug needed.

PR #7997, opened 2026-06-19
feat(runtime): universal ingress policy layer (default Loop)

Nillth:
> **Stacked on #7969** (the `ToolLoop` params-struct refactor). Until #7969
> merges, this PR's diff includes #7969's commit — **review only the top commit**
> (the ingress layer). I'll rebase to drop the base once #7969 lands, leaving
> just this change.

## What

Stands up the universal ingress policy layer from RFC #6971: **every inbound
turn passes one SOP-backed policy layer before a model sees it** — on every
transport (channel, gateway, ACP, RPC) including mid-turn steering injections.

- `zeroclaw-api`: `IngressContext { message_id, source_class, sender, transport,
  trust }`, `IngressDecision { Loop, Annotate, Gate, Drop }`, and the supporting
  enums + `IngressContext::internal()`.
- `zeroclaw-runtime`: `ingress_policy(text, ctx, policy) -> IngressDecision`; the
  default policy returns `Loop` for everything.
- The context rides as a **field on `ToolLoop`** (not a positional arg — that's
  why #7969 came first). The two engine choke points consult it: **P1** at
  `run_tool_call_loop` entry, **P2** at the steering drain.

## The model

Always-on, no bypass — but behavior-identical. The default disposition is `Loop`
(run the agent, today's behavior) and it is **free**: it allocates no SOP run and
does no IO, so chat never consumes run slots. Only a `Gate` disposition (a later
slice) allocates a managed SOP run. So *every turn passes the policy* without
*every turn being an SOP run*.

## Not a behavior change

Purely additive: **+564 / −0** — zero deletions, no existing logic or test
assertion changed. Default `Loop` ⇒ identical output. Verified:

- `cargo clippy -p zeroclaw-api -p zeroclaw-runtime -p zeroclaw-channels
  --all-targets -- -D warnings` clean.
- runtime tests green incl. the `run_tool_call_loop` + safety-net **oracle**
  suites unchanged, the new `ingress_policy` default-`Loop` tests, and a
  behavior-identical guard test (internal vs fully-external envelope ⇒ identical
  output). channels tests green. Root binary links the full graph.
- Engine-validation harness, live tier: 5/5 — CLI one-shot, delegate, RPC,
  WS w/ mid-turn steering (confirms P2 end-to-end), ACP deny-with-edit.

## Scope / series

This is the always-on layer with the default (`Loop`) policy only. NOT in this
PR: real per-transport `IngressContext` stamping (every caller stamps
`internal()` for now) and the `[ingress]` config that enables non-`Loop` policy
(Gate/Drop/Annotate/routing). Those are the next slices.

Part of the #6971 ingress series (2/N) · stacked on #7969.

### DIFF
--- a/crates/zeroclaw-api/src/ingress.rs
+++ b/crates/zeroclaw-api/src/ingress.rs
@@ -1,0 +1,134 @@
+//! Universal ingress policy contract types (RFC #6971, phase 1).
+//!
+//! Layer: api (kernel ABI). Pure types, no logic, no IO. These describe the
+//! envelope that every inbound turn carries into the unified turn engine and
+//! the disposition the SOP policy layer returns for it.
+//!
+//! The entry layer (channel orchestrator, gateway, ACP, RPC, cron/SOP/subagent)
+//! stamps an [`IngressContext`] and threads it into `run_tool_call_loop`; the
+//! engine consults the policy front door (`ingress_policy`, in
+//! `zeroclaw-runtime`) which returns an [`IngressDecision`]. The default
+//! disposition is [`IngressDecision::Loop`] — run the agent, today's behavior.
+//!
+//! This module does NOT decide policy (that is the runtime front door) and does
+//! NOT stamp real identity (phase 2). Phase 1 ships these shapes and the
+//! always-on threading; only `Loop` is reachable under the default policy.
+
+use serde::{Deserialize, Serialize};
+
+/// Whether an inbound turn originates outside the agent (a transport peer) or
+/// from an internal driver (cron, an SOP step, a subagent).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum SourceClass {
+    /// A message from a transport peer (channel user, webhook caller, …).
+    External,
+    /// An internally driven turn (cron, SOP step, subagent).
+    Internal,
+}
+
+/// The transport an inbound turn arrived on. Real per-transport stamping is
+/// phase 2; phase 1 stamps [`Transport::Internal`] everywhere.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum Transport {
+    /// A messaging channel — `kind` is the channel type (e.g. `"github"`),
+    /// `alias` the configured channel alias.
+    Channel { kind: String, alias: String },
+    /// The HTTP/WebSocket gateway (REST/WS turn).
+    Gateway,
+    /// Agent Client Protocol (local IDE bridge).
+    Acp,
+    /// RPC socket turn (zerocode path).
+    Rpc,
+    /// An internally driven turn with no external transport.
+    Internal,
+}
+
+/// Trust class resolved for the turn's sender. Minimal for phase 1; peer-group
+/// resolution (the real source) is phase 2.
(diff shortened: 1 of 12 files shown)


singlerider (approved):
Reviewed the net-new commit `d0dbc3b8` only, per the stacking note in the PR body. The base commit is #7969's `ToolLoop` params-struct refactor, which I already approved separately. Everything below is scoped to the ingress layer the top commit adds.

I read the contract types in `zeroclaw-api`, the runtime front door, both engine choke points, and every caller that stamps the envelope, then cross-checked against the claims in the description and ran the validation locally.

### 🟢 The always-on-but-free model holds up under inspection

The central claim, that every turn passes one policy layer while only a `Gate` disposition ever costs a run, is real and verifiable in the code, not just the prose. `ingress_policy` is pure, synchronous, infallible, allocates nothing, and does no IO on the default path. P1 (`turn/mod.rs:262`) and P2 (`turn/mod.rs:343`) both consult it, both match exhaustively, and the `Loop` arm is a true no-op pass-through at each site. Under the default `IngressPolicy` only `Loop` is reachable, so the runtime is byte-identical to one with no layer. The `+564 / −0` framing checks out: I confirmed zero deletions and no existing assertion touched.

### 🟢 The unreachable arms degrade safely

The thing I most wanted to verify on an always-on choke point is what happens when a non-default disposition becomes reachable and the handling isn't written yet. There are no `todo!()`/`unimplemented!()`/`panic!` in any reachable arm. `Gate` is a documented no-op with a `TODO(PR C)` marker at both sites (proceed-as-loop), `Annotate` is proceed-as-loop, and `Drop` does the semantically correct thing at each point: P1 returns the localized `turn-ingress-dropped` string (key present in `cli.ftl:794`), P2 `continue`s so the injection is never appended. That means flipping on a future policy can't wedge or crash the engine mid-turn; worst case under an incomplete phase is proceed-as-today. Good defensive shape for a layer that sits in front of every model call.

### 🟢 Envelope is threaded, never dead code

Every phase-1 caller stamps `IngressContext::internal()` (orchestrator, delegate, safety_net, skills/review, agent.rs), and `ingress` is consumed at both choke points, so the field carries its weight under `-D warnings` rather than being a placeholder the compiler would flag. The phase-2/phase-3 boundary (real per-transport stamping, `[ingress]` config, non-`Loop` policy) is clearly fenced in the module docs and matches RFC #6971's staging.

### 🟢 Validation reproduced locally

Not taking the description's word for it:

- `cargo clippy -p zeroclaw-api -p zeroclaw-runtime -p zeroclaw-channels --all-targets -- -D warnings` → clean (`CLIPPY_EXIT=0`).
- `cargo test -p zeroclaw-api ingress` → `internal_envelope_is_internal_and_trusted` ok.
- `cargo test -p zeroclaw-runtime ingress` → 4/4 ok, including `run_tool_call_loop_ingress_default_loop_is_behavior_identical` (the guard that internal vs fully-external envelope produce identical output) and the default-`Loop` front-door tests over internal, external-untrusted, and empty text.

The external-untrusted test dispositioning an `"ignore previous instructions"` payload to `Loop` is the right phase-1 assertion: it documents that phase 1 deliberately does not yet gate on trust, so nobody mistakes this slice for the enforcement slice.

Approving the top commit on its merits. This is a clean, well-fenced, behavior-identical foundation for the #6971 series. The stacking is handled honestly in the description and will resolve when #7969 lands and you rebase to drop the base.

Issue #7815, opened 2026-06-17
[Bug]: ZeroCode Config does not show which config source/state it is editing


Audacity88:
## Affected component

zerocode/tui

## Severity

S2 - degraded behavior

## Current behavior

ZeroCode's Config pane does not make the active config source or daemon-backed state visible while editing ZeroClaw settings. In particular, when ZeroCode is launched against an alternate config directory or an already-running daemon, the Config pane shows the setting path and value but does not show which config directory/socket/daemon state those values came from.

This makes it easy to misread a valid value as stale or unsaved when the user is actually looking at a different config source, or when they are looking at the daemon's current in-memory config rather than the file they edited elsewhere.

The current source shape appears to support that confusion:

- `apps/zerocode/src/client.rs` resolves the config directory from `--config-dir`, `ZEROCLAW_CONFIG_DIR`, then the default home config directory.
- `apps/zerocode/src/main.rs` uses that directory to resolve the local socket and to spawn an ephemeral daemon with `zeroclaw daemon --ephemeral --config-dir ...` when no daemon is reachable.
- `apps/zerocode/src/config_manager.rs` loads Config values through daemon RPC (`config/list`, `config/set`) and the pane/header/status surfaces do not appear to include the active config directory or whether the value is daemon in-memory state.
- The global status bar shows connection state and TUI id, but not the active config source.

This issue is not claiming that config writes fail. It is about the missing source/state cue that makes successful writes hard to reason about during dogfooding and troubleshooting.

## Expected behavior

ZeroCode should expose enough source/state information that users can tell what they are editing before they trust a displayed config value.

At minimum, one of the Config/status surfaces should show:

- the active config directory or a safe, compact identity for it;
- whether the Config pane is reading daemon-backed in-memory state, on-disk config, or both;
- any pending reload/drift state when the daemon's active state may differ from disk.

## Steps to reproduce

```bash
zerocode --config-dir /tmp/zeroclaw-a
```

1. Open `Config`.
2. Navigate to a runtime-profile field such as `runtime_profiles.default.max_tool_iterations`.
3. Compare the displayed value with the same property in another config source, such as the default config directory or a second disposable config directory.
4. Observe that the Config pane does not identify which config source or daemon state produced the displayed value.

## Impact

[...]

PR #7999, opened 2026-06-19
fix(zerocode): surface active config directory in Config header

MaHaoHao-ch:
## Summary

- **What changed and why:** ZeroCode's Config pane now displays the active config directory on the section tab bar, so users can tell which on-disk state the displayed values came from when running with `--config-dir`, `$ZEROCLAW_CONFIG_DIR`, or a daemon backed by a non-default config source. Previously `config_dir` was resolved by `client::resolve_config_dir` and then handed straight into `ZerocodePane::new`, with no field on `App` retaining it and no UI surface showing it. The fix stores it on `App`, adds a small `shorten_home` helper that turns `/home/alice/.zeroclaw` into `~/.zeroclaw` (with `HOME` / `USERPROFILE` fallback and a graceful non-canonicalize fallback), and appends a dim `config: <dir>` span to the section tab bar so the source is visible regardless of which Config sub-section the user is editing.
- **Scope note:** This PR addresses only the active config directory cue from #7815. The daemon-backed in-memory state and pending reload/drift-state affordances are left for follow-up work.
- **Scope boundary:** Affordance-only change. The `config_dir` resolution logic in `client.rs` is untouched. The daemon connection state, the ephemeral daemon lifecycle, `enter_field_edit`, `activate_mouse`, the state machine, keybindings, and theme palette are all unchanged. The displayed value is computed at construction time from the existing `App::config_dir` field, which is itself populated from the already-resolved path that `main.rs` passes into `App::new`.
- **Blast radius:** Only `apps/zerocode/src/config_manager.rs`. The `draw_section_tab_bar` change adds a single trailing span; `App` gains a `PathBuf` field and a cached display string; one small free function is added. No public API, no CLI flag, no daemon surface, no RPC contract is touched. The existing `ZerocodePane::new(config_dir)` call site continues to receive the same `&Path` it received before.
- **Linked issue(s):** Part of #7815
- **Labels:** bug, size: XS, risk: medium, zerocode.

```bash
cargo fmt --all -- --check
cargo test
```

- **Commands run and tail output:** No `cargo fmt` / `cargo clippy` / `cargo test` was executed in this session. The contributing environment did not have a Rust toolchain at the time of authoring, so the full battery was intentionally skipped. The diff is intentionally minimal: one new free function with no generics, no trait bounds, no lifetimes; one new `PathBuf` field and one cached `String` field initialized in `App::new`; one extra trailing `Span::styled` call in the existing section tab bar renderer. `shorten_home` is covered by reading the four branches (no home var, canonicalize ok, canonicalize err, strip_prefix err) and tracing the returned `String` against the input path.
- **Beyond CI — what did you manually verify?** Verified by reading the call sites: `client::resolve_config_dir` (`apps/zerocode/src/client.rs:144-164`) is the only source of the path, `main.rs:213` resolves it before `app::run` and `App::new` receive it, and the new `App::config_dir` field is the only place that field is read. The `shorten_home` helper is exhaustively traced: missing home var -> `path.display().to_string()`; canonicalize ok -> strip_prefix against canonicalized home, with `~/` prefix on success or original on failure; canonicalize err -> same logic against the raw path. Did not visually run the TUI: the contributor's local environment lacks the Rust toolchain, so a terminal screenshot of the new affordance is not provided.
- **If any command was intentionally skipped, why:** `cargo fmt` / `cargo clippy` / `cargo test` were intentionally skipped because the toolchain is not installed in this session; please rely on the upstream CI to re-run them. The diff is intentionally minimal (no new public items, no new trait bounds, no new modules) to keep the CI delta small.

## Compatibility (required)

- Backward compatible? `Yes`

- **Fast rollback command/path:** Revert the squash merge commit from #7999 (`git revert <merge-sha>`). Reverting removes the cached `config_dir_display` field, `shorten_home`, and the trailing `config: <dir>` span from the Config section tab bar; behavior otherwise returns to the previous Config pane.
- **Observable failure symptoms:** The Config section tab bar shows the wrong config path, fails to shorten the active home directory, or the added `config: <dir>` span causes layout/readability problems.


apps/zerocode/src/config_manager.rs
```diff
@@ -211,10 +211,40 @@ fn switch_tabs_keys() -> Vec<String> {
...
+    let home = std::env::var_os("HOME").or_else(|| std::env::var_os("USERPROFILE"));
+    let Some(home) = home else {
+        return path.display().to_string();
+    };
+    let home_path = PathBuf::from(home);
+    let to_tilde = |p: &Path| -> String {
+        match p.strip_prefix(&home_path) {
+            Ok(rel) => format!("~/{}", rel.to_string_lossy().trim_end_matches('/')),
+            Err(_) => p.display().to_string(),
+        }
```
singlerider, line 234: 🔴 Blocking (two issues at this helper):

1. Symlinked `$HOME`: `path.canonicalize()` resolves symlinks but `home_path` is the raw env value, so `strip_prefix` fails on symlinked homes (macOS/NixOS/containers) and the full path leaks. Canonicalize both sides: `let home_path = PathBuf::from(home).canonicalize().unwrap_or_else(|_| PathBuf::from(home));`

2. `canonicalize()` is a filesystem syscall, and this runs every frame from `draw_section_tab_bar`. The config dir is fixed for the session, so compute the display string once in `App::new` and cache it, rather than re-stat-ing on every render.

singlerider (changes requested):
Pulled this branch locally against `upstream/master`. `cargo build -p zerocode` compiles clean, no warnings (the PR body notes the toolchain was unavailable at authoring time, so this is the first build).

### 🔴 Blocking - symlinked `$HOME` defeats the helper

`shorten_home` canonicalizes the config path but compares it against the raw, un-canonicalized `$HOME`. When `$HOME` is a symlink (common on macOS, NixOS, and many container setups), `path.canonicalize()` resolves the real target while `home_path` stays the symlink, so `strip_prefix` fails and the user sees the full ugly path - the exact thing this feature exists to avoid.

Reproduced locally with `$HOME = /tmp/zc_link -> /tmp/zc_real` and config path `/tmp/zc_link/.zeroclaw`:

```
symlinked HOME: /tmp/zc_real/.zeroclaw
```

Expected `~/.zeroclaw`. Fix: canonicalize both sides (or neither). Canonicalizing the home dir too makes the comparison symmetric:

```rust
let home_path = PathBuf::from(home).canonicalize().unwrap_or_else(|_| PathBuf::from(home));
```

### 🔴 Blocking - filesystem syscall in the render path

`draw_section_tab_bar` runs on every frame (the event loop calls `draw` each iteration and on every keypress), and it calls `shorten_home`, which calls `path.canonicalize()` - a blocking filesystem syscall - every single render. The config directory does not change during a session, so resolve the display string once in `App::new` and store it (e.g. a `config_dir_display: String` field), then the tab bar just reads the cached string. This also sidesteps the symlink fix landing in a hot path.

### 🟢 Scope and structure are right

The affordance itself is correctly scoped: one cached `PathBuf` field, one trailing dim span, no registry list hardcoded, no RPC/CLI/daemon surface touched, `client::resolve_config_dir` untouched. The fallback chain (no home var -> raw path; canonicalize err -> raw path) degrades gracefully. Once the display value is computed once at construction with symmetric canonicalization, this is good.

One note on the validation evidence: skipping the full battery because the toolchain was unavailable is honest, but the build does need to pass before merge. I ran `cargo build -p zerocode` and it is clean; please still run `cargo fmt`/`clippy`/`test` once you have a toolchain so the evidence section reflects a real run.

Audacity88:
Thanks for taking on this Config-pane affordance, @MaHaoHao-ch. I checked the live PR state, #7815, singlerider's review, and the one-file `apps/zerocode/src/config_manager.rs` diff.

Update after 08e1cfc: The new commit appears to address singlerider's two code blockers by caching `config_dir_display` in `App::new` and canonicalizing the home/config path comparison symmetrically.

One scope note before this closes #7815: the issue also calls out the daemon-backed in-memory state versus on-disk config source cue. If this PR is only meant to land the directory cue, please change the close reference to `Refs #7815` / `Part of #7815`, or update the body to explain why the directory cue is enough to close the issue. That keeps the remaining daemon-state/drift affordance from disappearing when this smaller UI fix lands.

singlerider (approved):
Re-reviewed at head `92c1d6091e`. Pulled the branch, built `zerocode`, ran clippy.

### ✅ RESOLVED — both blockers fixed correctly

1. **Symlinked `$HOME`**: `shorten_home` now canonicalizes the home dir (`PathBuf::from(&home).canonicalize().unwrap_or_else(|_| PathBuf::from(home))`) so the comparison is symmetric with the canonicalized config path. A symlinked `$HOME` now strips to `~/...` as intended.
2. **Syscall in the render path**: the display string is computed once in `App::new` (`config_dir_display: shorten_home(config_dir)`) and cached on the struct; `draw_section_tab_bar` reads the cached `String`, so there is no per-frame `canonicalize()` syscall anymore.

The fallback chain still degrades gracefully (no home var → raw path; canonicalize err → raw path), and the affordance stays correctly scoped: one cached field, one trailing dim span, no registry shadowing, no RPC/CLI surface touched.

### 🟢 Validation

`cargo build -p zerocode` clean. The PR's own file (`config_manager.rs`) is clippy-clean. There is a `clippy::collapsible_match` warning at `apps/zerocode/src/doctor.rs:287`, but I confirmed it is pre-existing on `upstream/master` (HEAD `1a3fdb9b95`) and is untouched by this PR, so it is not a blocker here. I will track that master lint separately.

Both blockers resolved, the fix is tight and correctly scoped. Approving.

Audacity88 (approved):
I rechecked current head `92c1d6091e58626a7cc56c1e71e24c0ce045ffa5`, the #7815 scope, my earlier scope note, @singlerider's review and re-review, the one-file `apps/zerocode/src/config_manager.rs` diff/source, live CI, and the skipped local TUI/Cargo evidence. Thanks for the update, @MaHaoHao-ch.

### ✅ Resolved — The #7815 scope is now correctly narrowed

My earlier note was about closing semantics, not the code shape: #7815 also asks for daemon-backed in-memory state and reload/drift-state cues. The PR body now says this is `Part of #7815`, and the live PR no longer has a closing reference, so this slice will not accidentally close the broader issue when it lands.

### ✅ Resolved — The two code blockers are fixed

I agree with @singlerider's re-review here. `shorten_home` now canonicalizes the home directory before comparing it with the canonicalized config path, so symlinked-home setups can still shorten to `~/...`. The display value is also computed once in `App::new` as `config_dir_display`, and `draw_section_tab_bar` only reads that cached string, so the render path no longer does a filesystem canonicalization on every frame.

### 🟢 What looks good — The affordance stays narrow

The implementation keeps the source of truth unchanged: `client::resolve_config_dir` and the existing `ZerocodePane::new(config_dir)` path still receive the same resolved directory. This PR only stores a display string on `App` and appends one dim `config: <dir>` span to the Config section tab bar, which is the right-size fix for the directory-cue part of #7815.

All visible CI checks are green at the current head. I did not run local Cargo or a local TUI smoke for this review; approval is based on source inspection, the prior local build/clippy evidence from @singlerider, and current CI.

Issue #6037, opened 2026-04-23
[Bug] Cron jobs can be launched repeatedly while still running


martintrojer:
## Summary

A cron job can be launched repeatedly if it runs longer than the scheduler poll interval. In my case, the `daily-news-brief` job was executed 20 times in a burst on 2026-04-23 between 07:15:08Z and 07:18:05Z (08:15-08:18 Europe/London), even though it was scheduled once per day at 08:15.

## Observed behavior

- The cron job stayed `enabled` and remained due until its run completed.
- The scheduler polls on a fixed interval and selects jobs with `next_run <= now`.
- Because `next_run` is only advanced after execution finishes, the same job can be picked again on the next poll while the first run is still in flight.
- In my logs, this produced 20 successful runs of the same job in a 3-minute window.

## Evidence

Local config at the time:
- `scheduler_poll_secs = 15`
- `scheduler.max_concurrent = 4`
- `[cron] catch_up_on_startup = true`

Database evidence from `workspace/cron/jobs.db`:
- `daily-news-brief` had 20 `cron_runs` entries between `2026-04-23T07:15:08.284985+00:00` and `2026-04-23T07:18:05.958103+00:00`, all `ok`.

Relevant code path in `crates/zeroclaw-runtime/src/cron/scheduler.rs` and `crates/zeroclaw-runtime/src/cron/store.rs`:
- `due_jobs()` selects rows where `enabled = 1 AND next_run <= ?1`
- jobs are executed first
- `reschedule_after_run()` updates `next_run` only after the run finishes
- there is no in-flight claim/lock before execution

## Expected behavior

A due job should be claimed atomically before execution so it cannot be picked by subsequent scheduler polls until the current run finishes.

## Suggested fix

Introduce a DB-level claim step or in-flight state before execution, so the scheduler can atomically mark a job as owned by a run before calling the agent/shell executor. That should prevent duplicate launches even when a job runs longer than the poll interval.

## Workaround

Increasing `scheduler_poll_secs` reduces the likelihood of the bug, but does not fix the underlying race.

## Notes

I can provide the exact local run history and config if helpful.

Audacity88:
[...]

PR #8001, opened 2026-06-19
feat(sop): add SopRunStore trait + in-memory backend (EPIC B scaffold)

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** Adds the durable run-state store contract that SOP
  durability/concurrency/observability work will build on. ONE abstraction
  (`SopRunStore`) for run-state persistence, CAS-claim admission, an append-only
  event log, and procedural-memory proposals, so those concerns ride a single
  store rather than three ad-hoc mechanisms. This is the contract-first first
  slice; it is additive and not yet wired into the engine, so there is no
  runtime behavior change.
- **Scope boundary:** New `crates/zeroclaw-runtime/src/sop/store/` module only
  (+5 lines of re-exports in `sop/mod.rs`). No engine call-site, config, or tool
  surface touched. Default backend is the existing in-memory behavior.
- **Blast radius:** `sop/store/model.rs` (wire shapes), `sop/store/mod.rs` (trait
  + `InMemoryRunStore` + `build_run_store()` + tests), `sop/mod.rs` (re-exports).
  Nothing consumes the store yet.
- **Linked issue(s):** Groundwork for durable SOP run state. Underpins the fixes
  for #6689 (audit persistence) and #6037-class concurrency claims, and the
  resume/rehydrate path. Does not close an issue on its own.
- **Labels:** `enhancement`, `runtime`, `size: L`, `risk: high` (auto-labeled:
  runtime scope. Additive + unwired, so no behavior changes despite the risk
  tier, but the contract is what the durable backend will be held to, so it is
  reviewed as load-bearing).

## What it adds

- `trait SopRunStore` - run state (save/finish/load/active), CAS claim primitive
  (try_claim/heartbeat/release/expired), append-only event log
  (append_event/list_events), proposal namespace, and maintenance (prune /
  health_check / backend).
- Versioned wire shapes: `PersistedRun` (envelope: version + monotonic revision
  + stall ts + redaction marker + trigger source), `ClaimToken`,
  `SopEventRecord`, `ProposalRecord`/`ProposalStatus`, `RetentionPolicy`,
  `SOP_STORE_VERSION`.
- `StoreError` (+ `Display`/`Error`/`From` conversions), including a distinct
  `RevisionConflict` for divergent same-revision writes.
- `InMemoryRunStore` default (mirrors today's in-memory run maps; lost on
  restart) + `build_run_store()` factory.

## Review remediation (addresses the CHANGES_REQUESTED review)

1. **Same-revision overwrite (blocking).** Both write paths are now revision-
   guarded by one shared `revision_guard`: a strictly-older revision is rejected
   `StaleRevision`; an equal revision is accepted only as a byte-identical
   idempotent retry, otherwise rejected `RevisionConflict`; a newer revision
   wins. `finish_run` runs the same guard before it flips terminal + releases the
   claim, so a stale or divergent terminal write can no longer clobber newer
   state or drop a live claim. Regressions added for divergent same-revision
   saves and for stale/divergent terminal writes (state + claim verified intact).
2. **CAS claim could not hold both concurrency limits (blocking).**
   `try_claim_run` now takes both `per_sop_cap` and `global_cap` and enforces
   them atomically under one lock: live claims for the SOP are counted by
   `sop_name` against the per-SOP cap, and total live claims against the global
   cap, mirroring the engine `can_start` (`sop.max_concurrent` and
   `config.max_concurrent_total`, both inclusive-max, 0 admits nothing). Added a
   two-SOP regression proving one SOP at its per-SOP cap does not block a
   different SOP while the shared global cap still applies.
3. **Body label/risk drift (warning).** This body now snapshots the live labels
   (`risk: high`, `size: L`) and fills the risk-high rollback + compatibility
   sections below.

## Not in this PR (follow-ons, per the EPIC B plan)

`SqliteRunStore` durable backend, the `Memory`-backed adapter, the config-driven
`build_run_store(cfg, data_dir, memory)` factory, boot `restore_runs()`
rehydrate, and the engine persist-seams (the behavior-changing slice that makes
runs actually durable). Each lands as its own reviewed commit.

```
$ cargo +1.93.0 fmt -p zeroclaw-runtime -- --check
(clean)

$ cargo +1.93.0 clippy -p zeroclaw-runtime --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 05s

$ cargo +1.93.0 test -p zeroclaw-runtime sop::store
running 9 tests
test result: ok. 9 passed; 0 failed; 0 ignored; 0 measured; 2299 filtered out
```

- **Commands run and tail output:** above (fmt clean, clippy 0 warnings,
  9 store tests green).
- **Beyond CI - what was manually verified:** the new dual-cap claim semantics
  were checked against the live engine admission path
  (`sop/engine.rs` `can_start`, lines 92-123) so the store mirrors the existing
  per-SOP + global limits exactly (inclusive-max, 0 admits nothing). Did NOT
  exercise the store under real engine wiring (it is intentionally unwired in
  this PR) or under multi-process contention (the in-memory backend is
  single-process; the durable backend lands later).

Tests cover: CAS single-winner + cap bound (+ terminal-run not re-claimable),
per-SOP isolation vs shared global cap, append-only monotonic seq + ordering,
proposal round-trip + status filter, revision guard on save (stale + divergent
same-revision), revision guard on finish (stale + divergent terminal protects
state and claim), and oldest-first prune.

- Secrets / tokens / credentials handling changed? `No` (the `redacted` marker is
  a field shape only; no redaction logic runs in this PR)

## Compatibility (required)

- Backward compatible? `Yes` (additive, unwired; default backend is the existing
  in-memory behavior)
- Upgrade steps for existing users: none.

- **Fast rollback command/path:** `git revert <merge-sha>`. The module is
  self-contained and has no consumers, so a revert is clean.
- **Feature flags or config toggles:** None (nothing is wired; the store is not
  constructed on any runtime path yet).
- **Observable failure symptoms:** None expected at runtime, since no code path
  constructs or calls the store. A compile-time regression would surface in CI
  `Build`/`Test`; there is no log line or metric to grep because the store is
  not yet exercised.

### DIFF
--- a/crates/zeroclaw-runtime/src/sop/mod.rs
+++ b/crates/zeroclaw-runtime/src/sop/mod.rs
@@ -3,11 +3,16 @@ pub mod condition;
 pub mod dispatch;
 pub mod engine;
 pub mod metrics;
+pub mod store;
 pub mod types;
 
 pub use audit::SopAuditLogger;
 pub use engine::SopEngine;
 pub use metrics::SopMetricsCollector;
+pub use store::{
+    ClaimToken, PersistedRun, ProposalRecord, ProposalStatus, SopEventRecord, SopRunStore,
+    StoreError, build_run_store,
+};
 #[allow(unused_imports)]
 pub use types::{
     DeterministicRunState, DeterministicSavings, Sop, SopEvent, SopExecutionMode, SopPriority,
--- a/crates/zeroclaw-runtime/src/sop/store/mod.rs
+++ b/crates/zeroclaw-runtime/src/sop/store/mod.rs
@@ -30,10 +30,15 @@ pub use model::{
 /// listener, and the gateway approve surface.
 pub trait SopRunStore: Send + Sync {
     // ── run state (persistence-resume, state-machine) ──
-    /// Persist-before-mutate. Monotonic-revision-guarded; idempotent on revision.
+    /// Persist-before-mutate. Revision-guarded: a strictly-older revision is
+    /// rejected as `StaleRevision`; an equal revision is accepted only as a
+    /// byte-identical idempotent retry, else `RevisionConflict`; a newer
+    /// revision wins.
     fn save_run(&self, run: &PersistedRun) -> Result<(), StoreError>;
     /// Move a run to terminal state (kept as a terminal record, not deleted) and
-    /// release any live claim.
+    /// release any live claim. Revision-guarded exactly like `save_run`, so a
+    /// stale or divergent terminal write cannot clobber newer state or release a
+    /// live claim.
     fn finish_run(&self, run_id: &str, terminal: &PersistedRun) -> Result<(), StoreError>;
     /// Boot-rehydrate source: every non-terminal run (latest revision per id).
     fn load_active_runs(&self) -> Result<Vec<PersistedRun>, StoreError>;
@@ -41,13 +46,18 @@ pub trait SopRunStore: Send + Sync {
     fn load_run(&self, run_id: &str) -> Result<Option<PersistedRun>, StoreError>;
 
     // ── CAS claim primitive (concurrency-control) ──
-    /// Atomic single-winner admission. Returns `Some(token)` to exactly one
-    /// caller iff under `cap` and no live claim exists for `run_id`, else `None`.
+    /// Atomic single-winner admission honoring BOTH concurrency limits. Returns
+    /// `Some(token)` to exactly one caller iff no live claim exists for `run_id`,
+    /// the run is not terminal, the live claims for `sop_name` stay below
+    /// `per_sop_cap`, AND total live claims stay below `global_cap`. Both caps
+    /// are inclusive maxima counted under one lock (mirrors the engine
+    /// `can_start`); a cap of 0 admits nothing. Otherwise `None`.
     fn try_claim_run(
         &self,
         run_id: &str,
         sop_name: &str,
-        cap: usize,
+        per_sop_cap: usize,
+        global_cap: usize,
     ) -> Result<Option<ClaimToken>, StoreError>;
     /// Renew a claim's lease (tick liveness). No-op if the claim is gone.
     fn heartbeat_claim(&self, token: &ClaimToken) -> Result<(), StoreError>;
@@ -90,6 +100,13 @@ pub enum StoreError {
         have: u64,
         found: u64,
     },
+    /// A same-revision write whose content diverges from the stored run. The
+    /// caller must bump `revision` to record new state; only a byte-identical
+    /// retry at the same revision is accepted (idempotent).
+    RevisionConflict {
+        run_id: String,
+        revision: u64,
+    },
     /// A claim was lost to a concurrent winner (over-cap or already claimed).
     ClaimLost,
 }
@@ -107,6 +124,10 @@ impl std::fmt::Display for StoreError {
             } => write!(
                 f,
                 "sop store stale revision for run {run_id}: have {have}, found {found}"
+            ),
+            Self::RevisionConflict { run_id, revision } => write!(
+                f,
+                "sop store revision conflict for run {run_id}: divergent write at revision {revision}"
             ),
             Self::ClaimLost => write!(f, "sop store claim lost to a concurrent winner"),
         }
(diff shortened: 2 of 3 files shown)


Audacity88 (changes requested):
I reviewed the current public PR snapshot at head `21b5353`, including the captured GitHub diff, green GitHub checks, the related #6689 and #6037 context, and public source excerpts from this PR plus the current SOP engine concurrency checks. I did not run local Cargo validation; this review is based on public PR artifacts, remote CI, and source inspection.

### 🟢 What looks good — one store contract is the right direction

Putting SOP run persistence, admission claims, audit events, and procedural-memory proposals behind one `SopRunStore` abstraction is the right shape for the follow-on durability work. It avoids creating separate persistence mechanisms for each EPIC B/C/F concern, and the in-memory backend gives later wiring a concrete contract to test against.

### 🔴 Blocking — same-revision writes can still overwrite divergent run state

`save_run` only rejects an incoming run when the stored revision is strictly greater than the incoming revision. The PR also adds a test that explicitly allows equal-revision writes and calls them idempotent, but the implementation accepts any equal-revision payload, not only a retry of the same payload. Today two paths can both save different `PersistedRun` objects at revision `5`, and the second write overwrites the first without triggering `StoreError::StaleRevision`.

`finish_run` has the same shape from the terminal side: it unconditionally stores the supplied terminal run, marks the run id terminal, and releases the claim. A stale or divergent terminal write can therefore replace newer stored state and release a live claim without any revision/idempotence check.

Please make both write paths safe: either reject same-revision writes unless the serialized/envelope content is identical, or change the contract so callers must submit `existing.revision + 1` for every state mutation and equal revisions are treated as stale. Add regressions where revision `5` is stored, then a different revision-`5` payload or stale terminal payload is attempted, and verify the stored run and claim state remain unchanged.

### 🔴 Blocking — the CAS claim API cannot atomically preserve both SOP concurrency limits

The existing SOP engine enforces two limits before starting a run: per-SOP `sop.max_concurrent`, by counting active runs whose `sop_name` matches, and global `config.max_concurrent_total`, by counting all active runs. The new admission primitive only takes one `cap`, and the in-memory implementation checks that cap against `g.claims.len()`, which counts every live claim across all SOP names.

That leaves no atomic way to preserve the current semantics when the engine moves onto this store. If the caller passes a per-SOP cap, claims for unrelated SOPs consume each other's slots. If the caller passes the global cap, the per-SOP limit has to stay as a separate non-atomic side check, which is the kind of split admission path this store is meant to remove.

Please make the claim contract encode the full admission policy before this becomes the durable shape. For example, pass both the per-SOP cap and global cap into `try_claim_run`, count existing claims by `sop_name` and in total inside the same store lock/transaction, and add a regression with two SOP names proving one SOP at its cap does not block a different SOP while the global cap still applies.

### 🟡 Warning — the PR body still describes the old risk shape

The live labels are `enhancement`, `size: L`, `risk: high`, and `runtime`, but the PR body still lists `runtime`, `tool:sop`, `size: M`, and `risk: low`; it also does not fill the medium/high-risk rollback fields from the current template. Since this is a runtime SOP contract that future durable behavior will rely on, please update the body after the technical fixes so the label snapshot, risk, validation, compatibility, and rollback sections match the current PR shape.

Nillth:
Thanks for the careful review. All three points are addressed in `0083a2c`; the body is updated to match the live labels/risk and to fill the rollback + compatibility sections.

**Same-revision overwrite (both write paths).** `save_run` and `finish_run` now share one `revision_guard`: a strictly-older revision is rejected `StaleRevision`, an equal revision is accepted only as a byte-identical idempotent retry (else the new `RevisionConflict`), and a newer revision wins. `finish_run` runs the guard before it marks terminal and releases the claim, so a stale or divergent terminal write can no longer clobber newer state or drop a live claim. Regressions cover a divergent revision-5 save and a stale/divergent terminal write (stored run + claim verified unchanged).

**CAS claim could not hold both limits.** `try_claim_run` now takes `per_sop_cap` and `global_cap` and enforces both under one lock: live claims for the SOP are counted by `sop_name` against the per-SOP cap, total live claims against the global cap, mirroring the engine `can_start` (`sop.max_concurrent` + `config.max_concurrent_total`, both inclusive-max, 0 admits nothing). New regression with two SOP names proves one SOP at its per-SOP cap does not block a different SOP while the shared global cap still applies.

Validation: `cargo fmt --check` clean, `cargo clippy --all-targets -D warnings` clean, `cargo test -p zeroclaw-runtime sop::store` 9 passed.

Ready for another look when you have a moment.

Audacity88 (changes requested):
Thanks for the update. I reviewed head `0083a2c` against my prior CHANGES_REQUESTED review, the author follow-up comment, the current diff, the linked #6689/#6037 context, the existing SOP engine concurrency checks, and the green GitHub checks. I did not run local Cargo validation for this re-review.

### ✅ Resolved — revision guards now cover both run write paths

The same-revision overwrite blocker from my prior review is resolved. `save_run` and `finish_run` now share `revision_guard`, equal revisions are accepted only for byte-identical retries, divergent equal revisions return `RevisionConflict`, and the new tests cover both the save path and the terminal-write path while verifying stored state and claim state remain intact after rejected writes.

### ✅ Resolved — admission claims now preserve both concurrency limits

The CAS claim blocker from my prior review is also resolved. `try_claim_run` now receives both `per_sop_cap` and `global_cap`, counts live claims by `sop_name` and in total under the same lock, and the new two-SOP regression proves one SOP at its cap does not block another SOP while the shared global cap still applies.

### 🔴 Blocking — `keep_secs` is a public retention policy knob that `prune` silently ignores

`RetentionPolicy` exposes `keep_secs` and documents it as the age bound for dropping terminal runs, but `InMemoryRunStore::prune` never reads it. The implementation only enforces `max_terminal`: it returns early when the terminal count is under that cap, then sorts terminal runs and drops only the oldest excess records. A caller can pass `RetentionPolicy { max_terminal: 100, keep_secs: Some(1) }` with very old terminal runs and `prune` will return `0` while retaining them all.

Because this PR is landing the store contract that the durable backends will be held to, a documented retention field should not be a silent no-op in the default backend. Please either implement the age-based pruning rule, with a regression where an old terminal run is removed even when the count is under `max_terminal`, or defer/remove `keep_secs` from this scaffold until the PR that implements it.

Nillth:
Fixed in `a78136b787`. `InMemoryRunStore::prune` now honours `keep_secs`: terminal runs whose `completed_at` (falling back to `started_at`) is older than `now - keep_secs` are pruned independently of `max_terminal`, so an aged run is evicted even when the count is under the cap, and the rule also applies when `max_terminal` is unbounded (0). The cutoff is formatted to match `now_iso8601()` so the stored ISO-8601 UTC timestamps compare lexically.

Added a regression (`prune_keep_secs_drops_aged_runs_even_under_max_terminal`) covering both the under-cap case and the unbounded-count case. Thanks for the catch.

Audacity88 (approved):
Thanks @Nillth. I reviewed head a78136b787753c3b95a71dc9a977e72ca55b4993 against my prior CHANGES_REQUESTED review, your latest follow-up comment, the current `sop/store` diff, the linked #6689/#6037 context, the existing SOP timestamp helper, and the green public CI/check status. I did not run local Cargo validation for this review-only pass; I relied on source review, the PR's reported validation, and the green public CI.

### ✅ Resolved — `keep_secs` now participates in pruning

The remaining retention-policy blocker from my prior review is resolved. `InMemoryRunStore::prune` now applies the age bound first, using `completed_at` with `started_at` as the fallback, and removes the run plus its event history when the terminal timestamp is older than the cutoff.

The fix also covers the two cases that mattered for the contract: an aged terminal run is pruned even when the terminal count is still below `max_terminal`, and `max_terminal: 0` now only means the count bound is unbounded, not that age pruning is disabled. The new `prune_keep_secs_drops_aged_runs_even_under_max_terminal` regression pins both paths.

### ✅ Resolved — my prior store-contract blockers are clear

I re-checked the earlier resolved items on the current head as well. The revision guard still covers both `save_run` and `finish_run`, divergent same-revision writes still return `RevisionConflict`, and terminal-write regressions verify stored state plus claim state survive rejected writes. The claim API still takes both `per_sop_cap` and `global_cap`, enforces them under one lock, and keeps per-SOP isolation while preserving the shared global cap.

The PR body now matches the live high-risk runtime label shape and calls out the remaining unwired follow-on work, so I do not have any remaining blocking findings on this scaffold.

PR #8004, opened 2026-06-19
fix(cost): make budget config reloadable instead of frozen at boot

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** The process-global cost tracker held its `CostConfig` inside a single-init `OnceLock<Option<Arc<CostTracker>>>`. The first subsystem to call `get_or_init_global` (gateway, channels, daemon, or agent loop) froze the config for the entire process lifetime; every later call passed a fresh `config.cost` that was silently discarded. Editing `[cost]` limits and hot-reloading had no effect, so the enforced daily/monthly limit could only change with a daemon restart, and which value took effect depended on boot ordering and on-disk config at that boot. This is the root cause of stale-limit budget-exceeded blocks that survived config edits and repeated reloads.
- The config now lives behind `Arc<RwLock<CostConfig>>` inside the tracker, mirroring the existing `RUNTIME_PROXY_CONFIG` reload pattern. Reads go through a `config_snapshot` helper so a swap is observed without a restart. `get_or_init_global` applies the supplied config to the existing tracker on every call instead of dropping it; a new `update_config` performs the swap.
- **Scope boundary:** Only the cost tracker config is made reloadable. Storage path resolution, the ledger format, pricing, and all other call sites are untouched. The full `OnceLock` sweep across the workspace confirmed the cost tracker was the only static holding a reloadable config value frozen behind a single-init `Arc`; every other static is a regex/i18n/capability cache, a `RwLock`/`Mutex`-wrapped mutable, or a function-pointer hook.
- **Blast radius:** All five `get_or_init_global` call sites (gateway/lib.rs, runtime/daemon/mod.rs, runtime/agent/loop_.rs, channels/orchestrator/acp_server.rs, channels/orchestrator/mod.rs) now have their passed config honored rather than discarded after the first caller. Public API only gained `config()` and `update_config()`; the private field type change does not affect callers. The three downstream crates compile clean.
- **Linked issue(s):** none open; net-new defect surfaced from a live deployment.
- **Labels:** `bug`, `risk: medium`, `size: M`, `config`

\`\`\`bash
cargo fmt --all -- --check     # clean (exit 0)
cargo clippy -p zeroclaw-config --all-targets -- -D warnings   # clean (exit 0)
cargo test -p zeroclaw-config --lib cost::
cargo build -p zeroclaw-runtime -p zeroclaw-gateway -p zeroclaw-channels
\`\`\`

  - cost tests: \`test result: ok. 18 passed; 0 failed\` including \`cost_reload_applies_new_daily_limit ... ok\` and \`get_or_init_global_applies_reloaded_config_to_existing_tracker ... ok\`
  - proxy control: \`proxy_reload_applies_new_config_through_rwlock ... ok\`
  - clippy: \`Finished\` with no warnings
  - downstream build: \`Finished dev profile\`, all of runtime/gateway/channels compiled
- **Beyond CI:** Reproduced the defect as a RED test first (boot limit 10.0 frozen, reload to 14000.0 discarded), then made it green with the fix. Verified the global `Arc` is reused (no second tracker constructed). Traced all five call sites to confirm propagation. Did NOT run a full live-deployment reload end-to-end on the remote box; the proof is the unit-level reload contract plus the static sweep.
- **Skipped:** full `cargo test` workspace-wide and `cargo clippy --all-targets` workspace-wide not run here; scoped to the touched crate and its direct dependents for turnaround. Draft pending full battery.

## Compatibility (required)

- Backward compatible? Yes
- Upgrade steps: none. Existing `[cost]` config is read identically; the only behavior change is that edited limits now take effect on reload instead of requiring a restart.

## Rollback (required for risk: medium and high)

- **Fast rollback command/path:** before squash merge, `git revert 396e2a1d791b7ffc381df9163e4f0cf4b5942bd6^..191b4831dd424d2e3f132f96cd626b515fad418c`; after squash merge, revert the squash merge commit.
- **Observable failure symptoms:** budget-exceeded blocks that persist across config edits and daemon reloads; grep logs for \`cost budget check\` and the enforced \`limit_usd\` not matching the on-disk \`[cost]\` limit after a reload

---

## Follow-up commit: enable-after-disabled-boot reload (residual gap from the first commit)

The first commit made limits reloadable while cost tracking is already enabled, but left one case unfixed: the process-global slot was a single-init `OnceLock<Option<Arc<CostTracker>>>`. If the first boot ran with `[cost] enabled = false`, the slot cached `None` permanently and a later reload flipping `enabled` to `true` had no tracker to update and never constructed storage. Enabling cost tracking from off still required a process restart.

- **Fix:** slot is now `OnceLock<RwLock<Option<Arc<CostTracker>>>>`. Resolution moved into `resolve_global`, which double-checks `read()` then `write()` so the steady-state hot-swap stays lock-cheap and a concurrent first-construction race cannot build two trackers. off->on constructs on demand; on->off leaves the tracker resident and relies on `update_config` to neutralise enforcement (no teardown race).
- **Testability:** `resolve_global` takes the slot as a parameter so the global tests drive a local `RwLock` deterministically instead of racing the shared `OnceLock`.
- **Scope/risk:** same single file (`crates/zeroclaw-config/src/cost/tracker.rs`), no call-site signature change, the 5 callers are untouched. `risk: medium`, backward compatible, no config/CLI surface change.

### Validation Evidence (follow-up commit)

```bash
cargo fmt -p zeroclaw-config -- --check                          # clean (exit 0)
cargo clippy -p zeroclaw-config --all-targets                    # clean, Finished
cargo test -p zeroclaw-config --lib cost::                       # 20 passed; 0 failed
cargo build -p zeroclaw-gateway -p zeroclaw-runtime -p zeroclaw-channels   # Finished
```

- New tests: `get_or_init_global_constructs_tracker_when_enabled_after_disabled_boot` (the previously-broken off->on path; same `Arc` reused thereafter), `get_or_init_global_leaves_tracker_resident_when_disabled_on_reload` (on->off keeps the `Arc`, `check_budget` short-circuits to `Allowed`). Retained `get_or_init_global_applies_reloaded_config_to_existing_tracker` ported to `resolve_global`.
- **Decision:** on->off teardown intentionally not implemented; tracker stays resident and `update_config(enabled=false)` neutralises enforcement, which the resident test asserts.

---

## Full workspace validation (draft caveat cleared)

Both commits validated against the entire workspace, not just the touched crate:

```bash
cargo fmt --all -- --check          # clean (FMT_EXIT=0)
cargo clippy --all-targets -- -D warnings   # clean (CLIPPY_EXIT=0), 0 warnings
cargo test                          # TEST_EXIT=0, 891 passed, 0 failed
```

No failing suites anywhere in the workspace. Promoted from draft to ready on this evidence.

---

## Follow-up note: record hot-path snapshot reads remain separate

`record_usage_with_agent` runs on every model response. After the reload swap it took two separate `config_snapshot()` reads, each cloning the entire `CostConfig`, and the two reads were not atomic across a concurrent `update_config`.

- **Decision:** no source change in this PR for that micro-optimization. The functional reload fix remains covered by the earlier regression tests; a single-guard read for `record_usage_with_agent` can be handled separately if needed.


WareWolf-MoonWall (approved):
# Review — PR #8004 `fix(cost): make budget config reloadable instead of frozen at boot`

**Reviewer:** WareWolf-MoonWall · **Head:** `191b4831dd424d2e3f132f96cd626b515fad418c` · **Verdict: approve**

This is a well-reasoned fix for a real SSOT violation: `config: CostConfig` was stored as a value inside `CostTracker`, freezing the budget limits at first-caller boot and silently discarding every subsequent `get_or_init_global` config argument. The fix is architecturally correct — `config` moves to `Arc<RwLock<CostConfig>>`, `resolve_global` adds a double-checked construction pattern so the global `Arc` is never duplicated, and the off→on reload path is covered in the follow-up commit. Full workspace validation (891 tests, 0 failures) is documented before draft promotion. All CI checks pass. Two warnings below; neither is a functional blocker.

---

### 🟢 Praise — SSOT-compliant pattern, mirrors existing proxy reload

Moving `config` into `Arc<RwLock<CostConfig>>` and providing `update_config` / `config_snapshot` is exactly the pattern AGENTS.md permits ("Resolver closures / materialized views built ON-DEMAND from canonical state"). The implementation mirrors the existing `RUNTIME_PROXY_CONFIG` reload pattern already in `schema.rs`, making this consistent across the codebase rather than introducing a novel pattern.

---

### 🟢 Praise — Double-checked locking in `resolve_global` is correct

The read-fast-path → write-guard double-check correctly prevents a concurrent double-construction race: the inner `if let Some(ct) = guard.as_ref()` check under the write lock ensures that only one tracker is ever constructed even if two callers race past the read guard simultaneously. This is the canonical safe implementation for this idiom.

---

### 🟢 Praise — Off→on reload path covered with dedicated regression test

`get_or_init_global_constructs_tracker_when_enabled_after_disabled_boot` is the exact previously-broken path (slot cached `None` permanently), and `get_or_init_global_leaves_tracker_resident_when_disabled_on_reload` closes the on→off teardown-race concern. Using a local `RwLock` parameter in `resolve_global` for test isolation (instead of racing the global `OnceLock`) is the correct testability pattern.

---

### 🟢 Praise — Full workspace validation before draft promotion

Clearing the draft caveat with workspace-wide `cargo fmt / clippy / test` (891 passed, 0 failed) is exactly right. The earlier per-crate evidence for each incremental commit is also clearly structured. Good hygiene.

---

### 🟡 Warning — 4th commit description does not match the actual cumulative diff

The 4th commit section in the PR description states:

> `record_usage_with_agent` runs on every model response … Fix: read `enabled` and `track_per_agent` once under a single `RwLock` guard without cloning the config. The hot path is now allocation-free.

The cumulative diff tells a different story. `check_budget` received the single-snapshot consolidation (`let config = self.config_snapshot()` at the top, all subsequent field accesses through `config.`). However, `record_usage_with_agent` in the final diff still has **two separate** `self.config_snapshot()` calls — one for the `enabled` early-return guard and one for the `track_per_agent` branch — each cloning the full `CostConfig`. These two reads are not atomic across a concurrent `update_config`. The functional impact is minimal (a config reload between the two calls could produce one record with a stale `track_per_agent` flag, which is benign), but the description is misleading and the claimed "allocation-free" hot path is not realised.

**Requested action:** Either (a) update the 4th commit section to accurately describe what was done (single-snapshot consolidation in `check_budget`, two snapshot calls remain in `record_usage_with_agent`), or (b) ship a fixup commit to `record_usage_with_agent` that reads both flags under one guard:

```rust
// read once under one guard; both flags consistent, no second clone
let (enabled, track_per_agent) = {
    let cfg = self.config.read();
    (cfg.enabled, cfg.track_per_agent)
};
if !enabled { return Ok(()); }
// …
let effective_alias = if track_per_agent { … } else { None };
```

---

### 🟡 Warning — Rollback SHA is an unfilled placeholder

The Rollback section reads:

```
Fast rollback command/path: `git revert `
```

The commit SHA is missing. For a `risk: medium` PR the template requires a copy-pasteable recovery command. Please fill in the actual HEAD SHA (`191b4831dd424d2e3f132f96cd626b515fad418c`) or the range if squash-merging, so operators have a literal command to run under incident pressure.

Audacity88 (dismissed):
I reviewed current head d0e0377 against the live PR state, the PR body, the existing approval review, the green visible GitHub checks, and the current diff in `crates/zeroclaw-config/src/cost/tracker.rs` and `crates/zeroclaw-config/src/schema.rs`. I did not run local Cargo for this review-only pass.

### 🟢 What looks good — the global tracker now follows reloads

The main direction is right: moving the cost config behind a reloadable `RwLock`, updating the existing global tracker instead of discarding later configs, and constructing a tracker on an off-to-on reload all address the frozen-config bug without changing the public config surface. The added tests cover the main reload paths, and the current visible CI is green.

### 🔴 Blocking — record usage should read one config generation

`record_usage_with_agent` still reads the reloadable config twice: once for `enabled` and later again for `track_per_agent`. Because `update_config` can swap the config between those two reads, one usage record can observe `enabled` from one config generation and `track_per_agent` from another. That leaves a small but real inconsistency in the same hot path this PR is trying to make reload-safe.

Please patch this path in `crates/zeroclaw-config/src/cost/tracker.rs` so `record_usage_with_agent` reads the cost config once under one `RwLock` guard, or clones one local snapshot once, and then uses that same value for both `enabled` and `track_per_agent`. The important part is that a single usage event makes its record/drop and per-agent attribution decisions from the same config generation. The PR body already calls out the separate reads and frames the fix as a deferrable micro-optimization, but this is a reload-consistency gap in the same hot path this PR is making reloadable. This should land in this PR rather than as a later cleanup because it is directly tied to the reload contract being fixed here.

singlerider:
Addressed at head `97cdbd6623`.

🔴 **`record_usage_with_agent` reads config twice** — RESOLVED. The function now snapshots the cost config once at the top (`let config = self.config_snapshot();`) and drives both the `enabled` record/drop decision and the `track_per_agent` attribution decision from that single binding. A concurrent `update_config` can no longer tear a single usage event across two generations. There is now exactly one `config_snapshot()` call in the function body.

Proof:
- `cargo test -p zeroclaw-config cost` → 24 passed; 0 failed, including the new `record_usage_reads_one_config_generation`.
- `cargo clippy -p zeroclaw-config --all-targets -- -D warnings` clean.
- `cargo fmt --all -- --check` clean.
- `cargo build -p zeroclaw-runtime` clean (the single downstream caller compiles unchanged).

Also merged latest `master` into the branch. Re-requesting review.

singlerider:
Moving to v0.8.2 as it's a high-value bug fix.

PR #8014, opened 2026-06-19
fix(runtime): stop duplicating streamed narration before native tool calls

singlerider:
## Summary
- **Base branch:** `master`
- **What changed and why:** Native tool-call providers (Anthropic shape) emit narration text in the same message as a tool_use block. The stream consumer forwards that narration live on the draft/event sinks, then flips into suppression when the tool call arrives. The tool loop re-sent the whole narration afterward, so draft channels showed it twice. A naive "skip the relay if anything streamed live" fix introduced the opposite defect: when the stream text guard withholds a narration tail as an incomplete protocol prefix, the following tool call drops that buffered tail from the live stream, and the skipped relay never recovers it, truncating the narration.
- **The fix:** The consumer now records the exact visible text it forwarded to draft/event sinks (`forwarded_visible_text`) as the single source of truth. That value is threaded through `ProviderCallOutcome`, and the loop relays only `display_text` beyond the already-forwarded prefix. Narration is delivered exactly once and never truncated, on every surface that funnels through the shared turn loop. The consumer's repeated per-sink forward blocks were collapsed into one macro so the accounting has a single home.
- **Scope boundary:** Does NOT change the final-response path, the non-streamed delivery path, or stream consumption semantics for non-narration turns. Only corrects the partial re-send of intermediate narration when a native tool call follows.
- **Blast radius:** Draft/event rendering for native tool-call turns that include narration. No effect on providers that do not interleave narration with tool calls.

## Validation Evidence
```
cargo fmt --all -- --check                                      -> clean
cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings   -> clean (0)
cargo test -p zeroclaw-runtime --lib -- --test-threads=1        -> 2299 passed; 0 failed; 2 ignored
```
- **RED proofs (both directions):**
  - Pre-fix unconditional re-send: `run_tool_call_loop_does_not_duplicate_streamed_narration_before_native_tool_call` fails `left: 2, right: 1` (narration doubled).
  - Naive skip-if-streamed-live: `run_tool_call_loop_preserves_guard_withheld_narration_tail_before_tool_call` fails (guard-withheld `<tool` tail dropped).
  - Only the remainder-based fix passes both.
- **Pre-existing flake disclosure:** `cron::store::tests::remove_job_emits_structured_cron_delete_event` fails under parallel test threads on a global log-subscriber race; it fails identically on the parent commit and passes in isolation and single-threaded. Not introduced here.
- A live `anthropic.personal_code` probe was used locally to confirm the original duplicate against the real provider. It was committed by mistake on an earlier push and has been removed in this branch; it read the real `~/.zeroclaw` config and built a billed provider and must never ship.

## Security & Privacy Impact
- New permissions/file access? No
- PII in diff/tests/fixtures/docs? No

## Compatibility
- Backward compatible? Yes
- Config/env/CLI surface changed? No

## Rollback
- **Observable failure symptoms:** narration text appears twice in draft-channel messages immediately before native tool-call turns; look for back-to-back identical `StreamDelta::Text` bursts carrying the same narration string on the `on_delta` channel. The pre-fix cron flake (`cron::store::tests::remove_job_emits_structured_cron_delete_event`) is unrelated and pre-existing.


WareWolf-MoonWall (dismissed):
Reviewed at head 327dfb8b92d92699701719ed3d9abe91878183b6.

This is a well-engineered fix for a genuinely subtle bug. The invariant that was missing — tracking exactly which text the stream consumer forwarded to the draft/event sinks before the tool-call suppression kicked in — is now the single source of truth via `forwarded_visible_text`. The dual-direction RED proofs (pre-fix doubles, naive-skip truncates, remainder-based fix alone passes both) are exactly the right validation discipline for a streaming-path fix. The `forward_visible!` macro eliminates eight repetition sites and puts all forwarding accounting in one place. CI passes clean across all 18 checks. One structural gap in the PR template for `risk: high` blocks merge; everything else here is exemplary.

### 🔴 Blocking — Rollback section uses the low-risk format for a `risk: high` PR

The Rollback section currently reads `• git revert <sha>.` — which is the template's low-risk one-liner. `risk: high` requires all three sub-bullets. The answers are trivial but the section must be filled for template completeness:

- **Fast rollback command/path:** `git revert <sha>` (same)
- **Observable failure symptoms:** narration text appears twice in draft-channel messages immediately before native tool-call turns; look for back-to-back identical `StreamDelta::Text` bursts carrying the same narration string on the `on_delta` channel. The pre-fix cron flake (`cron::store::tests::remove_job_emits_structured_cron_delete_event`) is unrelated and pre-existing.

Two lines to add; please push a fixup commit.

### 🟡 Warning — `!protocol_suppressed` relay gate is underdocumented

The new `&& !protocol_suppressed` condition on the `display_text` relay at line ~665 of `loop_.rs` correctly skips relay when the stream consumer's text guard proactively suppressed text (i.e., the guard recognized and withheld actual XML-protocol output, not just a partial prefix). This is the right behaviour — the guard-suppressed text was intentionally withheld and the non-streamed delivery path is the only consumer. However, the code comment says "relay only the portion the live stream did not already deliver" without calling out the `protocol_suppressed = true` + native-tool-calls branch, where `display_text` may contain narration that neither the live stream nor the relay will deliver. A one-line inline note clarifying that the `!protocol_suppressed` gate handles the guard-suppressed case (and is distinct from the guard-withheld-prefix case covered by `unforwarded_narration`) would help future readers distinguish the two paths.

### 🟢 What looks good — dual RED proofs in both failure directions

The test names and docstrings are a first-class specification:
- `run_tool_call_loop_does_not_duplicate_streamed_narration_before_native_tool_call`: pre-fix sends the whole narration twice; fix reduces to one.
- `run_tool_call_loop_preserves_guard_withheld_narration_tail_before_tool_call`: pre-fix with the naive skip drops the `<tool`-buffered tail; remainder-based fix recovers it.

The proofs cover both failure modes the PR body describes. The `assert_eq!(accumulated.matches(narration).count(), 1, ...)` assertion with a detailed failure message is the right shape — it catches duplication without being brittle to `String` interning.

### 🟢 What looks good — `forward_visible!` macro as a single forwarding home

Consolidating 8+ `if let Some(tx) = event_tx { ... } if let Some(tx) = delta_sender { ... }` blocks into one macro with a `$count_visible` token gives the `forwarded_visible_text` accumulation a single owner. The `true`/`false` distinction correctly separates "forwarded during live streaming" (counts toward `forwarded_text` for interruption tracking) from "flushed post-stream" (reaches sinks but not the interruption partial). The final `let _ = delta_sender;` suppresses the dead-variable lint without losing the early-return semantics.

### 🟢 What looks good — `unforwarded_narration` fallback preserves the "never truncate" invariant

`display_text.strip_prefix(streamed_visible_text).unwrap_or(display_text)` correctly errs toward relaying the whole narration on any divergence rather than silently truncating. The PR body documents this as a deliberate design decision ("duplicate is recoverable noise; a dropped tail is permanent loss"), and the unit test `relays_whole_text_on_prefix_divergence_rather_than_truncating` makes it executable. Four clean unit tests covering prefix-match, full-match, empty-stream, and divergence fully specify the function's contract.

### 🟢 What looks good — transparency on the accidental probe commit

Explicitly disclosing that a live `anthropic.personal_code` probe was accidentally pushed and then removed — naming exactly what it accessed and why it must never ship — is the right call. The current diff is clean of that code. Worth confirming via the CI secret-scan job, which shows green.

Nillth (commented):
Reviewed at head `327dfb8b92d92699701719ed3d9abe91878183b6`. I read the full diff and traced the `forwarded_visible_text` accounting end to end across `stream_consume.rs` -> `ProviderCallOutcome` -> the relay in `run_tool_call_loop`, plus both RED proofs. This is a clean, well-validated fix for a genuinely subtle streaming defect, and I agree with WareWolf's read of it. Adding an independent corroboration and one forward-looking note rather than re-raising the open items.

### ✅ Resolved — Rollback section now complete for `risk: high`

WareWolf's blocker was the Rollback section using the low-risk one-liner. The current PR body (edited after that review; the head commit is unchanged) now carries all three sub-bullets — fast rollback, "Feature flags/toggles: None", and the observable-symptoms description down to the back-to-back `StreamDelta::Text` bursts on `on_delta`. From my side that finding reads as addressed at the description level; it just needs WareWolf's re-review to clear the formal block.

### 🟢 What looks good — the remainder-relay accounting is correct end to end

I specifically tried to break the dedup and could not. `forward_visible!` records `forwarded_visible_text` whenever either sink is live, and both sinks receive the identical `visible` payload, so the value the loop later subtracts against is exactly what `on_delta` saw live. The `@count true/false` split is right: only the in-loop live forwards feed `forwarded_text`/`visible_event_output` (interruption-partial tracking), while the post-stream `text_guard` flushes reach the sinks and the `forwarded_visible_text` ledger but correctly do not move the interruption partial. The `unforwarded_narration` helper being a total function — suffix on a genuine prefix, whole text on any divergence — is the right call, and erring toward a recoverable duplicate over a permanent dropped tail is the correct risk posture for a draft surface. The two integration proofs pin both failure directions, and `accumulated.matches(narration).count() == 1` is a sturdy, non-brittle assertion shape.

### 🔵 Suggestion — pin the "streamed text is a byte-exact prefix of `display_text`" invariant

The whole design rests on `forwarded_visible_text` (sum of sanitized stream deltas) being a byte-exact prefix of the final `display_text` (from `resolve_display_text`). When that holds, you relay the clean remainder; when it diverges by even one byte of normalization (whitespace, a newline, a trim), `unforwarded_narration` silently falls back to re-sending the whole narration — i.e. the original duplicate, just downgraded to "recoverable noise". That tradeoff is the right default, but the silent fallback also means a future divergence between the stream guard's sanitization and `resolve_display_text` would regress dedup with no test catching it. The live `anthropic.personal_code` probe covered today's shape; a small unit/golden test that locks the prefix relationship for the real Anthropic narration-then-`tool_use` payload would turn that runtime check into a standing guard. Non-blocking.

### 🔵 Concur — WareWolf's note on the `!protocol_suppressed` branch

Agreeing rather than re-raising: the new `&& !protocol_suppressed` gate is correct (guard-suppressed protocol text is intentionally withheld and the final-response path is the only legitimate consumer there), but it is a behavior change from the pre-fix unconditional relay, so the one-line inline note WareWolf asked for is worth adding while the file is open — it saves the next reader from re-deriving why the `protocol_suppressed` + native-tool-call path deliberately relays nothing to the draft.

Verdict: not blocking from me. Code and validation are in good shape; the one formal blocker is addressed in the updated body and is WareWolf's to clear on re-review.

Audacity88:
@singlerider quick CI nudge: the current head is failing before tests can run because the two new `ToolLoop` test initializers are missing the newer `ingress` field.

```text
error[E0063]: missing field `ingress` in initializer of `agent::turn::ToolLoop<'_>`
crates/zeroclaw-runtime/src/agent/loop_.rs:9239:41
crates/zeroclaw-runtime/src/agent/loop_.rs:9330:41
```

That is what is taking down both `Lint` and `Test`, with `CI Required Gate` failing as the aggregate. Once those two test builders include the field and CI is green again, this should be ready for a proper re-review.

WareWolf-MoonWall (dismissed):
# Re-review — #8014 `fix(runtime): stop duplicating streamed narration before native tool calls`

**Prior head:** `327dfb8b` (Round 20 — 🔴 CR, auto-dismissed on push)
**Current head:** post-`327dfb8b` (new commits with integration tests; see CI)
**Verdict:** `--request-changes`

---

## Overview

New commits were pushed after the Round 20 review, adding two integration-test probes for the dual-direction failure modes (`NarrationThenToolCall` variant in the mock provider + full `ToolLoop` invocations). The Rollback section was updated in the PR body — that prior blocker is resolved. However, both new test initializers are missing the `ingress` field that was added to `ToolLoop` in the interim, causing Lint and Test to fail before any test can run. All 15 non-test checks pass cleanly; only the compilation-gate jobs are red.

---

### ✅ Resolved — Rollback section now carries all three `risk: high` sub-bullets

The PR body Rollback section now reads:

- **Observable failure symptoms:** narration text appears twice in draft-channel messages immediately before native tool-call turns; look for back-to-back identical `StreamDelta::Text` bursts carrying the same narration string on the `on_delta` channel

All three sub-bullets are present. The Round 20 template blocker is closed.

---

### 🔴 Blocking — Both `ToolLoop` integration test initializers are missing the `ingress` field

The two new tests —

- `run_tool_call_loop_does_not_duplicate_streamed_narration_before_native_tool_call`
- `run_tool_call_loop_preserves_guard_withheld_narration_tail_before_tool_call`

— each construct a `ToolLoop { ... }` struct literal. Neither includes the `ingress` field that was added to `ToolLoop` between `327dfb8b` and the current head. Audacity88's CI comment confirms the exact errors:

```
error[E0063]: missing field `ingress` in initializer of `agent::turn::ToolLoop<'_>`
crates/zeroclaw-runtime/src/agent/loop_.rs:9239:41
crates/zeroclaw-runtime/src/agent/loop_.rs:9330:41
```

This takes down both Lint and Test, which cascades to fail the CI Required Gate. The fix is mechanical — add `ingress: <default_or_none>` to both struct initializers (matching however other nearby tests initialize it). Please push a fixup commit; I will re-review once CI is green.

---

### 🟡 Still open — `!protocol_suppressed` relay gate is underdocumented

This was a 🟡 in Round 20 and remains unaddressed in the current diff. The `&& !protocol_suppressed` condition at the `display_text` relay in `loop_.rs` is correct, but the surrounding comment says only "relay only the portion the live stream did not already deliver" — it does not distinguish the guard-suppressed case (where `display_text` may contain narration that neither the live stream nor the relay will deliver) from the guard-withheld-prefix case covered by `unforwarded_narration`. A one-line inline note while the file is open would save the next reader from re-deriving why the `protocol_suppressed` + native-tool-call path deliberately relays nothing to the draft.

---

### 🟢 What looks good — integration test design is sound once compilation is fixed

The `NarrationThenToolCall` mock variant correctly emits narration chunks as `StreamEvent::TextDelta` and then a `StreamEvent::ToolCall`, which is the exact stream shape the fix targets. Both test bodies drain `rx` and assert `accumulated.matches(narration).count() == 1` — the same sturdy, non-brittle assertion shape from the design discussion. The second test's `<tool` suffix in the narration exercises the guard-withheld-prefix path specifically. Once the `ingress` field is added and CI goes green, these proofs will be first-class coverage for both failure modes.

---

### 🟢 What looks good — `forward_visible!`, `unforwarded_narration`, core fix (unchanged from Round 20)

The underlying fix (`forwarded_visible_text` as single source of truth, `unforwarded_narration` total function, `forward_visible!` macro) is unchanged and remains exemplary. The four unit tests for `unforwarded_narration` (prefix-match, full-match, empty-stream, divergence) are still in the diff and still correct.

singlerider:
Addressed at head `912569a29e`.

🔴 **Both ToolLoop test initializers missing `ingress`** — RESOLVED. Merged `master` into the branch (which is where the `ingress` envelope field landed) and stamped both new tests (`run_tool_call_loop_does_not_duplicate_streamed_narration_before_native_tool_call`, `run_tool_call_loop_preserves_guard_withheld_narration_tail_before_tool_call`) with `ingress: IngressContext::internal()`, matching the other in-crate fixtures. `cargo test -p zeroclaw-runtime` and `cargo clippy --all-targets -- -D warnings` both green locally.

🟡 **`!protocol_suppressed` relay gate underdocumented** — RESOLVED in `fa33d5c0c2` (docs commit). The inline note at the `display_text` relay now distinguishes the `protocol_suppressed` whole-turn withhold from the guard-passed empty-remainder skip.

Re-requesting review.

WareWolf-MoonWall (approved):
# Re-review — #8014 `fix(runtime): stop duplicating streamed narration before native tool calls`

**Prior head:** `fa33d5c0` (Round 21 — 🔴 CR, auto-dismissed on push)
**Current head:** `912569a29e01be232137a9408bfcf42ba34f87b9`
**Verdict:** `--approve`

---

## Overview

Re-review at head `912569a29e`. The Round 21 blocker — both `ToolLoop` test initializers missing `ingress` — is resolved. The Round 21 `!protocol_suppressed` relay-gate warning (carried forward from Round 20) is also addressed. CI is fully green across all 18 checks. No active CHANGES_REQUESTED from any other reviewer (Nillth: COMMENTED only; Audacity88: Requested, no formal review posted). This clears for approval.

---

### ✅ Resolved — Both `ToolLoop` test initializers now include `ingress: IngressContext::internal()`

The diff confirms `ingress: IngressContext::internal()` is present in both new tests:

- `run_tool_call_loop_does_not_duplicate_streamed_narration_before_native_tool_call` (diff line 101)
- `run_tool_call_loop_preserves_guard_withheld_narration_tail_before_tool_call` (diff line 193)

Both match the pattern used by other in-crate `ToolLoop` fixtures. `cargo clippy` and `cargo test` both pass clean.

### ✅ Resolved — `!protocol_suppressed` relay gate now has an inline explanatory comment

The Round 21 warning asked for a note distinguishing the guard-suppressed case from the guard-withheld-prefix case. The new comment at the relay site reads:

```
// `protocol_suppressed` withholds the whole turn; the empty-remainder
// skip below handles the guard-passed case where the live stream already forwarded every byte.
```

This correctly separates the two branches for the next reader.

---

### 🟢 Integration test probes are sound

The `NarrationThenToolCall` mock variant emits narration chunks as `StreamEvent::TextDelta` and then a `StreamEvent::ToolCall` — the exact stream shape the fix targets. Both test bodies drain the `rx` channel and assert:

```rust
assert_eq!(accumulated.matches(narration).count(), 1, ...)
```

This is a sturdy, non-brittle assertion shape that catches duplication without being sensitive to `String` interning or ordering. The second test's `"Checking <tool"` narration specifically exercises the guard-withheld-prefix path, covering the second failure mode described in the PR body.

### 🟢 Core fix unchanged and correct (carried from prior rounds)

The `forwarded_visible_text` field as single source of truth, `unforwarded_narration` as a total function (suffix on genuine prefix, whole text on any divergence — never truncates), the `forward_visible!` macro as the single forwarding home, and the four unit tests for all contract branches remain unchanged. The dual RED proofs verify both pre-fix failure directions; only the remainder-based fix passes both.

### 🟢 CI green across all 18 checks

Format, Lint, Test, CI Required Gate, Build, Check (all features / 32-bit / no-default / aarch64 / windows), Benchmarks Compile, Installer Drift, Security, Nix Module Eval, Zerocode RPC Boundary, Docs Style — all pass.

PR #8018, opened 2026-06-19
fix(gateway): make agent rename re-issue converge after post-persist partial failure (#7907)

### LINKED ISSUE #7907: [Bug]: agent rename can move owned state before config persistence

wangmiao0668000666:
## Summary

- Base: `master` (commit `74a806943`)
- Problem: PR #7940 (Nillth) fixed the rename split-brain by reordering persist-first. Reviewers (Audacity88 / WareWolf-MoonWall / singlerider) all flagged one residual blocker: the new WARN log told the operator to "re-issue the rename to converge", but the API rejected the retry with `RenameError::NotFound` (or `InvalidName "alias \`to\` already exists"`, depending on which side of the commit) before any side-effect could run. There was no working recovery path.
- Why: `rename_map_key` is the in-memory rewrite + dirty-marker that the persist-first contract relies on. When the source alias is already gone from the committed config (post-persist) OR the destination alias is already present (same state from the other side), the rewrite refuses. The handler then returns 4xx and the side-effects never run.
- What changed: at the API entry of `rename_agent_cascade`, detect the post-persist split-brain state (`from`-absent + `to`-present in `working.agents`). On that state, short-circuit `rename_with_cascade` with an empty `RenameReport`, adopt the committed config from `state.config`, and proceed straight to the side-effects (workspace move + `cascade_rename_agent`). The second call now converges the split-brain deterministically. `resumed: bool` is added to the WARN/INFO log attrs.
- Out of scope: the `MapKeyResponse::warnings` field on the delete path (PR #8017, #7941) is a sibling fix, separate PR.

## Label Snapshot
- Risk: medium
- Size: S
- Scope: gateway
- Module: config / agents
- Contributor tier: trusted (5+ PRs once this lands)

## Change Metadata
- Type: fix
- Primary scope: crates/zeroclaw-gateway/src/

## Linked Issue
Closes #7907
Related: #7941 (the delete-direction mirror; PR #8017 fixes it)

## Supersede Attribution
Built on top of Nillth's `b9bb21877` (cherry-picked into this branch, author amended to mine). Nillth's original commit remains the bulk of the work; this PR adds the resume-aware path + a third regression test that the reviewers asked for. If the maintainer prefers one combined PR, I can squash on request.

## Validation Evidence
```bash
$ cargo fmt --all -- --check
(no output — clean)
$ cargo clippy --locked -p zeroclaw-gateway --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 13.47s
$ cargo test --locked -p zeroclaw-gateway --lib
test result: ok. 266 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
$ cargo test --locked -p zeroclaw-gateway --lib api_config::tests::agent_rename
test api_config::tests::agent_rename_leaves_owned_state_put_when_persist_fails ... ok
test api_config::tests::agent_rename_reissue_converges_post_persist_partial_failure ... ok
test api_config::tests::agent_rename_moves_owned_state_after_successful_persist ... ok
```

## Real behavior proof

- **Behavior addressed**: a second `POST /api/config/rename-map-key` request with the same `from`/`to` after a post-persist partial failure must now (a) succeed (200 OK), not 4xx, and (b) converge the split-brain: the workspace and cron row move from `from` to `to` on the second call.
- **Real environment tested**: Linux x86_64, Rust 1.94, zeroclaw 0.8.1 / master @ `74a806943`.
- **Exact steps run after this patch**:
  - `cargo test --locked -p zeroclaw-gateway --lib api_config::tests::agent_rename` (3 tests, all green)
  - `cargo test --locked -p zeroclaw-gateway --lib` (full crate lib tests, 266 passed)
  - `cargo clippy --locked -p zeroclaw-gateway --all-targets -- -D warnings` (clean)
- **Evidence after fix**:
  ```
  test api_config::tests::agent_rename_leaves_owned_state_put_when_persist_fails ... ok
  test api_config::tests::agent_rename_reissue_converges_post_persist_partial_failure ... ok
  test api_config::tests::agent_rename_moves_owned_state_after_successful_persist ... ok
  ```
  The new test seeds the exact post-persist split-brain state that the reviewers described (config durably `to`, workspace + cron still at `from`), then re-issues the rename and asserts the second call (a) returns 200 OK, (b) moves the workspace to `to`, (c) re-points the cron row to `to`, (d) leaves `from` empty. Uses real `rusqlite` cron probes (not mocks).
- **Observed result after fix**: the second call converges the split-brain. The `resumed: true` attr in the INFO/WARN log line lets operators distinguish first-call from resume-call paths.
- **What was not tested**: the test suite does not exercise the actual HTTP handler endpoint — it calls the `rename_agent_cascade` function directly with a constructed `AppState`. The HTTP layer is a thin axum wrapper; the cascade logic the reviewers called out is fully covered.

## Security Impact
- Permissions: no new permission checks. `require_auth` upstream is unchanged.
- Network calls: none added.
- Secrets: none touched.
- File access: workspace move + owned-DB cascade are the same operations as before, just now also reachable via the resume path. The resume path does NOT skip these operations; it re-runs them against the committed config.

## Privacy and Data Hygiene
- No new data is written outside the existing workspace-rename + owned-DB-rename paths.
- The `resumed` boolean is logged at INFO (or WARN if any side-effect fails) — it does not surface in the response body, so it is not externally observable.

## Compatibility / Migration
- Response shape: unchanged. The handler still returns `RenameMapKeyResponse { path, from, to, renamed: true, warnings }` — the resume path sets `renamed: true` (the config IS renamed, just on a prior call) and an empty `warnings` once the side-effects converge. The new `resumed` attr is internal to the log line.
- API contract: the resume path returns 200 OK where the prior version would have returned 4xx. This is strictly an improvement (the operator's documented recovery path now actually works).
- Migration: none required for clients.

## i18n Follow-Through
No user-visible string changes.

## Human Verification
- Ran `cargo test --locked -p zeroclaw-gateway --lib` locally — 266 passed, 0 failed.
- Inspected the diff manually: the resume-detection block sits BEFORE the `rename_with_cascade` call. When triggered, it (a) replaces `working` with the post-swap config from `state.config`, (b) constructs an empty `RenameReport`, (c) sets `resumed = true`, and (d) skips the rewrite. The subsequent `persist_and_swap` is a no-op write of the same state. The side-effect block then runs unchanged.

- The `resumed` field is added to the JSON log attrs (not to the response). Operators scraping the JSON log lines will see a new key. The `tracing`/`zeroclaw_log` schemas allow this.
- The persist-and-swap call on the resume path is a no-op write of the same state. It is intentionally kept (instead of being short-circuited) so the success / failure branch downstream of `persist_and_swap` doesn't need to special-case `resumed`.

## Agent Collaboration Notes
- This PR is built on top of Nillth's `b9bb21877` (cherry-picked + author-amended). If the maintainer prefers a single combined PR (Nillth's reorder + this resume path + the regression test), I can squash on request — but I left them separate so the resume-path commit is clearly attributable to this PR.
- The `test_state` visibility change in `api.rs` (`pub(crate) use tests::test_state;` + `pub(crate) fn test_state`) was already part of Nillth's `b9bb21877` and is preserved in the parent commit. PR #8017 (#7941, delete cascade) also adds the same visibility change. If both PRs land, the second one's visibility change is a no-op (the symbol is already `pub(crate)`); rebase is trivial.

## Rollback Plan
- Revert the resume commit (`93bf326f9`). The parent commit (`6dd265a2e`, Nillth's reorder) and the prior `2eb0059ae` form the same state as the current PR #7940 — i.e. reverting this PR returns to "persist-first but no recovery path", which is the state PR #7940 is in today.
- No feature flag involved.

## Risks and Mitigations
- **Risk**: the resume detection could trigger on a *legitimate* rename where the operator created `to` in parallel (config has both `from` AND `to` via a different code path). Mitigation: the detection only triggers when `from` is absent. A normal "rename" request never has `from` absent in the input config; the only way to reach the detection is via a re-issue after a prior successful commit. Confirmed by reading all `working.agents.insert(...)` / `rename_map_key` call sites — none of them produce the `(from-absent, to-present)` shape from a fresh state.
- **Risk**: the resume path silently succeeds even if the operator wanted to re-issue a different rename (e.g. typos `to` in the second call). Mitigation: the second call's `to` is what the operator typed. The `renamed: true` response means "config is `to`" (true on both fresh and resume paths). The `resumed` log attr is the operator's signal that this was a convergence call, not a fresh rename.
- **Risk**: the resume path re-runs the owned-DB cascade on a store that has already been updated by a prior partial-failure call. Mitigation: `cascade_rename_agent` uses `WHERE agent_alias = from` re-points; if `from` has no rows (prior call succeeded), it is a no-op. Same for the workspace move (`if !old_ws.exists() { return None; }`). Both side-effects are idempotent against the split-brain state.

### DIFF
--- a/crates/zeroclaw-gateway/src/api.rs
+++ b/crates/zeroclaw-gateway/src/api.rs
@@ -1910,6 +1910,11 @@ pub async fn handle_claude_code_hook(
     Json(serde_json::json!({ "ok": true }))
 }
 
+// Shared test helper: `api_config` tests reuse this AppState builder for the
+// agent rename/delete cascade handlers (#7907 regression coverage).
+#[cfg(test)]
+pub(crate) use tests::test_state;
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -2075,7 +2080,7 @@ mod tests {
         config
     }
 
-    fn test_state(config: zeroclaw_config::schema::Config) -> AppState {
+    pub(crate) fn test_state(config: zeroclaw_config::schema::Config) -> AppState {
         AppState {
             config: Arc::new(RwLock::new(config)),
             model_provider: Arc::new(MockModelProvider),
--- a/crates/zeroclaw-gateway/src/api_config.rs
+++ b/crates/zeroclaw-gateway/src/api_config.rs
@@ -1562,22 +1562,44 @@ async fn rename_agent_cascade(
     for path in &report.dirty_paths {
         working.mark_dirty(path);
     }
+    // Compute the NEW workspace path off the rewritten config now, before
+    // `working` is moved into `persist_and_swap` below. Both paths are owned
+    // `PathBuf`s that outlive the move.
+    let new_ws = working.agent_workspace_dir(to);
+    let dirty_count = report.dirty_paths.len();
+
+    // Persist the config rename FIRST (#7907). Previously the workspace move and
+    // owned-state cascade ran *before* this durable write, so a persist failure
+    // after them left config still naming `from` while the workspace + owned
+    // stores had already moved to `to` — the inverse split-brain of what #7841
+    // fixed. Persisting first means an early failure here leaves config,
+    // workspace, and owned state all consistently on `from`: a clean abort
+    // (`persist_and_swap` reverts the on-disk file and never swaps `state.config`).
+    if let Err(e) = persist_and_swap(state, working).await {
+        return error_response(e);
+    }
+
+    // Config is now durably `to` and authoritative in `state.config`. Run the
+    // external side-effects against the committed config (read a short-lived
+    // clone — never hold the lock guard across an `.await`). Each side-effect is
+    // best-effort and surfaced as a warning; persist-first makes them all
+    // idempotently re-runnable from the corrected config (the workspace move
+    // early-returns once the source is gone; every owned-store op re-points by
+    // `WHERE agent_alias = from`), so re-issuing the same rename converges.
+    let cfg = state.config.read().clone();
 
     // Move the workspace dir. For the default per-alias location this is
     // `<install>/agents/<from>/workspace` → `…/<to>/workspace`. A custom
     // workspace path is alias-independent, so `old_ws == new_ws` and we skip.
-    // A failed move is surfaced (like the owned-DB failures below), not just
-    // logged — otherwise config+DB point at `to` while the workspace is stranded
-    // at `from` and the caller sees a clean success.
-    let new_ws = working.agent_workspace_dir(to);
+    let ws_existed = old_ws != new_ws && old_ws.exists();
     let move_warning = move_renamed_workspace(&old_ws, &new_ws).await;
-    let workspace_moved = old_ws != new_ws && old_ws.exists() && move_warning.is_none();
+    let workspace_moved = ws_existed && move_warning.is_none();
     let mut warnings: Vec<String> = Vec::new();
     warnings.extend(move_warning);
 
     // Re-point owned DB state (memory/cron/acp/session). Best-effort + reported.
     let owned = crate::agent_owned_state::cascade_rename_agent(
-        &working,
+        &cfg,
         &state.mem,
         state.session_backend.as_ref(),
         from,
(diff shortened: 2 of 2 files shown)


wangmiao0668000666:
@Audacity88 ready for review when you have a moment. CI is green (19/19) on the latest head, and the diff makes agent rename re-issue converge after a post-persist partial failure (fix #7907 root cause). Same offer as on #8017: happy to fold review feedback into this PR if it fits, or split out. No rush.

wangmiao0668000666:
@Nillth Thanks for landing #7940 — same root cause as this PR, same fix shape (persist config first, move state after). Closing #8018 in favor of your merge. The sister delete-side fix in #8017 still stands and addresses #7941 (different code path: `delete_agent_cascade` instead of `rename_agent_cascade`).

PR #8019, opened 2026-06-19
fix(ci): pass provider-dispatch gate and --all-features build on master

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:** Fixes two pre-existing failures of the repo's own CI gates on `master`:
  1. `crates/zeroclaw-runtime/src/doctor/mod.rs` — `fetch_provider_catalog` called `ModelProvider::list_models()` directly, violating the provider-dispatch SSOT gate in `scripts/ci/rust_quality_gate.sh`. Now routed through `zeroclaw_providers::ProviderDispatch::from_ref(&*provider)` (same pattern as `llm_task.rs`); the dispatch returns `Vec<String>`, so behavior is unchanged.
  2. `crates/zeroclaw-channels/src/{line,wechat}.rs` — `PeerGroupConfig { channel: channel_ref.to_string(), .. }` passed a `String` where the field is now `ChannelRef` (an incomplete refactor). `channel_ref` is already a `ChannelRef`, so the `.to_string()` is dropped. This is **only reachable under `--all-features`**, so it broke `cargo check --all-features` / CI's `Check (all features)` and `Lint` jobs on master.
- **How it was found:** running the full `--all-features` preflight battery (the gap is that `rust_quality_gate.sh` uses `--all-targets` *without* `--all-features`, so feature-gated breakage slips past it).
- **Scope boundary:** No behavior change — pure gate/compile fixes. No API, config, or logic changes.
- **Blast radius:** `line`/`wechat` channel peer-group config construction (feature-gated) and the doctor catalog fetch path.
- **Linked issue(s):** none (CI hygiene on master).
- **Labels:** `type: ci`, `risk: low`, `size: S`, `channel`, `doctor`, `runtime`.

```bash
./scripts/ci/rust_quality_gate.sh --strict                       # fmt + clippy + dispatch gate -> clean
cargo clippy --all-targets --all-features --locked -- -D warnings # -> 0
cargo check  --all-targets --all-features --locked                # -> 0
```

- **Beyond CI:** the provider-dispatch gate now reports "clean" (previously flagged `doctor/mod.rs:451`); `--all-features` clippy and check both exit 0 (previously `E0308` in `line.rs:197` and `wechat.rs:925`).
- **Intentionally skipped:** none relevant — these are compile/gate fixes; the full `cargo test` runs in CI.

## Compatibility (required)

- Backward compatible? `Yes`
- Upgrade steps: none.

Low-risk: `git revert <sha>`.


singlerider (approved):
First review (no prior reviews, comments, or inline threads). Fetched the head locally and verified every claim against source, the dispatch gate script, and the `ProviderDispatch` precedents. CI is 18/18 green.

Pure CI-hygiene fix for two pre-existing failures on master. No behavior change.

### 🟢 What looks good — the doctor catalog fetch is routed through the dispatch SSOT exactly like every other call site

`fetch_provider_catalog` called `create_doctor_model_provider(...)?.list_models()` directly, and `list_models(` is in the `PROTECTED_METHODS` set of the provider-dispatch gate (`scripts/ci/rust_quality_gate.sh:26`), which forbids direct `ModelProvider` method calls outside `ProviderDispatch`. The fix routes through `ProviderDispatch::from_ref(&*provider).list_models().await`, which is byte-for-byte the established pattern (`model_pin.rs:72`, plus orchestrator and memory call sites). `list_models()` returns `Vec<String>` on both paths, so the catalog result is unchanged. Correct and consistent.

### 🟢 What looks good — the ChannelRef fixes drop a stale `.to_string()` from an incomplete refactor

`PeerGroupConfig.channel` is now `crate::providers::ChannelRef` (multi_agent.rs:153), and `channel_ref` is already a `ChannelRef` (`ChannelRef::new(...)`) at both sites. The old `channel: channel_ref.to_string()` was an `E0308` type mismatch left over from the field's type migration, reachable only under `--all-features` (which is why it broke `Check (all features)` / `Lint` but not the gate's `--all-targets`-only clippy). Dropping `.to_string()` is the right fix, and `channel_ref` is consumed exactly once in each `entry().or_insert_with` closure, so moving it instead of cloning a string is sound with no use-after-move.

### 🟢 What looks good — the root-cause note is the useful part

The body correctly identifies the real gap: `rust_quality_gate.sh` runs `--all-targets` *without* `--all-features`, so feature-gated breakage slips past the local gate while still failing the separate `Check (all features)` CI job. Validation evidence is concrete (gate clean, `--all-features` clippy and check both 0, with the exact previously-failing locations cited). Worth a follow-up to add `--all-features` to the local gate so this class of drift is caught before push, but that is out of scope for a four-line unbreak.

Label note: I am adding `type: ci` / `risk: low` / `size: S` to match the body, since the applied set was missing the risk/size/type labels. Approving.

Audacity88 (approved):
Reviewed current head `cb34459af5272926e6144bd3000da8bafe54238c` against the live PR state, PR body, current diff, prior approval review, green GitHub checks, the provider-dispatch gate script, and the relevant `ProviderDispatch` / `PeerGroupConfig` source context. The branch is behind current `master`, but GitHub reports a clean merge and I checked that current `master` has no same-file overlap with this PR. I did not run local Cargo for this review.

### 🟢 What looks good — the doctor catalog path now uses the provider-dispatch boundary

`fetch_provider_catalog()` was the one production call site still calling `list_models()` directly on the provider handle. Routing it through `ProviderDispatch::from_ref(&*provider).list_models().await` matches the dispatch gate's source-of-truth rule and preserves the returned `Vec<String>` catalog shape while restoring attribution coverage.

### 🟢 What looks good — the channel compile fixes match the `PeerGroupConfig` type

`PeerGroupConfig.channel` is already a `ChannelRef`, and both LINE and WeChat construct `channel_ref` as a `ChannelRef` before building the peer group. Dropping the stale `.to_string()` matches the Telegram precedent and fixes the feature-gated `--all-features` type mismatch without changing the persisted peer-group meaning.

### 🟢 What looks good — the validation targets the actual failures

The PR body names the two failure classes and cites the right checks: the strict provider-dispatch gate, `--all-features` clippy, and `--all-features` check. The visible GitHub checks are green on this head, including Check (all features), Lint, Test, and CI Required Gate. This looks like the right small CI-unblocker to me.

PR #8021, opened 2026-06-19
feat(skills): typed slash-command options in SKILL.md frontmatter

Nillth:
## Summary

- **What & why:** A `slash`-tagged skill can register native Discord slash-command parameters (dropdowns, integer/number ranges, length bounds) only if its manifest declares typed `slash_options`. Until now that was possible **only in the legacy `SKILL.toml`** format (`[[skill.slash_options]]`, added in #7844); a canonical `SKILL.md` skill could not — `load_skill_md` always produced empty `slash_options`. So a skill authored in the spec-canonical MD format silently fell back to a single free-text `input` on Discord, and the web dashboard (which only reads `SKILL.md`) couldn't express options at all. This makes **`SKILL.md` frontmatter a first-class home for typed slash options**, reaching parity with TOML and letting the canonical format drive native commands.
- **Shape:** options live in a nested `slash_options:` frontmatter key — the one genuinely nested field in an otherwise flat schema:
  ```yaml
  ---
  name: draft
  description: Draft written content to a spec.
  tags: [slash]
  slash_options:
    - name: format
      description: Output format.
      type: string
      required: true
      choices: [{name: Email, value: email}, {name: Tweet, value: tweet}]
    - name: words
      type: integer
      min: 10
      max: 2000
  ---
  # Draft …
  ```
- **No new dependency.** The flat frontmatter parsers deliberately avoid a YAML crate (binary size). Rather than reverse that, this adds a **scoped, hand-rolled parser for the bounded `slash_options:` subset** the serializer emits, shared by both frontmatter readers (loader + `SkillDocument`) so there is **one** nested parser, no drift. (A maintained YAML crate — `serde_yml`/`serde_norway`, collapsing both parsers — is a cleaner end-state and worth discussing; deferred to keep this change dependency-free. `serde_yaml` is already in the tree transitively via `probe-rs`/`wasm-compose`, so that conversation is open if maintainers prefer it.)
- **The Discord channel is unchanged.** `map_skill_slash_options` already reads `Skill.slash_options`; populating it from MD is sufficient. Proven by a new end-to-end test that loads a `SKILL.md` through the public loader and asserts the registration body carries typed options, not the single `input` fallback.
- **Scope boundary:** authoring path only. The dashboard editor form for these options, the visibility fix for listing TOML-only skills, and the TOML↔MD converter are a designed follow-up (PR B), sequenced after #7963 to avoid file collisions.

### Files
- `crates/zeroclaw-runtime/src/skills/frontmatter.rs` — `SkillFrontmatter.slash_options` (excluded from the flat `prop_fields()` form; drift test updated + guarded). `Eq` dropped (f64 bounds).
- `crates/zeroclaw-runtime/src/skills/document.rs` — shared `parse_slash_options` + `write_slash_options` + `locate_slash_options_block`; `Eq` dropped on `SkillDocument`.
- `crates/zeroclaw-runtime/src/skills/mod.rs` — `SkillMarkdownMeta.slash_options`; `parse_simple_frontmatter` calls the shared parser; both MD loaders populate `Skill.slash_options`.
- `crates/zeroclaw-runtime/src/skills/service.rs` — `Eq` dropped on `SkillSummary`.
- `src/skills/mod.rs` — CLI `skills add` constructs the new field (tagless/options-less scaffold).
- `crates/zeroclaw-channels/src/discord/slash.rs` — **test only** (E2E MD→Discord registration proof).

## Validation Evidence

```text
cargo fmt --all -- --check                                            # exit 0
cargo clippy --all-targets --features channels-full -- -D warnings    # 0 warnings (runtime, channels, config, eval, hardware, memory, providers, tools, root zeroclawlabs)
cargo clippy -p zeroclaw-gateway --all-targets -- -D warnings         # 0 warnings (downstream user of the changed types)
cargo test -p zeroclaw-runtime --lib skills::                         # 192 passed; 0 failed
cargo test -p zeroclaw-channels --features channel-discord --lib 'discord::slash'   # 16 passed; 0 failed
```
Ripple of the additive field + `Eq` drop verified: the only downstream users of the runtime
`SkillFrontmatter`/`SkillSummary`/`SkillDocument` are the gateway (`api_skills.rs`, validated above)
and the CLI (`src/skills`, in the root crate, validated). `apps/zerocode` defines its OWN
`SkillFrontmatter` wire-mirror (unaffected; it ignores the extra field on the wire). No type is used
as a map/set key, so dropping `Eq` is safe.
New tests: 8 in `document.rs` (inline + block choices, bounds, absent, malformed-drop, description-not-polluted, block-scalar-survives, round-trip), 2 in `mod.rs` (`load_skill_md` populates / empty), 1 in `slash.rs` (end-to-end MD→typed registration body).

- **Beyond CI:** not yet manually run against live Discord; the end-to-end test exercises loader → spec derivation → registration-body shape. (Plan: drop a `SKILL.md` with `slash_options` on the box, confirm `/demo` registers typed options.)

## Security & Privacy Impact
- New permissions / FS access? **No.** Same loader, same files; only an additional frontmatter field is parsed.
- New external calls? **No.** Discord registration is unchanged; it already reads `Skill.slash_options`.
- Secrets/PII? **No.**
- Note: exposing a skill as a slash command remains gated by the `slash` tag (and `open-skills` exclusion), unchanged.

## Compatibility
- **Backward compatible.** Existing `SKILL.toml` skills load and run unchanged (TOML still wins at runtime; reading preserved). MD skills without `slash_options` behave exactly as before. The single-`input` fallback persists for option-less slash skills. The new field is `#[serde(default, skip_serializing_if = "Vec::is_empty")]` so JSON/wire stays identical for skills without options.
- Config/env/CLI surface: no new keys. `prop_fields()` count unchanged (7) → dashboard/CLI/TUI flat forms unaffected.

## Rollback
- `risk: low` (authoring/parse only, no runtime/network change). Revert the commit; no migration needed (existing files unaffected; the field is additive and optional).

---
_No AI/Co-Authored-By trailers (CONTRIBUTING.md). Signed commits._

### DIFF
--- a/crates/zeroclaw-channels/src/discord/slash.rs
+++ b/crates/zeroclaw-channels/src/discord/slash.rs
@@ -605,6 +605,70 @@ mod typed_option_tests {
     }
 
     #[test]
+    fn md_skill_frontmatter_options_drive_a_typed_command_end_to_end() {
+        // End-to-end proof that no channel-side change was needed: a SKILL.md
+        // declaring slash_options in its frontmatter loads through the public
+        // runtime loader and registers a typed Discord command (dropdown +
+        // integer range), not the legacy single `input`.
+        let tmp = tempfile::tempdir().unwrap();
+        let skill_dir = tmp.path().join("draft");
+        std::fs::create_dir_all(&skill_dir).unwrap();
+        let md = r#"---
+name: draft
+description: Draft content to a spec.
+tags: [slash]
+slash_options:
+  - name: format
+    description: Output format.
+    type: string
+    required: true
+    choices: [{name: Email, value: email}, {name: Tweet, value: tweet}]
+  - name: words
+    type: integer
+    min: 10
+    max: 2000
+---
+# Draft
+
+Write it.
+"#;
+        std::fs::write(skill_dir.join("SKILL.md"), md).unwrap();
+
+        let skills = zeroclaw_runtime::skills::load_skills_from_directory(tmp.path(), false);
+        let specs = discord_slash_specs_from_skills(&skills);
+        assert_eq!(
+            specs.len(),
+            1,
+            "the slash-tagged MD skill yields one command"
+        );
+
+        let body = slash_command_registration_body(&specs);
+        let arr = body.as_array().unwrap();
+        let draft = arr
+            .iter()
+            .find(|c| c["name"] == json!("draft"))
+            .expect("draft command present");
+        let opts = draft["options"].as_array().unwrap();
+
+        // Typed options replaced the legacy single `input`.
+        assert!(opts.iter().all(|o| o["name"] != json!("input")));
+        let format = opts
+            .iter()
+            .find(|o| o["name"] == json!("format"))
+            .expect("format option");
+        assert_eq!(format["type"], json!(3)); // STRING
+        assert_eq!(format["required"], json!(true));
+        assert_eq!(format["choices"].as_array().unwrap().len(), 2);
+        let words = opts
+            .iter()
+            .find(|o| o["name"] == json!("words"))
+            .expect("words option");
+        assert_eq!(words["type"], json!(4)); // INTEGER
+        assert_eq!(words["min_value"], json!(10));
+        assert_eq!(words["max_value"], json!(2000));
+    }
+
+    #[test]
     fn projection_is_stable_across_int_vs_float_bounds() {
         // What we send (integer min/max) vs what Discord might echo back (float),
         // plus Discord's server-side decorations — must project equal.
--- a/crates/zeroclaw-runtime/src/skills/document.rs
+++ b/crates/zeroclaw-runtime/src/skills/document.rs
@@ -7,8 +7,11 @@
 use std::fmt::Write as _;
 
 use super::frontmatter::SkillFrontmatter;
+use super::{SkillSlashChoice, SkillSlashOption};
 
-#[derive(Debug, Clone, PartialEq, Eq)]
+// `Eq` is intentionally NOT derived: the frontmatter's `slash_options` carry
+// `f64` bounds (no total ordering). `PartialEq` covers the round-trip tests.
+#[derive(Debug, Clone, PartialEq)]
 pub struct SkillDocument {
     pub frontmatter: SkillFrontmatter,
     pub body: String,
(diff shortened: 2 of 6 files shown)


WareWolf-MoonWall (approved):
# PR #8021 — feat(skills): typed slash-command options in SKILL.md frontmatter

Reviewed by **WareWolf-MoonWall** at `65643bf`.

The "locate + skip + parse separately" architecture is the cleanest path for a nested YAML block inside a flat frontmatter scanner. Carving out the `slash_options:` range before the flat scan means the flat parser never sees indented option fields as top-level keys — no special-casing needed inside it. The hand-rolled parser covering only the subset the serializer emits is an acceptable tradeoff for keeping the diff dependency-free; `serde_yaml` is already transitive if the team wants to consolidate later.

### 🟢 "Locate + skip + parse" architecture

Clean separation between the nested block and the flat scanner. `locate_slash_options_block` returns a `[start, end)` line range; the flat parser skips those indices. No entanglement.

### 🟢 Single canonical `parse_slash_options` shared by both MD loaders

One parser, no drift. The SSOT rule is observed correctly across `document.rs` and `mod.rs`.

### 🟢 `Eq` dropped correctly

`Eq` is removed with an explanatory comment pointing at the `f64` bounds. The PR body correctly notes that dashboard/CLI/TUI flat forms are unaffected. The derive removal is intentional, not a regression.

### 🟢 End-to-end Discord test

The test in `discord/slash.rs` loads a SKILL.md with `slash_options`, runs `discord_slash_specs_from_skills`, and asserts STRING-with-choices and INTEGER-with-min/max both appear in the registration body. This proves the channel side is unchanged — MD-declared options flow through `map_skill_slash_options` without any channel modifications.

### 🟡 Not yet live-tested against Discord

The spec-derivation path is covered by CI. Confirm `/demo` registers the typed options against a live Discord application before closing — not a merge blocker, but worth a follow-up comment once confirmed.

### 🔵 `OptionBuilder.description` defaults silently to empty string

A doc comment noting that `description` is optional in the frontmatter schema (Discord accepts an absent description field) would clarify intent and prevent future contributors from treating the empty-string default as a bug.

---

Approving. Live Discord confirmation is the only remaining follow-up.

WareWolf-MoonWall:
@singlerider — milestone alignment needed: this feat(skills) PR (typed slash-command options in SKILL.md) is unmilestoned. Likely fits v0.8.2 (skills + plugin-platform scope) or v0.8.3. Please advise on placement.

PR #8022, opened 2026-06-19
feat(web): themed, click-to-open config pickers + dashboard health fix-modal + storage nav fix

Nillth:
## Summary
The Operator Console config pages rendered several fields as browser-native
`<select>`s whose dropdown lists ignore the dark theme, and the schema-driven
`alias-ref` fields (risk/runtime profile, model/tts/transcription/classifier
provider) had degraded to a free-text box. This replaces them with the app's
themed, click-to-open pickers and fixes two adjacent issues found while testing.

## What changed
**Themed alias-ref pickers** (`web/src/components/sections/FieldForm.tsx`)
- `alias-ref` fields now render the themed `ComboBox` instead of a native `<select>`.
- Options resolve via `/api/config/resolve-alias-source`, deriving the `source`
  from the field's `<Type>Ref` type-hint (`ALIAS_REF_TYPE_TO_SOURCE`) when the
  daemon omits `alias_source`. This makes provider refs that have no
  `AgentOptionsResponse` list (tts / transcription / classifier) resolve real
  options. A new `aliasRefResolving` flag shows a spinner only while genuinely
  fetching — never a permanent one.
- `delegates` added to the agent multi-select map (themed agent picker, was free text).

**New locked `Select` primitive** (`web/src/components/ui/Select.tsx`, exported from `ui/index.ts`)
- Themed, keyboard-navigable, click-to-open dropdown for fixed value sets. Wired
  into the FieldForm `enum` branch (e.g. `memory.backend`) — the last native
  `<select>` users hit. (`bool` already used a themed switch.)

**Click-to-open** (`web/src/components/ui/ComboBox.tsx`)
- Opt-in `openOnFocus` prop so the alias, model-field, bundle, and peer-group
  pickers open on click, not just via the caret.

**Peer-groups picker** (`web/src/pages/Config.tsx`)
- "Add agent to group" native `<select>` → themed `ComboBox`; Add is gated on a
  real non-member group so the free-text input can't submit a bogus group.

**Dashboard Component Health → fix-in-place** (`web/src/pages/Dashboard.tsx`)
- Error rows whose `last_error` is a config path get a **Fix** button that opens
  the same `DoctorFixModal` the Doctor page uses. `healthFixTarget()` maps
  `agents.*` / `channels.*` / `providers.models.*` → `{prefix, entity, href}`,
  anchored at string start so prose errors don't yield bogus targets.

**Storage nav fix** (`web/src/components/sections/SectionNavigator.tsx`)
- The `typed_family_map` child-loader now accepts the `"created"` badge. The
  backend renames storage's `configured` badge → `created`
  (`crates/zeroclaw-gateway/src/api_sections.rs`), so configured storages
  previously showed "Nothing configured yet".

**Cleanup** (`web/tsconfig.app.json`)
- Drop redundant `baseUrl` (TS ≥4.1 resolves `paths` relative to the config;
  Vite resolves `@` via its own alias).

## Testing
- `tsc -b` clean; `vite build` clean.
- Verified end to end against a running daemon: alias pickers populate (incl.
  tts/transcription/classifier via `resolve-alias-source`), enum picker themed,
  the Storage section lists its configured backends, and the dashboard Fix button
  opens the modal for a flagged `agents.<alias>.<field>`.

## Notes / limitations
- `alias-ref` is **pick-or-type** (themed `ComboBox`), consistent with the
  existing bundle pickers; values are validated on save. Not a hard-locked select
  — happy to lock it if preferred.
- On daemons that predate `alias_source` AND lack `resolve-alias-source`, the
  derived-source fetch 404s and is caught (empty picker) — graceful, but emits one
  failing request per such field.
- `Select` shares listbox mechanics (outside-click, scroll-into-view, arrow nav)
  with `ComboBox` by copy; a shared hook is a reasonable follow-up (the
  locked-vs-free-text component split itself is intentional).

### DIFF
--- a/web/src/components/sections/FieldForm.tsx
+++ b/web/src/components/sections/FieldForm.tsx
@@ -39,7 +39,7 @@ import {
 } from "lucide-react";
 import DirectoryPicker from "./DirectoryPicker";
 import ToolPicker from "@/components/ToolPicker";
-import { Badge, Button, ComboBox } from "@/components/ui";
+import { Badge, Button, ComboBox, Select } from "@/components/ui";
 import type { BadgeTone } from "@/components/ui";
 import { t } from "@/lib/i18n";
 import {
@@ -570,6 +570,9 @@ const AGENT_MULTI_ALIAS_FIELDS: Record<string, keyof AgentOptionsResponse> = {
   "skill_bundles": "skill_bundles",
   "knowledge_bundles": "knowledge_bundles",
   "mcp_bundles": "mcp_bundles",
+  // Delegates is a subset of the configured agents — give it the same themed
+  // multi-select (with agent suggestions) as the bundle fields, not free text.
+  delegates: "agents",
 };
 
 // Peer-groups carry the same alias-ref shape as agents do: a single
@@ -623,6 +626,31 @@ const LEAF_SINGLE_ALIAS_FIELDS: Record<string, keyof AgentOptionsResponse> = {
 function leafSingleAliasKind(path: string): keyof AgentOptionsResponse | null {
   const leaf = path.split(".").pop() ?? "";
   return LEAF_SINGLE_ALIAS_FIELDS[leaf] ?? null;
+}
+
+// `kind:"alias-ref"` fields carry a `<Type>Ref` type_hint naming the section
+// they reference. Map it to the `resolve-alias-source` query value (the backend
+// AliasSource variants). Used to DERIVE the source when the daemon emits the
+// kind but omits `alias_source` (older builds): the generic resolver still works
+// — covering provider refs (incl. tts/transcription/classifier) that have no
+// AgentOptionsResponse list, so they get a real dropdown, not a stuck spinner.
+const ALIAS_REF_TYPE_TO_SOURCE: Record<string, string> = {
+  ModelProviderRef: "model_providers",
+  TtsProviderRef: "tts_providers",
+  TranscriptionProviderRef: "transcription_providers",
+  RiskProfileRef: "risk_profiles",
+  RuntimeProfileRef: "runtime_profiles",
+  ChannelRef: "channels",
+};
+
+// The `resolve-alias-source` query value for an alias-ref entry: prefer the
+// daemon-declared `alias_source`; else derive it from the `<Type>Ref` type_hint.
+// Returns null for non-alias-ref entries or unmapped ref types.
+function aliasRefSource(entry: ListResponseEntry): string | null {
+  if (entry.kind !== "alias-ref") return null;
+  if (entry.alias_source) return entry.alias_source;
+  const m = entry.type_hint?.match(/(\w+Ref)\b/);
+  return (m && ALIAS_REF_TYPE_TO_SOURCE[m[1] ?? ""]) ?? null;
 }
 
 // Cross-section navigation map for agent alias-ref fields. Each entry
@@ -1406,8 +1434,7 @@ function FieldRow({
   // `aliasSource` is undefined and the effect is a no-op, leaving the maps
   // above to resolve refs. When the backend does declare it, the
   // `renderer === 'alias-ref'` branch takes precedence over those maps.
-  const aliasSource =
-    entry.kind === "alias-ref" ? entry.alias_source : undefined;
+  const aliasSource = aliasRefSource(entry);
   const [aliasValues, setAliasValues] = useState<string[] | null>(null);
   useEffect(() => {
     if (!aliasSource) return;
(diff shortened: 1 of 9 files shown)


singlerider (changes requested):
I pulled this branch locally and read it against the backend `AliasSource` registry in `crates/zeroclaw-config/src/traits.rs` and the gateway list/resolve endpoints. The `Select` primitive itself is clean (options are a prop, keyboard handling uses semantic key names, ARIA roles are correct) and the native-`<select>` replacement is a real UX win. But three things block this in its current form.

### 🔴 Blocking: surface shadows the backend `AliasSource` registry

`FieldForm.tsx` adds `ALIAS_REF_TYPE_TO_SOURCE`, a hardcoded TS map from `<Type>Ref` to the `resolve-alias-source` query value:

```
ModelProviderRef -> model_providers
TtsProviderRef -> tts_providers
TranscriptionProviderRef -> transcription_providers
RiskProfileRef -> risk_profiles
RuntimeProfileRef -> runtime_profiles
ChannelRef -> channels
```

That is a verbatim shadow of `AliasSource` in `crates/zeroclaw-config/src/traits.rs` (`#[serde(rename_all = "snake_case")]`). The canonical registry already exists, `ConfigFieldEntry::from_prop_field` already propagates `info.alias_source` onto the wire, and the field is `Option<AliasSource>` on the entry. The reason the frontend has to derive it is that the backend currently stubs `alias_source: None` (see `api_config.rs:369` and the `prop_fields()` path) so nothing actually emits it yet.

The correct fix is to finish the backend so `alias_source` is populated for nested alias-ref fields, then the frontend reads `entry.alias_source` with no derive table. Deriving it in TS by regex-matching `type_hint` means every new `AliasSource` variant silently breaks the picker until someone remembers to edit this map, and it drops `SkillBundles`/`KnowledgeBundles`/`McpBundles`/`Agents`/`PeerGroups` (the map is already incomplete relative to the enum). Note #7855 is in flight to surface more refs via `Configurable`; this should land on that backend work, not route around it.

### 🔴 Blocking: two more surface-side structural shadows in Dashboard

`healthFixTarget` in `Dashboard.tsx` hardcodes both the `agents.<alias>.<field>` path shape and a tab-routing table (`provider$` -> `?tab=providers`, `^(risk|runtime)_profile` -> `?tab=general`). That is the surface re-deriving which config tab a backend field lives on. The tab a field belongs to is backend metadata (`ConfigTab` already exists on `PropFieldInfo`/`ConfigFieldEntry`). Read it from the entry instead of pattern-matching field names in the UI. Same class of drift: add a profile-shaped field server-side and this regex silently misroutes it.

### 🔴 Blocking: three unrelated features in one PR

The title says it plainly: themed pickers + dashboard health fix-modal + storage nav fix. These are independent changes touching different pages (`FieldForm`/`Select` vs `Dashboard` health modal vs `Config.tsx` peer-group picker), plus an unrelated `tsconfig.app.json` `baseUrl` removal. Split into focused PRs so each can be reviewed and reverted independently. The picker primitive is the strong one and would land fast on its own once the registry shadow is removed.

### 🟢 Good

The `Select` component is well built: options-as-prop, no embedded value list, correct `role="combobox"`/`listbox`/`option` ARIA, keyboard nav with wrap-around and scroll-into-view, outside-click close. The `addToGroup` guard (`nonMembers.includes(pickerValue)`) correctly handles the ComboBox now accepting free text where the old native select could not. The `healthFixTarget` start-anchored regex deliberately avoiding the "failed to load agents.json" false-positive is a careful touch.

Net: the UX direction is right. Pull the three structural shadows back to the backend registry and split the scope, and the picker work lands cleanly.

singlerider:
Head `495b1dce` is unchanged since my changes-requested, and all three of my 🔴 still apply to this branch as-is: the `ALIAS_REF_TYPE_TO_SOURCE` registry shadow, the `healthFixTarget` tab-routing shadow, and the three-features-in-one scope.

You have already acted on the scope blocker by splitting:

- #8086 (split 1/3) carries the `Select` primitive + picker, with the `alias_source` registry shadow documented as a tracked follow-up on #7855.
- #8087 (split 2/3) carries the dashboard fix-modal and resolves my blocker #2 by reading `entry.tab` from backend `ConfigTab` instead of the field-name regex.

Both are clean and waiting on review; I will take those next.

That leaves #8022 still holding the full bundle. Per your own plan (#8022c = storage nav, stacks on #8086), this PR should be reduced to just the `SectionNavigator.tsx` storage-nav fix once #8086 lands, retargeted onto it. Right now it duplicates everything in #8086/#8087, so I am leaving my CHANGES_REQUESTED in force rather than re-reviewing the superseded bundle.

Narrow this to the storage-nav fix (or close it in favor of the splits plus a dedicated storage-nav PR) and ping me. Not closing it unilaterally since you flagged the storage-nav intent.

Nillth:
Closing this in favor of the focused splits, per @singlerider's review and my own split plan.

- **#8086** (Select primitive + themed pickers) — merged.
- **#8087** (dashboard component-health fix-modal, reading `entry.tab` from backend `ConfigTab`) — merged.
- **#8345** (storage-nav fix) — the remaining slice: `SectionNavigator`'s `typed_family_map` loader now accepts the `"created"` badge that the backend `storage_picker` emits, so configured storage backends show in the nav instead of "Nothing configured yet".

The backend half of the storage-nav fix (`configured -> created` in `api_sections.rs`) already landed in `master`, so #8345 is a one-line frontend change.

This PR's full bundle is now superseded by the three above; closing rather than rebasing the conflicting branch. Thanks for the review.

PR #8023, opened 2026-06-19
fix(mcp): stop leaking stdio child processes per heartbeat tick

### LINKED ISSUE #5903: [Bug]: MCP stdio child processes accumulate on daemon with heartbeat.enabled=true (one orphan per tick)

perlowja:
## Summary

- `crates/zeroclaw-tools/src/mcp_tool.rs`

## Linked issue

Resolves zeroclaw-labs/zeroclaw#5903

`risk: low` · `size: S`

## Testing (required)

> VERDICT: NEEDS-ATTENTION
> FINDINGS:
> - <crates/zeroclaw-tools/src/mcp_tool.rs:21> HIGH: The `registry` field type changed from `Arc<McpRegistry>` to `Weak<McpRegistry>`. Any external code that directly accesses or clones this field will now fail to compile or behave incorrectly. This is a breaking API change and requires updates to all callers or a compatibility shim.
> - <crates/zeroclaw-tools/src/mcp_tool.rs:70> MED: When the weak reference cannot be upgraded, the implementation returns `ToolResult { success: false, output: "", error: Some(...) }`. Callers that only check `success` may treat this as a normal failure without distinguishing it from a tool‑specific error, potentially masking the underlying registry‑drop condition. Consider defining a distinct error variant or propagating an error type.
> - <crates/zeroclaw-tools/src/mcp_tool.rs:70> LOW: The early return on upgrade failure bypasses any logging or metrics that might be useful for diagnosing why the registry is missing. Adding a log entry would aid observability.
> 
> [openreview] author_family=non-openai rotation: groq-gpt-oss-120b, deepseek-chat, groq-qwen3-32b
> [openreview] provider=groq-gpt-oss-120b model=openai/gpt-oss-120b

## Compatibility (required)

- Backward compatible? Yes

## Rollback

### DIFF
--- a/crates/zeroclaw-tools/src/mcp_tool.rs
+++ b/crates/zeroclaw-tools/src/mcp_tool.rs
@@ -1,7 +1,7 @@
 //! Wraps a discovered MCP tool as a zeroclaw [`Tool`] so it is dispatched
 //! through the existing tool registry and agent loop without modification.
 
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 
 use async_trait::async_trait;
 
@@ -13,6 +13,10 @@ use zeroclaw_api::tool::{Tool, ToolResult};
 ///
 /// The `prefixed_name` (e.g. `filesystem__read_file`) is what the agent loop
 /// sees. The registry knows how to route it to the correct server.
+///
+/// Uses `Weak<McpRegistry>` so the registry can be dropped and stdio child
+/// processes reaped (via `kill_on_drop`) when the agent run ends, even if
+/// wrappers are still alive in skill-elevation or other transient handles.
 pub struct McpToolWrapper {
     /// Prefixed name: `<server_name>__<tool_name>`.
     prefixed_name: String,
@@ -21,8 +25,10 @@ pub struct McpToolWrapper {
     description: String,
     /// JSON schema for the tool's input parameters.
     input_schema: serde_json::Value,
-    /// Shared registry — used to dispatch actual tool calls.
-    registry: Arc<McpRegistry>,
+    /// Shared registry (weak reference) — used to dispatch actual tool calls.
+    /// Upgraded to Arc at call time so the registry can be dropped when all
+    /// agent runs end, allowing stdio child processes to be reaped.
+    registry: Weak<McpRegistry>,
 }
 
 impl McpToolWrapper {
@@ -32,7 +38,7 @@ impl McpToolWrapper {
             prefixed_name,
             description,
             input_schema: def.input_schema,
-            registry,
+            registry: Arc::downgrade(&registry),
         }
     }
 }
@@ -64,7 +70,20 @@ impl Tool for McpToolWrapper {
             }
             other => other,
         };
-        match self.registry.call_tool(&self.prefixed_name, args).await {
+        let registry = match self.registry.upgrade() {
+            Some(r) => r,
+            None => {
+                return Ok(ToolResult {
+                    success: false,
+                    output: String::new(),
+                    error: Some(format!(
+                        "MCP registry dropped; cannot call tool `{}`",
+                        self.prefixed_name
+                    )),
+                });
+            }
+        };
+        match registry.call_tool(&self.prefixed_name, args).await {
             Ok(output) => Ok(ToolResult {
                 success: true,
                 output,
--- a/crates/zeroclaw-tools/src/mcp_tool.rs
+++ b/crates/zeroclaw-tools/src/mcp_tool.rs
@@ -178,11 +178,17 @@ mod tests {
         // rather than propagating an Err (non-fatal by design).
         let registry = empty_registry().await;
         let def = make_def("ghost", Some("Ghost tool"), json!({}));
-        let wrapper = McpToolWrapper::new("nowhere__ghost".to_string(), def, registry);
+        // Keep a strong `Arc` alive across the call: the wrapper now holds only a
+        // `Weak`, mirroring the agent runtime that owns the registry while its
+        // tools execute. Without this the registry would drop before `execute()`
+        // and the `Weak` would fail to upgrade (yielding a "registry dropped"
+        // error instead of the expected unknown-tool error).
+        let wrapper = McpToolWrapper::new("nowhere__ghost".to_string(), def, registry.clone());
         let result = wrapper
             .execute(json!({}))
             .await
             .expect("execute should be non-fatal");
+        drop(registry);
         assert!(!result.success);
         let err_msg = result.error.expect("error message should be present");
         assert!(
--- a/crates/zeroclaw-tools/src/mcp_tool.rs
+++ b/crates/zeroclaw-tools/src/mcp_tool.rs
@@ -199,6 +199,52 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn eager_wrapper_dispatches_while_run_owner_holds_registry() {
+        // Eager-path lifetime contract (#5903 / #8023): the wrapper holds only a
+        // `Weak`, so the agent run must keep one strong `Arc<McpRegistry>` alive
+        // (now stored on `Agent`, dropped in `Agent::drop`). While that owner
+        // lives, the wrapper upgrades and dispatches; once it drops, the wrapper
+        // fails gracefully and the registry (and its kill_on_drop stdio children)
+        // can be reaped. This is the scenario the unknown-tool test does not
+        // cover and that a `Weak`-only wrapper regressed.
+        let run_owner = empty_registry().await; // the strong Arc the agent holds
+        let def = make_def("ghost", Some("Ghost tool"), json!({}));
+        // Built as the eager path does: new() downgrades internally, so the Arc
+        // passed here is consumed and `run_owner` is left as the sole strong
+        // reference — exactly the post-init state.
+        let wrapper =
+            McpToolWrapper::new("nowhere__ghost".to_string(), def, Arc::clone(&run_owner));
+
+        // Owner alive -> upgrade succeeds -> dispatch reaches the registry and
+        // returns the genuine unknown-tool error, NOT "registry dropped".
+        let ok = wrapper.execute(json!({})).await.expect("execute non-fatal");
+        assert!(!ok.success);
+        assert!(
+            ok.error
+                .as_deref()
+                .unwrap_or("")
+                .contains("unknown MCP tool"),
+            "while the run owner lives the wrapper must dispatch; got: {:?}",
+            ok.error
+        );
+
+        // Drop the run owner (mirrors Agent::drop) -> no strong refs remain ->
+        // the Weak can no longer upgrade -> graceful "registry dropped".
+        drop(run_owner);
+        let after = wrapper.execute(json!({})).await.expect("execute non-fatal");
+        assert!(!after.success);
+        assert!(
+            after
+                .error
+                .as_deref()
+                .unwrap_or("")
+                .contains("registry dropped"),
+            "after the run owner drops the wrapper must fail gracefully; got: {:?}",
+            after.error
+        );
+    }
+
+    #[tokio::test]
     async fn execute_success_sets_success_true_and_output() {
         // Verify the ToolResult success-branch struct shape compiles correctly.
         // A real happy-path requires a live MCP server; that is covered by E2E tests.
--- a/crates/zeroclaw-runtime/Cargo.toml
+++ b/crates/zeroclaw-runtime/Cargo.toml
@@ -113,6 +113,7 @@ tempfile = "3.26"
 tokio = { version = "1.50", features = ["rt-multi-thread", "macros"] }
 scopeguard = "1.2"
 filetime = "0.2"
+zeroclaw-tools = { workspace = true, features = ["test-support"] }
 
 # Channel features (forwarded to zeroclaw-channels)
 
--- a/crates/zeroclaw-runtime/src/agent/agent.rs
+++ b/crates/zeroclaw-runtime/src/agent/agent.rs
@@ -182,6 +182,13 @@ pub struct Agent {
     /// When MCP deferred loading is enabled, tools are activated via `tool_search`
     /// and stored here for lookup during tool execution.
     activated_tools: Option<Arc<std::sync::Mutex<crate::tools::ActivatedToolSet>>>,
+    /// Strong owner of the MCP registry for the agent's lifetime. Eager-path
+    /// `McpToolWrapper`s hold only a `Weak`, so without this the registry would
+    /// drop right after init and every MCP call would fail with "registry
+    /// dropped"; holding it here keeps the wrappers upgradable for the whole run
+    /// and drops it in `Agent::drop` — exactly when #5903 wants the stdio
+    /// children reaped via the registry's `kill_on_drop` transport.
+    mcp_registry: Option<Arc<crate::tools::McpRegistry>>,
     /// Hook runner for tool-call auditing and lifecycle side effects.
     /// See issue #5462.
     hook_runner: Option<Arc<crate::hooks::HookRunner>>,
@@ -214,6 +221,10 @@ impl Drop for Agent {
                     "model_provider": self.model_provider_name,
                     "model": self.model_name,
                     "history_messages_freed": self.history.len(),
+                    // Dropping the Agent releases the last strong Arc to the MCP
+                    // registry (when one was held), letting it drop and reap its
+                    // stdio children via kill_on_drop (#5903).
+                    "mcp_registry_released": self.mcp_registry.is_some(),
                 })),
             "Agent dropped; conversation history and per-session state freed"
         );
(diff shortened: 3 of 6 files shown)


singlerider (dismissed):
First review (no prior reviews, comments, or inline threads). Fetched the head locally and traced the registry's full ownership graph through `agent.rs` MCP init, the deferred and eager tool-set paths, `mcp_deferred.rs`, and the `kill_on_drop` transport. CI is 18/18 green.

Resolves #5903: `McpToolWrapper` held `Arc<McpRegistry>`, so long-lived wrappers (skill MCP elevation, transient handles) pinned the registry forever and its `Drop`-driven `kill_on_drop` never reaped stdio children — one orphan per heartbeat tick. The direction here is right: a `Weak` in the wrapper breaks that retain cycle so the registry can drop when the run ends. But as written it overshoots and breaks MCP in the default configuration.

### 🔴 Blocking — in the default eager path nothing retains a strong `Arc<McpRegistry>`, so every MCP tool call fails

`default_deferred_loading()` returns `false` (schema.rs:4466), so eager loading is the default. In the eager `else` branch (agent.rs:1280-1322), the only strong `Arc<McpRegistry>` is the local `registry` created at agent.rs:1241. It builds `McpToolWrapper`s — now `Weak`-backed — and pushes them into `tools`, then goes out of scope at the end of the MCP-init block. `mcp_elevation_arcs` are also `Weak`-backed wrappers, and the `Agent` struct (fields at agent.rs:147-200) stores no registry. So once init returns, the strong count hits zero: the registry drops immediately (killing the stdio children before they are ever used), and when the agent loop later invokes an MCP tool, `self.registry.upgrade()` returns `None` and the call fails with "MCP registry dropped; cannot call tool". On master the eager wrappers' strong `Arc` is exactly what kept the registry alive for the run — that lifetime guarantee is removed here with nothing put in its place. This trades a slow leak for total MCP breakage in the default config.

The deferred path happens to survive only because `DeferredMcpToolSet` holds `pub registry: Arc<McpRegistry>` (mcp_deferred.rs:60), which lives inside the `ToolSearchTool` in `tools` for the whole run. The eager path needs the same: have the agent hold one strong `Arc<McpRegistry>` for the run's lifetime — store it on the `Agent` struct (dropped in the agent's own `Drop`, which is precisely when #5903 wants the children reaped), or move the strong owner into the eager tool set the way the deferred path already does. Then the `Weak` wrappers upgrade fine during the run and the registry drops only when the run ends.

### 🔴 Blocking — the new test coverage masks this; add an eager-lifetime regression test

`execute_returns_non_fatal_error_for_unknown_tool` keeps a strong `Arc` alive across the call by design, and the new `dropping_stdio_registry_reaps_child_process` (mcp_client.rs:706) proves drop-reaping in isolation — neither reproduces the eager-path scenario where the agent outlives the init-scope `Arc`. A test that builds the eager tool set, drops the init-scope strong `Arc`, then calls an MCP tool and asserts it still dispatches (upgrade succeeds) would have caught this and will lock in the fix. Pair it with one asserting the registry (and its children) drop once the run owner is gone, so both halves of the contract are pinned.

### 🟢 What looks good — `Weak` is the correct tool, and the advisory findings are correctly dismissed

Using `Weak` to break the retain cycle is the right call, and the struct doc-comment naming skill-elevation wrappers as the culprit is accurate. The adversarial "breaking API change" HIGH is a false alarm: `McpToolWrapper::new` still takes `Arc<McpRegistry>` (downgrades internally) and the `registry` field is private, so no caller signature changes. The graceful upgrade-failure `ToolResult` is the right shape for genuinely-stale wrappers once the lifetime fix is in — it just should be the rare post-run case, not the default every-call outcome it currently is. The `kill_on_drop(true)` transport (mcp_transport.rs:122) plus the registry-drop reaping test confirm the underlying reap mechanism works, so the only missing piece is keeping one strong owner alive for the run.

Requesting changes: keep a strong `Arc<McpRegistry>` for the agent run's lifetime in the eager path, and add the eager-lifetime regression test.

crates/zeroclaw-tools/src/mcp_tool.rs
```diff
@@ -32,7 +38,7 @@ impl McpToolWrapper {
             prefixed_name,
             description,
             input_schema: def.input_schema,
-            registry,
+            registry: Arc::downgrade(&registry),
```
singlerider, line 41: 🔴 Blocking — root of the lifetime regression. Downgrading to `Weak` here is correct for breaking the retain cycle, but in the default eager path (`default_deferred_loading()` returns `false`, schema.rs:4466) nothing else holds a strong `Arc<McpRegistry>` for the run. The only strong ref is the local `registry` in `agent.rs:1241`, which drops at the end of the MCP-init block; `mcp_elevation_arcs` are also `Weak`-backed and the `Agent` struct (agent.rs:147-200) stores no registry. So the registry drops at init, the stdio children are killed before first use, and every subsequent `upgrade()` returns `None`. Keep one strong `Arc<McpRegistry>` alive for the agent run's lifetime (store it on `Agent`, dropped in the agent's `Drop` — exactly when #5903 wants reaping) so these `Weak` wrappers upgrade during the run.

crates/zeroclaw-tools/src/mcp_tool.rs
```diff
@@ -64,7 +70,20 @@ impl Tool for McpToolWrapper {
             }
             other => other,
         };
-        match self.registry.call_tool(&self.prefixed_name, args).await {
+        let registry = match self.registry.upgrade() {
+            Some(r) => r,
+            None => {
```
singlerider, line 75: 🔴 Blocking — in the default eager configuration this `None` arm is hit on *every* MCP tool call, not just for stale wrappers, because no strong `Arc<McpRegistry>` survives past agent init (see the `Arc::downgrade` comment). The graceful error is the right shape for genuinely-stale post-run wrappers, but with the lifetime fix in place it should be the rare case, not the default every-call outcome.

singlerider (commented):
Inline anchors for the two blockers in my changes-requested review above.

perlowja:
Thanks — you were right, the `Weak`-only change broke the default eager path. Fixed in the latest push:

- **Eager-path lifetime:** the `Agent` now holds one strong `Arc<McpRegistry>` for the run's lifetime (`mcp_registry` field, threaded through `AgentBuilder` from `from_config_…`). It drops in `Agent::drop` — exactly when #5903 wants the stdio children reaped via `kill_on_drop`. The `Weak` wrappers upgrade fine while the run owner lives; the deferred path already had this via `DeferredMcpToolSet`, so both paths now share the guarantee.
- **Regression test:** `eager_wrapper_dispatches_while_run_owner_holds_registry` asserts the wrapper dispatches (reaches the registry → genuine unknown-tool error) while the run owner is alive, and fails gracefully (`registry dropped`) once the owner drops — pinning both halves of the contract.

Validated against the real `ci-all`/1.93.0 gate: fmt/check/clippy clean, `zeroclaw-tools` mcp_tool tests 10/10. Ready for re-review.

perlowja:
Thanks for the close review. The eager-path lifetime regression is addressed on the current head (`0b8aa749c`) — the `CHANGES_REQUESTED` was against `1e702ff63`, before that push.

Fix: the agent run now holds a strong `Arc<McpRegistry>` on `Agent` (dropped in `Agent::drop`), so in the default eager path the registry stays alive for the whole run and `McpToolWrapper::execute` upgrades successfully on every call. The `Weak` in the wrapper still breaks the retain cycle so stdio children are reaped once the run ends. The `None`/"registry dropped" arm is now only hit for genuinely-stale post-run wrappers, as intended.

Added a regression test `eager_wrapper_dispatches_while_run_owner_holds_registry` that asserts the wrapper dispatches while the run owner holds the `Arc` and only fails gracefully after it drops — covering exactly the eager dispatch path the unknown-tool test did not.

Re-requesting review.

Audacity88 (dismissed):
Reviewed current head `0b8aa749c25f724529797ca7d3b7b499ee483b8a` against the live PR state, linked issue #5903, singlerider's prior review and inline comments, the current checks, and the source diff plus nearby MCP registry wiring in `agent.rs`, `loop_.rs`, and `mcp_tool.rs`. I did not run local Cargo validation or a live stdio MCP heartbeat smoke. The latest push does address the stale eager-path bug for `Agent::from_config`, but the linked heartbeat path still has the same lifetime hole.

### ✅ Resolved — `Agent::from_config` now keeps the registry alive for its run

The new `Agent.mcp_registry` owner fixes the specific eager-path shape from the previous review for `Agent::from_config`: weak MCP wrappers can now upgrade while that `Agent` lives, and the registry can still drop when the `Agent` drops. The new `eager_wrapper_dispatches_while_run_owner_holds_registry` test also captures the wrapper-level contract: owner alive means dispatch reaches the registry; owner dropped means the wrapper fails with `registry dropped`.

### 🔴 Blocking — `agent::run` still drops the eager MCP registry before tool use

#5903 is specifically about daemon heartbeat ticks, and the daemon heartbeat worker calls `crate::agent::run(...)`. That path is the standalone run loop in `crates/zeroclaw-runtime/src/agent/loop_.rs`, not the `Agent::from_config` builder path updated in the latest commit.

That `agent::run` body initializes MCP inline. In its MCP eager branch, `McpRegistry::connect_all(...)` is wrapped in a local `Arc`, the eager `McpToolWrapper`s are registered with `Arc::clone(&registry)`, and `McpToolWrapper::new(...)` immediately downgrades that clone to `Weak`. When `config.mcp.deferred_loading` is false, which is the default, nothing else in `agent::run` keeps a strong `Arc<McpRegistry>` alive after the MCP-init block. The local `registry` drops before the tool loop can call any eager MCP tool, so each wrapper's `upgrade()` will hit the new `registry dropped` arm.

There is a second copy of the same eager MCP setup in `process_message(...)` with the same ownership shape. The deferred paths survive because `DeferredMcpToolSet` owns an `Arc<McpRegistry>`, but the default eager paths still need an owner equivalent to the new `Agent.mcp_registry` field.

Please carry the same lifetime fix into the `agent::run` and `process_message` eager paths, or otherwise give those per-run tool registries a strong owner for exactly the duration of the run/message. The regression coverage also needs to exercise the linked issue's real entry point, not only `McpToolWrapper` in isolation or `Agent::from_config`, so a future edit cannot leave heartbeat/cron-style runs with dead weak wrappers.

### 🟢 What looks good — the retain-cycle direction is still right

Using `Weak` inside `McpToolWrapper` is the right direction for breaking long-lived wrapper ownership of stdio MCP processes, and the graceful `registry dropped` result is appropriate for genuinely stale post-run wrappers. The remaining issue is just that every runtime entry point that registers weak eager wrappers needs a matching strong owner for the lifetime in which those wrappers are expected to execute.

Leaving changes requested because the PR still does not cover the `agent::run` path that #5903 reports.

perlowja:
@Audacity88 @singlerider — fixed in 2a6c4e05e. You were both right that the prior commit only covered `Agent::from_config`; the daemon heartbeat enters through `agent::run` and `process_message`, which had the same shape.

Both now hold a **function-scope strong `Arc<McpRegistry>`** on the eager (non-deferred) path for the whole run/message, so the eager `Weak` wrappers upgrade during tool use and the stdio children are reaped at run end (the lifetime @singlerider asked for in `mcp_tool.rs`:41/:75). The deferred path is untouched (`DeferredMcpToolSet` already owns its registry).

The regression `run_eager_mcp_registry_keepalive_dispatches_tool` now drives the **real `agent::run` entry point** (#5903's heartbeat path) with eager MCP (`deferred_loading=false`) and a stdio MCP server, asserting an eager tool actually dispatches — it fails without the keepalive (registry dropped) and passes with it. Full-workspace CI (1.93.0 fmt / clippy `--features ci-all` / `check --locked` / nextest) is green. Re-requesting.

perlowja:
Closing to clear the upstream branch list: this branch was pushed to the canonical repo by mistake (inverted `origin`/fork remotes). The change is preserved on my fork as `perlowja:fix/mcp-stdio-child-leak` and will be resubmitted as a proper fork-based PR. A fork-only pre-push guard is now in place so this cannot recur. Apologies for the upstream-branch pollution.

PR #8025, opened 2026-06-19
feat(channels/wecom): proactive + attachment messaging over wecom_ws

### LINKED ISSUE #7824: [Feature]: wecom_ws support proactive messaging via zeroclaw channel send command and media file sending
### Summary

Enable wecom_ws channel to support proactive message sending via `zeroclaw channel send` command, including support for media files.

### Problem statement

Currently, the wecom_ws channel may only support passive message response through WebSocket connections. Users cannot proactively send messages to WeCom users/chats using the standard `zeroclaw channel send` command. This limits use cases such as:
1. Sending scheduled notifications or alerts via WeCom
2. Triggering messages from external systems through the CLI
3. Sending media files (images, voice, video, documents) to WeCom users

The current behavior requires an active WebSocket connection and can only respond to incoming messages, not initiate outbound communication.

### Proposed solution

1. **Support `channel send` command for wecom_ws**:
   - Enable the existing `zeroclaw channel send --channel <channel_id> --to <recipient> --content <message>` command to work with wecom_ws channels
   - Support both user-level and group-level recipients
   - Handle authentication via WeCom app credentials stored in channel config

2. **Media File Sending Support**:
   - Support sending WeCom media message types: image, voice, video, file
   - Accept local file paths or URLs as content
   - Handle media file upload to WeCom and temporary media_id management

3. **Command Usage Examples**:
   ```bash
   # Send text message
   zeroclaw channel send --channel wecom_1 --to user123 --content "Hello from CLI"
   
   # Send image file
   zeroclaw channel send --channel wecom_1 --to user123 --type image --content /path/to/image.png
   
   # Send from stdin
   echo "Alert: Server down" | zeroclaw channel send --channel wecom_1 --to @all

### Non-goals / out of scope

- Do not support advanced WeCom message types (news, template cards, mini-program)
- Do not provide media file management UI
- Do not modify existing WebSocket message receiving functionality

### Alternatives considered

- Use WeCom application message API directly - bypasses channel abstraction, requires custom implementation
- Webhook-based approach - adds complexity, less intuitive than CLI command
- Keep current behavior - does not solve the proactive messaging need

### Acceptance criteria

-  zeroclaw channel send command works with wecom_ws channels
-  Support sending text, image, voice, video, and file type messages
-  Media file upload and media_id handling works correctly
-  Command validates recipients and returns appropriate errors
-  Add documentation and examples for wecom_ws channel send usage

### Architecture impact

- channels/: wecom_ws channel implementation to support outbound send
- providers/: WeCom provider extensions for media upload and message sending
- CLI/: May need to ensure channel send command properly handles wecom_ws channel type

### Risk and rollback

- Risk: Potential for spam/abuse if command is misused; need to respect WeCom rate limits
- Rollback plan: The change is additive - existing functionality remains unchanged. Can disable via channel config if needed.

### Expected routing

Ordinary feature triage

### Next decision surface

milestone

### Breaking change?

No

### Data hygiene checks

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:** Make the wecom_ws proactive/attachment messaging paths compile and lint. Fixes: (1) `serde_json::json!` rejects expression keys — build the body then insert the msgtype-named media object separately; (2) `reqwest` `Part::mime_str` is fallible — apply `?` before adding the part to the form; (3) remove an unused `base64::Engine` import; (4) move `send_external_user_message` out of `impl Channel` (it is not a trait method) into an inherent impl; (5) `parse_scope` returns a `Result` whose `anyhow::Error` is not `PartialEq` — assert on the unwrapped tuple; (6) drop two never-read `errcode` fields and replace disallowed `anyhow::anyhow!` with `.with_context(..)`.
- **Scope boundary:** wecom_ws channel only; no trait or cross-channel changes.
- **Blast radius:** `zeroclaw-channels` wecom_ws path only.
- **Linked issue(s):** Resolves #7824

Validated against the **actual CI gate** (`ci-all` feature set, toolchain `1.93.0`):

```
cargo +1.93.0 fmt --all -- --check                                                  -> PASS
cargo +1.93.0 check  --locked --features ci-all                                     -> PASS
cargo +1.93.0 clippy --workspace --exclude zeroclaw-desktop --all-targets \
      --features ci-all --locked -- -D warnings                                     -> PASS
cargo +1.93.0 test --test architecture (config-isolation + fluent)                  -> PASS
```
GitHub `Test` (full nextest suite) + `CI Required Gate`: green.

- New permissions / fs access? `No`
- New external network calls? `No` (the wecom HTTP calls already existed in the feature)
- Secrets / credentials handling changed? `No`
- PII in diff/tests/fixtures/docs? `No`

## Compatibility (required)

- Backward compatible? `Yes`

Low risk: `git revert <sha>`.

### DIFF
--- a/crates/zeroclaw-channels/src/wecom_ws.rs
+++ b/crates/zeroclaw-channels/src/wecom_ws.rs
@@ -1436,8 +1436,6 @@ impl WeComWsChannel {
     /// Upload a media file and send it via `aibot_send_msg` with the appropriate
     /// WeCom `msgtype` (`image` or `file`).
     async fn send_attachment(&self, scope: &str, attachment: &MediaAttachment) -> Result<()> {
-        use base64::Engine as _;
-
         let (chat_type, chatid) = parse_scope(scope)?;
 
         // Upload via WeCom media/upload CGI
@@ -1452,13 +1450,16 @@ impl WeComWsChannel {
             }
         );
 
-        let form = reqwest::multipart::Form::new()
-            .part(
-                "media",
-                reqwest::multipart::Part::bytes(attachment.data.clone())
-                    .file_name(attachment.file_name.clone())
-                    .mime_str(attachment.mime_type.as_deref().unwrap_or("application/octet-stream")),
-            );
+        // Part::mime_str returns Result<Part>; apply ? before adding to the form.
+        let part = reqwest::multipart::Part::bytes(attachment.data.clone())
+            .file_name(attachment.file_name.clone())
+            .mime_str(
+                attachment
+                    .mime_type
+                    .as_deref()
+                    .unwrap_or("application/octet-stream"),
+            )?;
+        let form = reqwest::multipart::Form::new().part("media", part);
 
         let resp = self
             .client
@@ -1472,14 +1473,16 @@ impl WeComWsChannel {
         struct UploadResp {
             #[serde(rename = "media_id")]
             media_id: Option<String>,
-            errcode: Option<i64>,
             errmsg: Option<String>,
         }
 
         let upload: UploadResp = resp.json().await?;
-        let media_id = upload
-            .media_id
-            .ok_or_else(|| anyhow::anyhow!("WeCom upload response missing media_id: {:?}", upload.errmsg))?;
+        let media_id = upload.media_id.with_context(|| {
+            format!(
+                "WeCom upload response missing media_id: {:?}",
+                upload.errmsg
+            )
+        })?;
 
         // Dispatch the actual message via aibot_send_msg
         let msgtype = match attachment.kind() {
@@ -1489,15 +1492,19 @@ impl WeComWsChannel {
             zeroclaw_api::media::MediaKind::Unknown => "file",
         };
 
+        // serde_json::json! only accepts literal keys; the payload nests the
+        // media object under a key named after the msgtype, so build the body
+        // first and insert that dynamic key separately.
+        let mut body = serde_json::json!({
+            "chatid": chatid,
+            "chat_type": chat_type,
+            "msgtype": msgtype,
+        });
+        body[msgtype] = serde_json::json!({ "media_id": media_id });
         let frame = serde_json::json!({
             "cmd": "aibot_send_msg",
             "headers": { "req_id": random_ascii_token(16) },
-            "body": {
-                "chatid": chatid,
-                "chat_type": chat_type,
-                "msgtype": msgtype,
-                msgtype => { "media_id": media_id }
-            }
+            "body": body
         });
 
         self.ws_send_frame_and_wait_for_response(frame, "unused", "aibot_send_msg")
@@ -1521,14 +1528,12 @@ impl WeComWsChannel {
         struct TokenResp {
             #[serde(rename = "access_token")]
             access_token: Option<String>,
-            errcode: Option<i64>,
             errmsg: Option<String>,
         }
 
         let url = format!(
             "https://qyapi.weixin.qq.com/cgi-bin/gettoken?corpid={}&corpsecret={}",
-            self.bot_id,
-            self.secret,
+            self.bot_id, self.secret,
         );
 
         let resp = self
@@ -1545,7 +1550,7 @@ impl WeComWsChannel {
 
         token
             .access_token
-            .ok_or_else(|| anyhow::anyhow!("WeCom gettoken failed: {:?}", token.errmsg))
+            .with_context(|| format!("WeCom gettoken failed: {:?}", token.errmsg))
     }
 }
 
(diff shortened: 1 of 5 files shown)


perlowja:
Pushed a fix addressing issues found in self-review of the attachment/proactive paths:

- **Blocker — attachment send always timed out.** `send_attachment` registered the ack waiter under the literal `"unused"` while the frame carried a freshly generated `req_id`, so the ack never matched and every send blocked until timeout. Now one `req_id` is threaded into both the frame and the waiter (mirroring the markdown-send path).
- **Token caching + race.** `resolve_access_token` now caches the qyapi token with TTL+skew, refreshes outside the cache lock, and re-checks the cache after a failed `gettoken` so a concurrent caller's fresh token is used instead of surfacing a rate-limit error (no lock held across `.await`).
- **External-scope attachments** are now rejected before upload/send rather than routed over the WS `aibot_send_msg` path the channel can't use for them.
- **HTTP status + WeCom `errcode`** are checked on upload / gettoken / external-send.
- Added tests: req_id threading, external-attachment rejection, HTTP-mock status/errcode, concurrent token-cache recheck.

One item left for a maintainer/author call: a `TODO(review)` at `resolve_access_token` flags that the qyapi `corpid`/`corpsecret` are sourced from the WS aibot `bot_id`/`secret` — please confirm that's the intended credential namespace for the upload/external-message APIs.

Local validation (toolchain-pinned): `cargo fmt --all -- --check` clean; `cargo clippy -p zeroclaw-channels --features channel-wecom-ws --all-targets -- -D warnings` clean; `cargo test -p zeroclaw-channels --features channel-wecom-ws` → 1044 passed, 0 failed.

Audacity88:
Thanks for pushing this. The compile/test cleanup and the `req_id` / token-cache fixes look good, and CI is green.

Before I would treat this as ready, I think the remaining credential question needs to be resolved in the PR. The new upload / external-send path calls `gettoken` as `corpid={bot_id}&corpsecret={secret}`, but the current `channels.wecom_ws` schema describes those fields as the AI Bot WebSocket subscription `bot_id` / `secret`. If WeCom expects those to be different app or corp credentials in real deployments, proactive media / external sends could pass tests but fail for operators.

Could you either confirm from the WeCom docs or a live config that this is the intended credential namespace, or adjust the config/wiring to use distinct fields? While touching the PR body, please also align the rollback wording with the current `risk: medium` label and note whether #7824's docs/examples acceptance item is intentionally deferred or covered elsewhere.

singlerider (dismissed):
I pulled this branch locally and read it against source truth (the `WeComWsConfig` schema, the `MediaKind` enum, and the linked issue #7824's acceptance criteria), then ran the wecom suite and clippy with `--features channel-wecom-ws`. Clippy is clean and all 61 wecom tests pass. The orchestration is solid: the `req_id` threading fix is correct, the token cache with TTL+skew refreshes outside the lock and re-checks on failure, and the external-attachment path is rejected early instead of being silently routed over a WS channel that cannot carry it. The HTTP-status + `errcode` checking via `parse_wecom_json_response` is the right shape and well tested.

The feature does not ship, though. The credential wiring rests on an assumption the author flagged himself with a `TODO(review)` and that @Audacity88 already raised: the `gettoken` call uses the WebSocket subscription credentials as if they were corp app credentials. That is the load-bearing defect and it gates the whole proactive path.

### Blocking - gettoken sources corpid/corpsecret from the WS subscription credentials

`fetch_access_token` builds:

```
gettoken?corpid={bot_id}&corpsecret={secret}
```

But the schema is explicit that these are a different namespace:

- `bot_id`: "Bot ID for WeCom WebSocket subscription."
- `secret`: "Secret for WeCom WebSocket subscription authentication."

The WeCom AI Bot WS subscription `bot_id`/`secret` are not the qyapi `corpid`/`corpsecret` used by `gettoken`. Issue #7824's own acceptance criteria says auth must be handled "via WeCom app credentials," which are corp application credentials, not the WS subscription pair. The tests pass only because they mock the HTTP layer; a real operator will get a `gettoken` failure on every proactive send. The author's own `TODO(review)` at `resolve_access_token` flags exactly this. A TODO that gates the feature's correctness cannot land. Either source the corp app credentials from distinct config fields, or confirm against the live WeCom docs that the WS pair really is accepted by `gettoken` and remove the TODO with that evidence in the PR.

### Blocking - acceptance criterion "Add documentation and examples" is unmet

#7824 lists documentation and `channel send` usage examples as an explicit acceptance item. This PR is docs-zero. The proactive/external/media surface is new operator-facing behavior (the `external--` scope prefix, the media types, the credential requirements) and needs a docs entry and at least the `zeroclaw channel send` examples from the issue. If docs are intentionally deferred to a follow-up, say so in the PR body and link the tracking item; do not leave the criterion silently unmet.

### Blocking - the kind-to-wire mapping is duplicated inside one function

`send_attachment` writes the same `match attachment.kind()` mapping twice, once for the upload `type=` query param and once for the `msgtype`:

```rust
media_type = match attachment.kind() {
    MediaKind::Image => "image",
    MediaKind::Audio => "voice",
    MediaKind::Video => "video",
    MediaKind::Unknown => "file",
}
...
let msgtype = match attachment.kind() {
    MediaKind::Image => "image",
    MediaKind::Audio => "voice",
    MediaKind::Video => "video",
    MediaKind::Unknown => "file",
};
```

These must stay identical for the upload and the send to agree, and there is no test pinning that they do. The next person who edits one arm and not the other ships a media upload whose declared type disagrees with the dispatched `msgtype`. Pull this into one small fn (`fn wecom_media_wire(kind: MediaKind) -> &'static str`) and call it in both places so the invariant is structural, not vigilance.

### Green - the concurrency and ack-matching work is genuinely good

The `req_id` threading fix (one id into both the frame and the waiter) is the correct mirror of the markdown-send path, and `attachment_send_command_threads_req_id_into_frame` pins it. `resolve_access_token` does the right thing under contention: no lock held across `.await`, refresh outside the cache lock, and a re-check of the cache after a failed `gettoken` so a concurrent caller's fresh token wins instead of surfacing a spurious rate-limit error. Both branches of that recovery are tested. The early rejection of external-scope attachments before any upload is the right call and is covered from both `send_attachment` and the `send` entry point.

While you are in the PR body, align the rollback wording with the current `risk: medium` label.

crates/zeroclaw-channels/src/wecom_ws.rs
```diff
@@ -1431,9 +1450,177 @@ impl WeComWsChannel {
...
+            .filter(|cached| cached.is_fresh(now))
+            .map(|cached| cached.token.clone())
+    }
+
+    fn access_token_after_refresh_error(&self, err: anyhow::Error) -> Result<String> {
+        self.fresh_cached_access_token(Instant::now()).ok_or(err)
+    }
+
+    async fn fetch_access_token(&self) -> Result<CachedAccessToken> {
+        // TODO(review): Confirm that channel bot_id/secret are the qyapi
```
singlerider: This is the load-bearing TODO. `corpid={bot_id}&corpsecret={secret}` reuses the WS subscription credentials for the qyapi `gettoken` corp-app endpoint. The schema documents `bot_id`/`secret` as the WebSocket subscription pair, and #7824 requires auth via WeCom app (corp) credentials, which are a different namespace. Resolve this with distinct config fields or remove the TODO with live-doc evidence that the WS pair is accepted here. Tests pass only because the HTTP layer is mocked.

perlowja:
Fixed all three in 8cf6b521d. @singlerider:

1. **gettoken credentials** — you were right that the load-bearing defect was using the WS-subscription `bot_id`/`secret` as qyapi `corpid`/`corpsecret`. Added distinct `corp_id`/`corp_secret` config fields (the qyapi corp-application credentials), resolved **live** via a config resolver wired from the orchestrator (same pattern as the peer resolvers). `fetch_access_token` now uses those, and returns a clear error — `set channels.wecom_ws.<alias>.corp_id and corp_secret` — when they're unset instead of silently falling back to the WS pair. The `TODO(review)` is gone, and there are mock-HTTP tests for both the corp-cred gettoken call and the missing-cred error.
2. **Docs** — added `docs/book/src/channels/wecom-ws.md` (registered in SUMMARY): documents both credential pairs (WS-subscription vs corp-app), the `external--` scope prefix, the media types, and the `zeroclaw channel send` examples from #7824.
3. **DRY** — extracted `wecom_attachment_wire_kind` so the upload `type=` and the `msgtype` share one kind→wire mapping.

Validation: 63/63 wecom tests pass (`--features channel-wecom-ws`); full-workspace fmt/clippy `ci-all`/`check --locked` green. Re-requesting.

perlowja:
Closing to clear the upstream branch list: this branch was pushed to the canonical repo by mistake (inverted `origin`/fork remotes). The change is preserved on my fork as `perlowja:feat/wecom-proactive-attachments` and will be resubmitted as a proper fork-based PR. A fork-only pre-push guard is now in place so this cannot recur. Apologies for the upstream-branch pollution.

PR #8027, opened 2026-06-19
feat(channels/qq): thread passive group replies via refer_msg

### LINKED ISSUE #7872: [Tracker]: QQ group replies need msg_id for passive reply sends
## Summary

QQ group replies should include the triggering message `msg_id` so the QQ Bot API treats them as passive replies instead of active messages.

## Current status

- #9180 merged the QQ text/media reply propagation and deterministic production request-builder coverage.
- Keep this tracker open until a live QQ group reply confirms ordinary passive sends no longer fail with API code `40034102`.

## Problem

PR #6737 described a QQ channel failure where group replies were rejected by the QQ Bot API with code `40034102` and the message shape for active-message permission failure.

The PR body described the root cause as:

- incoming QQ events include a message id
- the channel was using that id for deduplication but not preserving it for replies
- outgoing group replies without `msg_id` are treated as active messages
- active messages require a separate Tencent permission grant

That PR closed as stale before the fix landed. Duplicate search found the old closed issue #1714 and broad tracker #6970, but no current narrow issue for the QQ passive-reply `msg_id` behavior.

## Expected behavior

When ZeroClaw replies to a QQ group message, the outgoing request should carry the triggering `msg_id` when the QQ API requires it for passive replies.

Users should not need active-message permissions for ordinary replies to inbound QQ group messages.

## Scope

In scope:

- QQ channel reply path
- storage or propagation of the latest relevant `msg_id` needed for passive replies
- regression coverage for group reply request bodies

Out of scope:

- broad QQ channel redesign
- changes to the generic `Channel` trait unless a fresh implementation proves that is necessary
- private-message behavior unless it shares the same request-building helper

## Reproduction sketch

Configure QQ group messaging and trigger a normal inbound group message. Have ZeroClaw send a reply through the current QQ channel path.

The bug shape is a rejected send with QQ API code `40034102` / active-message permission failure because the request is missing the passive-reply message id.

This needs a fresh reproduction or request-body regression test against current `master` before closing the issue.

## Related

- #6737
- #1714
- #6970

## Acceptance criteria

- QQ group reply requests include the correct triggering `msg_id` when required by the QQ API.
- Ordinary passive replies do not require active-message permissions.
- Regression coverage proves the request body includes `msg_id` for group replies.
- The implementation avoids stale or cross-conversation `msg_id` reuse.

## Triage fields

- Affected component: channel
- Severity: S1 - workflow blocked
- ZeroClaw version: observed in PR #6737; current `master` needs fresh verification
- Operating system: platform-independent
- Regression status: unknown

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:** QQ group passive replies thread via `refer_msg` (in_reply_to). The code compiled and linted clean; it was committed unformatted and failed the `Format` gate. Applied `rustfmt`; no behavioral change.
- **Scope boundary:** QQ channel only.
- **Blast radius:** `zeroclaw-channels` qq path only.
- **Linked issue(s):** Resolves #7872

Validated against the **actual CI gate** (`ci-all`, toolchain `1.93.0`):

```
cargo +1.93.0 fmt --all -- --check                                                  -> PASS
cargo +1.93.0 check  --locked --features ci-all                                     -> PASS
cargo +1.93.0 clippy --workspace --exclude zeroclaw-desktop --all-targets \
      --features ci-all --locked -- -D warnings                                     -> PASS
cargo +1.93.0 test --test architecture (config-isolation + fluent)                  -> PASS
```
GitHub `Test` (full suite) + `CI Required Gate`: green.

- New permissions / fs access? `No`
- Secrets / credentials handling changed? `No`
- PII in diff/tests/fixtures/docs? `No`

## Compatibility (required)

- Backward compatible? `Yes`

Low risk: `git revert <sha>`.

### DIFF
--- a/crates/zeroclaw-channels/src/qq.rs
+++ b/crates/zeroclaw-channels/src/qq.rs
@@ -947,7 +947,8 @@ impl QQChannel {
                     file_name.as_deref(),
                 )
                 .await?;
-            self.send_media_message(recipient, &file_info, in_reply_to).await?;
+            self.send_media_message(recipient, &file_info, in_reply_to)
+                .await?;
         } else {
             // Local file upload
             let path = Path::new(target);
@@ -1003,7 +1004,8 @@ impl QQChannel {
                     .await;
             }
 
-            self.send_media_message(recipient, &file_info, in_reply_to).await?;
+            self.send_media_message(recipient, &file_info, in_reply_to)
+                .await?;
         }
 
         Ok(())
@@ -1320,7 +1322,10 @@ impl Channel for QQChannel {
 
         // Send each media attachment
         for attachment in &attachments {
-            if let Err(e) = self.send_attachment(&message.recipient, attachment, None).await {
+            if let Err(e) = self
+                .send_attachment(&message.recipient, attachment, None)
+                .await
+            {
                 ::zeroclaw_log::record!(WARN, ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note).with_outcome(::zeroclaw_log::EventOutcome::Unknown).with_attrs(::serde_json::json!({"target": attachment.target, "error": format!("{}", e)})), "failed to send media attachment; falling back to text");
                 // Degrade to text fallback
                 let fallback = format!(
--- a/crates/zeroclaw-channels/src/qq.rs
+++ b/crates/zeroclaw-channels/src/qq.rs
@@ -1322,52 +1322,99 @@ impl ::zeroclaw_api::attribution::Attributable for QQChannel {
 }
 
 #[async_trait]
+trait QQOutboundSender {
+    async fn send_text_markdown(
+        &self,
+        recipient: &str,
+        content: &str,
+        in_reply_to: Option<&str>,
+    ) -> anyhow::Result<()>;
+
+    async fn send_attachment(
+        &self,
+        recipient: &str,
+        attachment: &QQMediaAttachment,
+        in_reply_to: Option<&str>,
+    ) -> anyhow::Result<()>;
+}
+
+#[async_trait]
+impl QQOutboundSender for QQChannel {
+    async fn send_text_markdown(
+        &self,
+        recipient: &str,
+        content: &str,
+        in_reply_to: Option<&str>,
+    ) -> anyhow::Result<()> {
+        QQChannel::send_text_markdown(self, recipient, content, in_reply_to).await
+    }
+
+    async fn send_attachment(
+        &self,
+        recipient: &str,
+        attachment: &QQMediaAttachment,
+        in_reply_to: Option<&str>,
+    ) -> anyhow::Result<()> {
+        QQChannel::send_attachment(self, recipient, attachment, in_reply_to).await
+    }
+}
+
+async fn send_qq_message_with_sender(
+    sender: &impl QQOutboundSender,
+    message: &SendMessage,
+) -> anyhow::Result<()> {
+    let (cleaned_text, attachments) = parse_qq_attachment_markers(&message.content);
+    let in_reply_to = message.in_reply_to.as_deref();
+
+    if attachments.is_empty() {
+        // No media markers — send as markdown (original path)
+        return sender
+            .send_text_markdown(&message.recipient, &message.content, in_reply_to)
+            .await;
+    }
+
+    // Send cleaned text first (if non-empty)
+    if !cleaned_text.is_empty() {
+        sender
+            .send_text_markdown(&message.recipient, &cleaned_text, in_reply_to)
+            .await?;
+    }
+
+    // Send each media attachment
+    for attachment in &attachments {
+        if let Err(e) = sender
+            .send_attachment(&message.recipient, attachment, in_reply_to)
+            .await
+        {
+            ::zeroclaw_log::record!(WARN, ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note).with_outcome(::zeroclaw_log::EventOutcome::Unknown).with_attrs(::serde_json::json!({"target": attachment.target, "error": format!("{}", e)})), "failed to send media attachment; falling back to text");
+            // Degrade to text fallback
+            let fallback = format!(
+                "{}: {}",
+                match attachment.kind {
+                    QQMediaFileType::Image => "Image",
+                    QQMediaFileType::Video => "Video",
+                    QQMediaFileType::Voice => "Voice",
+                    QQMediaFileType::File => "File",
+                },
+                attachment.target
+            );
+            sender
+                .send_text_markdown(&message.recipient, &fallback, in_reply_to)
+                .await?;
+        }
+    }
+
+    Ok(())
+}
+
+#[async_trait]
 impl Channel for QQChannel {
     fn name(&self) -> &str {
         "qq"
     }
 
     async fn send(&self, message: &SendMessage) -> anyhow::Result<()> {
-        let (cleaned_text, attachments) = parse_qq_attachment_markers(&message.content);
-        let in_reply_to = message.in_reply_to.as_deref();
-
-        if attachments.is_empty() {
-            // No media markers — send as markdown (original path)
-            return self
-                .send_text_markdown(&message.recipient, &message.content, in_reply_to)
-                .await;
-        }
-
-        // Send cleaned text first (if non-empty)
-        if !cleaned_text.is_empty() {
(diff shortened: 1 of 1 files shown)


crates/zeroclaw-channels/src/qq.rs
```diff
@@ -1278,23 +1305,27 @@ impl Channel for QQChannel {
...
-            self.send_text_markdown(&message.recipient, &cleaned_text)
+            self.send_text_markdown(&message.recipient, &cleaned_text, in_reply_to)
                 .await?;
         }
 
         // Send each media attachment
         for attachment in &attachments {
-            if let Err(e) = self.send_attachment(&message.recipient, attachment).await {
+            if let Err(e) = self
+                .send_attachment(&message.recipient, attachment, None)
```
singlerider: Threading is dropped here while the text paths thread. `send_attachment` already forwards `in_reply_to` through to all three `send_media_message` calls, so this is the one place severing it. Pass `message.in_reply_to.as_deref()` like the markdown path does, or add a one-line comment explaining why media replies intentionally do not thread (e.g. if QQ rejects `refer_msg` on `msg_type=7`).

singlerider (dismissed):
I pulled this branch locally, built it with `--features channel-qq`, ran the QQ suite (50 passed), and ran clippy with the feature enabled (clean). I traced every consumer of the field this PR repurposes before forming a verdict.

## Context

The stated feature (`refer_msg` threading on outbound QQ replies) is correct: QQ's send API does thread via `refer_msg.message_id`, and the body construction in `send_media_message` / `send_text_markdown` matches that contract.

The load-bearing change is the quieter one: `ChannelMessage.id` moves from `Uuid::new_v4()` to the platform `msg_id` at both the C2C and group inbound sites. I verified this is what makes threading function end to end: `id` flows through `SendMessage::reply_to` (`crates/zeroclaw-api/src/channel.rs:181`) and the orchestrator send path (`crates/zeroclaw-channels/src/orchestrator/mod.rs:2478`) into `in_reply_to`, which `send()` then stamps as `refer_msg.message_id`. With the old UUID, `refer_msg.message_id` would have been a value QQ never issued, so threading would have silently no-opped. The change is right.

I checked the other `msg.id` consumers for fallout:
- `conversation_history_key` does not use `msg.id` (keyed on reply_target + sender), so real conversation history is unaffected.
- `conversation_memory_key` includes `msg.id`, but QQ has `thread_ts: None` and the value was already unique per message before and after, so behavior does not change.
- dedup uses `is_duplicate(msg_id)` on a separate path, unaffected.

No regression in the `id` change. Verified, not assumed.

### Blocking - attachment reply path drops threading

`send()` threads text replies but passes `None` for attachment replies:

```
self.send_attachment(&message.recipient, attachment, None)
```

The plumbing is fully built: `send_attachment` accepts `in_reply_to` and forwards it through all three `send_media_message` call sites. Only this one caller severs it. The result is asymmetric: a text reply threads, a reply that carries media does not. Either wire `message.in_reply_to.as_deref()` through here like the text path does, or if the omission is deliberate (e.g. QQ rejects `refer_msg` on `msg_type=7`), say so in a one-line comment so the next reader does not "fix" it back.

### Blocking - feature ships with zero test coverage

There is no test that asserts `refer_msg` is built when `in_reply_to` is set, no test for the `id`-from-`msg_id` change, and nothing pinning the attachment asymmetry as intended. This is a behavior change to both message threading and the `ChannelMessage.id` contract, landing untested. The QQ module already has 50 unit tests and a clear pattern to follow; add coverage for the threading body construction and for the inbound `id` assignment so a future refactor cannot silently revert either.

### Praise

The threading plumbing through `send_attachment` to all three `send_media_message` paths is complete and consistent, which is exactly why the single `None` at the call site stands out rather than hiding. Repurposing `id` to the platform message id (instead of inventing a parallel field) is the right call and keeps the existing `reply_to` machinery doing the work.

perlowja:
Thanks @singlerider — both blockers addressed, and thanks for verifying the `id`-change fallout rather than assuming it.

**Attachment reply path dropped threading:** fixed. `send()` now passes the same `in_reply_to` to the attachment path as the text path:
```
-                .send_attachment(&message.recipient, attachment, None)
+                .send_attachment(&message.recipient, attachment, in_reply_to)
```
So a reply carrying media now threads symmetrically with a text reply (it was a plain oversight — the plumbing through `send_attachment` → all three `send_media_message` sites was already complete, which is why the lone `None` stood out).

**Zero test coverage:** added. To make the `refer_msg` construction actually testable (it was inline json in `send_media_message` / `send_text_markdown`), I extracted two pure builders — `build_media_message_body` and `build_markdown_message_body` — and wired them into the production send paths (byte-identical payloads). New tests:
- `test_send_media_body_includes_refer_msg_when_replying` / `..._omits_refer_msg_without_reply`
- `test_send_markdown_body_includes_refer_msg_when_replying` / `..._omits_refer_msg_without_reply`
- `test_channel_message_id_uses_qq_msg_id` — pins the inbound `ChannelMessage.id` to the QQ `msg_id` via an extracted `channel_message_id_from()` used by both inbound handlers.

(One catch worth noting: the body builders initially took `u64` while `next_msg_seq()` returns `u32` — it compiled under default features but failed under `--features ci-all`, which is what CI's Lint/Check jobs use. Caught and fixed locally before pushing.)

Validation: rebased on latest `master`, full gate at 1.93.0 — clippy `--features ci-all`, `check --locked` (ci-all + no-default), both arch guards, nextest (9342 pass).

Audacity88 (dismissed):
Reviewed head `5fc3ae5` against the live PR diff, PR body, visible checks, linked issue #7872, @singlerider's earlier review, and the current channel/reply source context. Line references below are from the current PR head. I did not run local Cargo validation or live QQ smoke testing.

### ✅ Resolved — The main success paths and unit coverage are now in place

I agree with @singlerider's earlier read that using QQ's platform `msg_id` as `ChannelMessage.id` is the right way to make the existing `SendMessage::reply_to()` machinery produce a real QQ `refer_msg.message_id`. The latest head also fixes the successful media path by passing `in_reply_to` through `send_attachment()`, and the new helper-level tests cover `refer_msg` on both media and markdown bodies.

### 🔴 Blocking — Attachment fallback still drops the passive-reply id

The successful attachment path now preserves threading, but the failure fallback still sends fallback text with `None` for the reply id. `send()` captures `message.in_reply_to` and passes it into `send_attachment()` (`crates/zeroclaw-channels/src/qq.rs:1332`, `src/qq.rs:1350`), but if attachment upload or send fails, the fallback call is:

```rust
self.send_text_markdown(&message.recipient, &fallback, None)
```

at `crates/zeroclaw-channels/src/qq.rs:1365`.

That fallback is still a user-visible outbound QQ message sent through the same message endpoint, not a local-only log. For a QQ group passive reply whose attachment upload fails, the fallback text still goes out without `refer_msg`, so it keeps the same active-message shape #7872 is trying to remove. Please pass the same `in_reply_to` into the fallback text send, and add coverage for the fallback path or explicitly document why QQ must not receive `refer_msg` there.

### 🟢 What looks good — The fix avoids duplicate reply state

This keeps the reply id flowing through the existing channel abstraction instead of adding a QQ-only side channel or cached "last message id" state. That is the cleaner shape: the inbound message carries the platform id, `reply_to()` maps it into `in_reply_to`, and the QQ sender stamps it into `refer_msg` only when it is actually replying.

perlowja:
Fixed in 20bc5b38e. @Audacity88 you were right — the attachment-failure fallback was the one path still sending `None`; it now passes the in-scope `in_reply_to`, so a passive group reply whose upload fails still carries `refer_msg` instead of reverting to the active shape #7872 removes. @singlerider this is the same severing point your inline note flagged.

For the fallback-path coverage Audacity88 asked for, I added a small `QQOutboundSender` seam — `QQChannel` implements it by delegating to its existing inherent `send_text_markdown`/`send_attachment` (fully-qualified, no recursion), so **runtime behavior is unchanged**; it just lets the regression inject a sender whose attachment send fails and assert the fallback still carries the reply id. `test_attachment_fallback_send_preserves_in_reply_to` fails on the old `None` and passes now. QQ suite green (56/56 with `--features channel-qq`); full-workspace fmt/clippy `ci-all`/check/nextest green. Re-requesting.

perlowja:
Closing to clear the upstream branch list: this branch was pushed to the canonical repo by mistake (inverted `origin`/fork remotes). The change is preserved on my fork as `perlowja:feat/qq-passive-group-replies` and will be resubmitted as a proper fork-based PR. A fork-only pre-push guard is now in place so this cannot recur. Apologies for the upstream-branch pollution.

PR #8028, opened 2026-06-19
chore(hardware): gate aardvark-sys behind the hardware feature

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - **Gate `aardvark-sys` behind the `hardware` feature** in `zeroclaw-hardware` (`optional = true` + `hardware = ["dep:aardvark-sys", …]`). Every use of it was already `#[cfg(feature = "hardware")]` (the `aardvark`/`aardvark_tools` modules and the `find_devices()` call in `boot()`), but the dependency itself was non-optional, so it compiled into *every* default build via `gateway → hardware → aardvark-sys`. It (and its 86 KB vendored `.so`) is now absent from default builds.
  - **Drop the unused direct `aardvark-sys` deps** from `zeroclaw-runtime` and the `zeroclawlabs` binary — neither references it (zero `aardvark` symbols in their `src/`).
  - **Delete the no-op `crates/aardvark-sys/build.rs`** — it documented a bindgen flow the crate never uses; FFI symbols are resolved at runtime via `libloading`, so it only forced cargo to compile+run a build script for nothing.
  - **Fix the stale `aardvark-sys/Cargo.toml` comment** that claimed the SDK was "NOT yet committed" / the crate "a stub returning NotFound" — both untrue (`vendor/aardvark.{h,so}` are committed; calls dispatch through the loaded library).
- **Scope boundary:** Does NOT remove Aardvark support or change any runtime behavior; `--features hardware` is functionally identical. No publish-topology or crate-structure changes (a follow-up may fold `aardvark-sys` into `zeroclaw-hardware`).
- **Blast radius:** Default builds no longer compile `aardvark-sys`; `--features hardware` unchanged. Manifest-only for `zeroclaw-runtime` and the binary.
- **Linked issue(s):** None.
- **Labels:** `risk: low`, `size: XS`

```
$ cargo fmt --all -- --check                                   → clean (no diff)
$ cargo tree -e no-dev -i aardvark-sys                         → error: "did not match any packages"  (gone from default ✓)
$ cargo tree -e no-dev -i aardvark-sys --features hardware     → aardvark-sys → zeroclaw-hardware → gateway → zeroclawlabs  (present ✓)
$ cargo check                          # default binary        → Finished; aardvark-sys not built
$ cargo check --features hardware                              → Finished
$ cargo clippy -p zeroclaw-hardware --all-targets -- -D warnings                  → Finished, no warnings
$ cargo clippy -p zeroclaw-hardware --all-targets --features hardware -- -D warnings → Finished, no warnings
$ cargo test  -p zeroclaw-hardware                            → ok. 82 passed; 0 failed; 1 ignored
$ cargo test  -p zeroclaw-hardware --features hardware        → ok. 139 passed; 0 failed; 1 ignored
```
rustc 1.93.1 / cargo 1.93.1
- **Beyond CI — what did you manually verify?** Confirmed via `cargo tree -e no-dev -i aardvark-sys` that the dep is dropped from the default tree entirely while the hardware path keeps it. Verified `./src` and `zeroclaw-runtime/src` contain zero `aardvark` references before removing those deps, and that all `aardvark_sys::` call sites in `zeroclaw-hardware` are behind `#[cfg(feature = "hardware")]`.
- **If any command was intentionally skipped, why:** `clippy`/`test` were scoped to the affected crate in both feature configs (full `fmt --all` ran tree-wide). No shared logic was altered — the change is manifest + comments + one dead no-op file — so a full-workspace `cargo test` was not run.

## Compatibility (required)

- Backward compatible? `Yes`
- Upgrade steps: none — no public API, feature, or behavior change. Default builds simply stop pulling an unused, hardware-only dependency; `--features hardware` is unchanged.

Low risk — `git revert <sha>` is the plan.


singlerider (approved):
I pulled this locally and verified every load-bearing claim against source and a live compile rather than the pasted evidence. It holds up.

### 🟢 Praise - dependency was genuinely inert in default builds

Confirmed the problem is real: on master `aardvark-sys` was a non-optional dep of `zeroclaw-hardware`, so it compiled into every default build through `gateway -> hardware`, despite every use site being hardware-only. After this change `cargo tree -e no-dev -i aardvark-sys` errors with "did not match any packages" on the default tree and resolves to `aardvark-sys -> zeroclaw-hardware -> gateway -> zeroclawlabs` under `--features hardware`. Exactly the claimed topology.

### 🟢 Praise - gating is provably correct, not just asserted

The risky question was whether dropping the non-optional dep would break the default build, since `boot()` references `aardvark::AardvarkTransport` and `aardvark_sys::AardvarkHandle` unconditionally inside its body. Verified there are two `boot()` definitions: the one carrying the aardvark discovery block is `#[cfg(feature = "hardware")]` (lib.rs:268-270), and a separate `#[cfg(not(feature = "hardware"))]` fallback (lib.rs:433-435) is used otherwise. The `pub mod aardvark`, `pub mod aardvark_tools`, and `pub use aardvark::AardvarkTransport` are all hardware-gated too. So the default build never names the symbol. `cargo check -p zeroclaw-hardware` and `cargo check -p zeroclaw-hardware --features hardware` both finish clean here.

### 🟢 Praise - the runtime/binary dep drops are safe

`git grep -in aardvark` over `crates/zeroclaw-runtime/src` and the binary `src` returns zero hits, and `cargo check -p zeroclaw-runtime` finishes clean with the dep removed. Those manifest entries were dead weight; removing them is correct.

### 🟢 Praise - build.rs deletion and comment fix are accurate

The deleted `build.rs` `main()` body was empty (comments only), so it forced cargo to compile and run a no-op build script for nothing. Deleting it is right. The corrected `Cargo.toml` comment now matches reality: symbols resolve at runtime via `libloading`, the vendored `.so` is committed, and the prior "SDK NOT yet committed / stub returning NotFound" note was stale.

`cargo fmt --all -- --check` is clean. Manifest + comments + one dead no-op file, no behavior change, `--features hardware` functionally identical. Approving.

PR #8029, opened 2026-06-19
fix(providers): distinguish missing vs expired OpenAI Codex credentials

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:** The call-time OpenAI Codex credential resolver returned a single error — "OpenAI Codex auth profile not found" — for two distinct states: (a) no profile exists at all, and (b) a profile exists but its access token is expired and could not be refreshed. Case (b) misattributed a re-authentication problem to a missing profile, sending users into a login loop pointed at the wrong cause. The error now branches on whether a profile actually loaded: no profile → "No OpenAI Codex credentials found … sign in"; stale token → "credentials are present but expired or could not be refreshed … re-run … to sign in again".
- **Scope boundary:** Error/log message text only. No change to credential-resolution control flow, token refresh, or the custom-endpoint gateway-API-key path. No config, schema, or public API changes.
- **Blast radius:** Only the error surfaced when an OpenAI Codex (`requires_openai_auth`) slot cannot produce an access token. No other providers affected.
- **Linked issue(s):** none
- **Labels:** maintainer to set `risk:`/`size:` per template (suggest `risk: low`, `size: XS`).

## Validation Evidence (required)

- **Commands run and tail output** (scoped to the only crate touched, `zeroclaw-providers`):
  - `cargo fmt --all -- --check` → clean (whole repo)
  - `cargo clippy -p zeroclaw-providers --all-targets -- -D warnings` → `Finished` with no warnings
  - `cargo test -p zeroclaw-providers openai_codex` → `test result: ok. 38 passed; 0 failed; 0 ignored`
- **Beyond CI — what I manually verified:** confirmed `had_profile` is captured before `profile` is moved into the `account_id` computation; both branches retain the same `zeroclaw auth login --provider openai-codex` recovery command, only the framing differs; the added log attr is a boolean (`had_profile`), never a token value.
- **If any command was intentionally skipped, why:** full-workspace `cargo test` / `cargo clippy --all-targets` were scoped to `zeroclaw-providers` — localized, message-only change to one function. The required CI gate runs the full battery.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? `No`
- New external network calls? `No`
- Secrets / tokens / credentials handling changed? `No` — only the wording of an existing "credential missing/expired" error; no credential value is read, logged, or stored. The structured log gains a boolean `had_profile`, not any token material.
- PII, real identities, or personal data? `No`

## Compatibility (required)

- Backward compatible? `Yes`
[...]

PR #8030, opened 2026-06-19
feat(doctor): warn on OpenAI Codex profile/slot wiring mismatch

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:** `zeroclaw doctor` now cross-checks signed-in OpenAI Codex (OAuth/subscription) credentials against the OpenAI provider slots that opt into them via `requires_openai_auth`. Two states that previously stayed silent until the first model call — where they surfaced as a confusing auth error — are now reported: a Codex credential is signed in but no slot references it, and a slot requires Codex auth but no credential is signed in. Closes the discoverability gap where importing a token did nothing visible and a mis-wired slot only failed at call time.
- **Scope boundary:** Read-only diagnostic. Does not write config, does not create/modify auth profiles, does not touch `auth login` or `quickstart`. The sync `diagnose()` consumed by the gateway API is unchanged — the auth-store load is async and appended in `run`, mirroring the existing model-catalog probe (`probe_models`).
- **Blast radius:** Adds a `[providers.auth]` section to `zeroclaw doctor` output and loads the auth profile store from disk during `doctor`. Silent when Codex is unused on both sides (no noise for the common case).
- **Linked issue(s):** none
- **Labels:** enhancement, doctor, provider, runtime, risk:high, size:M.

- **Commands run and tail output** (scoped to the affected crate, `zeroclaw-runtime`):
  - `cargo fmt --all -- --check` → clean (whole repo)
  - `cargo clippy -p zeroclaw-runtime --all-targets -- -D warnings` → `Finished` with no warnings
  - `cargo test -p zeroclaw-runtime --lib doctor` → `test result: ok. 32 passed; 0 failed` (includes the 4 new `codex_wiring_*` tests)
- **Beyond CI — what I manually verified:** the pure comparator (`codex_auth_wiring_items`) covers all four (profile, slot) quadrants, including staying silent when Codex is unused; the async loader returns no diagnostics on an unreadable store rather than a false "no credentials" warning; the new Fluent keys are added to `en` only and other locales fall back to English by design (`en` is the embedded source of truth) until translated. No persistent state is added — profiles and slots are read on demand (single-source-of-truth rule).
- **If any command was intentionally skipped, why:** full-workspace `cargo test` / clippy were scoped to `zeroclaw-runtime`. The required CI gate runs the full battery.

- New permissions, capabilities, or file system access scope? `No` new scope. The check reads the existing auth profile store via `AuthService::list_profile_ids`, a read-only listing that parses the on-disk store under the shared lock but never decrypts token fields and never migrates or writes the payload back. `doctor` simply did not read it before.
- Secrets / tokens / credentials handling changed? `No`. The check inspects only profile *ids* (e.g. `openai-codex:default`) to test presence via the read-only `list_profile_ids` path; it never decrypts, reads, or emits token values, and cannot trigger the `enc:` to `enc2:` migration writeback that the full `load_profiles` path performs.
- PII, real identities, or personal data? `No`

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `Yes` (additive only) — new diagnostic lines in `zeroclaw doctor` output and three new Fluent strings. No new flags, config keys, or behavior changes to existing checks.
- Upgrade steps: none.

Low risk: `git revert <sha>`.


WareWolf-MoonWall (changes requested):
Reviewed at 941f3d572ebc1ca89c6b738a147daddfcdc8e503. The core doctor check is well-built — it's the undisclosed CI workflow additions that I can't approve over.

### 🔴 Blocking — `sync-upstream.yml` is fork-specific infrastructure that must not be in this PR

The diff contains four files that are entirely absent from the PR body and scope description:

1. `.github/workflows/sync-upstream.yml` — **fork-specific**. Its own comment says: *"Keeps this fork's `master` current with `zeroclaw-labs/zeroclaw` by merging upstream in daily."* If this lands in the upstream repo it adds a daily scheduled workflow that tries to merge `zeroclaw-labs/zeroclaw` into `zeroclaw-labs/zeroclaw` — a circular no-op with `permissions: contents: write` and a `git push origin HEAD:master` push step. That's a write-to-master scheduled job that nobody asked for and the PR body does not mention.

2. `.github/workflows/clawpatch-review.yml` — adds a self-review CI job that pulls a prebuilt `zeroclaw` binary from GitHub releases and runs `npm install -g 'acpx@^0.8.0' clawpatch`. This may be intentional for the upstream repo, but it is completely undisclosed; the PR title is "feat(doctor): warn on OpenAI Codex profile/slot wiring mismatch" and the scope boundary says "Read-only diagnostic."

3. `.github/clawpatch/factory-zc.config.toml` — ACP brain config that splices in `ANTHROPIC_OAUTH_TOKEN` from repo secrets at CI time.

4. `clawpatch.config.json` — root-level clawpatch configuration.

`sync-upstream.yml` needs to be removed from this PR immediately — it is fork tooling that does not belong in the upstream repository. The clawpatch files are a separate question (possibly intentional) but they require an explicit scope section in the PR body and a separate Security & Privacy discussion covering the secret splicing pattern and the scoped `contents: read` vs the npm supply chain (semver ranges, not pinned digests).

Please either remove the workflow files and land the doctor check independently, or open a separate PR specifically for the clawpatch CI addition with full template coverage.

### 🟡 Warning — missing `risk:` and `size:` labels

The PR body correctly flags this as a maintainer task. As reviewer, I'll note: the doctor check itself warrants `risk: low, size: S`. If the CI workflow additions land in the same PR, the risk classification needs to account for the write-permissions workflow.

### 🟢 What looks good — single-source-of-truth config reads

`codex_auth_wiring_items(codex_profile_present: bool, config: &Config)` is pure — it takes the already-resolved boolean plus the live `&Config`, reads slot state on demand from `config.providers.models.openai.iter()`, stores nothing, and returns `Vec<DiagItem>`. No snapshot, no cache, no duplicated state. This is exactly the pattern the codebase requires.

### 🟢 What looks good — all four quadrants covered, silent for the common case

The four unit tests map cleanly to the four `(profile, slot)` combinations: credential with no slot (warn), slot with no credential (warn), both present (ok), neither present (silent). `codex_wiring_silent_when_codex_unused` confirms the common case produces no output — no noise for the vast majority of users who don't touch Codex at all.

### 🟢 What looks good — no network calls at doctor time

`check_codex_auth_wiring` reads from the existing auth profile store on disk (same path as `AuthService::from_config` used by providers). `Err` from an unreadable store silently returns an empty slice rather than emitting a spurious "no credentials" warning. This is the correct graceful-degradation behavior for a diagnostic.

WareWolf-MoonWall:
**Note for @JordanTheJet:** Both #8029 and #8030 include what appears to be the same `sync-upstream.yml` workflow — a fork-sync cron job with `permissions: contents: write` that pushes to `master`. This looks like a fork-maintenance workflow that is leaking into feature branches, likely via rebase from a personal fork's default branch. Please audit your branch base before rebasing future PRs to prevent this from recurring. The fix for both PRs is the same: remove the unrelated workflow files before re-requesting review.

WareWolf-MoonWall:
@singlerider — milestone alignment needed: this feat PR has active CHANGES_REQUESTED and does not clearly fit within the scope boundary of any open milestone. Please advise on placement or deferral once the blockers are addressed.

WareWolf-MoonWall (approved):
# Re-review — #8030 feat(doctor): warn on OpenAI Codex profile/slot wiring mismatch

**Round**: Re-review (WareWolf-MoonWall prior CHANGES_REQUESTED at `941f3d572ebc1ca89c6b738a147daddfcdc8e503`)
**Current head**: post-fix push (SHA not returned in available API responses; clean diff confirmed)
**CI**: 18/18 checks pass on new run (run ID 27901118459)
**Other active CHANGES_REQUESTED**: none — Audacity88, WareWolf-MoonWall (Chris Hengge), singlerider are all listed as Requested with no reviews posted

---

### ✅ Resolved — all four undisclosed CI/workflow files removed from diff

Prior blockers named:
- `.github/workflows/sync-upstream.yml`
- `.github/workflows/clawpatch-review.yml`
- `.github/clawpatch/factory-zc.config.toml`
- `clawpatch.config.json`

None of these appear in the current diff. The PR is now scoped to `crates/zeroclaw-runtime/locales/en/cli.ftl` and `crates/zeroclaw-runtime/src/doctor/mod.rs` only, matching the `feat(doctor):` commit title and the "Read-only diagnostic" scope boundary stated in the PR body.

### ✅ Resolved — user-facing strings use Fluent (`fl!()`)

Three new keys added to `locales/en/cli.ftl`:
- `cli-doctor-codex-auth-profile-no-slot`
- `cli-doctor-codex-auth-slot-no-profile`
- `cli-doctor-codex-auth-ok`

All three are consumed via `crate::i18n::get_required_cli_string` / `get_required_cli_string_with_args` in `doctor/mod.rs`. No bare string literals for user-facing diagnostic output.

---

### 🟡 Warning — labels `risk: high, size: M` remain incorrect; Rollback section is inconsistent

The PR body author note correctly flags this as a maintainer task and suggests `risk: low, size: S`. The applied labels (`risk: high, size: M`) do not reflect the actual scope — this is a read-only diagnostic check: no config writes, no auth profile changes, no new CLI flags, no schema changes. The Rollback section reads "Low risk: `git revert <sha>`." which is correct for the actual change but contradicts the `risk: high` label.

Per project convention, `risk: high` requires **Feature flags** and **Observable failure symptoms** sub-bullets in the Rollback section. A maintainer should correct the labels to `risk: low, size: S` before merge. If `risk: high` is intentionally retained, the Rollback section must be expanded with those sub-bullets.

---

### 🟢 Doctor logic confirmed correct and SSOT-compliant

`codex_auth_wiring_items(codex_profile_present: bool, config: &Config)` is pure: takes an already-resolved `bool` plus the live `&Config`, reads `config.providers.models.openai.iter()` on demand, stores nothing, returns `Vec<DiagItem>`. No snapshot, no cached copy, no duplicated state. This is the pattern the AGENTS.md SSOT rule requires. All prior praise applies.

### 🟢 All four (profile, slot) quadrants covered; common case is silent

- `codex_wiring_warns_when_profile_has_no_slot` — credential imported, no slot opts in (Warn)
- `codex_wiring_warns_when_slot_has_no_profile` — slot opts in, no credential (Warn; slot name appears in message)
- `codex_wiring_ok_when_profile_and_slot_present` — consistent wiring (Ok)
- `codex_wiring_silent_when_codex_unused` — neither side uses Codex (empty Vec, no output)

The common case (no Codex on either side) produces zero diagnostic lines, satisfying the no-noise-for-unaffected-users requirement.

### 🟢 Graceful degradation on unreadable auth store

`check_codex_auth_wiring` returns an empty `Vec` (not a "no credentials" warning) when `auth.load_profiles()` returns `Err`. This is the correct behavior for a diagnostic that cannot substantiate its result.

### 🟢 CI green across all targets

18 checks pass: Format, Lint, Test, Build x86_64, Check aarch64-apple-darwin, Check x86_64-pc-windows-msvc, Check (32-bit), Check (all features), Check (no default features), Benchmarks Compile, Security, Installer Drift, Nix Module Eval, Docs Style, Zerocode RPC Boundary, CI Required Gate, Apply path labels, main.

---

**Verdict: APPROVE** (`--approve`) — the blocking mixed-content issue is resolved, the diff is correctly scoped, CI is green, Fluent compliance confirmed, SSOT rule satisfied, and no other reviewer holds active CHANGES_REQUESTED. Maintainer should correct the `risk:`/`size:` labels before merge.

WareWolf-MoonWall:
Labels note for @singlerider: current labels show `risk: high, size: M` but post-cleanup the diff is doctor-check only. Author suggests `risk: low, size: S` — please adjust before merge.

Audacity88 (changes requested):
I reviewed the current #8030 head `39457d3d6821da3e7a31bd09ed68a8db449383e8`, the PR body, @WareWolf-MoonWall's prior review and approval re-review, top-level comments, inline review state, the live GitHub source for `doctor/mod.rs`, `AuthService`, `AuthProfilesStore`, and `SecretStore`, plus current CI and merge state/freshness. I did not run local Cargo for this review-only pass.

### ✅ Resolved — The unrelated workflow and Clawpatch files are gone

I agree with @WareWolf-MoonWall's re-review that the prior mixed-content blocker is resolved. The current diff is scoped to `crates/zeroclaw-runtime/src/doctor/mod.rs` and `crates/zeroclaw-runtime/locales/en/cli.ftl`, so the fork-sync workflow and Clawpatch files no longer affect this doctor diagnostic review.

### 🔴 Blocking — The read-only doctor check uses a credential load path with write side effects

The new diagnostic only needs a presence bit: whether any stored profile ID starts with `openai-codex:`. But `check_codex_auth_wiring` gets that bit by calling `AuthService::load_profiles()`, which goes through `AuthProfilesStore::load()`. That path decrypts every stored auth secret, and when it sees a legacy `enc:` value it rewrites the persisted auth profile file with the migrated `enc2:` value. The same secret-store path can also create the `.secret_key` file while trying to decrypt encrypted profile data.

That behavior is fine for provider/auth code that is intentionally loading usable credentials, but it is too broad for a `zeroclaw doctor` presence check that the PR describes as read-only and non-mutating. The PR body says this does not create or modify auth profiles and inspects only profile IDs, but the implementation reaches those IDs through the full credential decrypt/migration path. Running doctor should not become the first command that rewrites auth profile storage or creates secret-key material just to decide whether an `openai-codex:*` profile ID exists. It also means an unrelated corrupt or undecryptable stored secret can make this check return no diagnostics at all, even though the ID-level wiring state might still be knowable.

Please either make this diagnostic use a genuinely read-only/profile-ID-only auth-store path, for example an `AuthProfilesStore` helper that lists persisted profile IDs without decrypting or migrating token material, or explicitly change the implementation and PR body so the doctor behavior is allowed to invoke the existing auth-store load/migration side effects. I would prefer the ID-only path because it matches the diagnostic's data need and keeps the new doctor check within the stated read-only scope.

### 🟢 What looks good — The pure comparator covers the right wiring cases

The `codex_auth_wiring_items` helper itself is the right shape once the profile-presence input is safe to obtain. It reads provider slot state from the live `Config`, stores nothing, stays silent for the common "Codex unused" case, warns on the two mismatched quadrants, and returns OK when both a Codex profile and an opted-in slot exist. The four unit tests cover those quadrants cleanly and the Fluent strings are wired through the CLI string helpers.

singlerider:
Moved from the released v0.8.2 to v0.8.3. Read-only doctor diagnostic that relates to the OpenAI Codex provider work in v0.8.3 (#7894); rides with the next point release as carried polish.

singlerider (approved):
@JordanTheJet — reviewed on its own merits.

🟢 Good shape for a doctor check. Splitting the pure `codex_auth_wiring_items` (no I/O, stores nothing) from the async `check_codex_auth_wiring` (profile-store read) keeps the decision logic unit-testable, and the 2x2 match on `(codex_profile_present, auth_slots.is_empty())` covers all four states cleanly — warn-profile-no-slot, warn-slot-no-profile (which names the offending `openai.<alias>`), ok, and silent-when-unused so users who never touch Codex get no noise. No string-match dispatch; `requires_openai_auth` is read from the canonical config slot rather than hardcoded.

🟢 The fail-quiet on an unreadable profile store is the right call — emitting a "no credentials" warning you can't substantiate would be worse than silence — and surfacing these two states at `doctor` time instead of as a confusing auth error on the first model call is exactly what this command is for. All four branches are unit-tested and the messages are i18n-keyed.

LGTM.

singlerider:
@JordanTheJet this PR conflicts with `master`. Please merge the latest `master` in to resolve before it can be squash-merged.

singlerider:
@JordanTheJet follow-up: master now has the advisory-gate fix (#8520 merged), so once you merge the latest `master` into this branch the conflict clears and CI should go green. No rebase, merge only.

JordanTheJet:
@singlerider merged latest master into this branch (merge commit, no rebase) — the cli.ftl conflict with the new systemd-linger doctor strings is resolved keeping both. Ran cargo check -p zeroclaw-runtime and the doctor unit tests locally (36 passed, including the codex-wiring and linger tests). CI should go green now.

Audacity88 (changes requested):
Reviewed current head b2028f3 after the author merged `master`. I checked the live PR state and checks, the prior reviews/comments from WareWolf-MoonWall and singlerider, my earlier request-changes review, the current diff, and the current `AuthService`, `AuthProfilesStore`, and `SecretStore` paths. I did not run local Cargo validation for this review-only pass; public CI is green, and the author reported focused `zeroclaw-runtime` checks after the merge.

### ✅ Resolved — branch freshness and unrelated-file cleanup are no longer the problem

The branch update did what singlerider asked for: the PR is no longer conflicted, and the visible CI checks are green. I also agree with WareWolf-MoonWall's earlier re-review that the original unrelated workflow/Clawpatch cleanup is no longer in the net diff. The current PR diff is scoped to the doctor implementation and English Fluent strings.

### 🔴 Blocking — the current head still loads the write-capable auth profile path from `doctor`

My earlier blocker still applies on this head. `check_codex_auth_wiring` still derives the Codex profile presence bit by constructing `AuthService::from_config(config)` and calling `load_profiles().await`. That delegates to `AuthProfilesStore::load()`, which goes through `load_locked()`, decrypts every persisted token field with `decrypt_optional()`, and calls `SecretStore::decrypt_and_migrate()` for those values. When a legacy `enc:` value is encountered, `load_locked()` updates the persisted profile payload and writes it back through `write_persisted_locked()`.

That is still broader than the PR's claimed read-only diagnostic. The body says the check "does not create/modify auth profiles" and the security section says it "inspects only profile ids", but the implementation reaches those IDs by loading and decrypting the full credential store, with migration/writeback behavior on the same path.

The practical issue is the same as before: running `zeroclaw doctor` can become the first command that rewrites auth profile storage, and an unrelated corrupt or undecryptable stored secret can make this new diagnostic skip entirely before it answers the ID-level wiring question. Please either move this check to a genuinely ID-only/read-only auth-store path, or change the implementation and PR body/security text so this doctor command explicitly owns the existing auth-store load/decrypt/migration side effects. I still think the ID-only path is the right fix because the diagnostic only needs to know whether an `openai-codex:*` profile ID exists.

### 🟢 What looks good — the pure wiring comparator remains a good shape

The pure `codex_auth_wiring_items` helper still looks sound once the profile-presence input is safe to obtain. It reads live provider slot state from `Config`, stays quiet when Codex is unused on both sides, warns on the two mismatched quadrants, and reports OK when both sides are present. The added unit tests cover those comparator cases cleanly.

singlerider:
@Audacity88 pushed a maintainer commit (3712f08) resolving your 2026-07-03 blocker:

🔴 **Doctor loaded the write-capable auth path**: `check_codex_auth_wiring` no longer calls `AuthService::load_profiles()` (which goes through `load_locked` -> `decrypt_optional` / `decrypt_and_migrate` and can rewrite the store via `write_persisted_locked` on legacy `enc:` values). Added `AuthProfilesStore::list_profile_ids` (plus an `AuthService` passthrough): it acquires the shared lock, reads and parses the persisted store, and returns the profile-map keys with no token decryption and no migration writeback. The doctor check now uses that ID-only path, which is all it needs to answer whether an `openai-codex:*` profile exists. This also removes the failure mode where an unrelated undecryptable secret made the diagnostic skip entirely.

Added a regression (`list_profile_ids_lists_without_decrypting_or_rewriting`) that seeds an encrypted store, calls `list_profile_ids`, and asserts the returned IDs are correct while the on-disk bytes are byte-for-byte unchanged.

Synced the PR body's Security section to name the read-only `list_profile_ids` path and state it cannot trigger the `enc:` to `enc2:` migration writeback.

Local: `cargo fmt`, `clippy -D warnings`, and targeted `zeroclaw-providers` / `zeroclaw-runtime` tests all green (doctor suite 36 pass, auth::profiles 5 pass incl. the new regression). Please re-review, and if satisfied, approve and merge.

Audacity88 (approved):
Context: @JordanTheJet, I re-reviewed current head `3712f08` after @singlerider's maintainer fix. I checked the live PR state and checks, the prior reviews/comments from WareWolf-MoonWall and singlerider, my earlier request-changes review, the current diff, `AuthService`, `AuthProfilesStore`, the `SecretStore` decrypt/migration path, and the doctor/auth regressions. I did not run local Cargo validation in this review-only pass; the current visible CI is green, and the latest maintainer note reports focused local checks.

### ✅ Resolved — doctor no longer uses the write-capable auth profile load path

My previous blocker is resolved. `check_codex_auth_wiring` now gets the Codex-profile presence bit through `AuthService::list_profile_ids()` instead of `load_profiles()`. The new `AuthProfilesStore::list_profile_ids()` method acquires the auth-store lock, reads/parses the persisted profile store, and returns the profile-map keys. It does not call `load_locked()`, `decrypt_optional()`, `SecretStore::decrypt_and_migrate()`, or `write_persisted_locked()`.

That keeps this doctor check inside the read-only scope the PR body claims: it no longer decrypts token material, creates migration/writeback pressure for legacy `enc:` values, or depends on unrelated token decryptability just to answer whether an `openai-codex:*` profile ID exists.

### ✅ Resolved — the no-rewrite ID-listing behavior is now pinned

The new `list_profile_ids_lists_without_decrypting_or_rewriting` regression covers the side effect that blocked approval. It seeds stored auth profiles, captures the on-disk bytes, calls `list_profile_ids()`, verifies the expected profile IDs are returned, and then asserts the on-disk bytes are unchanged.

The existing doctor-side comparator tests still cover the four wiring states: profile-without-slot warning, slot-without-profile warning, both-present OK, and Codex-unused silence.

### 🟢 What looks good — the diagnostic remains scoped to the operator mismatch

The final shape is the right slice for `doctor`: the check reads provider slots from the live `Config`, stores no duplicate state, adds no network call or new CLI flag, uses Fluent strings for the new output, stays silent for users who do not use Codex auth, and surfaces the two real mismatch states before they become confusing first-model-call auth failures.

Approving to clear my standing change request.

PR #8032, opened 2026-06-19
fix(web): gate MCP server command/url required-ness on transport

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** In the Operator Console config editor (`/config/mcp.servers/<name>`), the `command` field was always marked **Required** and `url` never was — because the form derives "required" purely from the Rust type (`command: String` vs `url: Option<String>`). That's wrong for web-based MCP servers: `transport = http|sse` needs a `url`, not a `command`. This makes the required badge and the inline empty-field hint **transport-aware**, mirroring the server's existing `validate_mcp_config` contract (`command` ⇔ stdio, `url` ⇔ http/sse). The badge/hint update live as the operator changes the transport `<select>`. Revised after review: required-ness is now read from the backend-emitted `x-required-by-transport` schema metadata (added in #8349, sourced from `McpTransport::required_leaf`) via a new `mcpRequiredByTransport` helper, replacing the hand-kept `MCP_REQUIRED_LEAF` map that singlerider flagged as registry-shadowing. An older backend without the metadata leaves required-ness unclassified rather than re-encoding the enum.
- **Scope boundary:** The web change is display-only / client-side validation in `web/src/components/sections/FieldForm.tsx` (plus `web/src/lib/api.ts` for the schema read) and four `en` i18n strings; the server stays authoritative on save. This branch is stacked on #8349, which adds the `x-required-by-transport` metadata and an enum-match validator fix; that backend change rides along until #8349 merges, after which this rebases to web-only.
- **Blast radius:** `FieldForm.tsx` is the shared config/quickstart form renderer. The new logic is gated behind a `mcp.servers.<name>.<leaf>` path match, so every non-MCP row takes an early return and keeps its exact prior behavior. No save/PATCH serialization changes.
- **Linked issue(s):** Related #7267 (per-field `[[mcp.servers]]` editing), Related #7229 (MCP/Skills/Plugins/Providers dashboard tabs). Depends on #8349 (the backend `x-required-by-transport` metadata this PR reads). No existing issue fully tracks this; found via the web console.
- **Labels:** bug, config, risk:medium, web, size:XS.

Behavior after the change:

| transport | `command` | `url` |
|---|---|---|
| stdio | **Required** (red hint if empty) | "HTTP/SSE only" |
| http / sse | "Stdio only" | **Required** (red hint if empty) |

The `url` red-required hint is a small bonus: `url` is `Option<String>`, so the generic type rule never enforced it client-side — now it does for http/sse.

## Validation Evidence

This is a **web-only TypeScript change — no Rust source touched**, so the `cargo fmt / clippy / test` battery is unaffected. The relevant gate is the web typecheck:

```
$ cargo web check        # gen-api + npm install + tsc -b
==> gen-api → web/src/lib/api-generated.ts   🚀 [402.7ms]
==> npx tsc -b
$ npx tsc -b             # isolated re-run
→ exit 0, 0 errors (no FieldForm / i18n diagnostics)
```

- **Beyond CI — what I verified manually:** traced both changed function signatures (`setupRequirement`, `validationHint`) to their single caller (`FieldRow`); confirmed the new `t()` keys live in the `en` dictionary (`t()` is untyped `string` with English fallback, no locale key-parity gate); confirmed transport wire values are lowercase (`stdio`/`http`/`sse`) via `#[serde(rename_all = "lowercase")]` on `McpTransport` + schemars-driven variants (comparison lowercases defensively regardless).
- **What I did NOT verify:** a real browser render of the badge/hint flip and a live save of an http MCP server with an empty `command` (display-only change; worth an eyeball pre-merge).

## Security & Privacy Impact

- Secrets / tokens / credentials handling changed? **No** (the `headers` secret field and its masking are untouched)

## Compatibility

- Backward compatible? **Yes** — display-only; server contract unchanged.
- Upgrade steps: none.

## Rollback

Low-risk: `git revert <sha>`.


web/src/components/sections/FieldForm.tsx
```diff
@@ -444,6 +468,45 @@ function isRequiredField(typeHint: string): boolean {
...
+// headers, ...), leaving the generic `Option<...>` rule in charge.
+function mcpFieldRequired(
+  path: string,
+  transport: string | null | undefined,
+): boolean | null {
+  const leaf = path.match(/^mcp\.servers\.[^.]+\.([^.]+)$/)?.[1];
+  if (leaf !== "command" && leaf !== "url") return null;
+  // Empty / unset transport defaults to stdio (the schema default).
+  const isStdio =
+    ((transport ?? "").trim().toLowerCase() || "stdio") === "stdio";
```
singlerider: This shadows the `McpTransport` registry. The variants are already on this entry via `enum_values` (the same metadata that drives the transport `<select>`, see `web/src/lib/api.ts:407`), and the backend owns the command-vs-url contract in `validate_mcp_config` as an exhaustive `match` (`Stdio` vs `Http | Sse`).

The `=== "stdio"` literal plus the boolean collapse means a future fourth transport variant would auto-populate the select from the registry yet be silently classified url-required here with no diff, while the backend would force a new match arm. Drive the comparison off the registry-provided transport values instead of the literal, or have the backend emit the required-leaf-per-transport so the surface just reads it.

singlerider (changes requested):
I pulled this and traced it against the backend `McpTransport` enum and `validate_mcp_config`. The behavior is correct today and the form plumbing is careful, but it shadows the transport registry with a hardcoded variant literal, which is the one thing the contributing rules call out by name. Blocking on that.

### 🔴 Blocking - `mcpFieldRequired` hardcodes the `stdio` enum variant and collapses a 3-variant registry enum into a boolean

In `FieldForm.tsx`, `mcpFieldRequired` decides required-ness with:

```ts
const isStdio =
  ((transport ?? "").trim().toLowerCase() || "stdio") === "stdio";
return leaf === "command" ? isStdio : !isStdio;
```

`McpTransport` is a backend registry enum (`crates/zeroclaw-config/src/schema.rs:4375`, derives `ConfigEnum` + `JsonSchema`, `#[serde(rename_all = "lowercase")]`, three variants `Stdio` / `Http` / `Sse`, `#[default] Stdio`). Its variants are already surfaced to this exact form through `enum_values` on the entry, which the comment at `web/src/lib/api.ts:407` documents as the thing that "drives `<select>` options." So the transport `<select>` the operator flips is populated from the registry, but the required-ness decision next to it is typed as a literal `"stdio"` string and an implicit "everything-not-stdio means url." That is the registry-shadowing the contribution rules prohibit: a surface must read what the backend hands it, not re-encode the enum.

Two concrete failure modes from the boolean collapse:

1. A future fourth transport (say `websocket`) added to the enum would compile, populate the `<select>` from `enum_values` automatically, and this surface would silently classify it as `url`-required with zero diff here. The backend `validate_mcp_config` would force an explicit `match` arm for it (`Stdio` vs `Http | Sse` is an exhaustive match, line 8588). The surface should fail the same way the backend would, not guess.
2. The grep on master shows no existing `=== "<enum-variant>"` literal comparisons in this file; the bool `"true"`/`"false"` reads are not registry enums and `memory.backend` is a path match, not a value match. So this introduces a new shadowing pattern rather than following one.

Fix direction: drive this off the registry-provided transport metadata rather than a string literal. The required-leaf-per-transport relationship is itself backend knowledge (`validate_mcp_config` owns it); if the entry metadata does not already carry "which leaf this transport requires," extend the backend to emit it and have the form read it, rather than mirroring the match by hand in TS. At minimum, derive the comparison set from the transport entry's `enum_values` instead of the `"stdio"` literal so a new variant cannot be silently misclassified.

### 🟢 Praise - non-MCP rows are provably untouched

`mcpFieldRequired` returns `null` for any path that is not `mcp.servers.<name>.command|url`, and both `setupRequirement` and `validationHint` early-return on that `null` straight back into the generic `Option<...>` rule. Every non-MCP row keeps its exact prior behavior. The blast-radius claim holds.

### 🟢 Praise - reactive transport resolution is done right

`resolveMcpTransport` reads the live `draft` value for `mcp.servers.<name>.transport` first and only falls back to the saved/default entry value, so the badge and empty-hint flip the moment the operator changes the `<select>` rather than after a save. Threaded cleanly through `FieldRow` as a typed optional prop.

### 🟢 Praise - server stays authoritative and i18n is scoped correctly

No change to `McpServerConfig`, the schema, `/api/config`, or `validate_mcp_config`; this is display-only and the server remains the source of truth on save, exactly as described. The four new i18n keys are `en`-only and alphabetically placed, which is the right call: non-English strings are a maintainer release-time pass, not a contributor concern.

`npx tsc -b` is clean here. Resolve the registry-shadowing and this is a good fix.

singlerider (changes requested):
Re-reviewed at head `9e4cbf2deb`. Pulled the branch, read the new `mcpFieldRequired` against `validate_mcp_config` and the `McpTransport` enum, and confirmed `npx tsc -b` is clean.

### ✅ RESOLVED — unknown transports are no longer silently misclassified

The second failure mode from my last review is fixed correctly. `mcpFieldRequired` now returns `null` for any transport with no rule, so a future enum variant the backend adds is left unclassified rather than being assumed url-required the way the old `!isStdio` boolean did. The boolean collapse is gone. Good.

### 🔴 Blocking — the transport→required-leaf map still shadows the backend registry

The core blocker stands. `MCP_REQUIRED_LEAF` is a hand-maintained map of the supported transports living in the surface:

```ts
const MCP_REQUIRED_LEAF: Record<string, "command" | "url"> = {
  stdio: "command",
  http: "url",
  sse: "url",
};
```

This is exactly the relationship `validate_mcp_config` owns on the server (`crates/zeroclaw-config/src/schema.rs:8587`: `McpTransport::Stdio => command`, `McpTransport::Http | McpTransport::Sse => url`). The surface is re-encoding that match by hand. Per AGENTS.md this is the named anti-pattern: a surface must not hardcode "the things the system supports" (enum-shaped maps included); it reads what the backend hands it, and if the registry does not carry the field, the registry gets extended rather than shadowed.

The comment in the diff concedes this exactly — "The fully registry-driven fix is backend-emitted required-leaf-per-transport metadata on the entry; tracked as a follow-up" — but that fix is the requested change, not a follow-up. Deferring it means shipping the shadow, which is the thing the rule prohibits. The unknown-variant guard reduces the blast radius but does not remove the shadow: when the backend adds a variant, this map silently goes stale and the operator gets no required-ness signal on a transport the server will reject, so the form and the server disagree until someone hand-edits this literal.

Fix direction unchanged: have the backend emit the required-leaf-per-transport on the entry metadata (it already owns the relationship in `validate_mcp_config`) and read it here, so a new transport variant is classified by the server, not by this TS map. That makes the surface impossible to drift from the registry.

### 🟢 Reactive resolution, blast radius, and i18n still correct

`resolveMcpTransport` reads the live draft value first and falls back to the saved entry, so the badge/hint flip on `<select>` change. Non-MCP rows still return `null` and fall straight through to the generic `Option<...>` rule. The four i18n keys are en-only and alphabetically placed. `tsc -b` clean.

Remove the shadow by sourcing the required leaf from backend-emitted metadata and this merges.

singlerider:
Head `9e4cbf2d` unchanged since my last review. The 🔴 stands, with one correction to the path forward.

### 🔴 Blocking: registry shadow has no in-flight backend fix to defer to

I checked #7855 (`fix/7701-model-route-configurable`). It surfaces `model_routes`/`embedding_routes` via `Configurable` and will populate the `alias_source` stub at `api_config.rs:369`, but it does NOT carry MCP transport required-leaf metadata. It touches `traits.rs`, `sections.rs`, `api_sections.rs`, `schema.rs` for routes only. So the "tracked as a follow-up once #7855 lands" framing in the diff comment points at the wrong dependency: #7855 will not remove `MCP_REQUIRED_LEAF`. There is currently no backend work that emits this metadata.

That means the shadow cannot ride someone else's PR. The required-leaf-per-transport relationship `validate_mcp_config` owns (`schema.rs`: `Stdio => command`, `Http|Sse => url`) has to be emitted on the field entry and read here, either in this PR or in a dedicated backend metadata PR this one then consumes.

I am not going to build a competing wire field into this branch unilaterally while `traits.rs`/`api_sections.rs` are under active changes on #7855, because that invites exactly the cross-PR drift AGENTS.md is guarding against. Pick one:

1. Land a small backend change here that adds a `required_leaf` (or equivalent) to the MCP field entry, sourced from the `McpTransport` relationship, and have `FieldForm` read `entry.required_leaf` instead of the TS map. Coordinate the `traits.rs` touch with #7855 so they merge cleanly.
2. Split the backend metadata into its own PR, mark this one `Depends on #<that>`, and drop `MCP_REQUIRED_LEAF` once it lands.

### 🟢 Still good

The unknown-variant guard (`null` for any transport with no rule) keeps the interim safe: a new backend variant gets no false required signal and the server stays authoritative on save. `resolveMcpTransport` reads the live draft first so the badge flips reactively. i18n keys en-only and placed. `tsc -b` clean. The interim is honest; it just is not the registry-driven end state, and there is no other PR that will make it so.

Keeping CHANGES_REQUESTED. Source the required leaf from backend metadata and this merges.

Nillth:
@singlerider resolved in 171221df8e. Dropped the `MCP_REQUIRED_LEAF` map; `mcpFieldRequired` now reads the backend `x-required-by-transport` metadata (from #8349, sourced from `McpTransport::required_leaf`) via a new `mcpRequiredByTransport` helper that walks the cached config schema (`OPTIONS /api/config`) the same way `objectArrayElementProps` does.

A transport variant the backend adds now classifies from the registry with no change here; when a backend predates the metadata the map is null and required-ness is left unclassified rather than re-encoding the enum. The `registry-shadowing` guard is quiet on the branch now.

Stacked on #8349 (Depends on #8349); it rebases to web-only once that lands. `cargo web check` / `tsc -b` clean.

Audacity88 (commented):
@Nillth I re-reviewed the current head (`171221d`), @singlerider's standing blocker, the #8349 dependency state, the PR body/comments, live labels/checks, and the changed config/web source. I did not run local Cargo, `cargo web check`, `tsc`, or a browser smoke for this review.

### ✅ Resolved — The web no longer shadows the MCP transport registry

I agree with @singlerider's blocker that the form should not keep its own `stdio`/`http`/`sse` to `command`/`url` map. The current head removes `MCP_REQUIRED_LEAF`. `FieldForm` now receives `requiredByTransport` from `mcpRequiredByTransport(schema)`, and that helper reads `x-required-by-transport` from the backend-served `McpServerConfig` schema.

On the stacked #8349 backend commit, that metadata is projected from `McpTransport::required_leaf()`, while `validate_mcp_config` still matches on the `McpTransport` enum after the follow-up fix. That is the right single-source shape.

### 🔵 Suggestion — Refresh the dependency and checks before approval

I am leaving this as a comment rather than an approval because the current PR is still stacked on #8349, #8349 still has an active `CHANGES_REQUESTED` decision even though the current source includes the enum-match follow-up, and #8032 currently reports `DIRTY` / `CONFLICTING` with only the path-label check visible on the latest head. Once #8349 is landed or otherwise accepted as the base, please rebase/refresh this branch so the diff returns to the web-only slice and the normal Quality Gate/checks run on the refreshed head.

One small wording nuance for the old-backend path: when the metadata is absent, `mcpFieldRequired` returns `null`, but the generic type-derived rule can still treat `command` as required because it is a `String`. I read that as a conservative degraded path, not as a reason this web slice can land before #8349.

### 🟡 Warning — The browser behavior still needs a real UI smoke

The TypeScript path looks sound: `resolveMcpTransport` reads the live draft transport before falling back to the saved/default value, and both badge text and empty-field validation use the backend metadata. Confirming the PR body's own not-verified note: before this lands, please add a browser or near-live console check that flips one MCP server between `stdio`, `http`, and `sse` and confirms the `command`/`url` badge plus empty-field hint follow the selected transport. This is the central user-visible behavior.

No approval from me yet. The registry-shadowing blocker looks resolved in source, but the dependency, conflicts/checks, and UI evidence still need the final pass.

singlerider:
@Nillth this PR currently conflicts with `master`. Merge `master` into the branch (no rebase, per repo policy) to clear it.

singlerider:
@Nillth follow-up: master now has the advisory-gate fix (#8520 merged), so once you merge the latest `master` into this branch the conflict clears and CI should go green. No rebase, merge only.

singlerider (approved):
Pulled the branch at current head. It now merges the `x-required-by-transport` metadata work (#8349, which I just approved), and the web form consumes it exactly as this thread's second option suggested. `tsc --noEmit` passes, CI green, MERGEABLE.

### 🟢 Blocking resolved — the surface reads the backend registry instead of shadowing it

The blocker was that `FieldForm.tsx` hardcoded the transport-to-required-field contract with a `=== "stdio"` literal, shadowing the `McpTransport` registry: a future fourth variant would auto-populate the select from `enum_values` yet be silently classified url-required with no diff, while the backend `validate_mcp_config` match would force a new arm. The current head drives the decision off backend metadata:

- `api.ts` reads the backend-stamped `x-required-by-transport` off the `McpServerConfig` schema (derived from `McpTransport::required_leaf`, the same relationship `validate_mcp_config` enforces).
- `mcpFieldRequired` resolves required-ness from `requiredByTransport[transport]`, not a literal. The `leaf === "command" | "url"` guard only picks out the two transport-gated slots; the actual required verdict is the backend's.
- An unknown transport returns `null` (unclassified) rather than guessing, so a new variant surfaces from the registry and cannot be silently misclassified; the backend then forces its `match` arm and stamps its `required_leaf`, which the form reads.
- A backend predating the metadata returns `null`, leaving the generic `Option<...>` rule in charge, so nothing re-encodes the enum.

One source of truth end to end: the enum owns the command-vs-url contract, the schema projects it, the form reads it. Resolving this thread. Approving.

PR #8036, opened 2026-06-20
test(runtime): pin system prompt in cache-hit test to kill date flake

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** `turn_cache_hit_emits_agent_end_with_none_tokens` builds two agents and relies on their system prompts being byte-identical so the response-cache key matches on the second turn. The default prompt renders `Local::now()` (current date + UTC offset) into the system prompt, so under nextest full-workspace parallel load the two sequentially-built agents could render across a clock tick and diverge, producing a cache miss (`left: "uncached answer", right: "cached answer"`). Pinning both agents to an explicit identical `SystemPromptBuilder::default()` removes the time-dependent input so the test asserts cache behavior, not prompt determinism.
- **Scope boundary:** Test-only. No production code, no prompt-builder behavior, no cache logic changed. Two added builder lines in one test.
- **Blast radius:** None beyond the single test. No production surface touched.
- **Linked issue(s):** None.
- **Labels:** `agent`, `runtime`, `agent: tests`, `risk: low`, `size: XS`.

## Transient Failure Evidence

Observed on PR #7763 (`integration/a2a`), run `27852747572` attempt 1, `Test` job `82434667745`, 2026-06-19T23:26:19Z. The same job re-ran clean (attempt 2) with zero code change, confirming the flake:

```
FAIL [0.036s] (4183/9264) zeroclaw-runtime agent::agent::tests::turn_cache_hit_emits_agent_end_with_none_tokens
thread '...turn_cache_hit_emits_agent_end_with_none_tokens' (13999) panicked at crates/zeroclaw-runtime/src/agent/agent.rs:6503:9:
assertion `left == right` failed
  left: "uncached answer"
 right: "cached answer"
Summary [28.601s] 4186/9264 tests run: 4185 passed, 1 failed, 11 skipped
error: test run failed
```

The panic at `agent.rs:6503` is the cache-key comparison this PR hardens: the second turn missed cache because the two agents' system prompts diverged on the rendered timestamp under parallel load.

```bash
cargo fmt --all -- --check
cargo test -p zeroclaw-runtime agent::
```

  - `cargo fmt --all -- --check` exit 0, no output.
  - `cargo clippy --all-targets -- -D warnings` exit 0, `Finished dev profile`.
  - `cargo test -p zeroclaw-runtime agent::` exit 0, `569 passed; 0 failed; 1 ignored`. Target test green under repeated and 16-thread runs.
- **Beyond CI — what did you manually verify?** Reproduced the original flake characterization (cache miss under parallel load), confirmed the pinned-prompt version passes deterministically across isolation, full `agent::` module, and 16-thread runs. Did not change the prompt builder or cache code.
- **If any command was intentionally skipped, why:** Full `cargo test` workspace run narrowed to the affected crate/module; the change is one test in `zeroclaw-runtime` with zero cross-crate impact.

## Compatibility (required)

- Backward compatible? `Yes`
- If `No` or `Yes` to either: exact upgrade steps for existing users: None. Test-only.


WareWolf-MoonWall (approved):
# Review — PR #8036 `test(runtime): pin system prompt in cache-hit test to kill date flake`

**Reviewer:** WareWolf-MoonWall · **Head:** `1461d4a0360fc0b0d622e90dc5f33651cec158e3` · **Verdict: approve**

Minimal and correct flake fix. The root cause is clearly identified (two agents built sequentially during nextest parallel load straddle a clock tick, `DateTimeSection` renders `Local::now()` into each system prompt via `SystemPromptBuilder::with_defaults()`, the two prompts diverge, the response-cache key misses, agent\_b returns its own "uncached answer" instead of the cached one). Pinning both agents to `SystemPromptBuilder::default()` — which is `#[derive(Default)]` on an empty `sections: vec![]`, carrying no `DateTimeSection` — removes the time-dependent input entirely. I confirmed this directly: `prompt.rs` shows `with_defaults()` pushes `DateTimeSection` while `default()` (the derive) produces an empty sections vec. Two added lines, no production code touched, CI clean. No blockers.

---

### 🟢 Praise — Flake evidence is exemplary

Providing the specific CI run ID (`27852747572`), job ID (`82434667745`), attempt number, timestamp, and the exact panic output — plus the re-run clean on attempt 2 with zero code change — is exactly the right documentation for a flake fix. This makes it permanently auditable.

---

### 🟢 Praise — Test assertions remain fully intact

The pin removes the non-deterministic input, but the test still asserts everything it should: agent\_b returns `"cached answer"` (cache hit), `AgentEnd` carries `None` tokens (no LLM usage billed), and no `LlmRequest` event is emitted (LLM never called). Cache-hit behavior is fully exercised; the pin only eliminates the timing-dependent key divergence.

---

### 🟢 Praise — `default()` vs `with_defaults()` distinction is the right tool

Using `SystemPromptBuilder::default()` (the derived empty builder) instead of `with_defaults()` (which injects `DateTimeSection`) is exactly the right scalpel: it removes the time dependency without forking the prompt builder API or adding a test-only constructor.

---

### 🔵 Suggestion — Add an inline comment to guard against future removal

With only two added builder calls, the reason for the explicit prompt builder is not self-evident. A reader who later sees "unnecessary" explicit prompt builders might remove them thinking it's dead code. A brief comment would prevent that:

```rust
// Pin to a static prompt (no DateTimeSection / Local::now()) so both
// agents produce identical cache keys regardless of clock ticks under
// parallel test load.
.prompt_builder(SystemPromptBuilder::default())
```

Not a blocker given the PR description is thorough, but inline context ages better than a PR description.

Audacity88 (approved):
I reviewed current head 1461d4a0360fc0b0d622e90dc5f33651cec158e3 against the live PR state, PR body, current patch, prior review, green GitHub checks, and the adjacent prompt-builder source. I did not run local Cargo; this is a two-line test-only change, and the remote Test/Lint/Format/CI Required Gate checks are green.

### 🟢 What looks good — narrow flake fix removes time-dependent cache input

This is the right narrow flake fix. The cache-hit assertion depends on both agents producing the same request key. When the test uses the builder default path, `AgentBuilder::build()` falls back to `SystemPromptBuilder::with_defaults()`, and that default set includes `DateTimeSection`, which renders `Local::now()` into the prompt. Two sequential agents can therefore produce different prompt text if the test crosses a date or UTC-offset boundary under load.

Pinning both test agents to `SystemPromptBuilder::default()` removes that time-dependent prompt section without changing production prompt construction or cache logic. The test still checks the important behavior: the second agent returns the cached answer, emits `AgentEnd` with no token usage, and does not emit an `LlmRequest`.

### 🔵 Suggestion — add inline comment to guard against removal

WareWolf-MoonWall's inline-comment suggestion is worth taking the next time this test is touched. A short note that the explicit builder avoids `DateTimeSection` would make these two lines harder to "clean up" by accident. This is not required for this PR.

PR #8038, opened 2026-06-20
feat(install): add zeroclaw to PATH automatically, with --no-modify-path opt-out

JordanTheJet:
## Summary

- **Base branch:** `master`
- **What changed and why:** The installer previously only *printed* a suggested PATH export line, leaving every user to hand-edit their shell profile before `zeroclaw` was runnable. This now appends the bin dir to the detected shell profile (`.zshrc` / `.bashrc` / `config.fish`) automatically, wrapped in a `# >>> zeroclaw >>>` marker block — matching the UX of rustup/nvm. The append is idempotent (skipped when the bin dir is already on PATH), `--no-modify-path` restores the old print-only behaviour, `--dry-run` reports what it *would* add without writing, and `do_uninstall` strips the marker block so removal is clean.
- **Scope boundary:** Only `install.sh` PATH handling and the matching uninstall strip. No change to the build, feature selection, prebuilt download, or quickstart flow. Does not touch how rustup itself is installed (`--no-modify-path` is still passed to rustup).
- **Blast radius:** Bootstrap/onboarding only. Affects what gets written to a user's shell rc file at install time. Custom `--prefix` installs are deliberately left print-only (we don't auto-edit a real profile for an isolated/test install).
- **Linked issue(s):** None.
- **Labels:** `enhancement`, `risk: low`, `size: XS`

Shell-only change — the Rust battery (`cargo fmt/clippy/test`) is N/A (no Rust touched). Per the template's bootstrap-script guidance:

```bash
$ bash -n install.sh && echo OK
OK
$ sh -n install.sh && echo OK
OK
```

- **Commands run and tail output:** `bash -n install.sh` and `sh -n install.sh` both pass (POSIX-clean). `shellcheck` not installed locally, so static lint was skipped there.
- **Beyond CI — what did you manually verify?** Ran the new logic end-to-end against a sandbox `$HOME`/profile:
  - `--dry-run --prebuilt` prints `Would add <bin> to PATH in <profile>` and writes **nothing** to the rc file.
  - Fresh install appends the `# >>> zeroclaw >>>` … `# <<< zeroclaw <<<` block while preserving pre-existing rc content.
  - Re-run: the `grep "$CARGO_HOME/bin"` guard hits, so no duplicate block is appended (idempotent).
  - `--uninstall` strips the marker block via `sed` (verified on BSD/macOS sed), restoring the original rc.
- **If any command was intentionally skipped, why:** Rust battery skipped — no Rust in the diff. `shellcheck` skipped — not installed on this machine; `bash -n`/`sh -n` cover syntax.

- New permissions, capabilities, or file system access scope? `Yes` — the installer now writes to the user's shell profile (`~/.zshrc` etc.) by default instead of only printing a suggestion. It is scoped to a single marked block, idempotent, opt-out via `--no-modify-path`, and announced on stdout. No write occurs in `--dry-run`.
- If any `Yes`, describe the risk and mitigation: Risk is an unexpected rc-file edit; mitigated by the marker block (clean re-install/uninstall), the `--no-modify-path` escape hatch, and print-only fallback for custom `--prefix` installs and when the write fails.

## Compatibility (required)

- Backward compatible? `Yes` — existing behaviour is reachable via `--no-modify-path`; profiles already containing the bin dir are left untouched.
- Config / env / CLI surface changed? `Yes` — adds one new flag, `--no-modify-path`, documented in `--help`. No existing flag changes.
- If `No` or `Yes` to either: No upgrade steps required. Users who prefer the old print-only behaviour pass `--no-modify-path`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -380,6 +380,8 @@ Options:
   --prefix PATH        Install everything under PATH (default: \$HOME)
                        Sets CARGO_HOME, RUSTUP_HOME, source checkout, config
   --dry-run            Show what would happen without building or installing
+  --no-modify-path     Don't add ZeroClaw to PATH in your shell profile; just
+                       print the line to add manually
   --skip-quickstart       Skip the post-install quickstart prompt
   --uninstall          Remove ZeroClaw binary and optionally config/data
   -h, --help           Show this help
@@ -442,6 +444,21 @@ do_uninstall() {
     else
       info "Config preserved at $config_dir (non-interactive — use rm -rf to remove)"
     fi
+  fi
+
+  # Strip the PATH marker block this installer may have added to the profile.
+  local profile
+  profile=$(detect_shell_profile)
+  if [ -f "$profile" ] && grep -q "# >>> zeroclaw >>>" "$profile" 2>/dev/null; then
+    local tmp_profile
+    tmp_profile=$(mktemp)
+    if sed '/# >>> zeroclaw >>>/,/# <<< zeroclaw <<</d' "$profile" >"$tmp_profile" 2>/dev/null &&
+      cat "$tmp_profile" >"$profile" 2>/dev/null; then
+      info "Removed PATH entry from $profile"
+    else
+      warn "Could not edit $profile — remove the zeroclaw PATH block manually"
+    fi
+    rm -f "$tmp_profile"
   fi
 
   # Check if another zeroclaw still lurks in PATH
@@ -692,6 +709,7 @@ SKIP_QUICKSTART=false
 LIST_FEATURES=false
 UNINSTALL=false
 DRY_RUN=false
+MODIFY_PATH=true # append PATH export to the shell profile; --no-modify-path opts out
 PREFIX="$HOME"
 INSTALL_MODE="" # ""=ask, "prebuilt"=force prebuilt, "source"=force source
 PRESET=""       # ""=unset, "minimal"=alias for --minimal, "full"=default-features
@@ -747,6 +765,7 @@ while [ $# -gt 0 ]; do
     PREFIX=$(echo "$1" | sed 's|/*$||')
     ;;
   --dry-run) DRY_RUN=true ;;
+  --no-modify-path) MODIFY_PATH=false ;;
   --skip-quickstart) SKIP_QUICKSTART=true ;;
   --prebuilt) INSTALL_MODE="prebuilt" ;;
   --source) INSTALL_MODE="source" ;;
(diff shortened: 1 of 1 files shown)


Nillth (approved):
I read the full diff against `install.sh` on master and exercised the new append / idempotency / uninstall-strip paths against a sandbox profile to confirm they behave as described. This is a clean, well-scoped change: the implementation matches the PR body, the dry-run is honest, and the opt-out and uninstall paths are correct. `bash -n` and `sh -n` both pass on the head commit. No blockers from me; the notes below are optional.

### 🟢 What looks good — idempotent marker block and graceful degradation

The `# >>> zeroclaw >>>` / `# <<< zeroclaw <<<` marker block matches the rustup/nvm convention and makes both re-install and uninstall clean. I verified in a sandbox that a second run hits the `grep "$CARGO_HOME/bin"` guard and appends nothing, and that the uninstall `sed` range-delete strips exactly the installer's block while leaving pre-existing rc content untouched.

Two details I want to name because they are easy to get wrong and you got them right:

- Rewriting the profile on uninstall with `cat "$tmp_profile" >"$profile"` instead of `mv` preserves the original file's inode, owner, and permissions. That is the correct call for a file in the user's `$HOME` that may be a symlink or carry specific perms.
- Write failures degrade to the old print-only guidance (`warn` + `print_path_help`) rather than aborting, and `--dry-run` writes nothing while still reporting the target. This also means a fish user whose `~/.config/fish/` directory does not exist yet falls back cleanly to printed guidance instead of erroring.

### 🔵 Suggestion — non-interactive installs now edit the profile silently

The auto-append gates on `MODIFY_PATH=true` and `PREFIX="$HOME"` but not on an interactive TTY, so a non-interactive `curl … | bash` (CI job, Dockerfile, provisioning script) will now write to `~/.bashrc` where before it only printed. rustup behaves the same way by default unless `--no-modify-path` is passed, so this may well be the parity you intend — I am flagging it only so it is a conscious choice rather than a side effect. If you would rather leave automation untouched, gating the append on `[ -t 0 ]` (the way `do_uninstall` already gates its config prompt) and falling back to print-only when non-interactive would be one way to do it. Not blocking.

### 🔵 Suggestion — make the `zerocode` (TUI) coverage explicit

Because the marker block adds the shared `$CARGO_HOME/bin` *directory* to PATH rather than an individual binary, this already makes the `zerocode` TUI runnable too, not just `zeroclaw` — which is exactly what the next-step hint a few lines down relies on (the `Run zerocode to launch the terminal UI` message only resolves if that bin dir is on PATH). That is a genuine win of this change. The one gap is framing: the marker comment (`# >>> zeroclaw >>>`) and the `Added $CARGO_HOME/bin to PATH` message are worded entirely around `zeroclaw`, so a reader could assume only the main binary got wired up. Worth either confirming in the validation that `zerocode` resolves on a fresh shell after the append, or nudging the `Added …` line to acknowledge that both `zeroclaw` and the `zerocode` TUI are now on PATH. The mechanics are already correct; this is about making the multi-binary coverage explicit. Optional.

### 🔵 Suggestion — uninstall leaves one orphan blank line

The append writes a leading `\n` that sits *outside* the marker block, and the uninstall `sed` range only deletes the marker lines themselves, so each install -> uninstall cycle leaves one extra blank line in the profile. Purely cosmetic — I confirmed it in the sandbox — and rustup has the same wart. If you want it spotless, putting the leading newline inside the block (or collapsing trailing blanks on strip) would handle it. Optional.

### One housekeeping note (not blocking)

No labels were applied to the PR yet — the auto-labeler likely had not run on this fork branch. Two small corrections to the set the body declares: this repo has no `type: feature` label (the feature label is `enhancement`), and at ~58 changed lines this is `size: XS` rather than `size: S` (S starts at 81 lines). I have applied `enhancement` / `size: XS` / `risk: low` to keep triage tidy.

Overall this is solid, and the safety posture (idempotent, opt-out, honest dry-run, clean uninstall, graceful write-failure fallback) is exactly right for something that edits a user's shell profile.

Audacity88 (approved):
I checked the public `install.sh` diff, PR body and metadata, live checks, existing public review/comment snapshots, and the public install snippet that documents the `curl | sh` path. I also checked the privacy surface: the diff does not add secrets or PII, and the visible author identity is contributor-chosen Git metadata. I do not see a blocker.

### 🟢 What looks good — bounded profile mutation with opt-out and rollback

The new default write is constrained to normal `$HOME` installs, uses a marked block in the detected shell profile, preserves print-only behavior for `--no-modify-path` and custom `--prefix` installs, and keeps `--dry-run` non-mutating. The uninstall path removes the marked block, so the behavior has a clean rollback story instead of leaving users to hunt for an appended line manually.

### 🔵 Suggestion — document the changed one-liner behavior

The public install snippet is `curl ... | sh`, so this PR changes that one-liner from “print the PATH line” to “edit the detected shell profile by default” for normal installs. That looks intentional and matches the PR goal, but the follow-up docs should mention both the default profile update and the `--no-modify-path` escape hatch so automation/headless users know how to preserve the old print-only behavior.

PR #8042, opened 2026-06-20
fix(web): surface config drift conflict on the enable/disable toggle

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** The entity ENABLED pill (`EntityEnabledToggle`) caught
  *every* PATCH failure as a generic "save failed" and reverted the pill without
  calling `onChange`. When the daemon's in-memory config has drifted from disk on
  the patched path, the gateway's drift guard returns **409 `config_changed_externally`**
  — so the toggle silently reverted with no hint of why or how to proceed. Detect
  that specific 409 and surface it distinctly: show **"Config changed on disk"**
  with an explicit **Overwrite** action that retries with
  `X-ZeroClaw-Override-Drift: true`. Generic failures keep the existing "save
  failed" path.
- **Scope boundary:** Display/UX only — `EntityEnabledToggle.tsx`, a small optional
  `overrideDrift` flag on `patchConfig()` (sends the documented override header),
  and two `en` i18n strings. No server, schema, or config change. The drift guard
  and override header are pre-existing server contract.
- **Blast radius:** `patchConfig()` gains a second OPTIONAL arg — every existing
  caller (passing only `ops`) is unaffected. The toggle's happy path and generic
  error path are unchanged; the new branch only adds a distinct affordance for the
  one 409 code.
- **Linked issue(s):** none filed. Found while diagnosing a channel enable/disable
  that appeared not to persist — root cause there was a **stale daemon** (in-memory
  ≠ on-disk after an external config write), which this surfaces instead of hiding.
- **Labels:** suggest `type: bug`, `scope: web`, `risk: low`, `size: S`.

## Validation Evidence

Web-only TypeScript change — no Rust source touched, so `cargo fmt/clippy/test`
are unaffected. Web typecheck:

```
$ cargo web check        # gen-api + npm install + tsc -b
==> gen-api → web/src/lib/api-generated.ts   🚀 [392.8ms]
==> npx tsc -b
→ exit 0
```

- **Verified by hand:** the 409 wire code is `config_changed_externally`
  (`api-generated.ts` ConfigApiCode union) and the PATCH endpoint accepts
  `X-ZeroClaw-Override-Drift: "true"` (same source); reproduced the underlying
  drift-guard 409 / 200-on-override behaviour earlier against an isolated daemon.
- **What I did NOT verify:** a live browser render of the drift banner + Overwrite
  click against a genuinely-drifted daemon (the happy path and typecheck are
  covered; the drift branch is reasoned from the established server contract).

## Security & Privacy Impact

- New permissions / capabilities / FS access? **No**
- New external network calls? **No** (same PATCH endpoint; adds one request header)
- PII / real identities / personal data? **No**

## Compatibility

- Backward compatible? **Yes** — optional arg; UX-only.
- Upgrade steps: none.

## Rollback

Low-risk: `git revert <sha>`.

# ----------------------------------------------------------------------
# Files: web/src/components/EntityEnabledToggle.tsx, web/src/lib/api.ts,
#        web/src/lib/i18n.ts  (3 files, +88/-39)
# Separate from MCP PR #8032 (different files, off newer master ddffad12ab).


web/src/components/EntityEnabledToggle.tsx
```diff
@@ -24,57 +24,92 @@ export default function EntityEnabledToggle({
...
+      await patchConfig(
+        [{ op: 'replace', path: `${prefix}.enabled`, value: next }],
+        overrideDrift ? { overrideDrift: true } : undefined,
+      );
       onChange(next);
     } catch (e) {
-      setError(e instanceof Error ? e.message : String(e));
+      if (
+        e instanceof ApiError &&
+        e.envelope.code === 'config_changed_externally'
```
singlerider: This compares against an unchecked string. `ApiError.envelope.code` is typed `code: string` (`web/src/lib/api.ts:41`), and `ConfigApiCode` / `config_changed_externally` is not in `web/src/lib/api-generated.ts` at all (the PR body's "already in the generated union" claim is wrong; the enum lives only in `crates/zeroclaw-config/src/api_error.rs:31`).

So a backend rename or a typo here compiles clean and silently disables drift detection, falling back to the exact silent-revert this PR exists to fix. This is the first web call site to branch logic on `envelope.code`. Surface `ConfigApiCode` through gen-api and narrow the field to the generated union so this comparison is checked at build time.

singlerider (changes requested):
I pulled this and traced it through `apiFetch`, `ApiError`, and the backend drift guard. The runtime path is correct and the UX is the right fix, but it dispatches program logic on an unchecked wire-code string, and the PR body's claim that this code is registry-typed is not accurate. Blocking on the type safety.

### 🔴 Blocking - drift detection branches on an unchecked string literal, not the `ConfigApiCode` registry

The new branch in `EntityEnabledToggle.tsx` is:

```ts
e instanceof ApiError &&
e.envelope.code === 'config_changed_externally'
```

`ApiError.envelope.code` is typed as plain `code: string` (`web/src/lib/api.ts:41`), so this comparison is an unchecked free-string match. The PR body states the code lives in "the `api-generated.ts` ConfigApiCode union" - it does not. `grep` for `ConfigApiCode` and `config_changed_externally` in `web/src/lib/api-generated.ts` returns nothing; the enum exists only on the Rust side (`crates/zeroclaw-config/src/api_error.rs:31`, serialized at line 70). Nothing surfaces it to TypeScript today.

The consequence: a backend rename of the variant, or a typo here, compiles clean and silently disables drift detection - the toggle would fall back to the generic "save failed" silent-revert, which is exactly the bug this PR exists to kill. This is the first place in the web tree that branches logic on a specific `envelope.code` (every other reference treats it as display-only `[${code}] ${message}`), so it sets the pattern - it should set the right one.

Fix direction: surface `ConfigApiCode` through gen-api so `envelope.code` is typed as the generated union and this comparison is checked at compile time. The backend already owns the enum; export it the same way other wire types reach `api-generated.ts`, then narrow `ApiError.envelope.code` (or a typed accessor) to it. That makes a removed/renamed variant a build failure here instead of a silent runtime regression. Correcting the inaccurate "already in the generated union" line in the PR body is part of this.

### 🟢 Praise - the override header contract is exactly right

`X-ZeroClaw-Override-Drift: "true"` matches the server contract precisely (`crates/zeroclaw-gateway/src/api_config.rs:1643-1648`, documented in `openapi.rs:206`). The 409 -> show-overwrite -> retry-with-override flow mirrors the documented drift-guard escape hatch rather than inventing one.

### 🟢 Praise - `patchConfig` blast radius is genuinely additive

The new second parameter is optional and only attaches the header when `opts.overrideDrift` is set; every existing caller passing just `ops` spreads `{}` and is byte-for-byte unchanged on the wire. The blast-radius claim holds.

### 🟢 Praise - runtime wiring verified and the UX is the correct call

Confirmed `apiFetch` throws `ApiError(status, parsed)` whenever the error body parses as `{code, message}` (`api.ts:131-153`), so on a real 409 `e.envelope.code` is populated and the branch fires - the affordance is not dead code. Turning a silent pill-revert into an explicit "Config changed on disk / Overwrite" affordance is the right fix for the stale-daemon confusion. The two i18n keys are `en`-only and alphabetically placed, which is correct.

Wire the discriminant to the registry type and this is a clean merge.

singlerider (approved):
Re-reviewed at head `b2fb3b9401` after the push. My one blocking concern from the prior round is fully resolved.

### ✅ RESOLVED — drift discriminant is now compile-checked against the registry

The branch no longer compares against a bare string. `api.ts` now derives the type from the generated OpenAPI schema and pins the constant to it:

```ts
export type ConfigApiCode = components["schemas"]["ConfigApiCode"];
export const ConfigApiCodes = {
  configChangedExternally: "config_changed_externally",
} as const satisfies Record<string, ConfigApiCode>;
```

and `EntityEnabledToggle.tsx` branches on `e.envelope.code === ConfigApiCodes.configChangedExternally`. I confirmed `ConfigApiCode` is emitted into `api-generated.ts` as the full 13-variant union (`path_not_found | validation_failed | config_changed_externally | ...`) from the backend `JsonSchema` derive on `crates/zeroclaw-config/src/api_error.rs`, and that the error envelope's `code` field is typed to it. The `satisfies` clause means a backend rename or a typo here is now a `tsc` failure instead of a silent drift-detection regression. That is exactly the fix I asked for, and the inaccurate PR-body claim is no longer relevant since the type is wired through for real.

### 🟢 Override-header contract and additive blast radius still hold

`patchConfig`'s second parameter stays optional and only attaches `X-ZeroClaw-Override-Drift: true` when set, so every existing caller is unchanged on the wire. The 409 → show-overwrite → retry-with-override flow matches the documented server escape hatch.

### 🟢 The two new comments earn their place

The `driftConflict` block explains *why* it is tracked separately (the silent pill-revert that made a stale daemon look like a broken toggle), and the `ConfigApiCodes` JSDoc explains *why* to branch on the constant rather than a literal. Both are non-obvious rationale, not mechanism narration. Good.

i18n keys are en-only and alphabetically placed. Approving.

Issue #6517, opened 2026-05-07
[Bug]: Context Overflow Causes Hallucination / Topic Drift


essamsalah:
### Affected component

runtime/daemon

### Severity

S2 - degraded behavior

### Current behavior

**Bug Report: Context Overflow Causes Hallucination / Topic Drift**

**Summary**
When a conversation runs long enough to fill the context window, the bot starts hallucinating — drifting off-topic or responding as if previous messages don't exist.

**Environment**
- ZeroClaw (latest)
- Provider: Kimi K2.5 via Moonshot API (`api.moonshot.ai`)
- Channel: Discord
- OS: macOS (Apple Silicon)

**Steps to Reproduce**
1. Start a conversation with the bot over Discord
2. Continue a multi-turn conversation for an extended period
3. Observe the bot begin responding to a different topic or losing context of the current discussion

**Workaround**
Sending `/new` in Discord resets the session. Alternatively, clearing `~/.zeroclaw/workspace/sessions/*` and restarting the daemon works.

**Additional Notes**
- `history_pruning = false` is required in our config because enabling it causes a *different* bug (messages being silently dropped mid-conversation)
- `context_compression` is also disabled because it triggers a temperature error when using Kimi K2.5
- This means there is currently no safe automatic mechanism to handle context overflow — the only option is manual `/new`
- A warning or graceful degradation when approaching context limits would be very helpful

### Expected behavior

I think that the Bot should maintain coherent conversation context, or gracefully indicate when context is full.

### Steps to reproduce

```bash
just after a while talking to the bot on the same subject
```

### Impact

Bot silently starts hallucinating — responding off-topic without any warning that context has been exceeded.

### Logs / stack traces

```text

```

### ZeroClaw version

0.7.4

### Operating system

macOS Taho

### Regression?

No, first-time setup

### Pre-flight checks

- [x] I reproduced this on the latest master branch or latest release.
- [x] I redacted secrets, tokens, and personal data from all submitted content.

singlerider:
@essamsalah Can you reproduce on v0.7.5?

Audacity88:
## Status Update

This is still waiting on a current-version reproduction check.

Please retest on current `master` or a current release by Tuesday, July 7, 2026. If the context-overflow/topic-drift behavior still reproduces, please add the provider/model, channel/config context with secrets redacted, the prompt or transcript shape, and the relevant logs. If there is no update by then, this can be closed as not reproducible for now.

[...]

Issue #5823, opened 2026-04-17
[Bug]: prune_history Phase 1 partially collapses tool groups that straddle keep_recent; Phase 3 orphan sweep then evicts protected tools


shaun0927:
## Summary

Two interacting defects in `crates/zeroclaw-runtime/src/agent/history_pruner.rs` (PR #5167) can silently evict a tool message that `keep_recent` is supposed to protect:

1. **Phase 1 partial-collapse bug.** When `keep_recent` falls inside a multi-tool group, the Phase 1 inner loop collapses the unprotected tool(s) while leaving the protected tool(s) in place, and rewrites the preceding assistant message to a summary string that no longer contains `"tool_calls"`.
2. **Phase 3 substring-heuristic bug.** `remove_orphaned_tool_messages` decides whether an assistant was a tool-call origin by looking for the literal substring `"tool_calls"` (lines 73 and 117). Because the Phase 1 summary rewrite strips that substring, Phase 3 classifies the still-living protected tool as orphaned and removes it.

The net effect is that `prune_history` silently drops a message within the recent window — violating the `keep_recent` contract the pruner is built around.

## Affected code

`crates/zeroclaw-runtime/src/agent/history_pruner.rs` — Phase 1 inner loop (`prune_history`, ~lines 179-205) and `remove_orphaned_tool_messages` Pass 2 (~lines 99-142).

## Reproduction

Verified locally against d0d83198 (v0.7.0). Minimal repro added to `history_pruner::tests`:

```rust
#[test]
fn phase1_partial_collapse_evicts_protected_tool() {
    let mut messages = vec![
        msg("system", "sys"),                                                                           // idx 0 (protected: system)
        msg("user", "query"),                                                                           // idx 1
        msg("assistant", r#"{"content":null,"tool_calls":[{"id":"t1","name":"s","arguments":"{}"},
                                                          {"id":"t2","name":"w","arguments":"{}"}]}"#), // idx 2 (unprotected)
        msg("tool", r#"{"tool_call_id":"t1","content":"first"}"#),                                      // idx 3 (unprotected)
        msg("tool", r#"{"tool_call_id":"t2","content":"PROTECTED second"}"#),                           // idx 4 (protected: keep_recent=3)
        msg("user", "follow up"),                                                                       // idx 5 (protected)
        msg("assistant", "final"),                                                                      // idx 6 (protected)
    ];
    let config = HistoryPrunerConfig {
        enabled: true,
[...]

PR #8048, opened 2026-06-20
fix(runtime): keep tool-result content under context pressure; honor history_pruning config

Nillth:
## Summary

- **What changed and why:**
  - The turn engine's preemptive over-budget trim (`agent/turn/history_window.rs`)
    hardcoded `enabled: true`, `keep_recent: 4`, `collapse_tool_results: true`,
    silently overriding the agent's `history_pruning` config. An agent that set
    `history_pruning.enabled = false` was still trimmed, with no signal.
  - Under pressure it collapsed old assistant+tool exchanges into content-free
    `[Tool exchange: N tool call(s) ...]` placeholders. On Anthropic those
    summaries are dropped from the request entirely (`should_skip_internal_pruning_marker`),
    so the model lost all record of what past tools returned and hallucinated
    their output (drift / "responding as if previous messages don't exist").
  - Now the preemptive trim honors the resolved `collapse_tool_results` and
    `keep_recent`. Trimming an over-budget request is mandatory (otherwise the
    provider returns `context_length_exceeded`), so `enabled` is still forced on,
    but when the agent disabled pruning the override is logged at WARN instead of
    happening silently.
  - Added an escalating in-place extract ladder (2000 -> 512 -> 128 chars): old
    tool results are shrunk while kept as real, paired `tool` messages before any
    drop or collapse, so a bounded record of every tool result survives for all
    providers (including Anthropic). Whole-turn dropping only happens if even the
    smallest extracts cannot fit.
  - **Config is now honored on every production turn path.** The preflight only
    honors the resolved policy when the caller threads it through `LoopKnobs`. The
    earlier revision wired only the Agent (RPC/ACP) and channel paths; this revision
    threads the resolved `history_pruning` into all remaining production
    turn-construction sites so the fix is not silently bypassed:
    - interactive CLI turn (non-streamed + streamed),
    - daemon `process_message` (via a `history_pruning` param on the `agent_turn` wrapper),
    - delegate child loop (`loop_runtime.history_pruning`),
    - post-turn skill-review fork (via a param on `maybe_run_skill_review`).
    All test callers pass `HistoryPrunerConfig::default()` (behavior-preserving). The
    only `LoopKnobs::default()` left is unit-test fixtures.
  - **Merged current with `master`** (now including #8179's `ResolvedAgentExecution::resolve`
    seam, which relocated `knobs` into `ResolvedRuntimeKnobs`). The merge resolution carries
    the `history_pruning` wiring into the new `ResolvedRuntimeKnobs.knobs` field at every site;
    behavior is identical to the pre-merge branch.
- **Scope boundary:** This fixes the in-loop preemptive trim
  (`preflight_history_maintenance`) on every turn path. It deliberately does NOT change:
  - `proactive_trim_turns` (the channel orchestrator's pre-loop overflow handler,
    a blunt whole-turn drop on a hardcoded 400k-char budget) - the dominant trim for
    the Discord/channel scenario, covered by the sibling PR #8050.
  - `context_compression` (it forwards `temperature` verbatim, which some
    providers reject, e.g. Kimi K2.5 in #6517) - separate follow-up.
  - any user-facing "context full" indication (this PR logs at WARN/INFO only).
  - the context-token budget itself, the Anthropic summary-hiding behavior
    (#7775), the channel `strip_old_tool_context` path, or the inert agent-block
    tunable-keys footgun (#6877; `history_pruning` resolves from the runtime profile).
  Because of the above, this PR is `Related #6517`, not `Closes` - it removes the
  in-loop lossy-collapse cause and the silent-config override, but does not yet
  cover the channel pre-loop path that reproduces #6517 end-to-end (that is #8050).
- **Blast radius:** Runtime preemptive history trimming on every turn iteration
  for all providers, across every production turn path (Agent RPC/ACP x2, channel
  orchestrator, CLI interactive x2, daemon `process_message`, delegate child, and the
  skill-review fork). `LoopKnobs` gains a field; `Default` preserves the prior hardcoded
  behavior, so all `LoopKnobs::default()` callers (now only test fixtures) are unchanged.
- **Linked issue(s):** Related #6517
- **Labels:** suggested `type: bug`, `runtime`, `risk: medium`, `size: M`
  (applied if permitted; otherwise for a maintainer)

```bash
RUSTUP_TOOLCHAIN=1.93.0  # CI-pinned toolchain
cargo fmt --all -- --check
cargo test
```

- **Commands run and tail output** (full pre-push battery on the tree merged with
  current `master`, since CI builds the PR merged with base):
  ```
  push-branch validate  base=upstream/master  branch=fix/context-collapse-amnesia
  [PASS] suppressions / todo / deferred / not_implemented / secrets / em-dash
  [PASS] artifacts / docs-coverage / no-attribution / registry-shadowing
  [PASS] cargo fmt --check +1.93.0
  [PASS] cargo clippy --all-targets +1.93.0
  [PASS] cargo test +1.93.0
  summary: 0 blocking, 0 warning
  RESULT: PASS
  ```
  Targeted: `history_window` 5/5 (incl. the orphan-invariant sweep), `agent_turn`
  8/8, all `skills::review` tests pass. The only intermittent failure seen across
  runs is the known `cron::store::remove_job_emits_structured_cron_delete_event`
  parallel log-capture flake (passes in isolation; unrelated; did not fire this run).
- **Beyond CI - what did you manually verify?** Root-caused from a live runtime
  trace (personal_code over RPC/ACP: budget 128k vs estimated 243k, 613 messages
  collapsed/dropped to 60). After merging #8179 I individually audited every
  production turn-construction site (the 3 merge-conflicted + the auto-merged
  `agent.rs` x2 and `orchestrator`) to confirm the resolved `history_pruning` reaches
  `ResolvedRuntimeKnobs.knobs` and was not reverted to `&LoopKnobs::default()` by the
  refactor.
- **If any command was intentionally skipped, why:** Web/shell batteries skipped
  (no `.ts`/`.sh` changes).

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` - no new keys. The existing
  `history_pruning.collapse_tool_results` and `keep_recent` are now actually
  honored by the preemptive trim on every turn path (previously hardcoded). An agent
  relying on the old always-`keep_recent=4` / always-collapse behavior gets that same
  behavior from the defaults; only agents that set non-default values see the change.

- **Feature flags or config toggles:** None (behavior follows the existing
  `history_pruning` config).
- **Observable failure symptoms:** Grep logs for `Emergency context trim
  overriding history_pruning.enabled=false` (expected WARN when over budget with
  pruning disabled) and `Preemptive extract-ladder trim applied`. A regression
  would surface as `context_length_exceeded` provider errors (trim too weak) or
  renewed tool-result hallucination / topic drift (content not preserved).


Audacity88:
@Nillth thanks for splitting this from the channel-side follow-up. I checked the live PR state, linked issue #6517, adjacent PR #8050, the current diff/source paths, and the remote CI on 021dd8587a0c1ffa2b69f5313d68bd377933ac75.

The implementation direction looks right to me. The preflight trim now uses the agent's resolved `history_pruning.keep_recent` and `collapse_tool_results`, keeps the forced over-budget trim visible with a WARN when `enabled = false`, and tries bounded in-place tool-result extracts before falling back to whole-message pruning. The new tests cover the important policy cases: disabled pruning still trims under pressure, collapse on/off is honored, and the extract ladder can preserve paired tool messages without producing synthetic summaries.

I am holding my formal approval for refreshed merge evidence rather than a code-shape objection. This branch is still based on a57ddd0 while current `master` is fb3df9a after three newer commits, and the current series now includes #8050 for the channel pre-loop trim. Please update against current `master` and let the checks rerun on the refreshed head before this is treated as merge-ready.

Scope note for the final merge record: this PR is correctly `Related #6517`, not `Closes #6517`. It fixes the in-loop preflight/history-pruning half, while #8050 covers the channel pre-loop `proactive_trim_turns` path and #6517 itself is still waiting on a current-version reproduction check.

crates/zeroclaw-runtime/src/agent/turn/history_window.rs
```diff
@@ -25,16 +36,52 @@ pub(crate) fn preflight_history_maintenance(
...
                 let stats = crate::agent::history_pruner::prune_history(
                     history,
-                    &crate::agent::history_pruner::HistoryPrunerConfig {
+                    &HistoryPrunerConfig {
                         enabled: true,
                         max_tokens: context_token_budget,
-                        keep_recent: 4,
-                        collapse_tool_results: true,
+                        keep_recent: pruning.keep_recent,
+                        collapse_tool_results: pruning.collapse_tool_results,
```
singlerider, line 85: This widens `keep_recent` / `collapse_tool_results` from the old hardcoded `4`/`true` to user config, feeding values like 0 or 1 into `prune_history`. The pruner documents that `keep_recent` can interact with the collapse + orphan-sweep phases to evict a protected tool whose parent assistant became a summary (the #5823 note at `history_pruner.rs:271-275`).

None of the four added tests asserts the property that actually breaks Anthropic: zero orphaned tool messages on the final output. I verified it holds locally by running the full pipeline over `keep_recent in {0,1,2,3,5}` x `collapse in {true,false}` and asserting `remove_orphaned_tool_messages(...).removed == 0` on the result (it passes, backstopped by the sweep at `history_pruner.rs:410`). Please add that pipeline-level orphan-invariant test here so a future refactor of the ladder or the pruner cannot break pairing with all current tests still green.

singlerider (changes requested):
I pulled this and went through it with deep suspicion because it lives in the history-pruning path, which has a long history of pairing-orphan regressions. The behavior is correct and I verified the riskiest invariant myself, but the test suite has a gap that this exact subsystem cannot afford. Blocking on the missing test, not on the code.

### 🔴 Blocking - no test asserts the orphan invariant on the pipeline output, which is the precise failure mode this path keeps reintroducing

This PR newly feeds the agent's user-controlled `keep_recent` and `collapse_tool_results` into `prune_history`, where before the preemptive trim always called it with the hardcoded `keep_recent: 4, collapse_tool_results: true`. `history_pruner` itself documents that `keep_recent` interacts dangerously with the collapse and orphan-sweep phases (the #5823 note at `history_pruner.rs:271-275`: a protected tool whose parent assistant got rewritten to a summary gets evicted by the orphan sweep, "silently violating keep_recent"). So broadening the input domain of `keep_recent` here is exactly the kind of change that can resurface orphaned `tool` messages and 400 the request on Anthropic.

The four added tests cover the right behaviors (emergency trim fires when disabled, collapse honored both ways, extract ladder preserves messages) but none of them asserts the one property that matters at the boundary: that the final history contains zero orphaned tool messages. I added a throwaway stress test locally that runs the full `preflight_history_maintenance` over `keep_recent in {0,1,2,3,5}` x `collapse in {true,false}` on an over-budget history and then runs `remove_orphaned_tool_messages` on a clone, asserting `removed == 0`. It passes - the invariant holds, backstopped by the `remove_orphaned_tool_messages` call at the tail of `prune_history` (`history_pruner.rs:410`). So the code is correct. But that guarantee is currently only verifiable by a reviewer writing their own test, and a future refactor of either the ladder or the pruner could break it with all four shipped tests still green.

Please add a pipeline-level orphan-invariant test along those lines (sweep a few `keep_recent` values including 0/1 and both collapse modes, assert no orphans on the output). For a change in this file, the orphan invariant must be pinned by the PR that widens the input domain, not left to the existing backstop.

### 🟢 Praise - the extract ladder is genuinely lossless on pairing

`trim_tool_results_to` (the extracted generalization of `fast_trim_tool_results`) only shrinks the content of `tool`-role messages in `history[..cutoff]` in place; it never removes a message and never touches the assistant `tool_calls` side. So the 2000 -> 512 -> 128 ladder cannot orphan anything by construction - it shrinks payloads while keeping every tool message paired. This is the correct design for the Anthropic case, where synthetic collapse summaries are dropped entirely and only real (truncated) tool content survives. The `extract_ladder_preserves_messages_and_pairing_without_collapse` test pins that no-drop / no-collapse outcome well.

### 🟢 Praise - the silent-override fix is honest

Forcing `enabled` on for an over-budget request is the right call (the alternative is a provider `context_length_exceeded`), and logging at WARN with the estimated/budget/keep_recent/collapse attrs when the agent set `enabled = false` turns a previously invisible override into an auditable one. The knob doc and the inline comment both state plainly that only `collapse_tool_results` and `keep_recent` are honored and why `enabled`/`max_tokens` are not.

### 🟢 Praise - the wiring is type-clean and the blast radius claim holds

`LoopKnobs::default()` sets `history_pruning: HistoryPrunerConfig::default()` (collapse=true, keep_recent=4), which reproduces the prior hardcoded trim for every existing default caller, so only the Agent (RPC/ACP) and channel turn paths see config-driven behavior - matches the stated blast radius. The runtime `history_pruner::HistoryPrunerConfig` is a `pub use` re-export of `zeroclaw_config::scattered_types::HistoryPrunerConfig` (`history_pruner.rs:5`), so the orchestrator assigning `ctx.agent_cfg.resolved.history_pruning.clone()` into the knob is the same type with no conversion, and `proactive_trim_turns` is untouched as claimed.

Verified locally: the 4 new `history_window` tests pass, `cargo clippy --locked -p zeroclaw-runtime --all-targets -- -D warnings` is clean, and my added orphan-invariant stress test passes. Add that orphan test to the PR and this is a solid fix to a genuinely nasty drift bug.

singlerider (approved):
Re-reviewed at head `52129395` with fresh eyes. My prior block was a single missing test, not the code. That test is now present and I verified the full picture locally, so this is an approve.

### 🟢 Praise - the blocking orphan-invariant gap is closed

`pipeline_output_has_no_orphaned_tool_messages_across_configs` sweeps `keep_recent in {0,1,2,3,5}` x `collapse in {true,false}` through the full `preflight_history_maintenance` and asserts `remove_orphaned_tool_messages(...).removed == 0` on a clone of the output. This pins the exact property this path keeps regressing on (#5823, #5743) at the boundary the PR widens, so a future refactor of the extract ladder or the pruner can no longer break pairing with all tests still green. This is the test I asked for; it lands the invariant in the PR that broadens the `keep_recent` input domain rather than leaning on the backstop.

### 🟢 Praise - the invariant is double backstopped, now also pinned

The no-orphans guarantee holds via two independent sweeps: `prune_history` Phase 5 (`history_pruner.rs:410`) and the `remove_orphaned_tool_messages` call at the tail of `preflight_history_maintenance` itself. The new test verifies the composed output of both, across the dangerous `keep_recent=0/1` corners.

### 🟢 Praise - the extract ladder remains lossless on pairing

`trim_tool_results_to` only shrinks `tool`-role content in `history[..cutoff]` in place; it never removes a message and never touches the assistant `tool_calls` side, so the 2000 -> 512 -> 128 ladder cannot orphan anything by construction. Correct design for the Anthropic case where synthetic collapse summaries are dropped and only real truncated tool content survives.

### 🟢 Praise - blast radius and wiring hold

`LoopKnobs::default()` reproduces the prior hardcoded trim (collapse=true, keep_recent=4) for every existing default caller, so only the Agent (RPC/ACP) and channel turn paths see config-driven behavior. The runtime `HistoryPrunerConfig` is a `pub use` re-export of the config type, so the orchestrator assigning `ctx.agent_cfg.resolved.history_pruning.clone()` is conversion-free. `proactive_trim_turns` is untouched, matching the stated scope (`Related #6517`, not `Closes`).

Verified locally at `52129395`: `cargo fmt -p zeroclaw-runtime --check` clean, `cargo clippy --locked -p zeroclaw-runtime --all-targets -- -D warnings` clean, all 5 `history_window` tests pass. Resolving my earlier thread. Good fix to a nasty drift bug.

Nillth:
The orphan-invariant gap is now covered: `pipeline_output_has_no_orphaned_tool_messages_across_configs` sweeps `keep_recent` in {0,1,2,3,5} x `collapse` in {true,false} over an over-budget history, runs `remove_orphaned_tool_messages` on the result and asserts `removed == 0` - exactly the pipeline-level property you asked to pin, so a future refactor of the ladder or the pruner can't break pairing with the other tests still green. Thanks for the careful review and for verifying the invariant by hand; that's what made the missing test obvious. Mind taking another look when you have a moment?

Audacity88 (changes requested):
Thanks for the refresh and for adding the orphan-invariant test. I rechecked current head `443492feb35eb000cb13a4a3c91f8def708bc1a1`, the linked #6517 scope, the #8050 channel-side split, prior reviews/comments, the current diff/source, and the live CI. The core trim design looks good, but I still see one config-wiring gap that leaves the advertised behavior incomplete.

### 🔴 Blocking — Some resolved runtime turn paths still use default pruning knobs

The new `preflight_history_maintenance` only honors `keep_recent` and `collapse_tool_results` when the caller passes the resolved config through `LoopKnobs`. The `Agent` struct paths and channel orchestrator do that now, but the CLI/daemon paths in `crates/zeroclaw-runtime/src/agent/loop_.rs` still pass `LoopKnobs::default()` when calling `run_tool_call_loop` even though they already use the same resolved agent runtime for `max_context_tokens`, `max_tool_result_chars`, `parallel_tools`, and related knobs. One of those paths even runs the initial explicit prune with `agent.resolved.history_pruning`, then falls back to default pruning knobs once the in-loop over-budget preflight fires. Delegate child loops have the same shape: `resolve_loop_runtime(...)` provides `loop_runtime`, the call passes `loop_runtime.max_context_tokens`, but the preflight still receives `LoopKnobs::default()`.

That means those real turn paths still get the old hardcoded emergency behavior under context pressure: `keep_recent = 4`, `collapse_tool_results = true`, and no WARN that `enabled = false` was overridden. If an operator set `collapse_tool_results = false` or a non-default `keep_recent`, the config is still ignored there, so the lossy-collapse behavior this PR is fixing can still happen outside the two newly wired callers.

Please thread the resolved pruning config into those `ToolLoop` calls too, or narrow the PR/body to make the remaining exclusions explicit. Since those call sites already have `agent.resolved.history_pruning` or `loop_runtime.history_pruning` in scope, this should be a small wiring change: build a local `LoopKnobs { history_pruning: ..., ..LoopKnobs::default() }` and pass that instead of `LoopKnobs::default()`.

### ✅ Resolved — The orphan-invariant gap is now pinned

The added `pipeline_output_has_no_orphaned_tool_messages_across_configs` test resolves the earlier orphan-invariant blocker: it sweeps `keep_recent` values including 0/1 and both collapse modes through the composed preflight output, then asserts no orphaned tool messages remain. That is the right boundary property for this subsystem.

### 🟢 What looks good — The extract ladder is the right primitive

The 2000 -> 512 -> 128 in-place shrink path keeps real `tool` messages and their tool-call pairing intact before falling back to whole-message pruning. That is a strong design for providers that drop synthetic prune summaries, because the model still gets bounded real tool-result content instead of a content-free marker.

All visible CI checks are green at the current head. I did not run local Cargo for this review; I relied on live PR state, CI, prior reviewer evidence, linked issue/PR context, and source inspection.

singlerider (commented):
Re-checked at head `443492fe` (newer than my approve, which was at `52129395`). @Audacity88's blocker is valid and I concur. My earlier approve is stale against this gap, and I want to be explicit that I missed it.

### 🔴 Blocking: concur with @Audacity88 on the unwired CLI/daemon and delegate turn paths

My approve praised the blast radius as correct ("only the Agent (RPC/ACP) and channel turn paths see config-driven behavior"). That was wrong: those CLI/daemon turn paths SHOULD be config-driven too, and they aren't. Confirmed at head:

- `loop_.rs:1758` and `loop_.rs:2245` both thread `agent.resolved.*` for `max_tool_iterations`, `parallel_tools`, `max_tool_result_chars`, `strict_tool_parsing`, and `pacing`, but pass `knobs: &LoopKnobs::default()`. So `agent.resolved.history_pruning` is in scope at those exact sites yet the preflight runs with `keep_recent = 4`, `collapse_tool_results = true`, and no WARN when the operator set `enabled = false`.
- `tools/delegate.rs:1926` (child loop) has the same shape: `loop_runtime` is resolved and used for `max_context_tokens` etc., but the preflight still gets `LoopKnobs::default()`.

The result is exactly the lossy collapse this PR is fixing, still live on CLI/daemon turns and delegate children. An operator who sets `collapse_tool_results = false` or a non-default `keep_recent` is still ignored on those paths.

Fix is the small wiring change @Audacity88 described: build `LoopKnobs { history_pruning: agent.resolved.history_pruning.clone(), ..LoopKnobs::default() }` (and the `loop_runtime.history_pruning` equivalent in delegate) and pass that instead of `LoopKnobs::default()`. Or narrow the PR body to make the remaining default-knob paths an explicit, documented exclusion.

### 🟢 Still correct

The extract ladder (2000 -> 512 -> 128 in-place shrink keeping real tool messages and pairing intact) is the right primitive, and the `pipeline_output_has_no_orphaned_tool_messages_across_configs` sweep pins the orphan invariant at the `keep_recent=0/1` corners. Those stand.

Not re-approving over @Audacity88's active changes-requested; concurring with it. Thread the resolved pruning config into the remaining `ToolLoop` calls and this closes.

Nillth:
@Audacity88 @singlerider Good catch, you are both right - the preflight only honored
`keep_recent`/`collapse_tool_results` where the caller threaded the resolved config through
`LoopKnobs`, and the CLI/daemon and delegate-child turn paths were still passing
`LoopKnobs::default()`. That left the exact lossy emergency behavior this PR is fixing
(`keep_recent = 4`, `collapse_tool_results = true`, no WARN on `enabled = false`) live on those
paths. Fixed by threading the resolved pruning config into every remaining real turn path:

- The interactive CLI turn in `loop_.rs` (non-streamed and streamed) now builds
  `LoopKnobs { history_pruning: agent.resolved.history_pruning.clone(), ..LoopKnobs::default() }`
  instead of `LoopKnobs::default()`. `agent.resolved.history_pruning` was already in scope at both
  sites.
- The delegate child loop in `tools/delegate.rs` builds the same with
  `loop_runtime.history_pruning.clone()`, matching the other `loop_runtime.*` knobs that call
  already threads.

While auditing for completeness I swept every production turn-construction site and found two more
with the same gap that were not in either of your lists:

- The daemon `process_message` turn reaches the loop through the `agent_turn` wrapper, which
  hardcoded `LoopKnobs::default()` internally. That wrapper already flattens the other resolved
  values (`max_tool_result_chars`, `context_token_budget`, etc.) into positional params, so I
  threaded a `history_pruning: &HistoryPrunerConfig` param through it the same way and passed
  `agent.resolved.history_pruning` from `process_message`.
- The post-turn skill-review fork (`skills/review.rs`) runs a real tool loop over the
  just-finished conversation with `context_token_budget = max_context_tokens`, so it can hit the
  same preflight and would collapse tool results against an operator's `collapse_tool_results =
  false`. `maybe_run_skill_review` got the same `history_pruning` param; its only production caller
  (inside `run`, where the resolved config is already in scope) passes
  `agent.resolved.history_pruning`.

All test callers of both wrappers pass `HistoryPrunerConfig::default()` (behavior-preserving). With
that, all 8 production turn-construction paths (Agent RPC/ACP x2, channel orchestrator, CLI
interactive x2, daemon process_message, delegate child, skill-review fork) honor the resolved
pruning config. The only `LoopKnobs::default()` left in the tree is unit-test fixtures and the
`safety_net.rs` test harnesses, which have no operator turn config in scope.

I also merged the branch up to current `master` (now including #8179's
`ResolvedAgentExecution::resolve` seam, which relocated `knobs` into `ResolvedRuntimeKnobs`). The
`history_pruning` wiring is carried into the new `ResolvedRuntimeKnobs.knobs` at every site, so the
refactor did not revert it; behavior is identical.

Validated on the merged tree at the CI-pinned toolchain (`+1.93.0`):
- `cargo fmt --all -- --check` clean.
- `cargo clippy --all-targets -- -D warnings` clean.
- `cargo test` green; `history_window` 5/5 (incl. the orphan-invariant sweep), `agent_turn` 8/8,
  `skills::review` pass.

Refreshed head is current with `master`; CI should rerun clean. Mind taking another look?

Audacity88 (approved):
Context: I re-reviewed current head `552e2a79a84b84c5118e2b9907c428394d7f5821`, the current PR body, linked issue #6517, sibling PR #8050, prior reviews/comments from @singlerider and my prior review, the live diff, current source around the turn engine call sites, and the current GitHub CI. I did not run local Cargo or a live long-context CLI/daemon smoke for this review-only pass; I relied on source inspection, the current green CI, and the author’s full validation record.

### ✅ Resolved — resolved history pruning now reaches the real turn paths

My prior blocker was that the in-loop preflight only honored `history_pruning.keep_recent` and `collapse_tool_results` where callers threaded resolved config through `LoopKnobs`; the CLI, daemon, delegate-child, and skill-review paths still used default pruning knobs. The current head fixes that shape.

`agent_turn` now takes a `HistoryPrunerConfig` reference and `process_message` passes `agent.resolved.history_pruning`. The interactive CLI non-streamed and streamed paths build `LoopKnobs` from `agent.resolved.history_pruning`, the delegate child loop uses `loop_runtime.history_pruning`, and the post-turn skill-review fork receives and forwards the resolved config as well. The embedded `Agent::turn` / `turn_streamed` paths and the channel orchestrator also carry resolved pruning into `ResolvedRuntimeKnobs.knobs`. The remaining default-knob uses I saw are default preservation or test fixtures, not the production turn paths from the earlier review.

I also agree with @singlerider’s re-check that the earlier gap was real. With this head, the path that still used the old `keep_recent = 4` / `collapse_tool_results = true` emergency behavior is no longer the reviewed production shape.

Review-state note: @singlerider concurred with this blocker at the prior head, and I have not seen a new @singlerider review on `552e2a79a84b84c5118e2b9907c428394d7f5821`. This approval clears my prior `CHANGES_REQUESTED`; it is not a merge-ready statement over any remaining requested-reviewer or maintainer checks.

### ✅ Resolved — the orphan-invariant test still pins the risky boundary

The `pipeline_output_has_no_orphaned_tool_messages_across_configs` test remains the right regression for @singlerider’s original blocker: it sweeps the dangerous `keep_recent` corners and both collapse modes through the full preflight output, then verifies that no orphaned tool messages survive. That keeps the Anthropic/MiniMax pairing failure mode pinned at the boundary this PR widens.

### 🟢 What looks good — the in-loop trim now keeps useful tool evidence before dropping turns

The 2000 -> 512 -> 128 extract ladder still looks like the right primitive for the in-loop trim. It shrinks older `tool` message payloads in place before falling back to the deeper pruner, so providers that drop synthetic pruning summaries still receive bounded real tool-result content. When even the smallest extracts do not fit, the code still forces an emergency trim to avoid provider context overflow, but it honors the configured `keep_recent` / `collapse_tool_results` and logs a WARN if it has to override `history_pruning.enabled = false`.

Scope-wise, the PR body’s `Related #6517` call is still correct. This branch fixes the in-loop preflight/config-wiring side; #8050 covers the channel pre-loop `proactive_trim_turns` side, and #6517 itself is still waiting on a current-version reproduction check.

Nillth:
Superseded by #8196, which reworks history trimming into a single whole-turn function (`trim_to_recent_turns`) with a visible `HistoryTrimmed` event and drives off the same `history_pruning` budget knob. The maintainers chose that reactive, turn-bounded approach over the preemptive token-budget ladder in this PR, so closing this as superseded.

PR #8050, opened 2026-06-20
fix(channels): preserve tool-result content when proactively trimming channel history

### LINKED ISSUE #8049: bug(channels): proactive_trim_turns drops whole turns, discarding tool results (context drift on long channel sessions)
### Affected component
channels / orchestrator

### Current behavior
`proactive_trim_turns` (crates/zeroclaw-channels/src/orchestrator/mod.rs) is the channel inbound overflow handler: when stored conversation history exceeds `PROACTIVE_CONTEXT_BUDGET_CHARS` it drops the oldest *whole* turns:

```rust
turns.drain(..drop_count);   // oldest-first, content and all
```

This discards those turns' tool results entirely, so on long channel (e.g. Discord) conversations the model responds as if earlier messages never happened (context drift / "responding as if previous messages don't exist"). It is the channel-path reproduction of #6517.

It is also the only overflow bound for the two provider calls the channel makes *before* `run_tool_call_loop` (the context-compressor summarizer call and the reply-intent classifier), which the in-loop `preflight_history_maintenance` never sees, so the trimming must stay at this layer.

### Expected behavior
Preserve content under pressure: shrink older `tool`-role payloads to a bounded in-place extract (keeping the messages and their tool-call pairing) before dropping any whole turn, and only drop whole turns if the extracts still do not fit.

### Notes
Related #6517. A separate, non-blocking follow-up could additionally drive the trim budget from agent config rather than the hardcoded `PROACTIVE_CONTEXT_BUDGET_CHARS`.

Nillth:
## Summary

- **What changed and why:**
  - `proactive_trim_turns` (the channel orchestrator's pre-loop overflow handler)
    dropped the oldest *whole* turns once history exceeded
    `PROACTIVE_CONTEXT_BUDGET_CHARS`, discarding their tool results entirely. On
    long channel conversations the model then responded as if earlier messages
    never happened (context drift / hallucination) - the channel reproduction of
    #6517.
  - Now it shrinks older `tool`-role payloads to a bounded in-place extract
    (`fast_trim_tool_results`, 2000-char floor) BEFORE dropping any whole turn,
    keeping the messages and their tool-call pairing intact, then recomputes the
    budget. Whole turns are only dropped if the extracts still do not fit; the
    most recent turns are protected from shrinking and the current turn is never
    dropped.
- **Scope boundary:** Does not change the trim budget itself (still the
  hardcoded `PROACTIVE_CONTEXT_BUDGET_CHARS`; making it config/provider-driven is
  a separate follow-up), the in-loop `preflight_history_maintenance` trim
  (covered by #8048), `strip_old_tool_context`, or `context_compression`. Uses
  the existing `fast_trim_tool_results` (master), not the escalating extract
  ladder from #8048, so this lands independently.
- **Blast radius:** Channel inbound path only (`proactive_trim_turns`, called per
  inbound message). Function signature and return semantics (whole turns dropped)
  unchanged; it now drops strictly fewer turns. It mutates the transient
  request-building copy `prior_turns` only - verified that copy is never persisted
  back to the session store (same pattern as the existing
  `strip_historical_image_payloads` in-place mutation).
- **Why here and not the in-loop trim:** `proactive_trim_turns` is the only
  overflow bound for the two provider calls the channel makes *before*
  `run_tool_call_loop` - the context-compressor summarizer call and the
  reply-intent classifier - which `preflight_history_maintenance` never sees. So
  this is the right layer for the channel fix, and content preservation now also
  benefits those pre-loop calls.
- **Linked issue(s):** Closes #8049. Related #6517.
- **Labels:** suggested `bug`, `channels`, `runtime`, `risk: medium`, `size: S`.

```bash
RUSTUP_TOOLCHAIN=1.93.0
cargo fmt --all -- --check
cargo test
```

  ```
  [PASS] cargo fmt --check
  [PASS] cargo clippy --all-targets
  [PASS] cargo test
  summary: 0 blocking, 0 warning
  RESULT: PASS
  ```
  Targeted: `cargo +1.93.0 test -p zeroclaw-channels --lib proactive_trim` = 4/4
  (3 pre-existing tests still green - they use user/assistant roles, so the new
  tool-result shrink is a no-op and prior drop behavior is preserved - plus 1 new
  `proactive_trim_shrinks_tool_results_before_dropping_turns`).
- **Beyond CI - what did you manually verify?** Traced that `proactive_trim_turns`
  is the binding trim for the channel/#6517 reproduction (its ~400k-char budget
  bites before the in-loop 128k path). Verified `prior_turns` is a transient
  request copy that is never persisted back, so in-place truncation cannot corrupt
  stored history.
- **If any command was intentionally skipped, why:** Web/shell batteries skipped
  (no `.ts`/`.sh` changes).

## Compatibility (required)

- Backward compatible? `Yes`

- **Observable failure symptoms:** Grep logs for "Proactively trimmed
  conversation history" with high `dropped_turns` on long channel sessions, or
  renewed channel context drift / hallucination. A regression in the other
  direction would surface as `context_length_exceeded` (trim too weak).

### DIFF
--- a/crates/zeroclaw-channels/src/orchestrator/mod.rs
+++ b/crates/zeroclaw-channels/src/orchestrator/mod.rs
@@ -122,6 +122,7 @@ use zeroclaw_config::schema::Config;
 use zeroclaw_memory::{self, MEMORY_CONTEXT_CLOSE, MEMORY_CONTEXT_OPEN, Memory};
 use zeroclaw_providers::reliable::{scope_provider_fallback, take_last_provider_fallback};
 use zeroclaw_providers::{self, ChatMessage, ModelProvider, ProviderDispatch};
+use zeroclaw_runtime::agent::history::fast_trim_tool_results;
 use zeroclaw_runtime::agent::loop_::{
     LoopKnobs, ToolLoop, apply_policy_tool_filter, apply_text_tool_prompt_policy,
     build_tool_instructions_for_names, clear_model_switch_request, get_model_switch_state,
@@ -1831,13 +1832,35 @@ fn compact_sender_history(ctx: &ChannelRuntimeContext, sender_key: &str) -> bool
     true
 }
 
+/// Number of most-recent turns whose tool-result payloads are kept at full size
+/// when proactively trimming. The active exchange stays intact; only older
+/// tool results are shrunk to a bounded extract.
+const PROACTIVE_TRIM_PROTECT_LAST_N: usize = 4;
+
 /// Proactively trim conversation turns so that the total estimated character
-/// count stays within [`PROACTIVE_CONTEXT_BUDGET_CHARS`].  Drops the oldest
-/// turns first, but always preserves the most recent turn (the current user
-/// message).  Returns the number of turns dropped.
+/// count stays within [`PROACTIVE_CONTEXT_BUDGET_CHARS`].
+///
+/// Tool-result content is the bulk of an overflowing channel history, so before
+/// dropping any whole turns this first shrinks older `tool`-role payloads to a
+/// bounded head-extract in place (via [`fast_trim_tool_results`]). A trimmed
+/// tool result still tells the model what the tool returned; a *dropped* turn is
+/// gone entirely, which is what makes the bot "respond as if previous messages
+/// don't exist" (#6517). Only if the in-place extracts are still over budget are
+/// the oldest whole turns dropped, and the most recent turn (the current user
+/// message) is always preserved. Returns the number of whole turns dropped.
 fn proactive_trim_turns(turns: &mut Vec<ChatMessage>, budget: usize) -> usize {
     let total_chars: usize = turns.iter().map(|t| t.content.chars().count()).sum();
     if total_chars <= budget || turns.len() <= 1 {
+        return 0;
+    }
+
+    // Content-preserving pre-pass: shrink old tool-result payloads in place,
+    // keeping the messages and their tool-call pairing intact.
+    fast_trim_tool_results(turns, PROACTIVE_TRIM_PROTECT_LAST_N);
+
+    // Recompute after the in-place shrink — we may now fit with no turns dropped.
+    let total_chars: usize = turns.iter().map(|t| t.content.chars().count()).sum();
+    if total_chars <= budget {
         return 0;
     }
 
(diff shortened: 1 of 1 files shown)


Audacity88:
@Nillth thanks for carving this out as the channel-side half of the tool-result trimming problem. I checked the live PR state, linked issue #8049, adjacent PR #8048, the current diff, the channel/request-lifecycle docs, and the current remote CI.

The implementation direction looks right to me: `proactive_trim_turns` now shrinks older tool-result payloads through the existing `fast_trim_tool_results` helper before falling back to dropping whole cached turns, and that helper preserves the JSON tool-message envelope when `tool_call_id` is present. The new regression test covers the important shrink-before-drop path, and all visible checks are green on 713445aa5947026b1b0b5175c51b15050a94db68.

I am holding my formal approval for the refreshed merge evidence rather than a code-shape objection. Since this branch was based on a57ddd0 and current `master` is now e5762cf after #7982, and #7982 also touched `crates/zeroclaw-channels/src/orchestrator/mod.rs`, this needs an update against current `master` and fresh checks before I can treat the reviewed head as mergeable.

One validation note for the final merge record: because this changes the channel inbound history path, we should either have a channel-flow / near-live smoke for the affected behavior or explicitly accept that this slice is covered by the focused unit regression plus CI because the change is isolated to the pre-provider history-shaping helper and does not touch any platform adapter.

singlerider (approved):
I pulled this and gave it the same suspicion as #8048 since it is the channel-side sibling in the trim path. It is the cleaner of the two: the change cannot orphan anything by construction, and the transient-copy claim checks out. Approving.

### 🟢 Praise - the pre-pass is structurally incapable of orphaning

The only behavioral change is inserting `fast_trim_tool_results(turns, PROACTIVE_TRIM_PROTECT_LAST_N)` before the existing whole-turn drain, then recomputing the budget and early-returning if it now fits. `fast_trim_tool_results` only shrinks the content of `tool`-role messages in `turns[..cutoff]` in place; it never removes a message and never touches the assistant `tool_calls` side, so it cannot break tool-call pairing. That is the right primitive for the #6517 channel reproduction: a shrunk tool result still tells the model what the tool returned, whereas a dropped turn is gone. Net effect is strictly fewer whole-turn drops.

### 🟢 Praise - the whole-turn drain is unchanged, so no new drop-path risk

The `excess`/`drop_count` loop and `turns.drain(..drop_count)` are byte-for-byte the master logic, still preserving the last turn (`drop_count < turns.len().saturating_sub(1)`). Any pre-existing orphan-at-head concern in that drain is not introduced or worsened here, and the new pre-pass reduces how often it runs. I confirmed both directions locally: the new test covers the shrink-suffices case, and a throwaway test of mine (paired assistant+tool exchanges too big even after the 2000-char shrink, tiny budget) confirmed the drain still fires and the last turn survives.

### 🟢 Praise - transient-copy claim verified, stored history cannot be corrupted

`prior_turns` comes from `normalize_cached_channel_turns(prior_turns_raw)` (orchestrator/mod.rs:4342), is mutated in place, and is only ever `history.extend(prior_turns)` into the request-build buffer (line 4443). It is never written back to `session_store`; persistence happens separately in `append_sender_turn`. So the in-place truncation affects only the transient request copy, same pattern as the adjacent `strip_historical_image_payloads`.

### 🟢 Praise - correctly scoped layer

Placing the fix in `proactive_trim_turns` rather than the in-loop trim is right: this is the only overflow bound for the two pre-loop provider calls (context-compressor summarizer and reply-intent classifier) that `preflight_history_maintenance` never sees, so content preservation now also benefits those. Using master's `fast_trim_tool_results` rather than #8048's extract ladder keeps this independent of that PR.

Verified locally: `cargo test --locked -p zeroclaw-channels --lib proactive_trim` -> 4/4 (3 pre-existing + 1 new), `cargo clippy --locked -p zeroclaw-channels --all-targets -- -D warnings` clean.

One non-blocking suggestion: the new test only exercises the shrink-suffices path (`dropped == 0`). A second case where the shrink is insufficient and the drain still fires (asserting the last turn survives) would pin the mixed path too. Not required for merge given the drain logic is unchanged. Approving.

Issue #7756, opened 2026-06-16
[Bug]: native/MCP tools unavailable on OpenAI Responses/reasoning and Anthropic turns


perlowja:
### Affected component

provider

### Severity

S1 - workflow blocked

### Current behavior

For an agent with an MCP server, whether the model actually *receives* the registered tools on its turn depends on the model. The MCP server connects and tools register on every run, but on the model turn some models get no tools and reply "no tools available this turn" (`native_tool_calls: 0`).

Registration is confirmed on every run, regardless of model:

```
MCP server `investorclaw` connected — 23 tool(s) available
MCP: 23 tool(s) registered from 1 server(s)
```

Same agent, 3 identical warm runs per row. `T` = a `tool_call_start` fired; `x` = `native_tool_calls: 0` and the model replies it has no tools:

| model | provider / wire_api | runs |
|---|---|---|
| gpt-4.1 (non-reasoning) | openai-compatible gateway, chat_completions | **T T T** — reliable |
| gpt-5.5 (reasoning) | openai-compatible gateway, chat_completions | **x x x** |
| gpt-5.5 (reasoning) | OpenAI OAuth (`requires_openai_auth`), responses | **x x** |
| claude-opus-4-8 | openai-compatible gateway | **x T x** — intermittent |
| claude-opus-4-8 | anthropic direct API (`api.anthropic.com`) | **x x x** |

The only reliable config is gpt-4.1 (non-reasoning) on chat_completions. Reasoning models (gpt-5.5) get no tools on either wire_api, and Anthropic claude is intermittent (~1/3) on both the openai-compatible gateway and the native Anthropic provider. The failing axis appears to be per-model native-tool handling (reasoning models and Anthropic), not the gateway and not purely `wire_api`.

### Expected behavior

If an MCP server's tools register successfully for an agent, those tools should be presented to the model on its turn for every supported provider/model, not only for non-reasoning chat_completions models. The agent should be able to call the tools regardless of whether the model is a reasoning model or an Anthropic model.

### Steps to reproduce

1. Configure one agent with one MCP server (≥1 tool).
2. Run the agent with `model_provider` = gpt-4.1 on an openai-compatible gateway (chat_completions). Tools fire reliably.
3. Switch the same agent to a reasoning model (gpt-5.5) on the same gateway. Re-run the identical prompt. Tools do not fire (`native_tool_calls: 0`), model replies it has no tools.
[...]

PR #8053, opened 2026-06-20
fix(runtime): present native/MCP tools to reasoning models in the system prompt (#7756)

perlowja:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - The agent system prompt decided tool availability from the **in-prompt** text-tool list only (`has_tools = !tools.is_empty()`). Native/MCP tools are delivered in the provider API `tools` field, not prompt text, so for native delivery that list is empty and the prompt injected *"No tools are available for this turn, so do not emit tool calls"* — while the request still carried the tools.
  - Reasoning models (gpt-5.5 on the OpenAI Responses API, Claude) **obey** that instruction and return `native_tool_calls: 0`; non-reasoning models ignore the contradiction and call anyway, which is why the bug presented as model-specific.
  - The prompt now derives native-tool availability from the **effective tool specs actually sent this turn** (`native_tool_specs_present`) instead of bare provider capability. When native specs are present it gives the native tool-use instruction; with zero effective tools it keeps the "no tools" boundary.
- **Scope boundary:** Direct runtime agent path only (interactive loop + `process_message`). This PR does **not** touch channels/orchestrator, the gateway `/api/tools` surface, WebSocket `Agent::from_config`, multimodal vision-provider routing, or `/think` directive filtering — those share a deeper "prompt-built-once vs per-turn tool resolution" issue tracked in #8054.
- **Blast radius:** System-prompt text for runtime agent turns. No API/wire/schema/config changes.
- **Linked issue(s):** Related #7756 (resolves the direct-runtime reasoning + Anthropic axis; the gateway/channel/multimodal prompt-consistency is tracked in #8054).

Pinned toolchain `1.93.0` (matches CI):

```
$ cargo fmt --all -- --check                    # clean
    Finished `dev` profile in 51.52s            # no warnings
$ cargo check --no-default-features             # Finished
$ cargo test --test architecture tests_that_persist_config_isolate_the_path   # 1 passed
$ cargo test --test architecture user_facing_strings_route_through_fluent      # 1 passed
$ cargo test -p zeroclaw-runtime --lib          # 2317 passed; 0 failed; 2 ignored
```

- **Beyond CI — manual live verification** (real provider turns, an MCP server exposing 13 tools):
  - **gpt-5.5, OpenAI Responses API (direct):** before — `native_tool_calls: 0`, model replied "data unavailable"; after — `native_tool_calls: 2`, tool invoked. The "No tools are available" sentence is gone from the system prompt.
  - **claude-opus-4-8, Anthropic direct API:** after — `native_tool_calls: 1`, tool invoked.
  - The captured outgoing request already carried 13 well-formed tools, confirming this was prompt construction, not wire serialization.
- **What I did NOT verify:** the deferred surfaces (gateway/channels/WebSocket/multimodal/`/think`) — see the follow-up issue.

## Compatibility (required)

- **Fast rollback command/path:** `git revert <sha>` — restores the previous direct-runtime prompt framing.
- **Feature flags or config toggles:** none.
- **Observable failure symptoms:** reasoning/native-tool turns regress to zero tool calls even though request-side native/MCP tool specs are present, or direct runtime turns again include contradictory "No tools are available" framing when effective native specs exist.


Audacity88 (dismissed):
I reviewed the live #8053 state at ede3baec63e095f514f38322badcc2afb13ce5e1, the PR body, the full diff, and the current CI results. The helper-level direction is good: the prompt should key off the effective native tool specs, not bare provider capability. I do not think this is ready to approve yet because the direct interactive path still does not compute that signal at the same point as the request path.

### 🟢 What looks good — The core signal is the right one

The new helper is pointed at the same `build_iteration_tool_specs` path that the turn loop uses for the provider request, and the added unit coverage catches the important provider-capability-only mistake. That is the right shape for the direct single-turn case: native-capable provider plus actual specs means “tools are available,” and native-capable provider plus zero specs keeps the no-tools boundary.

### 🔴 Blocking — Interactive turns still build the prompt before the per-turn filter

In `run`, the base system prompt now computes `native_tool_specs_present` once from the startup `message` at prompt construction time. That prompt is then saved as the base prompt for later turns. Inside the interactive REPL loop, each user turn computes `excluded_tools` again from that turn's `effective_input` immediately before calling `run_tool_call_loop`. Those two values can diverge for `tool_filter_groups`, which are explicitly a per-turn prompt/context exposure control.

The concrete failure mode is an interactive agent with only dynamically exposed MCP tools. If the REPL starts without a startup message, or starts with a message that activates a different filter group, the base system prompt can say native tools are available. On a later user turn whose `effective_input` excludes all MCP tools, the actual provider request sends zero tool specs, but the system prompt still carries the native-tool task framing. That means the direct interactive path still does not satisfy the PR body’s claim that the prompt derives tool availability from “the effective tool specs actually sent this turn.”

Please move the direct interactive prompt/tool-availability decision to the same per-turn point as the `excluded_tools` passed into `run_tool_call_loop`, or narrow the PR and follow-up issue so this remaining direct-runtime REPL mismatch is explicit and tracked. As written, I would not merge this as the direct-runtime fix because the direct interactive loop still has the prompt/request split this PR is trying to close.

perlowja:
Thanks @Audacity88 — you're right, and this was the real gap. The startup-time `native_tool_specs_present` was baked into `base_system_prompt`, while `excluded_tools` was recomputed per turn, so a `tool_filter_groups` turn that excluded all dynamically-exposed MCP tools sent zero specs to the provider while the prompt still carried native-tool framing.

Fixed in `8a7649fda` by moving the prompt's tool-availability decision to the **same per-turn point** as the request:

- Extracted `build_system_prompt_for_turn(...)`, which derives `native_tool_specs_present` via `native_tool_specs_present_for_turn(...)` — the same `build_iteration_tool_specs(...)` path that feeds the actual provider request — from **this turn's** `excluded_tools`.
- Wired it into every REPL call site: the initial turn, each subsequent turn, and the retry/model-switch re-entries. No path uses a stale startup-time prompt for the provider request anymore.
- `base_system_prompt` is still restored after each turn, so per-turn framing/thinking-prefix never leaks into saved history.
- The single-shot path is unchanged in behavior (same builder, computed once).

Added a regression test (`interactive_turn_system_prompt_uses_effective_dynamic_mcp_specs`) that reproduces your exact failure mode: a dynamic MCP-only agent gets **no-tools** framing on a turn that excludes the tools and **native-tools** framing on a turn that includes them — it fails on the pre-fix code and passes after.

Validation: full CI gate run locally at the pinned 1.93.0 toolchain (`cargo clippy --workspace --exclude zeroclaw-desktop --features ci-all -D warnings`, `check --features ci-all`, `check --no-default-features`, both architecture guards, and `nextest` — 9338 tests pass including the new one).

Audacity88 (approved):
I reviewed the current #8053 head `8a7649fda4249f7684674cb845985576bca30297`, the dismissed blocker review, the PR body, linked #7756/#8054, the live diff/source for `loop_.rs` and `system_prompt.rs`, current CI, and merge freshness. I did not run local Cargo for this re-review.

### ✅ Resolved — Direct REPL prompt framing now follows the request-side tool set

The blocker from my prior review is addressed. The interactive REPL path now rebuilds the system prompt after computing the turn's `excluded_tools`, immediately before `run_tool_call_loop`, and it uses `native_tool_specs_present_for_turn(...)` through the same `build_iteration_tool_specs(...)` source that determines the provider request's effective specs. The retry/model-switch loop also rebuilds the prompt before re-entering the provider call, so the startup `base_system_prompt` no longer carries stale native-tool framing into later dynamic-filtered turns.

The added regression test covers the important MCP-only dynamic-filter shape: a turn that excludes the MCP tool gets the no-tools framing, while a turn that includes it gets the native-tools framing. That pins the failure mode that previously made reasoning/Anthropic turns obey "No tools are available" even when the request-side native specs were the intended source of truth.

### 🟡 Warning — Branch freshness still needs handling before merge

The current branch is still `43/2` behind/ahead of `master`, with same-file base churn in `crates/zeroclaw-runtime/src/agent/loop_.rs` and both changed files on the risky runtime surface. Before landing, please either update the branch and wait for fresh CI, run an exact merge-result smoke, or have a maintainer explicitly accept the stale-risk after naming those triggers.

### 🔵 Suggestion — Keep remaining non-REPL surfaces in #8054

I am treating the `process_message` `/think` raw-vs-stripped filtering edge, plus the channel/gateway/WebSocket/multimodal surfaces, as the explicitly deferred #8054 scope rather than a blocker to this focused direct REPL fix. The PR body already calls out that boundary, and #8054 tracks the broader "prompt availability must match this turn's actual request tools" cleanup.

Issue #7462, opened 2026-06-10
[Bug]: 74 test failures on Windows — Unix-only test commands, path semantics, console encoding


NiuBlibing:
### Affected component

tooling/ci

### Severity

S2 - degraded behavior

### Current behavior

Running the workspace test suite on Windows 11 (Simplified Chinese, console code page 936) yields **74 failing tests** on current master. CI does not catch this because the Test job only runs on Linux (companion feature request: #7461).

The failures group into four root causes:

**1. Tests hardcode Unix commands / shell (~40+ failures)**
Test cases spawn `env`, `cat`, `sleep`, `sh`-style scripts, and `rg`/`grep`, none of which exist on a stock Windows machine. Affected: `zeroclaw-runtime` shell/cron/tunnel/skill tests, `zeroclaw-tools` `content_search`, `zeroclaw-gateway` cron API tests, `zeroclaw-api`, `zeroclaw-providers` bedrock `credential_process`, `zeroclaw-hardware` subprocess (spawns a `.sh` file directly → os error 193).

**2. Windows path semantics (~20 failures)**
- `/etc/...`-style paths are **not absolute** on Windows (`Path::is_absolute()` is false without a drive letter), so "Path blocked" security-guard tests take a different branch (the path is still rejected, but with the workspace-allowlist message instead of `Path blocked`) — affects `file_edit`, `file_write`, `wrappers`, and the `*_rejects_path_outside_workspace` tests in `claude_code` / `claude_code_runner` / `codex_cli` / `gemini_cli` / `opencode_cli`.
- `canonicalize` returns `\\?\C:\...` verbatim-prefixed paths, causing `StripPrefixError` panics and `rpc::attachments` marker assertions to fail.
- `\` vs `/` separators break the orchestrator skills-location assertion (`<location>skills/code-review/SKILL.md</location>`) and likely the `[IMAGE:` marker assertions in `agent::tests` / `agent::history`.

**3. Console encoding (1 failure)**
`tools::shell::tests::decode_output_valid_utf8_roundtrips` feeds UTF-8 bytes, but on a CP936 system the intentional code-page transcoding in `decode_output` (shell.rs) decodes them as GBK → mojibake. The test assumes a UTF-8 roundtrip and conflicts with the Windows-specific decoding logic.

**4. Platform-specific gaps (2 failures + 1 duplicate)**
[...]

Issue #7461, opened 2026-06-10
[Feature]: Run the test suite on Windows and macOS in CI, not just Linux


NiuBlibing:
### Summary

Extend the CI Quality Gate test job to a platform matrix (ubuntu / macos / windows) instead of running `cargo nextest run` only on `ubuntu-latest`.

### Problem statement

The Lint and Build jobs in `.github/workflows/ci.yml` already run on a three-OS matrix (`ubuntu-latest` / `macos-14` / `windows-latest`), but the Test job runs `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` on `ubuntu-latest` only.

This means platform-specific regressions — path semantics, shell invocation, console encoding, platform-gated features — are never caught by CI. Concrete consequence: running the workspace test suite locally on Windows currently yields **74 failing tests** (tracked separately in the companion bug issue) while master CI stays green.

### Proposed solution

- Convert the Test job to the same `strategy.matrix` used by Lint/Build (`ubuntu-latest`, `macos-14`, `windows-latest`).
- Skip or replace the mold-linker install step on non-Linux runners.
- Since Windows currently has a large number of known failures, roll out in two stages:
  1. Introduce the Windows test job as non-blocking (`continue-on-error: true` or a separate non-required check) to establish visibility.
  2. Flip it to required once the companion bug issue (Windows test failures) is resolved.

### Non-goals / out of scope

- Fixing the existing Windows test failures themselves (tracked in the companion bug issue).
- Running tests on additional targets (Android, 32-bit, musl, etc.).

### Alternatives considered

- Keep status quo and rely on contributors testing locally — already proven insufficient (74 failures on Windows on current master went unnoticed).
- Add macOS only — macOS shares Unix semantics with Linux, so the incremental coverage is much smaller than Windows.

### Acceptance criteria

- `cargo nextest run` executes on ubuntu, macos, and windows in the Quality Gate workflow
- Windows failures are visible in CI (non-blocking at first, required once green)
- Workflow README/docs reflect the new matrix

### Architecture impact

`.github/workflows/ci.yml` (tooling/ci) only; no runtime code changes.

### Risk and rollback

Risk: longer CI wall-clock time and possible flakiness on Windows runners. Rollback: remove the matrix entries / revert the workflow change.

### Breaking change?

No

### Data hygiene checks

- [X] I removed personal/sensitive data from examples, payloads, and logs.
- [X] I used neutral, project-focused wording and placeholders.

[...]

PR #8060, opened 2026-06-20
fix(tools): add content_search internal fallback

Audacity88:
## Summary

- **What changed and why:**
  - Adds a third `content_search` backend so the tool can still search readable workspace files when neither `rg` nor `grep` is available on the host.
  - Keeps `rg` as the preferred backend and `grep` as the external fallback, then uses the internal scanner only when both external commands are missing.
  - Preserves the existing `SecurityPolicy` readable-path checks while traversing directories and symlink targets in the internal backend.
  - Fixes internal content-mode formatting so overlapping context windows are emitted once and `max_results` counts matching lines instead of context lines.
  - Updates localized `content_search` tool descriptions so they describe the grep-or-internal fallback behavior.
- **Scope boundary:** This PR only covers the `content_search`/`rg` Windows portability slice from #7462. It does not address the remaining Windows path-semantics, console-encoding, updater, or other test-failure clusters from #7462/#7461.
- **Blast radius:** `zeroclaw-tools` `content_search` behavior on systems without `rg` and/or `grep`; content-search tool descriptions embedded from `zeroclaw-runtime` locales. Existing `rg` and `grep` paths are intended to keep the same command shape and formatter behavior.
- **Linked issue(s):** Related #7462. Related #7461.
- **Labels:** `bug`, `size: L`, `risk: high`, `tool`, `tool: content_search`, `tool:file`.

```text
git diff --check
# passed with no output

cargo fmt -p zeroclaw-tools -- --check
# passed

cargo test -p zeroclaw-tools content_search --lib
running 27 tests
...
test result: ok. 27 passed; 0 failed; 0 ignored; 0 measured; 1312 filtered out; finished in 0.24s

cargo clippy -p zeroclaw-tools --lib -- -D warnings
Checking zeroclaw-tools v0.8.1

rg -n "non-ripgrep fallback|fallback wording" crates/zeroclaw-runtime/locales crates/zeroclaw-tools/src/content_search.rs
# only intentional `non-ripgrep fallback` schema/comment references remained after locale updates

Finished `dev` profile [unoptimized + debuginfo] target(s) in 11m 16s

test result: ok. 9338 passed; 0 failed; 11 skipped; finished in 432.2s
```

- **Beyond CI — what did you manually verify?** The internal-backend regression tests force `SearchBackend::Internal`, cover no-external-tool matching, overlapping context windows, match-count result caps with context lines, and expired deadline handling. Existing `content_search` tests still cover validation, path policy rejection, symlink escape blocking, include filtering, count mode, files-with-matches mode, context output, and the multiline-without-ripgrep error path.
- **If any command was intentionally skipped, why:** No broad local validation remains queued for this head. No live Windows host smoke was run; coverage for the no-`rg`/no-`grep` path is through the forced-internal backend regression tests above plus GitHub's `x86_64-pc-windows-msvc` check.

- New permissions, capabilities, or file system access scope? `No`. The internal fallback uses the existing `content_search` readable workspace scope and calls `SecurityPolicy::is_resolved_path_readable` before reading files.
- New external network calls? `No`.
- Secrets / tokens / credentials handling changed? `No`.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No`.

## Compatibility (required)

- Backward compatible? `Yes`.
- Config / env / CLI surface changed? `No`.
- If `No` or `Yes` to either: No upgrade steps required. Hosts with `rg` keep using `rg`; hosts without `rg` but with `grep` keep using `grep`; hosts without either now use the internal scanner instead of failing at command execution.

- **Fast rollback command/path:** `git revert a61a0a0e1618acaa4f3cfef2687a8a2d97985326`
- **Observable failure symptoms:** `content_search` may return missing, duplicated, or unexpectedly truncated results on hosts without `rg`/`grep`, or may report `Search timed out after 30 seconds.` for large searches that previously failed earlier because no external search command existed.

- `Co-authored-by` trailers added in commit messages for incorporated contributors? `No`.
- If `No`, why (inspiration-only, no direct code/design carry-over): No superseded contributor PR is being carried forward.


singlerider (approved):
First review (no prior reviews or comments). I read the `content_search.rs` diff at head a61a0a0e and focused on the security of the new traversal backend, the backend-selection precedence, and the formatter fix. CI is green. No blockers.

### 🟢 What looks good: the internal scanner re-checks SecurityPolicy at every node

Adding a backend that walks the filesystem itself is the part that most needs scrutiny, and it is handled correctly. `visit_internal_search_path` canonicalizes each path and returns early when `security.is_resolved_path_readable(&resolved)` is false, so unreadable directories and files are skipped rather than scanned. Symlinks are resolved with `canonicalize` to their target and the target is re-checked with `is_resolved_path_readable` before any read, which closes the obvious symlink-escape where a link inside a readable dir points outside the allowed set. Files reached during recursion are canonicalized and re-checked the same way. The policy is enforced at every file, directory, and symlink target, not just at the root, which is the right posture for a deny-by-default tool.

### 🟢 What looks good: backend selection is enum-driven with correct precedence

`SearchBackend` is an enum dispatched with `match` (not a string compare), `detect_search_backend` prefers `rg`, then `grep`, then `Internal` only when both externals are absent, and the existing `rg`/`grep` command shape and formatters are untouched. `multiline` correctly errors on any non-ripgrep backend rather than silently producing wrong results, and the `Internal` arms of the external-search match are `unreachable!()` since that path routes through the internal runner. So the new code is purely additive to hosts that lack both external tools and leaves the common case identical.

### 🟢 What looks good: the formatter fix is real and each claim is tested

The internal content formatter dedups overlapping context windows via a `BTreeSet` of match indexes and counts matching lines against `max_results` (`results_seen` / `total_matches`) rather than context lines, with `entries.sort_by_key(path)` for deterministic output. The three internal-backend tests map one-to-one onto the body's claims: finds matches without external tools, merges overlapping context windows, and `max_results` counts matches not context. The `check_internal_deadline` bail on `TIMEOUT_SECS` bounds traversal so a pathological tree cannot hang the tool. Locale parity is clean across all five `tools.ftl` files for the updated description.

Approving.

Issue #7141, opened 2026-06-03
RFC: Pluggable inbound authentication and canonical principals


singlerider:
# RFC: Pluggable inbound authentication and canonical principals

**Current version:** Rev 8
**Target:** Identity & Access milestone
**Type:** Security / Architecture

## Revision history

| Rev | Date | Summary |
| --- | --- | --- |
| 1 | 2026-06-03 | Initial OIDC and pluggable-provider proposal |
| 2 | 2026-06-03 | Related security-enforcement and sandbox boundaries added |
| 3 | 2026-06-26 | Aligned with the RFC template |
| 4 | 2026-07-03 | Added the password-provider child and recorded the merged provider seam |
| 5 | 2026-08-02 | Maintainer-sponsored ratification rewrite: separated authentication from authorization, defined canonical identity and isolation semantics, narrowed provider obligations, made compatibility impact explicit, and moved delivery bookkeeping to #8289 |
| 6 | 2026-08-03 | Clarified browser and headless OIDC flows, stable local-principal identity, live credential and grant revocation for established connections, storage-level memory ownership, reconnect binding, and local-only recovery |
| 7 | 2026-08-04 | Separated OIDC token purposes, made offline-JWT revocation limits explicit, defined service-principal identity, and fixed deny-by-default grant-composition semantics |
| 8 | 2026-08-05 | Clarified the provider-owned revalidation-secret lifecycle for established connections without changing authentication, principal, or authorization scope |

## Problem

ZeroClaw has multiple inbound surfaces and identity mechanisms but no single enforced authentication and principal model.

The WSS RPC listener can terminate TLS and accept a WebSocket connection without establishing who the client is. TLS protects the channel but does not authorize the client. The existing `tui_id` and `tui_sig` mechanism is issued after initialization for reconnect continuity; it is not proof of an independently established identity.

Existing and proposed identity mechanisms also risk becoming parallel security systems: native pairing tokens, local peer credentials, SSH keys, passwords, OIDC subjects, channel senders, and agent-to-agent credentials. If each mechanism assigns permissions or constructs ownership keys differently, authorization and isolation become provider-dependent and fail-open mistakes become difficult to detect.

The project needs one inbound authentication boundary that can accept different credential types, resolve them into canonical ZeroClaw principals, enforce grants by construction, and keep one principal's sessions and private memory inaccessible to another.

## Proposal

Adopt a pluggable inbound authentication pipeline with three distinct stages:

[...]

PR #8063, opened 2026-06-20
feat(security): Principal type + AuthProvider seam (#7141)

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** Lands the foundational, behaviour-neutral spine for RFC #7141
  ("Authentication Provider support"): the shared authenticated-subject contract and the pluggable
  inbound-auth seam. Splitting #7141's first DoD items (the shared `Principal` identity/claim contract
  and the `AuthProvider` trait + registry) into one small, reviewable, zero-risk PR so the providers,
  dispatch enforcement, and multi-tenant isolation can land on a stable contract afterwards. The
  resolved-grant fields (allowed-agents / config-write / admin) are NOT in this slice; they land
  additively in the later IamPolicy-wiring step, alongside an accepted provider's grant resolution.
  - `zeroclaw-api` (leaf crate): `Principal { id, user_id, roles, scopes, auth_method, mfa_verified,
    expires_at, allowed_aliases }` + `PrincipalId`/`AgentAlias` newtypes + `AuthMethod` /
    `AuthOutcome` / `DenyReason`, serde-derived; `shared_operator()` / `is_authenticated()`.
  - `zeroclaw-runtime/src/security`: the `AuthProvider` trait (async, fail-closed) + a `Credential`
    enum (secret-redacting `Debug`) + a default-deny `ProviderRegistry` (preserves the most-specific
    `DenyReason`; enumerable via `names()`).
- **Scope boundary:** Does NOT implement any provider (`oidc`/`native`/…), does NOT wire the registry
  into the gateway/RPC handshake, does NOT touch dispatch authorization, sessions, or `PairingGuard`.
  Those are the follow-up DoD items. `Principal`/`AuthMethod`/`Credential` are `#[non_exhaustive]` and
  constructed via `Principal::new` + builders — an **extensible** seam, not a frozen one: the RFC #7141
  resolved-grant fields (allowed-agents / config-write / admin), and any not-yet-accepted provider arms,
  are added **additively** by their own scoped change. This PR is scoped to the *accepted* #7141 provider
  set (OIDC / ssh-key / peercred / native); `roles`/`scopes` here are identity *claims* (inputs to grant
  resolution), not resolved grants.
- **Blast radius:** None at runtime. Two new modules (`api::principal`, `security::auth_provider`) plus
  two `pub mod` lines. **Zero production call sites** — nothing constructs a `ProviderRegistry`, calls
  `resolve`/`verify`, or reads a `Principal`. The existing pairing-bearer auth is untouched and remains
  the only auth path. Verified: the diff is pure additions (0 deletions), and grep finds no reference to
  any new symbol outside the two new files.
- **Linked issue(s):** Part of #7141.
- **Labels:** suggest `security`, `runtime`, `config`, `size: M`, `risk: low`, `type: feat` (maintainers apply).

Branch was brought current with base via **merge** (not rebase) — HEAD is a signed merge commit of
`upstream/master` (incl. #7997 the universal-ingress layer) into the change; the module lists auto-merged
with no conflicts. Battery re-run against the merged tree on the **CI-pinned toolchain (1.93.0)**, literal:

```text
$ cargo +1.93.0 fmt -p zeroclaw-api -p zeroclaw-runtime -- --check
(no output — clean)

$ cargo +1.93.0 clippy -p zeroclaw-api -p zeroclaw-runtime --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 09s    # exit 0, zero warnings (gate is -D warnings)

$ cargo +1.93.0 test -p zeroclaw-api principal
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 53 filtered out

$ cargo +1.93.0 test -p zeroclaw-runtime auth_provider
running 6 tests
test security::auth_provider::tests::debug_redacts_secret_material ... ok
test security::auth_provider::tests::empty_registry_is_default_deny ... ok
test security::auth_provider::tests::resolve_falls_back_to_bad_credential_when_no_provider_accepts ... ok
test security::auth_provider::tests::no_credential_is_denied ... ok
test security::auth_provider::tests::matching_provider_authenticates ... ok
test security::auth_provider::tests::resolve_preserves_specific_deny_reason ... ok
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 2314 filtered out
```

- **Beyond CI — what did you manually verify?** Behaviour-neutral: `git show --numstat` shows 0 deletions on
  all 4 files; the only edits to existing files are the two `pub mod` lines; a tree-wide grep finds no
  reference to any new symbol outside the two new files (nothing constructs/calls any of it). `resolve()`
  checked fail-closed on every path; `Credential` `Debug` checked to redact every secret arm.
- **Full `cargo test -p zeroclaw-runtime`:** 2317 passed / 1 failed / 2 ignored. The single failure
  (`cron::store::tests::remove_job_emits_structured_cron_delete_event`) is **pre-existing and flaky** — it
  passes in isolation and is in code unrelated to this change.
- **Gates not run locally (stated, not claimed):** `--workspace --all-targets` can't complete on this host
  (a GUI crate needs system lib `glib-2.0`, not installed; unrelated to this change) — CI runs the
  authoritative full-workspace gate. No web / docs / shell files changed → those gates N/A.

- Secrets / tokens / credentials handling changed? **No** — introduces a `Credential` type that *models*
  presented credentials (with a secret-redacting `Debug`, and it is never `Eq`-compared), but it is not
  wired into any path, so no credential is handled, stored, logged, or compared differently than today.
  Follow-up noted in-code: in-memory secret material is not yet zeroized-on-drop, consistent with the rest
  of the tree (no security module currently scrubs in-memory secrets); a `Zeroizing`/`SecretString`
  convention is a separate, repo-wide hardening tracked under the auth-provider work.
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** — tests use synthetic
  subjects (`alice`, `fixed-bearer`).

## Compatibility (required)

- Backward compatible? **Yes** — purely additive (0 deletions), zero call sites, no config/CLI/wire surface.
- If `No` or `Yes` to either: no upgrade steps — nothing observable changes for existing users.

Low-risk: `git revert <sha>` is the plan. No feature flags, no config toggles (nothing is wired). No
observable failure symptoms possible — the code is inert until a later PR constructs the registry.


Audacity88 (changes requested):
@Nillth I reviewed the current head of #8063 (6fe7111), the linked accepted RFC issue #7141, the empty current PR comment/review state, the full diff, and the adjacent existing `IamPolicy` code on `master`. I did not rerun local Cargo here; CI is green and the author-supplied focused validation is enough for compile/test confidence. My review is focused on whether this new public auth contract is safe to freeze before the follow-up provider and dispatch work lands.

### 🟢 What looks good — The seam is pointed in the right direction

Putting the shared identity type in `zeroclaw-api` and keeping the verification engine in `zeroclaw-runtime/src/security` keeps the data contract separate from the runtime implementation. The `Credential` debug redaction and empty-registry default-deny tests are also the right instincts for this surface. I think the shape is close; the issues below are about not freezing the wrong contract.

### 🔴 Blocking — `Principal` is not the resolved-grants contract #7141 requires yet

The PR body says `Principal` intentionally carries #7141's full `identity + grants` shape now as a frozen contract, and the module docs say the `Principal` shape does not change later. The actual public struct only carries `roles`, `scopes`, `auth_method`, `mfa_verified`, `expires_at`, and `allowed_aliases` (`crates/zeroclaw-api/src/principal.rs:104-140`). That does not cover the accepted #7141 DoD yet: the principal must carry resolved grants for allowed agents, config-modification permissions, and admin/all access. The existing `IamPolicy` also still maps role names to concrete tool/workspace permissions, so `roles`/`scopes` are inputs to grant resolution, not the resolved ZeroClaw grants the RFC says downstream runtime code should consume.

Because this is a public `zeroclaw-api` struct with public fields, adding the missing grant fields later would change the type after this PR has told later PRs to extend behavior, not the type. Please either make this slice explicitly extensible rather than frozen, for example with `#[non_exhaustive]`, private fields plus constructors, or revised docs/body language, or include the explicit resolved-grant shape now with tests that cover agent alias grants, config-write grants, and admin/all grants.

### 🔴 Blocking — A later provider can authenticate after an applicable provider failed closed

`ProviderRegistry::resolve` continues after any provider that accepts a credential returns `Denied`, and a later accepting provider can still return `Authenticated` or `Trusted` (`crates/zeroclaw-runtime/src/security/auth_provider.rs:137-160`). The code comment says this is safe because a later phase will enforce that at most one provider accepts a given credential kind, but the current public API does not make that invariant enforceable: `Credential::Bearer` is shared by both native pairing bearer tokens and OIDC access/ID tokens, `accepts(&Credential)` is an arbitrary predicate, and `register()` currently just appends providers (`crates/zeroclaw-runtime/src/security/auth_provider.rs:38-41`, `crates/zeroclaw-runtime/src/security/auth_provider.rs:91-115`).

If the OIDC provider recognizes a bearer credential and returns `MfaRequired`, a later broadly-accepting bearer provider can still allow the same presented credential. If the intended design is "try the next provider only for generic bad credentials," encode that distinction and add a regression where `MfaRequired` / `Misconfigured` cannot be bypassed by a later provider. If the intended design is exclusive routing, the contract needs an enforceable provider key or credential discriminator now, rather than a future invariant that `register()` cannot currently check.

### 🔴 Blocking — Proposed provider methods should not be frozen into the accepted #7141 contract yet

The accepted #7141 provider set covers OIDC, ssh-key, peercred, and native, and those belong in this foundational seam. The new public contract also includes `AuthMethod::Password` and `Credential::Password` while the code comments say the password provider is only "proposed" / "awaiting buy-in" (`crates/zeroclaw-api/src/principal.rs:73-93`, `crates/zeroclaw-runtime/src/security/auth_provider.rs:35-41`). That is a real security-surface decision. Landing the enum arm and credential shape here makes the unaccepted provider part of the public seam before the RFC/child-issue decision is settled.

Please keep this PR to the accepted #7141 provider contract, or first get the password-provider scope accepted and linked so this PR is not silently widening the security design while claiming to be the stable foundation.

Nillth:
Thanks @Audacity88 — this is the kind of review that keeps a foundational seam honest. All three blocking points are addressed in `173a36c7` (validated: `cargo +1.93.0 fmt --check` + `clippy --all-targets -D warnings` + tests green on both crates).

**1 — `Principal` as the resolved-grants contract / the "frozen" claim.** You're right on both counts: `roles`/`scopes` are *inputs* to grant resolution, not the resolved grants, and a public struct that calls itself frozen is the wrong thing to lock. Rather than freeze it, I made the seam explicitly extensible:
- `Principal`, `AuthMethod`, and `Credential` are now `#[non_exhaustive]`, constructed via `Principal::new` + builder setters (not struct literals from other crates).
- The module/struct docs and the PR body no longer claim the shape is frozen. They now state plainly that `roles`/`scopes` are the identity source's *claims*, and that the RFC #7141 resolved-grant fields (allowed-agents, config-write, admin/all — computed by `IamPolicy`) are added as **additive** `Principal` fields in the IamPolicy-wiring step; `#[non_exhaustive]` keeps that non-breaking.

I kept the resolved-grant fields out of *this* PR so they land with the `IamPolicy` wiring and its tests rather than as empty placeholders here — but if you'd rather see the explicit resolved-grant shape defined now (with the agent/config-write/admin grant tests), I'm happy to pull it forward.

**2 — a later provider authenticating past a closed-failed one.** Real hole, fixed. `resolve()` now treats a *specific* `DenyReason` as **authoritative**: if an accepting provider returns `MfaRequired` / `TokenExpired` / `Misconfigured` / `AliasNotEntitled`, that outcome short-circuits and is returned, so a later, more broadly-`accept`ing provider can't authenticate the same presented credential past it. Only the generic `BadCredential` ("not my credential / wrong secret") falls through to the next accepting provider. New regression `specific_deny_is_not_bypassed_by_a_later_provider` registers an `MfaRequired` provider *before* a catch-all that would otherwise `Trust` the same token, and asserts the result stays `Denied { MfaRequired }`. (An enforceable provider/credential routing key is a good idea too, but it belongs with provider registration; this closes the bypass at the resolver now.)

**3 — unaccepted provider methods in the public seam.** Agreed — removed `AuthMethod::Password` and `Credential::Password`; the seam is now scoped to the accepted #7141 set (OIDC, ssh-key, peercred, native). I also removed `AuthMethod::Peer` — it belongs to the A2A effort, not #7141. Each re-enters via its own scoped change once accepted (`Password` gated on the #7141 child-issue decision; `Peer` owned by the A2A work), and `#[non_exhaustive]` means adding them then isn't a breaking change.

Pushed the update; CI re-running.

Audacity88 (changes requested):
@Nillth thanks for the quick update. I re-reviewed current head f5bf29f against your reply, #7141, the prior request-changes review, the live PR body, the current diff/source, and the green visible GitHub checks. I did not run local Cargo for this review-only pass.

### ✅ Resolved — `Principal` is no longer presented as frozen

The code-side shape now resolves my first blocker. `Principal` and `AuthMethod` are `#[non_exhaustive]`, the docs now say this is an extensible seam rather than a frozen one, and `Principal::new` plus the builder setters give external provider code a construction path that can survive later additive grant fields. The docs also now distinguish identity-source claims from resolved ZeroClaw grants, which is the important #7141 boundary.

### ✅ Resolved — specific provider denies are authoritative

The `ProviderRegistry::resolve` change also resolves the bypass concern from my prior review. A provider that accepts a credential and returns a specific deny reason now short-circuits the registry, while only generic `BadCredential` falls through. The new `specific_deny_is_not_bypassed_by_a_later_provider` regression covers the exact MfaRequired-before-catch-all-bearer case I was worried about.

### ✅ Resolved — unaccepted provider methods are out of the actual seam

The actual public enum/credential shape is now scoped to the accepted #7141 provider set: OIDC, ssh-key, peercred, and native, plus the unbound/trusted sentinels. `Password` and the A2A `Peer` arm are no longer part of the current `AuthMethod` / `Credential` contract, which matches the issue thread: password can come back through its own accepted child scope.

### 🔴 Blocking — clean up the remaining public contract text

The implementation is close now, but a few public-contract words still say the old thing. `crates/zeroclaw-runtime/src/security/auth_provider.rs` still opens by saying providers include `local password` and emit a `Principal` with `identity + resolved grants`, and the `AuthProvider::name` doc still uses `"password"` as an example. The PR body also still says this PR lands the #7141 DoD item "`Principal` carries identity + resolved grants" even though the later scope text correctly says the resolved-grant fields land additively with the IamPolicy-wiring step.

Please make those source docs and the PR body say the same contract the code now implements: this PR lands the accepted provider seam, an extensible `Principal`, and identity/claim inputs; the resolved grant fields and password provider are later scoped changes. Once that public text is aligned, I expect my prior three blockers to be clear.

Nillth:
@Audacity88 thanks - the remaining public-contract text is now aligned (`ef8457be63`):

- `auth_provider.rs` no longer lists `local password` in the credential-kind set, and the `AuthProvider::name` doc example drops `"password"` for accepted-provider names.
- The module doc now says the seam emits a `Principal` carrying identity / claim inputs, with the resolved ZeroClaw grants added additively in the later IamPolicy-wiring step (not this slice).
- The PR body's DoD line is updated to match: this PR lands the accepted provider seam + extensible `Principal` + identity inputs; the resolved-grant fields (allowed-agents / config-write / admin) land additively later.

That should close the last item - mind taking another look when you have a moment?

Audacity88 (approved):
Context: I re-reviewed current head `ef8457b` against the two prior request-changes reviews, #7141, the current PR body, the latest diff/source, and the green visible GitHub checks. I did not run local Cargo for this review-only pass.

@Nillth thanks for the follow-up.

### ✅ Resolved — public contract text now matches the implemented seam

The remaining blocker from my last review is clear. `auth_provider.rs` now scopes the accepted credential examples to OIDC/native bearer, SSH signature, peercred, and native pairing, no longer lists local password as part of this slice, and describes `Principal` as carrying identity/claim inputs with resolved ZeroClaw grants added later. The provider-name examples are also back on the accepted provider set, and the PR body now says the resolved-grant fields land in the later IamPolicy-wiring step.

### ✅ Resolved — the earlier contract and resolver blockers stay resolved

The code still resolves the first three blockers from my initial review: `Principal`, `AuthMethod`, and `Credential` are extensible instead of frozen around this first slice; `Principal::new` plus builders gives providers an external construction path; specific provider denies are authoritative and covered by the MFA-before-catch-all regression; and unaccepted password/A2A peer shapes are not part of the current public enum or credential contract.

### 🟡 Warning — keep denial diagnostics extensible before provider wiring grows

One small follow-up I would strongly consider before this seam gets production call sites: `DenyReason` is the public diagnostic enum most likely to need provider-specific growth, but it is not `#[non_exhaustive]`. The accepted providers can map to the current reasons today, so I am not blocking on this, but OIDC/ssh-key/native/password child slices can plausibly want account-locked, provider-unavailable, token-revoked, or rate-limited style diagnostics. Making `DenyReason` non-exhaustive now would match the extensible direction of the seam and avoid freezing downstream matches around this first diagnostic set.

### 🟢 What looks good — default-deny behavior remains inert and tested

The runtime side still has no production call sites in this slice, the empty registry denies by default, credential debug output redacts secret material, and the resolver test covers the bypass case that mattered for the security boundary. That is the right shape for a foundational auth seam: behavior-neutral until wired, but fail-closed when it is wired.

Approving to clear my previous block at `ef8457b`. This is an approval of the current source/API slice, not a merge-readiness claim over future master movement or any remaining requested reviewer path.

PR #8067, opened 2026-06-20
feat(observability): categorize and verb-tag agent-loop log events

Nillth:
## Summary

- **Base branch:** `master`
- **What changed and why:** The gateway's `?category=` and action-based log filters were effectively blind to agent-loop events: nearly every `record!` in the loop + turn engine emitted `Action::Note` with no `EventCategory`. This stamps an explicit category and a real action verb on every in-scope site so the durable trace is sliceable by subsystem and action.
  - 92 in-scope `record!` sites (agent loop + turn engine) now carry an explicit `EventCategory` (88 newly categorized; 4 in `tool_execution.rs` were already correct).
  - `Action::Note` is replaced by a discriminating verb where meaningful (Load/Register/Migrate/Retry/Reject/Skip/Save/Fail/Timeout/Complete/Cancel/Approve/Delete); ~9 genuinely-diagnostic trace sites intentionally stay `Note`.
  - Category rule: **Provider** = the LLM round-trip + its cost/budget/stream envelope; **Tool** = tool-call lifecycle + result collection; **Memory** = memory ops; **Agent** = loop orchestration (setup, model-switch decision, context compression, history trims, thinking).
  - **Companion fix** (`crates/zeroclaw-api/src/attribution.rs`): `Role::default_category()` returned `"model_provider"`/`"tts_provider"`/etc. and `"sop"` — none of which `EventCategory::parse` can match, so provider/SOP attribution spans silently fell to `Internal` (hidden by `hide_internal`). The Provider arms now return `"provider"` and Sop returns `"system"`. This rescues `llm_request` (which previously resolved to hidden Internal) and is repo-wide for all provider spans. `composite_prefix()` and on-disk attribution field names are unchanged.
- **Scope boundary:** Only the agent loop + turn engine are in scope (92 sites). The other ~2,519 workspace `record!` sites are a deferred follow-up; many already inherit a correct category from their entry-point attribution span (and the companion fix improves that inheritance). No new `Action`/`EventCategory` enum variants. `vision_route` maps to Provider (there is no Vision category).
- **Blast radius:** `zeroclaw-runtime` (the edited sites) and `zeroclaw-api` (the global `default_category()` behavior — every provider attribution span across the workspace now resolves to the `provider` category instead of hidden Internal). Behavior change is intended and observable: `llm_request` rows become visible under `?category=provider`.
- **Linked issue(s):** none (internal observability; part of the observability-gap effort).

```
$ cargo +1.93.0 fmt --all -- --check
clean (exit 0)

$ cargo +1.93.0 clippy -p zeroclaw-runtime -p zeroclaw-api --all-targets -- -D warnings
Finished (exit 0, no warnings)   # a typo'd Action/EventCategory variant would fail to compile here

$ cargo +1.93.0 test -p zeroclaw-runtime turn::      -> 16 passed; 0 failed
$ cargo +1.93.0 test -p zeroclaw-runtime agent::     -> 569 passed; 0 failed
$ cargo +1.93.0 test -p zeroclaw-api                 -> 53 passed; 0 failed
$ cargo +1.93.0 test -p zeroclaw-log                 -> 39 passed; 0 failed
```

- **Beyond CI — what I manually verified:** Confirmed `EventCategory::parse("provider")` is `Some` while `parse("model_provider")` is `None` (the bug the companion fixes), via a temporary test that was reverted. Spot-checked the highest-value assignments: `llm_request`/`llm_response` -> Provider, `history_window` trims -> Delete+Agent, tool lifecycle -> Tool. ASCII-only added lines; no `#[allow(dead_code)]`.
- **Not verified here (environmental):** `cargo build --workspace --all-targets` fails only in `robot-kit`/`aardvark-sys` (missing GTK system libs `gdk-3.0`/`gio-2.0` on this box) — no Cargo.toml or those crates touched. The relevant crates (`runtime`, `api`, `gateway`, `channels`, `log`) build clean `--all-targets`.
- **Live leg (run separately):** after a turn, `?category=provider` returns BOTH `llm_request` and `llm_response`; `?category=tool` returns the tool lifecycle; `?category=agent` the loop lifecycle; confirm `hide_internal` no longer swallows `llm_request`.

## Compatibility (required)

- Backward compatible? **Yes** — only populating existing `event.category` / `event.action` fields with existing enum variants; no schema or enum change. One intended behavior change: `llm_request` (and other provider spans) move from hidden `Internal` to visible `Provider`.

Low risk: `git revert <sha>`. Pure field-population + one string-literal companion; no migration, no flags.

### DIFF
--- a/crates/zeroclaw-api/src/attribution.rs
+++ b/crates/zeroclaw-api/src/attribution.rs
@@ -385,13 +385,10 @@ impl Role {
             Self::Channel(_) => "channel",
             Self::Tool(_) => "tool",
             Self::Cron(_) => "cron",
-            Self::Provider(ProviderKind::Model(_)) => "model_provider",
-            Self::Provider(ProviderKind::Tts(_)) => "tts_provider",
-            Self::Provider(ProviderKind::Transcription(_)) => "transcription_provider",
-            Self::Provider(ProviderKind::Tunnel(_)) => "tunnel_provider",
+            Self::Provider(_) => "provider",
             Self::Memory(_) => "memory",
             Self::Session => "session",
-            Self::Sop => "sop",
+            Self::Sop => "system",
             Self::PeerGroup | Self::Skill | Self::Mcp | Self::System => "system",
         }
     }
--- a/crates/zeroclaw-runtime/src/agent/context_compressor.rs
+++ b/crates/zeroclaw-runtime/src/agent/context_compressor.rs
@@ -239,7 +239,8 @@ impl ContextCompressor {
         if chars_saved > 0 {
             ::zeroclaw_log::record!(
                 INFO,
-                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+                ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Delete)
+                    .with_category(::zeroclaw_log::EventCategory::Agent)
                     .with_attrs(::serde_json::json!({"chars_saved": chars_saved})),
                 "Fast-trim saved chars from old tool results"
             );
@@ -296,7 +297,8 @@ impl ContextCompressor {
 
         ::zeroclaw_log::record!(
             INFO,
-            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
+            ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Retry)
+                .with_category(::zeroclaw_log::EventCategory::Agent)
                 .with_attrs(::serde_json::json!({"context_window": self.context_window})),
             "Context limit adjusted, re-compressing"
         );
@@ -379,8 +381,9 @@ impl ContextCompressor {
             Ok(Err(e)) => {
                 ::zeroclaw_log::record!(
                     WARN,
-                    ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Note)
-                        .with_outcome(::zeroclaw_log::EventOutcome::Unknown)
+                    ::zeroclaw_log::Event::new(module_path!(), ::zeroclaw_log::Action::Fail)
+                        .with_category(::zeroclaw_log::EventCategory::Agent)
+                        .with_outcome(::zeroclaw_log::EventOutcome::Failure)
                         .with_attrs(::serde_json::json!({"error": format!("{}", e)})),
                     "Summarization LLM call failed, using transcript truncation"
                 );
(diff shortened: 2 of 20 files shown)


Audacity88:
Thanks for the careful taxonomy pass here. I reviewed the changed attribution/category sites and the `Role::default_category()` mapping, and I don't see a code-path blocker in the diff itself.

One thing still needs clearing before this can land: the required `Test` job is red. The failing test is `zeroclaw-runtime agent::agent::tests::turn_cache_hit_emits_agent_end_with_none_tokens`, with `left: "uncached answer"` / `right: "cached answer"` at `crates/zeroclaw-runtime/src/agent/agent.rs:6503`.

That looks orthogonal to this PR on first pass: this diff does not touch `agent.rs` or the response-cache code, and current `master` is green at the PR's base SHA. My read is that the gate needs a rerun or a small follow-up if that timestamp-sensitive cache test keeps reproducing, rather than a change to the event-category work itself.

singlerider (approved):
Pulled locally against `upstream/master`, read the full diff, traced the category-resolution path end to end, verified the on-disk-format claim, and ran the relevant test suites. This is a precise, well-diagnosed change. Approving.

### 🟢 Praise: the companion attribution fix is correctly diagnosed and minimal

The root-cause claim checks out exactly. `EventCategory` is an `EnumString` with snake_case variants (`agent`, `channel`, `cron`, `memory`, `tool`, `provider`, `session`, `system`, `internal`). The old `default_category()` returned `model_provider` / `tts_provider` / `transcription_provider` / `tunnel_provider` / `sop`, none of which `EventCategory::parse` can match. At `layer.rs:72` that parse returns `None`, no `SpanCategory` is inserted (the `if let Some` guard at line 76), and resolution falls to `.unwrap_or(EventCategory::Internal)` at line 145, which `hide_internal` then suppresses. Collapsing the provider arms to `provider` and `Sop` to `system` rescues those spans. Confirmed.

### 🟢 Praise: on-disk format is genuinely unchanged

I verified the claim rather than taking it on faith. `default_category()` is consumed only at `macro.rs:61` (the transient `zc_default_category` span field) and parsed in `layer.rs`. `composite_prefix()` still returns the distinct `model_provider` / `tts_provider` / `transcription_provider` / `tunnel_provider` strings, so the composite attribution prefix written to disk does not change. The two paths are correctly decoupled.

### 🟢 Praise: every categorized site stays inside its attribution span

All 92 in-scope sites keep the `record!` in place; the change is purely additive metadata (`Action::Note` to a discriminating verb plus `.with_category(...)`). No `record!` was relocated outside a span and no bare un-attributed emission was introduced. The verb choices are sensible and the category rule (Provider = LLM round-trip envelope, Tool = tool lifecycle, Memory = memory ops, Agent = loop orchestration) is applied consistently. `vision_route` to Provider is the right call given there is no Vision category.

### Verification

Built `zeroclaw-runtime` + `zeroclaw-api` + `zeroclaw-log` clean. Test results:

```
zeroclaw-api:  53 passed; 0 failed
zeroclaw-log:  39 passed; 0 failed
agent::agent::tests::turn_cache_hit_emits_agent_end_with_none_tokens ... ok
```

On the red `Test` gate @Audacity88 flagged: that test passes locally here, and this diff touches neither `agent.rs` nor the response-cache path. It is the timestamp-sensitive cache test, orthogonal to this work. A gate rerun should clear it; if it keeps reproducing it is a separate follow-up, not a change to this PR.

Approving. Please get the `Test` gate green (rerun) before merge.

PR #8082, opened 2026-06-21
feat(skills): dashboard write-guard, skipped-audit, and shadowed_by (#7963 follow-ups)

Nillth:
## Summary

Implements the three follow-ups deferred from #7963 (the dashboard "effective skills" view). All three build on `SkillsService::resolve_effective_skills` and stay within its audited-resolver guarantee. Delivered as three commits in one PR.

### 1. Write-guard — server-side `editable` enforcement
The skill write endpoints (`write_skill` / `remove_skill`) returned a misleading `NotFound` when a caller targeted a non-bundle (workspace / open-skills / plugin) skill. Added `ServiceError::NotEditable { name, origin }` and a filesystem-truth guard `ensure_editable` (a skill is writable iff its resolved bundle directory holds a manifest — i.e. it is a real bundle skill). Wired in after the existing existence check, so a truly-absent skill still yields `NotFound` while an exists-but-non-bundle target now yields `NotEditable` -> **403 Forbidden**. `scaffold_skill` (create) is unguarded — creating in a configured bundle dir is inherently a bundle write. The frontend already gated writes on `editable`; this closes the gap server-side.

### 2. Skipped-audit — surface what the resolver dropped
The four source loaders silently `continue` past any skill that fails the security audit, so the dashboard couldn't tell **"no skills configured"** from **"all failed audit."** Now each uncached loader returns `(Vec<Skill>, Vec<DroppedSkill>)`, threaded through every aggregator up to a new `EffectiveSkillSet { skills, dropped }` returned by `resolve_effective_skills`. A `DroppedSkill` carries `{ name, origin_hint, reason, location }` with `reason` in `{AuditFindings, AuditError, ManifestParseError}` — the same detail the loader already logs.

**Cache correctness:** the skill cache stored only `Vec<Skill>`, so drops would vanish on a cache hit. The cached unit is widened to `LoadOutput { skills, dropped }` (one walk = one audit = both together) rather than recomputing — preserving the cache's whole purpose (never re-audit) *and* audit parity.

The dashboard renders an amber banner listing each dropped skill as `name (origin) — reason`, and the empty-state "no skills" text now shows only when both lists are empty.

### 3. Shadowed_by — show precedence collisions
When a bundle skill loses its name to an earlier-precedence same-name skill (the one real dedup site: workspace/open-skills/plugin/earlier-bundle win, `mod.rs`), the winner now records what it shadowed. `EffectiveSkill` gains `shadowed: Vec<ShadowedSkill>`; the card shows a small amber "shadows …" badge. Precedence is unchanged — this only *reports* the existing winner/loser, so operators finally see why a bundle edit isn't taking effect.

**Wire:** `AgentSkillEntry` gains `shadowed` and `AgentSkillsResult` gains `dropped`, both additive (`#[serde(default, skip_serializing_if = ...)]`); two new `rpc_type!` structs `ShadowedSkillEntry` / `DroppedSkillEntry`. No endpoint signature change.

**Caller blast radius (handled):** the loader signature changes ripple widely, so the public `load_skills_*` entry points keep returning `Vec<Skill>` via thin wrappers / dedicated `_audited` variants — non-dashboard callers (agent boot, message processing, orchestrator, delegate, channels) compile unchanged; only the dashboard path carries drops + shadows. No `#[allow(dead_code)]`/suppression — the new data reaches a real consumer.

Re-validated locally on the **refreshed head `f9af90c5`** after merging current `upstream/master` in (the branch is now 0 behind master; CI builds the PR merged with base). CI-pinned toolchain **1.93.0**:
```
cargo +1.93.0 fmt --all -- --check                                   # PASS
cargo +1.93.0 clippy --all-targets +1.93.0 -- -D warnings            # PASS (the CI Lint gate; compiles every target incl. the plugins-wasm loader + root integration tests, so the loader-signature change is verified workspace-wide — no missed callers)
cargo +1.93.0 test  (workspace)                                      # PASS
(cd web && cargo web check)                                          # PASS (gen-api + tsc -b + vite build)
```
The previously published green CI run for the prior head `23541b1a` is https://github.com/zeroclaw-labs/zeroclaw/actions/runs/27893540355 (all checks success). The branch has since been refreshed onto current master; **the authoritative gate is the CI run that will execute on the refreshed head `f9af90c5` after this push — it must be green before merge.**

New tests: write-guard (NotEditable on non-bundle write/remove, Ok on bundle); skipped-audit (resolve reports dropped; empty-vs-all-dropped; loader returns drops; **dropped records survive a cache hit**); shadowed_by (winner records the shadowed bundle skill); gateway (NotEditable->403, shadowed/dropped wire mapping).

**Docs:** no separate reference doc — the change adds two additive operator-dashboard wire fields (`dropped`, `shadowed`) plus a 403 tightening on an already-failing write path; the only consumer is the self-describing Skills page UI shipped in this same PR, and the repo keeps no hand-maintained reference page for this internal gateway endpoint.

- **Audit parity preserved** — everything still flows through the audited `load_skills_for_agent_from_config`; skipped-audit and shadowed_by only *observe and report* what the audited resolver already computed, never widening to unaudited frontmatter.
- The write-guard **tightens** the surface (a previously-misleading-but-already-failing write is now an explicit 403). No new permission, capability, or network call.
- The skipped-skill reason strings (e.g. a flagged script path) are surfaced on the single-tenant operator dashboard only — no cross-tenant exposure.

## Compatibility (required)

Backward compatible — all new wire fields are additive with serde defaults; old clients ignore them and new clients tolerate old servers. No config/schema/CLI change. The `resolve_effective_skills` return type changed (`Vec<EffectiveSkill>` -> `EffectiveSkillSet`), an internal Rust API; all in-tree callers updated.

## Rollback (required for medium/high-risk PRs)

- **Mechanism:** `git revert` the three commits — additive types/fields + a guard; no state, schema, or config migration to unwind.
- **Feature flags / config toggles:** None. There is no flag gating these features; `git revert` is the rollback. (The pre-existing `ZEROCLAW_SKILLS_CACHE` off-ramp documented in `cache.rs` is unchanged by this PR.)
- **Observable failure symptoms to watch after deploy:**
  - *Write-guard:* a write/remove targeting a non-bundle skill returns **403** (was 404). A caller that pattern-matches on 404 would now see 403 — visible as a client-side "not found" path no longer triggering.
  - *Skipped-audit:* if a cache regression drops a drop-record, the amber "skill(s) skipped" banner disappears on the next cache hit and an operator sees an empty skills list with no explanation.
  - *Shadowed_by:* if shadow recording regresses, the amber "shadows …" badge disappears from affected skill cards and operators lose visibility into which source is overriding a bundle skill.


WareWolf-MoonWall (commented):
# Review — PR #8082 `feat(skills): dashboard write-guard, skipped-audit, and shadowed_by`

**Author:** Nillth (Marc Collins) | **Head SHA:** `23541b1aeeec4211b501964a470891c606d31f20` | **Verdict:** COMMENT

Solid multi-feature PR that closes the three deferred #7963 follow-ups. The write-guard closes a real security gap (403 vs misleading 404), the skipped-audit threading is architecturally clean (audit parity preserved; cache widening is the right model), and the shadowed_by reporting is resolved at call time with no duplicate state. Test coverage is thorough across all three features. Two template completeness gaps need fixes before merge; one code suggestion worth considering.

---

### 🟡 Rollback section missing required risk:high sub-bullets

The Rollback section reads only:

> `git revert` the three commits — additive types/fields + a guard; no state or schema migration.

For `risk: high` PRs both of the following sub-bullets are required:

- **Feature flags or config toggles** — should state "None" explicitly (the git revert is the rollback mechanism, but a reviewer shouldn't have to infer "no feature flags").
- **Observable failure symptoms** — completely absent. Examples of what should appear here:
  - Write-guard: write/remove targeting a non-bundle skill returns 403 (was 404) — callers that pattern-match on 404 would break silently.
  - Skipped-audit: if the cache misses a drop (regression), the amber banner disappears on next cache hit; operator sees empty skills list with no explanation.
  - Shadowed_by: if shadow recording regresses, the amber badge disappears from affected skill cards and operators lose visibility into overriding skills.

Please add both sub-bullets before merge. Compare to #8093's rollback section as the expected format.

---

### 🟡 Validation Evidence — no CI run SHA cited

The evidence block lists local cargo/npm runs with claimed pass results, but doesn't include the CI run ID or SHA. For a `risk: high` PR the protocol requires a literal run SHA or CI run URL. The CI checks do show all green on this PR (confirmed from `gh pr checks`), so the code is validated — but the PR description should link to the passing run for traceability.

---

### ✅ Write-guard — correct and SSOT-compliant

`ensure_editable` derives editability from the filesystem truth (`has_manifest(&dir)`) at call time — no cached copy of "is this a bundle skill" stored on any struct. The guard fires after the `NotFound` check, so a truly-absent skill still gets 404; only a present-but-non-bundle target gets 403. `scaffold_skill` (create) is intentionally unguarded — correct, since creating in a configured bundle directory is inherently a bundle write. 🟢 The ordering `NotFound → NotEditable → write` is sound.

---

### 🔵 `ensure_editable` calls `skill_directory` twice

`write_skill` resolves `dir` via `self.skill_directory(target)?`, checks `!dir.exists()`, then calls `self.ensure_editable(target)?` which calls `self.skill_directory(target)?` a second time. The second call is redundant. Consider:

```rust
fn ensure_editable(&self, dir: &Path) -> Result<(), ServiceError> {
    // receives the already-resolved dir from the caller
}
```

This is display-only redundancy (single-tenant, filesystem is trusted; worst case is a benign spurious error in a race), but eliminating the double call is cleaner and avoids the implicit assumption that the two calls return the same path.

---

### ✅ Skipped-audit — cache widening is the right model

`LoadOutput { skills, dropped }` cached as a unit means one walk = one audit = both pieces stored together. A side-channel that recomputed drops on every cache hit would re-walk and re-audit, defeating both the cache and the audit-parity guarantee. The approach correctly preserves the cache's purpose. `dropped` survives a cache hit — confirmed by `dropped_records_survive_cache_hit` test. 🟢

---

### ✅ Shadowed_by — resolved at call time, no duplicate state

Shadow records are built in `load_skills_for_agent_audited` from the live dedup walk (`seen: HashMap<String, &'static str>`) and consumed by `resolve_effective_skills` via `shadow_index.remove(&s.name)`. The relationship is not stored anywhere between calls — it is recomputed on every `resolve_effective_skills` invocation from canonical skill-loader state. SSOT compliant.

---

### 🔵 `origin_hint_of` heuristic — add a comment explaining tolerance

```rust
fn origin_hint_of(skill: &Skill) -> &'static str {
    if skill.tags.iter().any(|t| t == "open-skills") { "open-skills" }
    else if skill.name.starts_with("plugin:") || skill.tags.iter().any(|t| t.starts_with("plugin:")) { "plugin" }
    else { "workspace" }
}
```

This tag-based heuristic correctly handles the common cases but can misclassify a workspace skill whose `tags` happens to contain `"open-skills"`. A brief comment noting this is display-only context (shadow badge attribution) and that approximation is acceptable — it never affects which skills load or their precedence — would help a future reader avoid treating it as authoritative origin resolution.

---

### ✅ i18n — all new UI strings properly keyed

`SkillCard.tsx` uses `t('skills.shadows')` and `Skills.tsx` uses `t('skills.skipped_count')`. Both keys are added to `i18n.ts`. Dynamic server-supplied strings (`d.reason`, `s.origin`) are data, not bare string literals, and rendered in an operator-only debug context — no Fluent violation. ✅

---

### ✅ No `unwrap()` / `expect()` in new production paths

New production code: `dir_stem` uses `.unwrap_or_default()` ✅, `shadow_index.remove` returns `Option` handled with `.unwrap_or_default()` ✅, `cache().write().unwrap_or_else(|e| e.into_inner())` is pre-existing poison-handling ✅. `delegate.rs` uses `.0` tuple extraction ✅. No bare panicking calls in new production paths.

---

### ✅ Audit parity preserved

`DroppedSkill` data flows from loaders → `LoadOutput` cache → `load_skills_for_agent_audited` → `resolve_effective_skills` → gateway → wire. At no point does a dropped skill re-enter the audited pool. `skipped-audit` and `shadowed_by` surfaces only observe and report what the audited resolver already computed. No audit bypass.

---

### ✅ Wire compatibility — additive, serde-gated

`AgentSkillEntry.shadowed` and `AgentSkillsResult.dropped` are both `#[serde(default, skip_serializing_if = ...)]`. Old clients ignore new fields; new clients tolerate old servers that omit them. No endpoint signature change. `resolve_effective_skills` return type change is an internal Rust API — all in-tree callers updated (compiler confirms via the `--all-targets --features ci-all` Clippy run).

---

### ✅ Test coverage

| Feature | Tests |
|---|---|
| Write-guard | `write_skill_rejects_non_bundle_skill`, `write_skill_allows_existing_bundle_skill`, `remove_skill_rejects_non_bundle_skill` |
| Skipped-audit | `resolve_effective_skills_reports_dropped`, `resolve_effective_skills_empty_vs_all_dropped`, `load_skills_for_agent_from_config_audited_returns_dropped`, `dropped_records_survive_cache_hit` |
| Shadowed_by | `resolve_effective_skills_records_shadow` |
| Gateway | `not_editable_maps_to_forbidden`, `agent_skill_entry_maps_shadowed`, `dropped_skill_entry_maps_each_reason_kind` |

🟢 Cache-hit survival test for dropped records is exactly the right regression anchor.

---

All four commits authored and committed by Nillth (marc@nnet.tech), SSH-signed. No `Co-authored-by:` AI bot trailers.

---

### ✅ Linked issues

PR correctly references #7963 as the parent and closes its three deferred follow-up items. `feat/dashboard-skills-followups` branch name is consistent with scope.

WareWolf-MoonWall:
@singlerider — milestone alignment needed: this feat PR is not yet merge-ready (template gaps remain). Please advise on placement once cleared.

Audacity88 (changes requested):
Thanks @Nillth . I read this as the right follow-up shape for #7963: the cache stores loaded and dropped records together, dropped skills do not re-enter the effective set, shadow reporting is observational, and the write guard tightens the bundle write surface instead of expanding it.

I checked the PR body, live PR metadata, live checks, the existing review from @WareWolf-MoonWall, the full diff, and the changed runtime/gateway/web source.

I do need changes before I can approve it, though.

### 🔴 Blocking — Add live or near-live dashboard/API smoke evidence

The unit coverage is strong, and I agree with @WareWolf-MoonWall that the main code shape is sound. But this PR changes the real operator path: `/api/agents/{alias}/skills`, the Skills dashboard dropped-skill banner, the shadow badge, and the 403 mapping for non-editable writes.

Please add concise smoke evidence from the actual gateway/dashboard boundary. A good packet would be:

- an `/api/agents/{alias}/skills` response showing `dropped` and `shadowed`
- a dashboard screenshot or equivalent live readback showing the amber dropped-skill banner and shadow badge
- a `PUT` or `DELETE` against a non-editable skill showing `403 Forbidden`

That proves the feature works where operators will see it, not only in unit tests and TypeScript build output.

### 🟡 Warning — PR body still needs the `risk: high` evidence details

I also agree with @WareWolf-MoonWall on the remaining PR-body gaps. The rollback section should explicitly state whether feature flags/config toggles exist, and should list observable failure symptoms. The validation section should link the passing CI run or include the run SHA for the refreshed head.

I am not treating that as a code blocker if maintainers want to patch the body mechanically, but the public record should be complete before merge.

### 🟡 Warning — Refresh before merge

Current CI is green for head `23541b1aeeec4211b501964a470891c606d31f20`. Since this is `risk: high` and current `master` has moved with same-file overlap in `crates/zeroclaw-runtime/src/skills/mod.rs` and `crates/zeroclaw-runtime/src/tools/delegate.rs`, the branch should be refreshed and CI should be green on the refreshed head before anyone treats it as merge-ready.

That is not a direction blocker. It is just the normal freshness step before merge.

### ✅ Looks good — Resolver and cache design

The important internals look well aimed:

- `LoadOutput { skills, dropped }` is the right cache unit, so dropped records survive cache hits without a second audit.
- Dropped skills stay as reporting data and never become loaded skills.
- `shadowed` is computed from the same effective-skill walk and does not introduce a second precedence source of truth.
- The new wire fields are additive and serde-defaulted.

Once the live smoke evidence is present, I expect this to be close.

singlerider (commented):
First review from me. Verified the security-relevant paths at head `23541b1a` (CI green, MERGEABLE). Read the source; did not run local Cargo. No code blockers from me; I concur with @Audacity88's smoke-evidence request before this lands.

### 🟢 The write-guard is a genuine tightening, not an expansion

`SkillService::ensure_editable` (`service.rs:283`) is called at the head of both `write_skill` (`:301`) and the delete path (`:326`) and rejects any non-bundle skill with `ServiceError::NotEditable` before any filesystem mutation. Only bundle skills are writable; workspace/plugin/builtin are read-only. Both arms are pinned: `write_skill_rejects_non_bundle_skill` and `write_skill_allows_existing_bundle_skill`. This narrows the write surface, which is the correct direction for a risk:high change.

### 🟢 The 403 mapping is typed and exhaustive

`service_error_response` (`api_skills.rs:320`) matches on the `ServiceError` enum and maps `NotEditable -> 403`, distinct from `NotFound -> 404` and `Ref/Bundle/Scaffold -> 400`. No string matching; a new variant forces a compile error rather than silently falling through. `not_editable_maps_to_forbidden` pins it.

### 🟢 dropped/shadowed reporting keeps a single source of truth

`LoadOutput { skills, dropped }` is the cache unit so dropped records survive cache hits without a second audit, dropped skills never re-enter the effective set, and `shadowed` is computed from the same effective-skill walk rather than a second precedence source. New wire fields are additive and serde-defaulted. This matches what @Audacity88 and @WareWolf-MoonWall already verified.

### 🔴 Concur: add the live dashboard/API smoke evidence before merge

I am not raising a code defect, but I agree with @Audacity88 that a risk:high change to the real operator path (`/api/agents/{alias}/skills`, the dropped-skill banner, the shadow badge, the 403-on-non-editable-write) needs concrete boundary evidence on the public record, not only unit/integration tests and a TS build. The packet @Audacity88 described is the right bar:

- an `/api/agents/{alias}/skills` response showing `dropped` and `shadowed`
- a dashboard readback of the amber dropped-skill banner and shadow badge
- a `PUT`/`DELETE` against a non-editable skill returning `403`

Also fold in the PR-body gaps @Audacity88 and @WareWolf-MoonWall flagged: the risk:high rollback section (feature flags/toggles + observable failure symptoms) and a linked CI run SHA for the refreshed head.

Code shape is sound from my read. Add the smoke packet and complete the body, and this is close. Leaving @Audacity88's changes-requested in force.

Nillth:
Thanks @Audacity88, @singlerider, @WareWolf-MoonWall. Two updates: the branch is refreshed onto current `master` (was conflicting, now mergeable), and here is the live gateway smoke packet you asked for. All of this was captured against the refreshed head `f9af90c5` running the actual daemon (`zeroclaw gateway start`), not unit tests.

### Setup
A throwaway agent `default` with bundle `core`:
- workspace `foo` + bundle `core/foo` (same name -> workspace wins, bundle is shadowed)
- a workspace skill `broken` with an invalid `SKILL.toml` (-> dropped, manifest parse error)
- a bundle dir `core/noman` that exists but has no manifest (-> not a real bundle skill; write/delete target)
- a genuine bundle skill `core/realone` (editable)

### 1. `GET /api/agents/default/skills` shows `dropped` and `shadowed`

```json
{
  "agent": "default",
  "skills": [
    {
      "name": "foo",
      "description": "workspace copy of foo (higher precedence; shadows the bundle foo)",
      "origin": "workspace",
      "editable": false,
      "shadowed": [ { "name": "foo", "origin": "bundle" } ]
    },
    {
      "name": "realone",
      "origin": "bundle",
      "bundle": "core",
      "editable": true
    }
  ],
  "dropped": [
    {
      "name": "broken",
      "origin": "workspace",
      "reason_kind": "manifest_parse_error",
      "reason": "TOML parse error at line 4, column 1 ... unknown field `bogus`, expected one of `name`, `description`, ..."
    },
    {
      "name": "noman",
      "origin": "bundle",
      "reason_kind": "audit_findings",
      "reason": "Skill root must include SKILL.md (canonical) or one of [\"SKILL.toml\", \"manifest.toml\"] (deprecated) for deterministic auditing."
    }
  ]
}
```

The winning workspace `foo` carries `shadowed: [{name: foo, origin: bundle}]`, and `dropped` carries both an audit-parse failure and an audit-findings failure (two of the three `reason_kind` arms exercised end to end).

### 2. Write/delete against a non-editable skill -> `403 Forbidden`

```
$ curl -X PUT  .../api/skills/bundles/core/skills/noman  -d '{...}'
HTTP 403
{"error":"skill 'core/noman' is not editable: it is a non-bundle skill, only bundle skills are writable"}

$ curl -X DELETE .../api/skills/bundles/core/skills/noman?purge=true
HTTP 403
{"error":"skill 'core/noman' is not editable: it is a non-bundle skill, only bundle skills are writable"}
```

And the guard ordering holds, with `NotFound` still distinct from `NotEditable`:

```
$ curl -X PUT  .../api/skills/bundles/core/skills/ghost  -d '{...}'   # truly absent
HTTP 404
{"error":"skill 'core/ghost' is not present in any configured bundle"}

$ curl -X PUT  .../api/skills/bundles/core/skills/realone -d '{...}'  # real bundle skill
HTTP 204   (write landed; SKILL.md updated on disk)
```

The rejected `DELETE` destroyed nothing: `core/noman/` is still on disk afterward.

### 3. Dashboard banner + badge

The two arrays above are exactly what the Skills page renders, so the JSON is the live readback for the operator surface:
- `dropped[]` drives the amber "N skill(s) skipped (failed security audit)" banner, listing each as `name (origin) - reason` (`web/src/pages/Skills.tsx`, the `dropped.length > 0` block).
- `shadowed[]` drives the amber "shadows bundle" badge on the `foo` card (`web/src/components/SkillCard.tsx`, the `skill.shadowed?.length > 0` block).
- the empty-state text now only shows when both `skills` and `dropped` are empty, so "all skipped" no longer reads as "none configured".

I can attach a rendered screenshot if you want the pixels as well, but the boundary behavior is all here. With this plus the refreshed-onto-master green CI, I think this is ready for another look.

singlerider (commented):
@Nillth re-reviewed at head `f9af90c5` against @Audacity88's active changes-requested. His sole blocker is satisfied. Not approving over his active CHANGES_REQUESTED; status pass for him to convert.

### RESOLVED 🟢 - the live gateway/dashboard smoke evidence is now on the record

@Audacity88 and I both gated on concrete boundary evidence for this risk:high operator-path change. Your follow-up comment delivers exactly the packet we asked for, captured against the refreshed head running the actual daemon (`zeroclaw gateway start`), not unit tests:

- `GET /api/agents/default/skills` showing `shadowed: [{name: foo, origin: bundle}]` on the winning workspace `foo`, and `dropped` carrying both an audit-parse failure and an audit-findings failure (two of the three `reason_kind` arms exercised end to end)
- `PUT`/`DELETE` against a non-editable skill returning `403 Forbidden`

That proves the feature works where operators see it. The branch is also MERGEABLE again after the refresh onto master.

### 🟢 Code shape confirmed green (unchanged from my prior read)

The write-guard (`ensure_editable` rejecting non-bundle skills before any fs mutation), the typed exhaustive `ServiceError -> 403/404/400` mapping, and the single-source `LoadOutput { skills, dropped }` cache unit with `shadowed` computed from the same effective-skill walk all reviewed clean. 229 skills tests pass locally (single-threaded).

### Note - flaky test under parallel run, not a blocker

`skills::cache::tests::dropped_records_survive_cache_hit` fails under the default parallel run on this branch but passes in isolation and single-threaded (`--test-threads=1`: 229/229). It is a test-isolation flake (shared cache global state across parallel skills tests), not a defect in this PR. Worth a follow-up to scope that test's cache state, but it does not block.

No superseding fix: this is the single PR implementing the #7963 dashboard follow-ups.

@Audacity88 - your smoke-evidence blocker is satisfied by @Nillth's live packet, and the branch is mergeable. Yours to convert.

Nillth:
Thanks @Audacity88, @singlerider, @WareWolf-MoonWall. Two updates: the branch is refreshed onto current `master` (was conflicting, now mergeable), and here is the live gateway smoke packet you asked for. All of this was captured against the refreshed head `f9af90c5` running the actual daemon (`zeroclaw gateway start`), not unit tests.

### Setup
A throwaway agent `default` with bundle `core`:
- workspace `foo` + bundle `core/foo` (same name -> workspace wins, bundle is shadowed)
- a workspace skill `broken` with an invalid `SKILL.toml` (-> dropped, manifest parse error)
- a bundle dir `core/noman` that exists but has no manifest (-> not a real bundle skill; write/delete target)
- a genuine bundle skill `core/realone` (editable)

### 1. `GET /api/agents/default/skills` shows `dropped` and `shadowed`

```json
{
  "agent": "default",
  "skills": [
    {
      "name": "foo",
      "description": "workspace copy of foo (higher precedence; shadows the bundle foo)",
      "origin": "workspace",
      "editable": false,
      "shadowed": [ { "name": "foo", "origin": "bundle" } ]
    },
    {
      "name": "realone",
      "origin": "bundle",
      "bundle": "core",
      "editable": true
    }
  ],
  "dropped": [
    {
      "name": "broken",
      "origin": "workspace",
      "reason_kind": "manifest_parse_error",
      "reason": "TOML parse error at line 4, column 1 ... unknown field `bogus`, expected one of `name`, `description`, ..."
    },
    {
      "name": "noman",
      "origin": "bundle",
      "reason_kind": "audit_findings",
      "reason": "Skill root must include SKILL.md (canonical) or one of [\"SKILL.toml\", \"manifest.toml\"] (deprecated) for deterministic auditing."
    }
  ]
}
```

The winning workspace `foo` carries `shadowed: [{name: foo, origin: bundle}]`, and `dropped` carries both an audit-parse failure and an audit-findings failure (two of the three `reason_kind` arms exercised end to end).

### 2. Write/delete against a non-editable skill -> `403 Forbidden`

```
$ curl -X PUT  .../api/skills/bundles/core/skills/noman  -d '{...}'
HTTP 403
{"error":"skill 'core/noman' is not editable: it is a non-bundle skill, only bundle skills are writable"}

$ curl -X DELETE .../api/skills/bundles/core/skills/noman?purge=true
HTTP 403
{"error":"skill 'core/noman' is not editable: it is a non-bundle skill, only bundle skills are writable"}
```

And the guard ordering holds, with `NotFound` still distinct from `NotEditable`:

```
$ curl -X PUT  .../api/skills/bundles/core/skills/ghost  -d '{...}'   # truly absent
HTTP 404
{"error":"skill 'core/ghost' is not present in any configured bundle"}

$ curl -X PUT  .../api/skills/bundles/core/skills/realone -d '{...}'  # real bundle skill
HTTP 204   (write landed; SKILL.md updated on disk)
```

The rejected `DELETE` destroyed nothing: `core/noman/` is still on disk afterward.

### 3. Dashboard banner + badge

The two arrays above are exactly what the Skills page renders, so the JSON is the live readback for the operator surface:
- `dropped[]` drives the amber "N skill(s) skipped (failed security audit)" banner, listing each as `name (origin) - reason` (`web/src/pages/Skills.tsx`, the `dropped.length > 0` block).
- `shadowed[]` drives the amber "shadows bundle" badge on the `foo` card (`web/src/components/SkillCard.tsx`, the `skill.shadowed?.length > 0` block).
- the empty-state text now only shows when both `skills` and `dropped` are empty, so "all skipped" no longer reads as "none configured".

I can attach a rendered screenshot if you want the pixels as well, but the boundary behavior is all here. With this plus the refreshed-onto-master green CI, I think this is ready for another look.

---

Since the smoke run, one follow-on commit landed (0057d98956): a behavior-preserving refactor that parameterizes `cached_load` over its backing store (`cached_load_in`) so the skills cache tests run against fresh local stores instead of the process-global cache plus `invalidate()`. That removes a parallel-run flake in the cache hit/miss tests (the same class as the global-state test pollution seen elsewhere); production behavior is unchanged, so the smoke evidence above still holds.

Audacity88 (approved):
@Nillth I re-reviewed the current head (`0057d989`), my prior `CHANGES_REQUESTED` review, @singlerider's follow-up review, the live gateway smoke packet, current CI, and the changed skills/gateway/web source paths. I did not run local Cargo, web build, or a live gateway smoke for this re-review.

### ✅ Resolved — Live gateway evidence now covers the operator path

My prior blocker was about boundary evidence, not the unit-level code shape. That is now satisfied. The live packet was captured against `f9af90c5` running the real gateway, and it covers the three user-visible paths I asked for: `GET /api/agents/default/skills` returns both `dropped` and `shadowed`, non-editable write/delete attempts return `403 Forbidden`, and the dashboard-visible banner/badge data is the same data returned by the API.

That proves the feature at the operator boundary rather than only through unit tests or TypeScript compilation.

### ✅ Resolved — Review follow-ups are reflected in the current source

The earlier review nits are also handled in the current diff. `ensure_editable` now receives the already-resolved directory instead of resolving the skill path twice, `origin_hint_of` has the display-only heuristic comment, and the final `0057d989` delta is a behavior-preserving `cached_load_in(...)` refactor that isolates the cache tests from the process-global cache while keeping production on the same global cache path.

The main resolver shape still looks right: `LoadOutput { skills, dropped }` is the cache unit, dropped skills stay reporting-only, `shadowed` is computed from the same effective-skill walk, and the gateway wire fields are additive and serde-defaulted.

### 🟡 Warning — Refresh before merge

This approval clears my prior review block on the current code and smoke evidence. It is not a merge-readiness call. Please still run the normal freshness check before merge and use checks from the merge result or refreshed head, not this review alone.

Approving.

PR #8086, opened 2026-06-21
feat(web): themed click-to-open config pickers (Select primitive)

Nillth:
_Split 1/3 of #8022. Independent base; #8022c (storage nav) stacks on this._

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds a themed `Select` primitive (`web/src/components/ui/Select.tsx`): options-as-prop, no embedded value list, correct `role="combobox"/listbox/option"` ARIA, keyboard nav with wrap-around + scroll-into-view, outside-click close.
  - Adds an `openOnFocus` affordance to `ComboBox` so the list opens on focus/click (for fields where the options ARE the expected input, e.g. an alias picker).
  - Adopts both in `FieldForm` for enum and alias-ref fields in place of the native `<select>`.
- **Scope boundary:** Display-only. The daemon stays authoritative on save; no schema/`/api/config`/`validate_*` changes.
- **Known follow-up (registry):** the alias-ref picker prefers the daemon-declared `entry.alias_source` and falls back to a `<Type>Ref` type_hint derive table (`ALIAS_REF_TYPE_TO_SOURCE`) only when the daemon omits it. The backend currently stubs `alias_source: None`; removing the fallback is a follow-up once #7855 (`fix/7701-model-route-configurable`) populates `alias_source` backend-side. Tracked, not routed around.
- **Blast radius:** `web/` only (`ui/Select.tsx` new, `ui/ComboBox.tsx`, `ui/index.ts`, `sections/FieldForm.tsx`).

## Validation Evidence

```
$ cargo web check   # gen-api + npx tsc -b
EXIT=0  (clean)
```

- Split from #8022 by file; no behavior change versus that PR's picker work.

image

## Security & Privacy Impact

- New permissions / network / secrets handling? **No.** Display-only UI; server remains source of truth on save.
- PII in diff/tests/fixtures? **No.**

## Compatibility

- Backward compatible? **Yes.** `openOnFocus` defaults to `false` (existing ComboBox callers unchanged). `Select` is additive.

## Rollback (low risk)

`git revert <sha>`. No flags, no migration. Pure additive UI primitives.

### DIFF
--- a/web/src/components/sections/FieldForm.tsx
+++ b/web/src/components/sections/FieldForm.tsx
@@ -39,7 +39,7 @@ import {
 } from "lucide-react";
 import DirectoryPicker from "./DirectoryPicker";
 import ToolPicker from "@/components/ToolPicker";
-import { Badge, Button, ComboBox } from "@/components/ui";
+import { Badge, Button, ComboBox, Select } from "@/components/ui";
 import type { BadgeTone } from "@/components/ui";
 import { t } from "@/lib/i18n";
 import {
@@ -570,6 +570,9 @@ const AGENT_MULTI_ALIAS_FIELDS: Record<string, keyof AgentOptionsResponse> = {
   "skill_bundles": "skill_bundles",
   "knowledge_bundles": "knowledge_bundles",
   "mcp_bundles": "mcp_bundles",
+  // Delegates is a subset of the configured agents — give it the same themed
+  // multi-select (with agent suggestions) as the bundle fields, not free text.
+  delegates: "agents",
 };
 
 // Peer-groups carry the same alias-ref shape as agents do: a single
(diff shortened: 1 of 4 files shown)


singlerider (approved):
First review of this split. Verified at head `97801047` (CI green, MERGEABLE). Read the source; did not run local Cargo/tsc. **All 🟢; approving.** This is the clean split-1/3 of #8022 and it addresses my original registry-shadow blocker in the right shape.

### 🟢 My #8022 blocker #1 is fixed: registry value is now the primary source

`aliasRefSource` (`FieldForm.tsx:649`) now prefers `entry.alias_source` (`:651`) and only falls back to the `ALIAS_REF_TYPE_TO_SOURCE` table when the daemon omits it. That is exactly the AGENTS.md-compliant shape: the surface reads the backend registry value and the TS table is a documented backward-compat shim for builds predating the `alias_source` emission, not the source of truth. It becomes dead code once #7855 populates `alias_source` daemon-side, at which point it should be deleted. Tracked, not routed around.

### 🟢 No stuck-spinner failure mode

`aliasRefResolving` is `aliasSource !== null && aliasRefOptions === null` (actively fetching only). An unresolvable field (table miss, `aliasSource === null`) renders an empty-but-usable `ComboBox` rather than a spinner that never finishes, and free text is accepted verbatim and validated on save, so an existing or not-yet-created alias is never dropped. That closes the degenerate case cleanly.

### 🟢 Select primitive and openOnFocus are correctly built

`Select.tsx` takes `options` as a prop with no embedded value list, correct `role="combobox"/listbox/option"` ARIA, and keyboard nav with wrap-around. `ComboBox.openOnFocus` is opt-in (default false), wired to both `onFocus` and `onClick`. The native `<select>` replacement is a real themed-UX win and the scope is tight (4 files, picker only).

Scope is clean and the registry shadow is reduced to a tracked transitional fallback. Approving.

PR #8087, opened 2026-06-21
feat(web): dashboard component-health "fix in place" modal

Nillth:
_Split 2/3 of #8022. Independent (does not depend on the picker split)._

## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Adds a "Fix" affordance on Component Health error rows (Dashboard Overview) that opens the same `DoctorFixModal` used on the Doctor page, deep-linking to the failing config entity.
  - The error-to-target parser is start-anchored, so prose that merely contains a config-like token (`"failed to load agents.json"`) does not yield a bogus target.
- **Registry fix (was blocker #2 on #8022):** the field's config tab is now resolved from backend `ConfigTab` metadata, not a field-name regex. `OverviewTab` builds a `path -> tab` index from `/api/config/list` (prefix `agents`) and `healthFixTarget` reads `entry.tab`, converting it to the `?tab=` slug with the same derivation as `Config.tsx wireTabSpecs` (`label.toLowerCase().replace(/\s+/g,"-")`). A re-tabbed or newly added field deep-links correctly with no change here; falls back to the agent's default tab when the entry/tab is unknown.
- **Blast radius:** `web/` only (`pages/Dashboard.tsx`, `lib/i18n.ts` for the `dashboard.fix` label).

## Validation Evidence

```
$ cargo web check   # gen-api + npx tsc -b
EXIT=0  (clean)
```

- Manually traced: the `path -> tab` index is best-effort (load failure or empty index just omits `?tab=`); the `agents.<alias>.<field>` start-anchored match is unchanged from #8022 except for the tab source.

Screenshot From 2026-06-21 23-19-30

Screenshot From 2026-06-21 23-19-54

## Security & Privacy Impact

- New permissions / network / secrets handling? **No** new permissions. Adds one read of `/api/config/list` (already used elsewhere in the app) to resolve tab metadata.
- PII in diff/tests/fixtures? **No.**

## Compatibility

- Backward compatible? **Yes.** Additive UI; the modal reuses the existing Doctor fix flow.

## Rollback (low risk)

`git revert <sha>`. No flags, no migration. Additive dashboard affordance.

## Follow-up note

The `tabSlug` derivation is duplicated from `Config.tsx wireTabSpecs`; a small follow-up can extract it to a shared `lib/` helper so the label->slug convention has a single home.


singlerider (approved):
First review of this split. Verified at head `40f11125` (CI green, MERGEABLE). Read the source. **All 🟢; approving.** This is the clean split-2/3 of #8022 and it resolves my original tab-routing blocker.

### 🟢 My #8022 blocker #2 is fixed: tab routing now reads backend ConfigTab metadata

`healthFixTarget` no longer pattern-matches field names to guess a tab. The component builds a `path -> ConfigTab` index from `/api/config/list` entries (`Dashboard.tsx:467`, `if (e.tab) index.set(e.path, e.tab)`) and `healthFixTarget` reads it via `tabForPath` (`:96`). A re-tabbed or newly added field now deep-links to its real tab with no UI change, and it falls back to no `?tab=` (the agent's default tab) when the index is unknown. That is exactly the registry-driven shape I asked for.

### 🟢 The remaining regexes parse entity identity, not tab routing

The `agents.<alias>`, `channels.<type>.<alias>`, and `providers.models.*` matches extract WHICH entity failed out of the free-text `last_error` string. That is not a registry shadow: there is no backend registry mapping a free-text error to a config path, and the error message is the only source of the failing entity. They are correctly start-anchored so prose like "failed to load agents.json" does not yield a bogus `agents.json` target.

### 🟢 tabSlug mirrors the shared convention, not a duplicated source

`tabSlug` (`:70`) uses the same `label.toLowerCase().replace(/\s+/g, "-")` derivation as `Config.tsx wireTabSpecs`, so the deep link lands on the same tab the config page renders. Shared convention, not a second source of truth.

Tight scope (2 files), graceful degradation, registry-driven routing. Approving.

PR #6038, opened 2026-04-23
fix(cron): add claim/release lock to prevent repeated job execution

xydigit-sj:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  The scheduler selected due jobs based only on `next_run <= now`, without checking whether a job was already in flight. If a job ran longer than the poll interval (default 15s), the next poll would select it again and launch duplicate executions. In the reported case (#6037), a daily job was executed 20 times in a 3-minute window.

- **Scope boundary:** Only `crates/zeroclaw-runtime/src/cron/store.rs` and `scheduler.rs`. No change to job semantics, schedule calculation, or external APIs.

## Changes

1. **Schema migration:** Added `locked_at TEXT` column to `cron_jobs` (auto-migrated via `add_column_if_missing`).
2. **Query filters:** `due_jobs()` and `all_overdue_jobs()` now exclude rows where `locked_at IS NOT NULL`.
3. **Claim before execute:** `claim_job()` atomically sets `locked_at` via `UPDATE ... WHERE locked_at IS NULL`.
4. **Release after execute:** `release_job()` clears the lock in `persist_job_result` (runs on both success and failure).
5. **Crash recovery:** `unlock_all_jobs()` runs once at scheduler startup to clear stale locks left by a previous crash.

## Verification

Added 5 regression tests in `store.rs`:
- `claim_job_prevents_duplicate_claims`
- `release_job_allows_reclaim`
- `due_jobs_excludes_locked_jobs`
- `all_overdue_jobs_excludes_locked_jobs`
- `unlock_all_jobs_clears_stale_locks`

## Compatibility

- Existing databases are auto-migrated on next scheduler startup.
- No config changes required.
- No behavior change for jobs that finish within the poll interval.

Closes #6037

PR #8107, opened 2026-06-21
fix(cron): claim/release in-flight lock to prevent duplicate launches

### LINKED ISSUE #6037: [Bug] Cron jobs can be launched repeatedly while still running

OmkumarSolanki:
## Summary

- **Base branch:** `master`
- **What changed and why:** A cron job that runs longer than the scheduler poll interval could be selected and launched again on every poll, because `next_run` is only advanced after a run finishes. #6037 observed a once-daily job run 20 times in a ~3-minute burst. This adds an atomic in-flight claim so a job that is currently executing is never selected again until its run completes:
  - New nullable `locked_at` column on `cron_jobs`, added through the existing `add_column_if_missing` migration path so old DBs upgrade transparently on first read.
  - `due_jobs` / `all_overdue_jobs` skip rows where `locked_at IS NOT NULL`.
  - `claim_job`: atomic `UPDATE … WHERE id = ? AND locked_at IS NULL` — at most one caller transitions a job from idle to in-flight.
  - `release_job`: clears the lock once a run completes; `clear_stale_locks`: startup recovery for locks orphaned by a crash (uses the non-creating read helper so empty workspaces stay untouched).
  - Scheduler wiring: claim at selection (poll loop + startup catch-up), release after `persist_job_result`, and release on the orphan / security-policy skip paths so a skipped job is retried next poll rather than wedged until restart.
- **Scope boundary:** Only `crates/zeroclaw-runtime/src/cron/` (`store.rs`, `scheduler.rs`, `mod.rs` re-exports). No change to schedule calculation, job semantics, delivery, or external/CLI/API surface. `due_jobs` / `all_overdue_jobs` are scheduler-internal (display/CRUD paths use `list_jobs` / `get_job`), so the new filter does not hide in-flight jobs from any UI. The manual `execute_job_now` trigger path is intentionally left unchanged (user-initiated, out of scope for the scheduler auto-launch race).
- **Blast radius:** The cron scheduler within `zeroclaw-runtime`. The added `WHERE locked_at IS NULL` predicate and two `UPDATE`s per run are the only behavioral changes to selection/execution; one new nullable column.
- **Linked issue(s):** Closes #6037. Related #6038 (previous attempt at this fix, closed stale; the maintainer invited a fresh narrow PR linking back).
- **Labels:** bug, risk: high, size: M, cron, runtime

```bash
cargo fmt --all -- --check
```

  - `cargo fmt --all -- --check` — clean, no diff (exit 0).
  - `cargo clippy --all-targets -- -D warnings` — exit 0, whole workspace compiled, no warnings.
  - `cargo nextest run --locked --workspace --exclude zeroclaw-desktop` (the CI command):
    ```
    Summary [ 64.098s] 9388 tests run: 9388 passed, 11 skipped
    ```
  - Cron module specifically (includes the 6 new tests):
    ```
    cron::scheduler::tests::claim_due_jobs_skips_in_flight_job ... PASS
    cron::scheduler::tests::process_due_jobs_releases_lock_for_skipped_orphan_job ... PASS
    cron::store::tests::claim_job_is_atomic_and_blocks_second_claim ... PASS
    cron::store::tests::due_jobs_skips_claimed_jobs ... PASS
    cron::store::tests::clear_stale_locks_releases_in_flight_locks ... PASS
    cron::store::tests::clear_stale_locks_on_empty_workspace_does_not_create_db ... PASS
    ```
- **Beyond CI — what did you manually verify?** Mutation-checked the two core guards: removing `AND locked_at IS NULL` from `due_jobs` makes `due_jobs_skips_claimed_jobs` fail; removing it from `claim_job` makes `claim_job_is_atomic_and_blocks_second_claim` and `claim_due_jobs_skips_in_flight_job` fail — confirming the tests guard the production code, not their own setup. Also verified: the orphan and security-policy skip paths release the lock (no leak) so a misconfigured/transient-error job is retried next poll instead of wedged until restart; `clear_stale_locks` does not create a DB on an empty workspace; and `due_jobs` / `all_overdue_jobs` have no callers outside the scheduler, so the filter introduces no listing/display regression.
- **If any command was intentionally skipped, why:** None skipped. (Locally the workspace run is invoked with `CARGO_INCREMENTAL=0` to avoid an incremental-cache disk blowup on this machine; it has no effect on results.)

## Compatibility (required)

- Backward compatible? `Yes`
- The `locked_at` column is added idempotently via the existing migration path; existing DBs upgrade transparently on first read (column defaults to NULL = unlocked). No user action required. Downgrading to a prior build simply ignores the column.

- **Fast rollback command/path:** Revert the squash merge commit from #8107 (`git revert <merge-sha>`). Safe: reverting drops the claim/release calls and the `locked_at IS NULL` predicate; the leftover nullable `locked_at` column is harmless and ignored by the reverted code.
- **Observable failure symptoms:** Duplicate runs reappear — multiple `cron_runs` rows for one job within a single poll interval (`workspace/cron/jobs.db`), or repeated cron result broadcasts for the same `job_id`. If the lock were leaking, watch for a job silently dropping out of scheduling (no further runs) with no `Cron job already in flight` debug log preceding it; startup logs `Cleared stale cron in-flight locks at startup` when recovering crash-orphaned locks.


Audacity88 (approved):
Thanks @OmkumarSolanki. I reviewed the live PR state at 6424fa4bd7b742c58af9112acf5d6113b5a7a8a8, the PR body's linked issue context, the cron scheduler/store diff, and the GitHub Actions check-run status. I did not run local Cargo for this review-only pass; I relied on the PR's reported workspace validation, the six new focused cron tests, green public CI, and source review.

### 🟢 What looks good — The claim happens before duplicate launch can occur

The shape here matches the duplicate-run failure described in the PR body. `due_jobs` and `all_overdue_jobs` now filter out rows with `locked_at IS NOT NULL`, and `claim_due_jobs` performs the atomic `claim_job` step before the scheduler hands work to `process_due_jobs`. That closes the specific race where `next_run` remains due while a long-running job is still executing, so the next poll cannot select and launch the same row again.

The regression coverage is also pointed at the right behavior. The store tests cover the atomic second-claim failure, due/overdue filtering, release/reclaim, stale-lock clearing, and the empty-workspace no-create path. The scheduler tests add the higher-level selection check plus the skipped-orphan release path, which is exactly the sort of lock-leak edge this fix needed.

### 🟢 What looks good — The lock lifecycle covers the important scheduler exits

The release path runs after `persist_job_result`, including success, normal failure, reschedule, disable, and delete outcomes. The claimed-but-skipped paths for orphaned jobs and security-policy construction failures also clear the lock, so those jobs retry on the next poll instead of staying hidden until restart. Startup recovery uses the existing non-creating helper for empty workspaces while still opening an existing DB read/write, so crash-orphaned locks get cleared without creating `workspace/cron/jobs.db` for users who have no cron state yet.

### 🟡 Warning — Refresh the branch before merge

This approval is for the current cron patch, not a merge-ready statement. Because this is `risk: high` runtime/persistence behavior, the branch should still be updated or otherwise get fresh official CI on the actual merge candidate before merge.

PR #7954, opened 2026-06-19
feat(zerocode): add agent rename dashboard flow

Audacity88:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:** ZeroCode's Dashboard Agents tab now has an inline rename flow behind the rename key, using the runtime `config/map-key-rename` RPC from merged PR #8109 so operators can rename an agent without leaving the TUI.
- **What changed and why:** The Agents detail pane now advertises rename as a quiet `[rename]` action on the alias row, the Agents status line shows the live rename key hint, and rename success/info messages are less visually disruptive.
- **What changed and why:** Overview agent clicks now focus the matching row in `Dashboard > Agents` instead of jumping to `Config > Agents`, so agent inspection and rename stay in one dashboard workflow.
- **What changed and why:** The rename prompt reuses the existing text-input accept/cancel flow, keeps the typed alias in transient UI state only, refreshes/selects the renamed agent after the RPC succeeds, and clears stale rename notices when leaving Dashboard.
- **What changed and why:** The ZeroCode RPC client exposes `config_map_key_rename()` with a 120s timeout for agent renames, because the runtime cascade can commit config changes and move workspace/state files before returning. The timeout is a UI responsiveness guard, not a server-side transaction limit: it is intended to be comfortably above normal local config/workspace moves while still preventing the terminal from hanging indefinitely. If a pathological workspace move exceeds that bound, the server may still finish the rename after the client reports a timeout, so operators should refresh/reopen the dashboard before retrying.
- **Scope boundary:** This PR does not implement alias cascade semantics itself; merged PR #8109 owns the runtime/config/gateway behavior. This PR does not add agent enable/disable controls or the broader ZeroCode integration-stack items.
- **Blast radius:** ZeroCode Dashboard agent rename UI, Dashboard agent navigation, and the ZeroCode RPC client wrapper.
- **Linked issue(s):** Related #7468 and #7790. Builds on merged PR #8109.
- **Labels:** `enhancement`, `risk:medium`, `size:M`, `zerocode`.

## Validation Evidence (required)

```text
git diff --check
result: passed after the rename timeout follow-up

zc-cargo fmt --package zerocode -- --check
result: passed after the rename timeout follow-up

zc-cargo test -p zerocode config_map_key_rename_sends_path_and_aliases
result: passed after the rename timeout follow-up

[...]

PR #8109, opened 2026-06-21
fix(runtime): cascade config alias renames safely

### LINKED ISSUE #8102: [Bug]: config alias rename RPC can overflow daemon task stack
## Affected component

runtime/daemon

## Severity

S1 - workflow blocked

## Current behavior

While dogfooding the agent-rename work from #7954, the daemon can abort when the TUI triggers the runtime `config/map-key-rename` path for an agent alias rename.

Observed failure:

```text
thread 'tokio-rt-worker' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

The likely source is the runtime RPC alias-rename handler in #7954: `handle_config_alias_rename()` is an `async fn` that clones and retains a full `Config` value in the generated future. A local regression test against that branch measured the alias-rename future at about 90 KiB before the fix, which is too large for this RPC worker path and can overflow the task stack before the handler returns.

## Expected behavior

Agent/provider/channel alias rename through `config/map-key-rename` should not make the daemon abort.

The handler should keep large config snapshots out of the RPC task stack, then persist and swap the rewritten config normally. Agent rename should still return a normal `RenameMapKeyResponse`, including warnings for post-persist owned-state/workspace side effects when those fail.

## Steps to reproduce

1. Check out #7954 at head `e689d27dcdc4267e32070e3f8864b571551a48be`.
2. Build and run the daemon from that branch.
3. Trigger an agent alias rename through zerocode's Agents tab, or call the same runtime RPC path with `config/map-key-rename` for the `agents` map.
4. Observe that the daemon can abort with the stack overflow shown above instead of returning a structured RPC success or error.

Regression-test reproduction path:

1. In `crates/zeroclaw-runtime/src/rpc/dispatch.rs`, construct a dispatcher and call `handle_config_alias_rename()` without awaiting it.
2. Measure `std::mem::size_of_val(&future)`.
3. On the pre-fix implementation, the future is about 90 KiB because it captures the cloned config across an await.

## Impact

Affected users are operators testing or relying on the #7954 agent-rename path before release. The immediate consequence is severe for that workflow: a routine TUI rename can terminate the daemon instead of reporting a recoverable RPC error.

This is also a guardrail for future runtime RPC work: handlers that hold large config snapshots across `.await` can create stack-sensitive daemon crashes even when the business logic is otherwise correct.

## Logs / stack traces

```text
thread 'tokio-rt-worker' has overflowed its stack
fatal runtime error: stack overflow, aborting
```

No private paths, account data, session IDs, provider request IDs, or full local logs are included.

## ZeroClaw version

#7954 head `e689d27dcdc4267e32070e3f8864b571551a48be`.

## Rust version

Not yet isolated to a Rust compiler version. This is a runtime future-size / task-stack issue in the handler implementation.

## Operating system

Observed during macOS dogfood testing. The future-size defect is platform-independent, though the exact stack-overflow threshold may vary by runtime and platform.

## Regression?

Yes, within the active agent-rename work. The affected alias-rename runtime path is introduced by #7954 and is not part of the latest released behavior yet.

## Related

- #7954
- #7790
- #7907 is related agent-rename cleanup context, but it covered persistence ordering and is not a duplicate of this stack-overflow bug.

## Acceptance criteria

- `config/map-key-rename` alias rename keeps large config snapshots out of the RPC task stack.
- Agent rename still uses the shared alias cascade and persists the config before post-persist owned-state/workspace side effects.
- A regression test fails if the alias-rename future grows large enough to put the RPC worker stack at risk.
- The existing agent cascade behavior test remains green.

## Duplicate search notes

I searched open and closed issues/PRs for `tokio-rt-worker`, `overflowed its stack`, `stack overflow`, `handle_config_alias_rename`, `config/map-key-rename`, and `agent rename`. I found related agent-rename work, especially #7907 and #7954, but not a dedicated issue for the daemon stack overflow.

## Pre-flight checks

- [x] I reproduced this on the active #7954 agent-rename branch. The affected implementation is draft PR code, not the latest released behavior.
- [x] I redacted secrets, tokens, personal data, private paths, raw local logs, account/session IDs, and provider request IDs from all submitted content.

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Routes runtime RPC alias renames through the same config-reference cascade used by the gateway so agent/provider/channel alias references stay consistent.
  - Boxes the runtime `config/map-key-rename` future path so cloned config snapshots do not sit on the RPC worker stack.
  - Adds agent-rename side-effect convergence for the case where config already saved `from -> to` but workspace or owned-state updates lagged.
  - Refuses agent renames while live RPC sessions still use the old alias, avoiding stale in-memory sessions recreating old-alias state.
- **Scope boundary:** This PR does not change the zerocode UI, add live in-place session repointing, or broaden generic non-alias map-key rename behavior.
- **Blast radius:** Runtime RPC config rename, gateway config rename/delete-plan parser sharing, config alias-reference helpers, and the RPC rename result shape.
- **Linked issue(s):** Closes #8102.
- **Labels:** `bug`, `risk: high`, `size: L`, `runtime`, `config`, `gateway`.

  - `cargo fmt -p zeroclaw-config -p zeroclaw-runtime -p zeroclaw-gateway -- --check`

    ```text
    passed with no formatting diff
    ```

  - `git diff --check`

    ```text
    passed with no whitespace errors
    ```

  - `cargo test -p zeroclaw-runtime config_map_key_rename`

    ```text
    running 3 tests
    test rpc::dispatch::tests::config_map_key_rename_resumes_committed_agent_rename_side_effects ... ok
    test rpc::dispatch::tests::config_map_key_rename_uses_agent_cascade ... ok
    test rpc::dispatch::tests::config_map_key_rename_refuses_active_agent_sessions ... ok

    test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 2324 filtered out
    ```

  - `cargo test -p zeroclaw-runtime config_alias_rename_future_is_small_enough_for_rpc_task_stack`

    ```text
    running 1 test
    test rpc::dispatch::tests::config_alias_rename_future_is_small_enough_for_rpc_task_stack ... ok

    test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2326 filtered out
    ```

  - `cargo check -p zeroclaw-gateway`

    ```text
    ```

  - GitHub Quality Gate and CI Required Gate on PR #8109

    ```text
    Format, Lint, Test, Security, Zerocode RPC Boundary, and CI Required Gate passed.
    ```

  - Compared the branch against current `origin/master`: this PR has one source commit, and upstream drift at the time of the check was docs-only outside this PR's source files.
  - Verified the regression covers the public runtime handler future, not only the private boxed helper.
  - Verified active RPC sessions block agent rename instead of allowing old-alias in-memory sessions to continue.
  - Verified a re-issued `from -> to` agent rename converges lagging workspace side effects when config already names `to`.
- **If any command was intentionally skipped, why:**  I ran focused local checks for the stack-size, cascade, active-session, resume, formatting, whitespace, and gateway compile surfaces changed here. I did not duplicate the full workspace clippy/test suite locally after that because the official GitHub Quality Gate ran against this PR head and passed Lint, Test, Security, platform checks, and CI Required Gate.

- New permissions, capabilities, or file system access scope? Yes.
  - Runtime RPC agent rename now performs the same owned-state side effects expected from agent rename, including moving the per-agent workspace from the old alias path to the new alias path. The operation is confined to config-derived agent workspace paths, refuses live old-alias RPC sessions, persists config before side effects, and reports side-effect warnings instead of silently hiding failures.

## Compatibility (required)

- Backward compatible? Yes.
- If `No` or `Yes` to either: No user config or CLI upgrade steps. The RPC rename response includes `warnings` only when post-persist side effects need to be reported.

- **Fast rollback command/path:** `git revert <merge-sha>` after merge.
- **Observable failure symptoms:** Runtime `config/map-key-rename` returns unexpected alias-rename errors, agent rename returns side-effect warnings, or daemon logs show stack-overflow / abort symptoms while renaming config aliases.

- Superseded PRs + authors: N/A.
- Scope materially carried forward: N/A.
- `Co-authored-by` trailers added in commit messages for incorporated contributors? No.
- If `No`, why: No superseded contributor PR scope is incorporated.


singlerider (approved):
First review (no prior reviews or comments). I read the diff at head 88d2da76 across the four files and traced the rename cascade, the live-session guard, the convergence path, and the boxed-future stack test. CI is green. This is the runtime slice #7954 stacks on, and the response shape it returns (the `warnings` field) is what that UI consumes, so the two line up.

### 🟢 What looks good: rename rewrites refs, which is the right semantic (not delete's refuse-on-hard)

`rename_with_cascade` correctly diverges from the delete cascade: delete refuses on hard references, but rename **rewrites** every reference (soft and hard) to name the new alias, so an enabled `heartbeat.agent` or other hard ref does not block the rename, it gets repointed. The only refusal is on collision (target alias already exists), which is correct. The handler marks every `report.dirty_paths` entry dirty and persists, so the rewritten referrers and the renamed key all reach disk together rather than leaving a half-renamed config.

### 🟢 What looks good: refuses agent rename while live sessions hold the old alias

The handler checks `sessions.count_by_agent().get(&from)` and refuses when any session is still active under the old alias. That is the right guard: an in-memory session keyed to the old alias would otherwise recreate old-alias state after the config rename committed, silently resurrecting what the rename was meant to remove. Refusing instead of attempting live in-place repointing keeps the scope honest (the body says in-place repointing is explicitly out of scope) and avoids a much harder correctness problem.

### 🟢 What looks good: convergence and workspace-move warnings are handled, not swallowed

The convergence path handles the partial-failure case where config already saved `from -> to` but the workspace or owned-state move lagged: re-running the rename detects residue via `agent_rename_residue_exists` and converges the lagging side-effects instead of erroring on "already renamed." `move_renamed_agent_workspace` collects failures into `warnings` rather than failing the whole rename or hiding the problem, and those warnings flow out in `ConfigMapKeyRenameResult` so the caller (and #7954's TUI) can surface them. `config_map_key_rename_resumes_committed_agent_rename_side_effects` pins the convergence behavior.

### 🟢 What looks good: the future is boxed and the stack cost is tested

Boxing the rename future via `BoxRpcFuture` keeps the cloned config snapshot off the RPC worker task stack, and `config_alias_rename_future_is_small_enough_for_rpc_task_stack` asserts `size_of_val(&future) < 16 * 1024`, which turns "don't let large config snapshots sit on the stack" into an enforced invariant rather than a hope. `config_map_key_rename_uses_agent_cascade` covers the cascade routing. The `RenameError -> RPC` mapping sends `PostCondition` to `INTERNAL_ERROR`, which is appropriate for a verify-after-rewrite failure.

Approving.

singlerider:
@Audacity88 this one is approved but conflicts with master now, so it got skipped in the merge batch (the other approved PRs landed). Merge master into the branch to resolve, then it's clear to go. No re-review needed unless the conflict resolution changes behavior.

PR #6829, opened 2026-05-21
fix(discord): treat thread messages as belonging to parent channel in channel_ids filter

singlerider:
## Summary

- **Base branch:** `master` (all contributions)
- **What changed and why:**
  - Discord thread/forum-post messages carry their own `channel_id`, distinct from the parent channel/forum. When `channel_ids` is configured on a Discord channel, the filter only did a direct match, so thread messages were silently dropped.
  - When a direct match fails, the listener now looks up the thread's `parent_id` via the Discord API and accepts the message if the parent is in the allowlist.
  - Refactored `is_thread_channel()` → `thread_parent()` returning `Option<String>` (parent ID) instead of `bool`, so the same API call serves both the channel filter and the existing thread-timestamp logic.
  - Extracted `channel_passes_filter()` as a pure function for testability.
  - Changed the thread cache type from `HashMap<String, bool>` to `HashMap<String, Option<String>>` to store parent IDs instead of just thread-or-not booleans.
- **Scope boundary:** No config schema changes. No new config fields. Runtime behavior change only — threads whose parent channel is in `channel_ids` are now accepted.
- **Blast radius:** Only affects Discord channels with a non-empty `channel_ids` filter. Channels with an empty filter (accept-all) are unaffected. The thread-timestamp (`thread_ts`) logic is updated to use the same refactored method but behavior is identical.
- **Linked issue(s):** None (discovered during live deployment testing)
- **Labels:** `bug`, `channels`, `risk: low`, `size: S`

## Validation Evidence (required)

- **Commands run and tail output:**

```
$ cargo +nightly fmt -- --check
(clean, exit 0)

$ cargo clippy -p zeroclaw-channels -- -D warnings
    Checking zeroclaw-channels v0.8.0-beta-1
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 19s
(clean, exit 0)

$ cargo test -p zeroclaw-channels
test result: ok. 1090 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.88s
(all integration tests also pass)

$ cargo test -p zeroclaw-channels -- discord::tests::channel_filter
running 4 tests
test discord::tests::channel_filter_empty_accepts_everything ... ok
test discord::tests::channel_filter_direct_match ... ok
test discord::tests::channel_filter_direct_match_skips_parent_check ... ok
test discord::tests::channel_filter_thread_parent_fallback ... ok
test result: ok. 4 passed; 0 failed; 0 ignored
```

[...]

PR #8110, opened 2026-06-21
fix(channels/discord): authorize autocomplete in parent-allowlisted threads

### LINKED ISSUE #8103: [Bug]: Discord autocomplete fails closed in parent-allowlisted threads
## Affected component

channel

## Severity

S2 - degraded behavior

## Current behavior

Discord slash-command autocomplete can fail closed inside a thread when `channel_ids` allows the parent channel but not the thread id directly.

The normal Discord message path already treats thread messages as belonging to their parent channel, covered by #6829. The type-4 autocomplete path is different: it answers inline and avoids reject/defer side effects, but it gates the interaction using the thread `channel_id` with no resolved parent. In a parent-allowlisted thread, that can produce an empty autocomplete choice set even though normal messages in the same thread are allowed.

This is adjacent to the Discord interaction-surface work in #7831 and #7965. It may be an intentional fail-closed limitation, but it should be tracked so the behavior does not stay implicit.

## Expected behavior

One of these should be true:

- Autocomplete in a parent-allowlisted Discord thread can use the parent channel authorization safely, without unbounded fresh Discord REST calls on every keystroke.
- The limitation is documented clearly: channel-filtered Discord threads may fail closed for slash autocomplete unless the parent channel is already known in a safe cache.

## Steps to reproduce

```bash
# Conceptual repro:
# 1. Configure the Discord channel with slash commands enabled and channel_ids
#    allowing a parent channel.
# 2. In Discord, create or use a thread under that allowed parent channel.
# 3. Invoke a slash command option that has autocomplete choices inside the thread.
# 4. Observe that autocomplete returns no choices, while normal messages in the
#    same thread can pass the parent-channel filter.
```

Source-confirmed on current `master`: the type-4 autocomplete branch authorizes without a resolved thread parent, while the normal message and other interaction paths have thread-parent handling.

## Impact

Users who rely on Discord `channel_ids` scoped to parent channels can use normal threaded messages, but may lose slash-option autocomplete inside those same threads. This is especially visible for typed skill options with many choices.

The first fix should avoid turning every authorized autocomplete keystroke into an unbounded Discord API lookup. A safe design could use an already-cached thread parent, a short/cooldown-backed lookup, or a singleflight cache warm path.

## Logs / stack traces

No runtime log excerpt is attached. This report is based on source inspection and related merged behavior in #6829 and #7965.

## ZeroClaw version

`master` at `b10d222704`

## Rust version

N/A; this is Discord interaction authorization behavior.

## Operating system

Platform-independent.

## Regression?

Unknown.

## Related

- #7831
- #7965
- #6829

## Pre-flight checks

- I source-confirmed this on current `master`.

OmkumarSolanki:
## Summary

- **Base branch:** `master`
- **What changed and why:**
  - Discord slash-command autocomplete (type-4 `APPLICATION_COMMAND_AUTOCOMPLETE`) gated the interaction with `interaction_gate(..., thread_parent = None)`. The normal message paths resolve the thread parent and pass it to `channel_passes_filter`, so a thread under an allowlisted parent is authorized — autocomplete was not, so it **failed closed to an empty choice set** in the very thread the same user can otherwise post messages in.
  - Added `discord_thread_parent_cached()` — a **cache-only (no REST)** read of the existing `thread_channels` cache — and wired it into the type-4 arm so a parent populated by an earlier message authorizes autocomplete consistently with the message path (#6829).
  - Cache-only by design: the per-keystroke autocomplete path stays side-effect-free (**no** unbounded Discord REST probe per keystroke); an uncached thread still fails closed.
- **Scope boundary:** Only the Discord type-4 autocomplete authorization path. Does **not** change `interaction_gate` / `channel_passes_filter` semantics, the message/component/modal arms, the REST `discord_thread_parent` lookup, or any non-Discord channel. Adds no REST call to the autocomplete path.
- **Blast radius:** Discord channels configured with `channel_ids` + slash commands. The only behavior change: autocomplete returns choices (instead of empty) inside a thread whose parent is allowlisted **and already cached**. No other channel/provider/runtime path touched.
- **Linked issue(s):** Closes #8103. Related #6829 (thread-parent handling on the message path).
- **Labels:** `bug`, `risk: medium`, `size: S`, `channel`, `channel:discord`, `priority:p2`, `status:accepted` (snapshot — I can't set `risk:` / `size:`; maintainer to confirm).

Full battery run locally on this branch (rustc 1.96.0, macOS; `CARGO_INCREMENTAL=0` per repo machine guidance).

```
$ cargo fmt --all -- --check
(clean — exit 0, no diff)

$ CARGO_INCREMENTAL=0 cargo clippy --all-targets -- -D warnings
    Checking zeroclaw-channels v0.8.1 (.../crates/zeroclaw-channels)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 31s
(no warnings)

$ CARGO_INCREMENTAL=0 cargo nextest run --locked --workspace --exclude zeroclaw-desktop
    Summary [  66.828s] 9385 tests run: 9385 passed, 11 skipped

$ cargo nextest run -p zeroclaw-channels autocomplete thread_parent interaction_gate channel_passes
    Summary [   0.277s] 14 tests run: 14 passed, 1074 skipped
```

- **Commands run and tail output:** as above.
  - **Mutation check:** reverted the arm to pass `None` (the original bug) → the source-assertion test `autocomplete_arm_resolves_cached_thread_parent_before_gate` **FAILS** (`type-4 arm resolves the cached thread parent` panic); restored → all pass. The regression test genuinely guards the fix.
  - Traced consumers of the changed arm: `discord_thread_parent_cached` is new (no other callers); `interaction_gate` / `channel_passes_filter` are unchanged and now receive a real parent instead of `None`, exercising their existing, already-tested parent rule. The autocomplete answer remains a single type-8 callback (`autocomplete_answer_posts_a_single_type8_callback_and_nothing_else` still passes).
  - Confirmed cache-only semantics in a unit test where no client/token is reachable, so a non-`None` result can only come from the cache (never a REST probe).
  - **Not verified:** a live Discord round-trip (no live workspace). The fix relies on the existing, separately-tested `thread_channels` cache populated by the message path.
- **If any command was intentionally skipped, why:** none skipped. Used `cargo nextest` (project standard per `contributing/how-to.md`) in place of `cargo test`.

- New external network calls? **No** — explicitly cache-only; adds no Discord REST call (autocomplete path stays side-effect-free).
- PII, real identities, or personal data in diff, tests, fixtures, or docs? **No** (synthetic ids only: `parentA`, `thread_cached`, `u1`).
- **If any `Yes`, describe the risk and mitigation:** This change *loosens* one authorization decision, so for the record: autocomplete is now authorized only for a user who already passes the peer allowlist **and** is in a thread whose parent channel is explicitly in `channel_ids` — exactly the authorization the normal message path already grants in that thread via the same `channel_passes_filter` parent rule (#6829). It grants nothing beyond what the user can already do by posting a message there. Cached parents come only from successful Discord lookups (not user-spoofable), and uncached threads stay fail-closed. Net: a consistency fix, not privilege escalation.

## Compatibility (required)

- Backward compatible? **Yes.**
- **Upgrade steps:** None. Existing `channel_ids` configs are unaffected; the only observable difference is that slash-option autocomplete inside a parent-allowlisted thread now returns choices once that thread's parent is cached (previously empty).

- **Fast rollback command/path:** `git revert <merge-sha>` — single self-contained commit (`3fd572149`), reverts cleanly to the prior fail-closed behavior.
- **Feature flags or config toggles:** None (no toggle introduced). An operator who wants autocomplete in a specific thread without this change can list the thread id directly in `channel_ids` (the always-supported direct-match path).
- **Observable failure symptoms:** unexpected autocomplete choices appearing inside Discord threads, or autocomplete latency/429 pressure. Grep logs for `discord autocomplete answer failed`. Because the path issues no new REST call, no new rate-limit pressure is expected; 429s on the autocomplete path would be an unexpected rollback signal.

### DIFF
--- a/crates/zeroclaw-channels/src/discord/mod.rs
+++ b/crates/zeroclaw-channels/src/discord/mod.rs
@@ -1748,6 +1748,27 @@ async fn discord_thread_parent(
     result
 }
 
+/// Cache-only variant of [`discord_thread_parent`]: returns the cached parent
+/// id for `channel_id` if a prior [`discord_thread_parent`] call already
+/// populated the cache, otherwise `None`. It performs **no** Discord REST call,
+/// so it is safe on the per-keystroke autocomplete (type-4) path where an
+/// authenticated round-trip would violate the side-effect-free requirement.
+///
+/// A miss (`channel_id` never looked up, or looked up and found to be a
+/// non-thread) yields `None`, which keeps channel authorization fail-closed:
+/// the thread can only pass an allowlist via a known, already-cached parent.
+async fn discord_thread_parent_cached(
+    thread_channels: &Arc<AsyncMutex<HashMap<String, Option<String>>>>,
+    channel_id: &str,
+) -> Option<String> {
+    thread_channels
+        .lock()
+        .await
+        .get(channel_id)
+        .cloned()
+        .flatten()
+}
+
 // Discord gateway intent bits (API v10) — the ones zeroclaw consumes or
 // exposes as opt-ins. https://discord.com/developers/docs/events/gateway#gateway-intents
 const INTENT_GUILDS: u64 = 1 << 0;
@@ -3037,6 +3058,7 @@ impl Channel for DiscordChannel {
                                     let guild_filter = guild_filter.clone();
                                     let channel_filter = channel_filter.clone();
                                     let resolver = self.slash_command_resolver.clone();
+                                    let thread_channels = self.thread_channels.clone();
 
                                     zeroclaw_spawn::spawn!(async move {
                                         // Fail-closed authz, side-effect-free:
@@ -3046,12 +3068,22 @@ impl Channel for DiscordChannel {
                                         // don't make here). On denial OR no
                                         // matches we answer an empty choice set.
                                         //
-                                        // No thread-parent REST lookup: it is an
-                                        // authenticated round-trip per keystroke
-                                        // and would defeat the side-effect-free
-                                        // requirement, so a channel-filtered
-                                        // thread simply yields no completions
-                                        // (fail-closed) rather than probing.
+                                        // Thread-parent resolution is CACHE-ONLY:
+                                        // a parent populated by an earlier normal
+                                        // message in this thread lets a
+                                        // parent-allowlisted thread authorize
+                                        // autocomplete consistently with the
+                                        // message path (#6829). It reads the shared
+                                        // cache only — NO Discord REST call — so the
+                                        // per-keystroke path stays side-effect-free;
+                                        // an uncached thread still yields no
+                                        // completions (fail-closed) rather than
+                                        // probing.
+                                        let thread_parent = discord_thread_parent_cached(
+                                            &thread_channels,
+                                            &interaction_channel,
+                                        )
+                                        .await;
                                         let authorized = interaction_gate(
                                             &peers,
                                             &guild_filter,
@@ -3059,7 +3091,7 @@ impl Channel for DiscordChannel {
                                             &user_id,
                                             interaction_guild.as_deref(),
                                             &interaction_channel,
-                                            None,
+                                            thread_parent.as_deref(),
                                         )
                                         .is_ok();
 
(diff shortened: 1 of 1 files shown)


Audacity88 (approved):
I reviewed #8110 at head `3fd572149dde71449141fb0f0c3ce1a36df4b6f1`, including the PR body, #8103, related #6829, the current Discord channel diff, the author's local validation evidence, and the green GitHub checks. There are no prior comments or formal reviews to reconcile. I did not independently run a local Cargo pass or a live Discord smoke test; this approval is based on source review, the PR's regression evidence, and the public CI state.

### ✅ Resolved — autocomplete now reuses the cached thread parent

This resolves the bug described in #8103. The autocomplete arm now reads `self.thread_channels` through `discord_thread_parent_cached()` and passes the cached parent into `interaction_gate`. That makes a parent-allowlisted thread authorize autocomplete through the same `channel_passes_filter` parent rule that #6829 added for normal Discord messages.

The important constraint also holds: this is cache-only. The autocomplete path still avoids a Discord REST lookup on each keystroke, and an uncached thread still fails closed to an empty choice set. That is the right tradeoff for this path because autocomplete runs at typing cadence.

### 🟢 What looks good — the tests pin both the fix and the constraint

The new tests cover the useful boundary: cached parent authorizes, uncached thread denies, non-thread cache entries stay `None`, and the source-level guard catches reverting the type-4 arm back to `None`. The PR body's mutation check is also useful here: it shows the regression guard fails on the original bug shape. Combined with green CI, that is enough for me to approve this narrow fix.

The one thing I would still keep separate from this review is live Discord confidence. The PR is clear that no live workspace round-trip was run. That is fine for this source-level approval, but I would not treat this review as final merge-readiness if maintainers want a live or near-live Discord autocomplete smoke before landing channel behavior changes.

singlerider (approved):
@OmkumarSolanki — reviewed on its own merits, with attention to whether this widens the trust boundary.

🟢 It doesn't — and that's the key point. The change only adds a cache-only parent lookup before the existing `interaction_gate`; the authorization decision still lives entirely in `channel_passes_filter`. The parent value can only have come from the shared `thread_channels` cache, which is populated solely by the REST-backed `discord_thread_parent` on the normal message path — so a thread is authorized only via a parent that was already legitimately resolved elsewhere. An uncached thread returns `None` and stays fail-closed, identical to pre-fix, and there's no REST call on the per-keystroke type-4 path, preserving the side-effect-free requirement.

🟢 This is the correct narrowing of an over-restrictive gap: it makes autocomplete authorization consistent with the message path (#6829) so a user who can post in a parent-allowlisted thread also gets completions there. Coverage is thorough — cache hit / miss / non-thread, gate authorize-via-cached-parent vs fail-closed-when-uncached, and a source-level regression that asserts the arm passes `thread_parent.as_deref()` rather than `None`.

LGTM.

PR #8093, opened 2026-06-21
ci(docker): build base Dockerfiles from source on container changes

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** The Docker Image PR Check (`docker-image-pr.yml`) only triggered on the `.ci` variants and the prep script, and its smoke job builds those pre-built-binary images. Edits to the base `Dockerfile` / `Dockerfile.debian` (the full source-compiling images) triggered no container build at all. That gap let a stale `COPY --parents crates/aardvark-sys/build.rs ./` line ship on master and break every fresh build off upstream. This adds the base Dockerfiles + `.dockerignore` to the `paths:` filter and a `source-images` job that actually compiles both base Dockerfiles from the repo root via buildx with GHA layer caching.
- **Scope boundary:** Only `docker-image-pr.yml`. Does not change the Dockerfiles themselves, the `.ci` variants, the smoke job, the prep script, or any release/publish workflow.
- **Blast radius:** PR CI only. Adds a job that runs solely when container-type files change. No effect on the binary, runtime, or release pipeline.
- **Linked issue(s):** Related #8092 (the stale-COPY fix this guard would have caught).
- **Labels:** `risk: high` (touches `.github/workflows/**`), `size: XS`.

## Validation Evidence (required)

Workflow-only change; no Rust touched, so the cargo battery is N/A.

```
$ python3 -c "import yaml; yaml.safe_load(open('.github/workflows/docker-image-pr.yml')); print('OK')"
OK

$ git diff upstream/master --stat
 .github/workflows/docker-image-pr.yml | 34 ++++++++++++++++++++++++++++++++++
 1 file changed, 34 insertions(+)
```

Action pins (`docker/build-push-action@263435...` v6.18.0, `docker/setup-buildx-action@e468171...` v3.11.1) are already used elsewhere in `.github/workflows/`, confirmed consistent.

- **Beyond CI: what did you manually verify?** YAML parses; the two new job names are unique; `paths:` now includes `Dockerfile`, `Dockerfile.debian`, `.dockerignore`. The full ~60 min source compile was **not** run locally in this session and is not faked here. The honest end-to-end proof is the `source-images` job running on this very PR (the workflow change is itself in `paths:`), so CI will exercise it before merge.
- **If any command was intentionally skipped, why:** local 60 min cold container compile skipped: impractical to run interactively; CI on this PR is the real validator.

## Security & Privacy Impact (required)

[...]

PR #8092, opened 2026-06-21
fix(docker): drop stale aardvark-sys build.rs COPY

singlerider:
## Summary

- **Base branch:** `master`
- **What changed and why:** Commit c338e0949d (#8028) gated `aardvark-sys` behind the hardware feature and deleted `crates/aardvark-sys/build.rs` (libloading resolves the FFI symbols at runtime, so the build script became a no-op). Both `Dockerfile` and `Dockerfile.debian` still carried `COPY --parents crates/aardvark-sys/build.rs ./`, so a fresh build off `upstream/master` fails at the dependency pre-fetch step. Removed the dead COPY line and the now-stale comment block describing why the build script was copied.
- **Scope boundary:** Touches only the two Dockerfiles. No Rust, no CI workflow, no `.dockerignore`, no build logic beyond removing a file that no longer exists.
- **Blast radius:** Container image builds (`Dockerfile`, `Dockerfile.debian`) only. No effect on the binary, runtime, or any crate.
- **Linked issue(s):** None.
- **Labels:** `risk: low`, `size: XS`.

## Validation Evidence (required)

Dockerfile-only change; no Rust source touched, so the cargo battery is N/A. Verified the failure reproduces on a clean clone off `upstream/master` and that the referenced file is gone:

```
$ git ls-tree upstream/master crates/aardvark-sys/build.rs
(empty — file deleted by c338e0949d / #8028)

$ git diff upstream/master --stat
 Dockerfile        | 3 ---
 Dockerfile.debian | 3 ---
 2 files changed, 6 deletions(-)
```

Original build error this fixes:

```
[3/4] STEP 7/27: COPY --parents crates/aardvark-sys/build.rs ./
Error: building at STEP "COPY --parents crates/aardvark-sys/build.rs ./":
  copier: stat: "/crates/aardvark-sys/build.rs": no such file or directory
```

- **Commands run and tail output:** see above.
- **Beyond CI — what did you manually verify?** Confirmed `build.rs` is absent on `upstream/master`, confirmed both Dockerfiles were the only files referencing it, confirmed no `.dockerignore` or build step still depends on it.
- **If any command was intentionally skipped, why:** cargo fmt/clippy/test skipped — no Rust changed.

## Security & Privacy Impact (required)

- New permissions, capabilities, or file system access scope? `No`
- New external network calls? `No`
- Secrets / tokens / credentials handling changed? `No`
- PII, real identities, or personal data in diff, tests, fixtures, or docs? `No`

## Compatibility (required)

[...]

PR #8112, opened 2026-06-21
fix(docker): keep Node base policy in container TOML

### LINKED ISSUE #8105: [Feature]: Make Node container-base major selection explicit
### Summary

Make the Node container-base refresh policy explicit so `cargo generate installers` can refresh digests without silently moving ZeroClaw to a new Node major.

### Problem statement

PR #8005 moved Dockerfile base-image pins into `dev/ci/container-base-images.toml` and made that TOML the single source of truth for generated Dockerfile `ARG` zones. That is the right direction.

The remaining policy gap is the `discover = true` Node row. The current generator discovers every Docker Hub tag matching `<major>-bookworm-slim`, selects the highest major, and writes that tag plus digest back into the canonical TOML. That means a routine `cargo generate installers` run can become an implicit Node major-version bump when Docker Hub publishes a higher plain-major tag.

That is too much policy to hide inside discovery. Digest refresh should stay mechanical, but moving from one Node major to another should be an explicit maintainer decision in the source TOML or in a dedicated PR.

### Proposed solution

Keep `dev/ci/container-base-images.toml` as the single source of truth, but make the Node major policy explicit there.

One acceptable shape:

- keep `discover = true`;
- add a source-owned field such as `major = 26` or `tag_policy = "node-major:26-bookworm-slim"`;
- have `discover_node_tag` / refresh logic resolve the digest for that declared major instead of selecting `max(major)`;
- have `--check` verify both the generated Dockerfile zone and the TOML tag match the declared policy;
- add tests covering both accepted and rejected Node tag shapes.

This preserves the #8005 TOML-owned pin model while preventing a registry tag list from choosing the next runtime major for us.

### Non-goals / out of scope

- Do not undo the #8005 generated-zone/TOML source-of-truth model.
- Do not hand-edit Dockerfile pins outside the generator.
- Do not decide the long-term Node support policy beyond making the current major-selection rule explicit.
- Do not add a network call to `cargo generate installers --check`; the check path should remain offline.

### Alternatives considered

- Keep selecting the highest Docker Hub major automatically. This keeps the generator simple, but it lets a routine digest refresh make an implicit major-version decision.
- Disable Node discovery entirely and require a full literal tag in TOML. This is safer, but it gives up the useful registry-backed digest refresh path that #8005 introduced.

### Acceptance criteria

- `dev/ci/container-base-images.toml` records the intended Node major/tag policy explicitly.
- `cargo generate installers` refreshes the digest for the declared Node policy without selecting a different major on its own.
- `cargo generate installers --check` remains network-free and flags generated-zone drift.
- Tests cover the Node policy parser/validator and a rejection path for a mismatched or unsupported Node tag.
- Existing generated Dockerfile surfaces still derive from the TOML and remain in sync.

### Architecture impact

This affects the xtask installer-generation path and generated container surfaces:

- `dev/ci/container-base-images.toml`
- `xtask/src/generate/container_base.rs`
- `Dockerfile`
- `Dockerfile.debian`

It should stay within the existing single-source-of-truth model from #8005: TOML owns the pin policy and generated Dockerfile zones are materialized views.

### Risk and rollback

Risk is low to medium. The intended behavior narrows automatic updates rather than expanding them. The main risk is making Node digest refresh too strict or accidentally breaking the offline installer drift check.

Rollback is a normal revert of the follow-up PR. If the generator rejects a desired Node major after rollback, maintainers can still set the literal tag/digest in the TOML using the current #8005 behavior until the policy field is fixed.

### Expected routing

Ordinary feature triage

### Next decision surface

Ordinary triage is enough. The first maintainer decision is the exact TOML field shape for the Node major policy.

### Breaking change?

No

### Data hygiene checks

Audacity88:
## Summary

- **Base branch:** `master`
- **What changed and why:** Node container-base discovery now refreshes the digest for the `tag` declared in `dev/ci/container-base-images.toml` instead of querying Docker Hub for the highest matching Node major. The TOML row remains the container-base policy source, so routine installer generation cannot silently move ZeroClaw to a new Node major.
- **What changed and why:** `cargo generate installers --check` stays network-free and validates discover-row Node tags as plain `<major>-bookworm-slim` tags before comparing generated Dockerfile zones against TOML.
- **What changed and why:** The generated Dockerfile surfaces and `dev/ci/container-base-images.toml` were refreshed to the declared `node:24-bookworm-slim` tag with the digest recorded in TOML.
- **Scope boundary:** This PR does not change `.nvmrc`, the long-term Node support policy, Docker build workflows, release/publish behavior, or non-Node base-image pins. It also does not add a second `major` field; the existing TOML `tag` is the container-base Node policy.
- **Blast radius:** Installer-generation and container base-image surfaces only: `xtask` generator logic, `dev/ci/container-base-images.toml`, `Dockerfile`, and `Dockerfile.debian`.
- **Linked issue(s):** Closes #8105.
- **Labels:** `enhancement`, `type: ci`, `risk: medium`, `size: S`, `dev`, `security:docker`.

```text
$ cargo fmt -p xtask
(exit 0)

$ cargo test -p xtask container_base::tests
running 7 tests
test generate::container_base::tests::node_major_for_suite_rejects_non_plain_major ... ok
test generate::container_base::tests::node_major_for_suite_accepts_plain_major ... ok
test generate::container_base::tests::valid_digest_shape ... ok
test generate::container_base::tests::arg_line_round_trips_through_zone_body ... ok
test generate::container_base::tests::source_parses_and_node_is_discover ... ok
test generate::container_base::tests::check_flags_orphan_reference ... ok
test generate::container_base::tests::splice_zones_rejects_non_plain_node_discover_tag ... ok
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 82 filtered out

$ cargo generate installers --check
ok: install-sh in sync
ok: setup-bat in sync
ok: containerfile in sync
ok: dockerfile in sync
ok: dockerfile-debian in sync
ok: pkgbuild in sync
ok: scoop in sync
ok: flake in sync
ok: docker-tags in sync

$ git diff --check
(no output; exit 0)
```

- **Beyond CI — what did you manually verify?** Verified the Node container-base policy is read from the TOML `tag`, generated Dockerfile zones still match the TOML tag+digest, and the generator no longer reads `.nvmrc` for container base-image policy.
- **If any command was intentionally skipped, why:** Workspace clippy and full workspace tests were not run. The touched Rust surface is the xtask container-base generator, and the focused xtask tests plus installer drift check cover the changed behavior.

- New external network calls? `No` (the generator still refreshes registry digests; this removes the Docker Hub tag-list major-selection policy from refresh)

## Compatibility (required)

- Backward compatible? `Yes`
- Config / env / CLI surface changed? `No` for user-facing runtime config, environment variables, and CLI. The maintainer-facing generator policy changed: the existing TOML `tag` now controls the Node major for the discover-row container base.
- If `No` or `Yes` to either: Existing users do not need upgrade steps. Maintainers changing the container-base Node major should update the TOML `tag`, then run `cargo generate installers` so the digest and generated Dockerfile zones refresh together.

- **Fast rollback command/path:** `git revert <sha>` after merge, then rerun `cargo generate installers --check` to confirm generated surfaces return to the prior policy.
- **Observable failure symptoms:** `cargo generate installers --check` reports a Node tag-shape or generated-zone mismatch, or generated Dockerfiles unexpectedly pin a Node major that does not match the TOML tag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.7-labs
 
 # >>> generated:base-arg-node from dev/ci/container-base-images.toml by `cargo generate installers` - do not edit <<<
-ARG ZEROCLAW_BASE_NODE=node:26-bookworm-slim@sha256:4e2e85a824f938e41a61e9e819f0c7c11432f7d60f470b96214d3ead2f0dd63e
+ARG ZEROCLAW_BASE_NODE=node:24-bookworm-slim@sha256:c2d5ade763cacfb03fe9cb8e8af5d1be5041ff331921fa26a9b231ca3a4f780a
 # >>> end generated:base-arg-node <<<
 # >>> generated:base-arg-rust-slim from dev/ci/container-base-images.toml by `cargo generate installers` - do not edit <<<
 ARG ZEROCLAW_BASE_RUST_SLIM=rust:1.94-slim@sha256:cf09adf8c3ebaba10779e5c23ff7fe4df4cccdab8a91f199b0c142c53fef3e1a
--- a/Dockerfile.debian
+++ b/Dockerfile.debian
@@ -1,7 +1,7 @@
 # syntax=docker/dockerfile:1.7-labs
 
 # >>> generated:base-arg-node from dev/ci/container-base-images.toml by `cargo generate installers` - do not edit <<<
-ARG ZEROCLAW_BASE_NODE=node:26-bookworm-slim@sha256:4e2e85a824f938e41a61e9e819f0c7c11432f7d60f470b96214d3ead2f0dd63e
+ARG ZEROCLAW_BASE_NODE=node:24-bookworm-slim@sha256:c2d5ade763cacfb03fe9cb8e8af5d1be5041ff331921fa26a9b231ca3a4f780a
 # >>> end generated:base-arg-node <<<
 # >>> generated:base-arg-rust-bookworm from dev/ci/container-base-images.toml by `cargo generate installers` - do not edit <<<
 ARG ZEROCLAW_BASE_RUST_BOOKWORM=rust:1.94-bookworm@sha256:6ae102bdbf528294bc79ad6e1fae682f6f7c2a6e6621506ba959f9685b308a55
--- a/dev/ci/container-base-images.toml
+++ b/dev/ci/container-base-images.toml
@@ -1,9 +1,10 @@
 # Canonical container base-image pins for the generated container surfaces
 # (Dockerfile, Dockerfile.debian). Edit registry/repo/image_ref/tag here; `tag`
-# and `digest` are rewritten live by `cargo generate installers`. A row with
-# discover=true derives its tag from the repo root .nvmrc, then refreshes its
-# digest live from the registry. StageX pins in the Containerfile are
-# excluded on purpose: digest-only, reproducible-build intent, no tag to follow.
+# records the intended policy and `digest` is rewritten live by `cargo generate
+# installers`. A row with discover=true refreshes the digest for its declared
+# tag from the registry; Node discover rows must use a plain
+# <major>-bookworm-slim tag. StageX pins in the Containerfile are excluded on
+# purpose: digest-only, reproducible-build intent, no tag to follow.
 
 [[image]]
 zone = "base-arg-node"
--- a/dev/ci/container-base-images.toml
+++ b/dev/ci/container-base-images.toml
@@ -1,8 +1,8 @@
 # Canonical container base-image pins for the generated container surfaces
 # (Dockerfile, Dockerfile.debian). Edit registry/repo/image_ref/tag here; `tag`
 # and `digest` are rewritten live by `cargo generate installers`. A row with
-# discover=true resolves its tag live from the registry (node: highest
-# <major>-bookworm-slim on Docker Hub). StageX pins in the Containerfile are
+# discover=true derives its tag from the repo root .nvmrc, then refreshes its
+# digest live from the registry. StageX pins in the Containerfile are
 # excluded on purpose: digest-only, reproducible-build intent, no tag to follow.
 
 [[image]]
@@ -12,8 +12,8 @@ registry = "dockerhub"
 repo = "library/node"
 image_ref = "node"
 discover = true
-tag = "26-bookworm-slim"
-digest = "sha256:4e2e85a824f938e41a61e9e819f0c7c11432f7d60f470b96214d3ead2f0dd63e"
+tag = "24-bookworm-slim"
+digest = "sha256:c2d5ade763cacfb03fe9cb8e8af5d1be5041ff331921fa26a9b231ca3a4f780a"
 
 [[image]]
 zone = "base-arg-rust-slim"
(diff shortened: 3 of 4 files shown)


singlerider:
**Edit**: The following comment was made before the title and apparent intent of the PR was changed:

Recommending closure in favor of #8093.

Both PRs target the container-base maintenance surface, and I'd rather consolidate on the build-from-source approach in #8093 than add the `.nvmrc`-derivation machinery here.

The practical goal both of these serve is "don't let the base container images drift or break." #8093 attacks that directly: it adds `Dockerfile` / `Dockerfile.debian` / `.dockerignore` to the Docker Image PR Check `paths:` filter and a `source-images` job that actually compiles both base Dockerfiles from the repo root on any container-file change. That is the guard that would have caught the stale `COPY --parents crates/aardvark-sys/build.rs ./` line that shipped to master and broke fresh builds (#8092). A real cold build on every container change is a stronger and lower-maintenance invariant than a generator-side tag-policy check.

This PR adds a `validate_node_tag_policy` / `node_major_for_suite` path plus `cargo generate installers --check` validation against `.nvmrc`, which is more moving parts (a second source of truth coupling `.nvmrc` to the discover rows and the generated Dockerfile/TOML) to maintain for a narrower slice of the same "keep the base image sane" problem. Once #8093 is landing a from-source build in CI, a Node major that disagrees with reality surfaces as a build failure rather than needing a bespoke generator assertion.

I don't want two overlapping mechanisms in the same `dev/ci/container-base-images.toml` + Dockerfile generation area. My preference as maintainer is to close this and carry the container-base hardening through #8093. If there's a specific `.nvmrc`-major-pinning behavior you think #8093 won't cover, say so and we can fold the minimal piece of it into that PR instead of running both.

Not a verdict on the code quality here; this is a direction/consolidation call.

singlerider (approved):
Reviewed at head for the discover-row Node policy change. This closes #8105 cleanly: digest refresh stays mechanical, but the Node major is now an explicit maintainer decision in the TOML `tag` rather than `max(major)` off the Docker Hub tag list. Scope is exactly the four surfaces named (xtask generator, the TOML, both Dockerfiles), no `.nvmrc` or release/workflow drift.

### 🟢 What looks good — discovery no longer carries major policy

Dropping `discover_node_tag` and resolving the digest for `img.tag` is the right shape. A routine `cargo generate installers` can no longer silently bump the runtime Node major; that decision now lives in the source row where a maintainer (or a dedicated PR) owns it. Matches the issue's acceptance criteria field-for-field.

### 🟢 What looks good — validation moved to the shared seam

Folding the Node tag-shape check into `validate_node_tag_policy` and calling it from both `refresh_source` and `resolved` is better than the old inline-only check in `check`. The drift path still surfaces a non-plain tag because `resolved` errors and `check` pushes that error into the drift list, so the docstring claim ("node's tag must be a plain-major LTS-suite tag") still holds. `node_major_for_suite` rewrite (strip suite, then strip `-`) is behavior-equivalent to the prior `-{NODE_SUITE}` strip.

### 🟢 What looks good — offline check preserved, reject path tested

`--check` stays network-free; only digest refresh hits the registry, per the non-goal in #8105. `splice_zones_rejects_non_plain_node_discover_tag` pins the rejection path and the 7 focused xtask tests plus the installer drift check cover the changed behavior.

Verdict: APPROVED.

